PR #4831, opened 2023-10-27
HIVE-27827: Improve performance of direct SQL implement for getPartitionsByFilter

wecharyu:
1. Add partition name filter for equals and not-equals in generating partition filter.
2. Add benchmark test `getPartitionsByFilter`.

Performance improvement.

No.

No.

1. Passing existing partition filter related tests.
2. Add benchmark test
```bash
java -jar ./hmsbench-jar-with-dependencies.jar -H localhost --savedata /tmp/benchdata --sanitize -N 10 -N 1000 -N 10000 -o bench_results_direct.csv -C -d testbench_http --params=100  -E 'drop.*' -E 'renameTable.*' -E 'getTableObjectsByName.*' -E 'listTables.*' -E 'listPartitions.*' -E 'getPartitionsByNames.*' -E 'getPartitionNames.*' -E 'listPartition' -E 'getPartition' -E 'getPartitions' -E 'getPartitions.10' -E 'getPartitions.1000' -E 'getPartitions.10000'  -E 'getNid' -E 'listDatabases' -E 'getTable' -E 'createTable' -E 'addPartitions.*' -E 'addPartition.*' -E 'openTxn.*'
```
- Before this patch
```bash
Operation                      Mean     Med      Min      Max      Err%
getPartitionsByFilter          4.522    4.461    4.160    6.563    6.850
getPartitionsByFilter.10       4.371    4.305    4.035    6.364    6.456
getPartitionsByFilter.1000     4.403    4.376    4.118    5.145    3.958
getPartitionsByFilter.10000    6.044    6.006    5.663    9.304    6.237
```

- After this patch
```bash
Operation                      Mean     Med      Min      Max      Err%
getPartitionsByFilter          4.886    4.864    4.271    6.641    7.648
getPartitionsByFilter.10       4.779    4.803    4.121    5.578    7.569
getPartitionsByFilter.1000     4.315    4.208    3.976    5.047    6.571
getPartitionsByFilter.10000    4.333    4.127    3.783    9.149    16.65
```


ql/src/test/queries/clientpositive/partition_coltype_literals.q
```diff
@@ -73,11 +73,11 @@ drop table if exists partcoltypeothers;
 create table partcoltypeothers (key int, value string) partitioned by (decpart decimal(6,2), datepart date);
 
 set hive.typecheck.on.insert=false;
-insert into partcoltypeothers partition (decpart = 1000.01BD, datepart = date '2015-4-13') select key, value from src limit 10;
+insert into partcoltypeothers partition (decpart = 1000.01BD, datepart = date '2015-04-13') select key, value from src limit 10;
```
wecharyu, line 76: When disable `hive.typecheck.on.insert`, we need ensure the partition value is in right format.

wecharyu:
@ayushtkn @saihemanth-cloudera @deniskuzZ  could you review this PR?

ql/src/test/queries/clientpositive/partition_coltype_literals.q
saihemanth-cloudera: I don't agree with this change. This would be a breaking change if users were to use this patch in their production env (because they would now have to change their scripts/data). We intentionally specified this config 'hive.typecheck.on.insert=false' so that regardless of data format, partitions are created the way the data is inserted.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -2689,11 +2690,32 @@ private void addPartitionsInternal(String catName, String dbName,
     List<MPartition> mParts = new ArrayList<>();
     List<List<MPartitionPrivilege>> mPartPrivilegesList = new ArrayList<>();
     List<List<MPartitionColumnPrivilege>> mPartColPrivilegesList = new ArrayList<>();
+    List<MFieldSchema> partitionKeys = table.getPartitionKeys();
+    int[] dateIndices = IntStream.range(0, partitionKeys.size())
+            .filter(i -> partitionKeys.get(i).getType().equalsIgnoreCase(ColumnType.DATE_TYPE_NAME))
```
saihemanth-cloudera, line 2695: You have to validate partition column types only if this config is set 'HIVE_TYPE_CHECK_ON_INSERT'.
See this for reference: https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java#L1719

wecharyu: This config `HIVE_TYPE_CHECK_ON_INSERT` seems a client conf to correct the user input.

The reason why we want to check `date` value format is that HMS query `date` value in format of `uuuu-MM-dd` strictly:
https://github.com/apache/hive/blob/5e6ce2a66404ef0267c27f407f14e601e566dfc0/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L1368-L1371
And many engines like Spark will guarantee the  `date` value in format of `uuuu-MM-dd`, which is also important when use date related udf on such partition column.

ql/src/test/queries/clientpositive/partition_coltype_literals.q
wecharyu: The `hive.typecheck.on.insert` config is set to `true` by default, meaning that this change won't impact the most tasks. If users utilizes the date`2015-5-15` and also set `hive.typecheck.on.insert` to false, they can conveniently enable this config to resolve the issue. This behavior aligns with the config description.So I think this change is OK.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
wecharyu: @saihemanth-cloudera After a double thought, we think it's not good to change the partition values in `add_partitions`. 
The original motivation to do it was that the unit test `org.apache.hive.hcatalog.pig.AbstractHCatLoaderTest#testDatePartitionPushUp` failed after this patch:
Pig engine add partition values with time data like `dt=2016-07-14 15%3A10%3A15`, and then the query by dt cast data to 'yyyy-MM-dd' since [HIVE-14282](https://issues.apache.org/jira/browse/HIVE-14282), then it could not match PART_NAME and return empty result.
This test will fail in JDO query too if we disable direct sql because JDO query also use PART_NAME filter to speed up query.

To solve this issue, we could add more rules to match PART_NAME, that is if the partition type is `date`, then we use `PART_NAME like dt=2023-10-10%` instead of `PART_NAME = 2023-10-10`.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1336,7 +1337,8 @@ public void visit(LeafNode node) throws MetaException {
         return;
       }
 
-      String colTypeStr = ColumnType.getTypeName(partitionKeys.get(partColIndex).getType());
+      FieldSchema partCol = partitionKeys.get(partColIndex);
```
saihemanth-cloudera, line 1340: I don't think this variable is not being used anywhere else except L#1341, so I think it is ok to keep L1339 without any changes.

wecharyu:
@deniskuzZ @zhangbutao could you also take a look on this PR?

wecharyu:
@nrg4878 @saihemanth-cloudera Updated the indexes in `PARTITIONS` table for the performance improvement:
1. update `UNIQUEPARTITION (PART_NAME, TBL_ID)` -> `UNIQUEPARTITION (TBL_ID, PART_NAME)`
2. remove `PARTITIONS_N49 (TBL_ID)`

saihemanth-cloudera (commented):
Agree with changing the partition index order.
Changes looks good to me. But the test failures seem to be related. Can you please check them?
Also, can you verify/post the output of explain plan for the following queries in Mysql DB?
`explain select * from PARTITIONS where TBL_ID=123 and PART_ID=123 and PART_NAME>8;` or some queries involving Partitions table

wecharyu:
Test failure seems unrelated,  rebase to mater to get green CI.
And add some query case of PARTITIONS in MySQL:
```sql
-- Case 1: with TBL_ID and PART_ID (hit PRIMARY index)
mysql> explain select * from PARTITIONS where TBL_ID=1 and PART_ID=1 and PART_NAME like 'p=1%';
+----+-------------+------------+------------+-------+-------------------------+---------+---------+-------+------+----------+-------+
| id | select_type | table      | partitions | type  | possible_keys           | key     | key_len | ref   | rows | filtered | Extra |
+----+-------------+------------+------------+-------+-------------------------+---------+---------+-------+------+----------+-------+
|  1 | SIMPLE      | PARTITIONS | NULL       | const | PRIMARY,UNIQUEPARTITION | PRIMARY | 8       | const |    1 |   100.00 | NULL  |
+----+-------------+------------+------------+-------+-------------------------+---------+---------+-------+------+----------+-------+

-- Case 2: with TBL_ID and partial partition values (hit UNIQUEPARTITION index)
mysql> explain select * from PARTITIONS where TBL_ID=1 and PART_NAME like 'p=1%';
+----+-------------+------------+------------+-------+-----------------+-----------------+---------+------+------+----------+-----------------------+
| id | select_type | table      | partitions | type  | possible_keys   | key             | key_len | ref  | rows | filtered | Extra                 |
+----+-------------+------------+------------+-------+-----------------+-----------------+---------+------+------+----------+-----------------------+
|  1 | SIMPLE      | PARTITIONS | NULL       | range | UNIQUEPARTITION | UNIQUEPARTITION | 779     | NULL |    2 |   100.00 | Using index condition |
+----+-------------+------------+------------+-------+-----------------+-----------------+---------+------+------+----------+-----------------------+

-- Case 3: with only TBL_ID (hit the prefix index of UNIQUEPARTITION)
mysql> explain select * from PARTITIONS where TBL_ID=1;
+----+-------------+------------+------------+------+-----------------+-----------------+---------+-------+------+----------+-------+
| id | select_type | table      | partitions | type | possible_keys   | key             | key_len | ref   | rows | filtered | Extra |
+----+-------------+------------+------------+------+-----------------+-----------------+---------+-------+------+----------+-------+
|  1 | SIMPLE      | PARTITIONS | NULL       | ref  | UNIQUEPARTITION | UNIQUEPARTITION | 9       | const |   10 |   100.00 | NULL  |
+----+-------------+------------+------------+------+-----------------+-----------------+---------+-------+------+----------+-------+

-- Case 4: with only PART_NAME (no index hit, it's acceptable because in the real calls we always provide the TBL_ID for partition query)
mysql> explain select * from PARTITIONS where PART_NAME like 'p=1%';
+----+-------------+------------+------------+------+---------------+------+---------+------+------+----------+-------------+
| id | select_type | table      | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra       |
+----+-------------+------------+------------+------+---------------+------+---------+------+------+----------+-------------+
|  1 | SIMPLE      | PARTITIONS | NULL       | ALL  | NULL          | NULL | NULL    | NULL |   10 |    11.11 | Using where |
+----+-------------+------------+------------+------+---------------+------+---------+------+------+----------+-------------+
```

saihemanth-cloudera (approved):
The Changes look good to me. 
Thanks for the explain plan, Unfortunately, the primary index (Instead of UniquePartition) is being used in Case 1 which is used most often.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1520,10 +1523,43 @@ public void visit(LeafNode node) throws MetaException {
...
+        } else {
+          if (partColCount == 1) {
+            // Case where partition column type is date and there is no other partition columns
+            params.add(escapedNameFragment);
+          } else if (partColIndex + 1 == partColCount) {
+            // Case where the partition column is at the end of the name.
+            params.add("%/" + escapedNameFragment);
+          } else if (partColIndex == 0) {
+            // Case where the partition column is at the beginning of the name.
+            params.add(escapedNameFragment + "/%");
```
cxzl25, line 1553: Thank you for your contribution. We also made similar optimizations last year, which greatly reduced the load of MySQL.

However, I suggest `params.add(escapedNameFragment + "/%");` This can be transformed into a multi-level partition filter, which can filter more data when scanning.

Now it looks like this
```
d=xxx/%
```

image


Advanced optimization
```
d=xxx/h=_%/b=_%
```

image

---

If you think this is a good idea, I will create a PR.  @wecharyu @saihemanth-cloudera

PR #4832, opened 2023-10-27
HIVE-27828: Iceberg integration: enable copy on write update when split update is on

kasakrisz:
* Split Update and Delete into two semantic analyzers.
* Split each type of update, delete and merge rewrites into separate classes.
* Move the decision of which implementation to use from `SemanticAnalyzerFactory` to factory classes. The decision is based on the setting `hive.split.update`, table type (native or non-native acid) and storage handler.
* Remove deprecated setting `hive.merge.split.update`

Copy-on-write mode should be independent from split update because split update uses positional delete.

No.

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestIcebergCliDriver -Dqfile=update_iceberg_copy_on_write_partitioned.q,update_iceberg_copy_on_write_unpartitioned.q -pl itests/qtest-iceberg -Piceberg -Pitests
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=split_update.q,sort_acid.q,sqlmerge.q -pl itests/qtest -Pitests
```


ql/src/java/org/apache/hadoop/hive/ql/hooks/WriteEntity.java
```diff
@@ -170,7 +170,7 @@ public WriteType getWriteType() {
 
   /**
    * Only use this if you are very sure of what you are doing.  This is used by the
-   * {@link org.apache.hadoop.hive.ql.parse.UpdateDeleteSemanticAnalyzer} to reset the types to
+   * {@link org.apache.hadoop.hive.ql.parse.UpdateSemanticAnalyzer} to reset the types to
```
deniskuzZ, line 173: should we mention DeleteSemanticAnalyzer here as well?

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -444,7 +444,7 @@ default List<VirtualColumn> acidVirtualColumns() {
   }
 
   /**
-   * {@link org.apache.hadoop.hive.ql.parse.UpdateDeleteSemanticAnalyzer} rewrites DELETE/UPDATE queries into INSERT
+   * {@link UpdateSemanticAnalyzer} rewrites DELETE/UPDATE queries into INSERT
```
deniskuzZ, line 447: same, check javadoc refs for UpdateSemanticAnalyzer and mention Delete as well

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
```diff
@@ -0,0 +1,165 @@
...
+    ASTNode where = null;
+    if (children.size() > 1) {
+      where = (ASTNode) children.get(1);
+      assert where.getToken().getType() == HiveParser.TOK_WHERE :
+          "Expected where clause, but found " + where.getName();
+    }
+
+    boolean shouldTruncate = HiveConf.getBoolVar(conf, HiveConf.ConfVars.HIVE_OPTIMIZE_REPLACE_DELETE_WITH_TRUNCATE)
+        && where == null;
+    if (shouldTruncate) {
```
deniskuzZ, line 73: could we create List<Predicate<>> for diff optimizations iterate and call test(), if that returns true return, otherwise go to next?

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    Rewriter<DeleteStatement> rewriter = rewriterFactory.createRewriter(table, getFullTableNameForSQL(tableName));
+    ParseUtils.ReparseResult rr = rewriter.rewrite(ctx, new DeleteStatement(table, where));
+
+    Context rewrittenCtx = rr.rewrittenCtx;
+    ASTNode rewrittenTree = rr.rewrittenTree;
+
+    analyzeRewrittenTree(rewrittenTree, rewrittenCtx);
+
```
deniskuzZ, line 87: redunant new line

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
```diff
@@ -0,0 +1,165 @@
...
+      return false;
+    }
+    ConvertAstToSearchArg.Result result = ConvertAstToSearchArg.createSearchArgument(ctx.getConf(), hiveFilter);
+    if (result.isPartial()) {
+      return false;
+    }
+    SearchArgument sarg = result.getSearchArgument();
+    if (!table.getStorageHandler().canPerformMetadataDelete(table, tableName.getTableMetaRef(), sarg)) {
+      return false;
+    }
```
deniskuzZ, line 146: could we extract this to getSarg method that returns an Optional?

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -17,40 +17,31 @@
...
- * they are actually inserts) and then doing some patch up to make them work as merges instead.
- */
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
```
deniskuzZ, line 40: we need some javadoc

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -188,23 +163,23 @@ WHEN NOT MATCHED THEN INSERT VALUES (source.a2, source.b2)
       switch (getWhenClauseOperation(whenClause).getType()) {
       case HiveParser.TOK_INSERT:
         numInsertClauses++;
-        handleInsert(whenClause, rewrittenQueryStr, targetNameNode, onClause,
+        handleInsert(whenClause, mergeRewriter, targetNameNode, onClause,
```
deniskuzZ, line 166: i think we need to extract the whole merge logic. 
check Merge COW patch, I had to pass 3 extra params in every handle method + create extra deleteCow that basically joins deletes from Update and Delete branches

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -226,17 +201,26 @@ WHEN NOT MATCHED THEN INSERT VALUES (source.a2, source.b2)
       throw new SemanticException(ErrorMsg.MERGE_PREDIACTE_REQUIRED, ctx.getCmd());
     }
 
-    boolean validating = handleCardinalityViolation(rewrittenQueryStr, targetNameNode, onClauseAsText, targetTable,
-        numWhenMatchedDeleteClauses == 0 && numWhenMatchedUpdateClauses == 0, columnAppender);
-    ReparseResult rr = ParseUtils.parseRewrittenQuery(ctx, rewrittenQueryStr);
+    if (!conf.getBoolVar(HiveConf.ConfVars.MERGE_CARDINALITY_VIOLATION_CHECK)) {
```
deniskuzZ, line 204: if that is false, i don't think we should invoke handleCardinalityViolation, WDYT?

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -264,27 +248,16 @@ WHEN NOT MATCHED THEN INSERT VALUES (source.a2, source.b2)
...
-  protected int addDestNamePrefixOfUpdate(int insClauseIdx, Context rewrittenCtx) {
-    rewrittenCtx.addDestNamePrefix(insClauseIdx, Context.DestClausePrefix.UPDATE);
-    return 1;
-  }
-
   /**
    * If there is no WHEN NOT MATCHED THEN INSERT, we don't outer join.
    */
-  private String chooseJoinType(List<ASTNode> whenClauses) {
+  private boolean hasWhenNotMatchedInsertClause(List<ASTNode> whenClauses) {
```
deniskuzZ, line 254: COW would be using FULL or LEFT JOIN, so that should be moved to rewriter

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -295,72 +268,27 @@ private String chooseJoinType(List<ASTNode> whenClauses) {
    * cardinality_violation() UDF throws an error whenever it's called killing the query
    * @return true if another Insert clause was added
    */
-  private boolean handleCardinalityViolation(StringBuilder rewrittenQueryStr, ASTNode target,
-      String onClauseAsString, Table targetTable, boolean onlyHaveWhenNotMatchedClause, ColumnAppender columnAppender)
-              throws SemanticException {
+  private boolean handleCardinalityViolation(boolean onlyHaveWhenNotMatchedClause) {
```
deniskuzZ, line 271: for COW that would be "AND EXISTS (SELECT ...)"

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -723,11 +588,88 @@ private String getPredicate() {
...
+
+  /**
+   * This allows us to take an arbitrary ASTNode and turn it back into SQL that produced it.
+   * Since HiveLexer.g is written such that it strips away any ` (back ticks) around
+   * quoted identifiers we need to add those back to generated SQL.
+   * Additionally, the parser only produces tokens of type Identifier and never
+   * QuotedIdentifier (HIVE-6013).  So here we just quote all identifiers.
+   * (') around String literals are retained w/o issues
+   */
+  private static class IdentifierQuoter {
```
deniskuzZ, line 607: should it be a standalone util class?

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -142,31 +125,23 @@ WHEN NOT MATCHED THEN INSERT VALUES (source.a2, source.b2)
...
-    ColumnAppender columnAppender = getColumnAppender(subQueryAlias, StringUtils.EMPTY);
-    columnAppender.appendAcidSelectColumns(rewrittenQueryStr, Context.Operation.MERGE);
-
-    rewrittenQueryStr.deleteCharAt(rewrittenQueryStr.length() - 1); // remove last ','
-    addColsToSelect(targetTable.getCols(), rewrittenQueryStr);
-    addColsToSelect(targetTable.getPartCols(), rewrittenQueryStr);
-    rewrittenQueryStr.append(" FROM ").append(getFullTableNameForSQL(targetNameNode)).append(") ");
-    rewrittenQueryStr.append(subQueryAlias);
-    rewrittenQueryStr.append('\n');
+    MergeRewriter mergeRewriter = new MergeRewriterFactory(conf).createMergeRewriter(
```
deniskuzZ, line 129: should we create a new factory instance every time analyze is called or init it in constructor where we set the conf obj?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzerFactory.java
```diff
@@ -94,17 +96,11 @@ private static BaseSemanticAnalyzer getInternal(QueryState queryState, ASTNode t
...
       case HiveParser.TOK_DELETE_FROM:
-        return new UpdateDeleteSemanticAnalyzer(queryState);
+        return new DeleteSemanticAnalyzer(queryState, new DeleteRewriterFactory(queryState.getConf()));
 
       case HiveParser.TOK_MERGE:
-        if (HiveConf.getBoolVar(queryState.getConf(), HiveConf.ConfVars.SPLIT_UPDATE) ||
-                HiveConf.getBoolVar(queryState.getConf(), HiveConf.ConfVars.MERGE_SPLIT_UPDATE)) {
-          return new SplitMergeSemanticAnalyzer(queryState);
-        }
         return new MergeSemanticAnalyzer(queryState);
```
deniskuzZ, line 104: dup of pev comment. why don't we follow the same approach as for Update/Delete -> pass or init in constructor?

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteDeleteRewriter.java
```diff
@@ -0,0 +1,85 @@
...
+    this.conf = conf;
+    this.cowWithClauseBuilder = cowWithClauseBuilder;
+  }
+
+  @Override
+  public ParseUtils.ReparseResult rewrite(Context context, DeleteStatement deleteBlock)
+      throws SemanticException {
+
+    String whereClause = context.getTokenRewriteStream().toString(
+        deleteBlock.getWhereTree().getChild(0).getTokenStartIndex(),
```
deniskuzZ, line 47: should we extract `deleteBlock.getWhereTree().getChild(0)` to localVar?

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteDeleteRewriter.java
```diff
@@ -0,0 +1,85 @@
...
+        deleteBlock.getWhereTree().getChild(0).getTokenStopIndex());
+    String filePathCol = HiveUtils.unparseIdentifier("FILE__PATH", conf);
```
deniskuzZ, line 49: my bad, could we ref the constant: VirtualColumn.FILE_PATH.name()

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
```diff
@@ -0,0 +1,132 @@
...
+    this.cowWithClauseBuilder = cowWithClauseBuilder;
+    this.setClausePatcher = setClausePatcher;
+  }
+
+  @Override
+  public ParseUtils.ReparseResult rewrite(Context context, UpdateStatement updateBlock)
+      throws SemanticException {
+
+    String whereClause = context.getTokenRewriteStream().toString(
+        updateBlock.getWhereTree().getChild(0).getTokenStartIndex(),
```
deniskuzZ, line 59: same reg local var

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
```diff
@@ -0,0 +1,132 @@
...
+        updateBlock.getWhereTree().getChild(0).getTokenStopIndex());
+    String filePathCol = HiveUtils.unparseIdentifier("FILE__PATH", conf);
```
deniskuzZ, line 61: dup, use constant

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
```diff
@@ -0,0 +1,132 @@
...
+        // later and patch it up.
+        // Add one to the index because the select has the ROW__ID as the first column.
+        setColExprs.put(columnOffset + i, setCol);
+      }
+    }
+
+    sqlBuilder.append(" from ");
+    sqlBuilder.appendTargetTableName();
+
+    if (updateBlock.getWhereTree() != null) {
```
deniskuzZ, line 96: isn't this redundant check since the above code doesn't do any checks when constructing `whereClause`

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/DeleteRewriter.java
```diff
@@ -0,0 +1,71 @@
...
+    sqlBuilder.append(sqlBuilder.getTargetTableFullName());
+    sqlBuilder.appendPartitionColsOfTarget();
+
+    sqlBuilder.append(" select ");
+    sqlBuilder.appendAcidSelectColumns(Context.Operation.DELETE);
+    sqlBuilder.removeLastChar();
+    sqlBuilder.append(" from ");
+    sqlBuilder.append(sqlBuilder.getTargetTableFullName());
+
+    sqlBuilder.appendSortBy(sqlBuilder.getSortKeys());
```
deniskuzZ, line 53: not needed in case of COW

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.hadoop.hive.ql.metadata.Table;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+
+import java.util.Map;
+
+public interface MergeRewriter {
+  void handleSource(boolean hasWhenNotMatchedClause, String sourceAlias, String onClauseAsText);
+
+  void handleWhenNotMatchedInsert(String targetFullName, String columnListText, String hintStr,
+                                  String valuesClause, String predicate, String extraPredicate);
```
deniskuzZ, line 32: is that standard formatting?

ql/src/java/org/apache/hadoop/hive/ql/Context.java
```diff
@@ -200,6 +201,11 @@ public void setOperation(Operation operation) {
     this.operation = operation;
   }
 
+  public void setOperation(Operation operation, boolean splitUpdate) {
```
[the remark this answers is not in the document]
kasakrisz: Copy-on-write and "default" is the same from operation standpoint. I would introduce enum only when it is necessary. I was also thinking about treating split update as merge in a follow-up patch.

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
kasakrisz: Updated.

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
kasakrisz: New lines are intentional here. I wanted `analyzeRewrittenTree(rewrittenTree, rewrittenCtx);` call highlighted.

ql/src/java/org/apache/hadoop/hive/ql/parse/ParseUtils.java
```diff
@@ -675,11 +675,17 @@ public static String stripIdentifierQuotes(String val) {
     return val;
   }
 
+  public static ReparseResult parseRewrittenQuery(Context ctx,
+                                                  StringBuilder rewrittenQueryStr)
```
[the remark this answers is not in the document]
kasakrisz: done

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: added

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: I removed `MergeRewriter` interface and implemented `Rewriter` instead and added `MergeStatement` to store the parsed parts of merge.

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: Moved this to MergeStatement

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: This class doesn't worthy for general purposes. This way we can not quote ids which has backtick:
```
        visitedNodes.put(n, n);
        trs.insertBefore(n.getToken(), "`");
        trs.insertAfter(n.getToken(), "`");
```

So this class must be refactored in the future when we want to add full support of quoted Identifiers in acid tables. To do it we need something like this
https://github.com/apache/hive/blob/cbc5d2d7d650f90882c5c4ad0026a94d2e586acb/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/constraint/ConstraintsUtils.java#L313C36-L345

Currently one of the blockers is `ValidReadTxnList` uses characters `:` and `$` as seprarator characters in the string representation. According to standard these can be part of quoted identifier hence a table name.
https://github.com/apache/hive/blob/cbc5d2d7d650f90882c5c4ad0026a94d2e586acb/storage-api/src/java/org/apache/hadoop/hive/common/ValidReadTxnList.java#L190

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: Moved factory creation to SemanticAnalyzerFactory

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteDeleteRewriter.java
kasakrisz: done

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteDeleteRewriter.java
kasakrisz: Changed.

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
kasakrisz: Could you please share the code style of Hive. I saw formatting of many ways in so far

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
kasakrisz: I think MetadataUpdate doesn't fall into that pattern easily. It requires a more generic flow something like:
init -> canHandlePreParse -> parse -> canHandleAftParse -> rewrite -> canHandleAftRewrite -> compile

It can be done in a follow-up patch.

deniskuzZ: ok

ql/src/java/org/apache/hadoop/hive/ql/Context.java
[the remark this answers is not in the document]
deniskuzZ: ok

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
deniskuzZ: move sarg construction into a method - Optiional< SearchArgument > getDeleteSarg(....)
````
if(!sarg.IsPresent()) {
    return
}
````

File: ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java (hunk 11 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java:541]: can we mark is as final?

File: ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java (hunk 9 above)
Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java]: Implemented Rewriter instead and added MergeStatement to store the parsed parts of merge. COWMergeRewiter should implement its own join type selection logic

File: ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java (hunk 10 above)
Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java]: Implemented Rewriter instead and added MergeStatement to store the parsed parts of merge. COWMergeRewiter should implement its own cardinality check logic

File: ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzerFactory.java (hunk 13 above)
Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzerFactory.java]: Added MergeRewriter

File: ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java (hunk 17 above)
Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java]: done

File: ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java (hunk 16 above)
Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java]: Currently Update with empty where clause is not supported. We need to implement it in a follow-up patch.

File: ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java [hunk 23]
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.ql.Context;
+import org.apache.hadoop.hive.ql.metadata.Hive;
+import org.apache.hadoop.hive.ql.metadata.Table;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+
+import java.util.Map;
+
+public interface MergeRewriter {
```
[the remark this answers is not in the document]
kasakrisz: Added MergeRewriter

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
kasakrisz: Extracted these checks and the Filter predicate conversion to a method. It returns `null` if the the conversion fails. It could return `Optional` but IMHO it is just a matter of coding style. I don's see much difference between a null check and calling `isPresent()`. Optional `sarg` can also be `null` and throw NPE though.

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: done

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
```diff
@@ -0,0 +1,166 @@
...
+    ASTNode where = null;
+    if (children.size() > 1) {
+      where = (ASTNode) children.get(1);
+      assert where.getToken().getType() == HiveParser.TOK_WHERE :
+          "Expected where clause, but found " + where.getName();
+    }
+
+    boolean shouldTruncate = HiveConf.getBoolVar(conf, HiveConf.ConfVars.HIVE_OPTIMIZE_REPLACE_DELETE_WITH_TRUNCATE)
+        && where == null;
+    if (shouldTruncate) {
```
deniskuzZ, line 69: maybe `rewriteUsingTruncate`

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
```diff
@@ -0,0 +1,166 @@
...
+      genTruncatePlan(table, tableName);
+      return;
+    } else if (tryMetadataUpdate(table, tableName, where)) {
```
deniskuzZ, line 72: remove else and rename to `rewriteUsingMetadataUpdate`

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -723,11 +513,88 @@ private String getPredicate() {
...
+
+  /**
+   * This allows us to take an arbitrary ASTNode and turn it back into SQL that produced it.
+   * Since HiveLexer.g is written such that it strips away any ` (back ticks) around
+   * quoted identifiers we need to add those back to generated SQL.
+   * Additionally, the parser only produces tokens of type Identifier and never
+   * QuotedIdentifier (HIVE-6013).  So here we just quote all identifiers.
+   * (') around String literals are retained w/o issues
+   */
+  private static final class IdentifierQuoter {
```
deniskuzZ, line 532: maybe move it to `RewriteSemanticAnalyzer `

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
```diff
@@ -0,0 +1,274 @@
...
+
+  public MergeRewriter(Hive db, HiveConf conf, SqlGeneratorFactory sqlGeneratorFactory) {
+    this.db = db;
+    this.conf = conf;
+    this.sqlGeneratorFactory = sqlGeneratorFactory;
+  }
+
+  @Override
+  public ParseUtils.ReparseResult rewrite(Context ctx, MergeStatement mergeStatement) throws SemanticException {
+
```
deniskuzZ, line 60: nit: new line

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  protected MergeWhenClauseSqlGenerator createMergeSqlGenerator(
+      MergeStatement mergeStatement, MultiInsertSqlGenerator sqlGenerator) {
+    return new MergeWhenClauseSqlGenerator(conf, sqlGenerator, mergeStatement);
+  }
+
+  private void handleSource(boolean hasWhenNotMatchedClause, String sourceAlias, String onClauseAsText,
+                            MultiInsertSqlGenerator sqlGenerator) {
+    sqlGenerator.append("FROM\n");
```
deniskuzZ, line 106: that won't work for COW: see https://github.com/apache/hive/pull/4852/files#diff-5612014abb89e13c290474798741828a8933edda57f9035f0656df47c4b63e7bR471

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeStatement.java
```diff
@@ -0,0 +1,291 @@
...
+
+  public interface MergeSqlGenerator {
+    void appendWhenNotMatchedInsertClause(InsertClause insertClause);
+    void appendWhenMatchedUpdateClause(UpdateClause updateClause);
+    void appendWhenMatchedDeleteClause(DeleteClause deleteClause);
+  }
+
+  public interface DestClausePrefixSupplier {
+    default List<Context.DestClausePrefix> getInsertDestClausePrefixes()
+    {
```
deniskuzZ, line 279: formatting: brackets

ql/src/java/org/apache/hadoop/hive/ql/parse/DeleteSemanticAnalyzer.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
```diff
@@ -0,0 +1,274 @@
...
+
+    @Override
+    public void appendWhenMatchedDeleteClause(MergeStatement.DeleteClause deleteClause) {
+      handleWhenMatchedDelete(mergeStatement.getOnClauseAsText(),
+          deleteClause.getExtraPredicate(), deleteClause.getUpdateExtraPredicate(), hintStr, sqlGenerator);
+      hintStr = null;
+    }
+
+    protected void handleWhenMatchedDelete(String onClauseAsString, String extraPredicate, String updateExtraPredicate,
+                                         String hintStr, MultiInsertSqlGenerator sqlGenerator) {
```
deniskuzZ, line 259: doesn't seem aligned

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
deniskuzZ: i have some style file from downstream :), so it probably doesn't apply here. I usually use 4 spaces(2 tabs) for params that shifted to a next line

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeStatement.java
kasakrisz: fixed

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
kasakrisz: removed

deniskuzZ (approved):
LGTM +1, pending tests

PR #4835, opened 2023-10-30
HIVE-27789: Iceberg: Add a way to expire snapshots with retain last.

ayushtkn:

Allow iceberg table expiry with retain last

Better usability

Yes, Iceberg snapshots can be expired using retain last

No

UT

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/AlterTableExecuteSpec.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/AlterTableExecuteSpec.java
@@ -113,9 +113,12 @@ public class AlterTableExecuteSpec<T> {
 
     private long fromTimestampMillis = -1L;
 
+    private int numRetainLast = -1;
+
     public ExpireSnapshotsSpec(long timestampMillis) {
       this.timestampMillis = timestampMillis;
     }
+
 
     public ExpireSnapshotsSpec(String ids) {
       this.idsToExpire = ids.split(",");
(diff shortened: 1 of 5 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/AlterTableExecuteSpec.java
```diff
@@ -146,13 +157,19 @@ public boolean isExpireByTimestampRange() {
       return timestampMillis != -1 && fromTimestampMillis != -1;
     }
 
+    public boolean isExpireByRetainLast() {
+      return numRetainLast != -1;
```
zhangbutao, line 161: What happened if `numRetainLast` <-1, e.g. -2
We can check this value must be >= 1 in somewhere.

ayushtkn: There is a check here:
https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/RemoveSnapshots.java#L134-L137

ql/src/java/org/apache/hadoop/hive/ql/parse/AlterTableExecuteSpec.java
```diff
@@ -113,10 +113,13 @@ public static class ExpireSnapshotsSpec {
 
     private long fromTimestampMillis = -1L;
 
+    private int numRetainLast = -1;
+
     public ExpireSnapshotsSpec(long timestampMillis) {
       this.timestampMillis = timestampMillis;
     }
 
+
```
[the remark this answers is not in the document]
ayushtkn: Done

deniskuzZ (approved):
LGTM +1

ql/src/java/org/apache/hadoop/hive/ql/parse/AlterTableExecuteSpec.java
deniskuzZ: shouldn't it be
````
return numSnapshots >= 1
````

ayushtkn: discussed offline, we are sorted

PR #4837, opened 2023-10-30
HIVE-27533: Incorrect FOREIGN KEY constraints in SHOW CREATE TABLE

soumyakanti3578:

Currently SHOW CREATE TABLE and EXPLAIN DDL outputs are incorrect when a table has a FK constraint on a composite key. An example is shown in the Jira: https://issues.apache.org/jira/browse/HIVE-27533

Multiple ALTER TABLE statements are printed for each column in the constraint, which is a bug.

FOREIGN KEY constraints will print correctly now.

No

`mvn test -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile=show_create_table.q`

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/DDLPlanUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/DDLPlanUtils.java
@@ -75,8 +75,10 @@ import org.stringtemplate.v4.ST;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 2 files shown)


ql/src/java/org/apache/hadoop/hive/ql/exec/DDLPlanUtils.java
```diff
@@ -691,20 +691,26 @@ public void getAlterTableStmtForeignKeyConstraint(ForeignKeyInfo fr, List<String
...
-        command.add(DATABASE_NAME_FR, fkc.parentDatabaseName);
-        command.add(PARENT_TABLE_NAME, fkc.parentTableName);
-        command.add(PARENT_COL_NAME, fkc.parentColName);
-        command.add(ENABLE, fkc.enable);
-        command.add(VALIDATE, fkc.validate);
-        command.add(RELY, fkc.rely);
-        constraints.add(command.render());
+      List<ForeignKeyInfo.ForeignKeyCol> fkCols = all.get(key);
+      List<String> parentCols = new ArrayList<>();
+      List<String> childCols = new ArrayList<>();
```
kasakrisz, line 696: How about using `LinkedHashSet<String>(fkCols.size())` since column names should be unique.

kasakrisz (commented):
Good catch. Left a minor comment.

aturoczy (approved):

PR #4475, opened 2023-07-10
HIVE-27487: NPE in Hive JDBC storage handler

dengzhhu653:
### Why are the changes needed?
When deserializing the row read from Hive JDBC storage handler, in case if the column names of the row are full qualified, then it's unable to find the matching value from the row by the registered column, npe would happen:
https://github.com/apache/hive/blob/5e46e80bc7d059093aece81e3886ba5ee425ee95/jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/JdbcSerDe.java#L228

### Does this PR introduce _any_ user-facing change?
No

### Is the change a dependency upgrade?
No

### How was this patch tested?
Able to run the query against a table based on hive jdbc storage handler in the testing env.

PR #4839, opened 2023-10-31
HIVE-25379 : Fix external_jdbc_table.q and external_jdbc_table_perf.q

rkirtir:
HIVE-25379 : Fix external_jdbc_table.q and external_jdbc_table_perf.q

Both q files needs fix

NO

NO

with q file and flakey job

http://ci.hive.apache.org/job/hive-flaky-check/750/
http://ci.hive.apache.org/job/hive-flaky-check/746/


aturoczy (approved):
You mainly here just enabled the test right? Did you find the root cause of it?

rkirtir:
Thanks for review. I have fixed  external_jdbc_table_perf.q by overrriding q.out file as it has now prehook properties and enabled  external_jdbc_table.q which fails here but not on flakey test job :). I will get green build by fixing it.

ayushtkn (changes requested):
The test being fixed is failing :
http://ci.hive.apache.org/job/hive-precommit/job/PR-4839/5/testReport/org.apache.hadoop.hive.cli.split8/TestMiniLlapLocalCliDriver/Testing___split_19___PostProcess___testCliDriver_external_jdbc_table_/

We should not just enable the tests but figure out the reason for previous failure as well, the tests were earlier failing with 
```
Caused by: MetaException(message:org.apache.hadoop.hive.serde2.SerDeException Caught exception while initializing the SqlSerDe)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreUtils.getDeserializer(HiveMetaStoreUtils.java:94)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreUtils.getDeserializer(HiveMetaStoreUtils.java:77)
	at org.apache.hadoop.hive.ql.metadata.Table.getDeserializerFromMetaStore(Table.java:323)
```
What fixed it? There is no prod change in this PR

rkirtir:
sorry for late reply!! The root cause has been fixed in below PR

NPE in JDBC storage handler has been fixed with https://issues.apache.org/jira/browse/HIVE-27487
https://github.com/apache/hive/pull/4475
I have reproduced it reverting above commit and then able to get stable results with pull/4475 on local.

This test is failing with assertion difference on jenkins. Will debug and fix it

PR #4841, opened 2023-10-31
HIVE-27833: Hive Acid Replication Support for Dell Powerscale

harshal-16:
Details:
* Isilon supports xAttr only on /.reserved/raw path. Because of that checkFileSystemXAttrSupport throws AccessControlException
* In case of HDFS encryption zone, a virtual path /.reserved/raw gives superusers direct access to underlying block data in filesystem.
This allows superusers to DistCp data without requiring access to encryption keys, and avoids the overhead of decrypting and re-encrypting data.
Isilon system only supports getXAttrs on /.reserved/raw paths

* Added Additional property dfs.xattr.supported.only.on.reserved.namespace which will be true in case of isilon
* Modified logic of checkFileSystemXAttrSupport to return false in case of isilon if path doesn't start with /.reserved/raw

* Isilon only supports xattr on /.reserved/raw path else it throws AccessControlException

* From User perspective flow will be same 

* Tested on Cluster
* Added Test case

### DIFF
--- a/shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
+++ b/shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
@@ -49,7 +49,7 @@ import org.slf4j.LoggerFactory;
 public class Utils {
 
   private static final Logger LOG = LoggerFactory.getLogger(Utils.class);
-
+  public static final String RAW_RESERVED_VIRTUAL_PATH = "/.reserved/raw/";
   private static final boolean IBM_JAVA = System.getProperty("java.vendor")
       .contains("IBM");
 
@@ -165,8 +165,12 @@ public class Utils {
   }
 
   public static boolean checkFileSystemXAttrSupport(FileSystem fs) throws IOException {
+    return checkFileSystemXAttrSupport(fs, new Path(Path.SEPARATOR));
+  }
+
+  public static boolean checkFileSystemXAttrSupport(FileSystem fs, Path path) throws IOException {
     try {
-      fs.getXAttrs(new Path(Path.SEPARATOR));
+      fs.getXAttrs(path);
       return true;
     } catch (UnsupportedOperationException e) {
       LOG.warn("XAttr won't be preserved since it is not supported for file system: " + fs.getUri());
--- a/common/src/java/org/apache/hadoop/hive/common/FileUtils.java
+++ b/common/src/java/org/apache/hadoop/hive/common/FileUtils.java
@@ -18,6 +18,8 @@
 
 package org.apache.hadoop.hive.common;
 
+import static org.apache.hadoop.hive.shims.Utils.RAW_RESERVED_VIRTUAL_PATH;
+
 import java.io.EOFException;
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -61,11 +63,13 @@ import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.PathExistsException;
 import org.apache.hadoop.fs.PathIsDirectoryException;
 import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
 import org.apache.hadoop.hive.shims.HadoopShims;
 import org.apache.hadoop.hive.shims.ShimLoader;
 import org.apache.hadoop.hive.shims.Utils;
 import org.apache.hadoop.io.IOUtils;
 import org.apache.hadoop.security.UserGroupInformation;
+import com.google.common.base.Preconditions;
 import org.apache.hadoop.util.StringUtils;
 import org.apache.hadoop.util.functional.RemoteIterators;
 import org.apache.hive.common.util.ShutdownHookManager;
@@ -767,7 +771,7 @@ public final class FileUtils {
       // is tried and it fails. We depend upon that behaviour in cases like replication,
       // wherein if distcp fails, there is good reason to not plod along with a trivial
       // implementation, and fail instead.
-      copied = doIOUtilsCopyBytes(srcFS, srcFS.getFileStatus(src), dstFS, dst, deleteSource, overwrite, shouldPreserveXAttrs(conf, srcFS, dstFS), conf, copyStatistics);
+      copied = doIOUtilsCopyBytes(srcFS, srcFS.getFileStatus(src), dstFS, dst, deleteSource, overwrite, shouldPreserveXAttrs(conf, srcFS, dstFS, src), conf, copyStatistics);
     }
     return copied;
   }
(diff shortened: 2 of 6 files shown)


shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
```diff
@@ -172,6 +173,10 @@ public static boolean checkFileSystemXAttrSupport(FileSystem fs) throws IOExcept
       LOG.warn("XAttr won't be preserved since it is not supported for file system: " + fs.getUri());
       return false;
     }
+    catch (AccessControlException e) {
+      LOG.warn(e.getMessage());
+      return false;
+    }
```
ayushtkn, line 179: Isn't catching this wrong? Other systems like HDFS throws ACE for permission issues, & here we will be catching them as well?

harshal-16: For Isilon, while calling `fs.getXAttrs`  it throws `org.apache.hadoop.security.AccessControlException` with message `Isilon only supports getXAttrs on /.reserved/raw paths`

should I add additional check also in catch? please suggest

harshal-16: this function just checks for xAttr support, if it has access issues then it will fail at a much earlier stage

ayushtkn: it won't fail, it won't preserve xAttr and replication will go on with DistCp not preserving xAttr, this method was added to check if the FS supports xAttrs or not, if it does support, the xAttrs were always preserved.

The FileSystem description of getXAttr
```
   * @throws UnsupportedOperationException if the operation is unsupported
   *         (default outcome).
   */
  public byte[] getXAttr(Path path, String name) throws IOException {
```
https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L3280-L3283

So, Ideally every HCFS should abide by this, so ignoring ACE ain't an option

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
```diff
@@ -172,6 +173,11 @@ public static boolean checkFileSystemXAttrSupport(FileSystem fs) throws IOExcept
       LOG.warn("XAttr won't be preserved since it is not supported for file system: " + fs.getUri());
       return false;
     }
+    catch (AccessControlException e) {
+      LOG.warn("Checking for XAttrSupport at {} Location", RAW_RESERVED_VIRTUAL_PATH);
+      fs.getXAttrs(new Path(RAW_RESERVED_VIRTUAL_PATH));
```
ayushtkn, line 178: How do you know you are gonna shoot a distcp on reserved path? Say you have Isilon system, but you aren't using the reserved path. In that case this will return true, but when you shoot an actual distcp, you will fail because you were using the normal paths

ayushtkn (commented):
Honestly playing with ACE & this logic doesn't look very safe to me. This method is used by other components apart from replication as well.

As far as I remember, the use of raw paths is governed by this conf (Can check **DirCopyTask**)
```
    REPL_ADD_RAW_RESERVED_NAMESPACE("hive.repl.add.raw.reserved.namespace", false,
        "For TDE with same encryption keys on source and target, allow Distcp super user to access \n"
            + "the raw bytes from filesystem without decrypting on source and then encrypting on target."),

```

So, maybe we should a coin a logic around this. Like the path to try can be root, if this conf is set we can use a reserved path & remove this ACE logic itself.

```
diff --git a/common/src/java/org/apache/hadoop/hive/common/FileUtils.java b/common/src/java/org/apache/hadoop/hive/common/FileUtils.java
index 18efe167a6..4f583e2595 100644
--- a/common/src/java/org/apache/hadoop/hive/common/FileUtils.java
+++ b/common/src/java/org/apache/hadoop/hive/common/FileUtils.java
@@ -75,6 +75,8 @@
 
 import javax.security.auth.login.LoginException;
 
+import static org.apache.hadoop.hive.conf.HiveConf.ConfVars.REPL_ADD_RAW_RESERVED_NAMESPACE;
+
 /**
  * Collection of file manipulation utilities common across Hive.
  */
@@ -896,7 +898,8 @@ private static void checkDependencies(FileSystem srcFS, Path src, FileSystem dst
   }
 
   public static boolean shouldPreserveXAttrs(HiveConf conf, FileSystem srcFS, FileSystem dstFS) throws IOException {
-    if (!Utils.checkFileSystemXAttrSupport(srcFS) || !Utils.checkFileSystemXAttrSupport(dstFS)){
+    Path path = conf.getBoolVar(REPL_ADD_RAW_RESERVED_NAMESPACE) ? new Path("/.reserved/raw/") : new Path("/");
+    if (!Utils.checkFileSystemXAttrSupport(srcFS, path) || !Utils.checkFileSystemXAttrSupport(dstFS, path)) {
       return false;
     }
     for (Map.Entry<String,String> entry : conf.getPropsWithPrefix(Utils.DISTCP_OPTIONS_PREFIX).entrySet()) {
diff --git a/shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java b/shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
index 339f0b5e9c..2513eee837 100644
--- a/shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
+++ b/shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
@@ -164,9 +164,9 @@ public static Filter getXSRFFilter() {
     return Utils.constructXSRFFilter();
   }
 
-  public static boolean checkFileSystemXAttrSupport(FileSystem fs) throws IOException {
+  public static boolean checkFileSystemXAttrSupport(FileSystem fs, Path path) throws IOException {
     try {
-      fs.getXAttrs(new Path(Path.SEPARATOR));
+      fs.getXAttrs(path);
       return true;
     } catch (UnsupportedOperationException e) {
       LOG.warn("XAttr won't be preserved since it is not supported for file system: " + fs.getUri());

```

This might hack your use case, if you are using reserved paths, but when not using reserved paths, it would throw you out with ACE, which your current code will also do.

For your current code->
For isilion, if not using reserve paths, you will set preserve Xattrs but during distcp it will throw you an ACE, because it doesn't support XAttrs in generic case, so your replication will anyway fail with your current code as well


--> Thoughts from gallery not from field, It has been long time since I checked the replication code, so I might have old info or wrong info, If other folks are convinced, please proceed no objections from my side

pudidic:
Looks good to me. Because the default value and behavior is kept same. Only if the configuration is enabled, it works. However, please fill the fields of the default PR template for reviewers.

harshal-16:
Filled the default PR template, please have a look @pudidic

PR #4851, opened 2023-11-03
HIVE-27847: Casting NUMERIC types to TIMESTAMP is prohibited

basapuram-kumar:
As per this jira [HIVE-27847](https://issues.apache.org/jira/browse/HIVE-27847) , this fix will prevent query failures for Casting NUMERIC types to TIMESTAMP.

To avoid query failures on master branch, 4.0.0-alpha-1 & 2 branches.

No

No

Reproducible steps are added in the jira [HIVE-27847](https://issues.apache.org/jira/browse/HIVE-27847), after applying this fix, the failed query gets succeded.


aturoczy (changes requested):
Could you please create a test around it? The ticket is well described what it is about. But would be great to have test(s) to check what are the new results. Depend of the input

okumin:
@basapuram-kumar @deniskuzZ Hi, we are also hitting this problem, and I found the issue happens when both of the following conditions are satisfied.

- `GenericUDFTimestamp` is evaluated at runtime where no hive-site.xml is located, which typically means TezChild
- It is vectorized. `VectorSelectOperator` is unlikely to have the issue

I guess the current implementation partially works but it could not be perfect as `intToTimestampInSeconds` can be still wrongly configured. Could you please try to fix the issue in the correct way? I can take it over if you prefer it.

This is the minimal reproduction in my environment.

```
CREATE TABLE test AS SELECT 0 AS ts;

set hive.strict.timestamp.conversion=false;
set hive.vectorized.execution.enabled=false;

SELECT CAST(ts AS TIMESTAMP) FROM test;
```

```
Caused by: java.lang.RuntimeException: Map operator initialization failed
	at org.apache.hadoop.hive.ql.exec.tez.MapRecordProcessor.init(MapRecordProcessor.java:351)
	at org.apache.hadoop.hive.ql.exec.tez.TezProcessor.initializeAndRunProcessor(TezProcessor.java:292)
	... 16 more
Caused by: org.apache.hadoop.hive.ql.exec.UDFArgumentException: Casting NUMERIC types to TIMESTAMP is prohibited (hive.strict.timestamp.conversion)
	at org.apache.hadoop.hive.ql.udf.generic.GenericUDFTimestamp.initialize(GenericUDFTimestamp.java:91)
	at org.apache.hadoop.hive.ql.udf.generic.GenericUDF.initializeAndFoldConstants(GenericUDF.java:149)
	at org.apache.hadoop.hive.ql.exec.ExprNodeGenericFuncEvaluator.initialize(ExprNodeGenericFuncEvaluator.java:184)
	at org.apache.hadoop.hive.ql.exec.Operator.initEvaluators(Operator.java:1073)
	at org.apache.hadoop.hive.ql.exec.Operator.initEvaluatorsAndReturnStruct(Operator.java:1099)
	at org.apache.hadoop.hive.ql.exec.SelectOperator.initializeOp(SelectOperator.java:74)
	at org.apache.hadoop.hive.ql.exec.Operator.initialize(Operator.java:360)
```

okumin:
Also, GenericUDFTimestamp.java is not the only file that accesses runtime configurations. We should review all UDFs accessing HiveConf.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFTimestamp.java
```diff
@@ -71,22 +71,20 @@ public class GenericUDFTimestamp extends GenericUDF {
...
 
-    strict = SessionState.get() != null ? SessionState.get().getConf()
-        .getBoolVar(ConfVars.HIVE_STRICT_TIMESTAMP_CONVERSION) : new HiveConf()
-        .getBoolVar(ConfVars.HIVE_STRICT_TIMESTAMP_CONVERSION);
     intToTimestampInSeconds = SessionState.get() != null ? SessionState.get().getConf()
         .getBoolVar(ConfVars.HIVE_INT_TIMESTAMP_CONVERSION_IN_SECONDS) : new HiveConf()
         .getBoolVar(ConfVars.HIVE_INT_TIMESTAMP_CONVERSION_IN_SECONDS);
 
-    if (strict) {
+    SessionState ss = SessionState.get();
```
zratkai, line 85: Please use more convenient names like sessionState here.

zratkai (changes requested):
Please check my comments!

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFTimestamp.java
```diff
@@ -71,22 +71,20 @@ public class GenericUDFTimestamp extends GenericUDF {
...
+
+    if (ss != null && ss.getConf().getBoolVar(ConfVars.HIVE_STRICT_TIMESTAMP_CONVERSION)) {
```
okumin, line 87: The root cause is SessionState is not available at runtime. So, we need to inject the proper config from anywhere else.
For example, we implement `GenericUDF#configure` for some UDFs in case that SessionState is unavailable.
https://github.com/apache/hive/blob/a1420ed816c315d98be7ebf05cdc3ba139a68643/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFToUnixTimeStamp.java#L125-L132
https://github.com/apache/hive/blob/a1420ed816c315d98be7ebf05cdc3ba139a68643/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFRegExp.java#L64-L70

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFTimestamp.java
```diff
@@ -71,22 +71,20 @@ public class GenericUDFTimestamp extends GenericUDF {
...
     checkArgsSize(arguments, 1, 1);
     checkArgPrimitive(arguments, 0);
     checkArgGroups(arguments, 0, tsInputTypes, STRING_GROUP, DATE_GROUP, NUMERIC_GROUP, VOID_GROUP, BOOLEAN_GROUP);
 
-    strict = SessionState.get() != null ? SessionState.get().getConf()
-        .getBoolVar(ConfVars.HIVE_STRICT_TIMESTAMP_CONVERSION) : new HiveConf()
-        .getBoolVar(ConfVars.HIVE_STRICT_TIMESTAMP_CONVERSION);
     intToTimestampInSeconds = SessionState.get() != null ? SessionState.get().getConf()
         .getBoolVar(ConfVars.HIVE_INT_TIMESTAMP_CONVERSION_IN_SECONDS) : new HiveConf()
         .getBoolVar(ConfVars.HIVE_INT_TIMESTAMP_CONVERSION_IN_SECONDS);
```
okumin, line 83: I guess this also needs to be correctly configured.

dengzhhu653:
Hello @basapuram-kumar, thank you for the contribution, could you please address the suggested comments if have cycles? Thank you in advance!

dengzhhu653:
Would anyone be willing to take over this PR since we have no heard from basapuram-kumar? cc @okumin @zratkai

okumin:
So, let me take a glance.

okumin:
@dengzhhu653 I created a new PR because I have no permission to directly update @basapuram-kumar 's branch.
https://github.com/apache/hive/pull/5278

PR #4852, opened 2023-11-03
HIVE-27794: Iceberg: Implement Copy-On-Write for Merge queries

deniskuzZ:


ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteMergeRewriter.java
```diff
@@ -0,0 +1,238 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.UnaryOperator;
+
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.apache.hadoop.hive.ql.parse.rewrite.sql.SqlGeneratorFactory.TARGET_PREFIX;
+
+public class CopyOnWriteMergeRewriter extends MergeRewriter {
```
kasakrisz, line 44: What is used from `MergeRewriter`? Isn't it better to implement `Rewriter<MergeStatement>` interface ?

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
```diff
@@ -45,40 +45,51 @@ public class CopyOnWriteUpdateRewriter implements Rewriter<UpdateStatement> {
...
-
     MultiInsertSqlGenerator sqlGenerator = sqlGeneratorFactory.createSqlGenerator();
 
-    cowWithClauseBuilder.appendWith(sqlGenerator, filePathCol, whereClause);
-
-    sqlGenerator.append("insert into table ");
+    String whereClause = null;
+    int columnOffset = 0;
+    
+    boolean shouldOverwrite = updateBlock.getWhereTree() == null;
```
kasakrisz, line 65: 1. I think it would be better to move this to another Jira and patch since this isn't really about Merge.
2. I would like to avoid these type of flags in the implementation. Why don't we give another implementation of `Rewriter<UpdateStatement>`? It can extend `CopyOnWriteUpdateRewriter` if makes sense or vice versa. Or the common parts could be extracted to an abstract base class or into a wrapped class similar to `COWWithClauseBuilder` but with a different name like `CopyOnWriteUpdaterBase`

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
```diff
@@ -217,13 +218,17 @@ public void appendWhenMatchedUpdateClause(MergeStatement.UpdateClause updateClau
...
+                             List<String> values, boolean aliasRhsExpr) {
       for (FieldSchema fieldSchema : targetTable.getCols()) {
+        String value = String.format("%s.%s", targetAlias, HiveUtils.unparseIdentifier(fieldSchema.getName(), conf));
         if (newValues.containsKey(fieldSchema.getName())) {
-          values.add(newValues.get(fieldSchema.getName()));
+          String rhsExp = newValues.get(fieldSchema.getName());
+          if (aliasRhsExpr){
+            rhsExp += String.format(" AS %s", value);
+          }
+          values.add(rhsExp);
```
kasakrisz, line 229: How about extracting this to method to get rid of the boolean flag `aliasRhsExpr` since UUIC it's value is implementation specific?
```
getNewValue(...) {
  return fieldSchema.getName();
}
```
and override it in `CopyOnWriteMergeWhenClauseSqlGenerator`
```
@Override
getNewValue(...) {
  String value = String.format("%s.%s", targetAlias, HiveUtils.unparseIdentifier(fieldSchema.getName(), conf));
  return String.format("%s AS %s", fieldSchema.getName(), value);
}
```

kasakrisz (changes requested):

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergV2.java
```diff
@@ -587,31 +602,13 @@ public void testUpdateForSupportedTypes() throws IOException {
...
-    testTables.createTable(shell, "customers", HiveIcebergStorageHandlerTestUtils.CUSTOMER_SCHEMA,
-        PartitionSpec.unpartitioned(), fileFormat, HiveIcebergStorageHandlerTestUtils.OTHER_CUSTOMER_RECORDS_2);
-    // insert one more batch so that we have multiple data files within the same partition
-    shell.executeStatement(testTables.getInsertQuery(HiveIcebergStorageHandlerTestUtils.OTHER_CUSTOMER_RECORDS_1,
-        TableIdentifier.of("default", "customers"), false));
-    AssertHelpers.assertThrows("should throw exception", IllegalArgumentException.class,
-        "Attempt to do update or delete on table", () -> {
-          shell.executeStatement("DELETE FROM customers WHERE customer_id=3 or first_name='Joanna'");
-        });
+  public void testDeleteStatementFormatV1() throws TException, InterruptedException {
```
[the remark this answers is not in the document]
deniskuzZ: it was there :) i just fixed it, we can refactor later, it's already a big one

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergV2.java
```diff
@@ -403,10 +407,21 @@ public void testDeleteForSupportedTypes() throws IOException {
...
     // create and insert an initial batch of records
     testTables.createTable(shell, "customers", HiveIcebergStorageHandlerTestUtils.CUSTOMER_SCHEMA,
-        PartitionSpec.unpartitioned(), fileFormat, HiveIcebergStorageHandlerTestUtils.OTHER_CUSTOMER_RECORDS_2, 2);
+        PartitionSpec.unpartitioned(), fileFormat, HiveIcebergStorageHandlerTestUtils.OTHER_CUSTOMER_RECORDS_2,
+        formatVersion);
+
+    // verify update mode set to merge-on-read
+    Assert.assertEquals(formatVersion == 2 ? HiveIcebergStorageHandler.MERGE_ON_READ : null,
+        shell.metastore().getTable("default", "customers")
+            .getParameters().get(TableProperties.UPDATE_MODE));
```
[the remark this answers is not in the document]
deniskuzZ: yes, v1 is copy-on-write

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergV2.java
```diff
@@ -587,31 +602,13 @@ public void testUpdateForSupportedTypes() throws IOException {
...
-        PartitionSpec.unpartitioned(), fileFormat, HiveIcebergStorageHandlerTestUtils.OTHER_CUSTOMER_RECORDS_2);
-    // insert one more batch so that we have multiple data files within the same partition
-    shell.executeStatement(testTables.getInsertQuery(HiveIcebergStorageHandlerTestUtils.OTHER_CUSTOMER_RECORDS_1,
-        TableIdentifier.of("default", "customers"), false));
-    AssertHelpers.assertThrows("should throw exception", IllegalArgumentException.class,
-        "Attempt to do update or delete on table", () -> {
-          shell.executeStatement("DELETE FROM customers WHERE customer_id=3 or first_name='Joanna'");
-        });
+  public void testDeleteStatementFormatV1() throws TException, InterruptedException {
+    executeDeleteAndValidateDataUnpartitioned(1);
```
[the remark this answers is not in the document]
deniskuzZ: that was a v1 update/delete - wasn't supported before

iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_copy_on_write_partitioned.q
```diff
@@ -0,0 +1,25 @@
...
+merge into target_ice as t using source src ON t.a = src.a
+when matched and t.a > 100 THEN DELETE
+when matched then update set b = 'Merged', c = t.c + 10
+when not matched then insert values (src.a, src.b, src.c);
+
+merge into target_ice as t using source src ON t.a = src.a
+when matched and t.a > 100 THEN DELETE
+when matched then update set b = 'Merged', c = t.c + 10
+when not matched then insert values (src.a, src.b, src.c);
+
```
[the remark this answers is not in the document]
deniskuzZ: i didn't catch, you want to split merge branches in 2?

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteMergeRewriter.java
```diff
@@ -0,0 +1,238 @@
...
+    sqlGenerator.indent().append("ON ").append(onClauseAsString);
+    sqlGenerator.append('\n');
+    sqlGenerator.append(")");
+    
+    sqlGenerator.addCteExpr();
+  }
+
+  private static String replaceColumnRefsWithTargetPrefix(String columnRef, String strValue) {
+    return strValue.replaceAll(columnRef + "\\.(`?)", "$1" + TARGET_PREFIX);
+  }
```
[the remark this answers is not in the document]
deniskuzZ: replaces alias ref in column with prefix:
````
 - WHERE `t`.`a` = `src`.`a` AND NOT(`t`.`a` > 100)
 + WHERE `t__a` = `src`.`a` AND NOT(`t__a` > 100)
````

File: ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteMergeRewriter.java [hunk 9]
```diff
@@ -0,0 +1,238 @@
...
+      sqlGenerator.append('\n');
+    }
+
+    @Override
+    public void appendWhenMatchedUpdateClause(MergeStatement.UpdateClause updateClause) {
+      Table targetTable = mergeStatement.getTargetTable();
+      String targetAlias = mergeStatement.getTargetAlias();
+      String onClauseAsString = mergeStatement.getOnClauseAsText();
+
+      UnaryOperator<String> columnRefsFunc = value -> replaceColumnRefsWithTargetPrefix(targetAlias, value);
```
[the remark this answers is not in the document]
deniskuzZ: nope, it uses the `targetAlias` param

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteMergeRewriter.java
deniskuzZ: at least a few methods are reused, i don't want to introduce duplication

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
deniskuzZ: ok

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
deniskuzZ: any suggestions for the class name: CopyOnWriteIOWRewriter?

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergV2.java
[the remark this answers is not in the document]
deniskuzZ: or I can move them into TestHiveIcebergV1 and maybe figure out issue with mode persistence

deniskuzZ:
@zhangbutao, could you please check if I haven't messed up the input params in `TestHiveIceberg` tests you created (branch, tag, etc).  tried to remove redundant combinations in some of them, however, I feel we still fire many duplicates

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
deniskuzZ: + added extra refactor

kasakrisz: How about 
```
MergeRwriter.addValues(...) {
      for (FieldSchema fieldSchema : targetTable.getCols()) {
        String quotedColumnName = String.format("%s.%s", targetAlias,
            HiveUtils.unparseIdentifier(fieldSchema.getName(), conf));
        if (newValues.containsKey(fieldSchema.getName())) {
          values.add(getValue(newValues.get(fieldSchema.getName()), quotedColumnName));
        } else {
          values.add(quotedColumnName);
        }
      }
}

protected String MergeRwriter.getValue(String newValue, String alias) {
  return newValue;
}
```
```
@Override
protected String CopyOnWriteMergeRewriter.getValue(String newValue, String alias) {
  return String.format("%s AS %s", newValue, alias);
}
```
every code part is on the place where it's suppose to be and the logic in `MergeRwriter.addValues` is less complex: `if (aliasRhsExpr){` and a method parameter is removed.

OR

If `it won't harm if we add alias in existing implementations` then why don't we add it always. It also means that the boolean parameter `aliasRhsExpr` can be removed.

kasakrisz (changes requested):

zhangbutao:
> @zhangbutao, could you please check if I haven't messed up the input params in `TestHiveIceberg*` tests you created (branch, tag, etc). tried to remove redundant combinations in some of them, however, I think we still have many of those remaining

Hi @deniskuzZ , I am ok with your current change. IMO, branch&tag&snapshot related tests are mostly DDL tasks, so it is safe to be tested by one time (HiveCatalog, Parquet, V1), and the test times can be reduced from 9 time to 1 time.
For other tests i also think it should be ok as I often see the Apache Iceberg souce code tested with just one catalog.
We can think more how to find a best way to optimize all tests in the future work.

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteMergeRewriter.java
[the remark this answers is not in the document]
deniskuzZ: that just adds extra complexity, we can just call `replaceColumnRefsWithTargetPrefix(targetAlias, value)` instead, it doesn't simplify anything

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
deniskuzZ: i don't see any benefit in this, `aliasRhsExpr` is an extra param, that is false in a base class and true in 1 of the children. btw, MergeRwriter - is a base class

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
deniskuzZ: I would prefer to keep the IOW optimization in COWUpdateRewriter in that PR, however, in follow-up PR we could move it to UpdateSemanticAnalyzer and use the same optimization for ACID

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteMergeRewriter.java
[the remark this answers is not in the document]
kasakrisz: Oh, you are right. Then why don't we call it?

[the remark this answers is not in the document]
deniskuzZ: it's used multiple times in a method with 1 static input param, so I replaced it with the lambda

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteUpdateRewriter.java
deniskuzZ: created ticket [HIVE-27902](https://issues.apache.org/jira/browse/HIVE-27902)

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
deniskuzZ: changed, however, I don't think that was a critical issue

PR #4854, opened 2023-11-06
HIVE-27854:Driver compile may causing memory leak

chenruotao:
HIVE-27854:Driver compile may causing memory leak

I describe it in this the issuse 
https://issues.apache.org/jira/browse/HIVE-27854

I describe it in this the issuse 
https://issues.apache.org/jira/browse/HIVE-27854

No

No

I describe it in this the issuse 
https://issues.apache.org/jira/browse/HIVE-27854

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/DriverState.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/DriverState.java
@@ -28,9 +28,6 @@ import java.util.concurrent.locks.ReentrantLock;
  * Represents the driver's state. Also has mechanism for locking for the time of state transitions.
  */
 public class DriverState {
-  private static final String CLASS_NAME = Driver.class.getName();
-  private static final Logger LOG = LoggerFactory.getLogger(CLASS_NAME);
-
   private static ThreadLocal<DriverState> tlInstance = new ThreadLocal<DriverState>() {
     @Override
     protected DriverState initialValue() {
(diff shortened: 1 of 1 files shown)


chenruotao:
@miklosgergely hi, could you tell me why my code test fail ? please

chenruotao:
@aturoczy Hi i do test on master, this commit code can stop driver go on compiling after be closed or destroyed, but master would not cause memory leak even no this commit code, anyway, i still think the driver should not go on compiling after that, maybe cause other problems, please take a look.

aturoczy (approved):
+1
It is follow-up on a different PR
cc: @deniskuzZ @ayushtkn @SourabhBadhya

deniskuzZ:
@chenruotao, code linked in a JIRA is not relevant for Hive 4. Could you reproduce the same issue in master branch?

chenruotao:
> in SQLOperation#cleanup we call driver.close before settting it to NULL
> 
> ```
>     if (driver != null) {
>       driver.close();
>       driver.destroy();
>     }
>     driver = null;
> ```

In my test, set driver = null can not stop the compiling thread, only set aborted = true can interrupt the thread, the thread still run after set null, and it is the root cause of  memory leak in Hive 3

ql/src/java/org/apache/hadoop/hive/ql/DriverState.java
```diff
@@ -89,7 +89,11 @@ public void compiling() {
   public void compilingWithLocking() {
     lock();
     try {
-      driverState = State.COMPILING;
+      if (isDestroyed() || isClosed()) {
+        abort();
```
deniskuzZ, line 93: `abort` is only used if the query is in the middle of compilation or execution 
````
if (driverState.isCompiling() || driverState.isExecuting()) {
        driverState.abort();
}
````

chenruotao: yes, but the question is that the driver may be set null when the query is in the middle of compilation, and if the compiling thread just runs before lock(), the driverState will be reset to State.COMPILING from State.CLOSED OR State.DESTROYED, and keep running.

PR #4856, opened 2023-11-07
HIVE-27854:Driver compile may causing memory leak

chenruotao:
HIVE-27854:Driver compile may causing memory leak

I describe it in this the issuse 
https://issues.apache.org/jira/browse/HIVE-27854

I describe it in this the issuse 
https://issues.apache.org/jira/browse/HIVE-27854

No

No

I describe it in this the issuse 
https://issues.apache.org/jira/browse/HIVE-27854

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/Driver.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/Driver.java
@@ -508,6 +508,10 @@ public class Driver implements IDriver {
     perfLogger.PerfLogBegin(CLASS_NAME, PerfLogger.COMPILE);
     lDrvState.stateLock.lock();
     try {
+      if (lDrvState.driverState == DriverState.DESTROYED || lDrvState.driverState == DriverState.CLOSED) {
+        LOG.warn("FAILED: Query command could not be compiled because driver has been cancelled, closed or destroyed.");
+        return;
+      }
       lDrvState.driverState = DriverState.COMPILING;
     } finally {
       lDrvState.stateLock.unlock();


ql/src/java/org/apache/hadoop/hive/ql/Driver.java
```diff
@@ -508,6 +508,10 @@ private void compile(String command, boolean resetTaskIds, boolean deferClose) t
     perfLogger.PerfLogBegin(CLASS_NAME, PerfLogger.COMPILE);
     lDrvState.stateLock.lock();
     try {
+      if (lDrvState.driverState == DriverState.DESTROYED || lDrvState.driverState == DriverState.CLOSED) {
+        LOG.warn("FAILED: Query command could not be compiled because driver has been cancelled, closed or destroyed.");
```
aturoczy, line 512: Maybe better:
_Query compilation failed as the driver was canceled, closed, or destroyed_

aturoczy (commented):
I would merge this PR with https://github.com/apache/hive/pull/4854 
Both PR are belong to the same issue. Does not need 2 PR for this. 

Btw is this resolved on master?

ql/src/java/org/apache/hadoop/hive/ql/Driver.java
chenruotao: But there is no driverState named canceled, i think the closed of driverStates include canceled

aturoczy: ok

chenruotao:
> I would merge this PR with #4854 Both PR are belong to the same issue. Does not need 2 PR for this.
> 
> Btw is this resolved on master?

sorry for 2 PR, because the related code in master  has a difference from brance 3.1.2, so I submit 2 PR, one for brance 3.1, one for master.  And I think there is the same problem on master, but it can not return directly, i set abort.

aturoczy:
> > I would merge this PR with #4854 Both PR are belong to the same issue. Does not need 2 PR for this.
> > Btw is this resolved on master?
> 
> sorry for 2 PR, because the related code in master has a difference from brance 3.1.2, so I submit 2 PR, one for brance 3.1, one for master. And I think there is the same problem on master, but it can not return directly, so i set abort.

Please check on master. It would be better to have this PR on master, and we after that backport to 3.1. It could be a valuable change for all version

chenruotao:
> > > I would merge this PR with #4854 Both PR are belong to the same issue. Does not need 2 PR for this.
> > > Btw is this resolved on master?
> > 
> > 
> > sorry for 2 PR, because the related code in master has a difference from brance 3.1.2, so I submit 2 PR, one for brance 3.1, one for master. And I think there is the same problem on master, but it can not return directly, so i set abort.
> 
> Please check on master. It would be better to have this PR on master, and we after that backport to 3.1. It could be a valuable change for all version

ok, I will do update on https://github.com/apache/hive/pull/4854 later, that is on master, please take a review

aturoczy:
Sure! Please ping me on that PR

chenruotao:
> @chenruotao, can we close this PR as it has a duplicate?

OK, if it doesn't need to be fixed in version 3.1

PR #4857, opened 2023-11-07
HIVE-27269: Create multiple MatchTracker when using VectorFastHashMapContainer

ngsg:

This patch changes VectorMapJoinFastHashMapContainer to create multiple MatchTracker, each of which tracks distinct HashMap in the HashMapContainer.
In addition, it fixes fixed size array allocation problem in 3 classes and adjusts the number of partitioned HashTable when it is not a power of 2.

Current VectorMapJoinFastHashMapContainers have 1 MatchTracker and multiple HashMaps. MatchTracker tracks MATCHED entries by marking the index of HashTable entry in a bitset. Since multiple HashMaps use the same MatchTracker, if one of the HashTables marks the index as MATCHED, then the other HashMap's entry on the same index is also treated as MATCHED regardless of actual join result. This leads to missing non-matched join results of FullOuterJoin.

No

No

I added unit tests for Long, String, and MultiKey VectorizedFastHashMapContainer. Also, I ran TPC-DS query97 on 100GB dataset for verifying the patch.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
@@ -120,7 +120,7 @@ public class VectorMapJoinFastHashTableLoader implements org.apache.hadoop.hive.
             adjustedValue, initialValue);
       }
 
-      this.numLoadThreads = initialValue;
+      this.numLoadThreads = adjustedValue;
     }
     this.totalEntries = new LongAccumulator(Long::sum, 0L);
     this.loadExecService = Executors.newFixedThreadPool(numLoadThreads,
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
@@ -27,6 +27,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.LongAccumulator;
 
+import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.hadoop.hive.common.Pool;
 import org.apache.hadoop.hive.llap.LlapDaemonInfo;
@@ -106,9 +107,21 @@ public class VectorMapJoinFastHashTableLoader implements org.apache.hadoop.hive.
   }
 
   private void initHTLoadingService(long estKeyCount) {
-    // Avoid many small HTs that will rehash multiple times causing GCs
-    this.numLoadThreads = (estKeyCount < VectorMapJoinFastHashTable.FIRST_SIZE_UP) ? 1 :
-        HiveConf.getIntVar(hconf, HiveConf.ConfVars.HIVEMAPJOINPARALELHASHTABLETHREADS);
+    if (estKeyCount < VectorMapJoinFastHashTable.FIRST_SIZE_UP) {
+      // Avoid many small HTs that will rehash multiple times causing GCs
+      this.numLoadThreads = 1;
+    } else {
+      int initialValue = HiveConf.getIntVar(hconf, HiveConf.ConfVars.HIVEMAPJOINPARALELHASHTABLETHREADS);
+      Preconditions.checkArgument(initialValue > 0, "The number of HT-loading-threads should be positive.");
+
+      int adjustedValue = Integer.highestOneBit(initialValue);
+      if (initialValue != adjustedValue) {
+        LOG.info("Adjust the number of HT-loading-threads to {}. (Previous value: {})",
+            adjustedValue, initialValue);
+      }
+
+      this.numLoadThreads = initialValue;
+    }
     this.totalEntries = new LongAccumulator(Long::sum, 0L);
     this.loadExecService = Executors.newFixedThreadPool(numLoadThreads,
         new ThreadFactoryBuilder()
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastLongHashMapContainer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastLongHashMapContainer.java
@@ -109,10 +109,13 @@ public class VectorMapJoinFastLongHashMapContainer extends VectorMapJoinFastHash
     private NonMatchedLongHashMapIterator(MatchTracker matchTracker,
         VectorMapJoinFastLongHashMap[] vectorMapJoinFastLongHashMaps, int numThreads) {
       super(matchTracker);
+
+      assert matchTracker.getIsPartitioned();
+
       hashMapIterators = new VectorMapJoinFastLongHashMap.NonMatchedLongHashMapIterator[numThreads];
       for (int i = 0; i < numThreads; ++i) {
-        hashMapIterators[i] = new VectorMapJoinFastLongHashMap.NonMatchedLongHashMapIterator(matchTracker,
-            vectorMapJoinFastLongHashMaps[i]);
+        hashMapIterators[i] = new VectorMapJoinFastLongHashMap.NonMatchedLongHashMapIterator(
+            matchTracker.getPartition(i), vectorMapJoinFastLongHashMaps[i]);
       }
       index = 0;
       this.numThreads = numThreads;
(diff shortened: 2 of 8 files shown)


ngsg:
@ayushtkn , I added a qfile which reproduces the issue using long and string key. Could you please review this PR again? Thank you.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
```diff
@@ -106,9 +107,21 @@ public void init(ExecMapperContext context, MapredContext mrContext,
...
+    if (estKeyCount < VectorMapJoinFastHashTable.FIRST_SIZE_UP) {
+      // Avoid many small HTs that will rehash multiple times causing GCs
+      this.numLoadThreads = 1;
+    } else {
+      int initialValue = HiveConf.getIntVar(hconf, HiveConf.ConfVars.HIVEMAPJOINPARALELHASHTABLETHREADS);
+      Preconditions.checkArgument(initialValue > 0, "The number of HT-loading-threads should be positive.");
+
+      int adjustedValue = Integer.highestOneBit(initialValue);
+      if (initialValue != adjustedValue) {
+        LOG.info("Adjust the number of HT-loading-threads to {}. (Previous value: {})",
```
deniskuzZ, line 119: what's the use of that, should we auto-tune?
````
 this.numLoadThreads=adjustedValue?
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastLongHashMapContainer.java [hunk 2]
```diff
@@ -109,10 +109,13 @@ public static class NonMatchedLongHashMapIterator extends VectorMapJoinFastNonMa
     private NonMatchedLongHashMapIterator(MatchTracker matchTracker,
         VectorMapJoinFastLongHashMap[] vectorMapJoinFastLongHashMaps, int numThreads) {
       super(matchTracker);
+
+      assert matchTracker.getIsPartitioned();
```
deniskuzZ, line 113: assertions should not be used in production code

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastLongHashMapContainer.java
```diff
@@ -185,8 +188,14 @@ public JoinUtil.JoinResult lookup(long key, VectorMapJoinHashMapResult hashMapRe
 
   public JoinUtil.JoinResult lookup(long key, VectorMapJoinHashMapResult hashMapResult,
       MatchTracker matchTracker) {
+    assert matchTracker == null || matchTracker.getIsPartitioned();
```
deniskuzZ, line 191: same, assertions should not be used in production code

deniskuzZ:
@ngsg, isn't it the same for VectorFastHashSetContainer?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
ngsg: Yes, I intended auto-tuning, and the code is wrong. Thanks for pointing it out.

ngsg:
@deniskuzZ , if my memory serves me correctly, this problem happens when Hive uses MatchTracker. Only FullOuterJoin uses MatchTracker, and FOJ always uses HashMap. So I think this issue is irrelevant to HashSet and HashMultiSet. I'll check it again and leave you a comment if there is something wrong.

ngsg:
@deniskuzZ , I fixed numHTTLoadhreads adjustment code and removed assertions from the patch. Also I have checked that HashSet/HashMultiSet containers do not have this issue. Could you please review the changes?

deniskuzZ:
> @deniskuzZ , I fixed numHTTLoadhreads adjustment code and removed assertions from the patch. Also I have checked that HashSet/HashMultiSet containers do not have this issue. Could you please review the changes?

Could you please explain why that issue is not applicable to HashSet/HashMultiSet containers? 
Based on the description in a PR I would assume that all classes introduced in `HIVE-25583: Support parallel load for HastTables` are affected.
````
Current VectorMapJoinFastHashMapContainers have 1 MatchTracker and multiple HashMaps. MatchTracker tracks MATCHED entries by marking the index of HashTable entry in a bitset. Since multiple HashMaps use the same MatchTracker, if one of the HashTables marks the index as MATCHED, then the other HashMap's entry on the same index is also treated as MATCHED regardless of actual join result.````

**Comment by ngsg** [comment]
As I mentioned above, MatchTracker is used only by FOJ, and FOJ does not use HashSet/HashMultiSet container.

In current implementation, `VectorMapJoinOperator` sets `MatchTracker` by VectorMapJoinOuterGenerateResultOperator.fullOuterHashTableSetup(), which is called only by `VMJFullOuter{Long/String/MultiKey}Operator`. According to the `Vectorizer.specializeMapJoinOperator()`, these operators are chosen when `joinType` is `JoinDesc.FULL_OUTER_JOIN` and they always use `HashTableKind.HASH_MAP`. Therefore, this issue does not happen when using HashSet/HashMultiSet.

--- The remaining part explains about current fast VectorMapJoinHashTable implementation in order to help understanding the patch.

HIVE-25583 introduces parallel HT loading by creating multiple HashTables and putting them into a single HashTableContainer. (e.g. VectorMapJoinFastLongHashMapContainer has an array of VectorMapJoinFastLongHashMap.) The number of HashTables in a container is equal to the number of HT-loading threads, and a HashTable for the given KV pair(or Key in case of HashSet/HashMultiSet) is determined by the following formula: `(Hash of key) & (# of HT threads - 1)`.
```
e.g.
key hash = 0 -> use HT #0
key hash = 7, numHTThreads = 2 -> use HT #1
key hash = 7, numHTThreads = 4 -> use HT #3
```

Every HashTable has a long array which represents entries of HashTable. HashTable inserts a KV pair in 3 steps:
1. write KV pair on somewhere, usually a buffer
2. compute a long value that refer to the position of the written KV pair.
3. find an empty HashTable entry and save the long value
The last step is done by quadratic hash probing starting from the following index: `(Hash of key) & (length of the array)`.
```
e.g.
Insert key hash = 0 -> use HT #0, use slot #0
Insert key hash = 0 -> use HT #0, use slot #1 (slot #0 is already filled)
Insert key hash = 0 -> use HT #0, use slot #3 (slot #0, #1 are already filled)
```

MatchTracker tracks the slot index. MapJoinOperator may have a MatchTracker, and if it has one, it passes MatchTracker to HashTable when performing lookup. If HashTable finds matched element, then it records the slot index of matched element to MatchTracker.
```
e.g.
HashTable entries: {5, 0, 0, 6} where 0 represents an empty entry.
Lookup 1 -> MatchTracker: no slots matched
Lookup 5 -> MatchTracker: slot #0 matched
Lookup 6 -> MatchTracker: slot #0, #3 matched
```

Current implementation uses a single MatchTracker for all HashTables, which makes false-negative errors when listing non-matched elements.
```
e.g.
HT #0: {1, 5, 0, 0}, HT #1: {0, 2, 0, 0}, 0 represents an empty entry
Lookup 1 -> MatchTracker: slot #0 matched
Lookup 2 -> MatchTracker: slot #0, #1 matched
HT #0, non-matched entries according to the above MatchTracker: empty <- wrong result, should be slot #1
```

deniskuzZ:
> slot #0 matched



> As I mentioned above, MatchTracker is used only by FOJ, and FOJ does not use HashSet/HashMultiSet container.
> 
> In current implementation, `VectorMapJoinOperator` sets `MatchTracker` by VectorMapJoinOuterGenerateResultOperator.fullOuterHashTableSetup(), which is called only by `VMJFullOuter{Long/String/MultiKey}Operator`. According to the `Vectorizer.specializeMapJoinOperator()`, these operators are chosen when `joinType` is `JoinDesc.FULL_OUTER_JOIN` and they always use `HashTableKind.HASH_MAP`. Therefore, this issue does not happen when using HashSet/HashMultiSet.
> 
> --- The remaining part explains about current fast VectorMapJoinHashTable implementation in order to help understanding the patch.
> 
> [HIVE-25583](https://issues.apache.org/jira/browse/HIVE-25583) introduces parallel HT loading by creating multiple HashTables and putting them into a single HashTableContainer. (e.g. VectorMapJoinFastLongHashMapContainer has an array of VectorMapJoinFastLongHashMap.) The number of HashTables in a container is equal to the number of HT-loading threads, and a HashTable for the given KV pair(or Key in case of HashSet/HashMultiSet) is determined by the following formula: `(Hash of key) & (# of HT threads - 1)`.
> 
> ```
> e.g.
> key hash = 0 -> use HT #0
> key hash = 7, numHTThreads = 2 -> use HT #1
> key hash = 7, numHTThreads = 4 -> use HT #3
> ```
> 
> Every HashTable has a long array which represents entries of HashTable. HashTable inserts a KV pair in 3 steps:
> 
> 1. write KV pair on somewhere, usually a buffer
> 2. compute a long value that refer to the position of the written KV pair.
> 3. find an empty HashTable entry and save the long value
>    The last step is done by quadratic hash probing starting from the following index: `(Hash of key) & (length of the array)`.
> 
> ```
> e.g.
> Insert key hash = 0 -> use HT #0, use slot #0
> Insert key hash = 0 -> use HT #0, use slot #1 (slot #0 is already filled)
> Insert key hash = 0 -> use HT #0, use slot #3 (slot #0, #1 are already filled)
> ```
> 
> MatchTracker tracks the slot index. MapJoinOperator may have a MatchTracker, and if it has one, it passes MatchTracker to HashTable when performing lookup. If HashTable finds matched element, then it records the slot index of matched element to MatchTracker.
> 
> ```
> e.g.
> HashTable entries: {5, 0, 0, 6} where 0 represents an empty entry.
> Lookup 1 -> MatchTracker: no slots matched
> Lookup 5 -> MatchTracker: slot #0 matched
> Lookup 6 -> MatchTracker: slot #0, #3 matched
> ```
> 
> Current implementation uses a single MatchTracker for all HashTables, which makes false-negative errors when listing non-matched elements.
> 
> ```
> e.g.
> HT #0: {1, 5, 0, 0}, HT #1: {0, 2, 0, 0}, 0 represents an empty entry
> Lookup 1 -> MatchTracker: slot #0 matched
> Lookup 2 -> MatchTracker: slot #0, #1 matched
> HT #0, non-matched entries according to the above MatchTracker: empty <- wrong result, should be slot #1
> ```

@ngsg, thanks again for the detailed explanation! I understand the problem, I was just worried that we'll get exactly the same issues with HashSet/HashMultiSet maybe not with FOJ, but somewhere else. Anyhow, I'll merge this PR once we get a green build.

PR #4858, opened 2023-11-07
HIVE-27856: Disable CTE materialization by default

ngsg:

This patch changes the default value of `hive.optimize.cte.materialize.threshold` to -1. (Original value was 3.)

According to HIVE-24167, Hive fails to compile some queries when CTE materialization is enabled. 

No

No

I didn't add any test as this patch does not introduce any new feature.


deniskuzZ:
@ngsg, WDYT if we push the config change only in 4.0.0 release branch?

ngsg:
@deniskuzZ , that makes sense to me. Is there something I have to do for that?

deniskuzZ (approved):
LGTM +1

PR #4859, opened 2023-11-07
HIVE-27848 - TestCrudCompactorOnTez.secondCompactionShouldBeRefusedBe…

tarak271:
…foreEnqueueing fails

Change error message thrown when compaction is not successfully initiated

Regression of enhancement https://issues.apache.org/jira/browse/HIVE-27598

No

No

Junit tests


zabetak (changes requested):
Shouldn't we remove the `@Ignore` annotation from
`org.apache.hadoop.hive.ql.txn.compactor.TestCrudCompactorOnTez#secondCompactionShouldBeRefusedBeforeEnqueueing` to ensure that the changes in this PR fix the problem?

tarak271:
@zabetak removed @ignore annotation, please help review & merge

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -80,15 +84,15 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
       Optional<String> partitionName =  partitionMap.keySet().stream().findFirst();
       partitionName.ifPresent(compactionRequest::setPartitionname);
     }
-    List<CompactionResponse> compactionResponses =
+    Map<CompactionResponse, String> compactionResponses =
         initiatorBase.initiateCompactionForTable(compactionRequest, table.getTTable(), partitionMap);
-    for (CompactionResponse compactionResponse : compactionResponses) {
+    for (Map.Entry<CompactionResponse, String> compactionResponseEntry : compactionResponses.entrySet()) {
```
zabetak, line 89: Instead of creating and passing a `Map` around why don't we extract the partition iteration here instead of having it inside the `initiateCompactionForTable`? This way we will not need a Map and will avoid the need for iterating over the partitions twice (once inside the `initiateCompactionForTable` and once over the respective responses here).

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -80,15 +84,15 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
...
+      CompactionResponse compactionResponse = compactionResponseEntry.getKey();
       if (!compactionResponse.isAccepted()) {
-        String message;
         if (compactionResponse.isSetErrormessage()) {
-          message = compactionResponse.getErrormessage();
           throw new HiveException(ErrorMsg.COMPACTION_REFUSED, table.getDbName(), table.getTableName(),
-              "CompactionId: " + compactionResponse.getId(), message);
+              compactionResponseEntry.getValue() == null ? "" :
+                  "(partition=" + compactionResponseEntry.getValue() + ")", compactionResponse.getErrormessage());
```
zabetak, line 95: Do we have a test case that verifies that partition is present in the Exception? If not can you please add one.

tarak271: Added new test case

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -69,40 +79,48 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
...
+    //Will directly initiate compaction if an un-partitioned table/a partition is specified in the request
+    if (desc.getPartitionSpec() != null || !table.isPartitioned()) {
+      if (desc.getPartitionSpec() != null) {
+        Optional<String> partitionName = partitionMap.keySet().stream().findFirst();
+        partitionName.ifPresent(compactionRequest::setPartitionname);
       }
-      context.getConsole().printInfo("Compaction enqueued with id " + compactionResponse.getId());
-      if (desc.isBlocking() && compactionResponse.isAccepted()) {
-        waitForCompactionToFinish(compactionResponse, context);
+      CompactionResponse compactionResponse = initiatorBase.initiateCompactionForTable(compactionRequest);
```
zabetak, line 88: The `CompactionRequest` provides ways to select a specific partition using its name. The method here `initiateCompactionForTable` seems capable of launching a compaction request targeting a specific table + partition. 

Why do we need to have an additional method (`initiateCompactionForPartition`) to do more or less the same thing? Why can't we use `initiateCompactionForTable` ?

tarak271: initiateCompactionForTable is used for straight forward compaction request for a non-partitioned table or a specific partition is given

Where as initiateCompactionForPartition is used to check whether the partition is eligible to get compacted or not, if eligible then only compaction will get initiated. This the reason why i had to pass Table, Partition objects to re-use methods from Initiator and decide whether to initiate compaction or not.

**Background**

Supported case for alter table compact command are
```
1. alter table <table_name> compact '<manjor/minor>'; // For Unpartitioned table
2. alter table <table_name> partition <partition_name> compact '<major/minor>'; // to compact specific partition of a partitioned table
```

This command will not work for partitioned tables if we don't provide a partition name in that command.
Command fails stating that the command won't work for partitioned table

This has been enhanced(https://issues.apache.org/jira/browse/HIVE-27598) to support even partitioned table if user want's to compact all the eligible partitions in one command instead of executing alter table 1000 times if they have 1000 partitions.

So the plan is to check each partition and initiate compaction if the partition is eligible to be compacted, otherwise unnecessarily compaction requests would get flooded if the partition count is high

`3. alter table <partitioned_table_name> compact '<major/minor>'; // will scan all partitions and initiate compactions for eligible partitions`

For case 2, Since user asked to compact a specific partition, we will initiate compaction straight away. we don't want to check the eligibility of the partition(which is existing old behaviour)

Use case could be that the partition needs to be compacted irrespective of the delta folder count threshold for some reason.

Use case:
```
create table test_acid(id int) partitioned by(dt string);
insert into test_acid values(1,"a");
insert into test_acid values(2,"a");
insert into test_acid values(3,"a");
insert into test_acid values(4,"a");
alter table test_acid partition(dt='a') compact 'major'; // partition get compacted even the number of delta folders is 4
```

before compaction, the table will contain 4 delta folders and after compaction the table will contain only single base directory

tarak271:
@zabetak could you please help guide me, on the next steps with this pull request

zabetak:
Hey @tarak271 apologies for the delay. I think all the comments are addressed so I have to do a careful review of the refactoring to ensure that nothing broke and we should be good to merge this. I see that there are some merge conflicts in the InitiatorBase so if you can resolve these as well that would be helpful.

tarak271:
@zabetak resolved merge conflicts. Please help review and let me know if any changes are required

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -56,6 +59,13 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
       throw new HiveException(ErrorMsg.NONACID_COMPACTION_NOT_SUPPORTED, table.getDbName(), table.getTableName());
     }
 
+    Map<String, org.apache.hadoop.hive.metastore.api.Partition> partitionMap =
+        convertPartitionsFromThriftToDB(getPartitions(table, desc, context));
+
+    InitiatorBase initiatorBase = new InitiatorBase();
+    initiatorBase.setConf(context.getConf());
+    initiatorBase.initialize();
+
```
zabetak, line 68: Is there any particular reason why this block was moved earlier up in the class?

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/TestCrudCompactorOnTez.java
```diff
@@ -648,6 +646,52 @@ public void secondCompactionShouldBeRefusedBeforeEnqueueing() throws Exception {
...
+
+    //Do a compaction directly and wait for it to finish
+    CompactionRequest rqst = new CompactionRequest(dbName, tableName, CompactionType.MAJOR);
+    rqst.setPartitionname("pt=test");
+    CompactionResponse resp = txnHandler.compact(rqst);
+    runWorker(conf);
+
+    //Try to do a second compaction on the same table before the cleaner runs.
+    try {
+      driver.run("ALTER TABLE " + tableName + " partition(pt='test') COMPACT 'major'");
```
zabetak, line 680: This test is mostly a copy of the previous one but not sure we need all of this stuff. The main part that we wanted to test is the exception message when a partition is set in the request. To do that it would suffice to just add this try-catch block in the previous test or rework this test and to keep only the necessary things.

zabetak (commented):
Overall LGTM, I left some comments/questions and once these are addressed I will merge the PR. The most important one is around the `initialize` method in `InitiatorBase` class.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
tarak271: No difference in logic, just changed for readability

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/InitiatorBase.java
```diff
@@ -50,59 +49,53 @@
...
-                + table.getTableName() + " Exception: " + e.getMessage());
-        throw new RuntimeException(e);
-      }
-    });
-    return compactionResponses;
-  }
-
-  public List<CompactionResponse> initiateCompactionForTable(CompactionRequest request, Table table, Map<String, Partition> partitions) throws Exception {
+  public void initialize() throws Exception {
+    super.init(new AtomicBoolean());
```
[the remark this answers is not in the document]
tarak271: Yeah, i have overridden init itself.

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/TestCrudCompactorOnTez.java
tarak271: 1st test case is for non-partitioned table. The newly created one is for partitioned table. I have made a new test case to make it different for partitioned table as issuing compact on a partitioned table is different from non-partitioned table logic wise

tarak271:
@zabetak addressed review comments and build is successful. Please help take a look

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/InitiatorBase.java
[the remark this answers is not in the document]
zabetak: The latest patch changes the behavior of `Initiator` forcing the latter to retrieve and set a `validTxnList` during initialization something that wasn't done before. I reviewed related code in `Initiator` class and the change appears safe but in general it is best to avoid such refactorings that implicitly change the behavior of the code.

@tarak271 Did you double check that there are no undesired effects from this change in behavior? If yes then we can proceed and merge this PR.

zabetak (approved):
The refactoring here is an improvement over the previous version of the code and it does restore the old behavior when compactions are refused thus I am +1 on merging this PR.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/InitiatorBase.java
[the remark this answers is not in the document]
tarak271: @zabetak I have moved all the reused functions into CompactorUtil.java by making them static. So now initiator as well as manual compaction will re-use the functions and no inheritance between these two, so no issues w.r.t Initiator. Removed InitiatorBase as well. Please take a look. Thanks a lot for your time on this. Sorry that I made you review more number of times

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -295,4 +324,232 @@ public static LockRequest createLockRequest(HiveConf conf, CompactionInfo ci, lo
...
+    } catch (IOException | InterruptedException | MetaException e) {
+      LOG.error("Error occurred while Checking if we should compact partition {} of table {}.{} Exception: {}",
+              compactionInfo.partName, table.getDbName(), table.getTableName(), e.getMessage());
+      throw new RuntimeException(e);
+    }
+    return compactionResponse;
+  }
+
+  public static CompactionResponse initiateCompactionForTable(CompactionRequest request, TxnStore txnHandler) throws MetaException {
+    return txnHandler.compact(request);
```
zabetak, line 553: In the previous version of the code before launching the compaction the code was setting the `VALID_TXNS_KEY`
```
ValidTxnList validTxnList = TxnCommonUtils.createValidReadTxnList(txnHandler.getOpenTxns(), 0);
    conf.set(ValidTxnList.VALID_TXNS_KEY, validTxnList.writeToString());
```
Is it intentional that now it is not done?

Most of the other comments are nitpicks but this is important to clarify.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -69,40 +80,48 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
...
-      if (desc.isBlocking() && compactionResponse.isAccepted()) {
-        waitForCompactionToFinish(compactionResponse, context);
+      CompactionResponse compactionResponse = CompactorUtil.initiateCompactionForTable(compactionRequest,txnHandler);
+      parseCompactionResponse(compactionResponse, table, compactionRequest.getPartitionname());
+    } else { // Check for eligible partitions and initiate compaction
+      for (Map.Entry<String, org.apache.hadoop.hive.metastore.api.Partition> partitionMapEntry : partitionMap.entrySet()) {
+        compactionRequest.setPartitionname(partitionMapEntry.getKey());
+        CompactionResponse compactionResponse =
+            CompactorUtil.initiateCompactionForPartition(table.getTTable(), partitionMapEntry.getValue(),
+                compactionRequest,JavaUtils.hostname(),txnHandler, context.getConf());
```
zabetak, line 96: There is an identical method with `JavaUtils.hostname` in `ServerUtils` which is in common. The latter is a better alternative cause it avoids the extra coupling with the standalone-metastore module.

zabetak: nit: There are some spaces missing between commas and parameters.

zabetak (changes requested):
Left a few additional comments for the last refactoring. Please also check the test failures cause they seem kinda of related

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/InitiatorBase.java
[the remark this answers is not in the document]
zabetak: Moving the methods in `CompactorUtil` is good idea many thanks for doing that. I added some comments about the last refactoring in the respective places.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -59,6 +86,8 @@
 public class CompactorUtil {
   private static final Logger LOG = LoggerFactory.getLogger(CompactorUtil.class);
   public static final String COMPACTOR = "compactor";
+  public static final String COMPACTOR_THRESHOLD_PREFIX = "compactorthreshold.";
```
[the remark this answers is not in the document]
tarak271: Done, only those methods, that need public access were made public

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -295,4 +324,232 @@ public static LockRequest createLockRequest(HiveConf conf, CompactionInfo ci, lo
         !conf.getBoolVar(HiveConf.ConfVars.TXN_WRITE_X_LOCK));
     return requestBuilder.build();
   }
+
+  public static String getInitiatorId(long threadId, String hostName) {
+    return hostName + "-" + threadId;
+  }
+
```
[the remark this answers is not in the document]
tarak271: Done

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -295,4 +324,232 @@ public static LockRequest createLockRequest(HiveConf conf, CompactionInfo ci, lo
...
+  public static CompactionResponse requestCompaction(CompactionInfo ci, String runAs, String hostname, String runtimeVersion, TxnStore txnHandler) throws MetaException {
```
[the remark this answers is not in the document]
tarak271: Applied eclipse formatter in dev-support in intellij and corrected all the indentation related issues

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -295,4 +324,232 @@ public static LockRequest createLockRequest(HiveConf conf, CompactionInfo ci, lo
...
+    GetValidWriteIdsRequest validWriteIdsRequest = new GetValidWriteIdsRequest(Collections.singletonList(fullTableName));
+    validWriteIdsRequest.setValidTxnList(validTxnList.writeToString());
+
+    return TxnUtils.createValidCompactWriteIdList(txnHandler.getValidWriteIds(validWriteIdsRequest).getTblValidWriteIds().get(0));
+  }
+
+  public static CompactionResponse scheduleCompactionIfRequired(CompactionInfo ci, Table t,
+                                                                Partition p, String runAs, boolean metricsEnabled, String hostName, TxnStore txnHandler, HiveConf conf)
+          throws MetaException {
+    StorageDescriptor sd = CompactorUtil.resolveStorageDescriptor(t, p);
```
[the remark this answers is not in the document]
tarak271: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
tarak271: Done

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
zabetak: The code before this refactoring is not failing so I am trying to understand what is causing this change in behavior. I don't want to let changes in behavior break in the code base accidentally. Can you please elaborate why it was necessary to do it before and it shouldn't be done now?

Other than that if the method is simply a call to `txnHandler.compact(request)` then it is not really necessary to wrap it in a utility. Using `txnHandler.compact(request)` directly is shorter and more explicit.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -295,4 +324,230 @@ public static LockRequest createLockRequest(HiveConf conf, CompactionInfo ci, lo
...
+    StorageDescriptor sd = resolveStorageDescriptor(t, p);
+    try {
+      ValidWriteIdList validWriteIds = resolveValidWriteIds(t, txnHandler, conf);
+
+      checkInterrupt(Initiator.class.getName());
+
+      CompactionType type = checkForCompaction(ci, validWriteIds, sd, t.getParameters(), runAs, txnHandler, conf);
+      if (type != null) {
+        ci.type = type;
+        return requestCompaction(ci, runAs, hostName, ci.initiatorVersion, txnHandler);
```
zabetak, line 500: nit: Since we are already passing the `ci` in the method the `ci.initiatorVersion` is redundant and can be dropped.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -32,23 +38,44 @@
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
 import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.hadoop.hive.metastore.api.CompactionResponse;
+import org.apache.hadoop.hive.metastore.api.CompactionRequest;
+import org.apache.hadoop.hive.metastore.api.CompactionType;
+import org.apache.hadoop.hive.metastore.api.GetValidWriteIdsRequest;
+import org.apache.hadoop.hive.metastore.api.NoSuchTxnException;
+
+
```
zabetak, line 47: nit: Is there a style guideline that mandates multiple spaces in imports? I don't understand with what criterion they are separated.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -32,23 +38,44 @@
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
 import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.hadoop.hive.metastore.api.CompactionResponse;
+import org.apache.hadoop.hive.metastore.api.CompactionRequest;
+import org.apache.hadoop.hive.metastore.api.CompactionType;
+import org.apache.hadoop.hive.metastore.api.GetValidWriteIdsRequest;
+import org.apache.hadoop.hive.metastore.api.NoSuchTxnException;
```
zabetak, line 45: nit: Is the import order correct? I was expecting that `api.CompactionResponse` would be before `api.Partition` etc.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -35,27 +37,37 @@
...
+    TxnStore txnHandler;
     Table table = context.getDb().getTable(desc.getTableName());
     if (!AcidUtils.isTransactionalTable(table) && !AcidUtils.isNonNativeAcidTable(table)) {
       throw new HiveException(ErrorMsg.NONACID_COMPACTION_NOT_SUPPORTED, table.getDbName(), table.getTableName());
     }
 
+    Map<String, org.apache.hadoop.hive.metastore.api.Partition> partitionMap =
+        convertPartitionsFromThriftToDB(getPartitions(table, desc, context));
+
+    txnHandler = TxnUtils.getTxnStore(context.getConf());
```
zabetak, line 69: nit: txnHandler definition could be moved here. It is a good practice to declare variables closer to their actual usages (see Item 57: Minimize the scope of local variables in Effective Java).

zabetak (commented):
Some clarifications are still needed for the usage of `validTxnList`. The other comments are a bunch of nits that I would treat myself if it was just those.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
tarak271: Sorry for not clarifying this part.

Before making changes as part of this pull request, 
`ValidTxnList validTxnList = TxnCommonUtils.createValidReadTxnList(txnHandler.getOpenTxns(), 0);
    conf.set(ValidTxnList.VALID_TXNS_KEY, validTxnList.writeToString());`

is in the method  (initiateCompactionForTable|https://github.com/apache/hive/blob/a348e5d86adc611e4f72d386815a21a96b667ab4/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/InitiatorBase.java#L96) of the InitiatorBase class 

So earlier also, this is not part of AlterTableCompactOperation.

So functionally there is no change now as it is moved to CompactorUtil.initiateCompactionForPartition

Earlier initialization of TxnHandler is taken care in (MetaStoreCompactorThread|https://github.com/apache/hive/blob/a4d4b9bf3cbd04b9bad13068a13fff22e6ad9e70/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/MetaStoreCompactorThread.java#L60) due to inheritance. Now we need to initialize it in AlterTableCompactOperation.

After setting `ValidTxnList.VALID_TXNS_KEY` in conf of AlterTableCompactOperation, test cases are failing with error,
`Need to record valid WriteID list but there is no valid TxnID list `

That's why moved this into the original method CompactorUtil.initiateCompactionForPartition

Hope this is clear now. Thank you

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
tarak271: Done

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
tarak271: formatter did not do it automatically. seems like my mistake

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
tarak271: changed the order of imports

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
zabetak: Actually, I see that in the [old code](https://github.com/apache/hive/blob/764c1a2fd2e645d8fa49ad6fbabf483342ccaaf7/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/InitiatorBase.java#L101) before calling `txnHandler.compact(request)` we used to do [conf.set](https://github.com/apache/hive/blob/764c1a2fd2e645d8fa49ad6fbabf483342ccaaf7/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/InitiatorBase.java#L97) and now we don't for this code path.

tarak271: Yes. 

setting conf with `ValidTxnList.VALID_TXNS_KEY` is only needed for queries like alter table compact on partitioned tables, which needs to check & identify eligible partitions to get compacted

Earlier we are doing that for `initiateCompactionForMultiplePartitions` as we are calling it within the method `initiateCompactionForTable`. But setting conf with `ValidTxnList.VALID_TXNS_KEY` is not needed for `txnHandler.compact(request)`.  That's why we are not doing it in the newer version where we have separated methods for table and partitions

> The new CompactorUtil.initiateCompactionForPartition method, which accepts context.getConf, modifies the original configuration and not the copy.
> 
> The change above may be related to the test failures that happen when setting the VALID_TXNS_KEY in various places.

Yes, earlier InitiatorBase is a separate Class/Thread runs as a service and so it has it's own conf due to inheritance.
But now we have switched to using static methods, need to use the same config from the AlterTableCompactOperation class. 

Please advise if any changes can be made w.r.t new config object creation in the static method `initiateCompactionForPartition`

zabetak: The main question is why it is not needed.

I checked the code for `txnHandler.compact(request)` in more detail and now I can see why it is not necessary to set  `VALID_TXNS_KEY` in this case. The ValidTxnList is gonna be retrieved later on by [GetValidWriteIdsFunction](https://github.com/apache/hive/blob/ffb1165f59defa66b31b4fd9cb6367b71050071b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/GetValidWriteIdsFunction.java#L64).
Even if it was set in the previous version of the code it seems that it didn't have really any effect on `txnHandler.compact(request)` so it was redundant in the first place.

Regarding the config modification, I will start (or comment) in a separate thread to avoid mixing up the discussions.

I will now mark this as resolved.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -69,40 +80,48 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
       compactionRequest.setNumberOfBuckets(desc.getNumberOfBuckets());
     }
 
-    InitiatorBase initiatorBase = new InitiatorBase();
-    initiatorBase.setConf(context.getConf());
```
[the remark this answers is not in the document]
zabetak: Utility methods should not implicitly change the arguments especially things like configurations cause this makes them harder to use and can easily lead to bugs.

I believe we can modify the `CompactorUtil.initiateCompactionForPartition` method to not modify the input configuration by creating the `ValidTxnList` when needed (i.e., inside the `resolveValidWriteIds` method) unless there is a specific reason that we need this to be done at the very beginning. 

Assuming that it is really necessary to do the initialization of `ValidTxnList` here then we could just propagate this object as it is to subsequent methods by adding an additional param.

A final option would be to make a copy of the whole config object right before modifying it but this would be costly and wasteful in terms of resources.

zabetak (changes requested):

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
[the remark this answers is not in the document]
tarak271: 1. Changing conf at `resolveValidWriteIds` will disturb initiator flow, Initiator will set VALID_TXNS_KEY in its config object every time it runs before calling resolveValidWriteIds. And VALID_TXNS_KEY is not only used in `resolveValidWriteIds` but also some other methods like `AcidUtils.getAcidState(sd, writeIds, conf);`
Secondly, `resolveValidWriteIds` is getting executed for every partition, so it is better to keep common code into `initiateCompactionForPartition` itself to execute code only once.
2. We cannot add a new property in config object and pass it since other methods like 
`AcidUtils.getAcidState(sd, writeIds, conf);`
|
 `return getAcidState(fs, location, conf, writeIds, Ref.from(false), false);`
 |
 `public static AcidDirectory getAcidState(FileSystem fileSystem, Path candidateDirectory, Configuration conf,
      ValidWriteIdList writeIdList, Ref<Boolean> useFileIds, boolean ignoreEmptyFiles) throws IOException {`
 |		
`public static AcidDirectory getAcidState(FileSystem fileSystem, Path candidateDirectory, Configuration conf,
		      ValidWriteIdList writeIdList, Ref<Boolean> useFileIds, boolean ignoreEmptyFiles, Map<Path,
		      HdfsDirSnapshot> dirSnapshots) throws IOException {`		    
`ValidTxnList validTxnList = getValidTxnList(conf);`

   expects VALID_TXNS_KEY in the config, so adding another param will need change here as well
3. So choosing the option to clone config and set VALID_TXNS_KEY in it and pass it on to the rest of the methods. The new object is created only once per alter table compact command irrespective of the number of partitions and will not disturb initiator flow

zabetak (approved):
LGTM! I will merge this soon.

Many thanks for your persistence and bearing through this long round of reviews.

PR #4860, opened 2023-11-07
HIVE-27857: Do not check write permission while dropping external table or partition

wecharyu:
When drop external table, we actually not need delete the table path. But now HMS will check write permission if `AUTHORIZATION_STORAGE_AUTH_CHECKS` enabled no matter the table is external or not.

Fix the bug that drop external table should not fail by the storage permission.

No,

No.

Add a UT.

### DIFF
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/client/TestTablesCreateDropAlterTruncate.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/client/TestTablesCreateDropAlterTruncate.java
@@ -1572,6 +1572,23 @@ public class TestTablesCreateDropAlterTruncate extends MetaStoreClientTest {
     client.dropTable("nosuch", testTables[0].getDbName(), testTables[0].getTableName(), true, false);
   }
 
+  @Test(expected = MetaException.class)
+  public void testDropManagedTableWithoutStoragePermission() throws TException, IOException {
+    String dbName = testTables[0].getDbName();
+    String tblName = testTables[0].getTableName();
+    Table table = client.getTable(dbName, tblName);
+    Path tablePath = new Path(table.getSd().getLocation());
+    FileSystem fs = Warehouse.getFs(tablePath, new Configuration());
+    fs.setPermission(tablePath.getParent(), new FsPermission((short) 0555));
+
+    try {
+      client.dropTable(dbName, tblName);
+    } finally {
+      // recover write permission so that file can be cleaned.
+      fs.setPermission(tablePath.getParent(), new FsPermission((short) 0755));
+    }
+  }
+
   @Test
   public void testDropExternalTableWithoutStoragePermission() throws TException, IOException {
     // external table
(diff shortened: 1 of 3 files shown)


aturoczy:
Hi @wecharyu 
Do you have a described hive ticket about it?

I'm a bit confused with this ticket. Imho the HMS should check the permission even for external table drop. 

Let's imagine I have a customer external table. I want to drop the table, but it prevents it as I as a user do not have permission on the file system. Yes, it does not act as a delete operation on the file system, also does not do anything on FS but still does a permission check on something that virtually does not belong to the table to prevent use case damage.

I do not think this ticket is not right, maybe there is a use case just it is a bit hard to imagine, where this could make any benefit. (Disadvantage could have)

Could you please elaborate here and on the hive ticket?

cc: @ayushtkn @abstractdog @SourabhBadhya @InvisibleProgrammer

ayushtkn:
Does this mean, like even if the external table had purge false, it was checking for the delete permissions on the files? & you plan to relax that check if the external table has purge false, since the files aren't gonna be deleted as part of the drop table?

wecharyu:
Yes, that's what I mean. @ayushtkn

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/client/TestTablesCreateDropAlterTruncate.java
```diff
@@ -1567,6 +1572,24 @@ public void dropTableBogusCatalog() throws TException {
     client.dropTable("nosuch", testTables[0].getDbName(), testTables[0].getTableName(), true, false);
   }
 
+  @Test
+  public void testDropExternalTableWithoutStoragePermission() throws TException, IOException {
```
ayushtkn, line 1576: add a test case where data was supposed to be deleted & the user didn't had permission & assert that the exception is thrown

ayushtkn (commented):
Sounds fair enough to me, but this behaviour is there since stone age, let me see if we can have any additional feedback, if there was any legacy reason behind this. Ideally if we aren't gonna delete the data, we shouldn't check for FS permissions.

cc. @nrg4878/ @deniskuzZ in case you folks have any pointers

nrg4878:
I believe we do the same on a create external table, when the location for the external table is already existing. (would be nice to confirm). So whether or not we create that location, we still check. If thats the case, it feels consistent to me. If we fix the create table,
I do recognize that create/drop are slightly different.

Also, another thing to check is for storage based handlers, we also authorize the URL before the drop. In think case, it is required to authorize that URL. @saihemanth-cloudera Any thoughts?

wecharyu:
> only if purge is set to false, then we can ignore the delete permissions on the external table path.

How about ignoring delete permissions if `external.table.purge` is not set to true? That's the current implementation logic. @ayushtkn @saihemanth-cloudera

saihemanth-cloudera:
@wecharyu - I'm good with the current implementation. We can ignore delete permission if external.table.purge is not enabled.

saihemanth-cloudera (commented):
LGTM +1.
Need @ayushtkn approval also on this one, since he started reviewing this patch.

ayushtkn:
Thanx @saihemanth-cloudera for holding it, if you are convinced, feel free to merge, no comments from my side :-)

saihemanth-cloudera (approved):

PR #4866, opened 2023-11-09
HIVE-27865: HMS in http mode shuts down silently with no errors

dengzhhu653:
Waits for the Jetty server to die.

Threads in the jetty pool are all daemon threads, such pool does not prevent the JVM from exiting when the main thread finishes his job.

No

No

Manually test on my laptop.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
@@ -490,13 +490,11 @@ public class HiveMetaStore extends ThriftHiveMetastore {
         HMSHandler.LOG.debug("Starting HTTPServer for HMS");
         server.setStopAtShutdown(true);
         server.start();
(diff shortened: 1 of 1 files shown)


aturoczy (approved):
PR is fine, just please fill out the PR description. 

Btw why it did not work previously?

dengzhhu653:
> PR is fine, just please fill out the PR description.
> 
> Btw why it did not work previously?

Done, filled the description.
The threads in the executor pool are daemon threads,
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java#L394

If the main thread exits, threads in this pool will die as a consequence, which makes HMS drop down silently.
In HS2, we join the server while starting to serve,
https://github.com/apache/hive/blob/master/service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpCLIService.java#L282

ayushtkn (approved):
LGTM.

"Drops" in the title is not very indicative, I felt it drops the table or something silently :-), maybe we can change it to shutdown, crashes or something similar, while merging :-)

PR #4867, opened 2023-11-09
HIVE-27866: HttpRequestInterceptorBase should not add an empty "Cookie:" header to the request if no custom cookies have been specified

g3rg0:

While debugging session cookies of a sticky session, I noticed that the JDBC driver adds an empty "Cookie" header to the request if no custom cookie is configured. This is both unnecessary and unfortunately intermittently interferes with the sticky session handling in the kubernetes nginx ingress controller, so I created this PR to omit the empty "Cookie" header in this case.

Without this change the nginx ingress controller is not able to route the request to the same hiveserver2 client in a stable way, so the jdbc session will be interrupted sooner or later.

No.

No.

Tested with a new unit test, and also tested manually with a snapshot build of jdbc driver ( - checked the headers logged by the apache http client on debug level with and without custom cookies).

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hive/service/cli/thrift/TestThriftHttpCLIServiceFeatures.java
+++ b/itests/hive-unit/src/test/java/org/apache/hive/service/cli/thrift/TestThriftHttpCLIServiceFeatures.java
@@ -19,6 +19,7 @@
 package org.apache.hive.service.cli.thrift;
 
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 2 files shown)


aturoczy (approved):
+1 and nice job.
@abstractdog @deniskuzZ  could you please also check? 

Imho this is a must have in any kubernetes environment.

abstractdog (approved):
+1
good catch
restarted tests to get a green run

PR #4869, opened 2023-11-09
HIVE-27867: Incremental materialized view throws NPE whew Iceberg source table is empty

kasakrisz:
* Use Objects.equals for comparing saved snapshot in MV metadata and table current snapshot.
* HiveStorageHandler.getCurrentSnapshotContext() returns null when table is empty and no snapshot exists. Add null check of snapshotId when creating filter predicate

If both saved snapshot in MV metadata and table current snapshot are null NPE was thrown.

No.

No.

```
mvn test -Dtest=TestHiveAugmentSnapshotMaterializationRule,TestHivePushdownSnapshotFilterRule -pl ql
mvn test -Dtest.output.overwrite -Dtest=TestIcebergLlapLocalCliDriver -Dqfile=mv_iceberg_orc.q -pl itests/qtest-iceberg -Piceberg -Pitests
```
and search for NPE in `hive.log`

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
@@ -131,6 +131,11 @@ public class HiveAugmentSnapshotMaterializationRule extends RelRule<HiveAugmentS
     Table table = hiveTable.getHiveTableMD();
 
     SnapshotContext mvMetaTableSnapshot = mvMetaStoredSnapshot.get(table.getFullyQualifiedName());
+    if (table.getStorageHandler() == null) {
+      throw new UnsupportedOperationException(String.format("Table %s does not have Storage handler defined. " +
+          "Mixing native and non-native tables in a materialized view definition is currently not supported!",
+          table.getFullyQualifiedName()));
+    }
     if (Objects.equals(mvMetaTableSnapshot, table.getStorageHandler().getCurrentSnapshotContext(table))) {
       return;
     }
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
@@ -138,11 +138,8 @@ public class HiveAugmentSnapshotMaterializationRule extends RelRule<HiveAugmentS
       return;
     }
 
-    Long snapshotId = null;
-    if (mvMetaTableSnapshot != null) {
-      snapshotId = mvMetaTableSnapshot.getSnapshotId();
-      table.setVersionIntervalFrom(Long.toString(snapshotId));
-    }
+    Long snapshotId = mvMetaTableSnapshot != null ? mvMetaTableSnapshot.getSnapshotId() : null;
+    table.setVersionIntervalFrom(Objects.toString(snapshotId, null));
 
     RexBuilder rexBuilder = call.builder().getRexBuilder();
     int snapshotIdIndex = tableScan.getTable().getRowType().getField(
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
@@ -148,10 +148,10 @@ public class HiveAugmentSnapshotMaterializationRule extends RelRule<HiveAugmentS
 
     final RelBuilder relBuilder = call.builder();
     relBuilder.push(tableScan);
-    final RexNode literalSnapshotId = rexBuilder.makeLiteral(
+    final RexNode snapshotIdLiteral = rexBuilder.makeLiteral(
         snapshotId, snapshotIdType(relBuilder.getTypeFactory()), false);
     final RexNode predicateWithSnapShotId = rexBuilder.makeCall(
-        SqlStdOperatorTable.LESS_THAN_OR_EQUAL, snapshotIdInputRef, literalSnapshotId);
+        SqlStdOperatorTable.LESS_THAN_OR_EQUAL, snapshotIdInputRef, snapshotIdLiteral);
     relBuilder.filter(singletonList(predicateWithSnapShotId));
     call.transformTo(relBuilder.build());
   }
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
@@ -41,6 +41,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -65,6 +66,8 @@ import java.util.Set;
  * @see HivePushdownSnapshotFilterRule
  */
 public class HiveAugmentSnapshotMaterializationRule extends RelRule<HiveAugmentSnapshotMaterializationRule.Config> {
+
+  public static final int NULL_SNAPSHOT_ID = -1;
 
   public static RelOptRule with(Map<String, SnapshotContext> mvMetaStoredSnapshot) {
     return RelRule.Config.EMPTY.as(HiveAugmentSnapshotMaterializationRule.Config.class)
@@ -125,11 +128,12 @@ public class HiveAugmentSnapshotMaterializationRule extends RelRule<HiveAugmentS
     Table table = hiveTable.getHiveTableMD();
 
     SnapshotContext mvMetaTableSnapshot = mvMetaStoredSnapshot.get(table.getFullyQualifiedName());
-    if (mvMetaTableSnapshot.equals(table.getStorageHandler().getCurrentSnapshotContext(table))) {
+    if (Objects.equals(mvMetaTableSnapshot, table.getStorageHandler().getCurrentSnapshotContext(table))) {
       return;
     }
 
-    table.setVersionIntervalFrom(Long.toString(mvMetaTableSnapshot.getSnapshotId()));
+    long snapshotId = mvMetaTableSnapshot != null ? mvMetaTableSnapshot.getSnapshotId() : NULL_SNAPSHOT_ID;
+    table.setVersionIntervalFrom(Long.toString(snapshotId));
 
     RexBuilder rexBuilder = call.builder().getRexBuilder();
     int snapshotIdIndex = tableScan.getTable().getRowType().getField(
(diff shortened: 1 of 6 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
```diff
@@ -125,11 +131,12 @@ public void onMatch(RelOptRuleCall call) {
     Table table = hiveTable.getHiveTableMD();
 
     SnapshotContext mvMetaTableSnapshot = mvMetaStoredSnapshot.get(table.getFullyQualifiedName());
-    if (mvMetaTableSnapshot.equals(table.getStorageHandler().getCurrentSnapshotContext(table))) {
+    if (Objects.equals(mvMetaTableSnapshot, table.getStorageHandler().getCurrentSnapshotContext(table))) {
```
deniskuzZ, line 134: is there a null check for `table.getStorageHandler()`?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HivePushdownSnapshotFilterRule.java
```diff
@@ -124,7 +126,8 @@ private boolean setSnapShotId(RexNode op1, RexNode op2) {
       }
 
       RelOptHiveTable hiveTable = (RelOptHiveTable) relOptTable;
-      hiveTable.getHiveTableMD().setVersionIntervalFrom(Long.toString(snapshotId));
+      String snapshotIdText = snapshotId != NULL_SNAPSHOT_ID ? Long.toString(snapshotId) : null;
```
deniskuzZ, line 129: why is this needed if eventually, you just map null-> NULL_SNAPSHOT_ID -> null ?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
kasakrisz: Added null check.
In theory this can not happen because this rule is used only if MV base tables are Iceberg. Mixed mode is not supported.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HivePushdownSnapshotFilterRule.java
deniskuzZ: got it, thanks! won't it work if we have `NOT(ISNULL($3)) AND <=($3, NULL)`? if not, that's OK with me

kasakrisz: it is still `null`:
```
<=($3, NULL) -> null
anything AND null -> null
```

deniskuzZ: so there is no lazy evaluation for AND operators? if left is false -> return false?

deniskuzZ (approved):
LGTM +1, since AND(FALSE, NULL) is not lazy evaluated to FALSE in Hive

kasakrisz:
@deniskuzZ 
I revisited your proposal and found that `NOT(ISNULL($3)) AND <=($3, NULL)` is actually always simplified to `<=($3, NULL)` at Filter operator creation becasue $3 is a referenc to SnapshotID virtual column which type is `bigint not null`. I also checked that we don't need an incremental MV rebuild when any of the base tables was empty when the MV was last rebuilt because when the MV definition has 
* only one base table and that has new records inserted it actually means that all records are new and a full scan is needed.
* join. It is always inner join (only this join type is supported by the rewrite algorithm) hence we need full scan on all tables and a full rebuild because those new records must be joined with the records coming from the other table.

mv_iceberg_orc.q tests this scenario. In this case the augmented plan has a filter with a predicate `<=($3, NULL)` on top of the previously empty table and the rewrite doesn't happen and we fall back to full rebuild which is expected.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
```diff
@@ -125,11 +129,20 @@ public void onMatch(RelOptRuleCall call) {
...
+          "Mixing native and non-native tables in a materialized view definition is currently not supported!",
+          table.getFullyQualifiedName()));
+    }
+    if (Objects.equals(mvMetaTableSnapshot, table.getStorageHandler().getCurrentSnapshotContext(table))) {
       return;
     }
 
-    table.setVersionIntervalFrom(Long.toString(mvMetaTableSnapshot.getSnapshotId()));
+    Long snapshotId = null;
+    if (mvMetaTableSnapshot != null) {
```
deniskuzZ, line 142:
````
Long snapshotId = mvMetaTableSnapshot != null ? mvMetaTableSnapshot.getSnapshotId() : null;
table.setVersionIntervalFrom(String.valueOf(snapshotId));
````

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HivePushdownSnapshotFilterRule.java [hunk 4]
```diff
@@ -116,15 +116,16 @@ private boolean setSnapShotId(RexNode op1, RexNode op2) {
...
+      Long snapshotId = literal.getValueAs(Long.class);
 
       RelOptTable relOptTable = getRelOptTableOf(op2);
       if (relOptTable == null) {
         return false;
       }
 
       RelOptHiveTable hiveTable = (RelOptHiveTable) relOptTable;
-      hiveTable.getHiveTableMD().setVersionIntervalFrom(Long.toString(snapshotId));
+      String snapshotIdText = snapshotId != null ? Long.toString(snapshotId) : null;
```
deniskuzZ, line 127:
````
hiveTable.getHiveTableMD().setVersionIntervalFrom(String.valueOf(snapshotId))
````

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java [hunk 5]
```diff
@@ -141,11 +154,8 @@ public void onMatch(RelOptRuleCall call) {
     relBuilder.push(tableScan);
     List<RexNode> conds = new ArrayList<>();
     final RexNode literalHighWatermark = rexBuilder.makeLiteral(
-        mvMetaTableSnapshot.getSnapshotId(), snapshotIdType(relBuilder), false);
-    conds.add(
-        rexBuilder.makeCall(
-            SqlStdOperatorTable.LESS_THAN_OR_EQUAL,
-            ImmutableList.of(snapshotIdInputRef, literalHighWatermark)));
+        snapshotId, snapshotIdType(relBuilder.getTypeFactory()), false);
+    conds.add(rexBuilder.makeCall(SqlStdOperatorTable.LESS_THAN_OR_EQUAL, snapshotIdInputRef, literalHighWatermark));
```
deniskuzZ, line 158: original code formatting (with new lines) was more readable

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
kasakrisz: `String.valueOf(null)` returns string literal `"null"`. This is not a valid snapshotId

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HivePushdownSnapshotFilterRule.java
kasakrisz: `String.valueOf(null)` returns string literal `"null"`. This is not a valid snapshotId

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java
kasakrisz: Sorry, where were the new lines?

deniskuzZ:
````
    conds.add(
        rexBuilder.makeCall(
            SqlStdOperatorTable.LESS_THAN_OR_EQUAL,
            ImmutableList.of(snapshotIdInputRef, literalHighWatermark)));
````

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java (hunk 3 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java:142]: sorry, `Objects.toString(obj, null)`

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HivePushdownSnapshotFilterRule.java (hunk 4 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HivePushdownSnapshotFilterRule.java:127]: Objects.toString(obj, null)

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java (hunk 3 above)
Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java]: refactored

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HivePushdownSnapshotFilterRule.java (hunk 4 above)
Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HivePushdownSnapshotFilterRule.java]: refactored

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java (hunk 5 above)
Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java]: refactored

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAugmentSnapshotMaterializationRule.java [hunk 6]
```diff
@@ -139,14 +148,11 @@ public void onMatch(RelOptRuleCall call) {
...
     relBuilder.push(tableScan);
-    List<RexNode> conds = new ArrayList<>();
-    final RexNode literalHighWatermark = rexBuilder.makeLiteral(
-        mvMetaTableSnapshot.getSnapshotId(), snapshotIdType(relBuilder), false);
-    conds.add(
-        rexBuilder.makeCall(
-            SqlStdOperatorTable.LESS_THAN_OR_EQUAL,
-            ImmutableList.of(snapshotIdInputRef, literalHighWatermark)));
-    relBuilder.filter(conds);
+    final RexNode literalSnapshotId = rexBuilder.makeLiteral(
```
deniskuzZ, line 151: since you'll have to re-trigger build again, should we rename to `snapshotIdLiteral`

kasakrisz: renamed

PR #4871, opened 2023-11-10
HIVE-27869: Iceberg: Select HadoopTables will fail at HiveIcebergStorageHandler::canProvideColStats

zhangbutao:

Please refer to HIVE-27869 to reproduce this issue.

The name of HadoopTables iceberg is the same as its location:
https://github.com/apache/iceberg/blob/7cec1d97965a2823878aef17d5c54fc768e8612f/core/src/main/java/org/apache/iceberg/hadoop/HadoopTables.java#L95

And we make up the colStatsPath using iceberg table name and its location:
https://github.com/apache/hive/blob/master/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java#L595-L596

For HadoopTable iceberg, its colStatsPath will like this and this is a invalid path to hdfs:
`hdfs://localhost:8028/tmp/testiceberg/default/ice_test_001/stats/hdfs:/localhost:8028/tmp/testiceberg/default/ice_test_0018020750642632422610`

I think we need check if it is HadoopTable and then return a proper path in the code snippet to avoid the invalid path.

Without this PR, we can not query on HadoopTable iceberg by launching Tez task.

No

No

I have not added the test. In curent codebase, in fact we have HadoopTable tests, but the tests is on the local path. e.g. file://XXX, and the local path won't throw the exception like hdfs:// path, so we did not find this issue in existing tests.

But I haven't figured out how to test using hdfs path in iceberg codebase. I have tested this PR in my local hadoop env, it works ok. Is it enough?

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -167,6 +167,7 @@ import org.apache.iceberg.expressions.Projections;
 import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.expressions.StrictMetricsEvaluator;
 import org.apache.iceberg.hadoop.HadoopConfigurable;
+import org.apache.iceberg.hadoop.HadoopTableOperations;
 import org.apache.iceberg.hive.HiveSchemaUtil;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.mr.Catalogs;
@@ -538,7 +539,7 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
     try {
       FileSystem fs = statsPath.getFileSystem(conf);
       return  fs.exists(statsPath);
-    } catch (IOException e) {
+    } catch (Exception e) {
       LOG.warn("Exception when trying to find Iceberg column stats for table:{} , snapshot:{} , " +
           "statsPath: {} , stack trace: {}", table.name(), table.currentSnapshot(), statsPath, e);
     }
@@ -593,6 +594,9 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   }
 
   private Path getColStatsPath(Table table, long snapshotId) {
+    if (((BaseTable) table).operations() instanceof HadoopTableOperations) {
+      return new Path(table.location() + STATS + snapshotId);
+    }
     return new Path(table.location() + STATS + table.name() + snapshotId);
   }
 
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -167,7 +167,6 @@ import org.apache.iceberg.expressions.Projections;
 import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.expressions.StrictMetricsEvaluator;
 import org.apache.iceberg.hadoop.HadoopConfigurable;
-import org.apache.iceberg.hadoop.HadoopTableOperations;
 import org.apache.iceberg.hive.HiveSchemaUtil;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.mr.Catalogs;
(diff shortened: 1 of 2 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -538,7 +539,7 @@ private boolean canProvideColStats(Table table, long snapshotId) {
     try {
       FileSystem fs = statsPath.getFileSystem(conf);
       return  fs.exists(statsPath);
-    } catch (IOException e) {
+    } catch (Exception e) {
```
zhangbutao, line 542: I think we need catch all exception to avoid failing here, as colStas should not stop we querying the table.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -593,6 +594,9 @@ private Path getColStatsPath(Table table) {
   }
 
   private Path getColStatsPath(Table table, long snapshotId) {
+    if (((BaseTable) table).operations() instanceof HadoopTableOperations) {
+      return new Path(table.location() + STATS + snapshotId);
```
zhangbutao, line 598: I am not sure this path is reasonable. Please provide your suggestion here. Thanks.

zhangbutao, line 598: I am not sure this path is reasonable. Please provide your suggestion here. Thanks.

deniskuzZ: AFAIK that is supposed to be the puffin stats file path for the provided snapshot

zhangbutao: Yes, this path is puffin stats file path. Hadoop tables and Hive-catalog tables are really no different, they both can use the stats to optimize query. After this PR, the puffin stats file for Hadoop tables is created correctly, and query can get stats from the puffin stats file. So i think we definitely need the stats file for Hadoop tables.

What i want to say, should we keep the same path format for both Hadoop tables and hive-catalog tables? Then we don't need to judge this table type.
 e.g.
```
  private Path getColStatsPath(Table table, long snapshotId) {
+    return new Path(table.location() + STATS + snapshotId);
-    return new Path(table.location() + STATS + table.name() + snapshotId);
  } 
```
But I'm not sure if there will be backward compatibility issues for hive-catalog tables.

deniskuzZ: makes sense if we use stats for Hadoop tables, I wasn't sure about that, most metadata optimizations are available when we use HiveCatalog.

but anyways, not sure why we needed table name in stats file, maybe we should use similar to manifestList naming:
"snap-%d-%s", snapshotId(), UUID.randomUUID().toString())

zhangbutao: We don't need table name in stats file as the parent path already has table name which can be used to identify the specific table.
Use the `snap` as prefix is reasonable for me.  And i think there is no compatibility issues if changing the hive-catalog colpath to keep the same path format as HadopTable. Changed in https://github.com/apache/hive/pull/4871/commits/b1ffb414699b1cf6778b10b7f74bbbdc02a5d6cd

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -593,7 +593,7 @@ private Path getColStatsPath(Table table) {
   }
 
   private Path getColStatsPath(Table table, long snapshotId) {
-    return new Path(table.location() + STATS + table.name() + snapshotId);
+    return new Path(table.location() + STATS + snapshotId);
```
deniskuzZ, line 596: would be nice to have some Unit tests that creates and checks the col stats file

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: 👍

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
zhangbutao: Added test https://github.com/apache/hive/pull/4871/commits/1d78a39c56a7fca223447b14033eb277d8dc9274

deniskuzZ: thanks @zhangbutao

PR #4873, opened 2023-11-11
HIVE-27862: Map propertyContent to a wrong column in package.jdo

dengzhhu653:

### DIFF
--- a/standalone-metastore/metastore-server/src/main/resources/package.jdo
+++ b/standalone-metastore/metastore-server/src/main/resources/package.jdo
@@ -1107,7 +1107,7 @@
         <column name="DESCRIPTION" length="1000" jdbc-type="VARCHAR" allows-null="false"/>
       </field>
       <field name ="propertyContent">
-        <column name="PROPERTY_CONTENT" jdbc-type="BLOB" allows-null="true"/>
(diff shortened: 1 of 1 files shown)


aturoczy (approved):

ayushtkn (commented):
Ideally the column name in ``METASTORE_DB_PROPERTIES`` should have been ``PROPERTY_CONTENT`` like other columns, but I think we have to live with it now, due to compat issues...

There is no test which was checking the output?

dengzhhu653:
> Ideally the column name in `METASTORE_DB_PROPERTIES` should have been `PROPERTY_CONTENT` like other columns, but I think we have to live with it now, due to compat issues...
> 
> There is no test which was checking the output?

There are some tests `HMSDirectTest` which are testing out the fetch or save of  `PROPERTY_CONTENT` on derby, but they get through the test, not sure why.

ayushtkn:
hmm, do we need to catch someone from HMS side to figure out why those tests aren't working? I think it isn't an ideal situation for us, these bugs will keep on coming if those tests aren't working the way they should

ayushtkn (approved):
LGTM.
Thanx @dengzhhu653

PR #4875, opened 2023-11-16
HIVE-27880: Iceberg: Support creating a branch on an empty table

zhangbutao:

After this Iceberg change https://github.com/apache/iceberg/pull/8072 which has been inclued iceberg1.4, we can create a branch on an empty. User can create an empty branch, and then write data into the branch.

Branche can be created more easily when table is empty.

No

No

qtest


ayushtkn (approved):
LGTM

PR #4876, opened 2023-11-18
HIVE-27885 : Cast decimal from string with space without digits before dot returns NULL

nareshpr:
Support valid decimal which has only scale.

Incorrect results

No

No

Yes, added testcase in cast2.q file


ramesh0201 (approved):

ramesh0201:
+1 LGTM

aturoczy (commented):
with the .0000 is applicable in SQL standard?
.00 is more likely a Culture Info and it is bit US specific. 
cc: @ayushtkn @abstractdog

nrg4878:
@ayushtkn any concerns with this or are we good to merge?

ayushtkn:
yep, no objections from my side. go ahed!!!

PR #4877, opened 2023-11-19
Upgrade to check-spelling to 0.0.22

jsoref:

* A couple of spelling fixes (See commits)
* Refresh to check-spelling workflow and metadata

* Upgrades check-spelling from an unsupported version to the latest version
  * the unsupported version is using deprecated GitHub components which are starting to trigger warnings and may eventually cease to function

Check Spelling's reporting has changed a bit, it now offers SARIF integration with GitHub.

I'm refreshing the advice, but that isn't necessary if you happen to like the current advice.

The spelling fixes may remove errors from JavaDoc generated content

No

check-spelling is used by a number of consumers daily, this specific PR was tested in https://github.com/check-spelling-sandbox/apache-hive/actions/runs/6918098524 but previous runs were also involved in generating the content.


.github/actions/spelling/patterns.txt
```diff
@@ -1,38 +1,86 @@
...
+\b(?:Africa|Atlantic|America|Antarctica|Asia|Australia|Europe|Indian|Pacific)(?:/\w+)+
+
 # decode
 \.decode\("[0-9a-zA-Z]+"
-#
-"[0-9a-f]{80,}"
+
 # pom.xml
-<(additionalClasspathElement|arg|(?:artifact|group)Id|id|installDir|jvmArgs|mainClass|shadedClassifierName|template(?:Base|Source|Output)Dir|version)>[^<]*</\1>
+<(additionalClasspathElement|arg|(?:artifact|group)Id|id|installDir|jvmArgs|mainClass|shadedClassifierName|template(?:Base|Source|Output)Dir|version)>[^<]*</\g{-1}>
```
jsoref, line 45: the `\1` notation is very fragile, the `\g{-1}` notation will find the most recent match instead of the first match (the `(\\?['"]|&quot;)[0-9a-f]{40,}\g{-1}` rule above shifted what `\1` referenced)

PR #4879, opened 2023-11-19
HIVE-27890 FIx: Tez Progress bar not displayed in Beeline upon setting session level execution engine to Tez.

shivjha30:
When executing queries via Beeline with the server-level execution engine set to MapReduce (MR) and the session-level engine set to Tez, the anticipated behavior involves the visibility of the Tez Progress bar. However, a discrepancy has been identified where the latest session value is not considered when instantiating the ProgressMonitorStatusMapper.

This proposed pull request (PR) addresses the issue by retrieving the value of the session-level configuration for HIVE_EXECUTION_ENGINE, as opposed to the server-level configuration in service/src/java/org/apache/hive/service/cli/thrift/ThriftCLIService.java.

The modification ensures that the instantiation of the ProgressMonitorStatusMapper is appropriately evaluated, with the TezProgressMonitorStatusMapper being instantiated when the Hive execution engine is set to Tez at the session level. 

When queries are executed through Beeline and the server-level execution engine is configured to MapReduce (MR), while the session-level execution engine is set to Tez, it has been observed that the Tez Progress bar is not rendered in the output. This goes against the expected behaviour for the Tez Progress Bar. It should be rendered in the output.

No

No

The test cases already exist. The visibility of the Progress Bar was tested by running Hive queries and changing the hive execution engine frequently to Tez and back to MR in several epochs.

tez progress


aturoczy (approved):
+1

vikramahuja1001:
+1,
Looks Good

aturoczy:
@deniskuzZ @ayushtkn @abstractdog @sourabh912 Could you pls check?

abstractdog:
patch makes sense, so if we run on tez, let's display the tez progress monitor by all means, but before making this patch happen, can't we reiterate on why server-side execution engine is configured to MR? MR is not supported anymore, so we're kind of hacking here:
1. we configure MR on HS2
2. we run queries on Tez (as MR is not supported), so configure Tez in the session
3. the 2 settings conflicts, but we're aware that we want to use Tez

ayushtkn (approved):
MR is not deprecated nor discontinued as an execution engine in Apache Hive, it is a Cloudera thing for us :-)

This patch makes sense to me, anyway if the job was executed via Tez, we should display the Tez Progress bar.

Regarding the changing of execution engine at Session level, that is also fine, right? that is users choice, anyway I believe that behaviour is independent of this PR.

pkumarsinha:
+1 LGTM

pkumarsinha:
Oops, missed to add the jira ID in the commit message !!

PR #4880, opened 2023-11-20
HIVE-27714: Iceberg: metadata location overrides can cause data breach - handling default locations.

ayushtkn:

Add ability to mask default location

Better usability  

Yes, if configured the auth uri will have default location masked

No

UT

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -2224,6 +2224,10 @@ public class HiveConf extends Configuration {
     HIVE_ICEBERG_EXPIRE_SNAPSHOT_NUMTHREADS("hive.iceberg.expire.snapshot.numthreads", 4,
         "The number of threads to be used for deleting files during expire snapshot. If set to 0 or below it uses the" +
             " defult DirectExecutorService"),
+
+    HIVE_ICEBERG_MASK_DEFAULT_LOCATION("hive.iceberg.mask.default.location", false,
+        "If this is set to true the URI for auth will have the default location masked with DEFAULT_TABLE_LOCATION"),
+
     HIVEUSEEXPLICITRCFILEHEADER("hive.exec.rcfile.use.explicit.header", true,
         "If this is set the header for RCFiles will simply be RCF.  If this is not\n" +
         "set the header will be that borrowed from sequence files, e.g. SEQ- followed\n" +
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -48,6 +48,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.common.FileUtils;
 import org.apache.hadoop.hive.common.StatsSetupConst;
 import org.apache.hadoop.hive.common.type.Date;
 import org.apache.hadoop.hive.common.type.SnapshotContext;
@@ -211,6 +212,8 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   public static final String COPY_ON_WRITE = "copy-on-write";
   public static final String MERGE_ON_READ = "merge-on-read";
   public static final String STATS = "/stats/snap-";
+
+  public static final String TABLE_DEFAULT_LOCATION = "TABLE_DEFAULT_LOCATION";
 
   /**
    * Function template for producing a custom sort expression function:
@@ -518,11 +521,11 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
         writer.finish();
         return true;
       } catch (IOException e) {
-        LOG.warn("Unable to write stats to puffin file", e.getMessage());
+        LOG.warn("Unable to write stats to puffin file {}", e.getMessage());
         return false;
       }
     } catch (InvalidObjectException | IOException e) {
-      LOG.warn("Unable to invalidate or merge stats: ", e.getMessage());
+      LOG.warn("Unable to invalidate or merge stats: {}", e.getMessage());
       return false;
     }
   }
@@ -1011,19 +1014,19 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
     Optional<String> metadataLocation =
         SessionStateUtil.getProperty(conf, BaseMetastoreTableOperations.METADATA_LOCATION_PROP);
     if (metadataLocation.isPresent()) {
-      authURI.append(encodeString(metadataLocation.get()));
+      authURI.append(getPathForAuth(metadataLocation.get()));
     } else {
       Optional<String> locationProperty =
           SessionStateUtil.getProperty(conf, hive_metastoreConstants.META_TABLE_LOCATION);
       if (locationProperty.isPresent()) {
         // this property is set during the create operation before the hive table was created
         // we are returning a dummy iceberg metadata file
-        authURI.append(encodeString(URI.create(locationProperty.get()).getPath()))
+        authURI.append(getPathForAuth(locationProperty.get()))
             .append(encodeString("/metadata/dummy.metadata.json"));
       } else {
         Table table = IcebergTableUtil.getTable(conf, hmsTable);
-        authURI.append(
-            encodeString(URI.create(((BaseTable) table).operations().current().metadataFileLocation()).getPath()));
+        authURI.append(getPathForAuth(((BaseTable) table).operations().current().metadataFileLocation(),
+            hmsTable.getSd().getLocation()));
       }
     }
     LOG.debug("Iceberg storage handler authorization URI {}", authURI);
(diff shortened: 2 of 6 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2224,6 +2224,10 @@ public static enum ConfVars {
     HIVE_ICEBERG_EXPIRE_SNAPSHOT_NUMTHREADS("hive.iceberg.expire.snapshot.numthreads", 4,
         "The number of threads to be used for deleting files during expire snapshot. If set to 0 or below it uses the" +
             " defult DirectExecutorService"),
+
+    HIVE_ICEBERG_MASK_DEFAULT_LOCATION("hive.iceberg.mask.default.location", false,
+        "If set to true the URI for auth will have the default location masked with DEFAULT_TABLE_LOCATION"),
```
aturoczy, line 2229: If **this is** set to true

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -212,6 +216,8 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   public static final String MERGE_ON_READ = "merge-on-read";
   public static final String STATS = "/stats/snap-";
 
+  public static final String TABLE_DEFAULT_LOCATION = "TABLE_DEFAULT_LOCATION";
```
aturoczy, line 219: What is the TABLE_DEFAULT_LOCATION here? It would not be a default path? Would not be /user/hive/warehouse?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1038,6 +1050,28 @@ static String encodeString(String rawString) {
     return HiveConf.EncoderDecoderFactory.URL_ENCODER_DECODER.encode(rawString);
   }
 
+  static String getLocation(String location, String defaultLocation, Configuration conf) {
+    if (location == null || defaultLocation == null) {
+      return location;
```
aturoczy, line 1055: If location is null or defaultLocation is null would not be better to throw an exception? Or it's value checked somewhere else?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1038,6 +1050,28 @@ static String encodeString(String rawString) {
...
+              TABLE_DEFAULT_LOCATION) :
+          location;
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  static String getQualifiedPath(String location, Configuration conf) throws IOException {
+    Path path = new Path(location);
+    FileSystem fs = path.getFileSystem(conf);
```
aturoczy, line 1071: Maybe it is not true for java. But this type of resource should be closable (disposable) in .net. Aka it should be close the resource. I guess there could be some mechanism in Java that could close automatically, but even if it has (don't know) but the best practice is always to close a resource that is not used anymore. Especially those that is coming from the OS

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -14215,17 +14216,18 @@ ASTNode analyzeCreateTable(
...
       }
       try {
         HiveStorageHandler storageHandler = HiveUtils.getStorageHandler(conf, storageFormat.getStorageHandler());
         if (storageHandler != null) {
-          storageHandler.addResourcesForCreateTable(tblProps, conf);
+          storageHandler.addResourcesForCreateTable(tblProps, conf, db, qualifiedTabName, isExt);
         }
-      } catch (HiveException e) {
-        throw new RuntimeException(e);
+      } catch (HiveException | MetaException e) {
```
aturoczy, line 14229: Rarely see the pipe in exception :) 👍

aturoczy (changes requested):
Minor changes + some small questions

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: FileSystem is cached, it isn't created always, it is created once & next time when you try to fetch it, it returns the same instance not a new one. In the end, all the FS instances are closed for the UGI which were created by Hive during the query execution, so there isn't no explicit need for closing the FS, it will get closed in the end.
If we close it in the middle, similar issues will arise like in: [HIVE-27447](https://issues.apache.org/jira/browse/HIVE-27447)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: We are typically attempting to mask the location here, if the location is null, we can't mask the default location or there is nothing to mask in the location, if the Client considers it invalid it will throw during auth

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
aturoczy: Ack

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: ``TABLE_DEFAULT_LOCATION`` is the constant that we will replace the default location with, say /user/hive/warehouse/database/table1 is the default location, so in the metadata.json path, this will be replaced by ``TABLE_DEFAULT_LOCATION``

so it will look like TABLE_DEFAULT_LOCATION/metadata/abc-metadata.json

That is what Janos use case is: In ranger he can configure policies to allow  paths within ``TABLE_DEFAULT_LOCATION``

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
ayushtkn: I will change, in the next iter :-)

ayushtkn:
I have addressed the comments. There was a test failure, it was basically the path was specified as unqualified & the output was qualified.
So, in case we don't do the masking, we should make sure that if the path specified is Qualified, the same is returned, if it isn't qualified the same form should be returned.

Even added a check the paths should be in same FS. 

Bit of if-else now, will await for further reviews :-)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1011,25 +1018,31 @@ public URI getURIForAuth(org.apache.hadoop.hive.metastore.api.Table hmsTable) th
     Optional<String> metadataLocation =
         SessionStateUtil.getProperty(conf, BaseMetastoreTableOperations.METADATA_LOCATION_PROP);
     if (metadataLocation.isPresent()) {
-      authURI.append(encodeString(metadataLocation.get()));
+      authURI.append(encodeString(getLocation(metadataLocation.get(), getDefaultCreateLocation(), conf)));
```
deniskuzZ, line 1021: could you please refactor `getLocation(....)` to `getPathForAuth(String path)`
````
return encodeString(location)
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 7]
```diff
@@ -1011,25 +1018,31 @@ public URI getURIForAuth(org.apache.hadoop.hive.metastore.api.Table hmsTable) th
...
+        authURI.append(encodeString(URI.create(
+            getLocation(((BaseTable) table).operations().current().metadataFileLocation(),
+                hmsTable.getSd().getLocation(), conf)).getPath()));
       }
     }
     LOG.debug("Iceberg storage handler authorization URI {}", authURI);
     return new URI(authURI.toString());
   }
 
+  private String getDefaultCreateLocation() {
```
deniskuzZ, line 1042: no need for this method, just use it inside `getPathForAuth`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1038,6 +1051,44 @@ static String encodeString(String rawString) {
     return HiveConf.EncoderDecoderFactory.URL_ENCODER_DECODER.encode(rawString);
   }
 
+  static String getLocation(String location, String defaultLocation, Configuration conf) {
+    if (location == null || defaultLocation == null) {
+      return location;
+    }
+    boolean maskDefaultLocation = conf.getBoolean(HiveConf.ConfVars.HIVE_ICEBERG_MASK_DEFAULT_LOCATION.varname,
+        HiveConf.ConfVars.HIVE_ICEBERG_MASK_DEFAULT_LOCATION.defaultBoolVal);
+    if (maskDefaultLocation) {
```
deniskuzZ, line 1060:
````
if (!maskDefaultLocation) {
  return location;
}
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 9]
```diff
@@ -1011,19 +1018,19 @@ public URI getURIForAuth(org.apache.hadoop.hive.metastore.api.Table hmsTable) th
     Optional<String> metadataLocation =
         SessionStateUtil.getProperty(conf, BaseMetastoreTableOperations.METADATA_LOCATION_PROP);
     if (metadataLocation.isPresent()) {
-      authURI.append(encodeString(metadataLocation.get()));
+      authURI.append(getPathForAuth(metadataLocation.get(), null, conf));
```
deniskuzZ, line 1021: you don't need to pass `conf`, it's an instance variable. remove static and make it private. Also, please create a default method with 1 arg.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: why not remove nesting?
````
if (!maskDefaultLocation) {
  return location;
}
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 10]
```diff
@@ -1640,7 +1681,17 @@ public boolean shouldOverwrite(org.apache.hadoop.hive.ql.metadata.Table mTable,
   }
 
   @Override
-  public void addResourcesForCreateTable(Map<String, String> tblProps, HiveConf hiveConf) {
+  public void addResourcesForCreateTable(Map<String, String> tblProps, HiveConf hiveConf, Hive db,
+      TableName qualifiedTabName, boolean isExt) throws MetaException, HiveException {
+    Optional<String> defaultTableLocation =
+        SessionStateUtil.getProperty(conf, hive_metastoreConstants.DEFAULT_TABLE_LOCATION);
+    if (!defaultTableLocation.isPresent()) {
+      Warehouse wh = new Warehouse(conf);
```
deniskuzZ, line 1689: why do you need all of this? in SA#analyzeCreateTable there is already available tblLocation.
````
storageHandler.addResourcesForCreateTable(tblProps, conf, tblLocation)
````

deniskuzZ: you can just extract in SA method `getDefaultTablePath`
````
        try {
          Warehouse wh = new Warehouse(conf);
          tblLocation = wh.getDefaultTablePath(db.getDatabase(qualifiedTabName.getDb()), qualifiedTabName.getTable(),
                  isExt).toUri().getPath();
        } catch (MetaException | HiveException e) {
          throw new SemanticException(e);
        } 
````

ayushtkn: Done, refactored the method & called from validateAndAddDefaultProperties

nrg4878:
@ayushtkn  the hive_metastoreConstants is a generated class. We cannot make change to it directly. Any changes will have to be made to hive_metastore.thrift file.

So this breaks the build if you do a clean build

ayushtkn:
Thanx @nrg4878 for pointing it out, I have quickly fixed it in addendum PR at https://github.com/apache/hive/pull/4921

If you can help check if it is working fine for you. Not sure why the build on this PR didn't fail nor did the subsequent PRs, maybe you were using Pthrift and the build doesn't use that, I can followup on that as well & make sure it doesn't happen next time.

In case the addendum PR doesn't make sense to you or has issues or doesn't fix the build. I can revert it or if it blocks you feel free to revert this commit & I can rework

PR #4881, opened 2023-11-20
[HIVE-27893] Add a range validator in hive.metastore.batch.retrieve.max to only have values >= 1

vikramahuja1001:
 Blocking the setting of the property hive.metastore.batch.retrieve.max to < 1. Added a Range validator when setting the conf as well in PartitionIterable Constructor to fallback to default value in case if the batchsize is < 1

PartitionIterable gives a NoSuchElementException when batchsize has been configured as 0 or less than 0. 

No

No

Added a test case for the same

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -1299,9 +1299,10 @@ public class HiveConf extends Configuration {
      */
     @Deprecated
     METASTORE_BATCH_RETRIEVE_MAX("hive.metastore.batch.retrieve.max", 300,
+         new RangeValidator(1, null),
         "Maximum number of objects (tables/partitions) can be retrieved from metastore in one batch. \n" +
         "The higher the number, the less the number of round trips is needed to the Hive metastore server, \n" +
-        "but it may also cause higher memory requirement at the client side."),
+        "but it may also cause higher memory requirement at the client side. Batch value should be > 0 "),
     /**
      * @deprecated Use MetastoreConf.BATCH_RETRIEVE_OBJECTS_MAX
      */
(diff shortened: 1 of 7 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -1299,9 +1299,10 @@ public static enum ConfVars {
      */
     @Deprecated
     METASTORE_BATCH_RETRIEVE_MAX("hive.metastore.batch.retrieve.max", 300,
+         new RangeValidator(1, null),
         "Maximum number of objects (tables/partitions) can be retrieved from metastore in one batch. \n" +
         "The higher the number, the less the number of round trips is needed to the Hive metastore server, \n" +
-        "but it may also cause higher memory requirement at the client side."),
+        "but it may also cause higher memory requirement at the client side. Batch value should be > 0 "),
```
chinnaraolalam, line 1305: Where did we use RangeValidator and upper value can be null

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
```diff
@@ -278,7 +278,7 @@ void checkTable(String catName, String dbName, String tableName, byte[] filterEx
         request.setCatName(table.getCatName());
         int batchSize = MetastoreConf.getIntVar(conf, MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX);
         if (batchSize > 0) {
-          parts = new PartitionIterable(getMsc(), table, batchSize).withProjectSpec(request);
+          parts = new PartitionIterable(getMsc(), table, batchSize, conf).withProjectSpec(request);
```
chinnaraolalam, line 281: validate batch size here only.

PR #4882, opened 2023-11-20
HIVE-27884: LLAP: Reuse FileSystem objects from cache across different tasks in the same LLAP daemon

abstractdog:
Close the filesystem for a specific ugi once per query + daemon instead of every task.
In particular this patch:
1. moves UGI creation from task running logic (TaskRunnerCallable) to the factory
2. introduces file system closure in the same factory
3. removes unnecessary null checks

Performance. Creating a new filesystem is costly, and closing it too frequently completely disables hadoop's FileSystem CACHE

No.

No.

Tested on LLAP cluster.
1. ran multiple tpcds queries
4. checked logs (only 1 ugi/fs creation happened for a query, every other tasks fell into reusing the same ugi)
5. filesystem is closed after query


llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
```diff
@@ -15,42 +15,115 @@
...
+    /**
+     * Creates an ugi for tasks in the same query and merges the credentials.
+     * This is valid to be done once per query: no vertex-level ugi and credentials are needed, both of them
+     * are the same within the same query.
+     * Regarding vertex user: LlapTaskCommunicator has a single "user" field,
+     * which is passed into the SignableVertexSpec.
+     * Regarding credentials: LlapTaskCommunicator creates SubmitWorkRequestProto instances,
+     * into which dag-level credentials are passed.
+     * The most performant way would be to use a single UGI for the same user in the daemon, but that's not possible,
+     * because the credentials can theoretically chance across queries.
```
kokila-19, line 56: **the credentials can theoretically chance across queries.** 
Wouldn't this be change across queries?

abstractdog: yes! nice catch, let me fix

RyanEugene (commented):
Is the credentials fix - the one Kokila pointed out going in before further merge? Otherwise looks good

abstractdog:
> Is the credentials fix - the one Kokila pointed out going in before further merge? Otherwise looks good

yes, tried to clarify this in code comments, but let me add links here for more clarity:
```
     * Regarding vertex user: LlapTaskCommunicator has a single "user" field,
     * which is passed into the SignableVertexSpec.
```
https://github.com/apache/hive/blob/e0bd9eac32851b27976a78cd83584cb867b83182/llap-tez/src/java/org/apache/hadoop/hive/llap/tezplugins/LlapTaskCommunicator.java#L125

https://github.com/apache/hive/blob/e0bd9eac32851b27976a78cd83584cb867b83182/llap-tez/src/java/org/apache/hadoop/hive/llap/tezplugins/LlapTaskCommunicator.java#L919

```
     * Regarding credentials: LlapTaskCommunicator creates SubmitWorkRequestProto instances,
     * into which dag-level credentials are passed.
```
https://github.com/apache/hive/blob/e0bd9eac32851b27976a78cd83584cb867b83182/llap-tez/src/java/org/apache/hadoop/hive/llap/tezplugins/LlapTaskCommunicator.java#L918

so assuming that the user is constant and credentials are DAG-level ones, we can make sure that the UserGroupInformation and the contained Credentials are not supposed to change withing a dag

RyanEugene (approved):

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/ContainerRunnerImpl.java
```diff
@@ -292,7 +292,8 @@ public SubmitWorkResponseProto submitWork(SubmitWorkRequestProto request) throws
 
       // Lazy create conf object, as it gets expensive in this codepath.
       Supplier<Configuration> callableConf = () -> new Configuration(getConfig());
-      UserGroupInformation fsTaskUgi = fsUgiFactory == null ? null : fsUgiFactory.createUgi();
+      UserGroupInformation fsTaskUgi =
+          fsUgiFactory.createUgi(queryIdentifier.toShortString(), vertex.getUser(), credentials);
```
deniskuzZ, line 296: could we have a null fsUgiFactory?

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
```diff
@@ -15,42 +15,115 @@
...
+
   private static final HadoopShims SHIMS = ShimLoader.getHadoopShims();
 
-  private static class KerberosUgiFactory implements UgiFactory {
+  /**
+   * This class implements abstract logic for maintaining a single ugi for a specific user in a query.
+   * Subclasses implement createNewUgiInternal for creating a new ugi object when needed.
+   */
+  private static abstract class AbstractLlapUgiFactory implements UgiFactory {
+    Map<String, UserGroupInformation> ugis = new HashMap<>();
```
deniskuzZ, line 45: should we pass `QueryIdentifier` instead of `queryIdentifier.toShortString()`? see knownAppMasters

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/ContainerRunnerImpl.java
```diff
@@ -488,6 +489,7 @@ public QueryCompleteResponseProto queryComplete(
           fragmentInfo.getFragmentIdentifierString());
         executorService.killFragment(fragmentInfo.getFragmentIdentifierString());
       }
+      fsUgiFactory.closeFileSystemsForQuery(queryIdentifier.toShortString());
```
deniskuzZ, line 492: `amReporter.queryComplete(queryIdentifier)` has a null check for queryIdentifier, so can we get a NullPointer here?

common/src/java/org/apache/hadoop/hive/common/UgiFactory.java
```diff
@@ -15,8 +15,20 @@
...
+   * Creates a UserGroupInformation instance for a user with credentials in a scope of a specific query.
+   * Subclasses might implement a cache for taking care of reusing an existing ugi for the same query if possible.
+   */
+  UserGroupInformation createUgi(String queryIdentifier, String user, Credentials credentials) throws IOException;
+
+  /*
+   * Closes all filesystems for a specific query.
+   * In LLAP daemons for the same dag user and credentials, this call typically closes a single FileSystem instance.
+   */
+  void closeFileSystemsForQuery(String queryIdentifier);
```
deniskuzZ, line 33: close method shouldn't be part of a factory pattern

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
```diff
@@ -15,42 +15,115 @@
...
+      UserGroupInformation ugi = SHIMS.cloneUgi(baseUgi);
+      ugi.addCredentials(credentials);
+      LOG.info("Created ugi in KerberosUgiFactory {} for queryIdentifier '{}', current ugis #: {}", ugi,
+          queryIdentifier, ugis.size());
+      return ugi;
     }
   }
 
-  private static class NoopUgiFactory implements UgiFactory {
+  private static class NoopUgiFactory extends AbstractLlapUgiFactory implements UgiFactory {
```
deniskuzZ, line 116: It looks like a dead code, I can't find any usages of it

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/TaskRunnerCallable.java
```diff
@@ -226,10 +226,6 @@ protected TaskRunner2Result callInternal() throws Exception {
 
       // TODO Consolidate this code with TezChild.
       runtimeWatch.start();
-      if (fsTaskUgi == null) {
```
deniskuzZ, line 229: should we at least add `Preconditions.checkArgument(fsTaskUgi != null)`?

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
```diff
@@ -15,42 +15,115 @@
...
 import org.apache.hadoop.hive.llap.LlapUtil;
 import org.apache.hadoop.hive.shims.HadoopShims;
 import org.apache.hadoop.hive.shims.ShimLoader;
+import org.apache.hadoop.security.Credentials;
 import org.apache.hadoop.security.UserGroupInformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** No Java application is complete until it has a FactoryFactory. */
 public class LlapUgiFactoryFactory {
```
deniskuzZ, line 35: name is crazy, could we rename to say `LlapUgiHelper`?

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
deniskuzZ, line 56: typo: 'change'

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
abstractdog: this craziness is already part of hive come, I'm fine with changing it to any of your suggestions

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
abstractdog: ack

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/TaskRunnerCallable.java
abstractdog: we can, however, it won't help if it's null, but at least I can add a custom message for user convenience's sake

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
deniskuzZ, line 45: why do we need a map? don't we create a new Factory per query?

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
```diff
@@ -15,42 +15,115 @@
...
+        throws IOException {
       // Make sure the UGI is current.
       baseUgi.checkTGTAndReloginFromKeytab();
       // TODO: the only reason this is done this way is because we want unique Subject-s so that
       //       the FS.get gives different FS objects to different fragments.
-      // TODO: could we log in from ticket cache instead? no good method on UGI right now.
-      return SHIMS.cloneUgi(baseUgi);
+      // TODO: could we log in from ticket cache instead? no good method on UGI right now
+      UserGroupInformation ugi = SHIMS.cloneUgi(baseUgi);
+      ugi.addCredentials(credentials);
```
deniskuzZ, line 109: `addCredentials` could be extracted to the base class

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/ContainerRunnerImpl.java
abstractdog: no, we create one or fail:
```
    UgiFactory fsUgiFactory = null;
    try {
      fsUgiFactory = LlapUgiFactoryFactory.createFsUgiFactory(daemonConf);
    } catch (IOException e) {
    }
```

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
abstractdog: right, the problem was that UgiFactory is in hive-common, which should not depend on hive-llap-server...however, now I double-checked, nothing else uses UgiFactory in hive repo, it's an llap specific thing, so I'm moving it and it should work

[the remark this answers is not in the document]
abstractdog: nice catch! right, it makes the whole hacking I made useless

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/ContainerRunnerImpl.java
abstractdog: not anymore,because I changed this to QueryIdentifier according to your other suggestion :D

common/src/java/org/apache/hadoop/hive/common/UgiFactory.java
abstractdog: considering factory pattern strictly, yes, in this case I think we should just rename this to something else
AbstractLlapUgiFactory which is the primary implementor of this interface has an internal state (ugis map), which is needed when we want to close all the Fs objects that belong to a query, so the method should be called on this object

so I guess all the factory hierarch should be renamed to something else, like a "manager", so from:
```
UgiFactory -> AbstractLlapUgiFactory -> KerberosUgiFactory / NoopUgiFactory
```
... it might become:
```
UgiManager -> AbstractLlapUgiManager -> KerberosUgiManager / NoopUgiManager
```
are you fine with this @deniskuzZ ?

abstractdog:
@deniskuzZ: your comments were addressed in https://github.com/apache/hive/pull/4882/commits/c84f536c64cf6b298109473c363434d7acbf816e

common/src/java/org/apache/hadoop/hive/common/UgiFactory.java
deniskuzZ: question, could we keep the factories and use them in UgiManager that would maintain the state?

llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java
deniskuzZ:
````
LlapUgiManager {
   private ConcurrentHashMap<Identifier, UserGroupInformation > fsTaskUgiMap;
   private UgiFactory factory;
   
  public LlapUgiManager getInstance(Configuration conf) {
    return new LlapUgiManager(conf);
  }
  private LlapUgiManager(Configuration conf){
      try {
        fsUgiFactory = LlapUgiFactoryFactory.createFsUgiFactory(daemonConf);
      } catch (IOException e) {
      }
   }
}
````

File: llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java (hunk 3 above)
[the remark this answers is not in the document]
Comment (deniskuzZ) [reply llap-server/src/java/org/apache/hadoop/hive/llap/security/LlapUgiFactoryFactory.java]: 👍 :)

File: llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/ContainerRunnerImpl.java (hunk 2 above)
Comment (deniskuzZ) [reply llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/ContainerRunnerImpl.java]: 👍

File: llap-server/src/java/org/apache/hadoop/hive/llap/LlapUgiManager.java [hunk 10]
```diff
@@ -0,0 +1,85 @@
...
+
+import java.io.IOException;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LlapUgiManager {
+  private static final Logger LOG = LoggerFactory.getLogger(LlapUgiManager.class);
+  private final ConcurrentHashMap<QueryIdentifier, UserGroupInformation> ugis = new ConcurrentHashMap<>();
```
deniskuzZ, line 32: minor, please use interface - `ConcurrentMap`

llap-server/src/java/org/apache/hadoop/hive/llap/LlapUgiManager.java
```diff
@@ -0,0 +1,85 @@
...
+   * This is valid to be done once per query: no vertex-level ugi and credentials are needed, both of them
+   * are the same within the same query.
+   * Regarding vertex user: LlapTaskCommunicator has a single "user" field,
+   * which is passed into the SignableVertexSpec.
+   * Regarding credentials: LlapTaskCommunicator creates SubmitWorkRequestProto instances,
+   * into which dag-level credentials are passed.
+   * The most performant way would be to use a single UGI for the same user in the daemon, but that's not possible,
+   * because the credentials can theoretically change across queries.
+   */
+  public UserGroupInformation getUgi(QueryIdentifier queryIdentifier, String user, Credentials credentials) {
```
deniskuzZ, line 58: minor: i think `getOrCreate` is more suitable here

llap-server/src/java/org/apache/hadoop/hive/llap/LlapUgiManager.java
```diff
@@ -0,0 +1,85 @@
...
+    return ugis.computeIfAbsent(queryIdentifier,
+        k -> {
+          try {
+            UserGroupInformation ugi = ugiFactory.createUgi(user);
+            ugi.addCredentials(credentials);
+            LOG.info("Created ugi {} for queryIdentifier '{}', current ugis #: {}", ugi, queryIdentifier, ugis.size());
```
deniskuzZ, line 64: why do we need to log `ugis.size` ?

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/TaskRunnerCallable.java
deniskuzZ: could we add this precondition check to the TaskRunnerCallable constructor?

deniskuzZ (commented):
LGTM, some minor stuff

llap-server/src/java/org/apache/hadoop/hive/llap/LlapUgiManager.java
abstractdog: ack

llap-server/src/java/org/apache/hadoop/hive/llap/LlapUgiManager.java
abstractdog: ack

llap-server/src/java/org/apache/hadoop/hive/llap/LlapUgiManager.java
abstractdog: a) it gives an overall picture about currently used ugis, e.g. 4 concurrent queries means 4 ugis
b) in the early stages of implementation I was paranoid about a leak here (ugis not removed due to any reason)

both in case of a) and b) this looks useful

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/TaskRunnerCallable.java
abstractdog: ack

abstractdog:
> LGTM, some minor stuff

thanks! addressed in https://github.com/apache/hive/pull/4882/commits/506c55d07c7e6f761c1295bfcdd6e6ec4d1c1f1e

deniskuzZ (approved):
+1, pending tests

kokila-19 (approved):

abstractdog:
ready to merge, I checked the refactored patch on cluster, everything works as expected

PR #4889, opened 2023-11-21
HIVE-27876 Incorrect query results on tables with ClusterBy & SortBy

ramesh0201:
Disable `hive.map.groupby.sorted` optimization.

The optimization leads to data correctness issues. See jira https://issues.apache.org/jira/browse/HIVE-27876 for details

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -2019,10 +2019,10 @@ public class HiveConf extends Configuration {
     HIVEMULTIGROUPBYSINGLEREDUCER("hive.multigroupby.singlereducer", true,
         "Whether to optimize multi group by query to generate single M/R  job plan. If the multi group by query has \n" +
         "common group by keys, it will be optimized to generate single M/R job."),
-    HIVE_MAP_GROUPBY_SORT("hive.map.groupby.sorted", true,
+    HIVE_MAP_GROUPBY_SORT("hive.map.groupby.sorted", false,
         "If the bucketing/sorting properties of the table exactly match the grouping key, whether to perform \n" +
-        "the group by in the mapper by using BucketizedHiveInputFormat. The only downside to this\n" +
-        "is that it limits the number of mappers to the number of files."),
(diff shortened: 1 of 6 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupByOptimizer.java
```diff
@@ -28,19 +28,11 @@
 import java.util.Set;
 import java.util.Stack;
 
+import org.apache.hadoop.hive.ql.exec.*;
```
kasakrisz, line 31: AFAIK Hive code style doesn't allow wild card imports.
Here is some discussion why https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad

Please revert this change.

kasakrisz (changes requested):

kasakrisz:
@ramesh0201 
There are a few tests for `GroupByOptimizer` named groupby_sort_*.q
Most of them has a step 
```
-- perform an insert to make sure there are 2 files
```
https://github.com/apache/hive/blob/7638cb1a3b07713cc490aa2909a37037f89e08b4/ql/src/test/queries/clientpositive/groupby_sort_3.q#L12-L13

However when I run these locally and monitored the file system I experienced that only one file remained after executing this step so there might be hidden data correctness issues in those cases too.

aturoczy:
I don't get it. So if the CLUSTER BY and SORT BY are in the same command the result is incorrect? In which scenario it needed for anybody in the same statement?

aturoczy:
I don't think this patch is correct. Please revisit this implementation.

aturoczy (changes requested):
-1

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupByOptimizer.java
ramesh0201: My bad, I missed this.

ramesh0201:
Thanks for the review @kasakrisz and @aturoczy. Let me rework on this.

kasakrisz (approved):

kasakrisz:
Pending tests

PR #4891, opened 2023-11-22
HIVE-27896: Remove common-lang usage

pan3793:

Remove common-lang usage, replaced with commons-lang3

Though there is a rule to band import, but the does not cover the full class name usage.
```
<restrictImports implementation="de.skuzzle.enforcer.restrictimports.RestrictImports">
  <reason>Do not use commons-lang</reason>
  <bannedImports>
    <bannedImport>org.apache.commons.lang.**</bannedImport>
  </bannedImports>
  <includeTestCode>true</includeTestCode>
</restrictImports>
```

No

No

Pass CI


ayushtkn (approved):
LGTM

PR #4893, opened 2023-11-22
HIVE-27892: Hive 'insert overwrite table' for multiple partition tabl…

mayankkunwar:
…e issue

Authorization is not working for Hive "insert overwrite table" for multiple partition table.

Due to an authorization bug need these changes.

No

No

Unit test

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/command/CommandAuthorizerV2.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/command/CommandAuthorizerV2.java
@@ -233,8 +233,12 @@ final class CommandAuthorizerV2 {
       break;
     case DUMMYPARTITION:
     case PARTITION:
-      // TODO: not currently handled
-      return;
+      Table tbl = privObject.getTable();
+      List<String> col = tableName2Cols == null ? null :
+              tableName2Cols.get(Table.getCompleteName(tbl.getDbName(), tbl.getTableName()));
+      hivePrivObject = new HivePrivilegeObject(privObjType, tbl.getDbName(), tbl.getTableName(),
+              null, col, actionType, null, null, tbl.getOwner(), tbl.getOwnerType());
+      break;
     case SERVICE_NAME:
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/security/authorization/command/CommandAuthorizerV2.java
```diff
@@ -234,7 +234,12 @@ private static void addHivePrivObject(Entity privObject, Map<String, List<String
     case DUMMYPARTITION:
     case PARTITION:
       // TODO: not currently handled
-      return;
+      Table tbl = privObject.getTable();
+      List<String> col = tableName2Cols == null ? null :
+              tableName2Cols.get(Table.getCompleteName(tbl.getDbName(), tbl.getTableName()));
+      hivePrivObject = new HivePrivilegeObject(privObjType, tbl.getDbName(), tbl.getTableName(),
```
saihemanth-cloudera, line 240: Consider a bulk partition insert statement into a table, I think we would generate so many hive privilege objects for a single table right?
If so, we should only be generating a single object per table.

mayankkunwar: @saihemanth-cloudera but I think already when normal row is inserted without partition [here](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/command/CommandAuthorizerV2.java#L198), I see there also hive privilege objects would be generated many time, isn't it?

saihemanth-cloudera: Bulk insert into a table without partitions will always generate a single hive privilege object for table.

mayankkunwar: I have added a test to check the same on partitioned table and it also generates single hive privilege object.

saihemanth-cloudera (commented):
LGTM +1, Pending green Ptests.

saihemanth-cloudera (approved):

PR #4894, opened 2023-11-22
HIVE-27512: CalciteSemanticException.UnsupportedFeature enum to capital

maheshrajus:
Modified enum fields to capital letters and removed un used enums.

For better code readability and maintainability.

No

No

Exists CI runs

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
@@ -30,12 +30,23 @@ public class CalciteSemanticException extends SemanticException {
   private static final long serialVersionUID = 1L;
 
   public enum UnsupportedFeature {
-    Distinct_without_an_aggregation, Duplicates_in_RR, Filter_expression_with_non_boolean_return_type,
-    Having_clause_without_any_groupby, Invalid_column_reference, Invalid_decimal,
-    Less_than_equal_greater_than, Others, Same_name_in_multiple_expressions,
-    Schema_less_table, Select_alias_in_having_clause, Select_transform, Subquery,
-    Table_sample_clauses, UDTF, Union_type, Unique_join,
-    HighPrecisionTimestamp // CALCITE-1690
+    DISTINCT_WITHOUT_AN_AGGREGATION,
+    DUPLICATES_IN_RR,
+    FILTER_EXPRESSION_WITH_NON_BOOLEAN_RETURN_TYPE,
+    HAVING_CLAUSE_WITHOUT_ANY_GROUPBY,
+    INVALID_COLUMN_REFERENCE,
+    INVALID_DECIMAL,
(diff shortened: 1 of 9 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
```diff
@@ -30,12 +30,20 @@ public class CalciteSemanticException extends SemanticException {
...
+    INVALID_DECIMAL,
+    SAME_NAME_IN_MULTIPLE_EXPRESSIONS,
+    SCHEMA_LESS_TABLE,
+    SELECT_ALIAS_IN_HAVING_CLAUSE,
+    SELECT_TRANSFORM,
+    SUBQUERY,
+    TABLE_SAMPLE_CLAUSES,
+    UNION_TYPE,
+    UNIQUE_JOIN
   };
```
ayushtkn, line 47: Won't this be an incompatible change, Anyone catching this exception will have problems, right?

aturoczy: Why won't be compatible? It is just Enum an Enum compared with number. Expect if one integrator for some reason compare with string. But if those do then they are stupid and deserve to go back and sit and the corner and think about what they did :)

aturoczy (approved):
I'm fine with this change 
+1

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
ayushtkn: I see this comment was resolved @maheshrajus , I think we don't just resolve a binding objection just like that, it looks like going pass or ignoring, like yahh yahh whatever, lets resolve

ayushtkn (changes requested):
-1,
 I don't think such a refactor should be done for a public enum, it might look good or a standard practice, but it is a public enum and anybody is open to use them rather than the number/ordinal, with this change their import itself with an older name will shoot them a compilation error. 

If it is creating big issues we can revisit and see if it is worth pushing once we are on a stable release now

PS. Lets not resolve comments objecting the change without conclusin

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
maheshrajus: @ayushtkn i resolved it by mistake, sorry for it. It is not any functional change.i just put the enum in proper readable format. If any other module is using this ENUM directly then better we can not disturb it. it causes compilation issues for them.

maheshrajus:
@ayushtkn thanks for the review.
It is not any functional change. i just put the enum in proper readable format. If any other module is using this ENUM directly then better we can not disturb it. it causes compilation issues for other modules which are importing this enums directly.

PR #4896, opened 2023-11-23
Some GenericUDFs wrongly cast ObjectInspectors

okumin:

We have some UDFs that don't correctly cast ObjectInspectors.

https://issues.apache.org/jira/browse/HIVE-27905

Because we'd like to show better error messages. It should be helpful for users, and administrators who are monitoring errors. The original message is too primitive.

```
0: jdbc:hive2://hive-hiveserver2:10000/defaul> select split(array('a,b,c'), ',');
Error: Error while compiling statement: FAILED: ClassCastException org.apache.hadoop.hive.serde2.objectinspector.StandardConstantListObjectInspector cannot be cast to org.apache.hadoop.hive.serde2.objectinspector.PrimitiveObjectInspector (state=42000,code=40000)
```

Yes. Error messages can change.

No.

I added integration tests.


ayushtkn (commented):
this might be a problem in other places I believe, I see similar code in GenericUDFElt/GenericUDFLocate and many other places as well,
can we fix there as well?

ayushtkn:
Yep

aturoczy (approved):

okumin:
@ayushtkn I changed the scope, checked all UDFs, and added validations to where we incorrectly cast ObjectInspectors. I checked them by my eyes and it is possible that I missed some other mistakes.
For your information, most UDFs don't raise ClassCastException. For example, [GenericUDFElt has a validation before it uses ObjectInspector](https://github.com/apache/hive/blob/276b37830f18cd34d89db27bc549594aecad3d56/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFElt.java#L55-L61).

ayushtkn (approved):
LGTM

okumin:
Thank you!

PR #4897, opened 2023-11-23
HIVE-27906: Iceberg: Implement Delete Orphan Files.

ayushtkn:

Implement delete orphan files

To clean orphan files.

adds a delete orphan files query for iceberg tables

No.

UT

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
@@ -1,0 +1,232 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.iceberg.mr.hive;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.net.URI;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.LocatedFileStatus;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.RemoteIterator;
+import org.apache.iceberg.DataTask;
+import org.apache.iceberg.FileScanTask;
+import org.apache.iceberg.HasTableOperations;
+import org.apache.iceberg.ManifestFile;
+import org.apache.iceberg.MetadataTableUtils;
+import org.apache.iceberg.ReachableFileUtil;
+import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.StructLike;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.TableScan;
+import org.apache.iceberg.actions.DeleteOrphanFiles;
+import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.relocated.com.google.common.util.concurrent.MoreExecutors;
+import org.apache.iceberg.util.Tasks;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.iceberg.MetadataTableType.ALL_ENTRIES;
+
+public class HiveIcebergDeleteOrphanFiles implements DeleteOrphanFiles {
+
+  public static final String METADATA_FOLDER_NAME = "metadata";
+  public static final String DATA_FOLDER_NAME = "data";
+  private static final Logger LOG = LoggerFactory.getLogger(HiveIcebergDeleteOrphanFiles.class);
+  private String tableLocation = null;
+  private long olderThanTimestamp = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(3);
+  private Consumer<String> deleteFunc;
+  private ExecutorService deleteExecutorService = MoreExecutors.newDirectExecutorService();
+
+  private final Configuration conf;
+  private final Table table;
+
+  public HiveIcebergDeleteOrphanFiles(Configuration conf, Table table) {
+    this.conf = conf;
+    this.table = table;
+    this.deleteFunc = file -> table.io().deleteFile(file);
+    this.tableLocation = table.location();
+  }
+
+  @Override
+  public HiveIcebergDeleteOrphanFiles location(String location) {
+    this.tableLocation = location;
+    return this;
+  }
+
+  @Override
+  public HiveIcebergDeleteOrphanFiles olderThan(long newOlderThanTimestamp) {
+    this.olderThanTimestamp = newOlderThanTimestamp;
+    return this;
+  }
+
+  // TODO: Implement later, if there is any use case.
+  @Override
+  public HiveIcebergDeleteOrphanFiles deleteWith(Consumer<String> newDeleteFunc) {
+    this.deleteFunc = newDeleteFunc;
+    return this;
+  }
+
+  @Override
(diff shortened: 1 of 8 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -850,12 +851,44 @@ public void executeOperation(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
...
+      if (numThreads > 0) {
+        LOG.info("Executing delete orphan files on iceberg table {} with {} threads", icebergTable.name(), numThreads);
+        deleteExecutorService = getDeleteExecutorService(icebergTable.name(), numThreads);
+      }
+
+      HiveIcebergDeleteOrphanFiles deleteOrphanFiles = new HiveIcebergDeleteOrphanFiles(conf, icebergTable);
+      deleteOrphanFiles.olderThan(timestampMillis);
+      if (deleteExecutorService != null) {
+        deleteOrphanFiles.olderThan(timestampMillis);
+      }
```
zhangbutao, line 879: Line 876 and Line 878 have a same code: `deleteOrphanFiles.olderThan(timestampMillis);`
Why?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -850,12 +851,44 @@ public void executeOperation(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
         IcebergTableUtil.performMetadataDelete(icebergTable, deleteMetadataSpec.getBranchName(),
             deleteMetadataSpec.getSarg());
         break;
+      case DELETE_ORPHAN_FILES:
```
zhangbutao, line 854: It means we regard the `Delete Orphan Files` action as a metadata operation, so all the delete operation is done by HS2.
 If the iceberg table has big data and many orphan files, will the HS2 suffer from performance issue? Can we lauch tez task to delete orphan files?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: copy-paste error, The plan was like in explire snapshot, if deleteExecutorService isn't null use that for delete

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: Theoretically yes, it has to store Path strings. I am not sure how much memory that can take, but if you see Hive Replication, It launches Distcp per table, so that does listing for each table in the Hs2 only, so that doesn't choke, & launch MR jobs for copy, not for listing.

I did some benchmarking stuff there, so for ~3.8 million it was taking about 250-300 mb, but there it stores some more stuff as well, so here it should be less.

Yep, but we can explore having a Tez job, I don't have a clear idea how we can get it done via that route, but I will create a follow-up & discuss with folks & figure out a way :-)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  @Override
+  public Result execute() {
+    LOG.info("Cleaning orphan files for {}", table.name());
+    HiveIcebergDeleteOrphanFilesResult result = new HiveIcebergDeleteOrphanFilesResult();
+    result.addDeletedFiles(cleanContentFiles(olderThanTimestamp));
+    result.addDeletedFiles(cleanMetadata(olderThanTimestamp));
+
+    LOG.debug("Deleting {} files while cleaning orphan files for {}", result.deletedFiles, table.name());
```
zhangbutao, line 110: `result.deletedFiles` maybe a very long character if table has many orphan files. Maybe we can use its size instead?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
```diff
@@ -0,0 +1,232 @@
...
+    Tasks.foreach(result.deletedFiles).executeWith(deleteExecutorService).retry(3)
+        .stopRetryOn(FileNotFoundException.class).suppressFailureWhenFinished().onFailure((file, thrown) ->
+          LOG.warn("Delete failed for file: {}", file, thrown)).run(deleteFunc::accept);
+    return result;
+  }
+
+  private Set<String> cleanContentFiles(long lastTime) {
+    Set<String> validFiles = Sets.union(getAllContentFilePath(), getAllStatisticsFilePath(table));
+    LOG.debug("Valid content file for {} are {}", table.name(), validFiles);
```
zhangbutao, line 119: `validFiles` maybe a very long character if table has many files. Maybe we can use its size instead?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+    return filesToDelete;
+  }
+
+  private Table getMetadataTable() {
+    return MetadataTableUtils.createMetadataTableInstance(((HasTableOperations) table).operations(), table.name(),
+        table.name() + "#" + ALL_ENTRIES.name(), ALL_ENTRIES);
+  }
+
+  private static Set<String> getValidMetadataFiles(Table internalTable) {
```
zhangbutao, line 191: What's the meaning of `internalTable`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
```diff
@@ -0,0 +1,232 @@
...
+import org.apache.iceberg.relocated.com.google.common.collect.Iterables;
+import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.relocated.com.google.common.util.concurrent.MoreExecutors;
+import org.apache.iceberg.util.Tasks;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.iceberg.MetadataTableType.ALL_ENTRIES;
+
+public class HiveIcebergDeleteOrphanFiles implements DeleteOrphanFiles {
```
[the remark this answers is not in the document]
ayushtkn: Done

ql/src/java/org/apache/hadoop/hive/ql/parse/AlterTableExecuteSpec.java
```diff
@@ -270,4 +271,27 @@ public SearchArgument getSarg() {
...
+   * Value object class, that stores the delete orphan files operation specific parameters.
+   * <ul>
+   *   <li>timestampMillis: the time before which files should be considered to be deleted</li>
+   * </ul>
+   */
+    public static class DeleteOrphanFilesDesc {
+    private final long timestampMillis;
+
+    public DeleteOrphanFilesDesc(long timestampMillis) {
+      this.timestampMillis = timestampMillis;
```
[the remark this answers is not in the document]
ayushtkn: I added a PreCondition above to check that, I am not sure if toEpochMillis can be -ve or not, but better to be safe

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
ayushtkn: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
ayushtkn: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergDeleteOrphanFiles.java
ayushtkn: I changed it icebergTable, no special meaning, it is the normal table, not sure why I did that :-)

zhangbutao (approved):
LGTM +1

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/actions/HiveIcebergDeleteOrphanFiles.java
```diff
@@ -0,0 +1,232 @@
...
+    this.deleteExecutorService = executorService;
+    return this;
+  }
+
+  @Override
+  public Result execute() {
+    LOG.info("Cleaning orphan files for {}", table.name());
+    HiveIcebergDeleteOrphanFilesResult result = new HiveIcebergDeleteOrphanFilesResult();
+    result.addDeletedFiles(cleanContentFiles(olderThanTimestamp));
+    result.addDeletedFiles(cleanMetadata(olderThanTimestamp));
```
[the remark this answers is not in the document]
ayushtkn: makes sense, I will create a ticket for that, thx

PR #4899, opened 2023-11-24
HIVE-27899: Killed speculative execution task attempt should not comm…

zhengchenyu:

Try canCommit after process.

We should avoid speculative tasks being committed at the same time. Detailed information see: https://issues.apache.org/jira/browse/HIVE-27899

No

No

test on our cluster

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
@@ -293,6 +293,12 @@ public class TezProcessor extends AbstractLogicalIOProcessor {
       rproc.init(mrReporter, inputs, outputs);
       rproc.run();
 
+      // Try to call canCommit to AM. If there is no other speculative attempt execute canCommit, then continue.
+      // If there are other speculative attempt execute canCommit first, then wait until the attempt is killed
+      // or the committed task fails.
+      while (!this.processorContext.canCommit()) {
+        Thread.sleep(100);
+      }
     } catch (Throwable t) {
       rproc.setAborted(true);
       originalThrowable = t;
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
@@ -137,7 +137,6 @@ public class TezProcessor extends AbstractLogicalIOProcessor {
     }
   }
 
-  protected ProcessorContext processorContext;
   private ReflectiveProgressHelper progressHelper;
 
   protected static final NumberFormat taskIdFormat = NumberFormat.getInstance();
@@ -185,46 +184,45 @@ public class TezProcessor extends AbstractLogicalIOProcessor {
     Configuration conf = TezUtils.createConfFromUserPayload(getContext().getUserPayload());
     this.jobConf = new JobConf(conf);
     this.jobConf.getCredentials().mergeAll(UserGroupInformation.getCurrentUser().getCredentials());
-    this.processorContext = getContext();
     initTezAttributes();
-    ExecutionContext execCtx = processorContext.getExecutionContext();
+    ExecutionContext execCtx = getContext().getExecutionContext();
     if (execCtx instanceof Hook) {
       ((Hook)execCtx).initializeHook(this);
     }
-    setupMRLegacyConfigs(processorContext);
+    setupMRLegacyConfigs();
     perfLogger.perfLogEnd(CLASS_NAME, PerfLogger.TEZ_INITIALIZE_PROCESSOR);
   }
 
 
   private void initTezAttributes() {
-    jobConf.set(HIVE_TEZ_VERTEX_NAME, processorContext.getTaskVertexName());
-    jobConf.setInt(HIVE_TEZ_VERTEX_INDEX, processorContext.getTaskVertexIndex());
-    jobConf.setInt(HIVE_TEZ_TASK_INDEX, processorContext.getTaskIndex());
-    jobConf.setInt(HIVE_TEZ_TASK_ATTEMPT_NUMBER, processorContext.getTaskAttemptNumber());
+    jobConf.set(HIVE_TEZ_VERTEX_NAME, getContext().getTaskVertexName());
+    jobConf.setInt(HIVE_TEZ_VERTEX_INDEX, getContext().getTaskVertexIndex());
+    jobConf.setInt(HIVE_TEZ_TASK_INDEX, getContext().getTaskIndex());
+    jobConf.setInt(HIVE_TEZ_TASK_ATTEMPT_NUMBER, getContext().getTaskAttemptNumber());
   }
 
-  private void setupMRLegacyConfigs(ProcessorContext processorContext) {
+  private void setupMRLegacyConfigs() {
     // Hive "insert overwrite local directory" uses task id as dir name
     // Setting the id in jobconf helps to have the similar dir name as MR
     StringBuilder taskAttemptIdBuilder = new StringBuilder("attempt_");
-    taskAttemptIdBuilder.append(processorContext.getApplicationId().getClusterTimestamp())
+    taskAttemptIdBuilder.append(getContext().getApplicationId().getClusterTimestamp())
         .append("_")
-        .append(jobIdFormat.format(processorContext.getApplicationId().getId()))
+        .append(jobIdFormat.format(getContext().getApplicationId().getId()))
         .append("_");
     if (isMap) {
       taskAttemptIdBuilder.append("m_");
     } else {
       taskAttemptIdBuilder.append("r_");
     }
-    taskAttemptIdBuilder.append(taskIdFormat.format(processorContext.getTaskIndex()))
+    taskAttemptIdBuilder.append(taskIdFormat.format(getContext().getTaskIndex()))
       .append("_")
-      .append(processorContext.getTaskAttemptNumber());
+      .append(getContext().getTaskAttemptNumber());
 
     // In MR, mapreduce.task.attempt.id is same as mapred.task.id. Go figure.
     String taskAttemptIdStr = taskAttemptIdBuilder.toString();
     this.jobConf.set("mapred.task.id", taskAttemptIdStr);
     this.jobConf.set("mapreduce.task.attempt.id", taskAttemptIdStr);
-    this.jobConf.setInt("mapred.task.partition", processorContext.getTaskIndex());
+    this.jobConf.setInt("mapred.task.partition", getContext().getTaskIndex());
   }
 
   @Override
@@ -296,7 +294,7 @@ public class TezProcessor extends AbstractLogicalIOProcessor {
       // Try to call canCommit to AM. If there is no other speculative attempt execute canCommit, then continue.
       // If there are other speculative attempt execute canCommit first, then wait until the attempt is killed
       // or the committed task fails.
-      while (!this.processorContext.canCommit()) {
+      while (!getContext().canCommit()) {
         Thread.sleep(100);
       }
     } catch (Throwable t) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
@@ -295,7 +295,12 @@ public class TezProcessor extends AbstractLogicalIOProcessor {
       // If there are other speculative attempt execute canCommit first, then wait until the attempt is killed
       // or the committed task fails.
       while (!getContext().canCommit()) {
-        Thread.sleep(100);
+        // If canCommit returns false and fall into this loop, it means another task attempt has committed.
+        // And this task attempt is only needs to sleep for a relatively long time to wait for being killed.
+        // However, we need to avoid low-probability events: the rare case where a task attempt fails after
+        // committed, so we can't set an excessively long delay so that this task attempt could react on time.
+        // 500ms is a trade-off value.
+        Thread.sleep(500);
       }
     } catch (Throwable t) {
       rproc.setAborted(true);
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
@@ -181,24 +181,25 @@ public class TezProcessor extends AbstractLogicalIOProcessor {
   @Override
   public void initialize() throws IOException {
     perfLogger.perfLogBegin(CLASS_NAME, PerfLogger.TEZ_INITIALIZE_PROCESSOR);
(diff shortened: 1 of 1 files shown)


zhengchenyu:
@abstractdog Can you please review this PR? It is reasonable to use AM’s synchronization mechanism for commit file.

zhengchenyu:
@deniskuzZ Hi, can you please review the [HIVE-27985](https://issues.apache.org/jira/browse/HIVE-27985)?

zhengchenyu:
@deniskuzZ @ayushtkn @abstractdog @glapark 

Here we have the following problem about repetition in our production problem:

* (1) When speculative execution is enabled, I found that the cluster occasionally has file duplication problem.

When speculative execution is turned on, the slower task attempt will be killed. I found that the killed task may commit unfinished files. So I submit HIVE-25561 to solve this problem.

> Note: In fact, kill task attempt will cause this problem, Enabling speculative execution just increases the probability of killing task attempts.

* (2) When I apply HIVE-25561, I found that duplicated file still happened, but the probability was very very low.

I found that the duplication of files was caused by two different attempts of the same task being committed at the same time. So I submit HIVE-27899 to solve this problem.

* (3) After the task attempt is committed and before the appMaster receives the task success message, if an exception occurs, file duplication may occur.  See for  information in HIVE-27985. 

We can use task-based but not attempt-based file names. Even if there are duplicate files, since the names are the same, the file committed later will delete the file committed earlier. And because HIVE-25561 is applied, we can make sure that all commit file is generated by successfull task attempt . So I submit HIVE-27986.

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
```diff
@@ -293,6 +293,13 @@ protected void initializeAndRunProcessor(Map<String, LogicalInput> inputs,
       rproc.run();
 
       perfLogger.perfLogEnd(CLASS_NAME, PerfLogger.TEZ_RUN_PROCESSOR);
+
+      // Try to call canCommit to AM. If there is no other speculative attempt execute canCommit, then continue.
+      // If there are other speculative attempt execute canCommit first, then wait until the attempt is killed
+      // or the committed task fails.
+      while (!this.processorContext.canCommit()) {
```
deniskuzZ: There is no time interval that could potentially lead to starvation. Also would be nice to use some callback instead of Thread.sleep()

zhengchenyu: @deniskuzZ Sorry, I don't understand you. Do you means replace sleep with callback and do not block?
I try to enable speculative execution. The purpose of this is to avoid simultaneous commits. If a commit occurs simultaneously, the slower attempt will sleep, and then when the faster attempt completes, the slower sleep attempt will be killed, thus exiting the sleep loop. If the slower attempt does not block here, the slower attempt and the faster attempt may commit the file in the close method at the same time.

deniskuzZ: > the purpose of this is to avoid simultaneous commits.

is it possible to order the commit requests (use queue or synchronize the execution block)? also i don't really follow this sleep retry logic used here. 
we are trying to call `canCommit` until it returns 'true'. what if it never reports true (stuck process)?

zhengchenyu: In tez source code [TaskImpl::canCommit](https://github.com/apache/tez/blob/6683866ea5e9a7b900ccbcd617e49588974019ae/tez-dag/src/main/java/org/apache/tez/dag/app/dag/impl/TaskImpl.java#L694), we can see only one task attempt can commit successfully. If one attempt is committed, canCommit for other attempt will return false, here I sleep until the speculative attempt was killed.

In fact, tez example use this way to avoid commit simultaneously. I just copy from [here](https://github.com/apache/tez/blob/6683866ea5e9a7b900ccbcd617e49588974019ae/tez-mapreduce/src/main/java/org/apache/tez/mapreduce/processor/SimpleMRProcessor.java#L71)。

deniskuzZ: 1) 
[TaskImpl::canCommit](https://github.com/apache/tez/blob/6683866ea5e9a7b900ccbcd617e49588974019ae/tez-dag/src/main/java/org/apache/tez/dag/app/dag/impl/TaskImpl.java#L694) is already wrapped with `writeLock` - OK

however, if `commitAttempt != taskAttemptID` (speculative attempt was already committed), how are we planning to leave the loop `while (!getContext().canCommit()) {Thread.sleep(100);}`?
````
LOG.debug("{} is current committer. Commit waiting for: {}", commitAttempt, taskAttemptID);
return false;
````
I've found additional explanation in a JIRA 
````
The problem is that both speculatively executed tasks commit the file. This will not happen in the Tez
 examples because they will try canCommit, which can guarantee that one and only one task attempt commit
 successfully. If one task attempt executes canCommit successfully, the other one will be stuck by canCommit 
until it receives a kill signal.
````
so, it is intentional that 2nd attempt would be in busy-waiting until it receives a kill signal? is it a jvm process kill?
why not use backoff?
````
private void backoff() {
    nextSleep = (nextSleep > Long.MAX_VALUE / 2) ? 
        Long.MAX_VALUE : nextSleep * 2;
    try {
      Thread.sleep(nextSleep);
    }
  }
````

2)
> In fact, tez example use this way to avoid commit simultaneously. I just copy from here。

I think, usage pattern there is a bit different, but I might be wrong (no speculative execution):
````
// This will loop till the AM asks for the task to be killed. As
// against, the AM sending a signal to the task to kill itself
// gracefully. The AM waits for the current committer to successfully
// complete and then kills us. Until then we wait in case the
// current committer fails and we get chosen to commit.
while (!getContext().canCommit()) {
  Thread.sleep(100);
}
````

Comment (deniskuzZ) [inline ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java]: can we drop this `processorContext` global var and use `getContext()`as it seems unnecessary and confusing?

Comment (zhengchenyu) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java]: 1 about backoff?
I think we  should not use backoff sleep. Because in tez source code, `TaskImpl::commitAttempt` may be reset to null due to failure, etc. Sleeping for too long will affect the overall running time. And sleep is very light.

2. For me, I think no difference. In fact, we already enable speculative execution for tez engine.

Comment (zhengchenyu) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java]: It seems that it doesn't matter. If this is the case, I also removed the processorContext parameter passing to avoid unnecessary use of the stack.
However, the large amount of code modified in this PR has nothing to do with this PR. I don’t know if this is a good idea?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java]: @zhengchenyu, sleep is light. what about  `canCommit()` invocation every 100 msec? @abstractdog was proposing a capped back-off not to sleep too long
````
static final long MAX_SLEEP = 500;    
  
long nextSleep = 50
while (!processorContext.canCommit()) {
  nextSleep = backoff(nextSleep);
}

private void backoff(long nextSleep) {
  try {
    Thread.sleep(nextSleep);
  }
  if (nextSleep < MAX_SLEEP) {
    nextSleep = Math.min(MAX_SLEEP, nextSleep * 2);
  }
  return nextSleep;
}
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java [hunk 2]
```diff
@@ -185,46 +184,45 @@ public void initialize() throws IOException {
...
-    jobConf.setInt(HIVE_TEZ_TASK_INDEX, processorContext.getTaskIndex());
-    jobConf.setInt(HIVE_TEZ_TASK_ATTEMPT_NUMBER, processorContext.getTaskAttemptNumber());
+    jobConf.set(HIVE_TEZ_VERTEX_NAME, getContext().getTaskVertexName());
+    jobConf.setInt(HIVE_TEZ_VERTEX_INDEX, getContext().getTaskVertexIndex());
+    jobConf.setInt(HIVE_TEZ_TASK_INDEX, getContext().getTaskIndex());
+    jobConf.setInt(HIVE_TEZ_TASK_ATTEMPT_NUMBER, getContext().getTaskAttemptNumber());
   }
 
-  private void setupMRLegacyConfigs(ProcessorContext processorContext) {
+  private void setupMRLegacyConfigs() {
```
deniskuzZ: please keep the original param

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
deniskuzZ: since you use the context in your code, I don't see why it's a bad idea to drop the redundant global var.

zhengchenyu: If initsleep 50ms, max sleep =500ms. And after 50+100+200+400=0.75s, we will always use 0.5s to sleep. It doesn't seem to make much sense. How about configuring this variable? And set it to a value not less than a certain value, such as 100ms.

abstractdog: configuring such details seems a configuration overkill to me in the initial implementation, let's just agree in reasonable values and set it, what do you think?
my idea about 50-500 was a tradeoff about small initial and then not too large final sleep, if you have more reasonable values, I'm open to ideas

deniskuzZ: we can also change the multiplier to square root and it will take 600+ iterations to reach 500 ms (208 sec)
````	
long nextSleep = 50
...
nextSleep = (long) Math.min(MAX_SLEEP, Math.sqrt(nextSleep * nextSleep + 20 * 20)));
````

Comment (zhengchenyu) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java]: If canCommit returns false, it means another task attempt has committed. And this task attempt is likely to be killed. Therefore, we can delay it for a relatively long time for example, 500ms. In my opinion, backoff is not necessary, because in most case, this task attempt will be killed!

However, we need to avoid low-probability events: the rare case where a task attempt fails after committed, so we can't set an excessively long delay.

Thus, choosing the sleep time is crucial. It should account for low-probability events. For example, if the time from task attempt commit to task attempt failure is `t1`. Here the sleep time should be related to `t1`, `0.5 * t1` seems reasonable. Perhaps a simple sleep time of 500ms is a good choice.

This is just my personal opinion. please correct me if I'm wrong. @abstractdog @deniskuzZ

Comment (abstractdog) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java]: thanks @zhengchenyu  for the explanation, it really makes sense, I agree that it's a rare edge-case that another task starts to commit, then it fails, but if it happens, the current task should react on time, so for this purpose, max. 500ms looks good, also the exponential backoff might be an overcomplicated approach, as if canCommit == false, we can expect this task to be killed anyway, and 500ms won't overload the AM, so I'm fine with Thread.sleep(500) with an extra comment above sleep call that contains your explanation

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java]: constant 500 ms is fine by me

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java [hunk 3]
```diff
@@ -185,22 +184,21 @@ public void initialize() throws IOException {
...
     if (execCtx instanceof Hook) {
       ((Hook)execCtx).initializeHook(this);
     }
-    setupMRLegacyConfigs(processorContext);
+    setupMRLegacyConfigs(getContext());
     perfLogger.perfLogEnd(CLASS_NAME, PerfLogger.TEZ_INITIALIZE_PROCESSOR);
   }
 
 
   private void initTezAttributes() {
```
deniskuzZ: maybe parametrize with `ProcessorContext processorContext` similar to `setupMRLegacyConfigs`

zhengchenyu: hmm... The same context with setupMRLegacyConfigs. I think it is unnecessary to pass this parameter, and it wastes stack space...

deniskuzZ: 4/8 bytes to store the object ref? it's nothing
````
Use a method parameter over class property for data that is needed for a single, specific operation. 
This is transient data that defines what a particular method does, not what the object is
````
but i don't insist on that change

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java [hunk 4]
```diff
@@ -295,7 +295,12 @@ protected void initializeAndRunProcessor(Map<String, LogicalInput> inputs,
       // If there are other speculative attempt execute canCommit first, then wait until the attempt is killed
       // or the committed task fails.
       while (!getContext().canCommit()) {
-        Thread.sleep(100);
+        // If canCommit returns false and fall into this loop, it means another task attempt has committed.
+        // And this task attempt is only needs to sleep for a relatively long time to wait for being killed.
+        // However, we need to avoid low-probability events: the rare case where a task attempt fails after
+        // committed, so we can't set an excessively long delay so that this task attempt could react on time.
+        // 500ms is a trade-off value.
```
abstractdog: Nit, with minor fixes:
```
// If canCommit returns false and we enter this loop, it means another task attempt has already committed.  
// This task attempt only needs to sleep for a relatively long time while waiting to be killed.  
// However, we must avoid a low-probability scenario: a rare case where a task attempt fails after committing.  
// In that situation, the delay must not be excessively long so this attempt can still react in time.  
// 500ms is chosen as a trade-off value.  
```

zhengchenyu: Thanks, update it.

deniskuzZ (approved):
+1, pending tests

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezProcessor.java
zhengchenyu: Do you means fix like this?
```
    ProcessorContext processorContext = getContext();
    setupMRLegacyConfigs(processorContext);
```

deniskuzZ:
````
ProcessorContext processorContext = getContext();
initTezAttributes(processorContext);
ExecutionContext execCtx = processorContext.getExecutionContext();
...
setupMRLegacyConfigs(processorContext);
````

zhengchenyu: OK, I got it and I have fix.

PR #4901, opened 2023-11-27
HIVE-27911 : Drop database query failing with Invalid ACL Exception

rkirtir:
In the query-executor, we create ephemeral nodes which have the special behaviour that they are automatically deleted when the query-coordinator zookeeper client session is terminated. If there are no other Hive VWs on the cluster, sooner or later the entire path will be deleted.
To avoid this, we need to have a persistent node on path /llap-sasl/user-hivefoo0/pnode0.

Hue or beeline session thows InvalidACL exception when executing drop database, drop table or alter table drop partition operations on a Hive Virtual Warehouse that is in Stopped state: "org.apache.zookeeper.KeeperException$InvalidACLException: KeeperErrorCode = InvalidACL for /llap-sasl/user-hive".
The exception appears because the Hive VW wants to evict the cache in the LLAP executors, but those computes in a Stopped warehouse are not running.

In the query-executor, we create ephemeral nodes which have the special behaviour that they are automatically deleted when the query-coordinator zookeeper client session is terminated. If there are no other Hive VWs on the cluster, sooner or later the entire path will be deleted.
To avoid this, we need to have a persistent node on path.

NO

NO

Manually. This was tested on private cluster with crating data catalog and VWs and stopping VWS manually

### DIFF
--- a/llap-client/src/java/org/apache/hadoop/hive/llap/registry/impl/LlapZookeeperRegistryImpl.java
+++ b/llap-client/src/java/org/apache/hadoop/hive/llap/registry/impl/LlapZookeeperRegistryImpl.java
@@ -168,6 +168,16 @@ public class LlapZookeeperRegistryImpl
     String uniqueId = UNIQUE_ID.toString();
     long znodeCreationTimeout = 120;
 
+    /* When no LLAP executors are running, the getInstances()/ llapRegistryService.getInstances() method results in
+     InvalidACL exception when Hive tries to evict the cache in DROP DATABASE/TABLE
+
+     As ContainerManager of zookeeper makes sure to cleanup znodes periodically, once query-coordinator zookeeper
+     client session is terminated,entire path will get deleted sooner or later. This results in InvalidACL exception
+
+     PersistentNode created on server will be preserved through restarts of query executor and makes sure
+     proactive eviction call of DROP DATABASE/ DROP TABLE go through
+     successfully. */
+    ensurePersistentNodePath(daemonZkRecord);
     initializeWithoutRegisteringInternal();
     // Create a znode under the rootNamespace parent for this instance of the server
     try {
--- a/llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
+++ b/llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
@@ -29,6 +29,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.api.ACLProvider;
 import org.apache.curator.framework.imps.CuratorFrameworkState;
@@ -55,6 +56,7 @@ import org.apache.hadoop.registry.client.types.ServiceRecord;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.KeeperException.InvalidACLException;
 import org.apache.zookeeper.KeeperException.NodeExistsException;
 import org.apache.zookeeper.ZooDefs;
@@ -114,6 +116,8 @@ public abstract class ZkRegistryBase<InstanceType extends ServiceInstance> {
   private PersistentNode znode;
   private String znodePath; // unique identity for this instance
 
+  final String namespace;
+
   private PathChildrenCache instancesCache; // Created on demand.
 
   /** Local hostname. */
@@ -160,7 +164,7 @@ public abstract class ZkRegistryBase<InstanceType extends ServiceInstance> {
     this.stateChangeListeners = new HashSet<>();
     this.pathToInstanceCache = new ConcurrentHashMap<>();
     this.nodeToInstanceCache = new ConcurrentHashMap<>();
-    final String namespace = getRootNamespace(conf, rootNs, nsPrefix);
+    this.namespace = getRootNamespace(conf, rootNs, nsPrefix);
     ACLProvider aclProvider;
     // get acl provider for most outer path that is non-null
     if (userPathPrefix == null) {
@@ -353,6 +357,30 @@ public abstract class ZkRegistryBase<InstanceType extends ServiceInstance> {
     }
   }
 
+  @VisibleForTesting
+  public String getPersistentNodePath() {
+    return "/" + PATH_JOINER.join(namespace, StringUtils.substringBetween(workersPath, "/", "/"), "pnode0");
+  }
+
+  protected void ensurePersistentNodePath(ServiceRecord srv) throws IOException {
+    String pNodePath = getPersistentNodePath();
+    try {
+      LOG.info("Check if persistent node  path {}  exists, create if not", pNodePath);
+      if (zooKeeperClient.checkExists().forPath(pNodePath) == null) {
+        zooKeeperClient.create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT)
+            .forPath(pNodePath, encoder.toBytes(srv));
+        LOG.info("Created persistent path at: {}", pNodePath);
+      }
+    } catch (Exception e) {
+      // throw exception if it is other than NODEEXISTS.
+      if (!(e instanceof KeeperException) || ((KeeperException) e).code() != KeeperException.Code.NODEEXISTS) {
+        LOG.error("Unable to create a persistent znode for this server instance", e);
+        throw new IOException(e);
+      } else {
+        LOG.debug("Ignoring KeeperException while ensuring path as the parent node {} already exists.", pNodePath);
+      }
+    }
+  }
 
   final protected void initializeWithoutRegisteringInternal() throws IOException {
     // Create a znode under the rootNamespace parent for this instance of the server
--- a/llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
+++ b/llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
@@ -29,6 +29,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.api.ACLProvider;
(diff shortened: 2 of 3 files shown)


rkirtir:
Hi @abstractdog, Requesting your review here.

llap-client/src/java/org/apache/hadoop/hive/llap/registry/impl/LlapZookeeperRegistryImpl.java
```diff
@@ -168,6 +168,10 @@ public String register() throws IOException {
     String uniqueId = UNIQUE_ID.toString();
     long znodeCreationTimeout = 120;
 
+    // PersistentNode created on server will be preserved on VW restarts which will make
```
abstractdog, line 171: "VW restart" is a Cloudera hive ware warehouse specific thing, let's try to explain this in another way which describes our scenario

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -114,6 +116,7 @@ public abstract class ZkRegistryBase<InstanceType extends ServiceInstance> {
   private PersistentNode znode;
   private String znodePath; // unique identity for this instance
 
+  final String namespace;
```
abstractdog, line 119: could namespace be final?

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -353,6 +356,23 @@ protected final void updateServiceRecord(
     }
   }
 
+  protected void ensurePersistentNodePath(ServiceRecord srv) throws IOException {
+    try {
+      String pNodePath = "/" + PATH_JOINER.join(namespace, StringUtils.substringBetween(workersPath, "/", "/"),
+          "pnode0");
+      LOG.info("Check if persistent node  path{}  exists, create if not", pNodePath);
```
abstractdog, line 363: whitespace is strange here, this should be:
```
Check if persistent node path {} exists, create if not
```

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -353,6 +356,23 @@ protected final void updateServiceRecord(
...
+      if (zooKeeperClient.checkExists().forPath(pNodePath) == null) {
+        zooKeeperClient.create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT)
+            .forPath(pNodePath, encoder.toBytes(srv));
+        LOG.info("Created persistent path at: {}", pNodePath);
+      }
+    } catch (Exception e) {
+      if (!(e instanceof KeeperException) || ((KeeperException) e).code() != KeeperException.Code.NODEEXISTS) {
+        LOG.error("Unable to create a persistent znode for this server instance", e);
+        throw new IOException(e);
+      }
```
abstractdog, line 373: what other exception can happen here?
I believe we should be future-proof by doing:
```
      if (!(e instanceof KeeperException) || ((KeeperException) e).code() != KeeperException.Code.NODEEXISTS) {
        LOG.error("Unable to create a persistent znode for this server instance", e);
        throw new IOException(e);
      } else{
         do something here, at least LOG.info
      }
```

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -353,6 +356,23 @@ protected final void updateServiceRecord(
...
+      String pNodePath = "/" + PATH_JOINER.join(namespace, StringUtils.substringBetween(workersPath, "/", "/"),
+          "pnode0");
+      LOG.info("Check if persistent node  path{}  exists, create if not", pNodePath);
+      if (zooKeeperClient.checkExists().forPath(pNodePath) == null) {
+        zooKeeperClient.create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT)
+            .forPath(pNodePath, encoder.toBytes(srv));
+        LOG.info("Created persistent path at: {}", pNodePath);
+      }
+    } catch (Exception e) {
+      if (!(e instanceof KeeperException) || ((KeeperException) e).code() != KeeperException.Code.NODEEXISTS) {
```
abstractdog, line 370: can you put a comment here telling what is this condition about?

llap-client/src/java/org/apache/hadoop/hive/llap/registry/impl/LlapZookeeperRegistryImpl.java
aturoczy: Just this comment is not proper for upstream

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -353,6 +356,23 @@ protected final void updateServiceRecord(
     }
   }
 
+  protected void ensurePersistentNodePath(ServiceRecord srv) throws IOException {
+    try {
+      String pNodePath = "/" + PATH_JOINER.join(namespace, StringUtils.substringBetween(workersPath, "/", "/"),
```
aturoczy, line 361: Is it possible to create a unit test for this join? This one line is bit overloaded with functions and results and better to have a validation against it

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
rkirtir: fixed

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
rkirtir: fixed

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
rkirtir: Fixed. curator framework.create call throws KeeperException.NODEExsists  exception, so apart from that evrything has been rethrown.

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
rkirtir: done

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -353,6 +356,26 @@ protected final void updateServiceRecord(
...
+        zooKeeperClient.create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT)
+            .forPath(pNodePath, encoder.toBytes(srv));
+        LOG.info("Created persistent path at: {}", pNodePath);
+      }
+    } catch (Exception e) {
+      // throw exception if it is other than NODEEXISTS.
+      if (!(e instanceof KeeperException) || ((KeeperException) e).code() != KeeperException.Code.NODEEXISTS) {
+        LOG.error("Unable to create a persistent znode for this server instance", e);
+        throw new IOException(e);
+      }else{
```
abstractdog, line 374: nit:
```
}else{
```
should be:
```
} else {
```
mind the spaces, btw you can auto-format according to this style definition:
https://github.com/apache/hive/blob/master/dev-support/eclipse-styles.xml

another nit:
```
LOG.info( "Ignore as the parent node already exists.");
```
will result an INFO level message without any context: Ignore "something" as the parent node of "something" already exists, I would switch to this message (and I realized DEBUG is fine as else branch is called under normal circumstances)
```
LOG.debug( "Ignoring KeeperException while ensuring path as the parent node {} already exists.", parentNode);
```

llap-client/src/java/org/apache/hadoop/hive/llap/registry/impl/LlapZookeeperRegistryImpl.java
```diff
@@ -168,6 +168,16 @@ public String register() throws IOException {
     String uniqueId = UNIQUE_ID.toString();
     long znodeCreationTimeout = 120;
 
+    // When no LLAP executors are running, the getInstances()/ llapRegistryService.getInstances() method results in
```
abstractdog, line 171: please use a slash-star comment block for comments made up of multiple lines

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -114,6 +116,7 @@ public abstract class ZkRegistryBase<InstanceType extends ServiceInstance> {
   private PersistentNode znode;
   private String znodePath; // unique identity for this instance
 
+  private String namespace;
```
abstractdog, line 119: this is supposed to be final as agreed in previous comment

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
abstractdog: I agree, this code area needs more clarity than hacking something with "workersPath" and joining it to namespace with slashes
an unit test is much better then a code comment in this case

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
rkirtir: fixed

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
rkirtir: fixed

llap-client/src/test/org/apache/hadoop/hive/llap/registry/impl/TestLlapZookeeperRegistryImpl.java
```diff
@@ -124,6 +126,17 @@ public void testUpdate() throws Exception {
             attributes.get(LlapRegistryService.LLAP_DAEMON_NUM_ENABLED_EXECUTORS));
   }
 
+  @Test
+  public void testPersistentNodePath() {
+
```
abstractdog, line 131: minor nit, sorry, we usually don't put a blank line as the first line of a method (you'll have to push again as tests failed unfortunately, sigh)

rkirtir: fixed

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/dbinstall/TestMariadb.java
```diff
@@ -0,0 +1,33 @@
+/*
```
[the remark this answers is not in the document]
rkirtir: sorry, my bad. these changes got pushed when i switched branch on local. Fixed

abstractdog (approved):
LGTM

PR #4902, opened 2023-11-27
Multi-stream JDBC changes

kdeschle:

Performance improvements to the Hive JDBC driver to allow for high performance fetching.

HIVE-28018: Don't require HiveConf for JDBC Metadata calls [Kurt Deschler]
HIVE-27886: Configure Hadoop security for Kerberos when required [Michael Smith]
HIVE-27887: Provide reasonable defaults for ResultSetMetaData [Michael Smith]
HIVE-27873: Fix getOperationStatus() handling and optimize result fetch path [Kurt Deschler]
HIVE-27872: Support multi-stream parallel fetch in JDBC driver [Kurt Deschler]
HIVE-27874: Support datatype conversion on fetch threads [Kurt Deschler]

JDBC performance with large datasets was very slow, especially when connected to Impala via HS2 protocol

-new JDBC connection parameter 'fetchThreads'
-fetchSize is enforce for prepared statements
-JDBC standalone JAR can be dynamically loaded
-custom hiveconf is no longer required to use the JDBC jar

No

Ran existing JDBC regression tests
Manual testing multi-stream fetch against Impala

### DIFF
--- a/jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
+++ b/jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
@@ -1,0 +1,135 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hive.jdbc;
+
+import java.math.BigDecimal;
+import java.sql.Date;
+import java.sql.Timestamp;
+import java.util.function.UnaryOperator;
+import java.util.ArrayList;
+import java.util.EnumMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.hadoop.hive.common.type.HiveIntervalDayTime;
+import org.apache.hadoop.hive.common.type.HiveIntervalYearMonth;
+import org.apache.hadoop.hive.common.type.TimestampTZUtil;
+
+import org.apache.hadoop.hive.serde2.thrift.Type;
+
+import org.apache.hive.service.rpc.thrift.TRowSet;
+import org.apache.hive.service.cli.RowSet;
+import org.apache.hive.service.cli.TableSchema;
+
+/**
+ * ConvertedResultSet
+ */
+public class ConvertedResultSet implements RowSet {
+
+  private final List<Object[]> rows;
+  private final int numColumns;
+
+  static final Map<Type, UnaryOperator<Object>> convertFuncs = new EnumMap<Type, UnaryOperator<Object>>(Type.class) {{
+    put(Type.BINARY_TYPE, o -> (o instanceof String) ? ((String) o).getBytes() : 0);
+    put(Type.TIMESTAMP_TYPE, o -> Timestamp.valueOf((String) o));
+    put(Type.TIMESTAMPLOCALTZ_TYPE, o -> TimestampTZUtil.parse((String) o));
+    put(Type.DECIMAL_TYPE, o -> new BigDecimal((String) o));
+    put(Type.DATE_TYPE, o -> Date.valueOf((String) o));
+    put(Type.INTERVAL_YEAR_MONTH_TYPE, o -> HiveIntervalYearMonth.valueOf((String) o));
+    put(Type.INTERVAL_DAY_TIME_TYPE, o -> HiveIntervalDayTime.valueOf((String) o));
+  }};
+
+
+  public ConvertedResultSet(RowSet rowSet, TableSchema schema) {
+    rows = new ArrayList<>();
+    numColumns = rowSet.numColumns();
+    Iterator<Object[]> srcIter = rowSet.iterator();
+    final int nCols = schema.getSize();
+    
+    UnaryOperator[] colConvert = new UnaryOperator[nCols];
+    for (int i = 0; i < nCols; ++i) {
+      colConvert[i] = convertFuncs.get(schema.getColumnDescriptorAt(i).getType());
+    }
+    while (srcIter.hasNext()) {
+      Object[] srcRow = srcIter.next();
+      Object[] dstRow = new Object[srcRow.length];
+      for (int i = 0; i < nCols; ++i) {
+        if (colConvert[i] != null && srcRow[i] != null) {
+          dstRow[i] = colConvert[i].apply(srcRow[i]);
+        } else {
+          dstRow[i] = srcRow[i];
+        }
+      }
+      rows.add(dstRow);
+    }
+  }
+
+  @Override
+  public ConvertedResultSet addRow(Object[] fields) {
+    throw new UnsupportedOperationException("addRow");
+  }
+
+  @Override
+  public int numColumns() {
+    return numColumns;
+  }
+
+  @Override
+  public int numRows() {
+    return rows.size();
+  }
+
+  public ConvertedResultSet extractSubset(int maxRows) {
+    throw new UnsupportedOperationException("extractSubset");
+  }
+
+  @Override
+  public long getStartOffset() {
+    throw new UnsupportedOperationException("getStartOffset");
+  }
+
+  @Override
+  public void setStartOffset(long startOffset) {
+    throw new UnsupportedOperationException("setStartOffset");
+  }
+
+  @Override
+  public TRowSet toTRowSet() {
+    throw new UnsupportedOperationException("toTRowSet");
+  }
+
+  @Override
+  public Iterator<Object[]> iterator() {
+    return new Iterator<Object[]>() {
+
+      final Iterator<Object[]> iterator = rows.iterator();
+
+      @Override
+      public boolean hasNext() {
+        return iterator.hasNext();
+      }
+
+      @Override
+      public Object[] next() {
+        return iterator.next();
+      }
+    };
+  }
+}
--- a/jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
+++ b/jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
@@ -19,6 +19,7 @@
 package org.apache.hive.jdbc;
 
 import static org.apache.hadoop.hive.conf.Constants.MODE;
+import static org.apache.hive.service.auth.HiveAuthConstants.AuthTypes;
 import static org.apache.hive.service.cli.operation.hplsql.HplSqlQueryExecutor.HPLSQL;
 
 import java.io.BufferedReader;
@@ -106,6 +107,7 @@ import org.apache.hive.service.auth.HiveAuthConstants;
 import org.apache.hive.service.auth.KerberosSaslHelper;
 import org.apache.hive.service.auth.PlainSaslHelper;
 import org.apache.hive.service.auth.SaslQOP;
+
 import org.apache.hive.service.cli.session.SessionUtils;
 import org.apache.hive.service.rpc.thrift.TCLIService;
 import org.apache.hive.service.rpc.thrift.TCancelDelegationTokenReq;
@@ -177,12 +179,15 @@ public class HiveConnection implements java.sql.Connection {
   private int loginTimeout = 0;
   private TProtocolVersion protocol;
   int fetchSize;
+  int fetchThreads;
   private String initFile = null;
   private String wmPool = null, wmApp = null;
   private Properties clientInfo;
   private Subject loggedInSubject;
   private int maxRetries = 1;
   private IJdbcBrowserClient browserClient;
+
+  public TCLIService.Iface getClient() { return client; }
 
   /**
    * Get all direct HiveServer2 URLs from a ZooKeeper based HiveServer2 URL
@@ -279,10 +284,24 @@ public class HiveConnection implements java.sql.Connection {
     sessConfMap = null;
     isEmbeddedMode = true;
     fetchSize = 50;
+    fetchThreads = 0;
   }
 
   public HiveConnection(String uri, Properties info) throws SQLException {
     this(uri, info, HiveJdbcBrowserClientFactory.get());
+  }
+
+  /**
+   * Create a new connection that shares the same session ID as the current connection.
+   */
+  public HiveConnection(HiveConnection hiveConnection) throws SQLException {
+    this(hiveConnection.getConnectedUrl(), hiveConnection.getClientInfo(), HiveJdbcBrowserClientFactory.get(), false);
+    // These are set/updated when the session is established.
+    this.sessHandle = hiveConnection.sessHandle;
+    this.connParams = hiveConnection.connParams;
+    this.protocol = hiveConnection.protocol;
+    this.fetchSize = hiveConnection.fetchSize;
+    this.fetchThreads = hiveConnection.fetchThreads;
   }
 
   @VisibleForTesting
@@ -293,6 +312,12 @@ public class HiveConnection implements java.sql.Connection {
   @VisibleForTesting
   protected HiveConnection(String uri, Properties info,
       IJdbcBrowserClientFactory browserClientFactory) throws SQLException {
+    this(uri, info, browserClientFactory, true);
+  }
+
+  protected HiveConnection(String uri, Properties info,
+      IJdbcBrowserClientFactory browserClientFactory,
+      boolean initSession) throws SQLException {
     try {
       connParams = Utils.parseURL(uri, info);
     } catch (ZooKeeperHiveClientException e) {
@@ -311,7 +336,7 @@ public class HiveConnection implements java.sql.Connection {
       // Ensure UserGroupInformation includes any authorized Kerberos principals.
       LOG.debug("Configuring Kerberos mode");
       Configuration config = new Configuration();
-      config.set("hadoop.security.authentication", "Kerberos");
+      config.set("hadoop.security.authentication", AuthTypes.KERBEROS.getAuthName());
       UserGroupInformation.setConfiguration(config);
 
       if (isEnableCanonicalHostnameCheck()) {
@@ -365,8 +390,10 @@ public class HiveConnection implements java.sql.Connection {
         throw new SQLException(new IllegalArgumentException(
             "Browser mode is not supported in embedded mode"));
       }
-      openSession();
-      executeInitSql();
+      if (initSession) {
+        openSession();
+        executeInitSql();
+      }
     } else {
       long retryInterval = 1000L;
       try {
@@ -388,8 +415,10 @@ public class HiveConnection implements java.sql.Connection {
           // set up the client
           client = new TCLIService.Client(new TBinaryProtocol(transport));
           // open client session
-          openSession();
-          executeInitSql();
+          if (initSession) {
+            openSession();
+            executeInitSql();
+          }
 
           break;
         } catch (Exception e) {
@@ -1255,10 +1284,10 @@ public class HiveConnection implements java.sql.Connection {
     protocol = openResp.getServerProtocolVersion();
     sessHandle = openResp.getSessionHandle();
 
-    ConfVars confVars = ConfVars.HIVE_SERVER2_THRIFT_RESULTSET_DEFAULT_FETCH_SIZE;
-    int serverFetchSize = Optional.ofNullable(openResp.getConfiguration().get(confVars.varname))
+    ConfVars fetchSizeConf = ConfVars.HIVE_SERVER2_THRIFT_RESULTSET_DEFAULT_FETCH_SIZE;
+    int serverFetchSize = Optional.ofNullable(openResp.getConfiguration().get(fetchSizeConf.varname))
         .map(size -> Integer.parseInt(size))
-        .orElse(confVars.defaultIntVal);
+        .orElse(fetchSizeConf.defaultIntVal);
     if (serverFetchSize <= 0) {
       throw new IllegalStateException("Default fetch size must be greater than 0");
     }
@@ -1266,6 +1295,19 @@ public class HiveConnection implements java.sql.Connection {
         .map(size -> Integer.parseInt(size))
         .filter(v -> v > 0)
         .orElse(serverFetchSize);
+
+    ConfVars fetchThreadsConf = ConfVars.HIVE_JDBC_FETCH_THREADS;
+    int serverFetchThreads = Optional.ofNullable(openResp.getConfiguration().get(fetchThreadsConf.varname))
+        .map(size -> Integer.parseInt(size))
+        .orElse(fetchThreadsConf.defaultIntVal);
+    if (serverFetchThreads <= 0) {
+      throw new IllegalStateException("Default fetch threads must be >= 0");
+    }
+    this.fetchThreads = Optional.ofNullable(sessConfMap.get(JdbcConnectionParams.FETCH_THREADS))
+        .map(size -> Integer.parseInt(size))
+        .filter(v -> v >= 0)
+        .orElse(serverFetchThreads);
+
   }
 
   /**
@@ -1582,7 +1624,7 @@ public class HiveConnection implements java.sql.Connection {
     if (isClosed) {
       throw new SQLException("Can't create Statement, connection is closed");
     }
-    return new HiveStatement(this, client, sessHandle, false, fetchSize);
+    return new HiveStatement(this, client, sessHandle, false, fetchSize, fetchThreads);
   }
 
   /*
@@ -1605,7 +1647,7 @@ public class HiveConnection implements java.sql.Connection {
     if (isClosed) {
       throw new SQLException("Connection is closed");
     }
-    return new HiveStatement(this, client, sessHandle, resultSetType == ResultSet.TYPE_SCROLL_INSENSITIVE, fetchSize);
+    return new HiveStatement(this, client, sessHandle, resultSetType == ResultSet.TYPE_SCROLL_INSENSITIVE);
   }
 
   /*
--- a/jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
+++ b/jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
@@ -84,6 +84,7 @@ import org.apache.hadoop.hive.common.auth.HiveAuthUtils;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.security.Credentials;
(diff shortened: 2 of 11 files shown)


jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
```diff
@@ -285,6 +289,16 @@ public HiveConnection(String uri, Properties info) throws SQLException {
     this(uri, info, HiveJdbcBrowserClientFactory.get());
   }
 
+  public HiveConnection Clone() throws SQLException {
```
aturoczy, line 292: In which use case the Clone is needed? It is not better just to create another connection with the same connection string?

jdbc/src/java/org/apache/hive/jdbc/HiveResultSetMetaData.java
```diff
@@ -30,6 +30,8 @@
  *
  */
 public class HiveResultSetMetaData implements java.sql.ResultSetMetaData {
+  private static final String DOT = ".";
```
aturoczy, line 33: It is a bit overloaded to me to create a constant for DOT that is .  - I'm fine with the  . :-) 

cc: @ayushtkn @deniskuzZ Is this a design rule to make this static final? Even in the simplest cases?

aturoczy:
The test failed with incorrect POM :(

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
aturoczy, line 292: Alos the clone is small c :)

jdbc/pom.xml
```diff
@@ -302,6 +295,8 @@
                       <exclude>org.apache.hadoop:hadoop-annotations</exclude>
                       <exclude>org.apache.hadoop:hadoop-auth</exclude>
                       <exclude>org.apache.hbase*:*</exclude>
+                      <exclude>org.apache.hive:hive-classification</exclude>  
+                      <exclude>org.apache.hive:hive-storage-api</exclude>
```
MikaelSmith, line 299: These have extra trailing whitespace.

```suggestion
                      <exclude>org.apache.hive:hive-classification</exclude>
                      <exclude>org.apache.hive:hive-storage-api</exclude>
```

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
MikaelSmith: It ensures both clients share a session ID.

MikaelSmith: I'm not sure overriding Object#clone makes sense. I've added a comment to describe it's behavior.

jdbc/src/java/org/apache/hive/jdbc/HiveResultSetMetaData.java
deniskuzZ: all constants should be marked as `static final`

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
deniskuzZ: if you don't won't to override Object#clone, create a copy constructor

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
```diff
@@ -307,6 +333,12 @@ protected HiveConnection(String uri, Properties info,
     sessConfMap = connParams.getSessionVars();
     setupLoginTimeout();
     if (isKerberosAuthMode()) {
+      // Ensure UserGroupInformation includes any authorized Kerberos principals.
+      LOG.debug("Configuring Kerberos mode");
+      Configuration config = new Configuration();
+      config.set("hadoop.security.authentication", "Kerberos");
```
deniskuzZ, line 339: use `AuthTypes.KERBEROS.getAuthName`

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
```diff
@@ -293,12 +311,20 @@ protected int getNumRetries() {
   @VisibleForTesting
   protected HiveConnection(String uri, Properties info,
       IJdbcBrowserClientFactory browserClientFactory) throws SQLException {
+    this(uri, info, browserClientFactory, true);
+  }
+
+  @VisibleForTesting
```
deniskuzZ, line 317: why do we need this annotation, I don't see use in tests for the new constructor

jdbc/src/java/org/apache/hive/jdbc/Utils.java
```diff
@@ -166,6 +166,7 @@ public static class JdbcConnectionParams {
     // Set the fetchSize
     static final String FETCH_SIZE = "fetchSize";
     static final String INIT_FILE = "initFile";
+    static final String FETCH_THREADS = "fetchThreads";
```
deniskuzZ, line 169: FETCH_THREADS_COUNT = "fetchThreadCount";

jdbc/src/java/org/apache/hive/jdbc/JdbcColumn.java
```diff
@@ -372,6 +372,20 @@ static int columnScale(Type hiveType, JdbcColumnAttributes columnAttributes)
     }
   }
 
+  static boolean columnSigned(Type hiveType, JdbcColumnAttributes columnAttributes)
+      throws SQLException {
+    int columnType = hiveTypeToSqlType(hiveType);
+    switch(columnType) {
+    case Types.TINYINT:
```
deniskuzZ, line 379: please format the code

jdbc/src/java/org/apache/hive/jdbc/HiveStatement.java
```diff
@@ -73,12 +74,16 @@ public class HiveStatement implements java.sql.Statement {
   private static final int DEFAULT_FETCH_SIZE =
       HiveConf.ConfVars.HIVE_SERVER2_THRIFT_RESULTSET_DEFAULT_FETCH_SIZE.defaultIntVal;
 
+// TODO: Should there be a hiveconf default?
+  private static final int DEFAULT_FETCH_THREADS = 1;
```
deniskuzZ, line 78: i think so

jdbc/src/java/org/apache/hive/jdbc/HiveResultSetMetaData.java
```diff
@@ -43,7 +45,9 @@ public HiveResultSetMetaData(List<String> columnNames,
   }
 
   public String getCatalogName(int column) throws SQLException {
-    throw new SQLFeatureNotSupportedException("Method not supported");
+    // Hive doesn't implement a catalog concept, see
+    // https://issues.apache.org/jira/browse/HIVE-3121.
+    return "";
```
deniskuzZ, line 50: should be `default`

jdbc/src/java/org/apache/hive/jdbc/HiveResultSetMetaData.java
```diff
@@ -95,11 +99,23 @@ public int getScale(int column) throws SQLException {
...
+      return tableName.substring(0, index);
+    }
+    // Impala usually doesn't return fully qualified column names. Return "" to avoid
+    // giving false results.
+    return "";
   }
 
   public String getTableName(int column) throws SQLException {
-    throw new SQLFeatureNotSupportedException("Method not supported");
+    String columnName = getColumnName(column);
```
deniskuzZ, line 113: same, use `HiveTableName.of(tableName).getTable()`

jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * ConvertedResultSet
+ */
+public class ConvertedResultSet implements RowSet {
+
+  private final ArrayList<Object[]> rows;
+  private final int numColumns;
+
+  private interface ConvertFunc {
```
deniskuzZ, line 47: why not use `UnaryOperator`?

jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
```diff
@@ -0,0 +1,151 @@
...
+    Object convert(Object o);
+  }
+
+  static final int nTypes = Type.values().length;
+  static final ConvertFunc convertFuncs [] = new ConvertFunc [nTypes];
```
deniskuzZ, line 52: why not use Map?
````
  static final Map<Type, UnaryOperator<?>> convertFuncs = new HashMap<Type, UnaryOperator<?>>() {{
      put(Type.BINARY_TYPE, o -> (o instanceof String) ? ((String) o).getBytes() : 0);
      ....
    }};
````

File: jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java [hunk 13]
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.hive.service.cli.RowSet;
+import org.apache.hive.service.cli.TableSchema;
+import org.apache.hive.service.cli.TypeDescriptor;
+
+/**
+ * ConvertedResultSet
+ */
+public class ConvertedResultSet implements RowSet {
+
+  private final ArrayList<Object[]> rows;
```
deniskuzZ, line 44: please use interface (code to interface, not an implementation)

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
```diff
@@ -360,8 +393,10 @@ protected HiveConnection(String uri, Properties info,
         throw new SQLException(new IllegalArgumentException(
             "Browser mode is not supported in embedded mode"));
       }
-      openSession();
-      executeInitSql();
+      if (initSession) {
```
deniskuzZ, line 396: could we extract this into a helper method?

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -98,6 +109,7 @@ public static class Builder {
     private List<String> colTypes;
     private List<JdbcColumnAttributes> colAttributes;
     private int fetchSize = 50;
+    private int fetchThreads = 1;
```
deniskuzZ, line 112: should we get this from HiveConf?

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -271,8 +299,18 @@ private void setSchema(List<String> colNames, List<String> colTypes,
...
+      fetchedRows = RowSetFactory.create(new TRowSet(), protocol);
+      fetchedRowsItr = fetchedRows.iterator();
+    } catch (TException e) {
+      throw new SQLException(e);
+    }
+  }
+
   @Override
   public void close() throws SQLException {
+    ShutdownPool();
```
deniskuzZ, line 313: please use `camel-case` namings

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -290,7 +328,7 @@ public void close() throws SQLException {
     client = null;
     stmtHandle = null;
     isClosed = true;
-    operationStatus = null;
+    InitEmptyIterator();
```
deniskuzZ, line 331: `camel-case`

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
 
-  /**
-   * Moves the cursor down one row from its current position.
-   *
-   * @see java.sql.ResultSet#next()
-   * @throws SQLException
-   *           if a database access error occurs.
-   */
-  public boolean next() throws SQLException {
+  ExecutorService pool = null;
```
deniskuzZ, line 348: could you please move it to the rest or class fields declarations

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
+          gotLastBatch = true;
+        }
+        fetchedRows = RowSetFactory.create(results, protocol);
+        if (useConvertedResultSet) {
+          fetchedRows = new ConvertedResultSet(fetchedRows, getSchema());
+        }
+
+        nextStartRow.set(results.getStartRowOffset() + 1 + fetchedRows.numRows());
+      } catch (TException ex) {
+        ex.printStackTrace();
```
deniskuzZ, line 408: use `Logger` instead

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
+        throw new SQLException("Error retrieving next row", ex);
+      }
+    } else {
+      if (!gotLastBatch) {
+        if (pool == null) {
+          pool = Executors.newFixedThreadPool(poolSize);
+        }
+        // Add another thread on each row batch up to the limit
+        if(threadsStarted < (hasStartRow ? poolSize : 1)) {
```
deniskuzZ, line 417: nit: space

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
+          final boolean useMainClient = (threadsStarted == 0);
+          threadsStarted++;
+          pool.execute(()->{
```
deniskuzZ, line 420: nit: space after `->`

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
+            }
+            // TODO: shutdown called here can make execute return
+            // RejectedExecutionException
+            // pool.shutdownNow();
+          });
+        }
+      }
+      try {
+        if (interruptException != null) {
+          throw interruptException;
```
deniskuzZ, line 500: show we `shutdown` the pool?

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
+        }
+        FetchResult result = resultQueue.take();
+        fetchDone = !result.hasMoreRows;
+        if (result.ex != null) {
+          ShutdownPool();
+          throw new SQLException(result.ex);
+        }
+        fetchedRows = result.fetchedRows;
+      } catch (InterruptedException e) {
```
deniskuzZ, line 509: should we `shutdown` the pool in `finally` block?

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
 
-      if (!fetchedRowsItr.hasNext()) {
-        return false;
+    orientation = TFetchOrientation.FETCH_NEXT;
+    fetchedRowsItr = fetchedRows.iterator();
+
+    return fetchedRowsItr.hasNext();
+  }
+
+  void DrainQueue() throws SQLException {
```
deniskuzZ, line 520: `camel-case`

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
+    }
+  }
 
-      row = fetchedRowsItr.next();
-      rowsFetched++;
-    } catch (SQLException eS) {
-      throw eS;
-    } catch (Exception ex) {
-      throw new SQLException("Error retrieving next row", ex);
+  void ShutdownPool() throws SQLException {
```
deniskuzZ, line 529: `camel-case`

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
+    if (pool != null) {
+      poolDone = true;
+      DrainQueue();
```
deniskuzZ, line 532: do we need to call it before pool shutdown?

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
```diff
@@ -307,66 +345,220 @@ private void closeOperationHandle(TOperationHandle stmtHandle) throws SQLExcepti
...
+      pool.shutdownNow();
+      try {
+        while (!pool.awaitTermination(1, TimeUnit.SECONDS)) {
+          DrainQueue();
```
deniskuzZ, line 536: why do you need to call it multiple times?

deniskuzZ:
thanks for the PR @MikaelSmith , @kdeschle! In general LGTM, left few comments

jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
kdeschle: That leads to generic array creation compile error.

jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
kdeschle: Map would be less efficient than the current ordinal lookup.

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
kdeschle: I don't think it makes sense to add a function here.

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
kdeschle: Not here. This is just early shutdown for error case. Unconditional shutdown is from close()

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
kdeschle: This sequence is necessary to ensure that nothing blocks and nothing is left in the queue.

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
kdeschle: This was part of prescribed strategies for shutting down ExecutorService.

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
kdeschle: switched to copy constructor

jdbc/src/java/org/apache/hive/jdbc/HiveQueryResultSet.java
kdeschle: Have to restore initial state in case result is re-opened.

deniskuzZ:
hi @kdeschle, thanks for addressing the review comments! 
Would you be able to rebase this PR?

jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java
kdeschle: changed constant.
I don't thing it should  always be kerberized.

jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
deniskuzZ: That is the usage pattern for the HashTable - do the lookups. I believe this code part does not need to be optimized at the expense of readability.
````
  static final Map<Type, UnaryOperator<Object>> convertFuncs = new HashMap<Type, UnaryOperator<Object>>() {{
    put(Type.BINARY_TYPE, o -> (o instanceof String) ? ((String) o).getBytes() : 0);
    put(Type.TIMESTAMP_TYPE, o -> Timestamp.valueOf((String) o));
    put(Type.TIMESTAMPLOCALTZ_TYPE, o -> TimestampTZUtil.parse((String) o));
    put(Type.DECIMAL_TYPE, o -> new BigDecimal((String) o));
    put(Type.DATE_TYPE, o -> Date.valueOf((String) o));
    put(Type.INTERVAL_YEAR_MONTH_TYPE, o -> HiveIntervalYearMonth.valueOf((String) o));
    put(Type.INTERVAL_DAY_TIME_TYPE, o -> HiveIntervalDayTime.valueOf((String) o));
  }};

  public ConvertedResultSet(RowSet rowSet, TableSchema schema) {
    rows = new ArrayList<>();
    numColumns = rowSet.numColumns();
    Iterator<Object[]> srcIter = rowSet.iterator();
    final int nCols = schema.getSize();
    
    List<UnaryOperator<Object>> colConvert = new ArrayList<>(nCols);
    for (int i = 0; i < nCols; ++i) {
      colConvert.add(convertFuncs.get(schema.getColumnDescriptorAt(i).getType()));
    }
    
    while (srcIter.hasNext()) {
      Object[] srcRow = srcIter.next();
      Object[] dstRow = new Object[srcRow.length];
      for (int i = 0; i < nCols; ++i) {
        if (colConvert.get(i) != null && srcRow[i] != null) {
          dstRow[i] = colConvert.get(i).apply(srcRow[i]);
        } else {
          dstRow[i] = srcRow[i];
        }
      }
      rows.add(dstRow);
    }
  }
````

File: jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java [hunk 28]
```diff
@@ -0,0 +1,151 @@
...
+public class ConvertedResultSet implements RowSet {
+
+  private final List<Object[]> rows;
+  private final int numColumns;
+
+  private interface ConvertFunc {
+    Object convert(Object o);
+  }
+
+  static final int nTypes = Type.values().length;
```
deniskuzZ, line 51: `nTypes` variable can be removed

jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
deniskuzZ: it doesn't, please see snippet in `convertFuncs` comment

jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
```diff
@@ -0,0 +1,151 @@
...
+        return iterator.hasNext();
+      }
+
+      @Override
+      public Object[] next() {
+        return iterator.next();
+      }
+
+      @Override
+      public void remove() {
```
deniskuzZ, line 146: no need to override, it's identical to parent method

deniskuzZ:
In general LGTM. Would be great to include this perf improvement in the next 4.1.0 Hive release. 

I've reopened 1 comment with the refactor recommendation:  https://github.com/apache/hive/pull/4902#discussion_r1886898843

Also, Sonar reported a few new issues, we should try to address those.

jdbc/src/java/org/apache/hive/jdbc/ConvertedResultSet.java
kdeschle: With this change there was a 1.1% decrease in performance. I can keep the change if you insist.

deniskuzZ: @kdeschle what Map implementation have you used? EnumMap to HashMap?
EnumMap is 1-to-1 match with your solution (vals array with key.ordinal() index access)

1.1% decrease in generic performance performance or collection access?

kdeschle: Nevermind. It seems it was just variation in my runs. Posted new rev with these changes and some fixes.

deniskuzZ (approved):
+1, pending green build

deniskuzZ:
Sonar reported a number of newly introduced unused import, methods and vars, could we please drop them:
1. https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4902&id=apache_hive&open=AZPVJuwuQje9E2BL3_-t
2. https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4902&id=apache_hive&open=AZPFqXK8BQZ6BS4X3Vch
3. https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4902&id=apache_hive&open=AZPFqXNlBQZ6BS4X3Vc4
4. https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4902&id=apache_hive&open=AZPFqXNlBQZ6BS4X3Vc5
5. https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4902&id=apache_hive&open=AZPFqXNlBQZ6BS4X3Vcv
6. https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4902&id=apache_hive&open=AZPFqXNlBQZ6BS4X3Vcw
7. https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4902&id=apache_hive&open=AZPFqXHnBQZ6BS4X3Vcg

aturoczy (approved):

MikaelSmith (approved):

deniskuzZ:
this PR causes issues on a Jenkins, build timeouts after 12hr

deniskuzZ:
i checked the test bundle and there are no failures. rebased with master

kdeschle:
Any idea what the build timeout was about?

deniskuzZ:
it seems that all splits are done except split-13. It's already more than 5hr

kdeschle:
Thanks for the insights. I was able to repro locally as well and fix the issues.

deniskuzZ:
> Thanks for the insights. I was able to repro locally as well and fix the issues.

Thanks for the fix @kdeschle!

I noticed that we have 2 commits for HIVE-27872, should I squash them into 1
1. Support multi-stream parallel fetch in JDBC driver
2. Don't require HiveConf for JDBC Metadata calls 

Also [HIVE-27882](https://issues.apache.org/jira/browse/HIVE-27882): Fix missing classes in standalone jar, mentioned in the PR description, is missing. Is that expected?

kdeschle:
Closed HIVE-27882 as dup of HIVE-28315 and remove from PR desc since that was already fixed.
Updated the other ticket to HIVE-28018.

I'd like to leave the stack as-is now as they address different problems and can be cherrypicked separately to other branches.

deniskuzZ:
merged to master
thanks @MikaelSmith  & @kdeschle for the contribution!

kdeschle:
Thanks Denis!

PR #4903, opened 2023-11-28
HIVE-27912: Include Iceberg module in nightly builds

zhangbutao:

http://ci.hive.apache.org/job/hive-nightly/

[HIVE-25715](https://issues.apache.org/jira/browse/HIVE-25715) added nightly builds, and it give users a chance to test the snapshot binary package.

But the builds didn't contain Iceberg module, It would be good to include Iceberg module. And then user like [HIVE-27898](https://issues.apache.org/jira/browse/HIVE-27898) can use the snapshot binary package to test some queries.

No

No

No need test.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -308,7 +308,7 @@ time docker rm -f dev_$dbType || true
 set -e
 dev-support/nightly
 '''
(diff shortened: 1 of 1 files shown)


ayushtkn (approved):
LGTM

PR #4904, opened 2023-11-28
HIVE-27913: Wrap NoSuchObjectException into HiveException in Hive.dro…

wecharyu:
…pDatabase and Hive.dropConnector

Wrap `NoSuchObjectException` into `HiveException` for dropDatabase and dropDataConnector api in `Hive`.

To align the `drop_database` and `drop_dataconnector` behavior with `drop_table`.

Yes, the Hive client user need handle the `NoSuchObjectException` inside `HiveException` after this change.

No

Pass existing tests.

### DIFF
--- a/hcatalog/core/src/test/java/org/apache/hive/hcatalog/HcatTestUtils.java
+++ b/hcatalog/core/src/test/java/org/apache/hive/hcatalog/HcatTestUtils.java
@@ -60,7 +60,7 @@ public class HcatTestUtils {
    * Removes all databases and tables from the metastore
    */
   public static void cleanupHMS(Hive hive, Warehouse wh, FsPermission defaultPerm)
-    throws HiveException, MetaException, NoSuchObjectException {
+    throws HiveException, MetaException {
     for (String dbName : hive.getAllDatabases()) {
       if (dbName.equals("default")) {
         continue;
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseOperation.java
@@ -39,28 +39,24 @@ public class DropDatabaseOperation extends DDLOperation<DropDatabaseDesc> {
 
   @Override
   public int execute() throws HiveException {
-    try {
-      String dbName = desc.getDatabaseName();
-      ReplicationSpec replicationSpec = desc.getReplicationSpec();
-      if (replicationSpec.isInReplicationScope()) {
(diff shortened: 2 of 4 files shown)


ayushtkn (commented):
> Does this PR introduce any user-facing change?
Yes, the Hive client user need handle the NoSuchObjectException inside HiveException after this change.


And why we would do that & introduce incompatibilities for the end client?

aturoczy:
> > Does this PR introduce any user-facing change?
> > Yes, the Hive client user need handle the NoSuchObjectException inside HiveException after this change.
> 
> And why we would do that & introduce incompatibilities for the end client?

Maybe it is not even incompatible. I try to find out which actor is throwing the NoSuchObjectException and I haven't found it. What was the purpose of this exception? The Object does not exist anymore?

wecharyu:
`dropDatabase` may encounter `NoSuchObjectException` if the database does not exist:
https://github.com/apache/hive/blob/6b2e21a93ef3c1776b689a7953fc59dbf52e4be4/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java#L1675-L1684

`dropDataConnector` is compatible because it already wrap the exception in master:
https://github.com/apache/hive/blob/6b2e21a93ef3c1776b689a7953fc59dbf52e4be4/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L992-L999

The reason to add this change is that most exceptions in `Hive.java` are wrapped into `HiveException`, such as `dropTable`. And when using Hive call we do not need catch extra exceptions except for HiveException. Also this change will not affect the user result because NoSuchObjectException will be caught in HiveException.

CC: @ayushtkn @aturoczy

ayushtkn:
if i have a call like
try {
do something
} catch(NoSuchObjectException ne) {
do some logic
}

now if I throw HiveException, such a code will break, right?

aturoczy:
> NoSuchObjectException

This is a dilemma to me. Actually, the NoSuchObjectException does not describe the nature of the exception. It is pretty undescribed, what kind of object. Table, Column, Database etc. An object is missing, but an object could be anything. Which imho is not clear for the API user. 

On the other hand, the HiveException is too general in the Hive. It wraps everything and does not describe the actual issue till you do not open the details of the exception. Which is also not the best design principals. 

Of course, it would need a wider agreement. But to me, it will be the following 

HiveExcption is the root and I would have DatabaseMissingException, TableMissingException, and more detailed exception for those scenarios where the API consumers would read it clearly.

cc: @deniskuzZ @ayushtkn @abstractdog

wecharyu:
Close it since there are some incompatible cases.

PR #4908, opened 2023-11-30
HIVE-27918: Iceberg: Push transforms for clustering during table writes

SourabhBadhya:

Iceberg: Push transforms for clustering during table writes

Currently transformed columns (except for bucket transform) are not pushed / passed as clustering columns. This can lead to incorrect clustering on such columns which can lead non-performant writes. Hence push transforms for clustering during table writes.

No

No

QTests

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -38,7 +38,6 @@ import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.commons.collections.MapUtils;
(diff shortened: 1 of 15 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -234,6 +240,77 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
             }
           };
 
+  private static final transient BiFunction<Integer, Integer, Function<List<ExprNodeDesc>, ExprNodeDesc>>
```
deniskuzZ, line 243: this class looks overloaded already, could we move these functions into some helper class?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -732,6 +809,16 @@ private void addCustomSortExpr(Table table,  org.apache.hadoop.hive.ql.metadata.
       int order = fieldOrderMap.get(spec.getColumnName());
       if (TransformSpec.TransformType.BUCKET.equals(spec.getTransformType())) {
         customSortExprs.add(BUCKET_SORT_EXPR.apply(order + offset, spec.getTransformParam().get()));
+      } else if (TransformSpec.TransformType.TRUNCATE.equals(spec.getTransformType())) {
```
deniskuzZ, line 812: that's enum, could we use switch instead?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
SourabhBadhya: Done.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
SourabhBadhya: I have replaced the if-else block with switch block.

PR #4909, opened 2023-11-30
HIVE-27712: GenericUDAFNumericStatsEvaluator throws NPE

liangyu-1:

I modified function init() in org/apache/hadoop/hive/ql/udf/generic/GenericUDAFComputeStats.java line 398- line 399.

As I have described in [HIVE-27712](https://issues.apache.org/jira/browse/HIVE-27712)

When we set config: 
`set hive.groupby.skewindata=true;
set hive.map.aggr=true; `
execute a sql with groupby execution and join execution, I got a NullPointerException.

No

no

I rebuild the package and runs the same sql, it didn't throw NPE


zhangbutao:
@liangyu-1 Thanks for this PR. Could you please add a test to help others to reproduce this issue?

ayushtkn (commented):
Prod code looks good, you need to add a test, there are bunch of q files with names like compute_stats_*.q add one for your use case

ql/src/test/queries/clientpositive/compute_stats_double.q
```diff
@@ -7,3 +7,13 @@ select count(*) from tab_double;
 
 -- compute statistical summary of data
 select compute_stats(a, 'fm', 16) from tab_double;
+
+-- create a partitioned table with a double column
+
+create table tab_double_part(a double) partitioned by (b string);
```
zhangbutao, line 13: And You need regenerate the output  `compute_stats_double.q.out`

liangyu-1:
> Copy my respose from JIRA comment: This problem is somewhat similar to that one https://issues.apache.org/jira/browse/HIVE-22633, which maybe be fixed by https://issues.apache.org/jira/browse/HIVE-23530
> 
> Could you please try this patch [HIVE-23530](https://issues.apache.org/jira/browse/HIVE-23530)?

It seems that this patch can solve the problem, but I think my solution is more helpful because this is a bug for function compute_stats, that patch just deprecated the function compute_stats and modified too many files which may cause more problems.

I think we should just fix the bug rather than deprecate it.

ayushtkn:
It should be in either of two states: either we remove it or fix it, if we are keeping it, it shouldn't be broken, if we plan to remove not now but in future then it should be in fixed state, I believe, why would we have broken code in our release

zhangbutao:
> It should be in either of two states: either we remove it or fix it, if we are keeping it, it shouldn't be broken, if we plan to remove not now but in future then it should be in fixed state, I believe, why would we have broken code in our release

@ayushtkn This function **compute_stats(GenericUDAFComputeStats)**  has been deprecated by HIVE-23530 in **Hive 4.0.0-alpha-1** and it is not used anywhere in master branch , mybe we can safely remove it now?

ayushtkn:
if it is deprecated in the prev release, it can be removed in the master branch now

zhangbutao:
@liangyu-1 Thanks for the fix. This function is no longer used and was deprecated since Hive 4.0.0-alpha-1, so  this fix doesn't make sense in the current master branch. 

https://github.com/apache/hive/pull/4928 I just raised a PR to remove this function in the master branch.

ayushtkn:
@zhangbutao I see the PR to remove the test has failures, do you think we can fix this in 4.0 & chase the removal in 4.1.0(master)?

zhangbutao:
> @zhangbutao I see the PR to remove the test has failures, do you think we can fix this in 4.0 & chase the removal in 4.1.0(master)?

@ayushtkn The `compute_stats` is used in multiple qtests as a udf to verify some bugs though it was not used in current hive source code, so it seems that removal is not easy & I am afraid of removal will make code instable now. Maybe need more investigation about the removal.

As for this fix, it maybe a trivial improvement from a user UDF usage, as the current master code will never have use this code except some qtests. So feel free to merge this fix if you think ok.
Thanks.

zhangbutao:
> Ok, lets chase the removal only For the removal can we remove it from being registered in FunctionRegistry & in the TestCliDriver just register it as a FunctionRegistry.registerTemporaryUDF(..), & use it only for tests maybe...

Looks like the temporaryUDF is a good workaround, at least this can prevent users from using the deprecated function again.
Will try this way soon.
Thanks.

zhangbutao:
Closed this PR as we have removed the function usage on master branch.

PR #4910, opened 2023-11-30
HIVE-27323: Iceberg: malformed manifest file or list can cause data breach.

ayushtkn:

Add a config which ensures all the data files are within the table location.

Security use-cases

Yes, if the config is turned on, the iceberg tables with data files outside the table directory won't be readable.

No

UT

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -2229,6 +2229,8 @@ public class HiveConf extends Configuration {
 
     HIVE_ICEBERG_MASK_DEFAULT_LOCATION("hive.iceberg.mask.default.location", false,
         "If this is set to true the URI for auth will have the default location masked with DEFAULT_TABLE_LOCATION"),
+    HIVE_ICEBERG_ALLOW_DATAFILES_IN_TABLE_LOCATION_ONLY("hive.iceberg.allow.datafiles.in.table.location.only", false,
+        "If this is set to true, then all the data files being read should be withing the table location"),
 
     HIVEUSEEXPLICITRCFILEHEADER("hive.exec.rcfile.use.explicit.header", true,
         "If this is set the header for RCFiles will simply be RCF.  If this is not\n" +
@@ -5575,7 +5577,8 @@ public class HiveConf extends Configuration {
             "hive.zookeeper.ssl.keystore.type," +
             "hive.zookeeper.ssl.truststore.location," +
             "hive.zookeeper.ssl.truststore.password," +
-            "hive.zookeeper.ssl.truststore.type",
+            "hive.zookeeper.ssl.truststore.type," +
+            "hive.iceberg.allow.datafiles.in.table.location.only",
         "Comma separated list of configuration options which are immutable at runtime"),
     HIVE_CONF_HIDDEN_LIST("hive.conf.hidden.list",
         METASTOREPWD.varname + "," + HIVE_SERVER2_SSL_KEYSTORE_PASSWORD.varname
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.hive.common.FileUtils;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.llap.LlapHiveUtils;
 import org.apache.hadoop.hive.ql.exec.Utilities;
+import org.apache.hadoop.hive.ql.metadata.AuthorizationException;
 import org.apache.hadoop.hive.ql.metadata.HiveUtils;
 import org.apache.hadoop.hive.ql.plan.MapWork;
 import org.apache.hadoop.mapred.JobConf;
@@ -234,7 +235,7 @@ public class IcebergInputFormat<T> extends InputFormat<Void, T> {
           checkResiduals(task);
         }
         if (dataFilesWithinTableLocationOnly) {
-          validateFilesWithinTableDirectory(task, tableLocation);
+          validateFileLocations(task, tableLocation);
         }
         splits.add(new IcebergSplit(conf, task));
       });
@@ -252,10 +253,10 @@ public class IcebergInputFormat<T> extends InputFormat<Void, T> {
     return splits;
   }
 
-  private static void validateFilesWithinTableDirectory(CombinedScanTask split, Path tableLocation) {
+  private static void validateFileLocations(CombinedScanTask split, Path tableLocation) {
     for (FileScanTask fileScanTask : split.files()) {
       if (!FileUtils.isPathWithinSubtree(new Path(fileScanTask.file().path().toString()), tableLocation)) {
-        throw new IllegalArgumentException("The table contains paths which are outside the table location");
+        throw new AuthorizationException("The table contains paths which are outside the table location");
       }
     }
   }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
@@ -221,7 +221,7 @@ public class IcebergInputFormat<T> extends InputFormat<Void, T> {
       scan = applyConfig(conf, createTableScan(table, conf));
     }
 
-    boolean dataFilesWithinTableLocationOnly =
+    boolean allowDataFilesWithinTableLocationOnly =
         conf.getBoolean(HiveConf.ConfVars.HIVE_ICEBERG_ALLOW_DATAFILES_IN_TABLE_LOCATION_ONLY.varname,
             HiveConf.ConfVars.HIVE_ICEBERG_ALLOW_DATAFILES_IN_TABLE_LOCATION_ONLY.defaultBoolVal);
     Path tableLocation = new Path(conf.get(InputFormatConfig.TABLE_LOCATION));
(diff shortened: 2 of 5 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergInputFormat.java
```diff
@@ -176,6 +179,20 @@ public RecordReader<Void, Container<Record>> getRecordReader(InputSplit split, J
     }
   }
 
+  private static void validateFilesWithinTableDirectory(InputSplit split, JobConf job) throws IOException {
+    boolean dataFilesWithingTableLocationOnly =
+        job.getBoolean(HiveConf.ConfVars.HIVE_ICEBERG_ALLOW_DATA_IN_TABLE_LOCATION_ONLY.varname,
+            HiveConf.ConfVars.HIVE_ICEBERG_ALLOW_DATA_IN_TABLE_LOCATION_ONLY.defaultBoolVal);
+    if (dataFilesWithingTableLocationOnly) {
+      Path tableLocation = new Path(job.get(InputFormatConfig.TABLE_LOCATION));
```
deniskuzZ, line 187: why do we restrict on reads - that is expensive? we shouldn't allow any additions outside of table dir

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2226,6 +2226,10 @@ public static enum ConfVars {
     HIVE_ICEBERG_EXPIRE_SNAPSHOT_NUMTHREADS("hive.iceberg.expire.snapshot.numthreads", 4,
         "The number of threads to be used for deleting files during expire snapshot. If set to 0 or below it uses the" +
             " defult DirectExecutorService"),
+
+    HIVE_ICEBERG_ALLOW_DATA_IN_TABLE_LOCATION_ONLY("hive.iceberg.allow.data.in.table.location.only", false,
```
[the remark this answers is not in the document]
jkovacs-hwx: No, unfortunately the issue that is addressed here is Iceberg specific.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergInputFormat.java
jkovacs-hwx: the read path constraint is to avoid to read other location's - aka other table's - data from such a malicious table. Such table can be constructed manually, not necessarily written by spark (actually most probably constructed with other methods). 
The problem here is that without this read limitation, the user can use hive's elevated privileges (doAs=false) to access secured data even if data doesn't belong to the user's own table.

deniskuzZ: Users can also place a custom jar into the classpath to bypass that check. It doesn't seem like a robust solution, but more like a hack.

deniskuzZ: > Would a custom/per-user jar could lead to the same class override?
yes, but that could be altered only by admin

> Why don't we consider storage-based authorization
why can't we use `doAs=true` as a quick fix?

> not expose the metadata
if the data location is sensitive - why should we leak it? Also, this way authorization would be done in 1 dedicated place (service designed for that), but not in every engine.

jkovacs-hwx: If jar injection to override behaviour is only possible by admin, then this should not be a blocker for this scope. E.g. admin could even inject jar to override the AuthN chain and dump username/password pairs from those who are connecting to HS2 e.g. via jdbc via this auth method.

doAs=true is not an option for Fine-Grained Access Control where this issues is the most significant (e.g. otherwise masked data breached as non-masked). Spark has no FGAC and elevated privilege based data access decoupling from end-user based file-access. 

> if the data location is sensitive - why should we leak it?

Historically it was not sensitive and even with Iceberg it should not be treated as sensitive. The issues comes from  Iceberg's new behaviour where instead of limiting the read to a directory in a Hive table format case, Iceberg now can read data files from anywhere the hive service user has access to, if the location is in its manifest file.

jkovacs-hwx: I agree, and it is also breaking the multi-data locations in case of tiered-storage usage or just the movement of the Iceberg table's _write.data.path_, but that's why this is behind a configuration flag not enabled by default and considered only as a temporary quick fix for those who are rather break it temporarily (especially if they use Iceberg as standard external table only not noticing the limitations), than experience data breach. We are also already discussing a more complex possible solution where neither the Iceberg API+functionality are limited, nor the malicious tables could expose other tables' data.

deniskuzZ: ok, also one of the findings (Implement LOAD data for partitioned tables via Append API) won't be possible under this config

jkovacs-hwx: Right. We can articulate the limitation better in the HiveConf.java change, like adding that this breaks all Iceberg tables with data located outside of the table's location.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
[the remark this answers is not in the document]
jkovacs-hwx: @ayushtkn Could we extend the description with the note that this breaks Iceberg tables with data files located outside of the table location?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergInputFormat.java
```diff
@@ -176,6 +179,20 @@ public RecordReader<Void, Container<Record>> getRecordReader(InputSplit split, J
...
+      for (String location : split.getLocations()) {
```
deniskuzZ, line 188: is that reliable?
````
 getLocations() won't be accurate when called on worker nodes and will always return "*"
````

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergInputFormat.java:188]: check the method code:
````
  public String[] getLocations() {
    // The implementation of getLocations() is only meant to be used during split computation
    // getLocations() won't be accurate when called on worker nodes and will always return "*"
    if (locations == null && conf != null) {
      boolean localityPreferred = conf.getBoolean(InputFormatConfig.LOCALITY, false);
      locations = localityPreferred ? Util.blockLocations(task, conf) : ANYWHERE;
    } else {
      locations = ANYWHERE;
    }

    return locations;
  }
````
when invoked 2nd time it returns "*" always :) nice design

File: common/src/java/org/apache/hadoop/hive/conf/HiveConf.java (hunk 2 above)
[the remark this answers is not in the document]
Comment (ayushtkn) [reply common/src/java/org/apache/hadoop/hive/conf/HiveConf.java]: Changed & added to restricted list. I think what it breaks is meant for the documentation, we don't usually mention the after effects of the configs in the description.
It is indicative the files should be within table location

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergInputFormat.java [hunk 4]
```diff
@@ -176,6 +179,20 @@ public RecordReader<Void, Container<Record>> getRecordReader(InputSplit split, J
     }
   }
 
+  private static void validateFilesWithinTableDirectory(InputSplit split, JobConf job) throws IOException {
+    boolean dataFilesWithingTableLocationOnly =
```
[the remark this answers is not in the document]
ayushtkn: sorted

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergInputFormat.java
```diff
@@ -152,8 +153,10 @@ public InputSplit[] getSplits(JobConf job, int numSplits) throws IOException {
   }
 
   @Override
-  public RecordReader<Void, Container<Record>> getRecordReader(InputSplit split, JobConf job,
-                                                               Reporter reporter) throws IOException {
+  public RecordReader<Void, Container<Record>> getRecordReader(InputSplit split, JobConf job, Reporter reporter)
+      throws IOException {
+    validateFilesWithinTableDirectory(split, job);
```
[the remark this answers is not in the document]
ayushtkn: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergInputFormat.java
ayushtkn: That wasn't a cool thing I believe, I changed to use the FileScanTasks, that is what contains the actual files to be read.

I even added a test case where a file is actually readable, so the path would be extracted here & then for normal read, to ensure some only one time read stuff doesn't break the normal reads

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
```diff
@@ -218,13 +220,22 @@ public List<InputSplit> getSplits(JobContext context) {
       scan = applyConfig(conf, createTableScan(table, conf));
     }
 
+    boolean dataFilesWithinTableLocationOnly =
```
deniskuzZ, line 223: please rename -> checkFilesWithinTableLocationOnly

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
```diff
@@ -241,6 +252,14 @@ public List<InputSplit> getSplits(JobContext context) {
     return splits;
   }
 
+  private static void validateFilesWithinTableDirectory(CombinedScanTask split, Path tableLocation) {
+    for (FileScanTask fileScanTask : split.files()) {
+      if (!FileUtils.isPathWithinSubtree(new Path(fileScanTask.file().path().toString()), tableLocation)) {
+        throw new IllegalArgumentException("The table contains paths which are outside the table location");
```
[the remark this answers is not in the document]
ayushtkn: changed to authorizationexception

PR #4912, opened 2023-12-01
HIVE-27893 Add a range validator in hive.metastore.batch.retrieve.max to only have values greater than 0

vikramahuja1001:
 Adding a range validator as the part of this Jira to keep the lower bound value of hive.metastore.batch.retrieve.max to be 1. Also blocking PartitionIterable call if any user/client ever calls the constructor directly with batch size <1. Currently there is no Validation for the property hive.metastore.batch.retrieve.max. In case if it is ever set to <= 0, PartitionIterable will throw up NoSuchElementException because the batch size will be 0 there. 

GetPartition can fail due to thrift having a size limitation of 2 GB. To bypass this PartitionIterable is used, PartitionIterable converts one getpartition call for a table with large number of partitions into multiple HMS calls of smaller getPartitions calls. PartitionIterable has a batchsize as a parameter. In case if the batchsize is ever set to 0 by any user or any other client, partitionIterable will throw a NoSuchElementException. 

No

No

Added a test case for the same

### DIFF
--- a/ql/src/test/org/apache/hadoop/hive/ql/exec/TestGetPartitionInBatches.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/exec/TestGetPartitionInBatches.java
@@ -23,11 +23,13 @@ import org.apache.hadoop.hive.metastore.HiveMetaStoreClient;
 
 import org.apache.hadoop.hive.metastore.api.GetPartitionsByNamesRequest;
 import org.apache.hadoop.hive.metastore.api.MetaException;
+import org.apache.hadoop.hive.metastore.api.MetastoreException;
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.metastore.client.builder.PartitionBuilder;
 import org.apache.hadoop.hive.ql.metadata.Hive;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
+import org.apache.hadoop.hive.ql.metadata.PartitionIterable;
 import org.apache.hadoop.hive.ql.session.SessionState;
 import org.junit.Assert;
 import org.junit.After;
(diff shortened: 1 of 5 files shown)


ql/src/test/org/apache/hadoop/hive/ql/exec/TestGetPartitionInBatches.java
```diff
@@ -256,4 +258,21 @@ public void testBatchingWhenException() throws Exception {
...
+
+    @Test
+    public void testBatchingWhenBatchSizeIsZero() throws MetaException {
+        HiveMetaStoreClient spyMSC = spy(msc);
+        hive.setMSC(spyMSC);
+        int batchSize = 0;
+        try {
+            new PartitionIterable(hive, hive.getTable(dbName, tableName), null, batchSize);
+        } catch (HiveException e) {
+            Assert.assertTrue(e.getMessage().contains("Batch Size for Partition Iterable should be > 0"));
```
aturoczy, line 270: In this case would need to change the error message.

**It is not belong to your PR:**
btw I feel we overuse the HiveException. I know it is bit out of the scope of this but I have seen several PR the Author checked the message of the exception which is an anti-pattern. Every exception should be clear about the nature of the exception. The message just provides details for the developer. It is not an end-user interface. 

If you are agree (@ayushtkn, @deniskuzZ @abstractdog @zratkai, @veghlaci05 ) , we should have a discussion in the future of the exceptions use in the hive. HiveException is oddly overused.

aturoczy (changes requested):

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PartitionIterable.java
```diff
@@ -164,6 +164,9 @@ public PartitionIterable(Collection<Partition> ptnsProvided) {
    * a Hive object and a table object, and a partial partition spec.
    */
   public PartitionIterable(IMetaStoreClient msc, Table table, int batch_size) throws MetastoreException {
+    if (batch_size < 1) {
+      throw new MetastoreException("Batch Size for Partition Iterable should be > 0");
```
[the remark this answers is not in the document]
vikramahuja1001: Changed the exception message.

ql/src/test/org/apache/hadoop/hive/ql/exec/TestGetPartitionInBatches.java
vikramahuja1001: I have just changed the error message here for now.
Though i do agree about the clarity of the nature for every exception.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4153,43 +4153,14 @@ public List<Partition> getPartitions(Table tbl) throws HiveException {
...
-   * Get all the partitions. Do it in batches if batchSize is more than 0 else get it in one call.
+   * Get all the partitions in batches. Does not include auth information.
    * @param tbl table for which partitions are needed
    * @return list of partition objects
    */
   public Set<Partition> getAllPartitionsOf(Table tbl) throws HiveException {
-    int batchSize= MetastoreConf.getIntVar(
-            Hive.get().getConf(), MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX);
-    if (batchSize > 0) {
-      return getAllPartitionsInBatches(tbl, batchSize, DEFAULT_BATCH_DECAYING_FACTOR, MetastoreConf.getIntVar(
```
chinnaraolalam, line 4188: Here there is a chance of coming batchSize <=0, as prior to this validation is not happening.   RangeValidator will check only if setMetaConf is executed.

vikramahuja1001: I have reverted this change for now. Better to add it as a part of separate JIRA.

chinnaraolalam:
Changes looks good other than discussed Exception handling. These Exception refactoring please do in a separate Jira.

chinnaraolalam:
+1 LGTM

vikramahuja1001:
Thanks for the review everyone

PR #4915, opened 2023-12-04
HIVE-27930: Insert/Load overwrite table partition does not clean up directory before overwriting

KiranVelumuri:

This PR adds an additional check for cleaning up the destination directory while executing an INSERT/LOAD OVERWRITE TABLE statement for those partitions which do not exist in the metastore.

For the partitions not in metastore but contain some files in the filesystem, INSERT OVERWRITE TABLE with partition does not clean the directory before overwriting.

No

No

mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=insert_and_load_overwrite_drop_partition.q

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -5542,6 +5542,13 @@ private void constructOneLBLocationMap(FileStatus fSta,
         throw new HiveException("Getting globStatus " + srcf.toString(), e);
       }
 
+      // For insert/load overwrite cases, where external.table.purge is disabled for the table, there may be stale
+      // partitions present in the table location after Alter table drop partition operation. In such cases, oldPath will be
+      // null, since those partitions will not be present in metastore. Added below check to clean up those stale partitions.
+      if (oldPath == null && isInsertOverwrite) {
+        deleteOldPathForReplace(destf, destf, conf, purge, deletePathFilter, isNeedRecycle);
+      }
+
       // the extra check is required to make ALTER TABLE ... CONCATENATE work
       if (oldPath != null && (srcs != null || isInsertOverwrite)) {
         deleteOldPathForReplace(destf, oldPath, conf, purge, deletePathFilter, isNeedRecycle);
--- a/ql/src/test/queries/clientpositive/insert_and_load_overwrite_drop_partition.q
+++ b/ql/src/test/queries/clientpositive/insert_and_load_overwrite_drop_partition.q
@@ -1,0 +1,61 @@
+CREATE EXTERNAL TABLE `table1`(
+  `name` string,
+  `number` string)
+PARTITIONED BY (
+  `part_col` string);
+
+CREATE EXTERNAL TABLE `table2`(
+  `name` string,
+  `number` string)
+PARTITIONED BY (
+  `part_col` string);
+
+insert into table table1 values ('a', '10', 'part1');
+insert into table table1 values ('b', '11', 'part1');
+insert into table table1 values ('a2', '2', 'part2');
+
+insert into table table2 values ('x', '100', 'part1');
+insert into table table2 values ('y', '101', 'part1');
+insert into table table2 values ('z', '102', 'part1');
+insert into table table2 values ('x2', '200', 'part2');
+insert into table table2 values ('y2', '201', 'part2');
+insert into table table2 values ('x3', '300', 'part3');
+
+--non empty input case
+alter table table2 drop partition(part_col='part1');
+
+select count(*) from table2 where part_col='part1';
+
+dfs -ls ${hiveconf:hive.metastore.warehouse.dir}/table2/part_col=part1;
+
+insert overwrite table table2 partition(part_col='part1') select name, number from table1 where part_col='part1';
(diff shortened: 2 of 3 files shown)


ql/src/test/queries/clientpositive/insert_overwrite_drop_partition.q
```diff
@@ -0,0 +1,46 @@
+set hive.exec.dynamic.partition.mode=nonstrict;
+set hive.stats.column.autogather=false;
```
Indhumathi27, line 2: are these configs (from line 2 to 4) needed to be set for this scenario ?

ql/src/test/queries/clientpositive/insert_overwrite_drop_partition.q
```diff
@@ -0,0 +1,46 @@
...
+insert into table table2 values ('y2', '201', 'part2');
+
+--should be 2
+select count(*) from table1 where part_col='part1';
+
+--should be 3
+select count(*) from table2 where part_col='part1';
+
+alter table table2 drop partition(part_col='part1');
+
```
Indhumathi27, line 35: Can you add a check for stale partition exists here or not

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -5542,6 +5542,12 @@ private void replaceFiles(Path tablePath, Path srcf, Path destf, Path oldPath, H
         throw new HiveException("Getting globStatus " + srcf.toString(), e);
       }
 
+      // in insert overwrite table query for partitions not in metastore but exist in filesystem,
+      // oldPath is null and destination directory needs to cleaned up
```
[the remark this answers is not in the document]
zhangbutao: Good catch! You need to clarify that this issue can appear in case of `TBLPROPERTIES ("external.table.purge"="false")`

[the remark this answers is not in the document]
KiranVelumuri: Yes, I have updated the scenarios in the comment.

ql/src/test/queries/clientpositive/insert_overwrite_drop_partition.q
KiranVelumuri: No, removed the configs as default value of hive.exec.dynamic.partition.mode is nonstrict

ql/src/test/queries/clientpositive/insert_overwrite_drop_partition.q
```diff
@@ -0,0 +1,46 @@
...
+insert overwrite table table2 partition(part_col='part1') select name, number from table1 where part_col='part1';
```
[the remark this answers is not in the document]
KiranVelumuri: Yes, I have checked this. Thank you for mentioning LOAD OVERWRITE PARTITION for which this issue exists as well, which has been solved with these changes. Attached q file for the same.

data/files/table.txt
```diff
@@ -0,0 +1,2 @@
+a
```
zhangbutao, line 1: I think we can reuse the existing text file. such as `kv1.txt`

ql/src/test/queries/clientpositive/load_drop_partition.q
```diff
@@ -0,0 +1,24 @@
+dfs -cp ${system:hive.root}/data/files/table.txt ${system:test.tmp.dir}/;
+
```
zhangbutao, line 2: I think one qfile is enough. Let's test all the cases in `insert_overwrite_drop_partition.q`

zhangbutao (commented):
Thanks for the fix. Left some minor comments.

ql/src/test/queries/clientpositive/load_drop_partition.q
KiranVelumuri: Added load overwrite partition case in one q file

ql/src/test/queries/clientpositive/insert_and_load_overwrite_drop_partition.q
```diff
@@ -0,0 +1,72 @@
...
+dfs -ls ${hiveconf:hive.metastore.warehouse.dir}/table2/part_col=part2;
+
+insert overwrite table table2 partition(part_col='part2') select name, number from table1 where part_col='dummy_part';
+
+select count(*) from table2 where part_col='part2';
+
+dfs -ls ${hiveconf:hive.metastore.warehouse.dir}/table2/part_col=part2;
+
+--load overwrite partition
+CREATE EXTERNAL TABLE `table3`(
```
zhangbutao, line 50: Nit: No need create a new table here. We can reuse `table2` with a new partition(part3) to keep the test clean.
Thanks.

KiranVelumuri: Updated the test to reuse table2. Please help in reviewing and merging the changes. Thank you.

zhangbutao (approved):
LGTM +1

Indhumathi27:
+1 LGTM

KiranVelumuri:
@zhangbutao Thank you for the review and approval. Please merge the changes.

PR #4917, opened 2023-12-04
HIVE-27934: Fix incorrect description about the execution framework in README.md

zhangbutao:

https://github.com/apache/hive/commit/4c7dbd560f9034fbce77d17a82bd0a4abc3091fd#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5
HIVE-26134 has removed spark execution, there are only 2 execution frameworks left, MR&Tez.

No

No

No need test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,7 @@ subqueries, common table expressions, and more.  Hive's SQL can also be
 extended with user code via user defined functions (UDFs), user defined
 aggregates (UDAFs), and user defined table functions (UDTFs).
 
-Hive users have a choice of 2 runtimes when executing SQL queries.
-Users can choose between Apache Hadoop MapReduce or Apache Tez
+Hive users can choose between Apache Hadoop MapReduce or Apache Tez
 frameworks as their execution backend. MapReduce is a
 mature framework that is proven at large scales. However, MapReduce
 is a purely batch framework, and queries using it may experience
--- a/README.md
+++ b/README.md
@@ -25,8 +25,9 @@ extended with user code via user defined functions (UDFs), user defined
 aggregates (UDAFs), and user defined table functions (UDTFs).
 
 Hive users can choose between Apache Hadoop MapReduce or Apache Tez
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,7 +24,7 @@ subqueries, common table expressions, and more.  Hive's SQL can also be
 extended with user code via user defined functions (UDFs), user defined
 aggregates (UDAFs), and user defined table functions (UDTFs).
 
-Hive users have a choice of 3 runtimes when executing SQL queries.
+Hive users have a choice of 2 runtimes when executing SQL queries.
```
zabetak, line 27: Removing the sentence completely would be even better so that we don't have to update the number every time something changes.

MapReduce is deprecate so I am not sure it it is really an option maybe worth mentioning this as well.

With a little effort the commit summary/JIRA title could become more specific and someone could understand what changed (or is going to change) without looking at the diff.

zabetak (approved):

README.md
zhangbutao: > Removing the sentence completely would be even better so that we don't have to update the number every time something changes.

Agreed! Removed this sentence. Thanks for suggestion!

> MapReduce is deprecate so I am not sure it it is really an option maybe worth mentioning this as well.

IMHO, at present, it seems that we never had a concrete plan for deprecating MapReduce, so I think it's better not to mention MR deprecation here. I am afraid of this will lead to some confusion for users.

> With a little effort the commit summary/JIRA title could become more specific and someone could understand what changed (or is going to change) without looking at the diff.

Will change te tilte. Thanks.

zabetak: MR was officially deprecated in Hive 2.0 (see HIVE-12300).

zhangbutao: Thanks for the info! I added a note  `Note that MapReduce framework has been deprecated since Hive 2, and Apache Tez is recommended.`

aturoczy (approved):

PR #4919, opened 2023-12-06
HIVE-27925: HiveConf: unify ConfVars enum and use underscore for better readability

kokila-19:
HiveConf: unify ConfVars enum and use underscore for better readability

Better readability

No

No

Build and Unit tests.

### DIFF
--- a/beeline/src/java/org/apache/hive/beeline/BeeLine.java
+++ b/beeline/src/java/org/apache/hive/beeline/BeeLine.java
@@ -1831,7 +1831,7 @@ public class BeeLine implements Closeable {
     String prompt;
     // read prompt configuration and substitute variables.
     HiveConf conf = getCommands().getHiveConf(true);
-    prompt = conf.getVar(HiveConf.ConfVars.CLIPROMPT);
+    prompt = conf.getVar(HiveConf.ConfVars.CLI_PROMPT);
     prompt = getCommands().substituteVariables(conf, prompt);
     return prompt + getFormattedDb() + "> ";
   }
--- a/beeline/src/java/org/apache/hive/beeline/BeeLineOpts.java
+++ b/beeline/src/java/org/apache/hive/beeline/BeeLineOpts.java
@@ -291,7 +291,7 @@ public class BeeLineOpts implements Completer {
       if (conf == null) {
         conf = beeLine.getCommands().getHiveConf(false);
       }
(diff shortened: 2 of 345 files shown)


aturoczy (approved):
+1 it is just a rename refactor which is nice

+1 for @abstractdog for this detailed describtion:
When I read something like "[BASICSTATSTASKSMAXTHREADSFACTOR](https://github.com/apache/hive/blob/70f34e27349dccf5fabbfc6c63e63c7be0785360/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java#L753)" I feel someone in the world laughs out loud thinking of me struggling. I can read it, but I hate it  imagine what if we have vars like [HIVE_MATERIALIZED_VIEW_ENABLE_AUTO_REWRITING_SUBQUERY_SQL](https://github.com/apache/hive/blob/70f34e27349dccf5fabbfc6c63e63c7be0785360/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java#L1921) without underscores...okay, let me help, it is: HIVEMATERIALIZEDVIEWENABLEAUTOREWRITINGSUBQUERYSQL

Made my day :D 

btw: This is a real and meaningful Hungarian word: megszentségteleníthetetlenségeskedéseitekért (same as our enums :) useless)

aturoczy:
@ayushtkn @abstractdog @deniskuzZ Could you please check this PR as well. Many files are touched but it is a simple rename.

abstractdog (approved):
+1

dom93dd:
This change seems to break the compatibility with apache.iceberg.hive dependency:

E.g. here:
https://github.com/apache/iceberg/blob/6f0d9dd47c693dc2aa13a38fab3d56473d925563/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java#L95

We have been using the HiveCatalog of apache.iceberg prior the newly introduced changes of 4.0.0 Alpha 1. With this change, this isn't possible anymore. Furthermore, if I switch to apache.hive HiveCatalog then the necessary included shading of fasterxml breaks everything. Because the JsonUtil class in apache.iceberg includes the fasterxml dependency without the relocated path. Anyone an idea what to do here?

Here is my feature request for apache.iceberg to introduce the ENV changes there too: https://github.com/apache/iceberg/issues/10254

Here is my jira issue where I specified the current problem I have with the shading:
https://issues.apache.org/jira/browse/HIVE-28229

Thanks!

PR #4920, opened 2023-12-06
HIVE-27937: Clarifying comments and xml configs around tez container size

abstractdog:
- added a useful comment in HiveConf for hive.tez.container.size
- cleaned up hive-site and tez-site xml files accordingly:
  - removed hive.tez.container.size from tez-site.xml (even if it works from there, it's a hive option, should be in hive-site for clarity's sake)
  - updated hive.tez.container.size in hive-site where a bigger value was removed from tez-site
- minimal refactor in DagUtils in the same area, added debug level messages

Clarity in HiveConf and qtests.

No.

No.

Precommit, and manually checked tez logs.

in DAGAppMaster log:
```
2023-12-05T09:39:06,211 INFO [DelayedContainerManager] rm.YarnTaskSchedulerService: Assigning container to task: containerId=container_1701797922416_0001_01_000002, task=attempt_1701797922416_0001_1_00_000000_0, containerHost=localhost:60444, containerPriority= 2, containerResources=<memory:128, vCores:1>, localityMatchType=NodeLocal, matchedLocation=localhost, honorLocalityFlags=true, reusedContainer=false, delayedContainers=1

```

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -2461,7 +2461,20 @@ public class HiveConf extends Configuration {
         "The default input format for tez. Tez groups splits in the AM."),
 
     HIVE_TEZ_CONTAINER_SIZE("hive.tez.container.size", -1,
-        "By default Tez will spawn containers of the size of a mapper. This can be used to overwrite."),
+        "The memory in MB that's used by a Tez task container (TezChild) in Tez container mode. Hive uses this \n"
+        + "property to create a Resource object which is accepted by Yarn (and used in TezAM to ask for TezChild \n"
+        + "containers). This should be distinguished from the Tez AM's (DAGAppMaster) memory, \n"
+        + "which is driven by tez.am.resource.memory.mb! \n"
+        + "Also, as Hive takes care of TezChild memory by setting this option, there is no need \n "
+        + "to set tez.task.resource.memory.mb differently. \n"
+        + "The final -Xmx arg for TezChild process is not equal to this setting, \n "
+        + "because Tez considers a heap fraction (80%), so by default: \n"
+        + "Xmx = hive.tez.container.size * tez.container.max.java.heap.fraction. \n"
+        + "In case of values <= 0, container size falls back to mapreduce.map.memory.mb. \n"
+        + "LLAP notes: while generating splits, the needed per-task resource is derived from this option \n"
+        + "(refer to HiveSplitGenerator, TezAvailableSlotsCalculator), so even if its value doesn't change the \n"
+        + "LLAP daemons' total physical size, it has to be configured properly. In this context \n"
+        + "4096 implies that you assume a single task will consume 4096MB from a daemon's shared heap."),
     HIVE_TEZ_CPU_VCORES("hive.tez.cpu.vcores", -1,
         "By default Tez will ask for however many cpus map-reduce is configured to use per container.\n" +
         "This can be used to overwrite."),
--- a/data/conf/iceberg/llap/hive-site.xml
+++ b/data/conf/iceberg/llap/hive-site.xml
@@ -38,10 +38,12 @@
         <description>A base for other temporary directories.</description>
     </property>
 
+    <!-- in LLAP mode hive.tez.container.size isn't used to actually determine container size, however while -->
+    <!-- calculating available slots (in split generation) in it's used through -->
+    <!-- the Vertex resource, so this has to be defined in order to get consistent test results -->
     <property>
-        <name>hive.tez.container.size</name>
-        <value>128</value>
-        <description></description>
+      <name>hive.tez.container.size</name>
+      <value>128</value>
     </property>
 
     <property>
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
@@ -23,6 +23,7 @@ import java.util.concurrent.ConcurrentHashMap;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
+import com.google.common.base.Strings;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
@@ -686,7 +687,9 @@ public class DagUtils {
         cpuCores = MRJobConfig.DEFAULT_MAP_CPU_VCORES;
       }
     }
-    return Resource.newInstance(memorySizeMb, cpuCores);
+    Resource resource = Resource.newInstance(memorySizeMb, cpuCores);
+    LOG.debug("Tez container resource: {}", resource);
+    return resource;
   }
 
   /*
@@ -716,19 +719,18 @@ public class DagUtils {
     }
     logLevel = sb.toString();
 
+    String finalOpts = null;
     if (HiveConf.getIntVar(conf, HiveConf.ConfVars.HIVE_TEZ_CONTAINER_SIZE) > 0) {
-      if (javaOpts != null) {
-        return javaOpts + " " + logLevel;
-      } else  {
-        return logLevel;
-      }
+      finalOpts = Strings.nullToEmpty(javaOpts) + " " + logLevel;
     } else {
       if (javaOpts != null && !javaOpts.isEmpty()) {
         LOG.warn(HiveConf.ConfVars.HIVE_TEZ_JAVA_OPTS + " will be ignored because "
                  + HiveConf.ConfVars.HIVE_TEZ_CONTAINER_SIZE + " is not set!");
       }
-      return logLevel + " " + MRHelpers.getJavaOptsForMRMapper(conf);
+      finalOpts = logLevel + " " + MRHelpers.getJavaOptsForMRMapper(conf);
     }
+    LOG.debug("Tez container final opts: {}", finalOpts);
+    return finalOpts;
   }
 
   private Vertex createVertexFromMergeWork(JobConf conf, MergeJoinWork mergeJoinWork,
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezAvailableSlotsCalculator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezAvailableSlotsCalculator.java
@@ -20,11 +20,15 @@ package org.apache.hadoop.hive.ql.exec.tez;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.tez.runtime.api.InputInitializerContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Default implementation of AvailableSlotsCalculator which relies on available capacity of the cluster
  */
 public class TezAvailableSlotsCalculator implements AvailableSlotsCalculator {
+  private static final Logger LOG = LoggerFactory.getLogger(TezAvailableSlotsCalculator.class);
+
     private InputInitializerContext inputInitializerContext;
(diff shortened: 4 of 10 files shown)


abstractdog:
can I get a review on this?
@deniskuzZ @ayushtkn @zabetak

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
```diff
@@ -716,19 +718,18 @@ private static String getContainerJavaOpts(Configuration conf) {
...
     } else {
       if (javaOpts != null && !javaOpts.isEmpty()) {
         LOG.warn(HiveConf.ConfVars.HIVETEZJAVAOPTS + " will be ignored because "
-                 + HiveConf.ConfVars.HIVETEZCONTAINERSIZE + " is not set!");
+            + HiveConf.ConfVars.HIVETEZCONTAINERSIZE + " is not set!");
       }
-      return logLevel + " " + MRHelpers.getJavaOptsForMRMapper(conf);
+      finalOpts = logLevel + " " + MRHelpers.getJavaOptsForMRMapper(conf);
     }
+    LOG.debug("Tez container log level: '{}', java opts: '{}', final opts: {}", logLevel, javaOpts, finalOpts);
```
zabetak, line 731: Do we need to log the values separately? Isn't finalOpts sufficient?

nit: in the log message finalOpts is not enclosed in single quotes as it happens for the rest.

zabetak (approved):
LGTM! Thanks for improving this area Laszlo!

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
abstractdog: final opts is a result of some minor magic here above, so if it's not obvious why it is what it is, the values of the passed-in logLevel and javaOpts variables might help

abstractdog:
thanks for the review @zabetak, I'm assuming you're fine with patch with my comment, let me know if you don't, also, I have to wait for a green test anyway :)

abstractdog:
seems like both TestTriggersNoTezSessionPool failures and qtest failures are related, so for currently unknown reasons, hive.tez.container.size looks to have an effect on llap local driver results somehow, which I need to investigate (I would ignore it and change back to original, but that would be against the purpose of this PR, which is to understand and clarify hive.tez.container.size as much as possible, so let me work on this further

abstractdog:
The last test failure was unrelated but I'm retriggering...
@zabetak : are you fine with the patch in its current form?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2461,7 +2461,17 @@ public static enum ConfVars {
...
+        + "property to create a Resource object which is accepted by Yarn (and used in TezAM to ask for TezChild \n"
+        + "containers. This should be distinguished from the Tez AM's (DAGAppMaster) memory, \n"
+        + "which is driven by tez.am.resource.memory.mb! \n"
+        + "Also, as Hive takes care of TezChild memory by setting this option, there is no need \n "
+        + "to set tez.task.resource.memory.mb differently. \n"
+        + "One more thing to note that the final -Xmx arg for TezChild process is not equal to this setting, \n "
+        + "because Tez considers a heap fraction (80%), so by default: \n"
+        + "Xmx = hive.tez.container.size * tez.container.max.java.heap.fraction. \n"
+        + "In case of values <= 0, container size falls back to mapreduce.map.memory.mb. \n"
+        + "LLAP note: this option has no effect in LLAP mode, because LLAP deamons are sized in another way."),
```
zabetak, line 2474: Based on the latest findings I am not sure if the `LLAP note` is still accurate. Should we remove it or mention the effect on split generation?

deniskuzZ: 1. no closing bracket after ` (and use`
2. could we drop everything after `One more thing to note that` and maybe add a link to https://blog.cloudera.com/optimizing-hive-on-tez-performance/

data/conf/iceberg/llap/hive-site.xml
```diff
@@ -38,10 +38,12 @@
         <description>A base for other temporary directories.</description>
     </property>
 
+    <!-- in LLAP mode hive.tez.container.size isn't used to actually determine container size, however while -->
```
deniskuzZ, line 41: are we estimating not optimal number of slots in case of LLAP? We are relying on `hive.tez.container.size` that's not used in LLAP sizing?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
```diff
@@ -716,19 +718,18 @@ private static String getContainerJavaOpts(Configuration conf) {
...
     logLevel = sb.toString();
 
+    String finalOpts = null;
     if (HiveConf.getIntVar(conf, HiveConf.ConfVars.HIVETEZCONTAINERSIZE) > 0) {
-      if (javaOpts != null) {
-        return javaOpts + " " + logLevel;
-      } else  {
-        return logLevel;
-      }
+      finalOpts = javaOpts != null ? javaOpts + " " + logLevel : logLevel;
```
deniskuzZ, line 723:
````
finalOpts = Strings.nullToEmpty(javaOpts) + " " + logLevel
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java (hunk 1 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java:731]: i don't see how`logLevel` & `javaOpts` could be useful. 
finalOpts always contains `logLevel` and depending on HIVETEZCONTAINERSIZE - `javaOpts`

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezAvailableSlotsCalculator.java [hunk 5]
```diff
@@ -39,6 +43,9 @@ public int getAvailableSlots() {
         }
         int totalResource = inputInitializerContext.getTotalAvailableResource().getMemory();
         int taskResource = inputInitializerContext.getVertexTaskResource().getMemory();
-        return totalResource / taskResource;
+        int availableSlots = totalResource / taskResource;;
+        LOG.debug("totalResource: {} / taskResource: {} =  availableSlots: {}", totalResource, taskResource,
```
deniskuzZ, line 47: could we please add the metric "Mb" since we output the resource information?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: 1. I'll fix

2. maybe, but linking a cloudera blog post is not the Apache way, so if we want to do it, we need to have an Apache page I believe...regarding the content here: having 2-3 extra clarifying sentences should not hurt, because it clarifies the most common confusing stuff here immediately, I still think it's useful

data/conf/iceberg/llap/hive-site.xml
abstractdog: good question, so in case of a testing llap driver (e.g. TestMiniLlapLocalCliDriver), where there is no classic llap deamons at all, this was what I saw...however, you're pointing out a new confusion, yeah, so if hive.tez.container.size is not inline with the independently sized LLAP deamons, then this is a discrepancy in available slot calculation...let me check this for a LLAP product, most probably Cloudera's CDW, and bring clarity here, at least in comments

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
abstractdog: yes, even better, I'll fix

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
abstractdog: @deniskuzZ , right, I believe we know hive.tez.container.size from different log lines, so finalOpts contains everything we need (as we can figure out javaOpts/logLevel from that easily), so I'm fine with making this log message simpler, as you and @zabetak suggested

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezAvailableSlotsCalculator.java
abstractdog: makes sense! thanks, I'll do

data/conf/iceberg/llap/hive-site.xml
zhangbutao: Intresting discussion. I want to learn something about llap too. :)
IMHO, `hive.tez.container.size` has no effect LLAP task(exector) memory size. It seems that llap task size is decided by this code:
https://github.com/apache/hive/blob/36ce858163a19e29eafe4a8d3307191bc28fc175/llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/ContainerRunnerImpl.java#L164-L165

abstractdog: yeah, hive.tez.container.size has no effect on the actual executor memory size (that's handled on LLAP codepath, e,g, what you linked, part of llap deamon sizing), but it seems to have an effect on available slots calculation, which is a crucial part of the split generation, that's why I'll have to clarify this area in code comments

zhangbutao: Thanks for explanation. Maybe i am not familar about llap implementation. But it seems strange that llap split generation is related `hive.tez.container.size` instead of per task(executor) size.
Maybe i needI to take some time to learn this part. :)

abstractdog: I know how it feels :) I hope by the end of this PR we'll all have a better understanding of this...

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: agreed, "One more thing to note that" is blog style, kind of useless in this context

deniskuzZ: tnx!

data/conf/iceberg/llap/hive-site.xml
abstractdog: @zhangbutao , @deniskuzZ : please find the latest comment and let me know if it's clearer

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: done

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
abstractdog: done

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezAvailableSlotsCalculator.java
abstractdog: done

data/conf/iceberg/llap/hive-site.xml
zhangbutao: Thx. I think this change is clear enough to me.

abstractdog: thanks @zhangbutao ! 
@zabetak , @deniskuzZ  : please let me know if the current state of this PR is fine enough to be merged (it's important, I wanted to bring clarity, so please confirm if I succeeded)

deniskuzZ (approved):
LGTM +1

PR #4923, opened 2023-12-07
HIVE-25054: (2.3) Upgrade `jodd-core` dependency to get rid of CVE-2018-21234

pan3793:

Backport HIVE-25054 to branch-2.3, to address CVE-2018-21234. 

jodd CVE is listed in https://issues.apache.org/jira/browse/SPARK-44757 in first place, with score 9.8.

Although I prefer to get rid of such dependency by https://github.com/apache/hive/pull/4888, but it is objected by the Hive community (please let me know if we can do it in branch-2.3)

No. (but the jodd APIs used by Hive are changed, it's a potential breaking change of the downstream project which consumes Hive and jodd API directly)

Yes, the upgraded jar has zero-deps, but itself name changed

Pass GA.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,7 +165,7 @@
     <jline.version>2.12</jline.version>
     <jms.version>2.0.2</jms.version>
     <joda.version>2.8.1</joda.version>
-    <jodd.version>3.5.2</jodd.version>
+    <jodd.version>6.0.0</jodd.version>
     <json.version>1.8</json.version>
     <junit.version>4.11</junit.version>
     <kryo.version>3.0.3</kryo.version>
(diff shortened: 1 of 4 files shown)


pan3793:
cc @sunchao @luciferyang @wangyum

LuciferYang (approved):

sunchao (approved):

pan3793:
> Although I prefer to get rid of such dependency by https://github.com/apache/hive/pull/4888, but it is objected by the Hive community (please let me know if we can do it in branch-2.3)

the reason for my preference of copying rather than upgrading jodd is, that new jodd changed the calculation to use Java 8 Time API, I'm not sure if this will bring behavior change or not.

pan3793:
Close and in favor https://github.com/apache/hive/pull/5151

PR #4928, opened 2023-12-07
HIVE-23558: Remove compute_stats UDAF

zhangbutao:

[HIVE-23530](https://issues.apache.org/jira/browse/HIVE-23530) replaces this function usage completely.  And it has been deprecated by since Hive 4.0.0-alpha-1, we can remove it safely.

No

No

No need test.


zhangbutao:
@ayushtkn Could you please check if this fix is ok?
Thanks.

ayushtkn (approved):
LGTM

PR #4929, opened 2023-12-07
HIVE-27943: NPE in VectorMapJoinCommonOperator.setUpHashTable when running query with join on date

zabetak:
Handle a `hashTableKeyType` of type `DATE` inside `setUpHashTable`

Avoid the NPE by ensuring that hashTable is initialized when the key is of type `DATE`

No runtime query failures when joining on dates using vectorized map join.

No

```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=mapjoin_date.q
```

### DIFF
--- a/ql/src/test/queries/clientpositive/mapjoin_date.q
+++ b/ql/src/test/queries/clientpositive/mapjoin_date.q
@@ -1,0 +1,6 @@
+set hive.auto.convert.join=true;
+
+CREATE TABLE person (fname string, birthDate date);
+INSERT INTO person VALUES ('Victor', '2023-11-27'), ('Alexandre', '2023-11-28');
+
+SELECT * FROM person p1 INNER JOIN person p2 ON p1.birthDate=p2.birthDate;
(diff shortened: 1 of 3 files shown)


aturoczy (approved):
+1

I know it is not necessary for the qtest, but I always prefer to do it in two separate tables. Logically the test is correct.

ql/src/test/queries/clientpositive/mapjoin_date.q
```diff
@@ -0,0 +1,6 @@
+set hive.auto.convert.join=true;
+
+CREATE TABLE person (fname string, birthDate date);
+INSERT INTO person VALUES ('Victor', '2023-11-27'), ('Alexandre', '2023-11-28');
+
+SELECT * FROM person p1 INNER JOIN person p2 ON p1.birthDate=p2.birthDate;
```
kasakrisz, line 6: Should explain of this query added? Just to confirm that this is really a mapjoin

zabetak: Added explain in
https://github.com/apache/hive/pull/4929/commits/be930513efd36ef6ea08b281abb43025ad2059e6

kasakrisz (approved):
+1, pending tests

PR #4932, opened 2023-12-08
HIVE-27919: Constant reduction in CBO does not work for FROM_UNIXTIME, DATE_ADD, DATE_SUB, TO_UNIX_TIMESTAMP

zabetak:
1. Turn DATE_ADD, DATE_SUB, and FROM_UNIXTIME to regular SqlFunctions and pass the correct return type inference strategy. The operand type inference and type checker can remain null as they were before since they are not used currently in Hive.
2. Change the type family for FROM_UNIXTIME to reflect that the function returns a string (and not date or time).
3. Create and pass an appropriate operand checker for FROM_UNIXTIME (minor since it's not used at the moment).
4. Remove isDynamicFunction override from TO_UNIX_TIMESTAMP (which is wrong), to enable constant reduction.

Constant reduction does not work because the functions are declared as dynamic (isDynamicFunction returns true). However, the dynamic declaration is wrong cause none of the above depends on context variables; they all operate on concrete parameters and require one or more inputs.

Moreover, DATE_ADD, DATE_SUB, and FROM_UNIXTIME are not time functions, so it is wrong to extend the SqlAbsctractTimeFunction class. The overrides in SqlAbsctractTimeFunction are not correct/relevant to these functions so the changes here address this as well.

Simpler and more efficient query plans.

No

```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=cbo_constantfolding.q -Dtest.output.overwrite
```


akshat0395:
The changes looks good +1,
We would probably need to update the corresponding documentation against it: https://cwiki.apache.org/confluence/display/hive/languagemanual+udf for Hive 4.

cc: @aturoczy

akshat0395 (approved):

zabetak:
Many thanks for the review @akshat0395 ! Can you elaborate more on what documentation we need to update? Constant folding/reduction is not something that should be visible to the end-user.

akshat0395:
@ayushtkn has create a space for all Hive-4 specific documentation, the above note was more for me to checkout that document linked and see what all has changed since the inception of that document to Hive-4 and maybe create a Hive-4 specific copy for that doc.
Please share your thoughts around this as well, that would certainly help in improving documentation for Hive-4

kasakrisz (approved):
LGTM

PR #4935, opened 2023-12-10
HIVE-27944:When HIVE-LLAP reads the ICEBERG table, a deadlock may occur.

BsoBird:
Modified the modifier of the cache property in the SplitGroup class to make it non-static.

JIRA: https://issues.apache.org/jira/browse/HIVE-27944

Since the Properties object implement HashTable interface, all the methods of the HashTable interface are synchronised.
In a multi-threaded environment, a deadlock will occur when propA.equals(propB)  and propB.equals(propA) occur at the same time.
The problem can be solved with minimal modification by changing SplitGroup.cache to a non-static property.

no

no

no need. Because this patch doesn't change the execution logic of any code

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
@@ -29,7 +29,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.hadoop.hive.ql.io.AcidUtils;
 import org.apache.hadoop.hive.ql.io.BucketizedHiveInputFormat;
@@ -72,7 +71,7 @@ public class SplitGrouper {
   // TODO This needs to be looked at. Map of Map to Map... Made concurrent for now since split generation
   // can happen in parallel.
   private final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
-      new ConcurrentHashMap<>();
+      new HashMap<>();
 
   private final TezMapredSplitsGrouper tezGrouper = new TezMapredSplitsGrouper();
 
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
@@ -71,7 +71,7 @@ public class SplitGrouper {
 
   // TODO This needs to be looked at. Map of Map to Map... Made concurrent for now since split generation
   // can happen in parallel.
-  private static final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
+  private final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
       new ConcurrentHashMap<>();
 
   private final TezMapredSplitsGrouper tezGrouper = new TezMapredSplitsGrouper();


BsoBird:
@deniskuzZ  @abstractdog  Hi.Could you help review this PR?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
```diff
@@ -71,7 +71,7 @@ public class SplitGrouper {
 
   // TODO This needs to be looked at. Map of Map to Map... Made concurrent for now since split generation
   // can happen in parallel.
-  private static final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
+  private final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
```
zhangbutao, line 74: Are you using Hive4-beta?
If it is possible, can we add a test?

BsoBird: I've looked through the code, and the way I've done it doesn't change any of the logical behaviour of the code. I just defined the cache object as a non-static property to avoid the scenario of multiple threads calling it at the same time.

BsoBird: 况且,如果遵循原本的定义方式,当多个线程同时访问cache,那么有很大概率会出现死锁.

BsoBird: Since the nature of the problem is that HashTable causes deadlocks in concurrent scenarios, I'm not sure what kind of test case would be appropriate to add ....

BsoBird: @zhangbutao  If there's anything else I can provide, please let me know.

deniskuzZ: @BsoBird, with that change the whole caching construct is useless when we have multiple SplitGrouper instances. 
The issue is with concurrent Hashtable.equals(). What if we change the type of PartitionDesc.properties to HashMap?

BsoBird: @deniskuzZ 
hi. I think,modified the Properties to HashMap is too expensive, because HashMap does not have the Properties class of the various related functions. For example: loadFromXML ().
But I agree with you that there are some advantages to be lost by making the cache non-static. So I have a second option:
org.apache.hadoop.hive.common.CopyOnFirstWriteProperties::equals
```
  @Override
  public synchronized boolean equals(Object o) {
    if(o instanceof Hashtable){
      o = new Hashtable<>((Hashtable<Object, Object>) o);
    }
    if (interned != null) return interned.equals(o);
    else return super.equals(o);
  }
```

BsoBird: @deniskuzZ 
The third option is exactly the one you described:
PartitionDesc.equals::
```
    Properties properties1 = getProperties();
    Properties properties2 = other.getProperties();
    cond = (properties1 == null && properties2 == null) ||
        (properties1 != null && Maps.difference(properties1,properties2).areEqual());
    if (!cond) {
      return false;
    }
```

BsoBird: > if removing static, multiple SplitGrouper instances will have each own cache; I am afraid of this will consume much memory of LLAP coordinator (Not sure i understand correctlly).
> 
> From the jstack, it seems that thread is blocked at `org.apache.hadoop.hive.common.CopyOnFirstWriteProperties::equals`, which was introduced to resolve high memory usage about `Properties `, [HIVE-16079](https://issues.apache.org/jira/browse/HIVE-16079). Maybe we can find more clues on `CopyOnFirstWriteProperties`? Just my subjective thought, for your information. Thanks.

@zhangbutao  Since CopyOnFirstWriteProperties is a proxy class for Properties and implements the HashTable interface, CopyOnFirstWriteProperties also triggers deadlocks in concurrent scenarios. You can see the code.

BsoBird: @deniskuzZ Hi, what do you think is the best way for us to proceed?

deniskuzZ: not sure why we need to check properties in `equals`, but the same is done in `TableDesc`. In my opinion, non-synchronized equals should be enough.

BsoBird: @deniskuzZ Okay. let me resubmit a PR.

BsoBird: @deniskuzZ We can use HashMap instead of ConcurrentHashMap, right?

BsoBird: image

deniskuzZ: yes,
@BsoBird , could you please add q test that triggers use of cache from the SplitGrouper.

BsoBird: @deniskuzZ   I'm a newbie in this field, for a long time I can't even debug HIVE code locally(because I always use windows OS computer). I would be grateful if you could guide me. Tks.

deniskuzZ: basically you can add the sql into q file to replicate the use-case and define the expectations in the q.out file
more details on how to execute those tests:
https://cwiki.apache.org/confluence/display/hive/hivedeveloperfaq#HiveDeveloperFAQ-HowdoIruntheclientpositive/clientnegativeunittests?

deniskuzZ: add the SQL you've used to repro the concurrency issue into that q.file
note, in case of iceberg, place the q file under `itests/qtest-iceberg`
{code}
mvn test -Dtest=TestIcebergLlapLocalCliDriver -Dqfile=mytest.q -Drat.skip=true -Dtest.output.overwrite=true
{code}

BsoBird: @deniskuzZ Hi. @deniskuzZ . My current problem is that I can't reproduce the problem consistently. I haven't worked out a way to reproduce the problem consistently yet. So I have no way to add a q.file test case at the moment.

deniskuzZ: @BsoBird, could you please share the repro steps that led to the issue? it doesn't have to be consistent. I just want to create at least one test covering that part of the code.  At this moment we have 0 coverage.

BsoBird: @deniskuzZ  Just Read IcebergTable. and using hive llap.  Before I used this patch, this problem occurred occasionally. After I used this patch, the problem disappeared.

BsoBird: @deniskuzZ If it is difficult to override this change with SQL, can we re-run the previous SQL to test the LLAP+ICEBERG reads and writes? Because our change must affect all the SQL that reads and writes to the ICEBERG table, we just need to verify that the output is the same as before.

BsoBird:
1



---Original---
From: ***@***.***&gt;
Date: Fri, Mar 15, 2024 08:19 AM
To: ***@***.***&gt;;
Cc: "PLASH ***@***.******@***.***&gt;;
Subject: Re: [apache/hive] HIVE-27944:When HIVE-LLAP reads the ICEBERG table,a deadlock may occur. (PR #4935)




 
This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.
 Feel free to reach out on the ***@***.*** list if the patch is in need of reviews.

BsoBird:
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Feel free to reach out on the [dev@hive.apache.org](mailto:dev@hive.apache.org) list if the patch is in need of reviews.

1

deniskuzZ:
I can't just approve this without the test. I wanted to repro the issue locally and then create a q test, but I didn't have time.
Do you have some repro steps that could be tried locally (data setup + selects)?

BsoBird:
Sir.I'm having a hard time triggering this bug directly via SQL because as of now, I'm not sure how to consistently reproduce the problem. in our production env, after we added the patch.The problem disappeared, so I never looked at how to reproduce the bug. Can we test this class in code without using SQL?

PR #4936, opened 2023-12-11
HIVE-27949: GenericUDFAddMonths doesn't accept DECIMAL with scale zer…

soumyakanti3578:
…o for num_months

Second argument of GenericUDFAddMonths `num_months` should accept DECIMAL values with scale zero

DECIMALS like `1.0, 2.0`, etc can be easily converted to `INT`s, so they should be valid arguments for the UDF

Yes, this change enables support of DECIMALS with scale=0 for num_months

No

`mvn test  -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile=udf_add_months.q`
`mvn test -Dtest=TestNegativeLlapCliDriver -Dtest.output.overwrite=true -Dqfile=udf_add_months_error_2.q`
`mvn test -Dtest=TestNegativeLlapCliDriver -Dtest.output.overwrite=true -Dqfile=udf_add_months_error_1.q`

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDF.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDF.java
@@ -38,6 +38,7 @@ import org.apache.hadoop.hive.ql.udf.UDFType;
 import org.apache.hadoop.hive.serde2.io.ByteWritable;
 import org.apache.hadoop.hive.serde2.io.DateWritableV2;
 import org.apache.hadoop.hive.serde2.io.DoubleWritable;
+import org.apache.hadoop.hive.serde2.io.HiveDecimalWritable;
 import org.apache.hadoop.hive.serde2.io.HiveIntervalDayTimeWritable;
 import org.apache.hadoop.hive.serde2.io.HiveIntervalYearMonthWritable;
 import org.apache.hadoop.hive.serde2.io.ShortWritable;
@@ -351,8 +352,12 @@ public abstract class GenericUDF implements Closeable {
     case INT:
     case VOID:
       break;
+    case DECIMAL:
+      if (inOi.scale() == 0) {
+        break;
+      }
     default:
-      throw new UDFArgumentTypeException(i, getFuncName() + " only takes INT/SHORT/BYTE types as "
+      throw new UDFArgumentTypeException(i, getFuncName() + " only takes INT/SHORT/BYTE/DECIMAL(X,0) types as "
           + getArgOrder(i) + " argument, got " + inputType);
     }
 
@@ -607,6 +612,8 @@ public abstract class GenericUDF implements Closeable {
       v = ((ShortWritable) constValue).get();
     } else if (constValue instanceof ByteWritable) {
       v = ((ByteWritable) constValue).get();
+    } else if (constValue instanceof HiveDecimalWritable) {
+      v = ((HiveDecimalWritable) constValue).intValue();
     } else {
       throw new UDFArgumentTypeException(i, getFuncName() + " only takes INT/SHORT/BYTE types as "
           + getArgOrder(i) + " argument, got " + constValue.getClass());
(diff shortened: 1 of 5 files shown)


ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDF.java
```diff
@@ -351,8 +352,12 @@ protected void obtainIntConverter(ObjectInspector[] arguments, int i,
     case INT:
     case VOID:
       break;
+    case DECIMAL:
+      if (inOi.scale() == 0) {
+        break;
+      }
```
okumin, line 358: I think `precision` should also be small enough to ensure type safety. If we accept any DECIMAL, we should semantically accept LONG just in my opinion.

ql/src/test/queries/clientpositive/udf_add_months.q
```diff
@@ -40,3 +40,15 @@ add_months(cast('2016-01-29 10:30:00' as timestamp), 1),
 add_months(cast('2016-02-29 10:30:00' as timestamp), -1),
 add_months(cast('2016-02-29 10:30:12' as timestamp), 2, 'YYYY-MM-dd HH:mm:ss'),
 add_months(cast(null as timestamp), 1);
+
+select
+add_months('2014-01-14 10:30:00', 1.0),
```
okumin, line 45: I'm just curious when we want to give a floating point to `ADD_MONTHS`.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDF.java
```diff
@@ -607,6 +612,8 @@ protected Integer getConstantIntValue(ObjectInspector[] arguments, int i)
       v = ((ShortWritable) constValue).get();
     } else if (constValue instanceof ByteWritable) {
       v = ((ByteWritable) constValue).get();
+    } else if (constValue instanceof HiveDecimalWritable) {
+      v = ((HiveDecimalWritable) constValue).intValue();
     } else {
       throw new UDFArgumentTypeException(i, getFuncName() + " only takes INT/SHORT/BYTE types as "
```
okumin, line 618: It would be more consistent if we updated this message

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDF.java
aturoczy: Agreed

aturoczy (changes requested):
Temporary -1

I do not understand what could be the use-case where the month comes as a decimal. Yes, it is possible to convert and mostly and easy conversion. Just overcomplicates a domain with a use case that is not realistic. When could be a month be 12.43? Not mentioning that this change is in the base GenericUDF base class. It could have side effects for other UDF's

I would rather throw a big red exception into the user face, because it should be something odd. I simply can't imagein that anybody store the month in decimal in any DB. That is a huge anti-pattern that should be punish in several way :) 

cc: @ayushtkn @deniskuzZ @kasakrisz @abstractdog @InvisibleProgrammer 
Please give your opinions as well. Maybe I do not see something, but this behavioural change is odd to me.

InvisibleProgrammer:
Firstly, I agree with @aturoczy , it is better to educate the users about a month is an integer type so please use integer types when you work with month. 

I have two concerns about this change: 
- Now it introduces a decimal -> integer conversation. (Side note, for the first sight, it looks like it would work with decimal -> short conversation as well). Can that extra conversation cause a performance overhead? 
- It modifies `GenericUDF.obtainIntConverter`. It is used in 4 other classes as well. Does it means that change adds decimal support for those as well? 
image

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDF.java
soumyakanti3578: Yes, I agree that we should limit precision. Since we are using `add(int field, int amount)` method of `java.util.Calendar`, where the `amount` type is an `int`, I think we should limit the value to max value of an int.

soumyakanti3578:
Thanks for commenting on the PR. I have seen a case in production where the customer uses a DECIMAL(5, 0) column of a table as the second argument of `add_months`. Since a decimal with scale zero can always be represented as a whole number, I think the customer's case can be supported. But I agree that this is a bad pattern on their end.

I also think we should limit the precision of the DECIMAL because we use `add(int field, int amount)` method of `java.util.Calendar`, where the `amount` type is an `int`.

Regarding the change being in the base class, maybe I can create a new child class and override the method `obtainIntConverter` only for `add_months`.

PR #4938, opened 2023-12-12
HIVE-27950: STACK UDTF returns wrong results when # of argument is no…

okumin:
…t a multiple of N

https://issues.apache.org/jira/browse/HIVE-27950

Fix a data integrity problem of the stack UDF.

This is a bug fix

Yes. This PR would prevent users from facing the issue and would change error messages a bit

No

I added test cases

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
@@ -30,7 +30,6 @@ import org.apache.hadoop.hive.serde2.objectinspector.ConstantObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorFactory;
 import org.apache.hadoop.hive.serde2.objectinspector.StructObjectInspector;
-import org.apache.hadoop.hive.serde2.objectinspector.primitive.WritableConstantIntObjectInspector;
 import org.apache.hadoop.io.IntWritable;
 
 /**
@@ -74,7 +73,7 @@ public class GenericUDTFStack extends GenericUDTF {
       throw new UDFArgumentTypeException(
           0,
           "The first argument to STACK() must be a constant integer (got " +
-              args[0].getTypeName() + " instead).");
+          args[0].getTypeName() + " instead).");
     }
     numRows = (IntWritable) value;
 
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 
 import org.apache.hadoop.hive.ql.exec.Description;
 import org.apache.hadoop.hive.ql.exec.UDFArgumentException;
+import org.apache.hadoop.hive.ql.exec.UDFArgumentTypeException;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.udf.generic.GenericUDFUtils.ReturnObjectInspectorResolver;
 import org.apache.hadoop.hive.serde2.objectinspector.ConstantObjectInspector;
@@ -63,13 +64,21 @@ public class GenericUDTFStack extends GenericUDTF {
     }
     if (!(args[0] instanceof ConstantObjectInspector)) {
       throw new UDFArgumentException(
-          "The first argument to STACK() must be a constant integer (got " +
-          args[0].getTypeName() + " instead).");
+          "The first argument to STACK() must be a constant.");
     }
-    numRows = (IntWritable)
-        ((ConstantObjectInspector)args[0]).getWritableConstantValue();
+    final Object value = ((ConstantObjectInspector) args[0]).getWritableConstantValue();
+    if (value == null) {
+      throw new UDFArgumentException("The first argument of STACK() must not be null.");
+    }
+    if (!(value instanceof IntWritable)) {
+      throw new UDFArgumentTypeException(
+          0,
+          "The first argument to STACK() must be a constant integer (got " +
+              args[0].getTypeName() + " instead).");
+    }
+    numRows = (IntWritable) value;
 
-    if (numRows == null || numRows.get() < 1) {
+    if (numRows.get() < 1) {
       throw new UDFArgumentException(
           "STACK() expects its first argument to be >= 1.");
     }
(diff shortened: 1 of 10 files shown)


ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
```diff
@@ -63,11 +64,16 @@ public StructObjectInspector initialize(ObjectInspector[] args)
     }
     if (!(args[0] instanceof ConstantObjectInspector)) {
       throw new UDFArgumentException(
+          "The first argument to STACK() must be a constant.");
+    }
+    final Object value = ((ConstantObjectInspector) args[0]).getWritableConstantValue();
+    if (!(value instanceof IntWritable)) {
```
okumin, line 70: I added a validation since the current implementation throws a ClassCastException. I can split PRs if we don't want to fix it in this PR.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
```diff
@@ -63,13 +64,21 @@ public StructObjectInspector initialize(ObjectInspector[] args)
     }
     if (!(args[0] instanceof ConstantObjectInspector)) {
       throw new UDFArgumentException(
+          "The first argument to STACK() must be a constant.");
```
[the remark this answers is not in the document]
okumin: The first argument, # of rows to be generated from a single row, defines the final schema. That's why it must be evaluated at compile time.

```
0: jdbc:hive2://hive-hiveserver2:10000/defaul> select stack(2, 'a', 'b', 'c', 'd', 'e', 'f');
...
+-------+-------+-------+
| col0  | col1  | col2  |
+-------+-------+-------+
| a     | b     | c     |
| d     | e     | f     |
+-------+-------+-------+
2 rows selected (3.979 seconds)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> select stack(3, 'a', 'b', 'c', 'd', 'e', 'f');
...
+-------+-------+
| col0  | col1  |
+-------+-------+
| a     | b     |
| c     | d     |
| e     | f     |
+-------+-------+
```

InvisibleProgrammer:
LGTM

InvisibleProgrammer (approved):

okumin:
@aturoczy @InvisibleProgrammer Do you know anyone who can get this PR approved? Thanks.

aturoczy:
@SourabhBadhya @deniskuzZ @abstractdog Could you pls check this?

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
```diff
@@ -63,13 +64,21 @@ public StructObjectInspector initialize(ObjectInspector[] args)
...
+    final Object value = ((ConstantObjectInspector) args[0]).getWritableConstantValue();
+    if (value == null) {
+      throw new UDFArgumentException("The first argument of STACK() must not be null.");
+    }
+    if (!(value instanceof IntWritable)) {
+      throw new UDFArgumentTypeException(
+          0,
           "The first argument to STACK() must be a constant integer (got " +
-          args[0].getTypeName() + " instead).");
+              args[0].getTypeName() + " instead).");
```
SourabhBadhya, line 77: nit: Unnecessary change

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
```diff
@@ -109,15 +118,15 @@ public StructObjectInspector initialize(ObjectInspector[] args)
...
+      throws HiveException {
     for (int ii = 0; ii < numRows.get(); ++ii) {
       for (int jj = 0; jj < numCols; ++jj) {
         int index = ii * numCols + jj + 1;
         if (index < args.length) {
           forwardObj[jj] = 
             returnOIResolvers.get(jj).convertIfNecessary(args[index], argOIs.get(index));
         } else {
-          forwardObj[ii] = null;
+          forwardObj[jj] = null;
```
SourabhBadhya, line 129: I think you could have fixed a bug here, is there a valid test case which exposes this bug which can be added here?

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
```diff
@@ -63,13 +64,21 @@ public StructObjectInspector initialize(ObjectInspector[] args)
     }
     if (!(args[0] instanceof ConstantObjectInspector)) {
       throw new UDFArgumentException(
+          "The first argument to STACK() must be a constant.");
+    }
+    final Object value = ((ConstantObjectInspector) args[0]).getWritableConstantValue();
+    if (value == null) {
+      throw new UDFArgumentException("The first argument of STACK() must not be null.");
+    }
+    if (!(value instanceof IntWritable)) {
```
SourabhBadhya, line 73: Can a long be passed here (LongWritable)?

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
okumin: I added a new statement to `ql/src/test/queries/clientpositive/udtf_stack.q`. The SQL on the current master can't return the correct result.

```
0: jdbc:hive2://hive-hiveserver2:10000/defaul> SELECT STACK(2, 'a', 'b', 'c', 'd', 'e');
...
+-------+-------+-------+
| col0  | col1  | col2  |
+-------+-------+-------+
| a     | b     | c     |
| d     | NULL  | c     |
+-------+-------+-------+
```

This is the correct result.

```
POSTHOOK: query: SELECT STACK(2, 'a', 'b', 'c', 'd', 'e')
POSTHOOK: type: QUERY
POSTHOOK: Input: _dummy_database@_dummy_table
a	b	c
d	e	NULL
```

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
okumin: Modified
https://github.com/apache/hive/pull/4938/commits/8b109aa427c65f8443d403f6c3288630a1dfc1ed

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
okumin: Yes, but I believe there is no valid use case for two reasons.
- This UDF can't process N > Integer.MAX_VALUE anyway, as the maximum length of Java's arrays is Integer.MAX_VALUE
- I don't think anyone prefers to write `stack(cast(3 as bigint), a, b, c, ...`

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDTFStack.java
SourabhBadhya: Nice, glad that a relevant test is added for this bug. Thank you.

SourabhBadhya (approved):
LGTM, you must probably rebase and retrigger the tests.

okumin:
@SourabhBadhya Thanks for your review! I rebased this branch and confirmed all tests passed 👍

okumin:
Thanks for reviewing and merging this PR 👍

PR #4943, opened 2023-12-12
HIVE-27924: Incremental rebuild goes wrong when inserts and deletes overlap between the source tables

kasakrisz:
To generate incremental materialized view rebuild plan when any of the source tables has delete operations deleted rows has to be identified. Hence `HiveTableScan` operators scan both insert and delete delta files and fetch deleted rows too. The boolean typed virtual column `ROW_IS_DELETED` indicates whether a row is deleted or not. Based on this column's value and a filter on `writeId` helps identify newly inserted and deleted rows. Rebuild plan in such cases is a plan of a multi insert statement representing a merge statement. In order to split the rows between insert/update and delete branches extra filter conditions are propagated into the plan. These conditions are using `ROW_IS_DELETED` column and predicates with `writeId`s. The propagation is implemented in `HiveRowIsDeletedPropagator`.
1. This patch fixes the propagator by adding filters on top of HiveJoin operators to filter out result rows of the join when the row coming from the left side is deleted and the one coming from the right side is newly inserted and the have the same join key values. Later these filter conditions may pushed down to join the condition by follow up optimizer rules.
2. Disable incremental materialized view rebuild when the MV definition does not have aggregate but any source tables has delete operation. 
1. Resolve data correctness issues as described in the jira. https://issues.apache.org/jira/browse/HIVE-27924.
2. Rows which should be deleted from MV can not be determined without a unique identifier of the rows in source tables at rebuild time. Currently we don't have such unique identifier which we could rely on:
* unique and primary key constraints are not enforced in Hive
* the other option would be to project ROW_ID from each source table but the writeId component is changed at delete.

Yes. Result set of materialized which definition contains join operators may change in order to show correct results. Execution plan of incremental materialized view rebuild may also change.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=materialized_view_create_rewrite_6.q,materialized_view_create_rewrite_6_aggr_2joins.q,materialized_view_create_rewrite_6_aggr_3joins.q -pl itests/qtest -Pitests
```


aturoczy (approved):
Pretty significant change.
This is must have for Hive 4
cc: @deniskuzZ @abstractdog

deniskuzZ:
> Pretty significant change. This is must have for Hive 4 cc: @deniskuzZ @abstractdog

checking

kasakrisz:
Depends on #5053

hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/HCatRecordReader.java
```diff
@@ -226,7 +226,7 @@ public void close() throws IOException {
   /**
    * Tracks number of of errors in input and throws a Runtime exception
    * if the rate of errors crosses a limit.
-   * <br/>
+   * <br>
```
deniskuzZ, line 229: why is this change, i don't see closing tag

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/AlterMaterializedViewRebuildAnalyzer.java
```diff
@@ -480,8 +515,6 @@ protected ASTNode fixUpAfterCbo(ASTNode originalAst, ASTNode newAst, CalcitePlan
         fixUpASTAggregateInsertDeleteIncrementalRebuild(fixedAST, getMaterializedViewASTBuilder());
         return fixedAST;
       case JOIN_INSERT_DELETE_REBUILD:
```
deniskuzZ, line 517: why do we need an empty `case` for JOIN_INSERT_DELETE_REBUILD?

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/AlterMaterializedViewRebuildAnalyzer.java
```diff
@@ -322,8 +373,7 @@ private RelNode applyRecordIncrementalRebuildPlan(
               }
               return applyAggregateInsertDeleteIncremental(basePlan, mdProvider, executorProvider);
             } else {
-              return applyJoinInsertDeleteIncremental(
-                      basePlan, mdProvider, executorProvider, optCluster, calcitePreMVRewritingPlan);
+              return calcitePreMVRewritingPlan;
```
deniskuzZ, line 376: can we remove nesting and combine 
````
if (acidView && visitor.isContainsAggregate()) {
....
}
return calcitePreMVRewritingPlan;
````

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAggregateInsertDeleteIncrementalRewritingRule.java [hunk 4]
```diff
@@ -116,7 +121,7 @@ protected IncrementalComputePlanWithDeletedRows createJoinRightInput(RelOptRuleC
     aggInput = HiveHepExtractRelNodeRule.execute(aggInput);
     aggInput = new HiveRowIsDeletedPropagator(relBuilder).propagate(aggInput);
 
-    int rowIsDeletedIdx = aggInput.getRowType().getFieldCount() - 1;
+    int rowIsDeletedIdx = aggInput.getRowType().getFieldCount() - 2;
```
deniskuzZ, line 124: should we add some comment for this magic number `-2`?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/MaterializedViewRewritingRelVisitor.java
```diff
@@ -50,7 +49,7 @@ public class MaterializedViewRewritingRelVisitor extends RelVisitor {
 
 
   private boolean containsAggregate;
-  private boolean fullAcidView;
+  private final boolean fullAcidView;
```
deniskuzZ, line 52: others don't have to be final?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -17,72 +17,139 @@
...
+ * This class is part of incremental rebuild of materialized view plan generation.
+ * <br>
+ * @see AlterMaterializedViewRebuildAnalyzer
+ * @see HiveAggregateInsertDeleteIncrementalRewritingRule
  */
-public class HiveRowIsDeletedPropagator extends HiveRelShuttleImpl {
+public class HiveRowIsDeletedPropagator implements ReflectiveVisitor {
 
   protected final RelBuilder relBuilder;
+  protected final Deque<RelNode> stack = new ArrayDeque<>();
```
deniskuzZ, line 68: should it be protected (also relBuilder)

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -17,72 +17,139 @@
...
-    if (column == null) {
-      // This should not happen since Virtual columns are propagated for all native table scans in
-      // CalcitePlanner.genTableLogicalPlan()
-      throw new ColumnPropagationException("TableScan " + scan + " row schema does not contain " +
-          VirtualColumn.ROWISDELETED.getName() + " virtual column");
+    return dispatcher.invoke(relNode, new Context());
+  }
+
+  protected RelNode visitChild(RelNode parent, int i, RelNode child, Context context) {
+    stack.push(parent);
```
deniskuzZ, line 82: what does it do? i don't see how it's being used

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -17,72 +17,139 @@
...
+      }
+      return parent;
+    } finally {
+      stack.pop();
     }
+  }
+
+  protected RelNode visitChildren(RelNode rel, Context context) {
+    for (Ord<RelNode> input : Ord.zip(rel.getInputs())) {
+      rel = visitChild(rel, input.i, input.e, context);
```
deniskuzZ, line 98: doesn't it override `rel` in a loop, is that expected?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -17,72 +17,139 @@
...
+    }
+    return rel;
+  }
+
+  public static class Context {
```
deniskuzZ, line 103: should it be public?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -92,86 +159,155 @@ public RelNode visit(HiveTableScan scan) {
...
-   */
-  @Override
-  public RelNode visit(HiveProject project) {
-    RelNode newProject = visitChild(project, 0, project.getInput());
+  // Add the new columns(_deleted, _inserted) to the original project
+  public RelNode visit(HiveProject project, Context context) {
+    RelNode newProject = visitChild(project, 0, project.getInput(), context);
     RelNode projectInput = newProject.getInput(0);
-    int rowIsDeletedIndex = projectInput.getRowType().getFieldCount() - 1;
+
```
deniskuzZ, line 166: shouldn't the size be `project.getProjects()+2`?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -92,86 +159,155 @@ public RelNode visit(HiveTableScan scan) {
...
+        rightContext.rowIdPredicates.put(entry.getKey() - originalLeftFieldCount,
+          new InputRefShifter(originalLeftFieldCount, -originalLeftFieldCount, relBuilder).apply(entry.getValue()));
+      }
+    }
+    tmpJoin = visitChild(join, 1, join.getInput(1), rightContext);
+    RelNode newRightInput = tmpJoin.getInput(1);
+    RelDataType newRightRowType = newRightInput.getRowType();
+
+    // Create input refs to propagated columns in left and right inputs
+    int rightAnyDeletedIndex = newRightRowType.getFieldCount() - 2;
```
deniskuzZ, line 243: could it be negative?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -92,86 +159,155 @@ public RelNode visit(HiveTableScan scan) {
...
     int newLeftFieldCount;
-    if (join.getInput(0).getRowType().getField(VirtualColumn.ROWISDELETED.getName(), false, false) == null) {
-      // Shift column references refers columns coming from right input by one in join condition since the new left input
-      // has a new column
-      newJoinCondition = new InputRefShifter(leftRowType.getFieldCount() - 1, relBuilder)
-          .apply(join.getCondition());
-
-      newLeftFieldCount = leftRowType.getFieldCount() - 1;
+    int shift;
+    if (newLeftRowType.getField("_rowId_greater_than", false, false) == null) {
```
deniskuzZ, line 260: is there a constant for "_rowId_greater_than"

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/MaterializedViewRewritingRelVisitor.java
kasakrisz: This class is not stateless. Fields
```
containsAggregate
rewritingAllowed
countIndex
```
are set during plan traversal but `fullAcidView` should not be changed since it's value depends on the MV metadata.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: In each iteration of the for loop a child of `rel` is updated. 
The method
```
visitChild(RelNode parent, int i, RelNode child, Context context)
```
returns the parent passed as the first parameter but the returned instance's ith child is updated.

So in the last iteration we end up with a rel which all children are updated. This is the goal of this method.

It is coming from Calcite https://github.com/apache/calcite/blob/abf462a7ddab11a44610827b0cd69510c099d9b6/core/src/main/java/org/apache/calcite/rel/RelShuttleImpl.java#L54-L74

My idea was to extend  the Calcite implementation but it doesn't allow passing extra parameters (`Context` in this case).

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: It is public because it is exposed by the `visit` methods.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: No. At this point the new right input schema has at least two new columns.

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/AlterMaterializedViewRebuildAnalyzer.java
kasakrisz: removed by #5053

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/AlterMaterializedViewRebuildAnalyzer.java
kasakrisz: Combined

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveAggregateInsertDeleteIncrementalRewritingRule.java
kasakrisz: This was removed by #5053

kasakrisz: Added comment.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: removed `stack` and changed `relBuilder` and some protected methods to private

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: This came from a plan traversal algorithm implemented in Calcite but we don't need the full functionality of that so I removed.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: Fixed.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -92,86 +159,155 @@ public RelNode visit(HiveTableScan scan) {
...
+      return visitChild(filter, 0, filter.getInput(0), context);
+    }
+
+    if (!condition.isA(SqlKind.OR)) {
+      return visitChild(filter, 0, filter.getInput(0), context);
+    }
+
+    for (RexNode operand : ((RexCall)condition).operands) {
+      RexInputRef inputRef = findPossibleRowIdRef(operand);
+      if (inputRef == null) {
```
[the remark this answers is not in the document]
kasakrisz: Changed.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: Removed this if statement because only the true branch is executed.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/MaterializedViewRewritingRelVisitor.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -17,72 +17,128 @@
...
+    } else {
+      // A row is deleted if ROW_IS_DELETED is true and rowId > <saved_rowId>
+      projects.add(rexBuilder.makeCall(SqlStdOperatorTable.AND, rowIsDeleted, rowIdPredicate));
+      // A row is newly inserted if ROW_IS_DELETED is false and rowId > <saved_rowId>
+      projects.add(rexBuilder.makeCall(SqlStdOperatorTable.AND,
+          rexBuilder.makeCall(SqlStdOperatorTable.NOT, rowIsDeleted), rowIdPredicate));
+    }
+    String rowIsDeletedName = projectNames.remove(rowIsDeletedField.getIndex());
+    projectNames.add(rowIsDeletedName);
+    projectNames.add("_deleted");
```
deniskuzZ, line 140: should we move that to constants?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
```diff
@@ -92,86 +148,146 @@ public RelNode visit(HiveTableScan scan) {
...
+    List<RexNode> projects = new ArrayList<>(newLeftFieldCount + newRightRowType.getFieldCount() + 1);
+    List<String> projectNames = new ArrayList<>(newLeftFieldCount + newRightRowType.getFieldCount() + 1);
+    populateProjects(rexBuilder, newLeftRowType, 0, newLeftFieldCount, projects, projectNames);
+    populateProjects(rexBuilder, newRightRowType, newLeftRowType.getFieldCount(),
+        newRightRowType.getFieldCount() - 2, projects, projectNames);
+
+    // Create derived expressions
+    projects.add(rexBuilder.makeCall(SqlStdOperatorTable.OR, leftDeleted, rightDeleted));
+    projects.add(rexBuilder.makeCall(SqlStdOperatorTable.OR, leftInserted, rightInserted));
+    projectNames.add("_any_deleted");
```
deniskuzZ, line 258: same

deniskuzZ (approved):
LGTM +1, minor comments

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: Moved.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/HiveRowIsDeletedPropagator.java
kasakrisz: Moved.

PR #4945, opened 2023-12-13
HIVE-27161: MetaException when executing CTAS query in Druid storage handler

kasakrisz:
In case of ctas commands when destination table has storage handler specified check whether that storage handler requires location to be set at table creation.

Druid storage handler doesn't allow setting location explicitly:
https://github.com/apache/hive/blob/5e9695cbdded7d7a40ecd9fc93ee6803625bb197/druid-handler/src/java/org/apache/hadoop/hive/druid/DruidStorageHandler.java#L217-L220

No

No

Temporary remove Ignore annotation from `TestMiniDruidCliDriver`
```
mvn test -Dtest=TestMiniDruidCliDriver -Dqfile=druidmini_expressions.q -pl itests/qtest -Pitests
mvn test -Dtest.output.overwrite -Dtest=TestIcebergCliDriver -Dqfile=ctas_iceberg_orc.q -pl itests/qtest-iceberg -Piceberg -Pitests
```

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -2014,4 +2014,9 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
       throw new SemanticException(String.format("Error while fetching the partitions due to: %s", e));
     }
   }
+
+  @Override
+  public boolean requiresLocation() {
+    return true;
+  }
 }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -73,6 +73,7 @@ import org.apache.hadoop.hive.ql.QueryState;
 import org.apache.hadoop.hive.ql.ddl.DDLOperationContext;
 import org.apache.hadoop.hive.ql.ddl.table.AbstractAlterTableDesc;
 import org.apache.hadoop.hive.ql.ddl.table.AlterTableType;
+import org.apache.hadoop.hive.ql.ddl.table.create.CreateTableDesc;
 import org.apache.hadoop.hive.ql.ddl.table.create.like.CreateTableLikeDesc;
 import org.apache.hadoop.hive.ql.ddl.table.misc.properties.AlterTableSetPropertiesDesc;
 import org.apache.hadoop.hive.ql.exec.ColumnInfo;
@@ -1648,6 +1649,11 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   }
 
   @Override
+  public void setTableLocationForCTAS(CreateTableDesc desc, String location) {
+    desc.setLocation(location);
+  }
+
+  @Override
   public Map<String, String> getNativeProperties(org.apache.hadoop.hive.ql.metadata.Table table) {
     Table origTable = IcebergTableUtil.getTable(conf, table.getTTable());
     Map<String, String> props = Maps.newHashMap();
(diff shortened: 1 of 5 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -7949,7 +7949,15 @@ protected Operator genFileSinkPlan(String dest, QB qb, Operator input)
           }
         } else {
           if (tblDesc.isCTAS() && tblDesc.getStorageHandler() != null) {
-            tblDesc.setLocation(getCtasOrCMVLocation(tblDesc, viewDesc, createTableUseSuffix).toString());
+            try {
```
deniskuzZ, line 7952: can't we simply do `if (tblDesc.isCTAS() && isIcebergTable(tblProps)) {`

deniskuzZ: btw instead `createTableUseSuffix` better to directly use `false`

aturoczy (changes requested):
Overall, the change is good, I would just adjust @deniskuzZ is requested. Everything else is good.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
kasakrisz: This part of the code is in the compiler and it should not depend on Iceberg or any storage format directly. Hive has `HiveIcebergStorageHandler` interface to hide storage format customizations.

kasakrisz: "btw instead createTableUseSuffix better to directly use false"
changed to false

deniskuzZ: I agree that compiler code should be `storage format` agnostic, however:
1. SA already has specialized logic for external / ACID / Iceberg (`isIcebergTable()`) --> requires refactor, we can ignore for now;
2. `HiveStorageHandler` is being abused and polluted with many meaningless methods. `SH#requiresLocation()` - what does it even mean? Could we please use self-explanatory naming in the interface?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -7949,7 +7949,15 @@ protected Operator genFileSinkPlan(String dest, QB qb, Operator input)
...
+              HiveStorageHandler storageHandler =
+                  HiveUtils.getStorageHandler(SessionState.getSessionConf(), tblDesc.getStorageHandler());
+              if (storageHandler.requiresLocationAtCreateTable()) {
+                tblDesc.setLocation(getCtasOrCMVLocation(tblDesc, viewDesc, false).toString());
+              }
+            } catch (Exception ex) {
+              LOG.warn("Error while creating HiveStorageHandler instance: {}", tblDesc.getStorageHandler(), ex);
```
deniskuzZ, line 7959: is that OK, just to swallow the exception?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -2014,4 +2014,9 @@ public List<Partition> getPartitionsByExpr(org.apache.hadoop.hive.ql.metadata.Ta
       throw new SemanticException(String.format("Error while fetching the partitions due to: %s", e));
     }
   }
+
+  @Override
+  public boolean requiresLocationAtCreateTable() {
```
deniskuzZ, line 2019: maybe `requiresLocationForCTAS` as it's only CTAS specific?

deniskuzZ: or maybe similar to   
````
default void setTableParametersForCTLT(org.apache.hadoop.hive.ql.metadata.Table tbl, CreateTableLikeDesc desc,
      Map<String, String> origParams) {
  }
````
setLocationForCTAS?

deniskuzZ (approved):
LGTM +1, pending tests

PR #4947, opened 2023-12-14
HIVE-27952: Use SslContextFactory.Server() instead of SslContextFactory

ngsg:

Use SslContextFactory.Server() instead of SslContextFactory() when constructing the factory.

With Jetty 9.4.40, the base constructor fails to handle a KeyStore with multiple certificates in it.
```
Caused by: java.lang.IllegalStateException:
KeyStores with multiple certificates are not supported on the base class org.eclipse.jetty.util.ssl.SslContextFactory.
(Use org.eclipse.jetty.util.ssl.SslContextFactory$Server or org.eclipse.jetty.util.ssl.SslContextFactory$Client instead)
```

No

No

manually tested on a secure cluster

### DIFF
--- a/common/src/java/org/apache/hive/http/HttpServer.java
+++ b/common/src/java/org/apache/hive/http/HttpServer.java
@@ -536,7 +536,7 @@ public class HttpServer {
     if (!b.useSSL) {
       connector = new ServerConnector(webServer, http);
     } else {
-      SslContextFactory sslContextFactory = new SslContextFactory();
+      SslContextFactory sslContextFactory = new SslContextFactory.Server();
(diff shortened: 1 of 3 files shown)


aturoczy:
Can we get an attention on this PR pls? I think it is a simple change to merge. All the test are passed and as I see it is valuable for our community.

cc: @nrg4878 @ayushtkn @deniskuzZ

aturoczy:
Can we get an attention on this PR pls? I think it is a simple change to merge. All the test are passed and as I see it is valuable for our community.

cc: @nrg4878 @ayushtkn @deniskuzZ

nrg4878 (approved):
Changes look good to me.

PR #4949, opened 2023-12-15
HIVE-27749: SchemaTool initSchema fails on Mariadb 10.2

SourabhBadhya:

HIVE-27749: SchemaTool initSchema fails on Mariadb 10.2

initSchema fails on MariaDB.

No

No

On a MariaDB SQL environment.


InvisibleProgrammer:
Hi, I understand that we don't expect other value for NNI_ID than 1. But I have some second thoughts about this change: Firstly, the original logic provided a default value. Now, If I understand it correctly, we change it to enforcing 1. But we only enforce it in case the customer uses MariaDb. 
I see two painful points in that: Firstly, it can surprise us in a future, if we want to change how notification is working. 
Secondly, we are creating different Metastore schema for Mariadb and other databases. I think we should target consistency when we support multiple database providers.

SourabhBadhya:
@InvisibleProgrammer Thanks for the comment. I have seen this behaviour wherein the NOTIFICATION_SEQUENCE table had multiple records and this pretty much messed up the notification log functionality which can spiral into an ugly situation in Hive clusters with an ugly workaround (deleting the record manually in this table). As a long term solution, I had added the one-row constraint on the NOTIFICATION_SEQUENCE table in HIVE-26324. Unfortunately, one solution on MySQL didn't work on the MariaDB database. Hence I am making it consistent in this PR.

InvisibleProgrammer:
> @InvisibleProgrammer Thanks for the comment. I have seen this behaviour wherein the NOTIFICATION_SEQUENCE table had multiple records and this pretty much messed up the notification log functionality which can spiral into an ugly situation in Hive clusters with an ugly workaround (deleting the record manually in this table). As a long term solution, I had added the one-row constraint on the NOTIFICATION_SEQUENCE table in [HIVE-26324](https://issues.apache.org/jira/browse/HIVE-26324). Unfortunately, one solution on MySQL didn't work on the MariaDB database. Hence I am making it consistent in this PR.

Got it, thx.

standalone-metastore/metastore-server/src/main/sql/mysql/hive-schema-4.0.0-alpha-2.mysql.sql
```diff
@@ -892,7 +892,7 @@ CREATE TABLE IF NOT EXISTS `NOTIFICATION_SEQUENCE`
     PRIMARY KEY (`NNI_ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
-ALTER TABLE `NOTIFICATION_SEQUENCE` MODIFY COLUMN `NNI_ID` BIGINT(20) GENERATED ALWAYS AS (1) STORED NOT NULL;
+ALTER TABLE `NOTIFICATION_SEQUENCE` ADD CONSTRAINT `ONE_ROW_CONSTRAINT` CHECK (`NNI_ID` = 1);
```
deniskuzZ, line 895: `ONLY_ONE_ROW`?

SourabhBadhya: @deniskuzZ This is already `ONE_ROW_CONSTRAINT` in other DB scripts so I think its best to not change here.

deniskuzZ:
@SourabhBadhya, have you checked if that syntax is compliant with all Hive supported versions of mariaDb and MySQL?

SourabhBadhya:
@deniskuzZ This syntax is supported by MySQL 8.0+ and MariaDB 10.2+. Any versions prior to this has effectively been declared End of Life (EOL) already.

deniskuzZ (approved):
LGTM +1

PR #4951, opened 2023-12-15
HIVE-27948: Wrong results when using materialized views with non-deterministic/dynamic functions

kasakrisz:
1. Collect available rewrite algorithms when validating materialized view definition query plan.
If none of the algorithms selected do not register the MV for automatic rewrites.
2. Print warning message to console when enabling MV for rewrite but definition query has unsupported parts for algebraic rewrite algorithm.
3. If MV definition has non-deterministic function call none of the rewrite algorithms are selected.

Enabling MVs having non-deterministic functions in automatic query rewrites leads to data correctness issues. See jira for details: https://issues.apache.org/jira/browse/HIVE-27948

Yes, query results and explain plan may change.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=materialized_view_no_cbo_rewrite.q,materialized_view_rewrite_by_text_5.q,materialized_view_create_rewrite_6.q,materialized_view_rewrite_3.q,materialized_view_rewrite_by_text_10.q -pl itests/qtest -Pitests
```


ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rewrite/AlterMaterializedViewRewriteOperation.java
```diff
@@ -64,9 +68,12 @@ public int execute() throws HiveException {
           }
           throw new HiveException(msg);
         }
-        if (!planner.isValidAutomaticRewritingMaterialization()) {
-          throw new HiveException("Cannot enable rewriting for materialized view. " +
-              planner.getInvalidAutomaticRewritingMaterializationReason());
+        AutomaticRewritingValidationResult validationResult = planner.getAutomaticRewritingValidationResult();
+        String validationErrorMessage = validationResult.getErrorMessage();
+        if (validationResult.getSupportedRewriteAlgorithms().isEmpty()) {
+          throw new HiveException(validationErrorMessage);
```
zabetak, line 74: There seems to be some inconsistency between CREATE and ALTER VIEW statements. Here we throw an exception when there are no rewrite algorithms but in `SemanticAnalyzer#validateCreateView` we simply log an error to the console.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -14668,11 +14669,9 @@ private void validateCreateView()
           }
           throw new SemanticException(msg);
         }
-        if (!isValidAutomaticRewritingMaterialization()) {
-          String errorMessage = "Only query text based automatic rewriting is available for materialized view. " +
-                  getInvalidAutomaticRewritingMaterializationReason();
-          console.printError(errorMessage);
-          LOG.warn(errorMessage);
+        String validationWarningMessage = automaticRewritingValidationResult.getErrorMessage();
```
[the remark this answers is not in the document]
zabetak: Why don't we check the supported algorithms here?

ql/src/test/org/apache/hadoop/hive/ql/metadata/TestMaterializedViewsCache.java
```diff
@@ -157,7 +157,7 @@ private Table getTable(String db, String tableName, String definition) {
   private static HiveRelOptMaterialization createMaterialization(Table table) throws ParseException {
     return new HiveRelOptMaterialization(
             new DummyRel(table), new DummyRel(table), null, asList(table.getDbName(), table.getTableName()),
-            EnumSet.allOf(HiveRelOptMaterialization.RewriteAlgorithm.class),
+            EnumSet.allOf(RewriteAlgorithm.class),
```
zabetak, line 160: Should we use RewriteAlgorithm.ALL instead?

ql/src/test/results/clientpositive/llap/materialized_view_no_cbo_rewrite.q.out
```diff
@@ -36,4 +36,8 @@ PREHOOK: query: alter materialized view cmv_mat_view enable rewrite
 PREHOOK: type: ALTER_MATERIALIZED_VIEW_REWRITE
 PREHOOK: Input: default@cmv_mat_view
 PREHOOK: Output: default@cmv_mat_view
-FAILED: Execution Error, return code 40000 from org.apache.hadoop.hive.ql.ddl.DDLTask. org.apache.hadoop.hive.ql.metadata.HiveException: Cannot enable rewriting for materialized view. Statement has unsupported clause: sort by.
+Only query text based automatic rewriting is available for materialized view. Statement has unsupported clause: sort by.
```
zabetak, line 39: It seems that with this patch we also enable more rewritting capabilities that we didn't support before. Maybe worth mentioning in the JIRA ticket as well.

ql/src/test/queries/clientpositive/materialized_view_rewrite_by_text_10.q
```diff
@@ -0,0 +1,11 @@
+-- Materialzed view definition has non-deterministic function
+set hive.support.concurrency=true;
+set hive.txn.manager=org.apache.hadoop.hive.ql.lockmgr.DbTxnManager;
+
+CREATE TABLE EMPS (ENAME STRING, BIRTH_EPOCH_SECS INT) STORED AS ORC TBLPROPERTIES ('transactional'='true');
+
+CREATE MATERIALIZED VIEW v_emp AS SELECT * FROM EMPS WHERE BIRTH_EPOCH_SECS <= UNIX_TIMESTAMP();
```
zabetak, line 7: Shouldn't the CREATE fail completely? I mentioned this in the code as well but I don't understand why CREATE and ALTER behave differently.

ql/src/test/queries/clientpositive/materialized_view_rewrite_by_text_11.q
```diff
@@ -0,0 +1,13 @@
...
+set hive.materializedview.rewriting=true;
+
+create table cmv_basetable (a int, b varchar(256), c decimal(10,2))
+stored as orc TBLPROPERTIES ('transactional'='true');
+
+insert into cmv_basetable values (1, 'alfred', 10.30),(2, 'bob', 3.14),(2, 'bonnie', 172342.2),(3, 'calvin', 978.76),(3, 'charlie', 9.8);
+
+create materialized view cmv_mat_view disable rewrite as select t1.a from cmv_basetable t1 left outer join cmv_basetable t2 on (t1.a = t1.b);
+
+alter materialized view cmv_mat_view enable rewrite;
```
zabetak, line 13: This appears as a new file although in fact it is a negative test that was moved. Did you try to use `git mv` for performing the change so that it appears better in git history/diff.

ql/src/test/queries/clientpositive/materialized_view_rewrite_by_text_10.q
zabetak: Actually thinking a bit more on this, it may be fine to allow the creation of the materialized view as long as it does not create issues with query rewriting and (incremental) refresh.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptMaterializationValidator.java
```diff
@@ -308,24 +318,19 @@ public boolean isValidForQueryCaching() {
...
-    if (isValidForAutomaticRewriting()) {
-      this.automaticRewritingInvalidReason = automaticRewritingInvalidReason;
-    }
+  public AutomaticRewritingValidationResult getAutomaticRewritingValidationResult() {
+    return automaticRewritingValidationResult;
   }
 
-  public void setAutomaticRewritingInvalidReason(RelNode node) {
+  public void unsupportedByCalciteRewrite(String sqlPartType, String sqlPart) {
     if (isValidForAutomaticRewriting()) {
```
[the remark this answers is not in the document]
kasakrisz: There are two types of failures:
1. disable both rewrite algs.
2. allow text base only
If type 1) is found we can stop traversing the tree.
If type 2) is found we record it if this is the first one from this type but we have to continue traversing because there might be a type 1).

Actually we print only one failure and if there are more than one from type 2) it doesn't matter which one is printed. I kept the logic avoid too many golden file changes.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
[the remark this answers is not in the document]
kasakrisz: Renamed and disable rewrite if no rewrite alg is available.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptMaterializationValidator.java
```diff
@@ -62,13 +65,16 @@
  * - References to non-deterministic functions.
  */
 public class HiveRelOptMaterializationValidator extends HiveRelShuttleImpl {
-  static final Logger LOG = LoggerFactory.getLogger(HiveRelOptMaterializationValidator.class);
+  public static final String UNSUPPORTED_BY_CALCITE_FORMAT =
```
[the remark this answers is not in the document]
kasakrisz: done

ql/src/java/org/apache/hadoop/hive/ql/metadata/AutomaticRewritingValidationResult.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.hadoop.hive.ql.metadata;
+
+import org.apache.calcite.plan.RelOptMaterialization;
+
+import java.util.EnumSet;
+import java.util.Set;
+
+/**
+ * Hive extension of {@link RelOptMaterialization}.
+ */
```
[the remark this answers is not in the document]
kasakrisz: Removed javadoc and rename the class to MaterializationValidationResult

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rewrite/AlterMaterializedViewRewriteOperation.java
kasakrisz: Changed that create mv not just log the error but also disables rewrite for MV.

ql/src/test/org/apache/hadoop/hive/ql/metadata/TestMaterializedViewsCache.java
kasakrisz: done

ql/src/test/queries/clientpositive/materialized_view_rewrite_by_text_10.q
kasakrisz: Yes. Creation should not fail but a message is logged and MV is disabled for rewrite if none of the algorithms are available.

In case of Alter the goal is to enable rewrite but if none of the algorithms are available I think it is ok to fail.

ql/src/test/queries/clientpositive/materialized_view_rewrite_by_text_11.q
kasakrisz: I tried it. Let's see the result :)

ql/src/test/results/clientpositive/llap/materialized_view_no_cbo_rewrite.q.out
kasakrisz: Actually these capabilities already exists if rewrite is not disabled at creation time. However Alter doesn't allow these. This patch tries to improve Alter too.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -15930,18 +15932,14 @@ public ColsAndTypes(String cols, String colTypes) {
...
+  public MaterializationValidationResult getAutomaticRewritingValidationResult() {
+    return materializationValidationResult;
   }
 
-  public boolean isValidAutomaticRewritingMaterialization() {
-    return (invalidAutomaticRewritingMaterializationReason == null);
+  public void setAutomaticRewritingValidationResult(
+      MaterializationValidationResult materializationValidationResult) {
+    this.materializationValidationResult =
+        materializationValidationResult;
```
zabetak, line 15942: nit: Since we dropped "Automatic" from the result we can as well drop it from the methods since we are renaming them anyways.

Feel free to disregard or address directly during merge without waiting for a new CI run.

zabetak (approved):
LGTM!

zabetak:
@kasakrisz By the way the test failure is relevant so make sure you address it before merging.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
kasakrisz: Renamed. 
There were test failures in the last CI result due to a deprecated q out file. I have to rerun it anyway.

PR #4348, opened 2023-05-22
HIVE-27355: Iceberg: Create table can be slow due to file listing for…

difin:
… stats

### What changes were proposed in this pull request?

Skipping stats calculation when creating an Iceberg table since Iceberg doesn't use HMS stats.

### Why are the changes needed?

Create table can have huge slowdown due to listing of all files for stats. Stats code isn't needed for creating iceberg tables.

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

Hive-precommit CI passing; New unit tests; Manual testing using MiniHS2Cluster

PR #4954, opened 2023-12-16
HIVE-27957: Better error message for STORED BY

okumin:

This PR would change an error message thrown when a user gives a wrong storage format in STORED BY.

The current messages are less informative. I got confused while I was testing Iceberg and non-Iceberg tables in turns.

```
0: jdbc:hive2://hive-hiveserver2:10000/defaul> CREATE TABLE test (a STRING) STORED BY ORC TBLPROPERTIES ("orc.compress"="ZSTD");
Error: Error while compiling statement: FAILED: SemanticException Cannot find class '' (state=42000,code=40000)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> CREATE TABLE test (a STRING) STORED BY DEFAULT;
Error: Error while compiling statement: FAILED: NullPointerException null (state=42000,code=40000)
```

It changes only error messages.

No.

I added test cases.


ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
```diff
@@ -48,9 +51,14 @@ public class StorageFormat {
 
   public enum StorageHandlerTypes {
     DEFAULT(),
-    ICEBERG("\'org.apache.iceberg.mr.hive.HiveIcebergStorageHandler\'",
+    ICEBERG("'org.apache.iceberg.mr.hive.HiveIcebergStorageHandler'",
         "org.apache.iceberg.mr.hive.HiveIcebergInputFormat", "org.apache.iceberg.mr.hive.HiveIcebergOutputFormat");
 
+    private static final List<StorageHandlerTypes> SUPPORTED_BY_STORED_BY = Arrays
```
akshat0395, line 57: nit: Can we rename this to SUPPORTED_STORED_BY?

ql/src/test/results/clientnegative/create_table_stored_by_invalid3.q.out
```diff
@@ -0,0 +1 @@
+FAILED: SemanticException Cannot find class 'ICEBERG'
```
akshat0395, line 1: Thanks for checking this out, the message looks clean and precise now.
Can you also handle case this ('ICEBERG') to keep the message consistent? We can have something like `Unrecognized storage handler in STORED BY clause: 'ICEBERG'. Supported types = ICEBERG or FQCN of a storage handler.`

This should be an indicator enough for the user to remove the quotes. WDYT

akshat0395 (changes requested):

ql/src/test/results/clientnegative/create_table_stored_by_invalid3.q.out
okumin: Thanks for sharing your thought. Now, I also think your idea sounds more informative. Let me revise it

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
okumin: I started feeling we may not need `StorageHandlerTypes.DEFAULT`. But it seems to be a result of a discussion.
https://github.com/apache/hive/pull/4348#discussion_r1239887908

I'm thinking to rename it into like `NON_DEFAULT` for simplicity.

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
```diff
@@ -157,17 +165,32 @@ public boolean fillStorageFormat(ASTNode child) throws SemanticException {
...
     }
-
-    return ensureClassExists(BaseSemanticAnalyzer.unescapeSQLString(name));
+    final String supportedTypes = StorageHandlerTypes
+        .SUPPORTED_BY_STORED_BY
+        .stream()
+        .map(Enum::toString)
+        .collect(Collectors.joining(", "));
+    throw  new SemanticException(String.format(
+        "Unrecognized storage handler in STORED BY clause: %s. Supported types = %s or FQCN of a storage handler",
```
aturoczy, line 191: I like this initiative because hive error message is too massy, and need a pilot certificate to understand it. So big plus for it!

On the other hand, I would recommend the following message like:
The storage handler specified in the STORED BY clause is not recognized: %s. Please use one of the supported types, which are %s, or provide the Fully Qualified Class Name (FQCN) of a valid storage handler.


Even thinking the second half of the sentence is too technical.

aturoczy (changes requested):
Minor change request

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
okumin: Done
https://github.com/apache/hive/pull/4954/commits/136035e9704ea9cd676d0722b83dbd4ee0310039

ql/src/test/results/clientnegative/create_table_stored_by_invalid3.q.out
okumin: Done
https://github.com/apache/hive/pull/4954/commits/136035e9704ea9cd676d0722b83dbd4ee0310039

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
okumin: I simply followed it as it sounds better. I will update more if anyone has a better idea
https://github.com/apache/hive/pull/4954/commits/136035e9704ea9cd676d0722b83dbd4ee0310039

aturoczy (approved):
+1

akshat0395:
Changes LGTM +1, Thanks

akshat0395 (approved):

aturoczy:
@deniskuzZ @ayushtkn @abstractdog Could you pls check this?

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
```diff
@@ -48,9 +51,14 @@ public class StorageFormat {
 
   public enum StorageHandlerTypes {
     DEFAULT(),
-    ICEBERG("\'org.apache.iceberg.mr.hive.HiveIcebergStorageHandler\'",
```
abstractdog, line 51: wow, what was this "\" used for and why?

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
```diff
@@ -157,17 +165,41 @@ public boolean fillStorageFormat(ASTNode child) throws SemanticException {
...
+        throw createUnsupportedStorageHandlerTypeError(node, e);
+      }
+    }
+    if (node.getType() == HiveParser.Identifier) {
+      // e.g. STORED BY ICEBERG
+      for (StorageHandlerTypes type : StorageHandlerTypes.NON_DEFAULT_TYPES) {
+        if (type.name().equalsIgnoreCase(node.getText())) {
+          inputFormat = type.inputFormat();
+          outputFormat = type.outputFormat();
+          assert type.className != null;
```
abstractdog, line 183: assert is not supposed for production AFAIK, please use explicit null checking here, e.g. [Objects.requireNonNull](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String-)

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
okumin: I'm not sure about the original intention. IntelliJ asks me to remove the backslash and it seems to be safe to remove.

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
okumin: In my opinion, either is fine as this doesn't validate users' input but does assert invariant conditions of enum values for readability. I mean the current implementation doesn't care about the possibility where we add an enum value with null or a wrong class name. What do you think?
As for the direct access, I will fix that.

deniskuzZ:
@abstractdog, could this be merged or some comments are still pending?

abstractdog:
LGTM +1

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
abstractdog: the problem with assert is that it's not meant for production, it can be completely bypassed with JVM args

abstractdog:
> @abstractdog, could this be merged or some comments are still pending?

there are minor comments unaddressed

abstractdog (approved):

abstractdog:
LGTM, pending tests

ql/src/java/org/apache/hadoop/hive/ql/parse/StorageFormat.java
okumin: Thanks. You mean we'd like this validation to be effective at runtime. I replaced the assertion, rebased it based on the latest master, and then confirmed tests succeeded.
https://github.com/apache/hive/pull/4954/commits/df90d42c98f2290ab84455d3a4725fe4f5b0f1a2

PR #4957, opened 2023-12-18
HIVE-25803: URL Mapping appends default Fs scheme even for LOCAL DIRECTORY ops.

ayushtkn:

Use Local FS to qualify a path if LOCAL is used

To push correct qualified path to Ranger when using LOCAL

No

No


deniskuzZ (approved):
LGTM +1, pending tests

PR #4958, opened 2023-12-19
HIVE-27988: Don't convert FullOuterJoin + filter to MapJoin

ngsg:

Do not convert FullOuter Join to MapJoin if it has filter expressions. This partially disables HIVE-18908 optimization, which introduces FullOuterJoin + MapJoin.

As described in HIVE-27226, current Hive produces wrong result when we have FullOuterJoin + MapJoin + filters.

No

No

No test was added as this is just a quick-fix.
One can check the execution of FullOuterJoin + MapJoin + filter by running queries in HIVE-27226 JIRA pages.


ql/src/test/results/clientpositive/llap/mapjoin_filter_on_outerjoin_tez.q.out
```diff
@@ -754,7 +754,7 @@ STAGE PLANS:
     Tez
 #### A masked pattern was here ####
       Edges:
-        Reducer 2 <- Map 1 (CUSTOM_SIMPLE_EDGE), Map 3 (CUSTOM_SIMPLE_EDGE)
+        Reducer 2 <- Map 1 (SIMPLE_EDGE), Map 3 (SIMPLE_EDGE)
```
deniskuzZ, line 757: was this producing incorrect result before?

ql/src/test/results/clientpositive/llap/mapjoin_filter_on_outerjoin_tez.q.out
```diff
@@ -790,26 +790,23 @@ STAGE PLANS:
...
                       Statistics: Num rows: 2 Data size: 24 Basic stats: COMPLETE Column stats: COMPLETE
-                      value expressions: _col1 (type: int), _col2 (type: boolean), (UDFToShort((not _col2)) * 1S) (type: smallint)
+                      value expressions: _col1 (type: int), _col2 (type: boolean)
             Execution mode: llap
             LLAP IO: all inputs
         Reducer 2 
             Execution mode: llap
             Reduce Operator Tree:
-              Map Join Operator
+              Merge Join Operator
```
deniskuzZ, line 799: was `Map Join` producing incorrect results here?

ngsg: MapJoin (+ FOJ + filter) produces wrong result when multiple rows in big tables share the same key but do not satisfy filter.
In this test case, every row has a unique key. So MapJoin produced correct result here.

This test covers all kinds of outer join and FiltertagAppenderProc still affects to other queries.
e.g.
https://github.com/apache/hive/blob/7ff15c9830f1a6988fda9363dc710d75766db889/ql/src/test/results/clientpositive/llap/mapjoin_filter_on_outerjoin_tez.q.out#L336
https://github.com/apache/hive/blob/7ff15c9830f1a6988fda9363dc710d75766db889/ql/src/test/results/clientpositive/llap/mapjoin_filter_on_outerjoin_tez.q.out#L303-L307
UDFToShort(...) is appended by FiltertagAppenderProc and corresponds to filter expression on table 1.
So I think the current test name is still valid.

deniskuzZ: 👍

PR #4959, opened 2023-12-19
HIVE-27775: DirectSQL and JDO results are different when fetching par…

dengzhhu653:
…titions by timestamp in DST shift

On JDO path, we select the matched partitions by the partition name, which is timezone agnostic, however on DirectSQL,
we do a cast partition value to timestamp, then select by the local time, for example:
```
cast("FILTER0"."PART_KEY_VAL" as TIMESTAMP) else null end) as TIMESTAMP) =  <'2023-03-26 03:30:00'>
```
This could lead to different result as described in the Jira.

No

No

mvn test -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile=partition_timestamp3.q -Dtest.metastore.db=postgres|derby|mssql|mysql|oracle


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
```diff
@@ -263,7 +265,9 @@ protected String toDate(String tableValue) {
 
   protected String toTimestamp(String tableValue) {
     if (isORACLE()) {
-      return "TO_TIMESTAMP(" + tableValue + ", 'YYYY-MM-DD HH:mm:ss')";
+      return "TO_TIMESTAMP(" + tableValue + ", 'YYYY-MM-DD HH:mi:ss')";
```
aturoczy, line 268: Timezone information does not need here?

dengzhhu653: I haven't tested against Oracle yet, let me try

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1341,29 +1339,34 @@ public void visit(LeafNode node) throws MetaException {
...
           valType = FilterType.Date;
-        } catch (Exception pe) { // do nothing, handled below - types will mismatch
+          if (dbType.isPOSTGRES() || dbType.isORACLE()) {
+            nodeValue0 = "date '" + nodeValue + "'";
+            nodeValue = null;
+          }
+        } catch (Exception e) {  // do nothing, handled below - types will mismatch
+        }
+      } else if (colType == FilterType.Timestamp) {
+        if (dbType.isDERBY() || dbType.isMYSQL()) {
```
dengzhhu653, line 1356: For Derby or MySQL, we don't push the filter down directly, as the `partition_timestamp3.q` gets
`java.lang.RuntimeException: org.apache.hadoop.hive.ql.parse.SemanticException: MetaException(message:Lists are not the same size: SQL 2, ORM 1)` on Derby and `Different results from SQL and ORM, see log for details` on MySQL.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
dengzhhu653: We don't need it here, I have tested against Oracle19

aturoczy (approved):

aturoczy:
Can we pay attention to this issue pls? It is a blocker for Hive4 release. 

cc: @dengzhhu653 @abstractdog @zabetak @ayushtkn @nrg4878

deniskuzZ:
@dengzhhu653 , has to be rebased

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
```diff
@@ -749,6 +753,22 @@ public Object getBoolean(boolean val) {
     return val;
   }
 
+  public Object convertDateValue(Object dateValue) {
+    assert dateValue instanceof String;
```
deniskuzZ, line 757: assert are ignored in production code

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
```diff
@@ -749,6 +753,22 @@ public Object getBoolean(boolean val) {
     return val;
   }
 
+  public Object convertDateValue(Object dateValue) {
```
deniskuzZ, line 756: that is a weird signature for a method that is supposed to accept a String and return String.
also not clear what kind of conversion it's doing

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
```diff
@@ -749,6 +753,22 @@ public Object getBoolean(boolean val) {
...
+    assert dateValue instanceof String;
+    Date date = MetaStoreUtils.convertStringToDate((String)dateValue);
+    Object result = MetaStoreUtils.convertDateToString(date);
+    return result;
+  }
+
+  public Object convertTimestampValue(Object timestampValue) {
+    assert timestampValue instanceof String;
+    MetaStoreUtils.convertStringToTimestamp((String) timestampValue);
```
deniskuzZ, line 765: what is this? are we mutating the input object?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1422,29 +1420,34 @@ public void visit(LeafNode node) throws MetaException {
         return;
       }
 
+      String nodeValue0 = "?";
```
deniskuzZ, line 1423: what is  `nodeValue0` vs `nodeValue`? can we come up with a better name.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1422,29 +1420,34 @@ public void visit(LeafNode node) throws MetaException {
...
       // if Filter.g does date parsing for quoted strings, we'd need to verify there's no
       // type mismatch when string col is filtered by a string that looks like date.
-      if (colType == FilterType.Date && valType == FilterType.String) {
-        // Filter.g cannot parse a quoted date; try to parse date here too.
+      if (colType == FilterType.Date) {
```
deniskuzZ, line 1426: why have you removed the valType check?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1422,29 +1420,34 @@ public void visit(LeafNode node) throws MetaException {
...
         try {
-          nodeValue = MetaStoreUtils.convertStringToDate((String)nodeValue);
+          nodeValue = dbType.convertDateValue(nodeValue);
           valType = FilterType.Date;
-        } catch (Exception pe) { // do nothing, handled below - types will mismatch
+          if (dbType.isPOSTGRES() || dbType.isORACLE()) {
+            nodeValue0 = "date '" + nodeValue + "'";
```
deniskuzZ, line 1431: shouldn't we hide the db sescifics inside of DatabaseProduct methods?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1422,29 +1420,34 @@ public void visit(LeafNode node) throws MetaException {
...
+            nodeValue = null;
+          }
+        } catch (Exception e) {  // do nothing, handled below - types will mismatch
+        }
+      } else if (colType == FilterType.Timestamp) {
+        if (dbType.isDERBY() || dbType.isMYSQL()) {
```
deniskuzZ, line 1437: same as above, diff db impl should be hidden

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/PartFilterVisitor.java
```diff
@@ -231,25 +229,27 @@ public Long visitIntegerLiteral(PartitionFilterParser.IntegerLiteralContext ctx)
   }
 
   @Override
-  public Date visitDateLiteral(PartitionFilterParser.DateLiteralContext ctx) {
+  public String visitDateLiteral(PartitionFilterParser.DateLiteralContext ctx) {
     PartitionFilterParser.DateContext date = ctx.date();
     String dateValue = unquoteString(date.value.getText());
     try {
-      return MetaStoreUtils.convertStringToDate(dateValue);
+      MetaStoreUtils.convertStringToDate(dateValue);
```
deniskuzZ, line 236: are you mutating the input object?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/PartFilterVisitor.java
```diff
@@ -231,25 +229,27 @@ public Long visitIntegerLiteral(PartitionFilterParser.IntegerLiteralContext ctx)
...
   }
 
   @Override
-  public Timestamp visitTimestampLiteral(PartitionFilterParser.TimestampLiteralContext ctx) {
+  public String visitTimestampLiteral(PartitionFilterParser.TimestampLiteralContext ctx) {
     PartitionFilterParser.TimestampContext timestamp = ctx.timestamp();
     String timestampValue = unquoteString(timestamp.value.getText());
     try {
-      return MetaStoreUtils.convertStringToTimestamp(timestampValue);
+      MetaStoreUtils.convertStringToTimestamp(timestampValue);
```
deniskuzZ, line 248: same as above

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/ExpressionTree.java
```diff
@@ -263,26 +235,124 @@ public String toString() {
...
+    /**
+     * Get partition column index in the table partition column list that
+     * corresponds to the key that is being filtered on by this tree node.
+     * @param partitionKeys list of partition keys.
+     * @param filterBuilder filter builder used to report error, if any.
+     * @return The index.
+     */
+    public static int getPartColIndexForFilter(String partitionKeyName,
+        List<FieldSchema> partitionKeys, FilterBuilder filterBuilder) throws MetaException {
+      assert (partitionKeys.size() > 0);
```
deniskuzZ, line 268: asserts are not executed. in production code

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/ExpressionTree.java
```diff
@@ -263,26 +235,124 @@ public String toString() {
...
+      int partitionColumnIndex;
```
deniskuzZ, line 269:
````
int partitionColumnIndex = Iterables.indexOf(partitionKeys, key -> partitionKeyName.equalsIgnoreCase(key.getName()))
````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/ExpressionTree.java [hunk 14]
```diff
@@ -263,26 +235,124 @@ public String toString() {
...
+
+    @Override
+    protected void endTreeNode(TreeNode node) throws MetaException {
+      filterBuilder.append(") ");
+    }
+
+    @Override
+    protected void visit(LeafNode node) throws MetaException {
+      beforeParsing();
+      keyName = node.keyName;
```
deniskuzZ, line 340: are you using `keyName`, `operator`, `value`, `isReverseOrder` somewhere?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/ExpressionTree.java
```diff
@@ -263,26 +235,124 @@ public String toString() {
...
+    public JDOFilterGenerator(Configuration conf, List<FieldSchema> partitionKeys,
+        FilterBuilder filterBuilder, Map<String, Object> params) {
+      this.conf = conf;
+      this.partitionKeys = partitionKeys;
+      this.filterBuilder = filterBuilder;
+      this.params = params;
+    }
+
+    private void beforeParsing() throws MetaException {
+      if (!onParsing && !filterBuilder.getFilter().isEmpty()) {
```
deniskuzZ, line 315: do we need `onParsing` , isn't it enough to check the `filter`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
dengzhhu653: exactly, this is for verifying the input in the unit tests

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
dengzhhu653: To check the literal `timestampValue` is a valid timestamp

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
dengzhhu653: As the valType is always a string in case of `colType == FilterType.Date`, so there is no need to add `valType` check

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
dengzhhu653: this could be a little hack, as we need to alter the method variables `nodeValue0` and `nodeValue` depending on the database.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/ExpressionTree.java
```diff
@@ -263,26 +235,124 @@ public String toString() {
...
+    }
+
+    @Override
+    protected void beginTreeNode(TreeNode node) throws MetaException {
+      beforeParsing();
+      filterBuilder.append("( ");
+    }
+
+    @Override
+    protected void midTreeNode(TreeNode node) throws MetaException {
```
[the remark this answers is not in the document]
dengzhhu653: this visitor is for generating the JDO filter, `and` or `or` may not meet in JDO path

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/PartFilterVisitor.java
dengzhhu653: `MetaStoreUtils.convertStringToDate(dateValue)` converts to the date with time zone, it causes trouble when comparing the value both on direct sql or jdo, and as far as I know, there is no way to convert it back to original string, I would prefer to push the convention down into the `TreeVisitor`(i.e, the `JDOFilterGenerator`), converting it to date type if necessary.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
deniskuzZ: 1. why don't we rename it to `normalizeDate` or something
2. why input and output types are Object and not String
3. why this util method is in DatabaseProduct class (not MetaStoreUtils), has nothing to do with it

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
deniskuzZ: same as above, it's not a convert method - `isValidTimestamp(String timestampValue)`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/PartFilterVisitor.java
deniskuzZ: why do we need to validate then date here?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/ExpressionTree.java
deniskuzZ: won't `!filterBuilder.getFilter().isEmpty()` cover that?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1505,14 +1507,9 @@ public void visit(LeafNode node) throws MetaException {
...
-
-        if (valType == FilterType.Date) {
-        	tableValue = dbType.toDate(tableValue);
-        } else if (valType == FilterType.Timestamp) {
-          tableValue = dbType.toTimestamp(tableValue);
-        }
       }
-      if (!node.isReverseOrder) {
+
+      if (!node.isReverseOrder && nodeValue != null) {
```
deniskuzZ, line 1512: why do we need 2 checks, can we change to
````
if (nodeValue != null) { 
    params.add(nodeValue);
}
````

**Comment by deniskuzZ** [comment]
@dengzhhu653, any chance of getting this PR ready for merge? looks like that is the only remaining blocker for 4.0 release

**Comment by dengzhhu653** [comment]
> @dengzhhu653, any chance of getting this PR ready for merge? looks like that is the only remaining blocker for 4.0 release

I was engaged into other things these days, will back to it in next two days.

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/ExpressionTree.java (hunk 15 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/ExpressionTree.java:315]: The `filterBuilder` could not be empty when begin parsing, e.g, https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L4905-L4917

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java (hunk 17 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java:1512]: it's used to avoid adding the same parameter twice to the statement, 
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L1482
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L1519

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java (hunk 6 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java:1423]: The `nodeValue0` is placeholder, `nodeValue` is designated parameter for the PrepareStatement, we've already used it elsewhere in this method...

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/PartFilterVisitor.java (hunk 10 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/parser/PartFilterVisitor.java:236]: We can avoid passing the invalid date down into the query filter generator given the context here is a date, and fail from the error quickly.

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java [hunk 18]
```diff
@@ -1424,30 +1422,36 @@ public void visit(LeafNode node) throws MetaException {
...
-        } catch (Exception pe) { // do nothing, handled below - types will mismatch
+          if (dbType.isPOSTGRES() || dbType.isORACLE()) {
+            nodeValue0 = "date '" + nodeValue + "'";
+            nodeValue = null;
+          }
+        } catch (Exception e) {  // do nothing, handled below - types will mismatch
+        }
+      } else if (colType == FilterType.Timestamp) {
+        if (dbType.isDERBY() || dbType.isMYSQL()) {
+          filterBuffer.setError("Filter pushdown on timestamp not supported for " + dbType.dbType.name());
```
deniskuzZ, line 1442: should we move this to constants?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
deniskuzZ: I mean, instead of
````
if (node.isReverseOrder && nodeValue != null) {
        params.add(nodeValue);
}
....
if (!node.isReverseOrder && nodeValue != null) {
     params.put("catName", catName);
 }
````
can't we refactor to
````
if (nodeValue != null) { 
    params.add(nodeValue); 
}
````

**Review by deniskuzZ — COMMENTED** [review]
LGTM, just 2 minor comments 
@dengzhhu653, could you please take a look at new sonar issues if those are relevant.

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java [hunk 19]
```diff
@@ -1424,30 +1422,36 @@ public void visit(LeafNode node) throws MetaException {
...
+      String nodeValue0 = "?";
       // if Filter.g does date parsing for quoted strings, we'd need to verify there's no
       // type mismatch when string col is filtered by a string that looks like date.
-      if (colType == FilterType.Date && valType == FilterType.String) {
-        // Filter.g cannot parse a quoted date; try to parse date here too.
+      if (colType == FilterType.Date) {
         try {
-          nodeValue = MetaStoreUtils.convertStringToDate((String)nodeValue);
+          // check the nodeValue is a valid date
+          nodeValue = MetaStoreUtils.convertDateToString(
```
deniskuzZ, line 1431: missed that part, why don't we do similar to PartFilterVisitor.java
````
MetaStoreUtils.convertStringToDate((String) nodeValue);
````
and use nodeValue if no exceptions;
if that does some date with TZ string reformatting, can we create a helper method in MetaStoreUtils

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java (hunk 17 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java:1516]: No.   As in the result filter, there could be some "?" in the `tableValue`,
```
String filter = node.isReverseOrder
              ? nodeValue0 + " " + node.operator.getSqlOp() + " " + tableValue
              : tableValue + " " + node.operator.getSqlOp() + " " + nodeValue0;
```
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L1502-L1508

We need to place the `nodeValue` at the right position in `params` according to the `isReverseOrder`

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java (hunk 18 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java:1442]: This string constant will be referenced only once, I would like to keep it as local variable.

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java (hunk 19 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java:1431]: if we return the `Date` value from `PartFilterVisitor`, we couldn't convert the date back to the original string in JDO path(we compare the `PART_NAME` on JDO path, which is timezone agnostic), this could result to query difference   as we met in this jira.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java:1429]: PartFilterVisitor returns String date representation after validation, isn't it?

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java (hunk 17 above)
Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java:1513]: 👍

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java (hunk 19 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java:1429]: This is for converting the string date to a timezone based string date, we need it on the direct sql:
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java#L263-L269

**Review by deniskuzZ — APPROVED** [review]
LGTM +1. Thanks for the PR and addressing the review comments, @dengzhhu653!

File: standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java [hunk 20]
```diff
@@ -107,6 +107,17 @@ public static Date convertStringToDate(String date) {
...
   }
 
+  /**
+   * Converts the string format date without a time-zone to
+   * a time-zone based string format date
+   * @param date the date without a time-zone
+   * @return time-zone based string format date
+   */
+  public static String normalizeDate(String date) {
+    Date val = MetaStoreUtils.convertStringToDate(date);
```
deniskuzZ, line 117: minor, up to you:
````
return convertDateToString(convertStringToDate(date));
````
also we could reuse the new method in `BaseSemanticAnalyzer#normalizeDateCol`

dengzhhu653: done

PR #4960, opened 2023-12-19
HIVE-27961: Beeline will print duplicate stats info when hive.tez.exec.print.summary is true

zhangbutao:

Remove the duplicate stats info from beeline console.

if `hive.tez.exec.print.summary` is true, beeline console will print duplicate stats info.  Please see HIVE-27961 to check the info.
This stats info is displayed by default([BasicStatsTask.java L315](https://github.com/apache/hive/blob/4a057a7cea11029dbda7adaec09117d46e56d925/ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java#L315)), and no need to be displayed again when `hive.tez.exec.print.summary` is true.
https://github.com/apache/hive/blob/4a057a7cea11029dbda7adaec09117d46e56d925/ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java#L312-L316

No

No

No need test

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
@@ -309,7 +309,11 @@ public class BasicStatsTask implements Serializable, IStatsProcessor {
         TransactionalStatsProcessor transactionalStatsProcessor = new TransactionalStatsProcessor(db, p);
         transactionalStatsProcessor.process(statsAggregator);
 
-        console.printInfo("Table " + tableFullName + " stats: [" + toString(p.getPartParameters()) + ']');
+        if (conf.getBoolVar(ConfVars.TEZ_EXEC_SUMMARY)) {
+          console.printInfo("Table " + tableFullName + " stats: [" + toString(p.getPartParameters()) + ']');
+        } else {
+          LOG.info("Table " + tableFullName + " stats: [" + toString(p.getPartParameters()) + ']');
+        }
 
         // The table object is assigned to the latest table object.
         // So that it can be used by ColStatsProcessor.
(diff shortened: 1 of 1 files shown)


aturoczy (approved):
+1 Easy change.
@zhangbutao Please move the jira ticket into the correct state.

@SourabhBadhya @ayushtkn @deniskuzZ Could you pls quickly check?

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
```diff
@@ -309,9 +309,6 @@ private int aggregateStats(Hive db, Table tbl) {
         TransactionalStatsProcessor transactionalStatsProcessor = new TransactionalStatsProcessor(db, p);
         transactionalStatsProcessor.process(statsAggregator);
 
-        if (conf.getBoolVar(ConfVars.TEZ_EXEC_SUMMARY)) {
```
SourabhBadhya, line 312: Isnt something similar happening on the partition level as well?
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java#L378-L381

Also `console` variable seems to wrap the logger (LOG) variable and essentially doing things that can be achieved via the logger object directly. Maybe remove the console variable?

aturoczy: Ohh this is an excellent catch @SourabhBadhya

zhangbutao: Definitely a good catch!
I also take a little time to check `LOG.info` and `console.printInfo`, and it seems that the` else if `is more reasonable.
Thanks.

SourabhBadhya (approved):
LGTM +1 (pending tests)

PR #4961, opened 2023-12-19
HIVE-26505: Case When Some result data is lost when there are common column conditions and partitioned column conditions

kasakrisz:
```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=ppr_conjunct_isnotfalse.q -pl itests/qtest -Pitests
```


deniskuzZ:
@kasakrisz do we need this in 4.1.0? if yes, please rebase and ping someone in hive dev

PR #4962, opened 2023-12-19
HIVE-27962: Iceberg: Runtime predicate pushdown doesn't work

deniskuzZ:

BugFix

Runtime predicate pushdown doesn't work after HIVE-27273

No

No

On a cluster with 1Tb dataset

### DIFF
--- a/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergFilterFactory.java
+++ b/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergFilterFactory.java
@@ -24,9 +24,6 @@ import java.sql.Date;
 import java.sql.Timestamp;
 import java.time.LocalDate;
 import java.time.ZoneOffset;
-import java.util.Collections;
-import java.util.List;
-import org.apache.hadoop.hive.ql.io.sarg.ExpressionTree;
 import org.apache.hadoop.hive.ql.io.sarg.PredicateLeaf;
 import org.apache.hadoop.hive.ql.io.sarg.SearchArgument;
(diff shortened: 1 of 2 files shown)


aturoczy (commented):
So, if the expression has a **BETWEEN** statement and leaf is less than 2 then returns true?
But **BETWEEN** should always contains at least 2 literals. Or maybe I do not understand correctly. 

Do you have a simple query for this?

aturoczy (approved):
+1

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergFilterFactory.java
```diff
@@ -138,21 +138,6 @@ public void testBetweenOperand() {
     assertEquals(actual.right().op(), expected.right().op());
   }
 
-  @Test
-  public void testUnsupportedBetweenOperandEmptyLeaves() {
-    SearchArgument.Builder builder = SearchArgumentFactory.newBuilder();
-    final SearchArgument arg =
-        new MockSearchArgument(
```
zhangbutao, line 145: Maybe this mock class also need to be removed.
https://github.com/apache/hive/blob/36ce858163a19e29eafe4a8d3307191bc28fc175/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergFilterFactory.java#L282

deniskuzZ: removed

zhangbutao (approved):
LGTM

PR #4965, opened 2023-12-20
HIVE-27492: HPL/SQL built-in functions like sysdate not working

mdayakar:
Currently the HPLSQL built-in functions like sysdate are not working, only the HPLSQL builtin functions which are also available in Hive are working. So in this PR changed the code in such away that other built-in functions specific to HPLSQL like sysdate also will work. While parsing the HPLSQL statements if found any HPLSQL builtin functions then those will be executed and the result will be added to the final query, if the function is also available in Hive then just forwarding it to Hive to parse and execute the query.

To support all the HPLSQL built-in functions like sysdate will work.

No

No

Added unit test cases.


hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionString.java
```diff
@@ -33,38 +33,30 @@ public FunctionString(Exec e, QueryExecutor queryExecutor) {
...
-    f.map.put("LENGTH", this::length);
-    f.map.put("LOWER", this::lower);
-    f.map.put("REPLACE", this::replace);
-    f.map.put("SUBSTR", this::substr);
-    f.map.put("SUBSTRING", this::substr);
     f.map.put("TO_CHAR", this::toChar);
     f.map.put("UPPER", this::upper);
-    
-    f.specMap.put("SUBSTRING", this::substring);
-    f.specMap.put("TRIM", this::trim);
```
aturoczy, line 47: As I see from the test this is required. Please check the test. Put this back or change the test accordingly.

aturoczy (changes requested):

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionString.java
mdayakar: @aturoczy Thanks for the review. Test cases are already added for SUBSTRING and TRIM, also for all functions.
https://github.com/apache/hive/pull/4965/commits/7c0a9f31477453d9974472f487dde4bfff577989#diff-4b45b06ec5de20b8929a80f493ead0990f1208e725d9a403774ecc5f27928227R798-R807
https://github.com/apache/hive/pull/4965/commits/7c0a9f31477453d9974472f487dde4bfff577989#diff-4b45b06ec5de20b8929a80f493ead0990f1208e725d9a403774ecc5f27928227R849-R852

amansinha100: @mdayakar  I suppose the reason for removing SUBSTRING, TRIM and a few other functions is that these are builtin functions in Hive, so is HPL/SQL going to forward the function call to HS2 ? If so,  it would be good to add comments in some common class regarding this.   One thing we should verify is if the HPL/SQL behavior of these functions is the same as what the Hive built-in function does or is there any small difference.  I am curious why these were implemented in HPL/SQL in the first place since these are SQL standard functions.  Was there additional precondition checks being done or any other variants that HPL/SQL supported that Hive did not ?

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionDatetime.java
```diff
@@ -166,30 +161,6 @@ void toTimestamp(HplsqlParser.Expr_func_paramsContext ctx) {
...
-    if (cnt == 0) {
-      evalNull();
-      return;
-    }
-    long epoch = evalPop(ctx.func_param(0).expr()).longValue();
-    String format = "yyyy-MM-dd HH:mm:ss";
-    if (cnt > 1) {
-      format = evalPop(ctx.func_param(1).expr()).toString();
-    }
-    evalString(new SimpleDateFormat(format).format(new Date(epoch * 1000)));
```
amansinha100, line 184: Previously, this was using the SimpleDateFormat.  Based on this change, since this function will be forwarded to Hive's built-in function, that one is now defaulting to DateTimeFormatter.  We would want to document the change in behavior for the HPL/SQL function.

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
```diff
@@ -610,12 +612,268 @@ public void testNullCast() throws Throwable {
...
+  }
+
+  @Test
+  public void testCHARHplSqlFunction() throws Throwable {
+    String SCRIPT_TEXT = "select CHAR(2023)";
+    testScriptFile(SCRIPT_TEXT, args(), "2023");
+  }
+
+  @Test
+  public void testCOALESCEHplSQLFunction() throws Throwable {
```
amansinha100, line 646: I notice this test is not the same as what coalesce.sql was covering.  That test file's contents have been removed, so don't we want those tests to carry over here ?  Are we going to lose any test coverage ?
The same applies to some of the other functions such as SUBSTRING, TRIM, FROM_UNIXTIME etc.

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionString.java
mdayakar: > I suppose the reason for removing SUBSTRING, TRIM and a few other functions is that these are builtin functions in Hive.
Yes Aman.

> so is HPL/SQL going to forward the function call to HS2 ?
Yes, in this case it will just forward the function call to HS2

> One thing we should verify is if the HPL/SQL behavior of these functions is the same as what the Hive built-in function does or is there any small difference.
Yes Aman, if there is any behaviour change or some extra feature for any function then that code is not removed. Actually for SUBSTRING also later found that it supports another functionality 'SUBSTRING(string FROM start_pos [FOR substring_len])' which is not present in HS2 so added back the SUBSTRING function code. Similar for CONCAT (if any param is NULL then HS2 returns NULL as output where as HPLSQL function ignores NULLs and concats other non-null strings) so this function also not removed.

>  I am curious why these were implemented in HPL/SQL in the first place since these are SQL standard functions.
As per my understanding HPLSQL initially provided as a separate entity (not part of Hive) and later it got merged as a part of Hive. When it was a separate entity, it can be used not only with Hive also can be with any other Databases like DB2, MySQL,... so they might provided implementation for all these functions now its part of Hive release so I assumed it can be used only with Hive so I removed the functions implementation which are already supported in Hive.

> Was there additional precondition checks being done or any other variants that HPL/SQL supported that Hive did not ?
The code is already present and I could see there are some precondition checks present.

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
mdayakar: COALESCE function returns first non-null value in the input list. coalesce.sql was covering scenarios like no null value, first param null with only second non-null param, first param null with two non-null params and two null params with one non-null param. All these scenarios can be added to TestHplSqlViaBeeLine.java also, will try to add them also.

kasakrisz (approved):
minor comments only

hplsql/src/main/java/org/apache/hive/hplsql/Var.java
```diff
@@ -601,7 +601,11 @@ else if (type == Type.STRING) {
       return (String)value;
     }
     else if (type == Type.DATE) {
-      return ((Date)value).toString();
+      String date = ((Date)value).toString();
+      StringBuilder dateSB = new StringBuilder("DATE '");
+      dateSB.append(date);
+      dateSB.append("'");
+      return dateSB.toString();
```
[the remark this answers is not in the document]
mdayakar: Fixed

hplsql/src/main/java/org/apache/hive/hplsql/Var.java
```diff
@@ -612,7 +616,10 @@ else if (type == Type.TIMESTAMP) {
       if (t.length() > len) {
         t = t.substring(0, len);
       }
-      return t;
+      StringBuilder tsSB = new StringBuilder("TIMESTAMP '");
+      tsSB.append(t);
+      tsSB.append("'");
+      return tsSB.toString();
```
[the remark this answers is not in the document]
mdayakar: Fixed

PR #4968, opened 2023-12-22
HIVE-27406: Addendum: Query runtime optimization

deniskuzZ:
### What changes were proposed in this pull request?

SQL query runtime improvement

### Why are the changes needed?

Performance optimization

### Does this PR introduce _any_ user-facing change?

No

### Is the change a dependency upgrade?

No

### How was this patch tested?

mvn test -Dtest=ITestDbTxnManager#testRemoveDuplicateCompletedTxnComponents -Dtest.metastore.db=mysql;
mvn test -Dtest=ITestDbTxnManager#testRemoveDuplicateCompletedTxnComponents -Dtest.metastore.db=postgres;
mvn test -Dtest=ITestDbTxnManager#testRemoveDuplicateCompletedTxnComponents -Dtest.metastore.db=oracle;

PR #4970, opened 2023-12-22
HIVE-27022 - Split removeDuplicateCompletedTxnComponents away from Ac…

tarak271:
…idHouseKeeper and onto a separate timer

In busy clusters removeDuplicateCompletedTxnComponents seems to be bottleneck in backend database. So splitting compaction related house keeping tasks into another house keeping service, so that we can adjust interval for this alone to avoid calling removeDuplicateCompletedTxnComponents too many times in a short period of time

To facilitate the ability to change frequency at which house keeping to be done on compaction related tables

No

No

Junit test cases


deniskuzZ:
-1
why do we need one more housekeeping service? tune params of existing service if required

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -97,6 +97,9 @@ public class MetastoreConf {
   static final String ACID_HOUSE_KEEPER_SERVICE_CLASS =
       "org.apache.hadoop.hive.metastore.txn.AcidHouseKeeperService";
   @VisibleForTesting
+  static final String ACID_COMPACTION_HOUSE_KEEPER_SERVICE_CLASS =
```
deniskuzZ, line 100: simply `COMPACTION_HOUSE_KEEPER_SERVICE_CLASS`

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -293,6 +296,9 @@ public enum ConfVars {
     ACID_HOUSEKEEPER_SERVICE_INTERVAL("metastore.acid.housekeeper.interval",
         "hive.metastore.acid.housekeeper.interval", 60, TimeUnit.SECONDS,
         "Time interval describing how often the acid housekeeper runs."),
+    ACID_COMPACTION_HOUSEKEEPER_SERVICE_INTERVAL("metastore.acid.compaction.housekeeper.interval",
```
deniskuzZ, line 299: simply `COMPACTION_HOUSEKEEPER_SERVICE_INTERVAL`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/leader/HouseKeepingTasks.java
```diff
@@ -61,6 +62,11 @@ public List<MetastoreTaskThread> getRemoteOnlyTasks() throws Exception {
     Collection<String> taskNames =
         MetastoreConf.getStringCollection(configuration, MetastoreConf.ConfVars.TASK_THREADS_REMOTE_ONLY);
     for (String taskName : taskNames) {
+      if (AcidCompactionHouseKeeperService.class.getName().equals(taskName) &&
```
deniskuzZ, line 65: move this logic to the service ifself

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/AcidCompactionHouseKeeperService.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Performs background tasks for Transaction management in Hive.
+ * Runs inside Hive Metastore Service.
+ */
+public class AcidCompactionHouseKeeperService extends AcidHouseKeeperService {
```
deniskuzZ, line 29: 1. rename to `CompactionHouseKeeperService`
2. move all services under `txn/service`package

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
tarak271: Done

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
tarak271: Done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/leader/HouseKeepingTasks.java
tarak271: Moved logic to service, my intention to keep it here is to avoid creating service itself when initiator is not enabled

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/AcidCompactionHouseKeeperService.java
tarak271: done

InvisibleProgrammer:
Hi, please let me putting some thoughts about the change here: 

Firstly, I like the idea of decoupling the AcidHouseKeeperService. As I see, it already does five different things so that change is one small steps in a direction of one object doing one thing:
```java
    performTask(txnHandler::performTimeOuts, "Cleaning timed out txns and locks");
    performTask(txnHandler::performWriteSetGC, "Cleaning obsolete write set entries");
    performTask(txnHandler::cleanTxnToWriteIdTable, "Cleaning obsolete TXN_TO_WRITE_ID entries");
    if (isCompactorEnabled) {
      performTask(txnHandler::removeDuplicateCompletedTxnComponents, "Cleaning duplicate COMPLETED_TXN_COMPONENTS entries");
      performTask(txnHandler::purgeCompactionHistory, "Cleaning obsolete compaction history entries");
    }
```

However, I have some second thoughts about how the performance issue is solved: the issue is that there is a clean up in a database that takes too much time to run. The solution: increase the time window. The thing that concerns me about that is it more putting the problem away stuff. Currently, we have run the clean up in every minute. After that change, the duplicated txn components clean up will run in every 5 minutes. And it will be the default. 

That can lead to larger intervals of table locks on the table that we use. What if it lead to deadlocks? At relational databases, the best practise is to keep our transactions short. 

What about considering an other approach? What if instead of running the clean up in one, large transaction, we try to run multiple small ones? 
I don't know what kind of database the customer has but for example, at MSSQL we did measures back then and found that when we want to delete large amount of records, it is way faster in batches under 5000 element (usually we used 4000). The reason why it was faster was the locking mechanism of the database: for large amount of records, it put exclusive locks and you cannot use the table in other processes and it can cause performance issues. 

The clean up in that case is a little bit complicated: it can be slow because of the time to take to delete the records or it can be the time to collect the records that we want to delete. 

For those kind of scenarios I would recommend to have two parameters for the clean up: 
- Batch size
- Number of iterations
And I would still keep a 1 minute interval as default. 

So that, it can be easily to fine tune the parameters for the customers: if there are too many records to delete, just increase the number of iterations. If it takes too large of time to collect what to delete, increase the time window and/or the batch size. 

What do you think about that?

tarak271:
@InvisibleProgrammer 
Thanks a lot for your valuable inputs.

1. This pull request is aimed at separating the compaction housekeeping tasks into a new service
2.  By doing this we can have a temporary workaround to run compaction related housekeeping tasks less frequently since we have the option to adjust the frequency
3. Coming to performance issue, tuning of the delete query used in compaction housekeeping is already done at https://github.com/apache/hive/pull/4968 which needs to be merged
4. Related to changing the way we do cleaning can be done as part of another apache jira created now https://issues.apache.org/jira/browse/HIVE-27976

InvisibleProgrammer:
@tarak271 , thank you for your reply. This looks great. 

I have only one question left: I assume the root cause of a change is a customer escalation. And having larger time window can solve their issue. Are you sure that we want to change the default time window? 
Or we have a recurring problem that affects multiple customers so it makes sense to change the default? 

Thank you

tarak271:
@InvisibleProgrammer Thanks for the reply

> we have a recurring problem that affects multiple customers so it makes sense to change the default?

Yes, that is the reason why we want to decouple compactor housekeeper, so that frequency can be changed without affecting rest of the housekeeping tasks

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -38,16 +40,22 @@ public class AcidHouseKeeperService implements MetastoreTaskThread {
   private static final Logger LOG = LoggerFactory.getLogger(AcidHouseKeeperService.class);
 
   private Configuration conf;
-  private boolean isCompactorEnabled;
   private TxnStore txnHandler;
+  private String serviceName;
+
+  public void setServiceName(String serviceName) {
```
deniskuzZ, line 46: why do we need to expose `setServiceName` & `getTxnHandler`?

tarak271: To get the service name printed in log line in run method as we are not overriding it in CompactionHouseKeeper.java
   ```
- LOG.info("Starting to run AcidHouseKeeperService.");
   + LOG.info("Starting to run {}", serviceName);
```

And reuse TxnHandler created already in AcidHouseKeeperService as we are overriding cleanTheHouse method

 ```
void cleanTheHouse() {
    if (isCompactorEnabled) {
      performTask(getTxnHandler()::removeDuplicateCompletedTxnComponents,
          "Cleaning duplicate COMPLETED_TXN_COMPONENTS entries");
      performTask(getTxnHandler()::purgeCompactionHistory, "Cleaning obsolete compaction history entries");
    }
  }
```

Please guide me if CompactionHouseKeeper needs to be completely decoupled with AcidHouseKeeperService code wise by not implementing inheritance

tarak271:
@deniskuzZ could you please help guide on the next steps with this pull request

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
deniskuzZ: yes, it should be an independent MetastoreTaskThread service

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -78,17 +86,13 @@ public void run() {
     }
   }
 
-  private void cleanTheHouse() {
+  void cleanTheHouse() {
```
deniskuzZ, line 89: should have `protected` modifier

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -78,17 +86,13 @@ public void run() {
...
     performTask(txnHandler::performWriteSetGC, "Cleaning obsolete write set entries");
     performTask(txnHandler::cleanTxnToWriteIdTable, "Cleaning obsolete TXN_TO_WRITE_ID entries");
-    if (isCompactorEnabled) {
-      performTask(txnHandler::removeDuplicateCompletedTxnComponents, "Cleaning duplicate COMPLETED_TXN_COMPONENTS entries");
-      performTask(txnHandler::purgeCompactionHistory, "Cleaning obsolete compaction history entries");
-    }
   }
 
-  private void performTask(FailableRunnable<MetaException> task, String description) {
+  void performTask(FailableRunnable<MetaException> task, String description) {
```
deniskuzZ, line 95: why have you removed private accessor?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/CompactionHouseKeeperService.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public long runFrequency(TimeUnit unit) {
+    return MetastoreConf.getTimeVar(getConf(), MetastoreConf.ConfVars.COMPACTION_HOUSEKEEPER_SERVICE_INTERVAL,
+        unit);
+  }
+
+  @Override
+  void cleanTheHouse() {
+    if (isCompactorEnabled) {
```
deniskuzZ, line 48: it shouldn't even be registered if `isCompactorEnabled` false

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: was it committed?
would be nice if we rename to
````
ACID_HOUSEKEEPER_SERVICE_CLASS - for aic txn
COMPACTION_HOUSEKEEPER_SERVICE_CLASS - for compaction
````

File: ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java [hunk 9]
```diff
@@ -1244,9 +1245,16 @@ void testInitiatorWithMultipleFailedCompactionsForVariousTblProperties(String tb
 
     MetastoreConf.setTimeVar(hiveConf, MetastoreConf.ConfVars.ACID_HOUSEKEEPER_SERVICE_INTERVAL, 10,
         TimeUnit.MILLISECONDS);
+    MetastoreConf.setTimeVar(hiveConf, MetastoreConf.ConfVars.COMPACTION_HOUSEKEEPER_SERVICE_INTERVAL, 10,
+        TimeUnit.MILLISECONDS);
+    MetastoreConf.setBoolVar(hiveConf, MetastoreConf.ConfVars.COMPACTOR_INITIATOR_ON, true);
+    MetastoreConf.setBoolVar(hiveConf, MetastoreConf.ConfVars.COMPACTOR_CLEANER_ON, true);
```
deniskuzZ, line 1251: no need for Cleaner & Initiator configs

ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
```diff
@@ -3508,9 +3509,15 @@ public void testRemoveDuplicateCompletedTxnComponents() throws Exception {
     Assert.assertEquals(TestTxnDbUtil.queryToString(conf, "select * from \"COMPLETED_TXN_COMPONENTS\""),
       5, TestTxnDbUtil.countQueryAgent(conf, "select count(*) from \"COMPLETED_TXN_COMPONENTS\""));
 
+    MetastoreConf.setBoolVar(conf, MetastoreConf.ConfVars.COMPACTOR_INITIATOR_ON, true);
```
deniskuzZ, line 3512: same here

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/CompactionHouseKeeperService.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Override
+  public long runFrequency(TimeUnit unit) {
+    return MetastoreConf.getTimeVar(getConf(), MetastoreConf.ConfVars.COMPACTION_HOUSEKEEPER_SERVICE_INTERVAL,
+        unit);
+  }
+
+  @Override
+  public void cleanTheHouse() {
```
deniskuzZ, line 42: protected

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -95,13 +95,16 @@ public class MetastoreConf {
       "org.apache.hadoop.hive.metastore.security.MetastoreDelegationTokenManager";
   @VisibleForTesting
   static final String ACID_HOUSE_KEEPER_SERVICE_CLASS =
-      "org.apache.hadoop.hive.metastore.txn.AcidHouseKeeperService";
+      "org.apache.hadoop.hive.metastore.txn.service.AcidHouseKeeperService";
+  @VisibleForTesting
+  static final String COMPACTION_HOUSE_KEEPER_SERVICE_CLASS =
```
deniskuzZ, line 100: HOUSE_KEEPER -> HOUSEKEEPER, same for ACID

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -78,17 +81,13 @@ public void run() {
...
     long start = System.currentTimeMillis();
     Functions.run(task);
```
deniskuzZ, line 92:
````
+import org.apache.commons.lang3.function.Failable;
+import org.apache.commons.lang3.function.FailableRunnable;
-import org.apache.commons.lang3.Functions;
replace Functions.run(task) with Failable.run(task)
````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java [hunk 14]
```diff
@@ -78,17 +81,13 @@ public void run() {
     }
   }
 
-  private void cleanTheHouse() {
+  protected void cleanTheHouse() {
     performTask(txnHandler::performTimeOuts, "Cleaning timed out txns and locks");
```
deniskuzZ, line 85: could we refactor to
    protected final Map<FailableRunnable<MetaException>, String> tasks;
    in constructor:
    tasks = ImmutableMap.<FailableRunnable<MetaException>, String>builder()
      .put(txnHandler::performTimeOuts, "Cleaning timed out txns and locks")
      .put(txnHandler::cleanTxnToWriteIdTable, "Cleaning obsolete TXN_TO_WRITE_ID entries")
      .build();
    
tasks.forEach(this::performTask);

deniskuzZ: corresponding change in Compaction service

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
tarak271: changing to protected, so that it can be accessed in sub class

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/CompactionHouseKeeperService.java
tarak271: Done

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
tarak271: Done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
tarak271: Done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
tarak271: We cannot initialize tasks in constructor as we need txnHandler which is only initialized after setConf method is called. So made tasks as non final and initializing tasks in a new method

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -38,16 +42,27 @@ public class AcidHouseKeeperService implements MetastoreTaskThread {
   private static final Logger LOG = LoggerFactory.getLogger(AcidHouseKeeperService.class);
 
   private Configuration conf;
-  private boolean isCompactorEnabled;
-  private TxnStore txnHandler;
+  protected TxnStore txnHandler;
```
deniskuzZ, line 45: could be private

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -78,19 +93,13 @@ public void run() {
...
 
-  private void cleanTheHouse() {
-    performTask(txnHandler::performTimeOuts, "Cleaning timed out txns and locks");
-    performTask(txnHandler::performWriteSetGC, "Cleaning obsolete write set entries");
-    performTask(txnHandler::cleanTxnToWriteIdTable, "Cleaning obsolete TXN_TO_WRITE_ID entries");
-    if (isCompactorEnabled) {
-      performTask(txnHandler::removeDuplicateCompletedTxnComponents, "Cleaning duplicate COMPLETED_TXN_COMPONENTS entries");
-      performTask(txnHandler::purgeCompactionHistory, "Cleaning obsolete compaction history entries");
-    }
+  protected void cleanTheHouse() {
```
deniskuzZ, line 96: private

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -78,19 +93,13 @@ public void run() {
...
+    tasks.forEach(this::performTask);
   }
 
-  private void performTask(FailableRunnable<MetaException> task, String description) {
+  protected void performTask(FailableRunnable<MetaException> task, String description) {
```
deniskuzZ, line 100: private

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/CompactionHouseKeeperService.java
```diff
@@ -0,0 +1,53 @@
...
+ * Performs background tasks for Transaction management in Hive.
+ * Runs inside Hive Metastore Service.
+ */
+public class CompactionHouseKeeperService extends AcidHouseKeeperService {
+
+  public CompactionHouseKeeperService() {
+    serviceName = this.getClass().getSimpleName();
+  }
+
+  protected void initTasks(){
```
deniskuzZ, line 40: add @Override

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -38,16 +42,27 @@ public class AcidHouseKeeperService implements MetastoreTaskThread {
   private static final Logger LOG = LoggerFactory.getLogger(AcidHouseKeeperService.class);
 
   private Configuration conf;
-  private boolean isCompactorEnabled;
-  private TxnStore txnHandler;
+  protected TxnStore txnHandler;
+  protected String serviceName;
+  protected Map<FailableRunnable<MetaException>, String> tasks;
```
deniskuzZ, line 47: add `final`

deniskuzZ (commented):
LGTM, some minor stuff

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
```diff
@@ -15,19 +15,23 @@
...
+import org.apache.hadoop.hive.metastore.txn.TxnStore;
+import org.apache.hadoop.hive.metastore.txn.TxnUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-import static org.apache.commons.lang3.Functions.FailableRunnable;
+import org.apache.commons.lang3.function.FailableRunnable;
```
deniskuzZ, line 34: optimize imports: move `Failable` here

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/CompactionHouseKeeperService.java
tarak271: Done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
tarak271: cannot use it CompactionHousekeeper if it is private. tasks variable is also protected only

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
tarak271: done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidHouseKeeperService.java
deniskuzZ: you don't use it there

deniskuzZ:
also please check sonar: https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=4970&resolved=false&inNewCodePeriod=true

PR #4973, opened 2024-01-02
HIVE-27972: Set 'tez' as default value in hive.execution.engine

abstractdog:
Changing the default value of hive.execution.engine to "tez" in HiveConf and handle all the mess that comes after.

this patch contains lots of changes according to a typical pattern of points below:
1. basic change in HiveConf
2. introducing HiveConfForTest, which is a convenience HiveConf for tests, which takes care of setting up proper Tez values without having any Junit magic.
3. using HiveConfForTest where it's applicable and needed
4. added TODO comments where it was too hard to adapt the test to Tez (long-long replication tests) or it doesn't make sense or isn't worth (e.g. hcatalog)
5. copied the default hive-site.xml from  data/conf/ to data/conf/mr, denoting that from now on, "mr" is a special value (because it's not the default anymore) - data/conf/hive-site.xml remains unchanged which means Tez there from now on
6. adapted some qtest drivers to change in 5), so using the new, "mr" related conf
7. remove all occurrences of hive.execution.engine=tez from site xmls, and commented where it's still present, double-checked by:
```
mvn clean -Pitests,iceberg
 grep -iRH "hive.execution.engine" --include="*.xml" --exclude-dir=target -C 2
```
8. added test scope dependencies where it was needed, not a surprise: unit tests ran on MR until now, so they had hadoop mapreduce as dependency, not they must have tez
9. code formatting fixes where it really bothered me (e.g. unused imports)
10. miscellaneous fixed in unit tests, which came from the differences between MR and Tez (and Hive's integration of them)

Tez is the supported engine, and needs to be reflected in code.

No. Depending on their config, customers running on real clusters already use Tez (or MR, sigh). HiveConf default affects code/UT mostly.

No upgrade, only tez dependencies were brought into module where it was needed.

Unit tests are all passing. Where I found challenges, I created follow-up tickets.


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -4564,7 +4564,7 @@ public static enum ConfVars {
     HIVE_DECODE_PARTITION_NAME("hive.decode.partition.name", false,
         "Whether to show the unquoted partition names in query results."),
 
-    HIVE_EXECUTION_ENGINE("hive.execution.engine", "mr", new StringSet(true, "mr", "tez"),
+    HIVE_EXECUTION_ENGINE("hive.execution.engine", "tez", new StringSet(true, "tez", "mr"),
         "Chooses execution engine. Options are: mr (Map reduce, default), tez. While MR\n" +
```
akshat0395, line 4568: Can we update the text as well,
"Chooses execution engine. Options are: tez (default), mr(Map reduce). MR\n" +
        "is a historical engine\n" +
        "and is deprecated in Hive 2 line. It may be removed without further warning."),

akshat0395 (changes requested):

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: sure, this first commit is just for a quick precommit sanity test

akshat0395: Make sense, thx

InvisibleProgrammer:
Hi, there are some possibly related stuff that came into my mind: 

Do we need to change some configuration files as well? 
My first guess is data/conf/hive-site.xml. It can overwrite the default values. 
And also, I'm thinking about the docker stuff as well. 
And StartMiniHS2Cluster overwrites it as well. 

The other thing that I don't know for sure is what is the default engine for qtests?

abstractdog:
> Hi, there are some possibly related stuff that came into my mind:
> 
> Do we need to change some configuration files as well? My first guess is data/conf/hive-site.xml. It can overwrite the default values. And also, I'm thinking about the docker stuff as well. And StartMiniHS2Cluster overwrites it as well.
> 
> The other thing that I don't know for sure is what is the default engine for qtests?

regarding config files: yes I guess, haven't checked though

regarding qtests: default engine for qtests depends on the actual driver, those have to be handled one-by-one:
https://github.com/apache/hive/blob/master/itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestMiniClusters.java

common/src/java/org/apache/hadoop/hive/conf/HiveConfForTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Joiner;
+
+/**
+ * This class is mostly used in unit test environment to prevent polluting the
+ * original HiveConf, and also to provide a way to set some default values for execution.
+ */
+public class HiveConfForTest extends HiveConf {
```
zratkai, line 34: As it is used only for tests, I think it should be in src/test/java folder.

abstractdog: yes, I was thinking about this, maybe it needs some pom.xml changes to refer to it, let me check

common/src/java/org/apache/hadoop/hive/conf/HiveConfForTest.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Get all overlay options as a query string.
+   * This can be used in jdbc connection related tests where we want to pass
+   * only the specific values which were set in this class.
+   */
+  public String getOverlayOptionsAsQueryString() {
+    String queryString = Joiner.on(";").withKeyValueSeparator("=").join(overlay);
+    System.out.println("QUERY STRING: " + queryString);
```
abstractdog, line 97: oops, leftover, I'll remove

aturoczy (approved):
[I want it all](https://www.youtube.com/watch?v=hFDcoX7s6rE)

common/src/java/org/apache/hadoop/hive/conf/HiveConfForTest.java
abstractdog: done

common/src/java/org/apache/hadoop/hive/conf/HiveConfForTest.java
abstractdog: removed

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: this has been addressed in recent commits

service/pom.xml
```diff
@@ -319,6 +319,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.tez</groupId>
+      <artifactId>tez-mapreduce</artifactId>
+      <scope>test</scope>
+    </dependency>
```
SourabhBadhya, line 326: Why is this required?

streaming/pom.xml
```diff
@@ -126,19 +126,43 @@
...
+    <dependency>
+      <groupId>org.apache.tez</groupId>
+      <artifactId>tez-mapreduce</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.tez</groupId>
+      <artifactId>tez-dag</artifactId>
+      <scope>test</scope>
+    </dependency>
```
SourabhBadhya, line 165: Why is this required?

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestAcidOnTez.java
```diff
@@ -111,7 +113,11 @@ public String toString() {
 
   @Before
   public void setUp() throws Exception {
-    hiveConf = new HiveConf(this.getClass());
+    hiveConf = new HiveConfForTest(getClass());
+
+    hiveConf.set("hive.tez.container.size", "128");
+    hiveConf.setBoolean("hive.merge.tezfiles", false);
```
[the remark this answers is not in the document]
abstractdog: it was already part of this test class, I didn't want to touch them
I'll change them to ConfVars

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestAcidOnTez.java
```diff
@@ -1002,22 +984,6 @@ private void runQueries(String engine, String joinType, HiveConf confForTez, Hiv
...
-    conf.setBoolean("tez.local.mode", true);
-    conf.setBoolean("tez.local.mode.without.network", true);
-    conf.set("fs.defaultFS", "file:///");
-    conf.setBoolean("tez.runtime.optimize.local.fetch", true);
-    conf.set("tez.staging-dir", TEST_DATA_DIR);
-    conf.setBoolean("tez.ignore.lib.uris", true);
-    conf.set("hive.tez.container.size", "128");
-    conf.setBoolean("hive.merge.tezfiles", false); 
-    conf.setBoolean("hive.in.tez.test", true);
-  }
```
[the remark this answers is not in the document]
abstractdog: yeah! some of the opts are now taken care of there

service/pom.xml
abstractdog: otherwise test will fail with MRInput class not found or something similar

streaming/pom.xml
abstractdog: tez-mapreduce: MRInput (as mentioned above)
tez-dag: I cannot recall I'll double-check
hive-exec: I cannot recall I'll double-check
hive-common: HiveConfForTest

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestAcidOnTez.java
```diff
@@ -57,6 +58,7 @@
 import org.apache.hadoop.hive.ql.io.orc.OrcInputFormat;
 import org.apache.hadoop.hive.ql.io.orc.Reader;
 import org.apache.hadoop.hive.ql.session.SessionState;
+
```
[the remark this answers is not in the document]
abstractdog: will remove

hcatalog/pom.xml
```diff
@@ -79,6 +79,13 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-common</artifactId>
+      <version>${project.version}</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
```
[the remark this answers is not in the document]
abstractdog: HiveConfForTest usage
I'm checking if any of the catalog tests were adapted to tez, if so, this is needed, if it isn't yet, I'll remove

abstractdog:
> I have complied hive with this PR, built a docker image using docker compose with postgres and tested out a few queries, everything worked fine.

thanks a lot!!

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestAcidOnTez.java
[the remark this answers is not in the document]
abstractdog: done

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestAcidOnTez.java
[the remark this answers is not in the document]
abstractdog: done

service/pom.xml
abstractdog: checked, now all the affected unit tests seem to pass without this dependency, so removing from this PR (maybe a UT which needed this was finally reverted to MR due to some issues and will be taken care of by one of the followup tickets)

streaming/pom.xml
abstractdog: without tez-mapreduce it fails, it's needed
```
java.lang.NoClassDefFoundError: org/apache/tez/mapreduce/hadoop/InputSplitInfo
	at org.apache.hadoop.hive.ql.plan.TezWork.<init>(TezWork.java:91)
	at org.apache.hadoop.hive.ql.parse.GenTezProcContext.<init>(GenTezProcContext.java:176)
	at org.apache.hadoop.hive.ql.parse.TezCompiler.generateTaskTree(TezCompiler.java:634)
	at org.apache.hadoop.hive.ql.parse.TaskCompiler.compile(TaskCompiler.java:302)
...
Caused by: java.lang.ClassNotFoundException: org.apache.tez.mapreduce.hadoop.InputSplitInfo
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	... 67 more
```

without tez-dag:
```
Caused by: java.lang.ClassNotFoundException: org.apache.tez.dag.app.TezTaskCommunicatorImpl
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382) ~[?:1.8.0_292]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418) ~[?:1.8.0_292]
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352) ~[?:1.8.0_292]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351) ~[?:1.8.0_292]
	... 83 more
```

hive-exec: removed, it was a leftover from early versions of this patch when HiveConfForTest was in the hive-exec module

hive-common: needed because of HiveConfForTest
without it:
```[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M4:test (default-test) on project hive-streaming: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M4:test failed: org.junit.platform.commons.JUnitException: TestEngine with ID 'junit-jupiter' failed to discover tests: ClassSelector [className = 'org.apache.hive.streaming.TestStreamingDynamicPartitioning'] resolution failed: org/apache/hadoop/hive/conf/HiveConfForTest: org.apache.hadoop.hive.conf.HiveConfForTest -> [Help 1]
```

hcatalog/pom.xml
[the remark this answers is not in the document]
abstractdog: checked, needed for TestHCatAuthUtil for example
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M4:test (default-test) on project hive-hcatalog-core: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M4:test failed: java.lang.NoClassDefFoundError: org/apache/hadoop/hive/conf/HiveConfForTest: org.apache.hadoop.hive.conf.HiveConfForTest -> [Help 1]
```

pom.xml
```diff
@@ -1334,11 +1334,36 @@
...
+            <artifactId>hadoop-mapreduce-client-common</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-hdfs</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-yarn-client</artifactId>
+          </exclusion>
```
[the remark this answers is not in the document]
abstractdog: this is added to the dependencyManagement section, so from now on, every <dependency> in sub-modules will exclude the same Hadoop transitive dependencies
the reason I added these exclusions was to make it inline with other tez dependencies defined here as a best effort (assuming that exclusions were added for a reason earlier):
https://github.com/apache/hive/blob/1c9969a003b09abc851ae7e19631ad208d3b6066/pom.xml#L1344-L1387

abstractdog:
thanks a lot, @SourabhBadhya  for the review, good catches, I addressed all your comments in 
https://github.com/apache/hive/pull/4973/commits/59ad09bc8d1b91514953f07fd5207d3387d36815

simhadri-g (approved):
LGTM +1 (pending tests)

abstractdog:
> LGTM +1 (pending tests)

thanks @simhadri-g
let me wait for @SourabhBadhya if he has any comments, objections

SourabhBadhya (approved):
LGTM +1 (pending tests)

streaming/pom.xml
qris: I just found this commit because I'm experimenting with Hive and I couldn't insert any data into a new instance because tez is now the default execution engine, but Hive doesn't ship with the required jars and the installation instructions make no mention of the need to install them, or how, so I get this same error message:
```
Caused by: java.lang.NoClassDefFoundError: org/apache/tez/mapreduce/hadoop/InputSplitInfo
```
In the end I followed the instructions [here](https://hive.apache.org/docs/latest/manual-installation_283118363/#tez) but using a tez binary distribution tarball instead of compiling it, which seems to partly work but Hive seems to be stuck on this instead:
```
INFO  : Tez session hasn't been created yet. Opening session
... <very long wait>
Error: Error while compiling statement: FAILED: Execution Error, return code 1 from org.apache.hadoop.hive.ql.exec.tez.TezTask. Cannot invoke "org.apache.hadoop.yarn.api.records.ApplicationId.toString()" because "this.sessionAppId" is null; Query ID: chris_20250808150149_4ba0ae15-fb90-4253-bc14-cbbf5c6eb05d (state=08S01,code=1)
0: jdbc:hive2://localhost:10000> INSERT INTO TABLE students (name) VALUES ('fred flintstone');
```
So I'm reverting to mr. Please could someone update the installation instructions?

PR #4976, opened 2024-01-03
HIVE-27974: Fix flaky test - TestReplicationMetricCollector.testSuccessStageFailure

InvisibleProgrammer:
Fix flaky test. Details in the ticket.

Flaky test on the branch: http://ci.hive.apache.org/job/hive-flaky-check/786/
And also on master: http://ci.hive.apache.org/job/hive-flaky-check/784/

Note: it seems, it became flaky only on downstream. 

Per test case object shouldn't be static

To fix the test

No

No

By running tests and flaky test checker

### DIFF
--- a/ql/src/test/org/apache/hadoop/hive/ql/parse/repl/metric/TestReplicationMetricCollector.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/parse/repl/metric/TestReplicationMetricCollector.java
@@ -49,7 +49,6 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
-import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
 
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


ayushtkn (approved):
LGTM

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/metric/TestReplicationMetricCollector.java
```diff
@@ -60,6 +60,7 @@
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.*;
```
kokila-19, line 63: Please remove the wildcart import from the static import.

InvisibleProgrammer: Hi, thank you for spotting this. I fixed this and the other import-only changes.

PR #4978, opened 2024-01-04
HIVE-27979: HMS alter_partitions log adds table name

cxzl25:
add log

1. `alter_partitions` outputs multiple partition values, but there is no table name. 
2. When updating the statistics of the partition, there is only the table name, but the partition name is not known.

No

No

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -5919,7 +5919,8 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
 
     if (LOG.isInfoEnabled()) {
       for (Partition tmpPart : new_parts) {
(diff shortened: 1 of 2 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -5919,7 +5919,7 @@ private void alter_partitions_with_environment_context(String catName, String db
 
     if (LOG.isInfoEnabled()) {
       for (Partition tmpPart : new_parts) {
-        LOG.info("New partition values:" + tmpPart.getValues());
+        LOG.info("New partition values: {}.{}.{}:{} ", catName, db_name, tbl_name, tmpPart.getValues());
```
ayushtkn, line 5922: Can we put values like: catalog: {} database: {} table: {} partition: {} rather than using .

ayushtkn (approved):
LGTM

zhangbutao (approved):
LGTM

PR #4980, opened 2024-01-06
HIVE-27960:Invalid function error when using custom udaf

tc-gaoxiong:
[HIVE-27960](https://issues.apache.org/jira/browse/HIVE-27960) : Invalid function error when using custom udaf

Why are the changes needed?
When a permanent udaf used before over() function, hive will throw invalid function error.

no

no

How was this patch tested?
on aliyun emr hive 2.3.9


tc-gaoxiong:
sorry for this !

PR #4981, opened 2024-01-07
HIVE-27960:Invalid function error when using custom udaf

tc-gaoxiong:
[HIVE-27960](https://issues.apache.org/jira/browse/HIVE-27960) : Invalid function error when using custom udaf

Why are the changes needed?
When a permanent udaf used before over() function, hive will throw invalid function error.

no

no

How was this patch tested?
aliyun emr hive 2.3.9


ql/src/test/queries/clientpositive/ptf_register_use_2.q
```diff
@@ -0,0 +1,2 @@
+-- session 2: use function
```
zhangbutao, line 1: Looks like your qtest depends on the two different session. I am not sure if it is a good way to create two qtest to achieve this, then this needs the two qtests must be ordered, that is to say `ptf_register_use_1.q` must be in front of `ptf_register_use_2.q`. But i think the qtests framework may not work as you expected sometimes, and will lead your test flaky.

I may suggest you adding the create function in https://github.com/apache/hive/blob/master/data/files/datasets/temp_udf/load.hive.sql (Like [HIVE-23558](https://github.com/apache/hive/pull/4928/files#top)), and add `--! qt:dataset:temp_udf` in another qtest to invoke this funtion.  This can aslo create two seesions and can reproduce your issue.

cc @ayushtkn may give some thought.

ql/src/test/queries/clientpositive/ptf_register_use_2.q
```diff
@@ -0,0 +1,2 @@
...
+select ptf_register_use_func() over();
```
zhangbutao, line 2: before the query, you need add its explain test `explain select ptf_register_use_func() over();`

ql/src/java/org/apache/hadoop/hive/ql/exec/Registry.java
```diff
@@ -385,7 +385,9 @@ public WindowFunctionInfo getWindowFunctionInfo(String functionName) throws Sema
     // First try without qualifiers - would resolve builtin/temp functions
     FunctionInfo info = getFunctionInfo(WINDOW_FUNC_PREFIX + functionName);
     // Try qualifying with current db name for permanent functions
-    if (info == null) {
+    if (info == null
+            // and try register function to session
+            && FunctionRegistry.getFunctionInfo(functionName) != null) {
```
zhangbutao, line 390: Suggested to change
```
    // Try qualifying with current db name for permanent functions and registering functions to session
    if (info == null && FunctionRegistry.getFunctionInfo(functionName) != null) {
```

okumin:
Reading the description of HIVE-27960 and quickly testing the test case on my machine, I would say it should be merged into the 2.3 branch. The current master is likely to work without the problem.

### Hive 2.3

```
Beeline version 2.3.8 by Apache Hive
0: jdbc:hive2://hive-hiveserver2:10000/defaul> create function row_number2 as 'org.apache.hadoop.hive.ql.udf.generic.GenericUDAFRowNumber';
No rows affected (0.79 seconds)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> select row_number2() over();
Error: Error while compiling statement: FAILED: SemanticException Failed to breakup Windowing invocations into Groups. At least 1 group must only depend on input columns. Also check for circular dependencies.
Underlying error: Invalid function row_number2 (state=42000,code=40000)
```

### Hive 3.1

```
Beeline version 3.1.2 by Apache Hive
0: jdbc:hive2://hive-hiveserver2:10000/defaul> create function row_number2 as 'org.apache.hadoop.hive.ql.udf.generic.GenericUDAFRowNumber';
...
No rows affected (0.995 seconds)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> select row_number2() over();
...
+-----------------------+
| row_number2_window_0  |
+-----------------------+
| 1                     |
+-----------------------+
1 row selected (12.93 seconds)
```

### Hive 4(latest master)

```
Beeline version 4.0.0-beta-2-SNAPSHOT by Apache Hive
0: jdbc:hive2://hive-hiveserver2:10000/defaul> create function row_number2 as 'org.apache.hadoop.hive.ql.udf.generic.GenericUDAFRowNumber';
...
No rows affected (1.582 seconds)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> select row_number2() over();
...
+-----------------------+
| row_number2_window_0  |
+-----------------------+
| 1                     |
+-----------------------+
1 row selected (16.311 seconds)
```

zhangbutao:
@okumin On hive master, If you open another beeline and execute the select query, you will find this issue.

ql/src/test/queries/clientpositive/ptf_register_use_2.q
tc-gaoxiong: Thanks for your review! As you suggest, this pr should add `data/files/datasets/temp_udaf/load.hive.sql` file to create function, and a `ptf_register_use.q` file to invoke function.  

I test it local. By log, hive does use two session to exeuct create an invoke.

```txt
2024-01-15T22:45:28,041  INFO [main] SessionState: Hive Session ID = bbc2a0ea-227a-4c07-a7a6-52c3a50e1e06
...
2024-01-15T22:45:29,451  INFO [main] SessionState: Hive Session ID = e8ffa509-9272-483b-a885-ef7d92e705cd
```

Also, the create does execute before invoke :
https://github.com/apache/hive/blob/06ef7c8231501c0314875d640f081adc35a70db6/itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CoreCliDriver.java#L112-L115

So, are we agreed on the idea?

zhangbutao: Yes, if you don't have other better ways, let's use this idea. I think this idea is better than creating two qtests at least.

tc-gaoxiong: Another question, can i use `git push --force` to amend this modify on hive repository?

zhangbutao: sure

zhangbutao:
@tc-gaoxiong You can rebase this PR to get a green CI.

tc-gaoxiong:
> @tc-gaoxiong You can rebase this PR to get a green CI.

Thanks !!!

tc-gaoxiong:
Just code one line, but so hard. I am sorry for those stupid operation.

zhangbutao (approved):
LGTM

tc-gaoxiong:
ci green get, can this pr merge?

PR #4984, opened 2024-01-08
HIVE-27984: Support backward compatibility of hms thrift struct about column stats

zhangbutao:

Please see details from HIVE-27984.
HIVE-22046 indroduced the must required `engine` field in hms thrift api, which broken the backward compatibility. This lead to hive3 hms thrift client can not call the hive4 column stats api.

This PR try to make `engine ` field optional and then this will reduce the incompatibility issues.

No

No

Test with local cluster. And this change has no negative effect to Hive.

### DIFF
--- a/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
+++ b/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
@@ -600,7 +600,7 @@ struct ColumnStatistics {
 2: required list<ColumnStatisticsObj> statsObj,
 3: optional bool isStatsCompliant, // Are the stats isolation-level-compliant with the
                                                       // the calling query?
-4: optional string engine
+4: optional string engine = "hive"
 }
 
 // FileMetadata represents the table-level (in case of unpartitioned) or partition-level
@@ -993,7 +993,7 @@ struct GetPartitionsByNamesRequest {
   4: optional bool get_col_stats,
   5: optional list<string> processorCapabilities,
   6: optional string processorIdentifier,
-  7: optional string engine,
+  7: optional string engine = "hive",
   8: optional string validWriteIdList,
   9: optional bool getFileMetadata,
   10: optional i64 id=-1,  // table id
@@ -1710,7 +1710,7 @@ struct GetTableRequest {
   7: optional bool getColumnStats,
   8: optional list<string> processorCapabilities,
   9: optional string processorIdentifier,
-  10: optional string engine,
+  10: optional string engine = "hive",
   11: optional i64 id=-1 // table id
 }
 
--- a/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
+++ b/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
@@ -725,7 +725,7 @@ struct SetPartitionsStatsRequest {
 2: optional bool needMerge, //stats need to be merged with the existing stats
 3: optional i64 writeId=-1,         // writeId for the current query that updates the stats
 4: optional string validWriteIdList, // valid write id list for the table for which this struct is being sent
-5: required string engine //engine creating the current request
+5: optional string engine = "hive" //engine creating the current request
 }
 
 struct SetPartitionsStatsResponse {
@@ -901,7 +901,7 @@ struct TableStatsRequest {
  3: required list<string> colNames
  4: optional string catName,
  5: optional string validWriteIdList,  // valid write id list for the table for which this struct is being sent
- 6: required string engine, //engine creating the current request
+ 6: optional string engine = "hive", //engine creating the current request
  7: optional i64 id=-1 // table id
 }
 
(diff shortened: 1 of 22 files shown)


standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
```diff
@@ -725,7 +725,7 @@ struct SetPartitionsStatsRequest {
 2: optional bool needMerge, //stats need to be merged with the existing stats
 3: optional i64 writeId=-1,         // writeId for the current query that updates the stats
 4: optional string validWriteIdList, // valid write id list for the table for which this struct is being sent
-5: required string engine //engine creating the current request
+5: optional string engine = "hive" //engine creating the current request
```
zhangbutao, line 728: Make the `engine` field optional, and give it a default value `hive` which is consistent with the current existing code logic.

pan3793:
@zhangbutao hi, have you observed backward compatibility issues with Hive 2.3 client and Hive 4 HMS? also cc @sunchao

zhangbutao:
> @zhangbutao hi, have you observed backward compatibility issues with Hive 2.3 client and Hive 4 HMS? also cc @sunchao

@pan3793 I think it depends your use case, for this change, TrinoDB&StarRocks will use its customized thrift client and call `ThriftHiveMetastore$Client.recv_get_table_statistics_re` to fetch hive tables column stats, so it will occur compatibility issues. But as far as i know, Spark won't call this api to get hive tables column stats, so it is safe for Spark to integrate Hive4 using its embedded Hive2.3 client.
BTW, we have used Spark3 to query hive4 tables for a long time, and we have not found any compatibility issues.

pan3793:
@zhangbutao great! thanks for such information

zhangbutao:
Can i request your review? @ayushtkn @dengzhhu653 @deniskuzZ 
Thx.

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
dengzhhu653: could we set the `engine` to default value `hive` without changing it to `optional`?

zhangbutao: That doesn't work. `required `will enforce check `engine `field in construction method though it has a default value. For some others framework(Trino) will use this class `SetPartitionsStatsRequest` but don't want to care the `engine `field, and they just reuse hive stats.

`optional `has no any problem in current hive code. only need partial trival changes in `HiveMetaStoreClient.java`. I think it is safe. :)

okumin: Though I am not familiar with Thrift, I expect the following behaviors.

| Client protocol | HMS version | Behavior |
|-|-|-|
| Hive 3 (no engine is set) | Hive 4 w/o HIVE-27984 | Error |
| Hive 4 w/o HIVE-27984 | Hive 4 w/o HIVE-27984 | The engine type specified by the client is used |
| Hive 4 w/ HIVE-27984 | Hive 4 w/o HIVE-27984 | The engine type specified by the client is used |
| Hive 3 (no engine is set) | Hive 4 w/ HIVE-27984 | engine=hive |
| Hive 4 w/o HIVE-27984 | Hive 4 w/ HIVE-27984 | The engine type specified by the client is used |
| Hive 4 w/ HIVE-27984 | Hive 4 w/ HIVE-27984 | The engine type specified by the client is used |

okumin (approved):
Thanks for handling this issue!

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
zhangbutao: Yes, you are almost right. To be more specific, for example, Trino will get hms table column statistics by its thrift hms client which is compatible with hive3.
https://github.com/trinodb/trino/blob/057f2d435d6461855ff0f904dbd675b39db97d1e/plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftHiveMetastoreClient.java#L331-L336

But `TableStatsRequest` in Hive4 need pass `engine` through constructor function, and the `engine` field is required, therefore:

1. for hive3 thrift client, this api call will fail due to the missed `engine `field if HMS server doesn't have this PR, As there is no place for hive3 thrift client to pass `engine `field. But this api call will sucessfully if HMS has this PR(`engine` field is optional and has the defalut value).
2. for hive4 thrift client without this PR, you also must pass `engine `field in `TableStatsRequest` to call this api correctly no matter if HMS server has this PR. Because `engine` field is needed in `TableStatsRequest` constructor function.
3. for hive4 thrift client with this PR. Whether HMS server have this PR or not
, you can call the api sucessfully without pass `engine` as the `engine` field is not mandatory in `TableStatsRequest` constructor function, and you can explicitly reset the `engine` fileld by `TableStatsRequest::setEngine` if you don't want the default `hive engine` column statistics. But i think most  framework(like trino) will want reuse hive engine column statistics.

Sounds a little verbose.. :) In one word, this PR can make hive3 thrift hms client worked well with hive4 hms when calling statistics related api. I have tested this PR with Trino, and i think it is good.

dengzhhu653 (approved):

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
okumin: Thanks for elaborating the behaviors. I'd say it is flexible and less surprising for people.

PR #4985, opened 2024-01-08
HIVE-27987: Implement YESTERDAY UDF in Hive

MRKKmrkk:
Implement YESTERDAY UDF in Hive

In my daily work, I frequently need to examine data from the previous day. Currently, in Hive, the only available method to obtain the date of yesterday is by using calculations like date_sub(current_date, 1). It would be convenient if Hive could introduce a method named 'yesterday' to facilitate the quick and easy retrieval of the date corresponding to the previous day.

No

No

Created  qtests as part of this change

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/FunctionRegistry.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/FunctionRegistry.java
@@ -452,6 +452,7 @@ public final class FunctionRegistry {
     system.registerGenericUDF("current_schema", GenericUDFCurrentSchema.class);
     system.registerGenericUDF("current_catalog", GenericUDFCurrentCatalog.class);
     system.registerGenericUDF("current_date", GenericUDFCurrentDate.class);
+    system.registerGenericUDF("yesterday", GenericUDFYesterday.class);
     system.registerGenericUDF("current_timestamp", GenericUDFCurrentTimestamp.class);
     system.registerGenericUDF("current_user", GenericUDFCurrentUser.class);
     system.registerGenericUDF("current_groups", GenericUDFCurrentGroups.class);
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFYesterday.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFYesterday.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.apache.hadoop.hive.ql.udf.generic;
 
 import org.apache.hadoop.hive.common.type.Date;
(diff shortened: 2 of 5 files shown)


ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFYesterday.java
```diff
@@ -0,0 +1,76 @@
...
+                    "The function YESTERDAY does not take any arguments, but found "
+                            + arguments.length);
+        }
+
+        if (yesterday == null) {
+            SessionState ss = SessionState.get();
+            ZonedDateTime dateTime = ss.getQueryCurrentTimestamp().atZone(
+                    ss.getConf().getLocalTimeZone()).minusDays(1L);
+            Date dateVal = Date.valueOf(
+                    dateTime.toString().substring(0, 10));
```
okumin, line 50: I think it is more straightforward and robust to use this method in this UDF and CURRENT_DATE.
https://github.com/apache/hive/blob/rel/release-4.0.0-beta-1/common/src/java/org/apache/hadoop/hive/common/type/Date.java#L205-L207

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFYesterday.java
```diff
@@ -0,0 +1,76 @@
+package org.apache.hadoop.hive.ql.udf.generic;
```
okumin, line 1: We need a license.

ql/src/java/org/apache/hadoop/hive/ql/exec/FunctionRegistry.java
```diff
@@ -452,6 +452,7 @@ public final class FunctionRegistry {
     system.registerGenericUDF("current_schema", GenericUDFCurrentSchema.class);
     system.registerGenericUDF("current_catalog", GenericUDFCurrentCatalog.class);
     system.registerGenericUDF("current_date", GenericUDFCurrentDate.class);
+    system.registerGenericUDF("yesterday", GenericUDFYesterday.class);
```
okumin, line 455: We need to update this file when we add a new UDF.
https://github.com/okumin/hive/commits/master/ql/src/test/results/clientpositive/llap/show_functions.q.out

okumin:
Googling some database systems, I didn't find ones with YESTERDAY or a similar UDF. But I imagine this UDF would be helpful

MRKKmrkk:
> Googling some database systems, I didn't find ones with YESTERDAY or a similar UDF. But I imagine this UDF would be helpful at this point. I'm not confident though.

MRKKmrkk:
> Googling some database systems, I didn't find ones with YESTERDAY or a similar UDF. But I imagine this UDF would be helpful at this point. I'm not confident though.

Thank you for your review! I understand your confusion, actually clickhouse has similar UDF: https://clickhouse.com/docs/en/sql-reference/functions/date-time-functions#yesterday, and this function is widely used by both me and my colleagues in our daily work.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFYesterday.java
MRKKmrkk: Done!

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFYesterday.java
MRKKmrkk: Done!

ql/src/java/org/apache/hadoop/hive/ql/exec/FunctionRegistry.java
MRKKmrkk: Done!

MRKKmrkk:
@okumin Could you please check if this fix is ok? Thanks.

okumin:
@MRKKmrkk Thanks for sharing the use case. I'm putting some thoughts here.

1. CI is failing. Could you please check [this](http://ci.hive.apache.org/blue/rest/organizations/jenkins/pipelines/hive-precommit/branches/PR-4985/runs/4/nodes/520/log/?start=0) and [this](http://ci.hive.apache.org/blue/rest/organizations/jenkins/pipelines/hive-precommit/branches/PR-4985/runs/4/nodes/567/log/?start=0)? I guess you need to generate the result using `TestMiniLlapLocalCliDriver`
2. I put some minor comments
3. I can personally give +1 if we resolve the above two points. But note that I am not a committer and you would need to find another person to take another look

PR #4988, opened 2024-01-09
HIVE-27991: Utilise FanoutWriters when inserting records in an Iceberg table when the records are unsorted

SourabhBadhya:

Utilise FanoutWriters when inserting records in an Iceberg table when the records are unsorted

FanoutWriter is a writer in Iceberg which can be used to write records in a table. This writer keeps all the file handles open, until the write is finished. FanoutWriters is used as the writer when the incoming records are unsorted. We can by default have some mechanism to switch to using FanoutWriters instead of ClusteredWriters when custom sort expressions are not present for the given table/query.
Similar stuff is already implemented in Spark - https://github.com/apache/iceberg/pull/8621

No

No

QTests

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
@@ -29,9 +29,11 @@ import org.apache.iceberg.Schema;
 import org.apache.iceberg.data.Record;
 import org.apache.iceberg.io.ClusteredDataWriter;
 import org.apache.iceberg.io.DataWriteResult;
+import org.apache.iceberg.io.FanoutDataWriter;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.io.FileWriterFactory;
 import org.apache.iceberg.io.OutputFileFactory;
+import org.apache.iceberg.io.PartitioningWriter;
 import org.apache.iceberg.mr.hive.FilesForCommit;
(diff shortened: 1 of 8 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
```diff
@@ -41,9 +42,9 @@ class HiveIcebergRecordWriter extends HiveIcebergWriterBase {
 
   HiveIcebergRecordWriter(Schema schema, Map<Integer, PartitionSpec> specs, int currentSpecId,
       FileWriterFactory<Record> fileWriterFactory, OutputFileFactory fileFactory, FileIO io,
-      long targetFileSize) {
-    super(schema, specs, io,
-        new ClusteredDataWriter<>(fileWriterFactory, fileFactory, io, targetFileSize));
+      long targetFileSize, boolean fanoutEnabled) {
+    super(schema, specs, io, fanoutEnabled ? new FanoutDataWriter<>(fileWriterFactory, fileFactory, io, targetFileSize)
```
zhangbutao, line 46: Nit: It would be better if we can use a separate function to extract&simplify this long code.

zhangbutao (approved):
LGTM

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
SourabhBadhya: Done.

PR #4994, opened 2024-01-10
HIVE-27977: Fix ordering flakiness in TestHplSqlViaBeeLine

abstractdog:
Introduce ORDER BY clause in some unit tests.

Bumped into intermittent test failures in TestHplSqlViaBeeLine while moving it from MR to Tez, and realized that the ordering is not defined in the test cases, so it makes no sense to look for the execution engine differences as long as we don't have a stable golden output.

No.

No.

With the touched unit tests.


zhangbutao (approved):
LGTM

ayushtkn (approved):
LGTM

PR #4995, opened 2024-01-10
HIVE-27994: Optimize renaming the partitioned table

dengzhhu653:

### DIFF
--- a/ql/src/test/queries/clientpositive/rename_table.q
+++ b/ql/src/test/queries/clientpositive/rename_table.q
@@ -1,0 +1,40 @@
+--! qt:dataset:src
+--! qt:dataset:part
+-- This test verifies that if the table after rename can still fetch the column statistics
+set hive.stats.kll.enable=true;
+set metastore.stats.fetch.bitvector=true;
+set metastore.stats.fetch.kll=true;
+set hive.stats.autogather=true;
+set hive.stats.column.autogather=true;
+
+CREATE TABLE rename_partition_table0 (key STRING, value STRING) PARTITIONED BY (part STRING)
+STORED AS ORC;
+
+INSERT OVERWRITE TABLE rename_partition_table0 PARTITION (part = '1') SELECT * FROM src where rand(1) < 0.5;
+ALTER TABLE rename_partition_table0 ADD COLUMNS (new_col INT);
+INSERT OVERWRITE TABLE rename_partition_table0 PARTITION (part = '2') SELECT src.*, 1 FROM src;
+
+ALTER TABLE rename_partition_table0 RENAME TO rename_partition_table1;
+DESCRIBE FORMATTED rename_partition_table1;
+DESCRIBE FORMATTED rename_partition_table1 PARTITION (part='1') key;
+DESCRIBE FORMATTED rename_partition_table1 PARTITION (part='1') value;
+DESCRIBE FORMATTED rename_partition_table1 PARTITION (part='2') key;
+DESCRIBE FORMATTED rename_partition_table1 PARTITION (part='2') value;
+DESCRIBE FORMATTED rename_partition_table1 PARTITION (part='2') new_col;
+
+CREATE EXTERNAL TABLE rename_partition_table_ext0 (key STRING, value STRING) PARTITIONED BY (part STRING)
+STORED AS ORC;
+
+INSERT OVERWRITE TABLE rename_partition_table_ext0 PARTITION (part = '1') SELECT * FROM src where rand(1) < 0.5;
+ALTER TABLE rename_partition_table_ext0 CHANGE COLUMN value val STRING CASCADE;
+INSERT OVERWRITE TABLE rename_partition_table_ext0 PARTITION (part = '2') SELECT * FROM src;
+
+ALTER TABLE rename_partition_table_ext0 RENAME TO rename_partition_table_ext1;
+DESCRIBE FORMATTED rename_partition_table_ext1;
+DESCRIBE FORMATTED rename_partition_table_ext1 PARTITION (part='1') key;
+DESCRIBE FORMATTED rename_partition_table_ext1 PARTITION (part='1') val;
+DESCRIBE FORMATTED rename_partition_table_ext1 PARTITION (part='2') key;
+DESCRIBE FORMATTED rename_partition_table_ext1 PARTITION (part='2') val;
+
+DROP TABLE rename_partition_table1;
+DROP TABLE rename_partition_table_ext1;
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
@@ -64,6 +64,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.locks.ReentrantLock;
@@ -181,7 +182,7 @@ class DirectSqlUpdatePart {
             e -> e.partitionId).collect(Collectors.toList()
     );
 
-    prefix.append("select \"PART_ID\", \"COLUMN_NAME\" from \"PART_COL_STATS\" WHERE ");
+    prefix.append("select \"PART_ID\", \"COLUMN_NAME\", \"ENGINE\" from \"PART_COL_STATS\" WHERE ");
     TxnUtils.buildQueryWithINClause(conf, queries, prefix, suffix,
             partIdList, "\"PART_ID\"", true, false);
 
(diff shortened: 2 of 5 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -181,7 +182,7 @@ private void populateInsertUpdateMap(Map<PartitionInfo, ColumnStatistics> statsP
             e -> e.partitionId).collect(Collectors.toList()
     );
 
-    prefix.append("select \"PART_ID\", \"COLUMN_NAME\" from \"PART_COL_STATS\" WHERE ");
+    prefix.append("select \"PART_ID\", \"COLUMN_NAME\", \"ENGINE\" from \"PART_COL_STATS\" WHERE ");
```
zhangbutao, line 185: Why do we need `ENGINE ` field here?
Isn't the `PART_ID ` field enough to identify the specific partition?

dengzhhu653: It could be possible,
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L10130
in the above example, we use different engines to fetch the statistics with the same partition and column.

soumyakanti3578:
@dengzhhu653 There are some minor issues found by SonarCloud related to readability and maintainability: https://sonarcloud.io/project/issues?cleanCodeAttributeCategories=INTENTIONAL&resolved=false&pullRequest=4995&id=apache_hive&open=AYz34ifC28xE78Z9jBVO

Fixing these are optional since these are very minor but good to have! :)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
zhangbutao: Make sense.  if so, Shouldn't we add the `ENGNE `field in the `WHERE `sub-statement?
Current direct SQL query:
`select "PART_ID", "COLUMN_NAME" from "PART_COL_STATS" WHERE ("PART_ID" in (141))`

Maybe can be changed to:
`select "PART_ID", "COLUMN_NAME" from "PART_COL_STATS" WHERE ("PART_ID" in (141)) and ENGINE = hive`

BTW, we also have another field `CAT_NAME` in `PART_COL_STATS` to differentiate column stas between multi catalog. Should we also consider it here?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -221,25 +223,46 @@ private void populateInsertUpdateMap(Map<PartitionInfo, ColumnStatistics> statsP
...
-      try {
-        pst = dbConn.prepareStatement(update);
-        StatObjectConverter.initUpdatedColumnStatement(mPartitionColumnStatistics, pst);
-        LOG.debug("Going to execute update " + update);
-        int numUpdate = pst.executeUpdate();
-        if (numUpdate != 1) {
-          throw new MetaException("Invalid state of  PART_COL_STATS for PART_ID " + partId);
+      StringBuilder update = new StringBuilder("UPDATE \"PART_COL_STATS\" SET ")
+          .append(StatObjectConverter.getUpdatedColumnSql(mPartitionColumnStatistics))
+          .append(" WHERE \"PART_ID\" = ? AND \"COLUMN_NAME\" = ? AND \"ENGINE\" = ?");
```
zhangbutao, line 231: Same as above. Should we also consider addding `CAT_NAME` field here?

dengzhhu653:
> @dengzhhu653 There are some minor issues found by SonarCloud related to readability and maintainability: https://sonarcloud.io/project/issues?cleanCodeAttributeCategories=INTENTIONAL&resolved=false&pullRequest=4995&id=apache_hive&open=AYz34ifC28xE78Z9jBVO
> 
> Fixing these are optional since these are very minor but good to have! :)

Thank you @soumyakanti3578, will fix them.
cc @saihemanth-cloudera @henrib @deniskuzZ @ayushtkn cloud you take a look as well if have cycles please?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
dengzhhu653: Morning @zhangbutao!
In this method we want to get the insert or the update statistics from `Map<PartitionInfo, ColumnStatistics> statsPartInfoMap`, there is no guarantee that all of the `statsPartInfoMap` are for the engine hive or the same engine, so PartColNameInfo needs to feed with the engine info when compared with the stats in `statsPartInfoMap`.

> BTW, we also have another field CAT_NAME in PART_COL_STATS to differentiate column stas between multi catalog. Should we also consider it here?

I think we don't need to, the `PART_ID` here has the same effect for clarifying the catalog.

zhangbutao: Thanks for explanation!

> there is no guarantee that all of the statsPartInfoMap are for the engine hive or the same engine,

Is this code snippet only for `insert` statement?

If so, the one `insert` statement is must from the one same `ENGINE` (like hive or other compute engine), so the `statsPartInfoMap` will have the same `ENGINE `value.

dengzhhu653: > Is this code snippet only for insert statement?

No, as you see, this PR aims for optimizing the partitioned table rename. Even the `statsPartInfoMap` has the same `ENGINE` value, we must be careful not to override the stats(for other engines) for the same partition column on the database.

zhangbutao: I did some code debug. Only found that `insert` into statement gone through this code block(method `populateInsertUpdateMap`), but `alter table rename` not. Maybe i missed some configuration?

Since `statsPartInfoMap` has the same `ENGINE `value, why not we use this direct sql to filter the PART_ID based on the `ENGINE` (e.g. `hive`)? 
`select "PART_ID", "COLUMN_NAME" from "PART_COL_STATS" WHERE ("PART_ID" in (141)) and ENGINE = hive`

dengzhhu653: @zhangbutao, you need to apply this change to see `populateInsertUpdateMap` is invoked by the rename operation, there is a batch rename for stats.
In my idea, there is no document that all the stats in `statsPartInfoMap` must be the same engine, so we cannot simply using the `ENGINE = hive` for all stats, besides `ENGINE = hive` in the filter is almost the same as in `select \"ENGINE\"` when there has limited numbers of engines on this partition.

zhangbutao: ok,i didn't apply your change to do test. No need to consideration about this comment, i just want to explore more details about stats usage in Hive.
Thanks. 😃

ql/src/test/queries/clientpositive/rename_table.q
```diff
@@ -0,0 +1,40 @@
+--! qt:dataset:src
+--! qt:dataset:part
+-- This test verifies that if the table after rename can still fetch the column statistics
+set hive.stats.kll.enable=true;
+set metastore.stats.fetch.bitvector=true;
+set metastore.stats.fetch.kll=true;
+set hive.stats.autogather=true;
+set hive.stats.column.autogather=true;
+
+CREATE TABLE rename_partition_table0 (key STRING, value STRING) PARTITIONED BY (part STRING)
```
saihemanth-cloudera, line 10: Can we also test managed tables?

saihemanth-cloudera (commented):
Changes looks good to me. +1
Will wait for @zhangbutao's approval.

saihemanth-cloudera:
@dengzhhu653 -- Can you also create a follow-up jira to optimize this further where we directly update the db_name and table_name in the 'tab_col_stats' and 'part_col_stats' tables instead of fetching the stats, updating db/table names and then persisting it to the DB?

zhangbutao (approved):
LGTM +1 
Batch commit is a good way to improve the performance of alter partitioned tables.
Let's go ahead!

dengzhhu653:
> @dengzhhu653 -- Can you also create a follow-up jira to optimize this further where we directly update the db_name and table_name in the 'tab_col_stats' and 'part_col_stats' tables instead of fetching the stats, updating db/table names and then persisting it to the DB?

Created HIVE-28011 for tracking this

dengzhhu653:
Thank you @zhangbutao @soumyakanti3578 and @saihemanth-cloudera for the comment and review!

PR #3117, opened 2022-03-18
HIVE-26047: Vectorized LIKE UDF should use Re2J regex to address JDK-8203458

nareshpr:
### What changes were proposed in this pull request?
Vectorized LIKE udf is taking proportionately higher time depending on the size of input string in UDF.

### Why are the changes needed?
To support filter condition based on input data.

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
Added testcase as part of PR.

PR #4998, opened 2024-01-11
HIVE-26339: HIVE-26047 Related LIKE pattern issues

ryukobayashi:

Vectorized LIKE udf is taking proportionately higher time depending on the size of input string in UDF. And, I found a problem with some regular expressions.

To support filter condition based on input data.

No.

Added testcase as part of PR.


ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +133,93 @@ public Checker tryCreate(String pattern) {
       return new ComplexChecker("^" + UDFLike.likePatternToRegExp(pattern) + "$");
     }
   }
+
+  private enum PatternType {
```
deniskuzZ, line 137: can we reuse the enum from `UDFLike`?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +133,93 @@ public Checker tryCreate(String pattern) {
...
+      this.type = type;
+      this.pattern = pattern;
+    }
+
+    public String check() {
+      PatternType lastType = PatternType.NONE;
+      int length = pattern.length();
+      int beginIndex = 0;
+      int endIndex = length;
+      char lastChar = 'a';
```
deniskuzZ, line 163: it doesn't seem to be a generic code. why do we hardcode here?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
ryukobayashi: OK, I will reuse it.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
ryukobayashi: This is a value for initialization. There is no problem if it is a value other than `\`, so Initialize it to 0 for clarity.

ryukobayashi:
@deniskuzZ I have corrected what you pointed out, so could you please check it?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -58,12 +56,11 @@ protected List<CheckerFactory> getCheckerFactories() {
...
    */
   private static class BeginCheckerFactory implements CheckerFactory {
-    private static final Pattern BEGIN_PATTERN = Pattern.compile("([^_%]+)%");
-
     public Checker tryCreate(String pattern) {
-      Matcher matcher = BEGIN_PATTERN.matcher(pattern);
-      if (matcher.matches()) {
-        return new BeginChecker(matcher.group(1));
+      CheckerPattern checkerPattern = new CheckerPattern(UDFLike.PatternType.BEGIN, pattern);
+      String matche = checkerPattern.check();
```
deniskuzZ, line 61: can we transform CheckerPattern to enum and use it like
````
UDFLikePattern.BEGIN.apply(pattern)
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java [hunk 4]
```diff
@@ -140,4 +133,84 @@ public Checker tryCreate(String pattern) {
...
+      this.pattern = pattern;
+    }
+
+    public String check() {
+      UDFLike.PatternType lastType = UDFLike.PatternType.NONE;
+      int length = pattern.length();
+      int beginIndex = 0;
+      int endIndex = length;
+      char lastChar = 0;
+      String strPattern = new String();
```
deniskuzZ, line 155: use `StringBuilder`

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +133,84 @@ public Checker tryCreate(String pattern) {
...
+      String simplePattern;
+
+      for (int i = 0; i < length; i++) {
```
deniskuzZ, line 158: @ryukobayashi, are you saying that your code with char matching is more efficient than regex?  
Could you please create a benchmark (JMH) and share the results?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
ryukobayashi: OK, that's how I fixed it.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
ryukobayashi: Fixed.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
ryukobayashi: @deniskuzZ  This was originally a bug in `java.util.regex`, so it was a fix to use `com.google.re2j` instead of using regex as is. The comments below led me to implement a custom algorithm: https://github.com/apache/hive/pull/3117#issuecomment-1076183992

So, I don't have any particular attachment to custom algorithms. Which is better?

deniskuzZ: 👍, do we have any benchmarks that prove efficiency of implemented algorithm? also, usually for such kind of tasks (pattern matching) state machines are one of the recommended techniques.

ryukobayashi: @deniskuzZ Sorry for the late reply as I was unwell. This algorithm is not something I directly implemented. It's already implemented in a UDF that runs when Vectorized is off. So I don't know if there are any benchmarks to prove its efficiency: https://github.com/apache/hive/blob/release-4.0.0-rc0/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java#L108-L158

deniskuzZ: oh ok, please give me some time to check

deniskuzZ: shouldn't `parseSimplePattern` logic be the same? if yes, could we move this to util class and reuse in both places?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +108,100 @@ public Checker tryCreate(String pattern) {
...
+      int beginIndex = 0;
+      int endIndex = length;
+      char lastChar = 0;
+      StringBuilder strPattern = new StringBuilder();
+      String simplePattern;
+
+      for (int i = 0; i < length; i++) {
+        char n = pattern.charAt(i);
+        if (n == '_') { // such as "a_b"
+          if (lastChar != '\\') { // such as "a%bc"
```
deniskuzZ, line 152: should it be `a_bc`?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +108,100 @@ public Checker tryCreate(String pattern) {
...
+            return null;
+          } else { // such as "abc\%de%"
```
deniskuzZ, line 154: `abc\_de%` ?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +108,100 @@ public Checker tryCreate(String pattern) {
...
+            if (lastChar != '\\') { // such as "a%bc"
+              lastType = UDFLike.PatternType.CHAINED;
+            } else if (lastChar == '\\') { // such as "a\%bc"
+              strPattern.append(pattern.substring(beginIndex, i - 1));
+              beginIndex = i;
+              if (lastType == UDFLike.PatternType.CHAINED) {
+                lastType = UDFLike.PatternType.COMPLEX;
+              }
+            } else {
+              lastType = UDFLike.PatternType.COMPLEX;
```
deniskuzZ, line 172: i won't ever enter here

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +108,100 @@ public Checker tryCreate(String pattern) {
...
+      StringBuilder strPattern = new StringBuilder();
+      String simplePattern;
+
+      for (int i = 0; i < length; i++) {
+        char n = pattern.charAt(i);
+        if (n == '_') { // such as "a_b"
+          if (lastChar != '\\') { // such as "a%bc"
+            return null;
+          } else { // such as "abc\%de%"
+            strPattern.append(pattern.substring(beginIndex, i - 1));
```
deniskuzZ, line 155: strPattern.append(pattern, beginIndex, i - 1);

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +108,100 @@ public Checker tryCreate(String pattern) {
...
+            if (lastChar != '\\') {
+              endIndex = length - 1;
+              if (lastType == UDFLike.PatternType.END) { // such as "%abc%"
+                lastType = UDFLike.PatternType.MIDDLE;
+              } else if (lastType != UDFLike.PatternType.CHAINED &&
+                         lastType != UDFLike.PatternType.COMPLEX) {
+                lastType = UDFLike.PatternType.BEGIN; // such as "abc%"
+              }
+            } else { // such as "abc\%"
+              strPattern.append(pattern.substring(beginIndex, i - 1));
```
deniskuzZ, line 184: strPattern.append(pattern, beginIndex, i - 1);

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +108,100 @@ public Checker tryCreate(String pattern) {
...
+              beginIndex = i;
+              endIndex = length;
+            }
+          }
+        }
+        lastChar = n;
+      }
+
+      strPattern.append(pattern.substring(beginIndex, endIndex));
```
deniskuzZ, line 193: strPattern.append(pattern, beginIndex, endIndex);

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +108,100 @@ public Checker tryCreate(String pattern) {
...
+      int endIndex = length;
+      char lastChar = 0;
+      StringBuilder strPattern = new StringBuilder();
+      String simplePattern;
+
+      for (int i = 0; i < length; i++) {
+        char n = pattern.charAt(i);
+        if (n == '_') { // such as "a_b"
+          if (lastChar != '\\') { // such as "a%bc"
+            return null;
```
deniskuzZ, line 153: why return `null` and not 'COMPLEX' as in `UDFLike`?

deniskuzZ: please add the. test for `\\_`

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -140,4 +108,100 @@ public Checker tryCreate(String pattern) {
...
+            beginIndex = 1;
+          } else if (i < length - 1) {
+            if (lastChar != '\\') { // such as "a%bc"
+              lastType = UDFLike.PatternType.CHAINED;
+            } else if (lastChar == '\\') { // such as "a\%bc"
+              strPattern.append(pattern.substring(beginIndex, i - 1));
+              beginIndex = i;
+              if (lastType == UDFLike.PatternType.CHAINED) {
+                lastType = UDFLike.PatternType.COMPLEX;
+              }
```
deniskuzZ, line 170: should we return here?

deniskuzZ:
@ryukobayashi, found bunch of issues, please check below snippet
````
for (int i = 0; i < length; i++) {
        char n = pattern.charAt(i);
        if (n == '_') { // such as "a_b"
          if (lastChar != '\\') { // such as "a_bc"
            return null;
          } else { // such as "abc\%de%"
            strPattern.append(pattern, beginIndex, i - 1);
            beginIndex = i;
          }
        } else if (n == '%') {
          if (i == 0) { // such as "%abc"
            lastType = UDFLike.PatternType.END;
            beginIndex = 1;
          
          } else if (i < length - 1) {
            if (lastChar != '\\') { // such as "a%bc"
              lastType = UDFLike.PatternType.CHAINED;
            } else { // such as "a\%bc"
              strPattern.append(pattern, beginIndex, i - 1);
              beginIndex = i;
            }
          } else {
            if (lastChar != '\\') {
              endIndex = length - 1;
              if (lastType == UDFLike.PatternType.END) { // such as "%abc%"
                lastType = UDFLike.PatternType.MIDDLE;
              } else if (lastType != UDFLike.PatternType.CHAINED) {
                lastType = UDFLike.PatternType.BEGIN; // such as "abc%"
              }
            } else { // such as "abc\%"
              strPattern.append(pattern, beginIndex, i - 1);
              beginIndex = i;
              endIndex = length;
            }
          }
        }
        lastChar = n;
      }
      
      strPattern.append(pattern, beginIndex, endIndex);
      simplePattern = (lastType == UDFLike.PatternType.CHAINED) ? pattern : strPattern.toString();
      
      return (type == lastType) ? simplePattern : null;
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java (hunk 6 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java:152]: Yes, will fix.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java (hunk 7 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java:154]: Yes, will fix.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java (hunk 8 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java:172]: Good point, will fix.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java (hunk 9 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java:155]: will fix.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java (hunk 10 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java:184]: will fix.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java (hunk 11 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java:193]: will fix.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java (hunk 12 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java:153]: will do.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java (hunk 5 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java:149]: I will check and try this.

**Comment by ryukobayashi** [comment]
@deniskuzZ 
> found a bunch of issues, please check if we can replace with the below snippet

I will try it.

> looks like we get Complex type only when "_" is present, is that correct?

Yes, it hasn't changed from the original logic.

**Comment by ryukobayashi** [comment]
@deniskuzZ I have fixed it as per your suggestion. But, we did not integrate it with UDFLike because it would require major fixes to the logic. Please let me know if you think it should be fixed.

**Comment by deniskuzZ** [comment]
> @deniskuzZ I have fixed it as per your suggestion. But, we did not integrate it with UDFLike because it would require major fixes to the logic. Please let me know if you think it should be fixed.

hi @ryukobayashi, since we directly identify the pattern, why do we need multiple matcher factories, can we have just 1 factory? 
````
public class FilterStringColLikeStringScalar extends AbstractFilterStringColLikeStringScalar {
  private static final long serialVersionUID = 1L;

  private transient final static List<CheckerFactory> checkerFactories = ImmutableList.of(
    pattern -> {
      UDFLikePattern udfLike = UDFLikePattern.matcher(pattern);
      try {
        return udfLike.checker.getConstructor(String.class).newInstance(
          udfLike.format(pattern));
      } catch (Exception e) {
        throw new IllegalArgumentException("unable to initialize Checker");
      }
    });

  public FilterStringColLikeStringScalar() {
    super();
  }

  public FilterStringColLikeStringScalar(int colNum, byte[] likePattern) {
    super(colNum, null);
    super.setPattern(new String(likePattern, StandardCharsets.UTF_8));
  }

  @Override
  protected List<CheckerFactory> getCheckerFactories() {
    return checkerFactories;
  }

  private enum UDFLikePattern {
    BEGIN(BeginChecker.class),
    END(EndChecker.class),
    MIDDLE(MiddleChecker.class),
    NONE(NoneChecker.class),
    CHAINED(ChainedChecker.class),
    COMPLEX(ComplexChecker.class);

    Class<? extends Checker> checker;

    UDFLikePattern(Class<? extends Checker> checker) {
      this.checker = checker;
    }
    
    private static UDFLikePattern matcher(String pattern) {
      UDFLikePattern lastType = NONE;
      int length = pattern.length();
      char lastChar = 0;

      for (int i = 0; i < length; i++) {
        char n = pattern.charAt(i);
        if (n == '_' && lastChar != '\\') { // such as "a_bc"
          return COMPLEX;
        } else if (n == '%') {
          if (i == 0) { // such as "%abc"
            lastType = END;
          } else if (i < length - 1) {
            if (lastChar != '\\') { // such as "a%bc"
              lastType = CHAINED;
            }
          } else {
            if (lastChar != '\\') {
              if (lastType == END) { // such as "%abc%"
                lastType = MIDDLE;
              } else if (lastType != CHAINED) {
                lastType = BEGIN; // such as "abc%"
              }
            }
          }
        }
        lastChar = n;
      }
      return lastType;
    }

    private String format(String pattern) {
      int startIndex = 0;
      int endIndex = pattern.length();

      switch (this) {
        case BEGIN:
          endIndex--;
          break;
        case END:
          startIndex++;
          break;
        case MIDDLE:
          startIndex++;
          endIndex--;
          break;
        case COMPLEX:
          return "^" + UDFLike.likePatternToRegExp(pattern) + "$";
      }

      return pattern.substring(startIndex, endIndex);
    }
  }
}
````

**Comment by ryukobayashi** [comment]
@deniskuzZ This is original and I don't know why multiple matcher factories are needed. As you opinion, one factory should be no problem. So, I fixed.

**Comment by deniskuzZ** [comment]
> @deniskuzZ This is original and I don't know why multiple matcher factories are needed. As you opinion, one factory should be no problem. So, I fixed.

👍 nice,
1 last thing, I forgot to post it, could we please override `format` method per enum
````
private enum UDFLikePattern {
    BEGIN(BeginChecker.class) {
      @Override
      String format(String pattern) {
        return pattern.substring(0, pattern.length() - 1);
      }
    },
    END(EndChecker.class) {
      @Override
      String format(String pattern) {
        return pattern.substring(1);
      }
    },
    MIDDLE(MiddleChecker.class) {
      @Override
      String format(String pattern) {
        return pattern.substring(1, pattern.length() - 1);
      }
    },
    COMPLEX(ComplexChecker.class) {
      @Override
      String format(String pattern) {
        return "^" + UDFLike.likePatternToRegExp(pattern) + "$";
      }
    },
    CHAINED(ChainedChecker.class),
    NONE(NoneChecker.class);

    Class<? extends Checker> checker;

    UDFLikePattern(Class<? extends Checker> checker) {
      this.checker = checker;
    }

    String format(String pattern) {
      return pattern;
    }
````

File: ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java [hunk 14]
```diff
@@ -45,11 +45,12 @@ public class UDFLike extends UDF {
 
   // Doing characters comparison directly instead of regular expression
   // matching for simple patterns like "%abc%".
-  private enum PatternType {
+  public enum PatternType {
```
deniskuzZ, line 48: please revert changes in that class

ryukobayashi: Got it. I will fix each.

ryukobayashi:
@deniskuzZ Thanks for review. I fixed.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/AbstractFilterStringColLikeStringScalar.java
```diff
@@ -226,7 +226,7 @@ protected interface CheckerFactory {
   protected static final class NoneChecker implements Checker {
     final byte [] byteSub;
 
-    NoneChecker(String pattern) {
+    public NoneChecker(String pattern) {
```
deniskuzZ, line 229: @ryukobayashi, you can revert this. as well, no need for public modifier here

deniskuzZ, line 229: @ryukobayashi, not a big deal, but that is not required.

deniskuzZ (approved):
LGTM +1, pending tests

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java
```diff
@@ -51,93 +53,83 @@ public FilterStringColLikeStringScalar(int colNum, byte[] likePattern) {
...
-  }
 
-  /**
-   * Accepts any LIKE patterns and creates corresponding checkers.
-   */
-  private static class ComplexCheckerFactory implements CheckerFactory {
-    public Checker tryCreate(String pattern) {
-      // anchor the pattern to the start:end of the whole string.
-      return new ComplexChecker("^" + UDFLike.likePatternToRegExp(pattern) + "$");
+    public String format(String pattern) {
```
deniskuzZ, line 131: no need for `public`, package-private is enough

ryukobayashi: I will fix. Then run the test again.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/AbstractFilterStringColLikeStringScalar.java
ryukobayashi: @deniskuzZ If it is not `public`, an access violation will result in an exception.
https://github.com/apache/hive/blob/ceafcc411fe3471019a05f2509b5d7a6b242f371/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/FilterStringColLikeStringScalar.java#L38-L39

deniskuzZ: oh yes, sorry about that

ryukobayashi:
@deniskuzZ Thanks! test passed.

PR #5000, opened 2024-01-11
HIVE-27995: Fix inconsistent behavior of LOAD DATA command for partitoned and non-partitioned tables

shivjha30:
Earlier, the code flows for partitioned and non-partitioned tables were different. The partitioned tables skipped constraints checks before submitting the job for execution. This Pull request ensures that both, the partitioned and non-partitioned tables go through the constraints validations in `applyConstraintsAndGetFiles` function.

For partitioned tables, while executing LOAD DATA/ LOAD DATA LOCAL commands, the check for file existence is not executed on HiveServer2, and this in turn throws `java.io.FileNotFoundException` during Runtime once the job is launched. 
This PR prevents such cases at compile time.

No

No

The test cases already exist. The error messages prompted back to the user are now consistent if the file is not found at HiveServer2

Load Data Error (Non Partitioned Tables)
Load Data Error

File Not Found Exception (Partitioned Tables)
Load

Fixed: For partitioned tables
Load (1)

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestJdbcDriver2.java
+++ b/itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestJdbcDriver2.java
@@ -3428,4 +3428,26 @@ public class TestJdbcDriver2 {
       stmt.close();
     }
   }
+  /**
+   * These test methods validate the error handling when attempting to load data from a non-existent file into different types of Hive tables.
+   * Each test creates a specific type of table ( dynamically partitioned, or bucketed), and then attempts to load data from a non-existent file.
+   * The tests pass if an exception is thrown with a message containing the string "Invalid path".
+   */
+  @Test
+  public void testLoadDataNegativeForDynamicPartition() throws Exception {
+    HiveStatement stmt = (HiveStatement) con.createStatement();
+    try {
+      stmt.execute("drop table if exists T");
+      stmt.execute("create table T (a int, b int) partitioned by (p int) stored as orc");
+      try {
+        stmt.execute("load data local inpath '/path/to/nonexistent/file.txt' into table T");
+        fail("Expected an exception to be thrown");
+      } catch (Exception e) {
+        assertTrue("load data inpath",
+                e.getMessage() != null && e.getMessage().contains("Invalid path"));
+      }
+    } finally {
+      stmt.close();
+    }
+  }
 }
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
@@ -155,7 +155,12 @@ public class LoadSemanticAnalyzer extends SemanticAnalyzer {
     return new URI(fromScheme, fromAuthority, path, null, null);
   }
 
-  private List<FileStatus> applyConstraintsAndGetFiles(URI fromURI, Table table) throws SemanticException {
+  /**
+   * Verifies the file path existence. If the path is a directory, then it is
+   * verified to not be a glob pattern. If the path is a file, then it is
+   * verified to be a valid file name.
+   */
+  private FileStatus[] verifyFilePathExists(URI fromURI) throws SemanticException {
 
     FileStatus[] srcs;
 
@@ -166,7 +171,6 @@ public class LoadSemanticAnalyzer extends SemanticAnalyzer {
           ErrorMsg.ILLEGAL_PATH.getMsg(), fromTree,
           "Source file system should be \"file\" if \"local\" is specified"));
     }
-
     try {
       FileSystem fileSystem = FileSystem.get(fromURI, conf);
       srcs = matchFilesOrDir(fileSystem, new Path(fromURI));
@@ -175,6 +179,15 @@ public class LoadSemanticAnalyzer extends SemanticAnalyzer {
             ErrorMsg.INVALID_PATH.getMsg(), fromTree,
             "No files matching path " + fromURI));
       }
+    } catch (IOException e) {
+      throw new SemanticException(ASTErrorUtils.getMsg(
+              ErrorMsg.INVALID_PATH.getMsg(), fromTree), e);
+    }
+    return srcs;
+  }
+  private List<FileStatus> applyConstraintsAndGetFiles(URI fromURI, Table table, FileStatus[] srcs)
+          throws SemanticException {
+    try {
 
       for (FileStatus oneSrc : srcs) {
         if (oneSrc.isDir()) {
@@ -182,6 +195,7 @@ public class LoadSemanticAnalyzer extends SemanticAnalyzer {
           return null;
         }
       }
+      FileSystem fileSystem = FileSystem.get(fromURI, conf);
       AcidUtils.validateAcidFiles(table, srcs, fileSystem);
       // Do another loop if table is bucketed
       List<String> bucketCols = table.getBucketCols();
@@ -284,6 +298,8 @@ public class LoadSemanticAnalyzer extends SemanticAnalyzer {
       throw new SemanticException(ASTErrorUtils.getMsg(
           ErrorMsg.INVALID_PATH.getMsg(), fromTree, e.getMessage()), e);
     }
+    // validate the arguments
+    FileStatus[] srcs = verifyFilePathExists(fromURI);
 
     // initialize destination table/partition
     TableSpec ts = new TableSpec(db, conf, (ASTNode) tableTree);
@@ -338,7 +354,7 @@ public class LoadSemanticAnalyzer extends SemanticAnalyzer {
     }
 
     // make sure the arguments make sense
-    List<FileStatus> files = applyConstraintsAndGetFiles(fromURI, ts.tableHandle);
+    List<FileStatus> files = applyConstraintsAndGetFiles(fromURI, ts.tableHandle, srcs);
     if (queryReWritten) {
       return;
     }
--- a/ql/src/test/org/apache/hadoop/hive/ql/TestTxnLoadData.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/TestTxnLoadData.java
@@ -391,16 +391,16 @@ public class TestTxnLoadData extends TxnCommandsBaseForTests {
    */
   @Test
   public void testValidations() throws Exception {
-    runStatementOnDriver("drop table if exists T");
-    runStatementOnDriver("drop table if exists Tstage");
-    runStatementOnDriver("create table T (a int, b int) clustered by (a) into 2 buckets stored as orc tblproperties('transactional'='true')");
+    runStatementOnDriver("drop table if exists t");
+    runStatementOnDriver("drop table if exists tstage");
+    runStatementOnDriver("create table t (a int, b int) clustered by (a) into 2 buckets stored as orc tblproperties('transactional'='true')");
     File createdFile= folder.newFile("myfile.txt");
     FileUtils.writeStringToFile(createdFile, "hello world");
-    runStatementOnDriver("create table Tstage (a int, b int) stored as orc tblproperties('transactional'='false')");
+    runStatementOnDriver("create table tstage (a int, b int) stored as orc tblproperties('transactional'='false')");
     //this creates an ORC data file with correct schema under table root
-    runStatementOnDriver("insert into Tstage values(1,2),(3,4)");
+    runStatementOnDriver("insert into tstage values(1,2),(3,4)");
     // This will work with the new support of rewriting load into IAS.
-    runStatementOnDriver("load data local inpath '" + getWarehouseDir() + "/Tstage' into table T");
+    runStatementOnDriver("load data local inpath '" + getWarehouseDir() + "/tstage' into table t");
   }
 
   private void checkExpected(List<String> rs, String[][] expected, String msg) {
--- a/ql/src/test/queries/clientnegative/load_data_partition.q
+++ b/ql/src/test/queries/clientnegative/load_data_partition.q
@@ -1,0 +1,6 @@
+drop table if exists validate_load_data;
+CREATE TABLE validate_load_data(key int, value string) partitioned by (hr int) STORED AS TEXTFILE;
+LOAD DATA INPATH '/tmp/kv1.txt' INTO TABLE validate_load_data;
+SELECT * FROM validate_load_data;
+DROP TABLE validate_load_data;
+``````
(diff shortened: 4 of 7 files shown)


ayushtkn (commented):
you need to add a test, there are NegativeCLIDrivers which can be used.

There are bunch of return statements below the current line & the place where it is there right now, So, if that returned was called earlier, this `` applyConstraintsAndGetFiles(fromURI, ts.tableHandle);`` must not have been called, but now you would be calling that, right?. So, that is some additional cost just for error message?

shivjha30:
@ayushtkn Thanks for the review, i have added the test case as mentioned above.please review once

shivjha30:
There is a test failure in Jenkins, i ran the test in in local and that test is passing which can be viewed in the below screenshot:
image

itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestJdbcDriver2.java
```diff
@@ -3422,4 +3422,26 @@ public void testHeaderFooterNonTextFiles() throws Exception {
       stmt.close();
     }
   }
+  /**
```
chinnaraolalam, line 3425: add all the use cases like partitioned , non-partitioned and load data use cases.

chinnaraolalam:
This scenario will fail before fixing this issue at job execution and now it is validated at Hiveserver and it will throw SemanticExceptoin to the client. What is the exception received by client before fixing this. I think exception will be different, can you add what exception client is receiving before fixing this issue.

ql/src/test/org/apache/hadoop/hive/ql/TestTxnLoadData.java
```diff
@@ -394,16 +394,16 @@ public void loadDataPartitioned() throws Exception {
    */
   @Test
   public void testValidations() throws Exception {
-    runStatementOnDriver("drop table if exists T");
-    runStatementOnDriver("drop table if exists Tstage");
```
chinnaraolalam, line 398: Why these changes are needed

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
```diff
@@ -344,7 +358,7 @@ private void analyzeLoad(ASTNode ast) throws SemanticException {
     }
 
     // make sure the arguments make sense
-    List<FileStatus> files = applyConstraintsAndGetFiles(fromURI, ts.tableHandle);
+    List<FileStatus> files = applyConstraintsAndGetFiles(fromURI, ts.tableHandle, srcs);
```
chinnaraolalam, line 361: What is behaviour of below scenarios before and after fix and what exception it is throwing

Partitioned - Local path
Partitioned - HDFS path
Non-Partitioned - Local path
Non-Partitioned - HDFS path

shivjha30: The changes were introduced only for Partition tables, there is no change in behaviour for Non Partition.
For partition tables for both local and hdfs path , Before the fix the exception thrown was File Not Found from Node Manager. After the fix the exception will be thrown from Hive Server itself, and the exception now would be  Semantic Exception.

ql/src/test/org/apache/hadoop/hive/ql/TestTxnLoadData.java
shivjha30: when we create a table **Tstage**, in warehouse directory the table will be created in small case letters like "tstage". While we are performing Load Data we are appending warehousedir + "Tstage"  in https://github.com/apache/hive/blob/0bc624ade3020826b4e7101b7ea4789be7e8585b/ql/src/test/org/apache/hadoop/hive/ql/TestTxnLoadData.java#L406

As Linux is case sensitive, it couldnt find the path of warehousedir + "Tstage" but there is a path of warehousedir + "tstage" exists.

Previously, the file path wasn't being validated in the Hive server. Now, after implementing a fix, the test has failed because the Hive server is now validating whether the file path exists or not.

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
```diff
@@ -181,7 +184,16 @@ private List<FileStatus> applyConstraintsAndGetFiles(URI fromURI, Table table) t
...
+    } catch (IOException e) {
+      throw new SemanticException(ASTErrorUtils.getMsg(
+              ErrorMsg.INVALID_PATH.getMsg(), fromTree), e);
+    }
+    return srcs;
+  }
+    private List<FileStatus> applyConstraintsAndGetFiles(URI fromURI, Table table, FileStatus[] srcs)
+            throws SemanticException {
+    try {
+      FileSystem fileSystem = FileSystem.get(fromURI, conf);
```
deniskuzZ, line 196: why do you need `fileSystem` here?

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
```diff
@@ -325,6 +337,8 @@ private void analyzeLoad(ASTNode ast) throws SemanticException {
     if(ts.tableHandle.isStoredAsSubDirectories()) {
       throw new SemanticException(ErrorMsg.LOAD_INTO_STORED_AS_DIR.getMsg());
     }
+    // validate the arguments
+    FileStatus[] srcs = verifyFilePathExists(fromURI);
```
deniskuzZ, line 341: why don't we validate `fromURI` sooner?

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
shivjha30: Down the line in the applyConstraintsAndGetFiles, filesystem is getting used in "AcidUtils.validateAcidFiles(table, srcs, fileSystem);"

shivjha30: @deniskuzZ Can you look at the changes and review it ?

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
chinnaraolalam: So earlier in partition case, even though file not exists query was successful. Can you check why it is not throwing earlier. 

After fix it was validated at server level and it will throw error for non existing file paths. So if any such cases are there in existing applications, those queries need to corrected.

shivjha30: In the partition case if the file doesn't exist then number of splits will be zero, as the splits are zero the mappers also will be equal to zero. As there are no mappers the CombineHiveInputformat will not be called.

Now after the fix we would be validating at the server level, thus in all such cases in the existing applications, those queries needs to be corrected.

ql/src/test/results/clientnegative/load_wrong_noof_part.q.out
```diff
@@ -6,4 +6,4 @@ POSTHOOK: query: CREATE TABLE loadpart1(a STRING, b STRING) PARTITIONED BY (ds S
 POSTHOOK: type: CREATETABLE
 POSTHOOK: Output: database:default
 POSTHOOK: Output: default@loadpart1
-FAILED: SemanticException [Error 10006]: Line 2:82 Partition not found ''2009-05-05''
+#### A masked pattern was here ####
```
chinnaraolalam, line 9: How it is related to this change.

shivjha30: In HIVE-5601, the file path in the q file was given an incorrect path, although the purpose of the test was to check for number of  partitions. With this PR change that wrong path would throw the semantic exception even before checking the number of partitions.
Instead of aiming for the file path and raising a semantic exception, I have updated the q file with the right file path, fulfilling the task of merely verifying the number of partitions.

ql/src/test/queries/clientnegative/load_data_partition.q
```diff
@@ -0,0 +1,7 @@
+--load data if the file does not exist
+
+CREATE TABLE validate_load_data(key int, value string) partitioned by (hr int) STORED AS TEXTFILE;
+LOAD DATA INPATH 'hdfs:///validateload/filedoesnt.txt' INTO TABLE validate_load_data partition (hr);
+SELECT * FROM validate_load_data;
+DROP TABLE validate_load_data;
```
chinnaraolalam, line 6: Can you check this, without fix also it is throwing error.

shivjha30: @chinnaraolalam the following scenario is now handled, can you review this once

itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestJdbcDriver2.java
```diff
@@ -3422,4 +3422,51 @@ public void testHeaderFooterNonTextFiles() throws Exception {
...
+   * Each test creates a specific type of table ( dynamically partitioned, or bucketed), and then attempts to load data from a non-existent file.
+   * The tests pass if an exception is thrown with a message containing the string "Invalid path".
+   */
+  @Test
+  public void testLoadDataNegativeForDynamicPartition() throws Exception {
+    HiveStatement stmt = (HiveStatement) con.createStatement();
+    try {
+      stmt.execute("drop table if exists T");
+      stmt.execute("create table T (a int, b int) partitioned by (p int) stored as orc");
+      stmt.execute("set hive.exec.dynamic.partition=true");
```
chinnaraolalam, line 3436: Why we need to set these parameters explicitly, we can test with default values only. All the test cases have these parameters every where, we can test with default values unless any explicit case we need test.

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
chinnaraolalam: > For partition tables for local ,
In Local path case, If file would be present at hs2 what is the behaviour after fix

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
chinnaraolalam: Move this to above the line where it is used.

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
shivjha30: For partition case if the the file does exist at hs2, and the container looks for the file at other machines  it can throw the execution error from hdfs end, that scenario is not covered in this PR

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
shivjha30: Comment is updated and resolved

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
shivjha30: This comment  is update and resolved

chinnaraolalam:
Overall changes looks good to me, But locally when i run newly added test case it falling

`[ERROR]   TestNegativeLlapCliDriver.testCliDriver:59 Client Execution succeeded but contained differences (error code = 1) after executing load_data_partition.q 
15,21c15
< FAILED: SemanticException Line 2:17 Invalid path ''hdfs://### HDFS PATH ###'': No files matching path hdfs://### HDFS PATH ###
< FAILED: AssertionError java.lang.AssertionError: Client Execution succeeded but contained differences (error code = 1) after executing load_data_partition.q 
< 15c15
< < FAILED: SemanticException Line 2:17 Invalid path ''hdfs://### HDFS PATH ###'': No files matching path hdfs://### HDFS PATH ###
`

ql/src/test/results/clientnegative/load_data_partition.q.out
```diff
@@ -0,0 +1,15 @@
...
+POSTHOOK: Output: database:default
+PREHOOK: query: CREATE TABLE validate_load_data(key int, value string) partitioned by (hr int) STORED AS TEXTFILE
+PREHOOK: type: CREATETABLE
+PREHOOK: Output: database:default
+PREHOOK: Output: default@validate_load_data
+POSTHOOK: query: CREATE TABLE validate_load_data(key int, value string) partitioned by (hr int) STORED AS TEXTFILE
+POSTHOOK: type: CREATETABLE
+POSTHOOK: Output: database:default
+POSTHOOK: Output: default@validate_load_data
+#### A masked pattern was here ####
```
chinnaraolalam, line 15: Here there is no error,  I think it is failing before the expected scenario.

shivjha30: After masking the failure scenario is coming like this, before that 
the error is coming as expected 
"FAILED: SemanticException Line 2:17 Invalid path ''/tmp/kv1.txt'': No files matching path hdfs://localhost:65046/tmp/kv1.txt"

deniskuzZ (approved):
LGTM +1

ql/src/test/org/apache/hadoop/hive/ql/TestTxnLoadData.java
deniskuzZ: @shivjha30, you should lowercased just the `inpath` dir

shivjha30: The problem at hand pertains to the case sensitivity of the table name, which is currently in uppercase. Despite modifying the directory in the input path to lowercase, the case sensitivity for the table name continues to persist.

chinnaraolalam:
+1  Thanks for the review Ayush Saxena, Denys Kuzmenko

ql/src/test/org/apache/hadoop/hive/ql/TestTxnLoadData.java
deniskuzZ: ok,  wasn't the issue only with the inpath for Load Data? 
````
load data local inpath '" + getWarehouseDir() + "/tstage' into table T
````

chinnaraolalam: Yes lowercased inpath dir should work, but need to comment why we are lowering the table name from camel notation without using as it is, it will help for further extension of the test case.  IIUC test case scenario wise camel notation of table name is not needed.

All the test cases in this class uses camel notation, only to make it consistent I agree with denis maintain camel notation of table name here also and lowercased only inpath dir.

PR #5005, opened 2024-01-15
HIVE-27489: HPL/SQL does not support table aliases on column names in loops.

mdayakar:
If either table name or table alias is used in column name in the loops then selecting only column name to retrieve the data from cursor by assuming _<table/alias_name>.<column_name>_ is the pattern.

To fix the issue HIVE-27489.

No

No

Using Unit test code.
_mvn test -Dtest=TestHplSqlViaBeeLine#testTableAliasInColumnName -pl itests/hive-unit -Pitests_

### DIFF
--- a/hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
+++ b/hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
@@ -957,7 +957,11 @@ public class Stmt {
       int cols = query.columnCount();
       Row row = new Row();
       for (int i = 0; i < cols; i++) {
-        row.addColumnDefinition(query.metadata().columnName(i), query.metadata().columnTypeName(i));
+        String columnName = query.metadata().columnName(i);
+        if (columnName.contains(".")) {
+          columnName = columnName.substring(columnName.lastIndexOf('.') + 1);
+        }
+        row.addColumnDefinition(columnName, query.metadata().columnTypeName(i));
       }
       Var var = new Var(cursor, row);
       exec.addVariable(var);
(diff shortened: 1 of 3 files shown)


kasakrisz (approved):
+1 LGTM, pending tests

hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
```diff
@@ -957,7 +957,11 @@ public Integer forCursor(HplsqlParser.For_cursor_stmtContext ctx) {
       int cols = query.columnCount();
       Row row = new Row();
       for (int i = 0; i < cols; i++) {
-        row.addColumnDefinition(query.metadata().columnName(i), query.metadata().columnTypeName(i));
+        String columnName = query.metadata().columnName(i);
+        if (columnName.contains(".")) {
+          columnName = columnName.substring(columnName.lastIndexOf('.') + 1);
```
aturoczy, line 962: Just try to be a stupid engineer?
How about table like

col1.
col2...
col3.tab1.nice

I would add negative tests as well.

mdayakar: @aturoczy we can not create a table or a column with .(dot) in its name so here if .(dot) is there in the name means its _<table_name>.<column_name>_.
```
0: jdbc:hive2://localhost> create table tab.emp(id int, name string);
Error: Error while compiling statement: FAILED: SemanticException [Error 10072]: Database does not exist: tab (state=42000,code=10072)
0: jdbc:hive2://localhost> create table tab(id.new int, name.first string);
Error: Error while compiling statement: FAILED: ParseException line 1:19 cannot recognize input near '.' 'new' 'int' in column type (state=42000,code=40000)
0: jdbc:hive2://localhost>
```

PR #5006, opened 2024-01-15
HIVE-27993 : Netty4 ShuffleHandler: should use 1 boss thread

AnmolSun:
Similar to Tez [TEZ-4394](https://issues.apache.org/jira/browse/TEZ-4394), and [Hadoop](https://github.com/apache/hadoop/blob/7a0903b7436e7f3fb4f07c40bed2218a2653ba38/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-shuffle/src/main/java/org/apache/hadoop/mapred/ShuffleHandler.java#L353), [Hive LLAP ShuffleHandler](https://github.com/apache/hive/blob/cc3f4d1d3b766f444606a4b06341eb8c4ae8148c/llap-server/src/java/org/apache/hadoop/hive/llap/shufflehandler/ShuffleHandler.java#L299) should also use a single thread for boss thread pool.


zhangbutao (commented):
Can you rebase this change to get a green CI?

AnmolSun:
Thanks for helping with review @zhangbutao 
Rebased and now we have a green run!

zhangbutao (approved):
LGTM

zhangbutao:
@abstractdog Could you please take a look as well? Thx.

abstractdog (approved):
+1

PR #5008, opened 2024-01-16
HIVE-28002: Make hive.metastore.batch.retrieve.max property to be read from hiveconf in HS2

vikramahuja1001:
The property hive.metastore.batch.retrieve.max is used by both the client(Hs2) and the HMS when fetching objects(tables/partitions) in batches. This property is present in both hiveConf.java as well as metastoreConf.java. For some use cases in the client(Hs2), this value is referred from the metastoreConf. Changing those use cases to HiveConf.java to have a set demarkation between HS2 and HMS use cases for this property.


aturoczy (commented):
I do not understanding why it is better to read this config from HiveConfig than for MetaStore config? Sorry, but I do not understand the benefits of this change. Could you please ellaborate why this change has a benefit?

deniskuzZ:
-1

vikramahuja1001:
The main motivation behind this PR was to increase readability wrt HS2 accessing only HiveConf and HMS accessing only MetastoreConf to have a clear demarkation.

I will close this PR wrt the above comments.

PR #5014, opened 2024-01-18
HIVE-27741 handle Invalid timezone value in to_utc_timestamp() UDF

zratkai:
Change-Id: I68ee1783614b4b74093df0dddb613fa860ed202a

The to_utc_timestamp() UDF used UTC when the given timezone was invalid. It throws exception in this case to avoid data inconsistency.

To avoid data inconsistency in case of invalid timezone.

Yes. The  to_utc_timestamp() UDF will throw exception in case of invalid timezone.

No.

Unit test.


aturoczy (approved):
The change is fine. Need a test re-run and green output.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -88,18 +90,15 @@ public Object evaluate(DeferredObject[] arguments) throws HiveException {
...
-
-    TimeZone fromTz;
-    TimeZone toTz;
-    if (invert()) {
-      fromTz = timezone;
-      toTz = tzUTC;
-    } else {
-      fromTz = tzUTC;
-      toTz = timezone;
+    if(ZoneInfo.getTimeZone(tzStr)==null){
```
okumin, line 93: It would be nice if we added spaces between parentheses/operators.

```
if (ZoneInfo.getTimeZone(tzStr) == null) {
```

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -34,6 +35,7 @@
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.PrimitiveObjectInspectorFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import sun.util.calendar.ZoneInfo;
```
okumin, line 38: I'm curious what are the advantages of this package. In my first impression, I would use `java.time.ZoneId` because it sounds more standard and consistent with the TimeZone API.

zratkai: TimeZone.getTimeZone uses ZoneInfo to get the proper timezone. But if the input is not a valid timezone, then it gives back the default one, and this is what this fix is about. Instead of the default it will throw exception to avoid data inconsistency.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
zratkai: Fixed.

zratkai (commented):
Please check the comments and fixes.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.hive.ql.udf.generic;
 
+import java.util.Arrays;
```
simhadri-g, line 20: Unused import, please remove.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -113,6 +112,14 @@ public Object evaluate(DeferredObject[] arguments) throws HiveException {
     return result;
   }
 
+  protected TimeZone getToTimeZone(TimeZone timezone) {
+    return timezone;
+  }
+
+  protected TimeZone getFromTimeZone(TimeZone timezone) {
```
simhadri-g, line 119: do we need to pass timezone in the arguments? We are returning UTC.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
simhadri-g: According to this : https://stackoverflow.com/a/55149856/7279976 
`sun.util.calendar.ZoneInfo; ` seems to be  replaced by [java.time.zone.ZoneRules](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/zone/ZoneRules.html) 

Can you please check using ZoneRules ?

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFToUtcTimestamp.java
```diff
@@ -44,7 +46,11 @@ public String getName() {
   }
 
   @Override
-  protected boolean invert() {
-    return true;
+  protected TimeZone getToTimeZone(TimeZone timezone) {
+    return tzUTC;
```
simhadri-g, line 50: do we need to pass timezone in the arguments? We are returning UTC here.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -88,18 +90,15 @@ public Object evaluate(DeferredObject[] arguments) throws HiveException {
...
-
-    TimeZone fromTz;
-    TimeZone toTz;
-    if (invert()) {
-      fromTz = timezone;
-      toTz = tzUTC;
-    } else {
-      fromTz = tzUTC;
-      toTz = timezone;
+    if (ZoneInfo.getTimeZone(tzStr)==null  && !tzStr.startsWith("GMT")) {
```
simhadri-g, line 93: nit: checkstyle

zratkai: Fixed.

zratkai:
@okumin @simhadri-g could you please review?

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
zratkai: Refactored. Thanks for the input!

zratkai (commented):
Refactored to use ZoneId instead of TimeZone.

zratkai:
@okumin @simhadri-g could you please review? Refactored to use ZoneId instead of TimeZone.

simhadri-g (approved):
LGTM +1

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -88,31 +88,42 @@ public Object evaluate(DeferredObject[] arguments) throws HiveException {
...
-    } else {
-      fromTz = tzUTC;
-      toTz = timezone;
-    }
 
-    // inputTs is the year/month/day/hour/minute/second in the local timezone.
-    // For this UDF we want it in the timezone represented by fromTz
-    TimestampTZ fromTs = TimestampTZUtil.parse(inputTs.toString(), fromTz.toZoneId());
+    ZoneId zoneId;
+    try{
```
okumin, line 93: We may prefer to have a space before/after a parenthesis.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -88,31 +88,42 @@ public Object evaluate(DeferredObject[] arguments) throws HiveException {
...
+      zoneId = ZoneId.of(tzStr);
+    }catch (Exception exception) {
```
okumin, line 95: We may prefer to have a space before/after a parenthesis.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -88,31 +88,42 @@ public Object evaluate(DeferredObject[] arguments) throws HiveException {
...
+      String timeZoneId = ZoneId.SHORT_IDS.get(tzStr);
+      if(timeZoneId == null){
```
okumin, line 97: We may prefer to have a space before/after a parenthesis.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -88,31 +88,42 @@ public Object evaluate(DeferredObject[] arguments) throws HiveException {
...
-      fromTz = tzUTC;
-      toTz = timezone;
-    }
 
-    // inputTs is the year/month/day/hour/minute/second in the local timezone.
-    // For this UDF we want it in the timezone represented by fromTz
-    TimestampTZ fromTs = TimestampTZUtil.parse(inputTs.toString(), fromTz.toZoneId());
+    ZoneId zoneId;
+    try{
+      zoneId = ZoneId.of(tzStr);
```
okumin, line 94: `ZoneID.of(tzStr, ZoneId.SHORT_IDS` could be shorthand. If no, please ignore this comment.
https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#of-java.lang.String-java.util.Map-

okumin (approved):
+1 with some minor cosmetic comments.
This could be an incompatible change. But I understand the current behavior is surprising and it sounds better to throw an error(or put NULL at least). The final judgment will be done by a committer.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
zratkai: Done.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
zratkai: Good catch, thanks!

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
zratkai: Done.

zratkai:
@okumin @simhadri-g could you please review this?

okumin:
Looks good to me. Note that you finally need approval from a committer

okumin (approved):

ql/src/test/queries/clientpositive/udf_from_utc_timestamp.q
```diff
@@ -8,8 +8,6 @@ from_utc_timestamp('2012-02-11 04:30:00', 'PST'),
 from_utc_timestamp('2012-02-11 04:30:00', 'Europe/Moscow'),
 from_utc_timestamp('2012-02-11 04:30:00', 'GMT+8'),
 from_utc_timestamp('2012-02-11 04:30:00', 'GMT'),
-from_utc_timestamp('2012-02-11 04:30:00', ''),
-from_utc_timestamp('2012-02-11 04:30:00', '---'),
```
deniskuzZ, line 12: was that some special case??

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFFromUtcTimestamp.java
```diff
@@ -86,33 +86,33 @@ public Object evaluate(DeferredObject[] arguments) throws HiveException {
...
-      toTz = tzUTC;
-    } else {
-      fromTz = tzUTC;
-      toTz = timezone;
-    }
-
-    // inputTs is the year/month/day/hour/minute/second in the local timezone.
-    // For this UDF we want it in the timezone represented by fromTz
-    TimestampTZ fromTs = TimestampTZUtil.parse(inputTs.toString(), fromTz.toZoneId());
+    ZoneId zoneId = ZoneId.of(tzStr,ZoneId.SHORT_IDS);
```
deniskuzZ, line 90: nit: space

deniskuzZ (commented):
LGTM, minor code-style comments

RyanEugene (approved):

ql/src/test/queries/clientpositive/udf_from_utc_timestamp.q
zratkai: This are invalid formats. Previously in such cases UTC was used, but now instead an exception is thrown:
java.time.DateTimeException: Invalid ID for ZoneOffset, invalid format

But it can not be tested here. It is tested with TestGenericUDFFromUtcTimestamp.java

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFToUtcTimestamp.java
```diff
@@ -44,7 +46,11 @@ public String getName() {
   }
 
   @Override
-  protected boolean invert() {
-    return true;
+  protected ZoneId getToZoneId(ZoneId zoneId) {
+    return this.zoneIdUTC;
+  }
+  @Override
```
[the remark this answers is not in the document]
zratkai: Fixed.

deniskuzZ (approved):
+1, pending tests

PR #5015, opened 2024-01-18
HIVE-28009: Shared work optimizer ignores schema merge setting in case of virtual column difference

kasakrisz:
Extend table scan schema comparing with comparing virtual columns in shared work optimizer.

Shared work optimizer can merge table scan operators scanning the same table with different schema. This feature is controlled by `hive.optimize.shared.work.merge.ts.schema` but the current implementation ignores the value of the flag and always merge.

Only in execution plans.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=sharedwork_virtualcol_schema_merge.q -pl itests/qtest -Pitests
mvn test -Dtest.output.overwrite -Dtest=TestIcebergCliDriver -Dqfile=merge_iceberg_copy_on_write_unpartitioned.q,update_iceberg_copy_on_write_partitioned.q,update_iceberg_copy_on_write_unpartitioned.q -pl itests/qtest-iceberg -Piceberg -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
@@ -649,9 +649,9 @@ public class SharedWorkOptimizer extends Transform {
   }
 
   private static boolean compatibleSchema(TableScanOperator tsOp1, TableScanOperator tsOp2) {
-    return tsOp1.getNeededColumns().equals(tsOp2.getNeededColumns())
-        && tsOp1.getNeededColumnIDs().equals(tsOp2.getNeededColumnIDs())
-        && tsOp1.getConf().getVirtualCols().equals(tsOp2.getConf().getVirtualCols());
+    return Objects.equals(tsOp1.getNeededColumns(), tsOp2.getNeededColumns())
+        && Objects.equals(tsOp1.getNeededColumnIDs(), tsOp2.getNeededColumnIDs())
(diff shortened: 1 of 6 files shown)


deniskuzZ:
@kasakrisz, could we please add a test for this

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -650,7 +650,8 @@ protected void mergeSchema(TableScanOperator discardableTsOp, TableScanOperator
 
   private static boolean compatibleSchema(TableScanOperator tsOp1, TableScanOperator tsOp2) {
     return tsOp1.getNeededColumns().equals(tsOp2.getNeededColumns())
-        && tsOp1.getNeededColumnIDs().equals(tsOp2.getNeededColumnIDs());
+        && tsOp1.getNeededColumnIDs().equals(tsOp2.getNeededColumnIDs())
+        && tsOp1.getConf().getVirtualCols().equals(tsOp2.getConf().getVirtualCols());
```
deniskuzZ, line 654: is it possible to get NULLs here?

kasakrisz: Column pruning always set an empty array but I changed these to use `Obejcts.equals` for safety.

deniskuzZ (approved):
LGTM +1

PR #5022, opened 2024-01-23
HIVE-28019 Fix query type information in proto files for load queries

ramesh0201:
Query Type information for certain load queries are currently tagged as 'QUERY' operation type. As part of this change, load queries will be tagged as 'LOAD' query type.

Query Type information for load queries are not corrected tagged. This information needs to rightly set since it also gets logged to hive proto files, which in turn might get used to generate reports on the kind of queries that are being run.

Yes, log files will contain the query type field for load queries

No

Most of the existing tests results that contain any load queries were modified righty and verified.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
@@ -58,6 +58,7 @@ import org.apache.hadoop.hive.ql.metadata.HiveStorageHandler;
 import org.apache.hadoop.hive.ql.metadata.Partition;
 import org.apache.hadoop.hive.ql.metadata.Table;
 import org.apache.hadoop.hive.ql.plan.BasicStatsWork;
+import org.apache.hadoop.hive.ql.plan.HiveOperation;
 import org.apache.hadoop.hive.ql.plan.LoadTableDesc;
 import org.apache.hadoop.hive.ql.plan.LoadTableDesc.LoadFileType;
 import org.apache.hadoop.hive.ql.plan.MoveWork;
@@ -244,6 +245,7 @@ public class LoadSemanticAnalyzer extends SemanticAnalyzer {
     } else {
       analyzeLoad(ast);
     }
+    queryState.setCommandType(HiveOperation.LOAD);
   }
 
   private void analyzeLoad(ASTNode ast) throws SemanticException {
--- a/iceberg/iceberg-handler/src/test/results/positive/iceberg_load_data.q.out
+++ b/iceberg/iceberg-handler/src/test/results/positive/iceberg_load_data.q.out
@@ -15,11 +15,11 @@ POSTHOOK: type: CREATETABLE
 POSTHOOK: Output: database:default
 POSTHOOK: Output: default@ice_parquet
 PREHOOK: query: explain LOAD DATA LOCAL INPATH '../../data/files/parquet_partition' OVERWRITE INTO TABLE ice_parquet
-PREHOOK: type: QUERY
+PREHOOK: type: LOAD
 PREHOOK: Input: default@ice_parquet__temp_table_for_load_data__
 PREHOOK: Output: default@ice_parquet
 POSTHOOK: query: explain LOAD DATA LOCAL INPATH '../../data/files/parquet_partition' OVERWRITE INTO TABLE ice_parquet
-POSTHOOK: type: QUERY
+POSTHOOK: type: LOAD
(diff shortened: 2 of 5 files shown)


niteshy:
There are a lot of test files changed in the PR. Might be it is easier to verify the code/script that is used to update the test files.

ramesh0201:
> There are a lot of test files changed in the PR. Might be it is easier to verify the code/script that is used to update the test files.

The repo to test the changes are here https://github.infra.cloudera.com/rameshkumar/TestQueryTypes

ramesh0201:
Just FYI:
https://issues.apache.org/jira/browse/HIVE-28128
https://issues.apache.org/jira/browse/HIVE-28129

Above are pending issues related to the query type information

ql/src/java/org/apache/hadoop/hive/ql/QueryState.java
```diff
@@ -88,6 +88,11 @@ public class QueryState {
    */
   private final ReentrantLock resolveConditionalTaskLock = new ReentrantLock(true);
 
+  /**
+   * command type that requires the authorization.
+   */
+  private HiveOperation authorizationCommandType;
+
```
zabetak, line 95: The `HiveOperation` class was explicitly introduced for authorization purposes. Adding a separate field in `QueryState` is confusing and will be hard to maintain. It becomes unclear when/how to update `commandType` and `authorizationCommandType`, why we need both, when and under what context we should use each of those.

jfsii:
I tend to lead towards keeping things as is for subjective things (unless there is a good reason to change).
If we do end up changing the PRE/POST hook - it might make sense to do something like:
console.printInfo("PREHOOK: type: " (isExplain ? "EXPLAIN " : "") + queryState.getCommandType(), false);
So we get both bits of information - and as for the contents of the proto files maybe adding an isExplain? But as @zabetak pointed out it is pretty easy for people to do a prefix check on the query string (and it would be backwards compatible between versions of hive)

Just chiming in with my opinion - not sure what is the "right way" or even if there is a "right way".

ramesh0201:
Thank you @zabetak and @jfsii for the review. I have modified this change to contain only the changes for the load queries. For the explain queries we will still fallback to the previous type information it had.

ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java
```diff
@@ -244,6 +245,7 @@ public void analyzeInternal(ASTNode ast) throws SemanticException {
     } else {
       analyzeLoad(ast);
     }
+    queryState.setCommandType(HiveOperation.LOAD);
```
zabetak, line 248: As far as I see we obtain a `LoadSemanticAnalyzer` from https://github.com/apache/hive/blob/e980c7a5e6ba3adec473a30e3ea0a539a5eae257/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzerFactory.java#L75
and few lines before we set the command type using https://github.com/apache/hive/blob/e980c7a5e6ba3adec473a30e3ea0a539a5eae257/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzerFactory.java#L62
so I suppose initially the type is set correctly to `LOAD`.

Why do need to reset the type here? Who is changing it to `QUERY` and why? Was it intentional/necessary?

ramesh0201: Yes for certain load queries we transform to insert into queries by calling the reparseAndSuperAnalyze method at https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java#L470. 
Example, one of the cases is during the load on the bucketed tables at https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/parse/LoadSemanticAnalyzer.java#L341

Because we transform to insert query which is always tagged as QUERY type. But we may want to preserve the actual query's type in this case, as it is what is provided by the user.

zabetak: The `HiveOperation` primarily purpose is to ensure that authorization and transaction information is passed and used correctly during the compilation and execution of Hive queries. `HiveOperation` is not meant for end-user consumption so display purposes are a secondary concern.

A `LOAD` operation and a `QUERY` operation have different implications in terms of authorization and transactions.
* Is it OK to change the query state at the end of the analysis phase? 
* Will compilation and execution run under the same authorization/transaction semantics?
* Do we have any kind of authorization/transaction tests that ensure that things work correctly for `LOAD` statements after this change?

PR #5028, opened 2024-01-23
HIVE-27980: Hive Iceberg Compaction: add support for OPTIMIZE TABLE s…

difin:
…yntax

Added support for Hive Iceberg major compaction using `OPTIMIZE TABLE tableName REWRITE DATA` syntax.
This new syntax is implemented as a syntactic sugar which under the hood is being translated into `ALTER TABLE tableName COMPACT 'MAJOR' AND WAIT` command.

Iceberg Compaction in Impala is implemented using OPTIMIZE TABLE syntax.
Implementing the same syntax in Hive for consistency.

No

No

New query test is introduced as part of this patch.

### DIFF
--- a/iceberg/iceberg-handler/src/test/queries/positive/iceberg_optimize_table_unpartitioned.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/iceberg_optimize_table_unpartitioned.q
@@ -1,0 +1,58 @@
+-- SORT_QUERY_RESULTS
+-- Mask neededVirtualColumns due to non-strict order
+--! qt:replace:/(\s+neededVirtualColumns:\s)(.*)/$1#Masked#/
+-- Mask the totalSize value as it can have slight variability, causing test flakiness
+--! qt:replace:/(\s+totalSize\s+)\S+(\s+)/$1#Masked#$2/
+-- Mask random uuid
+--! qt:replace:/(\s+uuid\s+)\S+(\s*)/$1#Masked#$2/
+-- Mask a random snapshot id
+--! qt:replace:/(\s+current-snapshot-id\s+)\S+(\s*)/$1#Masked#/
+-- Mask added file size
+--! qt:replace:/(\S\"added-files-size\\\":\\\")(\d+)(\\\")/$1#Masked#$3/
+-- Mask total file size
+--! qt:replace:/(\S\"total-files-size\\\":\\\")(\d+)(\\\")/$1#Masked#$3/
+-- Mask current-snapshot-timestamp-ms
+--! qt:replace:/(\s+current-snapshot-timestamp-ms\s+)\S+(\s*)/$1#Masked#$2/
+-- Mask the enqueue time which is based on current time
+--! qt:replace:/(MAJOR\s+succeeded\s+)[a-zA-Z0-9\-\.\s+]+(\s+manual)/$1#Masked#$2/
+-- Mask compaction id as they will be allocated in parallel threads
+--! qt:replace:/^[0-9]/#Masked#/
+
+set hive.llap.io.enabled=true;
+set hive.vectorized.execution.enabled=true;
+set hive.optimize.shared.work.merge.ts.schema=true;
+
+create table ice_orc (
+    first_name string, 
+    last_name string
+ )
+stored by iceberg stored as orc 
+tblproperties ('format-version'='2');
+
+insert into ice_orc VALUES ('fn1','ln1');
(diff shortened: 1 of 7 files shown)


zhangbutao:
Hi, @difin you need to rebase yout PR to get the green Sonar check.

aturoczy (approved):
+1 Rerun test

parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
```diff
@@ -99,6 +100,19 @@ alterTblPartitionStatementSuffix[boolean partition]
   | alterStatementSuffixAddCol
   | alterStatementSuffixUpdateColumns
   ;
+    
+optimizeTableStatementSuffix
```
zhangbutao, line 104: We already have `alter table execute` syntax for partial iceberg maintance operations:
https://github.com/apache/hive/blob/94f318959ca6ec5f815e6326241bd2269ce91b80/parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g#L473

How do we distinguish between the two syntax? 
Is the `optimize` only used for iceberg compaction?

I'm not opposed to adding new syntax, but i think we need to clarify the specific use cases to avoid users confusion.

cc @ayushtkn @deniskuzZ any thoughts?

difin: Optimize syntax as proposed in this PR is a syntactic sugar for existing compaction syntax in Hive and it works for both Iceberg and ACID tables, but it is meant for Iceberg tables and the reason is that many Hive competitors use Optimize syntax for Iceberg compaction and it may help with Hive Iceberg adoption.
We are planning to extend the support for Optimize syntax in the future to support parameters like minimum number of files in the partition and file size threshold parameters which will be only for Iceberg:

```
OPTIMIZE TABLE <table_name>
REWRITE DATA [USING BIN_PACK]
[ ( { FILE_SIZE_THRESHOLD | MIN_INPUT_FILES } = <value> [, ... ] ) ]
WHERE category = 'c1'
```

ayushtkn: what is bad in our current syntax? we use that for acid, we will use it for iceberg as well, syntatic sugar makes sense when things go easy.

why don't the competitor use our syntax :-) 

no objections though from my side, btw. is it coming from spark or impala?

aturoczy: I think it is fine if the big data world has common syntax. It provides an easier learning curve.

deniskuzZ: looks good to me

iceberg/iceberg-handler/src/test/queries/positive/iceberg_optimize_table_unpartitioned.q
```diff
@@ -0,0 +1,58 @@
...
+update ice_orc set last_name = 'ln7a' where first_name='fn7';
+
+delete from ice_orc where last_name in ('ln5a', 'ln6a', 'ln7a');
+
+select * from ice_orc;
+describe formatted ice_orc;
+
+explain optimize table ice_orc rewrite data;
+optimize table ice_orc rewrite data;
+
```
deniskuzZ, line 55: should we check snapshots and files before and after compaction?

difin: It is checked using DESCRIBE FORMATTED command.
The iceberg compaction q-files contains that command before and after compaction and .q.out files compare number of data and delete files:

`current-snapshot-summary	{\"replace-partitions\":\"true\",\"added-data-files\":\"1\",\"added-records\":\"4\",\"added-files-size\":\"#Masked#\",\"changed-partition-count\":\"1\",\"total-records\":\"4\",\"total-files-size\":\"#Masked#\",\"total-data-files\":\"1\",\"total-delete-files\":\"0\",\"total-position-deletes\":\"0\",\"total-equality-deletes\":\"0\"}`

deniskuzZ: 👍

deniskuzZ (approved):
LGTM +1

PR #5032, opened 2024-01-24
HIVE-28024: Support sources profile for hive metastore modules

wecharyu:
Maven profile **sources** in Hive supports to generate source jar for all submodules. Since Hive Metastore modules are separated from hive parent module, it would be convenient to add this profile in standalone-metastore module.

To support generate sources in hive metastore submodules by profile `-Psources`

No.

No.

Test locally to check if the HMS source jar was generated by
```bash
mvn clean install -DskipTests -Psources
```

### DIFF
--- a/standalone-metastore/pom.xml
+++ b/standalone-metastore/pom.xml
@@ -658,5 +658,24 @@
         </plugins>
       </reporting>
     </profile>
+    <profile>
+      <id>sources</id>
+      <build>
+        <plugins>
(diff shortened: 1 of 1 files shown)


wecharyu:
@deniskuzZ @zabetak pls take a look on this PR.

zabetak (changes requested):
I am not sure why it's needed. Please see comments/questions under the JIRA case.

PR #5037, opened 2024-01-25
HIVE-24167: TPC-DS query 14 fails while generating plan for the filter

okumin:

Make it possible for PlanMapper to link related Operators or RelNodes correctly.

For your reference, I made a Gist page explaining the behaviors of PlanMapper and alternative approaches in my mind.
https://gist.github.com/okumin/b111fe0a911507bdf6a7204f49b9cb72

See also [the PR where I implemented another option, "Combine RelNodes with RelTreeSignature"](https://github.com/apache/hive/pull/5077).

No. 14 of TPC-DS or other complex queries with CTE materialized fail due to this bug.

No. This is a bug fix.

No.

I added a new test case, which fails with the current master branch, and restored No. 14 of TPC-DS.


ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -200,54 +199,30 @@ public void merge(Object o1, Object o2) {
...
-    for (Object object : keySet) {
-      EquivGroup group = objectMap.get(object);
-      if (group != null) {
-        mGroups.add(group);
-      }
-    }
-    if (mGroups.size() > 1) {
+    final EquivGroup group1 = objectMap.get(o1);
+    final EquivGroup group2 = objectMap.get(o2);
+    if (group1 != null && group2 != null && group1 != group2) {
```
aturoczy, line 204: I would change the logic a bit. In my mind, from Clean Code perspective if you check the true path it is easier to process and maintain the code.
i.e:
if (group1 == null || group2 == null || group1 == group2) 
return;

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -200,54 +199,30 @@ public void merge(Object o1, Object o2) {
...
       if (!mayMerge) {
         throw new RuntimeException("equivalence mapping violation");
```
aturoczy, line 206: If we touch this change please the exception message to Uppercase the first word

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -200,54 +199,30 @@ public void merge(Object o1, Object o2) {
...
-      }
+      group1.members.forEach(newGrp::add);
+      group2.members.forEach(newGrp::add);
       groups.add(newGrp);
-      groups.removeAll(mGroups);
+      groups.remove(group1);
+      groups.remove(group2);
     } else {
-      EquivGroup targetGroup = mGroups.isEmpty() ? new EquivGroup() : mGroups.iterator().next();
+      EquivGroup targetGroup = group1 != null ? group1 : (group2 != null ? group2 : new EquivGroup());
```
aturoczy, line 217: The logic is great, but my eyes hurted with this amount of ? : operator :)

aturoczy (changes requested):
Minimal change request, and comment

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
okumin: Thanks for your review but you may not need to review this PR yet.
I'm still investigating the issue and trying several approaches. I created this PR to trigger CI as I can't know which test cases could be affected because of the complexity of SemanticAnalyzer. It turned out that the current approach would break something else, and I am trying another approach.
See also https://gist.github.com/okumin/b111fe0a911507bdf6a7204f49b9cb72 to trace the current findings and my activity logs.

okumin: I made some refactoring and will resolve the current conversations once.

ql/src/test/queries/clientpositive/cbo_cte_materialization.q
```diff
@@ -0,0 +1,27 @@
...
+WITH materialized_cte AS (
+  SELECT key, value FROM src WHERE key != '100'
+),
+another_materialized_cte AS (
+  SELECT key, value FROM src WHERE key != '100'
+)
+SELECT a.key, a.value, b.key, b.value
+FROM materialized_cte a
+JOIN another_materialized_cte b ON a.key = b.key
+ORDER BY a.key;
```
zabetak, line 27: The problem is in the compilation phase so I don't think we need to actually run the query. Consider dropping the execution. If you opt to keep it then I would suggest crafting a much simpler test (without 1K lines in the output). It would be nice to keep the execution time of our test suite as low as possible. Moreover, it is not easy to see what the `src` table contains so verifying that the result is indeed correct is cumbersome.

ql/src/test/queries/clientpositive/cbo_cte_materialization.q
```diff
@@ -0,0 +1,27 @@
...
```
zabetak, line 16: Consider adding (or replacing this with) a traditional `EXPLAIN` where we can see the effect of `hive.optimize.cte.materialize.threshold` and the full plan for the materialized ctes.

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -200,54 +199,46 @@ public void merge(Object o1, Object o2) {
...
-    for (Object object : keySet) {
-      EquivGroup group = objectMap.get(object);
-      if (group != null) {
-        mGroups.add(group);
-      }
+    // Caches signatures on the first access. A signature of an Operator could change as optimizers could mutate it,
+    // keeping its semantics. The current implementation caches the signature before optimizations so that we can
+    // link Operators with their signatures at consistent timing.
+    registerSignature(o1);
+    registerSignature(o2);
```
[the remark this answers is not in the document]
kgyrtkirk: > This change reverts that logic and takes signatures out of the equation for determining equivalent groups.

haven't read all of it - but if thats true - please don't do that...
looking at the volume of changes - it must be changing something fundamental...

what's the issue - is it an equiv  violation? can you put the stacktrace somewhere?
I think that should be fixed without altering this part - it might be a missing link; or a missing signature annotation somewhere - you should not change this at all... 

all these things could enable to back-map runtime statistict up to the calcite join planning phase...but I think that feature was never completed/enabled

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -286,8 +271,13 @@ public Iterator<EquivGroup> iterateGroups() {
...
-    OpTreeSignature sig = signatureCache.getSignature(op);
-    return sig;
+    return signatureCache.getSignature(op);
+  }
+
+  private void registerSignature(Object o) {
+    if (o instanceof Operator) {
+      getSignatureOf((Operator<?>) o);
+    }
   }
```
kgyrtkirk, line 281: whats the point of this method? the cache will compute it if needed...what the point of pre-registering - if that's needed its not a *cache* anymore!

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -200,54 +199,46 @@ public void merge(Object o1, Object o2) {
...
-    Set<EquivGroup> mGroups = Collections.newSetFromMap(new IdentityHashMap<EquivGroup, Boolean>());
-
-    for (Object object : keySet) {
-      EquivGroup group = objectMap.get(object);
-      if (group != null) {
-        mGroups.add(group);
-      }
+    // Caches signatures on the first access. A signature of an Operator could change as optimizers could mutate it,
+    // keeping its semantics. The current implementation caches the signature before optimizations so that we can
+    // link Operators with their signatures at consistent timing.
```
kgyrtkirk, line 204: please remove this comment and don't pre-heat caches unless it gives real benefits
the purpose of the cache is not something like this ; but to prevent `O(N^2)` computation ( and also to reduce storage size by making it more serialize friendly...)

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -200,54 +199,46 @@ public void merge(Object o1, Object o2) {
...
+    registerSignature(o1);
+    registerSignature(o2);
+    final EquivGroup linkedGroup1 = objectMap.get(o1);
+    final EquivGroup linkedGroup2 = objectMap.get(o2);
```
kgyrtkirk, line 208: these changes will weaken this stuff - so that 2 OPs which have the same signature will be allowed -> that's not good....
its not causing much issues yet - because almost all existing operators have good signatures; but if that degrades...runtime stats will be applied incorrectly

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -200,54 +199,46 @@ public void merge(Object o1, Object o2) {
...
-        }
-      }
-      groups.add(newGrp);
-      groups.removeAll(mGroups);
-    } else {
-      EquivGroup targetGroup = mGroups.isEmpty() ? new EquivGroup() : mGroups.iterator().next();
-      groups.add(targetGroup);
-      targetGroup.add(o1);
-      targetGroup.add(o2);
+    if (linkedGroup1 == null || linkedGroup2 == null || linkedGroup1 == linkedGroup2) {
```
kgyrtkirk, line 217: what's all these changes do better than the old?

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -200,54 +199,46 @@ public void merge(Object o1, Object o2) {
...
+      final EquivGroup group = linkedGroup1 != null ? linkedGroup1 : linkedGroup2;
+      group.add(o1);
+      group.add(o2);
+      return;
     }
 
+    if (!mayMerge) {
+      throw new RuntimeException(String.format(
+          "Failed to link %s and %s. This error mostly means a bug of Hive",
```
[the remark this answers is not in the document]
kgyrtkirk: I find this error message pretty cloudy on what the issue is... 

I think the sentecte "Equivalence mapping violation" should be there - as that's what the problem is...we have 2 groups; not allowed to put them into the same equiv group -> big trouble...

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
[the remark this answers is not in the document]
kgyrtkirk: to get a failing test for these changes you would need 2 ops which have conflicting signatures stored into the metastore ; loaded back and they might get applied incorrectly...
possibly pretty hard to do...

kgyrtkirk:
just seen the details you linked to the PR.

I believe the `link` should be allowed - but in this the merge should be allowed due to the fact that these things represent the same thing.
I see now why the signature become an obstacle - I believe the SWO was able to match the same plan-parts to only scan the table once; if you don't have `link` calls there to explain the situation -> that could cause `equiv violations`  here...could you please check that?

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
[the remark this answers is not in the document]
okumin: > Assuming that we do not allow implicit linking via signatures I don't know what's the point of registering the signature at this stage

That's because some optimization can mutate Operators, changing their true signatures. [One example I remember is TableScanPPD](https://github.com/apache/hive/blob/rel/release-4.0.0-beta-1/ql/src/java/org/apache/hadoop/hive/ql/ppd/OpProcFactory.java#L414). I'm not 100% sure which signatures should be used in that case, but it sounds more reasonable and predictable for me to use ones before optimization than ones in the middle of optimization. I would say Operators should not be mutable for the signing purpose, but it is a too fundamental change.

I will make a reply to @kgyrtkirk in another thread as I guess those comments were written before reading my gist.

okumin:
> I believe the link should be allowed - but in this the merge should be allowed due to the fact that these things represent the same thing.



I see now why the signature become an obstacle - I believe the SWO was able to match the same plan-parts to only scan the table once; if you don't have link calls there to explain the situation -> that could cause equiv violations here...could you please check that?

okumin:
@kgyrtkirk

> I believe the link should be allowed - but in this the merge should be allowed due to the fact that these things represent the same thing.

I guess this mentions the link on creating a FilterOperator?

> I see now why the signature become an obstacle - I believe the SWO was able to match the same plan-parts to only scan the table once; if you don't have link calls there to explain the situation -> that could cause equiv violations here...could you please check that?

I guess the SWO stands for SharedWorkOptimizer. With CTE materialization, each CTE is compiled separately. It means `materialized_cte`, `another_materialized_cte`, and the body part are compiled one by one in a dedicated manner. So, SharedWorkOptimizer has no chance as it's invoked in analyzeInternal.

Just for a note. We use CTE because we need to support some cases which SharedWorkOptimizer doesn't support. An example is complex CTE + UNION ALL. We may replace CTE materialization with SharedWorkOptimizer when it becomes more mature.

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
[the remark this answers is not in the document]
okumin: I missed mentioning what are affected by such optimizations. Many test cases on the following revision failed with `equivalence mapping violation`.
https://github.com/apache/hive/pull/5037/commits/217b26db7eee95a0d25725df985adc79705a80d6

That's because it tries to LINK a pre-optimized operator with a post-optimized operator.
1. Before optimization
    - Operator A has a signature SA and is grouped as GA
    - Operator B has a signature SB and is grouped as GB
2. AuxSignatureLinker doesn't MERGE GA and GB as A and B have different signatures. Note that non aux signatures are not cached here
3. After optimization
    - Operator B is optimized, and its signature becomes SA
4. StatsRulePsrocFactory tries to LINK GA and GB as A and B share the same signature SA -> equivalence mapping violation

Potentially, GA and GB are actually mergeable though the current implementation doesn't do so.

kgyrtkirk:
> each CTE is compiled separately
got it now - I was  thinking its the other way around

I was thinking about this a bit in the last couple days; here is a summary of my thoughts:

* to have this feature cause trouble the following is needed:
   * a failing query
   * an incorrectly taken stat entry
   * the bad stat entry causes further issues
   * ....not very likely :D
* keeping multiple equiv groups which should be merged is not fortunate - but may not cause trouble
* at the end of the day all a `Map<Signature, RuntimeStat>` will be created from all these details
  * if there is are 2 groups which have the same `Signature` that will mean there will be multiple candidates for the key in the map => which means those should have been in a single group...that's why I didn't really like to relax the check on the signature
    * I believe lookup-by-signatue will either return one a representative group ; even when there are more

from the listed approaches in the doc:
* I liked the best the `RelTreeSignature` related; I think that would be the best if possioble
   * this might be a way to achieve a merge between the 2 groups
   * have you tried somehow relax the alias name of `HiveTableScan` in the `RelTreeSignature` ? 
      * that class creates the signature of the relnodes 1-by-1 ; so you could access/customize/etc based on the type
         * possibly use the `@Signature` annotation
         * use something to filter out poisonous stuff `@IgnoreSig` or something
         * by-hand stuff
* on the other hand after thinking it thru what might be the outcomes - I find that the idea to not pull the signatures during linking might actually work correctly...

okumin:
@kgyrtkirk 
I tried the option using `RelTreeSignature` in the following Pull Request. I still have some concerns, but it will likely work from my perspective.
https://github.com/apache/hive/pull/5077

Could you please take a look and see if it seems feasible? If we prefer #5077 , I will void this PR and try to get it merged.

okumin:
Let me close this one because we will likely go with #5077 or another approach.

PR #5038, opened 2024-01-26
HIVE-27778: Alter table command gives error after computer stats is r…

dengzhhu653:
…un with Impala

Loose the stats validation when alter the acid table, as it suggests from:
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L5115

Impala cannot alter the table even created by himself.

No

No

*Test1*
**hive**: 
drop table if exists test;
CREATE TABLE `test`(
   `trade_id` string,
   `version` int)
   partitioned by (part int)
 STORED AS ORC
 TBLPROPERTIES (
   'transactional'='true');   
insert into table test partition(part) values('aa', 1, 1), ('bb', 1, 1), ('cc', 2, 1), ('cc', 2, 2), ('dd', 3, 2), ('ee', 4, 1);
analyze table test COMPUTE STATISTICS FOR COLUMNS;
desc formatted test partition(part=1) version; --make sure hive partitions have col stats
**impala**:
compute stats test;
alter table test rename to test2;  
**hive**:
alter table test2 rename to test_y;

*Test2*
**impala**:
CREATE TABLE `test`(
   `trade_id` string,
   `version` int)
 STORED AS PARQUET
 TBLPROPERTIES (
   'transactional'='true',
   'transactional_properties'='insert_only');
**hive**:
insert into table test values('a', 1),('b',1),('c',2),('d',2),('c',3);
analyze table test COMPUTE STATISTICS FOR COLUMNS;
**impala**:
compute stats test;
alter table test rename to test_x;
**hive**:
alter table test_x rename to test;

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
@@ -5094,26 +5094,22 @@ public class ObjectStore implements RawStore, Configurable {
     if (validWriteIds != null && writeId > 0) {
       return null; // We have txn context.
     }
-    String oldVal = oldP == null ? null : oldP.get(StatsSetupConst.COLUMN_STATS_ACCURATE);
-    String newVal = newP == null ? null : newP.get(StatsSetupConst.COLUMN_STATS_ACCURATE);
-    // We don't need txn context is that stats state is not being changed.
-    if (StringUtils.isEmpty(oldVal) && StringUtils.isEmpty(newVal)) {
+
+    if (!StatsSetupConst.areBasicStatsUptoDate(newP)) {
+      // The validWriteIds can be absent, for example, in case of Impala alter.
+      // If the new value is invalid, then we don't care, let the alter operation go ahead.
       return null;
     }
+
+    String oldVal = oldP == null ? null : oldP.get(StatsSetupConst.COLUMN_STATS_ACCURATE);
+    String newVal = newP == null ? null : newP.get(StatsSetupConst.COLUMN_STATS_ACCURATE);
     if (StringUtils.equalsIgnoreCase(oldVal, newVal)) {
       if (!isColStatsChange) {
         return null; // No change in col stats or parameters => assume no change.
       }
-      // Col stats change while json stays "valid" implies stats change. If the new value is invalid,
-      // then we don't care. This is super ugly and idiotic.
-      // It will all become better when we get rid of JSON and store a flag and write ID per stats.
-      if (!StatsSetupConst.areBasicStatsUptoDate(newP)) {
-        return null;
-      }
     }
+
     // Some change to the stats state is being made; it can only be made with a write ID.
-    // Note - we could do this:  if (writeId > 0 && (validWriteIds != null || !StatsSetupConst.areBasicStatsUptoDate(newP))) { return null;
-    //       However the only way ID list can be absent is if WriteEntity wasn't generated for the alter, which is a separate bug.
     return "Cannot change stats state for a transactional table " + fullTableName + " without " +
             "providing the transactional write state for verification (new write ID " +
             writeId + ", valid write IDs " + validWriteIds + "; current state " + oldVal + "; new" +


dengzhhu653:
@deniskuzZ @kasakrisz could you please take a look if have cycles?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -5094,26 +5094,22 @@ public static String verifyStatsChangeCtx(String fullTableName, Map<String, Stri
     if (validWriteIds != null && writeId > 0) {
       return null; // We have txn context.
     }
-    String oldVal = oldP == null ? null : oldP.get(StatsSetupConst.COLUMN_STATS_ACCURATE);
-    String newVal = newP == null ? null : newP.get(StatsSetupConst.COLUMN_STATS_ACCURATE);
-    // We don't need txn context is that stats state is not being changed.
-    if (StringUtils.isEmpty(oldVal) && StringUtils.isEmpty(newVal)) {
+
+    if (!StatsSetupConst.areBasicStatsUptoDate(newP)) {
+      // The validWriteIds can be absent, for example, in case of Impala alter.
```
zhangbutao, line 5099: Since some components(such as Impala&Trino) may don't care the field `validWriteIds` , can we set the fileld a default value to fix this issue? Like `engine `field i have done:
https://github.com/apache/hive/blob/0f8190fefa56513a042566ef87b852e8343610aa/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift#L903-L904

It seems that we have introduced some extra filelds in stats related class since Hive4, but the stats related class are often used by other components(Impala&Trino, etc), so some incompatible issues are reported by the other components(Impala&Trino, etc). In one word, if we can set a default value for these new added field, maybe we can fix  similar issues at its root?

deniskuzZ: won't it cause an issue if we set engine = "hive" for Impala? Despite Impala stats is stored in HMS, it can't be used by Hive.

zhangbutao: Impala could retset the engine fileld value if it doesn't want to use hive stats.

dengzhhu653: As my understanding, there is no easy way to  set `validWriteIds` to a default value, and if the client is HiveServer2, we should always get a non-null `validWriteIds` on the HMS side.
The main idea of the change is that as long as the stats is marked as invalid(stale) on the new target table, we should just let the alter operation go regardless of whether the `validWriteIds` is null or not, as it won't pollute the Hive stats.

deniskuzZ: can we use for that purpose: `TableStatsRequest.isSetValidWriteIdList()`

deniskuzZ: @dengzhhu653 why do we need to lose validation? Would it work if we drop Impala stats on rename?

dengzhhu653: @deniskuzZ, sorry for the late reply!
The `validWriteIds` from impala engine is null, when the impala alters the acid table(like the Test2 in the description), it would be failed as the message shows:
>
ERROR : Failed
org.apache.hadoop.hive.ql.metadata.HiveException: Unable to alter table. Cannot change stats state for a transactional table default.parqtest without providing the transactional write state for verification (new write ID 6, valid write IDs null; current state null; new state {} 

If `StatsSetupConst.areBasicStatsUptoDate(newP)` is false for the new target table, then the alter operation should go as it has marked the stats as stale in the target table, it won't pollute the stats, nor cause trouble for Hive query optimization.

deniskuzZ (approved):
LGTM +1

PR #5040, opened 2024-01-26
HIVE-28038: Disable fallback to jdo for DeadlineException

wecharyu:
Mark `DeadlineException` as unrecoverable exception to avoid falling back to jdo query.

`DeadlineException` should be terminated immediately, instead of falling back to jdo query.
https://github.com/apache/hive/blob/b75a59779dcf9c5ef74890c8916dbe9a2c13aef4/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RetryingHMSHandler.java#L137-L142

No.

No.

Passing test:
```bash
mvn test -Dtest=org.apache.hadoop.hive.metastore.TestObjectStore -pl :hive-standalone-metastore-server
```

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
@@ -28,7 +28,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Stream;
-import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.hadoop.conf.Configurable;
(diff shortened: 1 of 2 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
```diff
@@ -52,9 +51,10 @@
  * */
 public class DatabaseProduct implements Configurable {
   static final private Logger LOG = LoggerFactory.getLogger(DatabaseProduct.class.getName());
-  private static final Class<SQLException>[] unrecoverableSqlExceptions = new Class[]{
-          // TODO: collect more unrecoverable SQLExceptions
-          SQLIntegrityConstraintViolationException.class
+  private static final Class<Exception>[] unrecoverableExceptions = new Class[]{
+          // TODO: collect more unrecoverable Exceptions
+          SQLIntegrityConstraintViolationException.class,
+          DeadlineException.class
```
zhangbutao, line 57: Just a thought.
I am thinking if direct sql is timeout due to some bugs, Is there a chance that the jdo retry will succeed?

wecharyu: We may need avoid such bugs, and even if there was such bug, we should fail-fast because this call was already timeout.

zhangbutao (approved):
LGTM

PR #5044, opened 2024-01-29
HIVE-28041 : Upgrade jetty to 9.4.53 due to CVE-2023-44487

AnmolSun:


AnmolSun:
Hive currently uses Apache DirServer 1.5.7, which is not compatible with Jetty 9.4.53.

The Apache DirServer version compatible with this Jetty upgrade is 2.0.0.AM25+. Unfortunately, the tests failing upon using Apache DirServer 2.0.0.AM25 is genuine and caused by what seems to be a change in behaviour. Lodged a Jira to DirServer here - [DIRSERVER-2386](https://issues.apache.org/jira/browse/DIRSERVER-2386)

PR #5052, opened 2024-01-30
HIVE-28047: Iceberg: Major QB Compaction with a single commit

difin:

Improvements to Hive Iceberg QB Major Compaction to perform compaction in one commit instead of two commits as was done till now. 

Existing implementation of compaction creates 2 commits which creates 2 snapshots: first snapshot with all the files deleted and second snapshot with compacted files. If a user queries the table based on snapshot id of the first snapshot, the result would be invalid as no data is present in the table in that snapshot. To avoid this problem this PR is proposed.

No

No

Hive contains 4 query tests for testing Hive Iceberg QB Major Compaction. The outputs of these q-tests were updated as part of this PR.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -553,7 +553,7 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
 
         List<DataFile> existingDataFiles = Lists.newArrayList();
         List<DeleteFile> existingDeleteFiles = Lists.newArrayList();
-        IcebergTableUtil.getFiles(table, existingDataFiles, existingDeleteFiles);
+        IcebergTableUtil.getDataAndDeleteFiles(table, existingDataFiles, existingDeleteFiles);
 
         RewriteFiles rewriteFiles = transaction.newRewrite();
         rewriteFiles.validateFromSnapshot(table.currentSnapshot().snapshotId());
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -60,6 +60,7 @@ import org.apache.iceberg.DeleteFile;
 import org.apache.iceberg.DeleteFiles;
 import org.apache.iceberg.OverwriteFiles;
 import org.apache.iceberg.ReplacePartitions;
+import org.apache.iceberg.RewriteFiles;
 import org.apache.iceberg.RowDelta;
 import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.SnapshotRef;
@@ -549,16 +550,27 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
     if (!results.dataFiles().isEmpty()) {
       Transaction transaction = table.newTransaction();
       if (rewritePolicy == RewritePolicy.ALL_PARTITIONS) {
-        DeleteFiles delete = transaction.newDelete();
-        delete.deleteFromRowFilter(Expressions.alwaysTrue());
-        delete.commit();
+
+        List<DataFile> existingDataFiles = Lists.newArrayList();
+        List<DeleteFile> existingDeleteFiles = Lists.newArrayList();
+        IcebergTableUtil.getFiles(table, existingDataFiles, existingDeleteFiles);
+
+        RewriteFiles rewriteFiles = transaction.newRewrite();
+        rewriteFiles.validateFromSnapshot(table.currentSnapshot().snapshotId());
+
+        existingDataFiles.stream().forEach(rewriteFiles::deleteFile);
+        existingDeleteFiles.stream().forEach(rewriteFiles::deleteFile);
+        results.dataFiles().stream().forEach(rewriteFiles::addFile);
+
+        rewriteFiles.commit();
+      } else {
+        ReplacePartitions overwrite = transaction.newReplacePartitions();
+        results.dataFiles().forEach(overwrite::addFile);
+        if (StringUtils.isNotEmpty(branchName)) {
+          overwrite.toBranch(HiveUtils.getTableSnapshotRef(branchName));
+        }
+        overwrite.commit();
       }
-      ReplacePartitions overwrite = transaction.newReplacePartitions();
-      results.dataFiles().forEach(overwrite::addFile);
-      if (StringUtils.isNotEmpty(branchName)) {
-        overwrite.toBranch(HiveUtils.getTableSnapshotRef(branchName));
-      }
-      overwrite.commit();
       transaction.commitTransaction();
       LOG.info("Overwrite commit took {} ms for table: {} with {} file(s)", System.currentTimeMillis() - startTime,
           table, results.dataFiles().size());
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -558,9 +558,9 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
         RewriteFiles rewriteFiles = transaction.newRewrite();
         rewriteFiles.validateFromSnapshot(table.currentSnapshot().snapshotId());
 
-        existingDataFiles.stream().forEach(rewriteFiles::deleteFile);
-        existingDeleteFiles.stream().forEach(rewriteFiles::deleteFile);
-        results.dataFiles().stream().forEach(rewriteFiles::addFile);
+        existingDataFiles.forEach(rewriteFiles::deleteFile);
+        existingDeleteFiles.forEach(rewriteFiles::deleteFile);
+        results.dataFiles().forEach(rewriteFiles::addFile);
 
         rewriteFiles.commit();
       } else {
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg.mr.hive;
 
+import com.sun.tools.javac.util.Pair;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
@@ -65,7 +66,6 @@ import org.apache.iceberg.RowDelta;
 import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.SnapshotRef;
 import org.apache.iceberg.Table;
-import org.apache.iceberg.Transaction;
 import org.apache.iceberg.exceptions.NotFoundException;
 import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.hadoop.Util;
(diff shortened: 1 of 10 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -328,4 +334,25 @@ public static PartitionData toPartitionData(StructLike key, Types.StructType key
...
+      }
+      for (DeleteFile delete : fileScanTask.deletes()) {
+        // filter repeated delete files
+        if (!deleteFilesPath.contains(delete.path().toString())) {
+          deleteFiles.add(delete);
+          deleteFilesPath.add(delete.path().toString());
+        }
+      }
+    }
+  }
```
zhangbutao, line 357: Can be simplified:
```suggestion
    fileScanTasks.forEach(fileScanTask -> {
      // filter repeated data files
      if (dataFilesPath.add(fileScanTask.file().path().toString())) {
        dataFiles.add(fileScanTask.file());
      }

      // filter repeated delete files
      fileScanTask.deletes().stream()
              .filter(delete -> deleteFilesPath.add(delete.path().toString()))
              .forEach(deleteFiles::add);
    });
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -328,4 +334,25 @@ public static PartitionData toPartitionData(StructLike key, Types.StructType key
...
     return data;
   }
+
+  public static void getFiles(Table table, List<DataFile> dataFiles, List<DeleteFile> deleteFiles) {
+    CloseableIterable<FileScanTask> fileScanTasks = table.newScan().planFiles();
+    Set<String> dataFilesPath = Sets.newHashSet();
+    Set<String> deleteFilesPath = Sets.newHashSet();
+
+    for (FileScanTask fileScanTask : fileScanTasks) {
+      // filter repeated data files
```
zhangbutao, line 344: Under what cases do we have repeated files?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -549,16 +550,27 @@ private void commitOverwrite(Table table, String branchName, long startTime, Fil
...
+        List<DataFile> existingDataFiles = Lists.newArrayList();
+        List<DeleteFile> existingDeleteFiles = Lists.newArrayList();
+        IcebergTableUtil.getFiles(table, existingDataFiles, existingDeleteFiles);
+
+        RewriteFiles rewriteFiles = transaction.newRewrite();
+        rewriteFiles.validateFromSnapshot(table.currentSnapshot().snapshotId());
+
+        existingDataFiles.stream().forEach(rewriteFiles::deleteFile);
+        existingDeleteFiles.stream().forEach(rewriteFiles::deleteFile);
+        results.dataFiles().stream().forEach(rewriteFiles::addFile);
```
zhangbutao, line 563:
```suggestion
        existingDataFiles.forEach(rewriteFiles::deleteFile);
        existingDeleteFiles.forEach(rewriteFiles::deleteFile);
        results.dataFiles().forEach(rewriteFiles::addFile);
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -328,4 +334,25 @@ public static PartitionData toPartitionData(StructLike key, Types.StructType key
     }
     return data;
   }
+
+  public static void getFiles(Table table, List<DataFile> dataFiles, List<DeleteFile> deleteFiles) {
```
zhangbutao, line 338: Maybe naming `getDataAndDeleteFiles()`? or orher more specific name. :)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -549,16 +550,27 @@ private void commitOverwrite(Table table, String branchName, long startTime, Fil
...
-        delete.commit();
-      }
-      ReplacePartitions overwrite = transaction.newReplacePartitions();
-      results.dataFiles().forEach(overwrite::addFile);
-      if (StringUtils.isNotEmpty(branchName)) {
-        overwrite.toBranch(HiveUtils.getTableSnapshotRef(branchName));
+
+        List<DataFile> existingDataFiles = Lists.newArrayList();
+        List<DeleteFile> existingDeleteFiles = Lists.newArrayList();
+        IcebergTableUtil.getFiles(table, existingDataFiles, existingDeleteFiles);
```
zhangbutao, line 556: Can we reuse `FilesForCommit results` to get the dataFiles and deleteFiles?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -328,4 +334,25 @@ public static PartitionData toPartitionData(StructLike key, Types.StructType key
...
+    CloseableIterable<FileScanTask> fileScanTasks = table.newScan().planFiles();
+    Set<String> dataFilesPath = Sets.newHashSet();
+    Set<String> deleteFilesPath = Sets.newHashSet();
```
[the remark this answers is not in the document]
difin: We can't, it causes Checkstyle exceptions:

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Makes sense, done.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: I am not sure under which cases there can be repeated files, I added that because I saw such duplicates prevention when getting data and delete files in a code of another engine which also implements Iceberg compaction.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: I don't think so, because FilesForCommit contains only the compacted, new files. They don't contain existing data and delete files.

zhangbutao: Since the Major QB Compaction is essentially a `IOW `operation, why don't we follow the regular `IOW `flow?  Why do we need the extra operation to find the existing data and delete files? 
Just reuse the  `ReplacePartitions  `api:
https://github.com/apache/hive/blob/e08a60029f93e52182908b61fff32e8659e78daa/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java#L556-L561

difin: Because IOW isn't supported on tables that had schema evolution as per https://issues.apache.org/jira/browse/HIVE-26133, but we want to support compaction even for tables that have had schema evolution.

zhangbutao: Got it, thx.

It seems that we have tradeoff about the change.
(twice commit) vs (single commit: loop to get and store all existing data and delte files)

I know twice commit will generate two snapshot and maybe get inconsistency of data when users reading table.
But for this change, if the original iceberg table has many many small data and delete files, and we need to loop and store these files in the `List`,  can this behavior would give memory pressure on HS?
I am not sure which way is the best one. ;(

Can other folks give some other thought? @deniskuzZ @ayushtkn @SourabhBadhya

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -549,16 +550,27 @@ private void commitOverwrite(Table table, String branchName, long startTime, Fil
...
-      ReplacePartitions overwrite = transaction.newReplacePartitions();
-      results.dataFiles().forEach(overwrite::addFile);
-      if (StringUtils.isNotEmpty(branchName)) {
-        overwrite.toBranch(HiveUtils.getTableSnapshotRef(branchName));
+
+        List<DataFile> existingDataFiles = Lists.newArrayList();
+        List<DeleteFile> existingDeleteFiles = Lists.newArrayList();
+        IcebergTableUtil.getDataAndDeleteFiles(table, existingDataFiles, existingDeleteFiles);
+
+        RewriteFiles rewriteFiles = transaction.newRewrite();
```
deniskuzZ, line 558: why do you need a transaction API in that case?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -328,4 +334,22 @@ public static PartitionData toPartitionData(StructLike key, Types.StructType key
     }
     return data;
   }
+
+  public static void getDataAndDeleteFiles(Table table, List<DataFile> dataFiles, List<DeleteFile> deleteFiles) {
```
deniskuzZ, line 338: code style violation, getter should return something, like Pair<dataFiles,deleteFiles>

deniskuzZ (commented):
In my opinion this approach looks more expensive, maybe we should reach out to iceberg community with the. proposal to extend current API with atomic IOW semantic?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: Removed

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Fixed

difin:
> In my opinion this approach looks more expensive, maybe we should reach out to iceberg community with the. proposal to extend current API with atomic IOW semantic?

I agree it is a little bit more expensive approach, but it doesn't have the correctness issue like the existing approach and this approach is used in another engines like Amoro, Trino and Impala is also in process of changing to this approach.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -328,4 +336,26 @@ public static PartitionData toPartitionData(StructLike key, Types.StructType key
     }
     return data;
   }
+
+  public static Pair<List<DataFile>, List<DeleteFile>> getDataAndDeleteFiles(Table table) {
```
deniskuzZ, line 340: you should be doing
````
    long startingSnapshotId = table.currentSnapshot().snapshotId();
    
    StructLikeMap<List<List<FileScanTask>>> fileGroupsByPartition =
    planFileGroups(startingSnapshotId);

    CloseableIterable<FileScanTask> fileScanTasks =
        table
            .newScan()
            .useSnapshot(startingSnapshotId)
            .filter(filter)
            .ignoreResiduals()
            .planFiles();
````
see RewriteDataFilesSparkAction

PR #5053, opened 2024-01-31
HIVE-28050: Disable Incremental non aggregated materialized view rebuild in presence of delete operations

kasakrisz:
Remove code of incremental rebuild of materialized views without aggregate in presence of delete operations

The current implementations leads to data correctness issues when a source table does not have a unique column.

No.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=materialized_view_create_rewrite_5.q,materialized_view_create_rewrite_8.q -pl itests/qtest -Pitests
```


ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/AlterMaterializedViewRebuildAnalyzer.java
```diff
@@ -480,8 +460,6 @@ protected ASTNode fixUpAfterCbo(ASTNode originalAst, ASTNode newAst, CalcitePlan
         fixUpASTAggregateInsertDeleteIncrementalRebuild(fixedAST, getMaterializedViewASTBuilder());
         return fixedAST;
       case JOIN_INSERT_DELETE_REBUILD:
-        fixUpASTJoinInsertDeleteIncrementalRebuild(fixedAST, getMaterializedViewASTBuilder());
-        return fixedAST;
```
amansinha100, line 484: This will end up throwing the UnsupportedOperationException.  Don't we want to fall back to the full rebuild ?

kasakrisz: At this point CBO plan is already transformed to AST hence we can not fall back. However `JOIN_INSERT_DELETE_REBUILD` is never set in the code any more.

amansinha100: Ok, just for my reference, for the non-groupby queries that this patch is addressing, could you point to where the fall back to the full rebuild (using insert-overwrite) happens ?

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/AlterMaterializedViewRebuildAnalyzer.java
```diff
@@ -322,8 +318,7 @@ private RelNode applyRecordIncrementalRebuildPlan(
               }
               return applyAggregateInsertDeleteIncremental(basePlan, mdProvider, executorProvider);
             } else {
-              return applyJoinInsertDeleteIncremental(
-                      basePlan, mdProvider, executorProvider, optCluster, calcitePreMVRewritingPlan);
+              return calcitePreMVRewritingPlan;
```
kasakrisz, line 321: @amansinha100 
This is the point where we fall back to full rebuild.

amansinha100 (approved):
+1

ql/src/test/queries/clientpositive/materialized_view_create_rewrite_5.q
```diff
@@ -11,6 +11,8 @@ set hive.materializedview.rewriting=true;
 create table cmv_basetable_n6 (a int, b varchar(256), c decimal(10,2), d int) stored as orc TBLPROPERTIES ('transactional'='true');
 
 insert into cmv_basetable_n6 values
+ (1, 'alfred', 10.30, 2),
+ (1, 'alfred', 10.30, 2),
```
zabetak, line 15: Instead of modifying existing tests to add duplicates in order to trigger the problem I think it would be better to add a new minimal test like the one you posted last under the JIRA ticket.

zabetak (approved):
LGTM, left few minor comments but feel free to disregard at your discretion.

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/AlterMaterializedViewRebuildAnalyzer.java
kasakrisz: Removed.

ql/src/test/queries/clientpositive/materialized_view_create_rewrite_5.q
kasakrisz: I added the 2nd example from the jira as a new q test but I also kept the change of this one because we ususally face extra challenges in case of MVs with Join.

PR #5054, opened 2024-01-31
HIVE-28051: LLAP: cleanup local folders on startup and periodically

abstractdog:
Implement a LocalDirCleaner which can remove old files from LLAP local dirs.

Because LLAP cannot take care of cleaning up old files in every failure scenario, this was shown in a customer problem. When I investigated [HIVE-24272](https://issues.apache.org/jira/browse/HIVE-24272) I found that local files for failed queries/DAGs are cleaned up, so this is clearly about leftovers after daemon crashes.

No.

No.

Unit test added, tested on llap daemon, added a file as:
```
mkdir -p  /apps/llap/work/usercache/hive/appcache/application_1707917402901_0001/3/output
touch  /apps/llap/work/usercache/hive/appcache/application_1707917402901_0001/3/output/file.out
```

set quick intervals like:
```
hive.llap.local.dir.cleaner.file.modify.time.threshold=60s
hive.llap.local.dir.cleaner.cleanup.interval=30s
```

checked logs:
```
query-executor <14>1 2024-02-15T07:37:59.351Z query-executor-0-0 query-executor 1 79184f83-fdb0-4982-9bd9-4e297438f8be [mdc@18060 class="impl.LocalDirCleaner" level="INFO" thread="pool-14-thread-1"] Cleaning up files older than 2024-02-15T07:36:59.351485Z from /apps/llap/work
query-executor <14>1 2024-02-15T07:38:29.351Z query-executor-0-0 query-executor 1 79184f83-fdb0-4982-9bd9-4e297438f8be [mdc@18060 class="impl.LocalDirCleaner" level="INFO" thread="pool-14-thread-1"] Cleaning up files older than 2024-02-15T07:37:29.351488Z from /apps/llap/work
query-executor <14>1 2024-02-15T07:38:59.351Z query-executor-0-0 query-executor 1 79184f83-fdb0-4982-9bd9-4e297438f8be [mdc@18060 class="impl.LocalDirCleaner" level="INFO" thread="pool-14-thread-1"] Cleaning up files older than 2024-02-15T07:37:59.351487Z from /apps/llap/work
query-executor <14>1 2024-02-15T07:38:59.352Z query-executor-0-0 query-executor 1 79184f83-fdb0-4982-9bd9-4e297438f8be [mdc@18060 class="impl.LocalDirCleaner" level="INFO" thread="pool-14-thread-1"] Delete old file: /apps/llap/work/usercache/hive/appcache/application_1707917402901_0001/3/output/file.out
query-executor <14>1 2024-02-15T07:39:29.351Z query-executor-0-0 query-executor 1 79184f83-fdb0-4982-9bd9-4e297438f8be [mdc@18060 class="impl.LocalDirCleaner" level="INFO" thread="pool-14-thread-1"] Cleaning up files older than 2024-02-15T07:38:29.351486Z from /apps/llap/work
query-executor <14>1 2024-02-15T07:39:59.351Z query-executor-0-0 query-executor 1 79184f83-fdb0-4982-9bd9-4e297438f8be [mdc@18060 class="impl.LocalDirCleaner" level="INFO" thread="pool-14-thread-1"] Cleaning up files older than 2024-02-15T07:38:59.351485Z from /apps/llap/work

```

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -5508,6 +5508,17 @@ public class HiveConf extends Configuration {
     LLAP_TASK_TIME_SUMMARY(
         "hive.llap.task.time.print.summary", false,
         "Display queue and runtime of tasks by host for every query executed by the shell."),
+
+    LLAP_LOCAL_DIR_CLEANER_CLEANUP_INTERVAL(
+        "hive.llap.local.dir.cleaner.cleanup.interval", "2h", new TimeValidator(TimeUnit.HOURS),
+      "Interval by which the LocalDirCleaner service in LLAP daemon checks for stale/old files." +
+      "Under normal circumstances, local files are cleaned up properly, so it's not recommended to" +
+      "set this more frequent than a couple of hours. Default is 2 hours."),
+    LLAP_LOCAL_DIR_CLEANER_FILE_MODIFY_TIME_THRESHOLD("hive.llap.local.dir.cleaner.file.modify.time.threshold", "24h",
+        new TimeValidator(TimeUnit.HOURS),
+      "Threshold time for LocalDirCleaner: if a regular file's modify time is older than this value, the file gets deleted." +
+      "Defaults to 86400s (1 day), which is a reasonable period for a local file to be considered as a stale one."),
+
     HIVE_TRIGGER_VALIDATION_INTERVAL("hive.trigger.validation.interval", "500ms",
       new TimeValidator(TimeUnit.MILLISECONDS),
       "Interval for validating triggers during execution of a query. Triggers defined in resource plan will get\n" +
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
@@ -1,0 +1,113 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hive.llap.daemon.impl;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.FileTime;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.service.AbstractService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * LocalDirCleaner is an LLAP daemon service to clean up old local files. Under normal circumstances,
+ * intermediate/local files are cleaned up (typically at end of the DAG), but daemons crash sometimes,
+ * and the attached local disk might end up being the same when a new daemon starts (this applies to
+ * on-prem as well as cloud scenarios).
+ */
+public class LocalDirCleaner extends AbstractService {
+  private static final Logger LOG = LoggerFactory.getLogger(LocalDirCleaner.class);
+
+  private List<String> localDirs;
+
+  private long cleanupIntervalSec;
+  private long fileModifyTimeThresholdSec;
+
+  ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+
+  public LocalDirCleaner(String[] localDirs, Configuration conf) {
+    super("LocalDirCleaner");
+    this.localDirs = Arrays.asList(localDirs);
+    this.cleanupIntervalSec = getInterval(conf);
+    this.fileModifyTimeThresholdSec = getFileModifyTimeThreshold(conf);
+    LOG.info("Initialized local dir cleaner: interval: {}s, threshold: {}s", cleanupIntervalSec,
+        fileModifyTimeThresholdSec);
+  }
+
+  @Override
+  public void serviceStart() throws IOException {
+    scheduler.scheduleAtFixedRate(this::cleanup, 0, cleanupIntervalSec, TimeUnit.SECONDS);
+  }
+
+  @Override
+  public void serviceStop() throws IOException {
+    // we can shutdown this service now and ignore leftovers, because under normal circumstances,
+    // files from the local dirs are cleaned up (so LocalDirCleaner is a best effort utility)
+    scheduler.shutdownNow();
+  }
+
+  private long getFileModifyTimeThreshold(Configuration conf) {
+    return HiveConf.getTimeVar(conf, HiveConf.ConfVars.LLAP_LOCAL_DIR_CLEANER_FILE_MODIFY_TIME_THRESHOLD,
+        TimeUnit.SECONDS);
+  }
+
+  private long getInterval(Configuration conf) {
+    return HiveConf.getTimeVar(conf, HiveConf.ConfVars.LLAP_LOCAL_DIR_CLEANER_CLEANUP_INTERVAL, TimeUnit.SECONDS);
+  }
+
+  private void cleanup() {
+    Instant deleteBefore = Instant.now().minus(fileModifyTimeThresholdSec, ChronoUnit.SECONDS);
+
+    localDirs.forEach(localDir -> cleanupPath(deleteBefore, Paths.get(localDir)));
+  }
+
+  private void cleanupPath(Instant deleteBefore, Path pathLocalDir) {
+    LOG.info("Cleaning up files older than {} from {}", deleteBefore, pathLocalDir);
+
+    try (Stream<Path> files = Files.walk(pathLocalDir)) {
+      files.filter(f -> {
+        try {
+          FileTime modified = Files.getLastModifiedTime(f);
+          LOG.debug("Checking: {}, modified: {}", f, modified);
+          return Files.isRegularFile(f) && modified.toInstant().isBefore(deleteBefore);
+        } catch (IOException ex) {
+          LOG.warn("IOException caught while checking file for deletion", ex);
+          return false;
+        }
+      }).forEach(f -> {
+        try {
+          LOG.info("Delete old local file: {}", f);
+          Files.delete(f);
+        } catch (IOException ex) {
+          LOG.warn("Failed to delete file", ex);
+        }
+      });
+    } catch (IOException e) {
+      LOG.warn("IOException caught while walking over local files", e);
+    }
+  }
+}
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LlapDaemon.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LlapDaemon.java
@@ -383,6 +383,7 @@ public class LlapDaemon extends CompositeService implements ContainerRunner, Lla
     // AMReporter after the server so that it gets the correct address. It knows how to deal with
     // requests before it is started.
     addIfService(amReporter);
+    addIfService(new LocalDirCleaner(localDirs, daemonConf));
   }
 
   private static long determineXmxHeadroom(
--- a/llap-server/src/test/org/apache/hadoop/hive/llap/daemon/impl/TestLlapDaemon.java
+++ b/llap-server/src/test/org/apache/hadoop/hive/llap/daemon/impl/TestLlapDaemon.java
@@ -34,9 +34,14 @@ import org.mockito.Mock;
 import org.mockito.internal.util.reflection.Fields;
 import org.mockito.internal.util.reflection.InstanceField;
 
+import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
@@ -55,6 +60,11 @@ public class TestLlapDaemon {
       MetricsUtils.METRICS_PROCESS_NAME
   };
 
+  public static final String TEST_LOCAL_DIR = new File(System.getProperty("java.io.tmpdir") +
+      File.separator + TestLlapDaemon.class.getCanonicalName()
+      + "-" + System.currentTimeMillis()
+  ).getPath().replaceAll("\\\\", "/");
+
   private Configuration hiveConf = new HiveConf();
 
   @Mock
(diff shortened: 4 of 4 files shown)


simhadri-g:
I think you missed updating the commit message with Hive jira

abstractdog:
> I think you missed updating the commit message with Hive jira

LOL, thanks, fixed

abstractdog:
unrelated test failure, I can rerun eventually, this can be reviewed

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -5508,6 +5508,17 @@ public static enum ConfVars {
     LLAP_TASK_TIME_SUMMARY(
         "hive.llap.task.time.print.summary", false,
         "Display queue and runtime of tasks by host for every query executed by the shell."),
+
+    LLAP_LOCAL_DIR_CLEANER_CLEANUP_INTERVAL(
```
zhangbutao, line 5512: Can we reuse the existing cleaning feature?
https://github.com/apache/hive/blob/e892ce7212b2f0c9e55092d89076b129c86c8172/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java#L3894-L3898

abstractdog: I'll check: maybe the utility itself can be reused but with keeping a separate config

abstractdog: checked [ClearDanglingScratchDir](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/session/ClearDanglingScratchDir.java), it seems to be specific to scratch dir cleanup and it's too smart for this scenario, I feel it would be more complicated to make it general and reuse than having a simple utility for the cases: ClearDanglingScratchDir checks lock files, works on hdfs, has a command line utility, etc., whereas LocalDirCleaner is just about to remove old files from a local folder

zhangbutao: Ok,will come back to check this after vacation. :)

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
```diff
@@ -0,0 +1,114 @@
...
+      }).forEach(f -> {
+        try {
+          LOG.info("Delete old local file: {}", f);
+          Files.delete(f);
+        } catch (IOException ex) {
+          LOG.warn("Failed to delete file", ex);
+        }
+      });
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
zhangbutao, line 111: Maybe return a more user-friendly exception? And aslo this cleaner is not a very necessary thread, maybe we can just `LOG.warn/error` instead of throw exception?

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public class LocalDirCleaner extends AbstractService {
+  private static final Logger LOG = LoggerFactory.getLogger(LocalDirCleaner.class);
+
+  private String[] localDirs;
+
+  private long cleanupIntervalSec;
+  private long fileModifyTimeThresholdSec;
+
+  ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
```
zhangbutao, line 48: Is 1 thread enough to clean the old files?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -5508,6 +5508,17 @@ public static enum ConfVars {
...
+        "hive.llap.local.dir.cleaner.cleanup.interval", "7200s", new TimeValidator(TimeUnit.SECONDS),
+      "Interval by which the LocalDirCleaner service in LLAP daemon checks for stale/old files. \n" +
+      "Under normal circumstances, local files are cleaned up properly, so it's not recommended to \n" +
+      "set this more frequent than a couple of hours. Default is 2 hours."),
+    LLAP_LOCAL_DIR_CLEANER_FILE_MODIFY_TIME_THRESHOLD("hive.llap.local.dir.cleaner.file.modify.time.threshold", "86400s",
```
zhangbutao, line 5517: Nit: If the usual period is at least one day, can we use hour as the default unit?
`LLAP_LOCAL_DIR_CLEANER_FILE_MODIFY_TIME_THRESHOLD("hive.llap.local.dir.cleaner.file.modify.time.threshold", "24h", new TimeValidator(TimeUnit.HOURS),`

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
```diff
@@ -0,0 +1,114 @@
...
+    for (String localDir : localDirs) {
+      Path pathLocalDir = Paths.get(localDir);
+      cleanupPath(deleteBefore, pathLocalDir);
+    }
+  }
+
+  private void cleanupPath(Instant deleteBefore, Path pathLocalDir) {
+    LOG.info("Cleaning up files older than {} from {}", deleteBefore, pathLocalDir);
+
+    try (Stream<Path> files = Files.walk(pathLocalDir)) {
```
zhangbutao, line 92: I am not sure the how the loca dir is generated. So this maybe a question: Can we use `Files.list()` instread of `Files.walk()`?
Files.walk() will  retrieve all subdirectories recursively, and maybe a expencive operation if `pathLocalDir` has many sub-directories.

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
```diff
@@ -0,0 +1,114 @@
...
+    return HiveConf.getTimeVar(conf, HiveConf.ConfVars.LLAP_LOCAL_DIR_CLEANER_CLEANUP_INTERVAL, TimeUnit.SECONDS);
+  }
+
+  private void cleanup() {
+    Instant deleteBefore = Instant.now().minus(fileModifyTimeThresholdSec, ChronoUnit.SECONDS);
+
+    for (String localDir : localDirs) {
+      Path pathLocalDir = Paths.get(localDir);
+      cleanupPath(deleteBefore, pathLocalDir);
+    }
```
zhangbutao, line 86: Can be simplified by stream api:
```suggestion
localDirs.forEach(localDir -> cleanupPath(deleteBefore, Paths.get(localDir)));
```

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
abstractdog: right, logging is enough, we're not handling this exception anyway

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
abstractdog: yes, I think so, this is not something to be over-optimized, old files are not expected to be present at all, but if any, we don't have a requirement for the performance of this

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
abstractdog: good catch, I'll change

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: makes sense, I can change it to hour as default unit

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
abstractdog: we have a root folder like:
```
/apps/llap/work/
```

and we'll have dangling files like:
```
/apps/llap/work/usercache/hive/appcache/application_1707917402901_0001/3/output/attempt_1707917402901_0001_3_05_000002_0_10414:
total 16
drwxr-xr-x  2 hive hive   44 Feb 14 13:41 .
drwxr-xr-x 63 hive hive 4096 Feb 14 13:41 ..
-rw-r--r--  1 hive hive  425 Feb 14 13:41 file.out
-rw-r--r--  1 hive hive   32 Feb 14 13:41 file.out.index

/apps/llap/work/usercache/hive/appcache/application_1707917402901_0001/3/output/attempt_1707917402901_0001_3_05_000005_0_10416:
total 16
drwxr-xr-x  2 hive hive   44 Feb 14 13:41 .
drwxr-xr-x 63 hive hive 4096 Feb 14 13:41 ..
-rw-r--r--  1 hive hive  383 Feb 14 13:41 file.out
-rw-r--r--  1 hive hive   32 Feb 14 13:41 file.out.index

```

so the point is to find all dangling intermediate files which might be a few layers below then the root work dir, so Files.walk()mseemed to be the way to find them all and I'm a bit concerned about how complicated it would be to achieve the same with Files.list()

zhangbutao: Can `/apps/llap/work/usercache/hive/appcache/` be the root folder? Can we only check the first directory layer?
List the root folder  `/apps/llap/work/usercache/hive/appcache/` and get all the sub-directory whose name starts with `application_`?
new File("/apps/llap/work/usercache/hive/appcache/").listFiles(fn -> fn.getName().startsWith("application_"));

abstractdog: delete an application folder might be risky: for long-running apps it's possible that we're using the same application folder for a day, so a modification/creation date doesn't really help there, I cannot really think of a solution where we can safely an easily remove unused folders

zhangbutao: yes, might be risky. There seems to be no better way.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -5508,6 +5508,17 @@ public static enum ConfVars {
...
+
+    LLAP_LOCAL_DIR_CLEANER_CLEANUP_INTERVAL(
+        "hive.llap.local.dir.cleaner.cleanup.interval", "2h", new TimeValidator(TimeUnit.HOURS),
+      "Interval by which the LocalDirCleaner service in LLAP daemon checks for stale/old files." +
+      "Under normal circumstances, local files are cleaned up properly, so it's not recommended to" +
+      "set this more frequent than a couple of hours. Default is 2 hours."),
+    LLAP_LOCAL_DIR_CLEANER_FILE_MODIFY_TIME_THRESHOLD("hive.llap.local.dir.cleaner.file.modify.time.threshold", "24h",
+        new TimeValidator(TimeUnit.HOURS),
+      "Threshold time for LocalDirCleaner: if a regular file's modify time is older than this value, the file gets deleted." +
+      "Defaults to 86400s (1 day), which is a reasonable period for a local file to be considered as a stale one."),
```
zhangbutao, line 5520: nit:
```suggestion
      "Defaults to 24h, which is a reasonable period for a local file to be considered as a stale one."),
```

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
```diff
@@ -0,0 +1,113 @@
...
+  @Override
+  public void serviceStop() throws IOException {
+    // we can shutdown this service now and ignore leftovers, because under normal circumstances,
+    // files from the local dirs are cleaned up (so LocalDirCleaner is a best effort utility)
+    scheduler.shutdownNow();
+  }
+
+  private long getFileModifyTimeThreshold(Configuration conf) {
+    return HiveConf.getTimeVar(conf, HiveConf.ConfVars.LLAP_LOCAL_DIR_CLEANER_FILE_MODIFY_TIME_THRESHOLD,
+        TimeUnit.SECONDS);
```
zhangbutao, line 75: nit:
```suggestion
        TimeUnit.HOURS);
```

zhangbutao (approved):
LGTM +1
Left several nit comments. Once the commets are fixed, we can merge this change without waitting CI.

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
```diff
@@ -0,0 +1,113 @@
...
+ * and the attached local disk might end up being the same when a new daemon starts (this applies to
+ * on-prem as well as cloud scenarios).
+ */
+public class LocalDirCleaner extends AbstractService {
+  private static final Logger LOG = LoggerFactory.getLogger(LocalDirCleaner.class);
+
+  private List<String> localDirs;
+
+  private long cleanupIntervalSec;
+  private long fileModifyTimeThresholdSec;
```
zhangbutao, line 48: Maybe
`private long fileModifyTimeThresholdHour;`
or
`private long fileModifyTimeThreshold;`

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
```diff
@@ -0,0 +1,113 @@
...
+
+  ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+
+  public LocalDirCleaner(String[] localDirs, Configuration conf) {
+    super("LocalDirCleaner");
+    this.localDirs = Arrays.asList(localDirs);
+    this.cleanupIntervalSec = getInterval(conf);
+    this.fileModifyTimeThresholdSec = getFileModifyTimeThreshold(conf);
+    LOG.info("Initialized local dir cleaner: interval: {}s, threshold: {}s", cleanupIntervalSec,
```
zhangbutao, line 57: Maybe the LOG also should use `hour `unit.

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
```diff
@@ -0,0 +1,113 @@
...
+    return HiveConf.getTimeVar(conf, HiveConf.ConfVars.LLAP_LOCAL_DIR_CLEANER_FILE_MODIFY_TIME_THRESHOLD,
+        TimeUnit.SECONDS);
+  }
+
+  private long getInterval(Configuration conf) {
+    return HiveConf.getTimeVar(conf, HiveConf.ConfVars.LLAP_LOCAL_DIR_CLEANER_CLEANUP_INTERVAL, TimeUnit.SECONDS);
+  }
+
+  private void cleanup() {
+    Instant deleteBefore = Instant.now().minus(fileModifyTimeThresholdSec, ChronoUnit.SECONDS);
```
zhangbutao, line 83: can use `HOUR `unit ?
If we want use `HOUR `as unit, should  we use `HOUR`  anywhere?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: right, I forgot to update this one :)

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
abstractdog: we cannot use this as HOUR default, because that way, every config with a shorter period will truncated to 0h:
```
2024-02-23T23:03:24,716  INFO [main] impl.LocalDirCleaner: Initialized local dir cleaner: interval: 0h, threshold: 0h
```
this happened when I tried to test this (in the unit test I need to configure SECONDS to get a result in a timely manner)
```
    HiveConf.setTimeVar(hiveConf, HiveConf.ConfVars.LLAP_LOCAL_DIR_CLEANER_CLEANUP_INTERVAL, 2, TimeUnit.SECONDS);
    HiveConf.setTimeVar(hiveConf, HiveConf.ConfVars.LLAP_LOCAL_DIR_CLEANER_FILE_MODIFY_TIME_THRESHOLD, 1,
        TimeUnit.SECONDS);
```

I think we can change the default unit in HiveConf to HOURS, but we're free to use SECONDS inside LocalDirCleaner, as it's the class' internal behavior

abstractdog: let's agree on this, and then we can resolve other comments I think

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
abstractdog: this depends on another review comment, I think we'll keep this SECONDS, so we can choose between fileModifyTimeThresholdSec and fileModifyTimeThreshold

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/LocalDirCleaner.java
abstractdog: HOUR is just the default unit for config values like "2"
the unit used here should correspond to the minus operation's first argument

abstractdog:
@zhangbutao : thanks for you comments, according to the last comment, I'm considering this as approved, let me know if it's otherwise

zhangbutao:
> @zhangbutao : thanks for your comments, according to the last comment, I'm considering this as approved, let me know if it's otherwise

Yes, free to merge this change. :)

PR #5055, opened 2024-02-01
HIVE-26818: Beeline module misses transitive dependencies due to shading

zabetak:
1. Do not generate the dependency-reduced-pom.xml and publish the regular pom.xml file along with the main jar file.
2. Remove some redundant explicit dependency declarations that were added to workaround the missing dependency problem from the beeline module.

Due to shading, the dependecy-reduced-pom.xml file is installed in the local maven repository (~/.m2/repository/org/apache/hive/hive-beeline/4.0.0-SNAPSHOT/) for beeline. The latter indicates that the module doesn't have any transitive dependencies. If we were publishing the shaded jar that would be true but we publish the regular jar. At this point, modules which include hive-beeline as a maven dependency are broken leading to build problems and classpath errors.

For more details see HIVE-26818.

No

No

1. Build separately main and itest modules
```
mvn clean install -DskipTests
cd itests && mvn clean install -DskipTests
```
2. Check that `TestBeelineDriver` does not throw `NoClassDefFoundError`
```
cd itests/qtest && mvn test -Dtest=TestBeeLineDriver -Dqfile=smb_mapjoin_1.q
```

### DIFF
--- a/beeline/pom.xml
+++ b/beeline/pom.xml
@@ -243,6 +243,7 @@
               <goal>shade</goal>
             </goals>
             <configuration>
+              <createDependencyReducedPom>false</createDependencyReducedPom>
               <descriptorRefs>
                 <descriptorRef>jar-with-dependencies</descriptorRef>
               </descriptorRefs>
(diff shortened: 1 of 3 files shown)


aturoczy (approved):

kokila-19 (approved):

ayushtkn (approved):
LGTM

PR #5056, opened 2024-02-01
HIVE-28053: Incorrect shading configuration for beeline jar-with-dependencies

zabetak:
1. Remove descriptorRefs cause it is not valid configuration parameter for maven-shade-plugin and has no effect on the build.
2. Set correct mainClass in the MANIFEST file making the jar runnable without extra options and avoid errors due to missing jmh class.

1. Developers will not see errors when browsing the respective pom.xml in the IDE
2. java -jar jar-with-dependencies now works and launches correctly beeline

No

Build and then run:
```
cd beeline && java -jar target/jar-with-dependencies
```

### DIFF
--- a/beeline/pom.xml
+++ b/beeline/pom.xml
@@ -243,14 +243,11 @@
               <goal>shade</goal>
             </goals>
             <configuration>
-              <descriptorRefs>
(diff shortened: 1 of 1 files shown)


asolimando (approved):
LGTM

PR #5057, opened 2024-02-02
HIVE-28054: SemanticException for join condition in subquery

soumyakanti3578:

Cleaned up `getPosition` method in `RexNodeExprFactory`

`getPosition` method didn't return the position of a column as soon as it found it in one of the RowResolvers. It threw an Exception if it encountered the column again in `outerRR`. Now it returns early.

No

No

`mvn test -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile=subquery_join.q`

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
@@ -165,22 +165,23 @@ public class RexNodeExprFactory extends ExprFactory<RexNode> {
 
   private int getPosition(ColumnInfo colInfo, List<RowResolver> rowResolverList)
       throws SemanticException {
-    ColumnInfo tmp;
-    ColumnInfo cInfoToRet = null;
     int position = 0;
-    for (RowResolver rr : rowResolverList) {
-      tmp = rr.get(colInfo.getTabAlias(), colInfo.getAlias());
-      if (tmp != null) {
-        if (cInfoToRet != null) {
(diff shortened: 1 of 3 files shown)


henrib (approved):
LGTM 👍 : clean code with comment and understandable result

ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
```diff
@@ -165,22 +165,23 @@ protected RexNode createColumnRefExpr(ColumnInfo colInfo, List<RowResolver> rowR
...
-          throw new CalciteSemanticException("Could not resolve column name");
-        }
-        cInfoToRet = tmp;
-        position += rr.getPosition(cInfoToRet.getInternalName());
-      } else if (cInfoToRet == null) {
+
+    for (RowResolver rr: rowResolverList) {
+      ColumnInfo tmp = rr.get(colInfo.getTabAlias(), colInfo.getAlias());
+      if (tmp == null) {
+        // if column in not present in the RR, increment position by size of RR
```
kasakrisz, line 173: typo?
is not present

kasakrisz (approved):
LGTM

ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
soumyakanti3578: Yes! I will fix it! Thanks for catching it :)

kasakrisz (approved):

PR #5062, opened 2024-02-06
HIVE-28061: Fix PerfLogger for recursive methods

soumyakanti3578:

Added a Map in `PerfLogger` that tracks recursion stack (calls) to `perfLogBegin` and `perfLogEnd` for recursive methods.
This fixes the issue described in HIVE-28061.

If we surround a recursive method with `perfLogBegin` and `perfLogEnd`, since the `method` doesn't change for each stack, `perfLogBegin` will overwrite the value in `startTimes`, and then for all `perfLogEnd`, the start value will be same, giving us wrong results. With this change, we compute the duration of the whole recursion stack, and output it in a single log.

No

No

`mvn test -Dtest=org.apache.hadoop.hive.ql.log.PerfLoggerTest#testRecursive`

Here are the relevant logs from my local machine
`2024-02-05T23:18:37,907 DEBUG [main] log.PerfLogger: <PERFLOG method=snoozeRecursive from=org.apache.hadoop.hive.ql.log.PerfLoggerTest>`
`2024-02-05T23:18:38,219 DEBUG [main] log.PerfLogger: </PERFLOG method=snoozeRecursive start=1707203917907 end=1707203918219 duration=312 from=org.apache.hadoop.hive.ql.log.PerfLoggerTest>`

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/ql/log/PerfLogger.java
+++ b/common/src/java/org/apache/hadoop/hive/ql/log/PerfLogger.java
@@ -93,6 +93,18 @@ public class PerfLogger {
   protected final Map<String, Long> startTimes = new ConcurrentHashMap<>();
   protected final Map<String, Long> endTimes = new ConcurrentHashMap<>();
 
+  /**
+   * This Map is used to keep track of the number of active calls to a method.
+   * Each call to perfLogBegin and perfLogEnd will increase and decrease the
+   * value respectively.
+   * <p>
+   * A start time will be logged only when the value for a method is 1, i.e.,
+   * the first call, and an end time will be logged only when the value is 0.
+   * So, this will behave like a stack for each method and prevent overwriting
+   * of start times during recursive calls.
+   */
+  protected final Map<String, Integer> activeMethodCalls = new ConcurrentHashMap<>();
+
   private static final Logger LOG = LoggerFactory.getLogger(PerfLogger.class.getName());
   protected static final ThreadLocal<PerfLogger> perfLogger = new ThreadLocal<>();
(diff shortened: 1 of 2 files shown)


zabetak (commented):
Reference counting is a smart idea to avoid reporting wrong execution times when there are nested calls to the same ID/method. However, it comes with a few disadvantages as well:
* It makes `PerfLogger` more complex since it increases the responsibilities of the class. Moreover, since we are using a `ConcurrentMap` I assume that the `PerfLogger` may be shared across threads which makes me wonder how reference counting can work correctly when the counts may be increased/decreased by different threads (and not necessarily by nested invocations of the same method).
* An additional map makes the class more expensive in terms of memory / cpu.
* Skipping nested invocations of methods may fix the elapsed times but it could potentially throw off other tracked `Metrics` such as `com.codahale.metrics.Counter`s (see `beginMetrics` and `endMetrics`)
* If the sequence of begin/end calls is not called as expected we may keep stale count references in the map creating another kind of bug.

Despite the potential risks listed above I am not against the changes here since they do seem to have some positive outcomes as well. I am just wondering if instead of modifying this class we should rather revisit explicitly the problematic calls to `beginPerfLog` and `endPerfLog` and reduce the scope of measurements and/or add more specific identifiers (e.g., not measure the whole method but parts of it).

@soumyakanti3578 What do you think is the best path forward?

soumyakanti3578:
Thanks @zabetak, I think the concerns you raised are significant, especially the one where different threads can call `perfLogBegin` with the same key. I am closing this PR for now.
cc: @ramesh0201

PR #5063, opened 2024-02-06
HIVE-28062: Optimize get_partitions_by_names in direct sql

wecharyu:
1. Add partition number check in `get_partitions_by_names`
2. Skipping get partition ids in `MetaStoreDirectSql#getPartitionsViaPartNames` to reduce the Database calls.

1. `get_partitions_by_names` also need check partition number in case the request number is too large
2. remove unnecessary db calls can improve the performance.

Yes, `get_partitions_by_names` api could also hit partition limit after this PR.

No.

1. Passing all existing tests.
2. Run the `getPartitionsByNames` benchmark tests:
```bash
java -jar ./hmsbench-jar-with-dependencies.jar -H localhost --savedata /tmp/benchdata --sanitize -N 100 -N 10000 -o bench_results_direct.csv -C -d testbench_http --params=100  -E 'drop.*' -E 'renameTable.*' -E 'getTableObjectsByName.*' -E 'listTables.*' -E 'listPartitions.*' -E 'getPartitionNames.*' -E 'listPartition' -E 'getPartition' -E 'getPartitions' -E 'getPartitions.100' -E 'getPartitions.10000' -E 'getPartitionsByFilter.*' -E 'addPartition.*' -E 'addPartitions.*' -E 'alterPartitions.*' -E 'getNid' -E 'listDatabases' -E 'getTable' -E 'createTable' -E 'openTxn.*'
```
- Before this patch:
```bash
getPartitionsByNames    7.47374 7.31857 6.96766 10.2396 7.39022
getPartitionsByNames.100        14.4165 14.2468 13.6072 17.9526 4.67421
getPartitionsByNames.10000      524.812 524.127 504.923 555.049 2.02143
```
- After this patch:
```bash
getPartitionsByNames    7.11728 7.02464 6.78734 8.59082 4.32347
getPartitionsByNames.100        13.4596 13.3253 12.8197 16.3828 3.77006
getPartitionsByNames.10000      460.498 459.805 438.384 500.921 2.57908
```

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestMetaStoreLimitPartitionRequest.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestMetaStoreLimitPartitionRequest.java
@@ -73,6 +73,9 @@ public class TestMetaStoreLimitPartitionRequest {
     conf.setBoolVar(HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL, true);
     conf.setBoolVar(HiveConf.ConfVars.DYNAMIC_PARTITIONING, true);
     conf.setBoolVar(HiveConf.ConfVars.HIVE_CBO_ENABLED, false);
+    // Disable loading dynamic partitions from partition names in this test
+    // because get_partitions_by_names will also hit partition limit since HIVE-28062.
+    conf.setBoolVar(HiveConf.ConfVars.HIVE_LOAD_DYNAMIC_PARTITIONS_SCAN_SPECIFIC_PARTITIONS, false);
 
     miniHS2 = new MiniHS2.Builder().withConf(conf).build();
     Map<String, String> overlayProps = new HashMap<String, String>();
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -3199,19 +3199,24 @@ private void constructOneLBLocationMap(FileStatus fSta,
           partitionNames.add(Warehouse.makeDynamicPartNameNoTrailingSeperator(details.fullSpec));
         }
       }
-      List<Partition> partitions = Hive.get().getPartitionsByNames(tbl, partitionNames);
-      for(Partition partition : partitions) {
-        LOG.debug("HMS partition spec: {}", partition.getSpec());
-        partitionDetailsMap.entrySet().parallelStream()
-            .filter(entry -> entry.getValue().fullSpec.equals(partition.getSpec()))
-            .findAny().ifPresent(entry -> {
-          entry.getValue().partition = partition;
-          entry.getValue().hasOldPartition = true;
-        });
+      try {
+        List<Partition> partitions = Hive.get().getPartitionsByNames(tbl, partitionNames);
+        for(Partition partition : partitions) {
+          LOG.debug("HMS partition spec: {}", partition.getSpec());
+          partitionDetailsMap.entrySet().parallelStream()
+              .filter(entry -> entry.getValue().fullSpec.equals(partition.getSpec()))
+              .findAny().ifPresent(entry -> {
+            entry.getValue().partition = partition;
+            entry.getValue().hasOldPartition = true;
+          });
+        }
+        // no need to fetch partition again in tasks since we have already fetched partitions
+        // info in getPartitionsByNames()
+        fetchPartitionInfo = false;
+      } catch (HiveException e) {
+        // Failed to fetch partitions in some cases (e.g., partition number limit)
+        LOG.warn("Fetching partitions by names failed.", e);
       }
-      // no need to fetch partition again in tasks since we have already fetched partitions
-      // info in getPartitionsByNames()
-      fetchPartitionInfo = false;
     }
 
     boolean isTxnTable = AcidUtils.isTransactionalTable(tbl);
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -3199,24 +3199,19 @@ private void constructOneLBLocationMap(FileStatus fSta,
           partitionNames.add(Warehouse.makeDynamicPartNameNoTrailingSeperator(details.fullSpec));
         }
       }
-      try {
-        List<Partition> partitions = Hive.get().getPartitionsByNames(tbl, partitionNames);
-        for(Partition partition : partitions) {
-          LOG.debug("HMS partition spec: {}", partition.getSpec());
-          partitionDetailsMap.entrySet().parallelStream()
-              .filter(entry -> entry.getValue().fullSpec.equals(partition.getSpec()))
-              .findAny().ifPresent(entry -> {
-            entry.getValue().partition = partition;
-            entry.getValue().hasOldPartition = true;
-          });
-        }
-        // no need to fetch partition again in tasks since we have already fetched partitions
-        // info in getPartitionsByNames()
-        fetchPartitionInfo = false;
-      } catch (HiveException e) {
-        // Failed to fetch partitions in some cases (e.g., partition number limit)
-        LOG.warn("Fetching partitions by names failed.", e);
+      List<Partition> partitions = Hive.get().getPartitionsByNames(tbl, partitionNames);
+      for(Partition partition : partitions) {
+        LOG.debug("HMS partition spec: {}", partition.getSpec());
+        partitionDetailsMap.entrySet().parallelStream()
+            .filter(entry -> entry.getValue().fullSpec.equals(partition.getSpec()))
+            .findAny().ifPresent(entry -> {
+          entry.getValue().partition = partition;
+          entry.getValue().hasOldPartition = true;
+        });
       }
+      // no need to fetch partition again in tasks since we have already fetched partitions
+      // info in getPartitionsByNames()
+      fetchPartitionInfo = false;
     }
 
     boolean isTxnTable = AcidUtils.isTransactionalTable(tbl);
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -7466,7 +7466,7 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     return result;
   }
 
-  public List<Partition> get_partitions_by_names(final String dbName, final String tblName,
+  private List<Partition> get_partitions_by_names(final String dbName, final String tblName,
       boolean getColStats, String engine,
       List<String> processorCapabilities, String processorId,
       GetPartitionsArgs args) throws TException {
@@ -7478,14 +7478,14 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     Table table = null;
     Exception ex = null;
     boolean success = false;
-    startTableFunction("get_partitions_by_names", parsedCatName, parsedDbName,
-        tblName);
+    startTableFunction("get_partitions_by_names", parsedCatName, parsedDbName, tblName);
     try {
       getMS().openTransaction();
       authorizeTableForPartitionMetadata(parsedCatName, parsedDbName, tblName);
 
       fireReadTablePreEvent(parsedCatName, parsedDbName, tblName);
 
+      checkLimitNumberOfPartitions(tblName, args.getPartNames().size(), -1);
       ret = getMS().getPartitionsByNames(parsedCatName, parsedDbName, tblName, args);
       ret = FilterUtils.filterPartitionsIfEnabled(isServerFilterEnabled, filterHook, ret);
       table = getTable(parsedCatName, parsedDbName, tblName);
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
@@ -739,8 +739,7 @@ class MetaStoreDirectSql {
     return Batchable.runBatched(batchSize, partNames, new Batchable<String, Partition>() {
       @Override
       public List<Partition> run(List<String> input) throws MetaException {
-        return getPartitionsByNames(catName, dbName, tblName, null,
-                partNames, Collections.emptyList(), false, args);
+        return getPartitionsByNames(catName, dbName, tblName, partNames, false, args);
       }
     });
   }
@@ -764,8 +763,7 @@ class MetaStoreDirectSql {
     return Batchable.runBatched(batchSize, partitionIds, new Batchable<Long, Partition>() {
       @Override
       public List<Partition> run(List<Long> input) throws MetaException {
-        return getPartitionsFromPartitionIds(catName, dbName,
-            tableName, null, input, Collections.emptyList(), isAcidTable, args);
+        return getPartitionsByPartitionIds(catName, dbName, tableName, input, isAcidTable, args);
       }
     });
   }
@@ -919,7 +917,7 @@ class MetaStoreDirectSql {
     List<Partition> result = Batchable.runBatched(batchSize, partitionIds, new Batchable<Long, Partition>() {
       @Override
       public List<Partition> run(List<Long> input) throws MetaException {
-        return getPartitionsFromPartitionIds(catName, dbName, tblName, null, input, Collections.emptyList(), false, args);
+        return getPartitionsByPartitionIds(catName, dbName, tblName, input, false, args);
       }
     });
     return result;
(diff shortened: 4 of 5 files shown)


wecharyu:
> The failed tests seem to be related.
http://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5063/1/tests

@zhangbutao The issue arises due to the introduction of a partition number check in the `getPartitionsByNames()` API, fixed it in the client side.

wecharyu:
@zhangbutao could you take a look again?

zhangbutao:
+ @dengzhhu653 @deniskuzZ @ayushtkn @saihemanth-cloudera may be interested in this improvement. :)

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -3199,19 +3199,24 @@ public Map<Map<String, String>, Partition> loadDynamicPartitions(final LoadTable
...
+            entry.getValue().partition = partition;
+            entry.getValue().hasOldPartition = true;
+          });
+        }
+        // no need to fetch partition again in tasks since we have already fetched partitions
+        // info in getPartitionsByNames()
+        fetchPartitionInfo = false;
+      } catch (HiveException e) {
+        // Failed to fetch partitions in some cases (e.g., partition number limit)
+        LOG.warn("Fetching partitions by names failed.", e);
```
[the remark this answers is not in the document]
deniskuzZ: i don't see how this change is relevant to the PR, also we should propagate the exception back to the client

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7516,14 +7516,14 @@ public List<Partition> get_partitions_by_names(final String dbName, final String
...
-    startTableFunction("get_partitions_by_names", parsedCatName, parsedDbName,
-        tblName);
+    startTableFunction("get_partitions_by_names", parsedCatName, parsedDbName, tblName);
     try {
       getMS().openTransaction();
       authorizeTableForPartitionMetadata(parsedCatName, parsedDbName, tblName);
 
       fireReadTablePreEvent(parsedCatName, parsedDbName, tblName);
 
+      checkLimitNumberOfPartitions(tblName, args.getPartNames().size(), -1);
```
deniskuzZ, line 7526: this checks limit regardless of name filter, isn't it?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -742,14 +742,8 @@ public List<Partition> getPartitionsViaPartNames(final String catName, final Str
...
       public List<Partition> run(List<String> input) throws MetaException {
-        String filter = "" + PARTITIONS + ".\"PART_NAME\" in (" + makeParams(input.size()) + ")";
-        List<Long> partitionIds = getPartitionIdsViaSqlFilter(catName, dbName, tblName,
-            filter, input, Collections.<String>emptyList(), null);
-        if (partitionIds.isEmpty()) {
-          return Collections.emptyList(); // no partitions, bail early.
-        }
-        return getPartitionsFromPartitionIds(catName, dbName, tblName, null,
-                partitionIds, Collections.emptyList(), false, args);
+        return getPartitionsFromPartitionNames(catName, dbName, tblName, null,
```
deniskuzZ, line 745: maybe `getPartitionsByNames`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1027,15 +1021,42 @@ public <T> List<T> getPartitionFieldsViaSqlFilter(
...
+  /** Should be called with the list short enough to not trip up Oracle/etc. */
+  private List<Partition> getPartitionsFromPartitionNames(String catName, String dbName,
+      String tblName, Boolean isView, List<String> partNameList, List<String> projectionFields,
+      boolean isAcidTable, GetPartitionsArgs args) throws MetaException {
+    // Get most of the fields for the partNames provided.
+    // Assume db and table names are the same for all partition, as provided in arguments.
+    String partNames = partNameList.stream()
+            .map(name -> "'" + name + "'")
+            .collect(Collectors.joining(","));
+    String queryText =
```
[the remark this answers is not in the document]
wecharyu: Yes, this query is familiar with following: 
https://github.com/wecharyu/hive/blob/a348e5d86adc611e4f72d386815a21a96b667ab4/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java#L448-L455

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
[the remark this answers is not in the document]
wecharyu: There are two ways to fetch partitions from hms in `loadDynamicPartitions()`:
1. get partitions by names
2. get partitions one by one
I think it's reasonable to fallback to the second if the first one failed. And we have a warn log to remind client what error happened in getPartitionsByNames call.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -742,14 +742,8 @@ public List<Partition> getPartitionsViaPartNames(final String catName, final Str
     return Batchable.runBatched(batchSize, partNames, new Batchable<String, Partition>() {
       @Override
       public List<Partition> run(List<String> input) throws MetaException {
-        String filter = "" + PARTITIONS + ".\"PART_NAME\" in (" + makeParams(input.size()) + ")";
-        List<Long> partitionIds = getPartitionIdsViaSqlFilter(catName, dbName, tblName,
```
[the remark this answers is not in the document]
wecharyu: Does not find the clues, I guess it's because there are existing `getPartitionIdsViaSqlFilter()` and `getPartitionsFromPartitionIds` methods that can be used directly.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: Yes, it check by the total size of input names, we assume that all input partition names are valid and related partitions exist.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
wecharyu: I named this new method in the format of `getPartitionsFromPartitionIds()`, do we need change it?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
deniskuzZ: ok

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
deniskuzZ: i think that would be more readable, btw same naming is used across the code

deniskuzZ:
No

We didn't have a limit before, didn't we? Right now user might get an exception when the limit is exceeded

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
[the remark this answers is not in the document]
deniskuzZ: oh, is the fallback code below? i don't see we use the same `partitionNames` list there.
what would happen if limit is exceeded? we'll catch the exception and try to load 1-by-1?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1027,15 +1021,42 @@ public <T> List<T> getPartitionFieldsViaSqlFilter(
...
+  private List<Partition> getPartitionsFromPartitionNames(String catName, String dbName,
+      String tblName, Boolean isView, List<String> partNameList, List<String> projectionFields,
+      boolean isAcidTable, GetPartitionsArgs args) throws MetaException {
+    // Get most of the fields for the partNames provided.
+    // Assume db and table names are the same for all partition, as provided in arguments.
+    String partNames = partNameList.stream()
+            .map(name -> "'" + name + "'")
+            .collect(Collectors.joining(","));
+    String queryText =
+        "select " + PARTITIONS + ".\"PART_ID\", " + SDS + ".\"SD_ID\", " + SDS + ".\"CD_ID\"," + " "
```
deniskuzZ, line 1035: can we reformat the query for readability? put joins on new lines

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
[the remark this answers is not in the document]
zhangbutao: I didn't see any fallback code block. BTW, i think no one will care about the warn log, so client side(spark & hiveserver2) maybe miss this warn exception, and then they maybe lose some partition data.

Therefore, if this PR must change the client behavior, such as the Introduced `partition limit exception` , (see the previous ci http://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5063/1/tests/ ), we need throw this exception. But as i said above, it would better to not to introduce this incompatible behavior if possible.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
wecharyu: Addressed.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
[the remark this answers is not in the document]
deniskuzZ: The only way the client realizes that there is a problem (limit exceeded) is if he checks the logs. I think we should throw an exception

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1023,33 +1023,43 @@ public <T> List<T> getPartitionFieldsViaSqlFilter(
...
+        + SERDES + ".\"SLIB\"," + PARTITIONS + ".\"WRITE_ID\"" + " from " + PARTITIONS
+        + " left outer join " + SDS + " on " + PARTITIONS + ".\"SD_ID\" = " + SDS + ".\"SD_ID\" "
+        + " left outer join " + SERDES + " on " + SDS + ".\"SERDE_ID\" = " + SERDES + ".\"SERDE_ID\" "
+        + " inner join " + TBLS + " on " + TBLS + ".\"TBL_ID\" = " + PARTITIONS + ".\"TBL_ID\" "
+        + " inner join " + DBS + " on " + DBS + ".\"DB_ID\" = " + TBLS + ".\"DB_ID\" where ");
+    suffix.append(
+        " and " + TBLS + ".\"TBL_NAME\" = ? and " + DBS + ".\"NAME\" = ? and " + DBS
+        + ".\"CTLG_NAME\" = ? order by \"PART_NAME\" asc");
+
+    TxnUtils.buildQueryWithINClauseStrings(conf, queries, prefix, suffix, quotedPartNames,
```
deniskuzZ, line 1054: sorry, I didn't notice that it's already under `Batchable.runBatched`. Not sure we want to use `TxnUtils.buildQueryWithINClauseStrings` here as it's used to construct mini-batches as well. 
Simple `IN` should be sufficient:
````
 String quotedPartNames = partNameList.stream()
        .map(DirectSqlUpdatePart::quoteString)
        .collect(Collectors.joining(","));

"where \"PART_NAME\" IN (" + quotedPartNames + ") "
````

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java (hunk 1 above)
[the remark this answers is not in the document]
Comment (wecharyu) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java]: Reverted this change.

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java (hunk 7 above)
Comment (wecharyu) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java]: Addressed.

File: itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestMetaStoreLimitPartitionRequest.java [hunk 8]
```diff
@@ -73,6 +73,9 @@ public static void beforeTest() throws Exception {
     conf.setBoolVar(HiveConf.ConfVars.METASTORE_TRY_DIRECT_SQL, true);
     conf.setBoolVar(HiveConf.ConfVars.DYNAMIC_PARTITIONING, true);
     conf.setBoolVar(HiveConf.ConfVars.HIVE_CBO_ENABLED, false);
+    // Disable loading dynamic partitions from partition names in this test
+    // because get_partitions_by_names will also hit partition limit since HIVE-28062.
+    conf.setBoolVar(HiveConf.ConfVars.HIVE_LOAD_DYNAMIC_PARTITIONS_SCAN_SPECIFIC_PARTITIONS, false);
```
zhangbutao, line 78: `hive.load.dynamic.partitions.scan.specific.partitions` was inroduced by HIVE-25178 to fetch only specific relevant **external** tables partitions by partition names(using `getPartitionsByNames `method):
https://github.com/apache/hive/blob/ffb1165f59defa66b31b4fd9cb6367b71050071b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L3187-L3202

Here setting `hive.load.dynamic.partitions.scan.specific.partitions` to false will give others a false impression that this propertity can be used to control the partitions limit. However, It's just a coincidence as it just skip to  invoke `getPartitionsByNames `method when this propertity is false, and it has nothing to do with partitions limit.

Am i right?
If it is ok, i would suggest to think another better to fix this.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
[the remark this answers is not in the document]
zhangbutao: Please see comment https://github.com/apache/hive/pull/5063/files#r1528710963

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestMetaStoreLimitPartitionRequest.java
wecharyu: It's an option to determine whether use `getPartitionsByNames` to load partitions, in this test we disable it because  the setup of this test would be failed by partition limit, which is not related to the real tests actually.

zhangbutao: I know it's just a workaround to fix the failed UT and not related to the real cases. But IMO, `hive.load.dynamic.partitions.scan.specific.partitions` should not be bound together with `getPartitionsByNames `, and we may can use other methods like `getPartitionsBy...` to achieve this in future.
This part maks me little confusing, but this is just a small fix about the UT, so i think it does't matter. :)

zhangbutao (approved):
LGTM +1

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1050,17 +1073,26 @@ private List<Partition> getPartitionsFromPartitionIds(String catName, String dbN
...
             + SERDES + ".\"SERDE_ID\" " + "where \"PART_ID\" in (" + partIds
             + ") order by \"PART_NAME\" asc";
+    // Keep order by name, consistent with JDO.
+    ArrayList<Partition> orderedResult = new ArrayList<Partition>(partIdList.size());
+    populatePartitionsByQuery(catName, dbName, tblName, isView, queryText, null, projectionFields,
+        isAcidTable, args, orderedResult);
+    return orderedResult;
+  }
+
+  private void populatePartitionsByQuery(String catName, String dbName, String tblName,
```
deniskuzZ, line 1083: why not return `List<Partition> getPartitionsByQuery(..)` instead of passing it as an input param?

wecharyu: Addressed. The initial thought was that the result list size can be determined by partition names size, so I want to init the result list before calling `populatePartitionsByQuery`, but actually the result size can be fetched from the sqlResult.

deniskuzZ (approved):
LGTM +1

zhangbutao:
http://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5063/13/tests/ 
`org.apache.iceberg.mr.hive.TestConflictingDataFiles::testMultiFiltersUpdate`   I think this is a flaky test. I have seen this faliure many times.
Run a flaky check to see results. http://ci.hive.apache.org/job/hive-flaky-check/831/

zhangbutao:
> http://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5063/13/tests/ `org.apache.iceberg.mr.hive.TestConflictingDataFiles::testMultiFiltersUpdate` I think this is a flaky test. I have seen this faliure many times. Run a flaky check to see results. http://ci.hive.apache.org/job/hive-flaky-check/831/

http://ci.hive.apache.org/job/hive-flaky-check/831/testReport/  test report shows that this is a flay test. Hi @simhadri-g , do you have any suggestion about this flaky test, as i see you wrote this code not long ago. Should we disable it now?

Thanks.

PR #5065, opened 2024-02-06
HIVE-27845:Upgrade protobuf to 3.24.4 to fix CVEs

tanishq-chugh:
In order to resolve the CVEs caused by 3.21.7 version of protobuf-java, we are trying to upgrade it to 3.24.4 version.

Why are the changes needed?
We need to update the protobuf-java version in hive's pom.xml.

No

No

How was this patch tested?
Manual tests.
After updating the protobuf-java version, we build the hive project again.
After manually testeing some DDL and SQL commands, hive is working as expected.

### DIFF
--- a/beeline/pom.xml
+++ b/beeline/pom.xml
@@ -243,6 +243,7 @@
               <goal>shade</goal>
             </goals>
             <configuration>
+              <createDependencyReducedPom>false</createDependencyReducedPom>
               <finalName>jar-with-dependencies</finalName>
               <transformers>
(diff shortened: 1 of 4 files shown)


Aggarwal-Raghav:
@tanishq-chugh, if you are upgrading the protobuf version then, you also need to update the generated protobuf code with the new version.

Aggarwal-Raghav:
In standalone-metastore/pom.xml also, you need to upgrade the version to have same dependency version across the modules.
https://github.com/apache/hive/blob/32a79270c33e151826b43eec0daa985b159fc568/standalone-metastore/pom.xml#L93

PR #5067, opened 2024-02-06
HIVE-28064: Add cause to ParseException for diagnosability purposes

zabetak:
The ParseException contains high level information about problems encountered during parsing but currently the stacktrace is pretty shallow. The end-user gets a hint about what the error might be but the developer has no way to tell how far we went into parsing the given statement and which grammar rule failed to pass.

Add RecognitionException (when available) as cause in ParseException to provide better insights around the origin of the problem and grammar rules that were invoked till the crash.

People who read the HS2 logs will see the additional stacktrace from the cause of the ParseException.

No

Run queries with parse errors via `TestMiniLlapLocalCliDriver` and inspect logs.


okumin (approved):

zhangbutao (approved):
LGTM +1

PR #5068, opened 2024-02-06
HIVE-28057: Iceberg: Branches with non-lowercase characters can't be accessed.

ayushtkn:

Make branches with non lowercase characters accessible.

You can create/have branches with lower cases, but can't query them

Yes, they can now query branches with non lowercase characters.

No

UT

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
@@ -472,7 +472,7 @@ public final class HiveUtils {
   public static String getLowerCaseTableName(String refName) {
     String[] refParts = refName.split("\\.");
     if (refParts.length == 3 && SNAPSHOT_REF.matcher(refParts[2]).matches()) {
-      return (refParts[0].toLowerCase() + "." + refParts[1]).toLowerCase() + "." + refParts[2];
+      return (refParts[0] + "." + refParts[1]).toLowerCase() + "." + refParts[2];
     }
     return refName.toLowerCase();
   }
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
@@ -468,4 +468,12 @@ public final class HiveUtils {
     Matcher ref = TAG.matcher(refName);
     return ref.matches();
   }
+
+  public static String getLowerCaseTableName(String refName) {
+    String[] refParts = refName.split("\\.");
+    if (refParts.length == 3 && SNAPSHOT_REF.matcher(refParts[2]).matches()) {
+      return (refParts[0].toLowerCase() + "." + refParts[1]).toLowerCase() + "." + refParts[2];
+    }
+    return refName.toLowerCase();
+  }
 }
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
@@ -1137,7 +1137,7 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
 
     ASTNode tableTree = (ASTNode) (tabref.getChild(0));
 
-    String tabIdName = getUnescapedName(tableTree).toLowerCase();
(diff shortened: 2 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -1137,7 +1137,7 @@ private String processTable(QB qb, ASTNode tabref) throws SemanticException {
 
     ASTNode tableTree = (ASTNode) (tabref.getChild(0));
 
-    String tabIdName = getUnescapedName(tableTree).toLowerCase();
+    String tabIdName = HiveUtils.getLowerCaseTableName(getUnescapedName(tableTree));
```
deniskuzZ, line 1140: what will happen if we simply remove toLowerCase() here?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_branch_with_different_case.q
```diff
@@ -0,0 +1,57 @@
+-- SORT_QUERY_RESULTS
```
deniskuzZ, line 1: i think it's better to do in UT, TestHiveIcebergBranchOperation, q tests are expensive

ayushtkn: Changed.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
ayushtkn: I am not very sure, that code is pretty old, so I preserved the old behaviour, it is used as a key in a map {{aliasToCTEs}}, not sure if it can break or lead to some perf penalties, if the query has same table with different cases or so....

iceberg/iceberg-handler/src/test/queries/positive/write_iceberg_branch.q
```diff
@@ -1,60 +0,0 @@
--- SORT_QUERY_RESULTS
```
deniskuzZ, line 1: why did you drop this q test? should we move the testcases from `iceberg_branch_with_different_case.q` in here?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
deniskuzZ: ok

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
```diff
@@ -468,4 +468,12 @@ public static Boolean isTableTag(String refName) {
     Matcher ref = TAG.matcher(refName);
     return ref.matches();
   }
+
+  public static String getLowerCaseTableName(String refName) {
+    String[] refParts = refName.split("\\.");
+    if (refParts.length == 3 && SNAPSHOT_REF.matcher(refParts[2]).matches()) {
+      return refParts[0].toLowerCase() + "." + refParts[1].toLowerCase() + "." + refParts[2];
```
deniskuzZ, line 475: use DatabaseName.CAT_DB_TABLE_SEPARATOR instead of "."

deniskuzZ, line 475: return (refParts[0] + "." + refParts[1]).toLowerCase() + "." + refParts[2];

deniskuzZ (approved):
LGTM +1, pending tests

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
```diff
@@ -468,4 +468,12 @@ public static Boolean isTableTag(String refName) {
     Matcher ref = TAG.matcher(refName);
     return ref.matches();
   }
+
+  public static String getLowerCaseTableName(String refName) {
+    String[] refParts = refName.split("\\.");
+    if (refParts.length == 3 && SNAPSHOT_REF.matcher(refParts[2]).matches()) {
+      return (refParts[0].toLowerCase() + "." + refParts[1]).toLowerCase() + "." + refParts[2];
```
deniskuzZ, line 475: redundant `.toLowerCase()` in refParts[0]

PR #5077, opened 2024-02-09
HIVE-24167: TPC-DS query 14 fails while generating plan for the filter

okumin:

Link Calcite's RelNodes with the same RelTreeSignature.

It is known that the current unification flow could fail when CTEs are materialized. I summarized the problem in [this gist](https://gist.github.com/okumin/b111fe0a911507bdf6a7204f49b9cb72).

This PR implements the option "Combine RelNodes with RelTreeSignature" in the gist.

See also [the PR where I implemented another option, "Don't pull groups of Operators by signatures"](https://github.com/apache/hive/pull/5037).

No.

No.

I added a new test case to reproduce the issue and restore query14 of TPC-DS which fails on the current master.


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveTableScan.java
```diff
@@ -202,6 +203,12 @@ public HiveTableScan copyIncludingTable(RelDataType newRowtype) {
   // Also include partition list key to trigger cost evaluation even if an
   // expression was already generated.
   @Override public RelWriter explainTerms(RelWriter pw) {
+    if (pw instanceof RelTreeSignatureWriter) {
+      return super.explainTerms(pw)
+          .item("tableScanTrait", this.tableScanTrait)
+          .itemIf("fromVersion", ((RelOptHiveTable) table).getHiveTableMD().getVersionIntervalFrom(),
+              isNotBlank(((RelOptHiveTable) table).getHiveTableMD().getVersionIntervalFrom()));
```
okumin, line 210: HiveTableScan doesn't retain the equivalents of `TableScanDesc#getPredicateString` or `TableScanDesc#getRowLimit`.
So, we can't unify ASTNodes or RelNodes based on `RelTreeSignature#equals`. Otherwise, Operators which should not belong to the same group will be unified later.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
```diff
@@ -277,8 +277,8 @@ private ASTNode convert() throws CalciteSemanticException {
     if (where != null) {
       ASTNode cond = where.getCondition().accept(new RexVisitor(schema, false, root.getCluster().getRexBuilder()));
       hiveAST.where = ASTBuilder.where(cond);
-      planMapper.link(cond, where);
       planMapper.link(cond, RelTreeSignature.of(where));
+      planMapper.link(cond, where);
```
okumin, line 281: We link RelTreeSignature first so that we can safely unify multiple filters.

ql/src/java/org/apache/hadoop/hive/ql/parse/TezCompiler.java
```diff
@@ -1046,11 +1046,16 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx, Obje
...
         // we will not see the actual incoming rowcount which was processed - so we may not use it for relNodes
-        TableScanOperator ts = (TableScanOperator) nd;
         if (ts.getConf().getPredicateString() != null) {
-          planMapper.link(ts, new OperatorStats.MayNotUseForRelNodes());
+          invalidateForRelNodes(ts, false);
+        }
+        // If sampling is configured, the table scan could be canceled in the middle. We avoid using runtime stats
+        // for HiveTableScan and its descendants as it is not pushed down to HiveTableScan RelNodes
+        if (ts.getConf().getRowLimit() >= 0) {
+          invalidateForRelNodes(ts, true);
```
okumin, line 1058: This might be too defensive as currently no RelNodes are created when TABLESAMPLE is used.

zhangbutao:
Sorry, I think I canceled wip by mistake when I was browsing the web...

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
kgyrtkirk: cool! :D

kgyrtkirk (approved):

okumin:
I applied two changes. I'm waiting for CI to finish.
1. Rebased this branch based on the latest master
2. Added https://github.com/apache/hive/pull/5077/commits/b11e3f88fef2e27934dc8389b8d1e80b6da5fc98
    - Now, we can populate correct stats of materialized tables thanks to #5089

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveTableScan.java
```diff
@@ -202,6 +203,12 @@ public HiveTableScan copyIncludingTable(RelDataType newRowtype) {
   // Also include partition list key to trigger cost evaluation even if an
   // expression was already generated.
   @Override public RelWriter explainTerms(RelWriter pw) {
+    if (pw instanceof RelTreeSignatureWriter) {
```
deniskuzZ, line 206: @okumin, why don't you define RelWriter var and extract the common initialization part to avoid code duplication.

ql/src/java/org/apache/hadoop/hive/ql/parse/TezCompiler.java
deniskuzZ: @okumin, is that possible/OK to link multiple MayNotUseForRelNodes to the same TableScanOperator

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveTableScan.java
okumin: Thanks for the suggestion. I have no objections. Could you please check if this update follows your suggestion?
https://github.com/apache/hive/pull/5077/commits/bd22d94e3cf1bc4375e7f91c91026ecc036b0ac6

ql/src/java/org/apache/hadoop/hive/ql/parse/TezCompiler.java
deniskuzZ: 👍

zabetak (changes requested):
Hey @okumin,

I was doing some testing with the proposed changes by setting the following properties in a wider scale.
```
set hive.optimize.cte.materialize.threshold=1;
set hive.optimize.cte.materialize.full.aggregate.only=false;
```
It seems that some queries still fail with `java.lang.RuntimeException: equivalence mapping violation`. 

For instance running cbo_query23.q with the above properties set raises an exception.
```
mvn test -Dtest=TestTezTPCDS30TBPerfCliDriver -Dqfile="cbo_query23.q" -Dtest.output.overwrite
```

Can you please look at the failures to see if this patch needs to be adapted. 

Also it would be nice to do a sanity run by setting the above properties temporarily for all qtests to see if the problem disappears completely or if there are still cases that need to be handled.

Assuming that there are still failures to be address please create additional qtests.

okumin:
> For instance running cbo_query23.q with the above properties set raises an exception.

Thanks. Predicate Push-Down and Partition Condition Remover cause the problem. I modified it on my local machine, and I will explain the details later.

> Also it would be nice to do a sanity run by setting the above properties temporarily for all qtests to see if the problem disappears completely or if there are still cases that need to be handled.

Thanks for the great suggestion! I definitely will do it. Now, I have run positive LLAP tests and TPC-DS, and I found that No. 64 of TPD-DS fails. I created another ticket because I found the problem can happen regardless of materialization or PlamMapper while I was minimizing the test query. I would be glad if the community could take a look.
https://github.com/apache/hive/pull/5207

I'm still testing the remaining qtests where CTEs are involved.

ql/src/test/queries/clientpositive/cbo_cte_materialization.q
```diff
@@ -0,0 +1,28 @@
+--! qt:dataset:src
+
+set hive.optimize.cte.materialize.threshold=1;
+set hive.optimize.cte.materialize.full.aggregate.only=false;
```
zabetak, line 4:
```
set hive.query.planmapper.link.relnodes=false;
```
By setting this property the test fails with the same `RuntimeException: equivalence mapping violation`.

zabetak (changes requested):
I had another look in this proposal relying on `RelTreeSignature` but it seems that it is not general enough to get rid of the problem for good. Changing the `hive.query.planmapper.link.relnodes` to false makes the new tests fail.

I was trying to come up with alternatives but don't have something viable yet.

okumin:
I also think the current one could fail somewhere especially if we follow minor options. Now, I have achieved some additional experience and knowledge, and may come up with the to-be approach. I will work on it tomorrow(I am unavailable because a tech event for Hive 4 takes place in my city today!) if I don't see very viable options here.

ql/src/test/queries/clientpositive/cbo_cte_materialization.q
okumin: I'm checking the whole execution and what the flag does. Disabling `hive.query.planmapper.link.relnodes`, no aux signature is created, and no merge across Operators doesn't happen. I feel it is not consistent with the description, `Whether to link Calcite nodes to runtime statistics`.
I guess the direct problem will be resolved if we skip linking RelNodes with Operators when `hive.query.planmapper.link.relnodes=false` is configured. That sounds more consistent with the description.

I'm putting my additional notes here. I tried to put some to-be solutions but it could not be very easy.
https://docs.google.com/document/d/1LCST23cSBZglBzjhnCqHlpcLv6xrXRBxU_wszSDAk9w/edit?usp=sharing

okumin: I guess `hive.query.planmapper.link.relnodes` can be implemented like this. @kgyrtkirk might have some more contexts.
https://github.com/okumin/hive/commit/cdab2c1941411caa9cef45c2821689982cb2dfe3

okumin:
While running negative tests with `hive.optimize.cte.materialize.threshold=1` and `hive.optimize.cte.materialize.full.aggregate.only=false`, I found another error. I created another ticket because it is also unrelated to the PlanMapper.
https://github.com/apache/hive/pull/5232

ql/src/test/queries/clientpositive/cbo_cte_materialization.q
okumin: I found this fails if we disable `hive.query.planmapper.link.relnodes`.

```
create table src (key string, value string);

set hive.cbo.fallback.strategy=NEVER;
set hive.query.planmapper.link.relnodes=false;
set hive.optimize.ppd=false;

EXPLAIN
SELECT * FROM src WHERE key = '5'
UNION ALL
SELECT * FROM src WHERE key = '5';
```

@zabetak As a potential workaround, I'm wondering if it makes sense to relax the validation and disable features with which PlanMapper is involved when "equivalence mapping violation" happens.
I presume PlanMapper succeeds in 99.9% cases as most qtests succeed. Currently, the remaining 0.1% fails and it sounds like an excessive penalty. We can keep the validation for qtests so that we can minimize the risk of degradation.
https://github.com/okumin/hive/commit/dd5be1543b9426f5d6ab94a37d5cdf8136c5fec7

dengzhhu653:
What's the current status of this PR? it looks like we have some problems with the testing.

okumin:
@dengzhhu653 @zabetak Thanks. I understand we need some more consensus on the expectations of HIVE-24167.

We know the current PR doesn't work with some sets of Hive parameters. I have a question here, "Should HIVE-24167 let any SQLs with any set of parameters successfully run?". If the answer is yes, I recommend [avoiding the current approach "let the entire query crush"](https://github.com/apache/hive/pull/5077#discussion_r1598642918) and fixing known problems incrementally in the following tickets because it is unrealistic to prove that no issue happens under arbitrary conditions. If the answer is no, we may need a consensus on what queries/parameters we should support in HIVE-24167.

PR #5085, opened 2024-02-14
HIVE-28078: TestTxnDbUtil should generate csv files when we query the metastore database

InvisibleProgrammer:
queryToString separates the columns with 3 spaces. It would improve the debugging experience if we have a method that shows the data in CSV format so that the output can be easily imported into a spreadsheet.

Two new method in TestTxnDbUtil: queryToCsv (with or without header). 

To easily be able to import the output into a sheet during a debugging session.

No

No

Manually

### DIFF
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestTxnDbUtil.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestTxnDbUtil.java
@@ -338,10 +338,24 @@ public final class TestTxnDbUtil {
       closeResources(conn, stmt, rs);
     }
   }
+
   public static String queryToString(Configuration conf, String query) throws Exception {
     return queryToString(conf, query, true);
   }
-  public static String queryToString(Configuration conf, String query, boolean includeHeader)
+
+  public static String queryToString(Configuration conf, String query, boolean includeHeader) throws Exception {
+    return queryToString(conf, query, includeHeader, "   ");
+  }
+
+  public static String queryToCsv(Configuration conf, String query) throws Exception {
+    return queryToString(conf, query, true, ",");
+  }
+
+  public static String queryToCsv(Configuration conf, String query, boolean includeHeader) throws Exception {
+    return queryToString(conf, query, includeHeader, ",");
+  }
+
+  public static String queryToString(Configuration conf, String query, boolean includeHeader, String columnSeparator)
       throws Exception {
     Connection conn = null;
     Statement stmt = null;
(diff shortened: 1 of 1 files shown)


standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestTxnDbUtil.java
```diff
@@ -370,6 +384,8 @@ public static String queryToString(Configuration conf, String query, boolean inc
     return sb.toString();
   }
 
+
```
abstractdog, line 387: nit: is this needed? we already have 1 line space between the methods

InvisibleProgrammer: Thank you for spotting. It was a mistake. I kept one line as a spacer between this and the next method.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestTxnDbUtil.java
```diff
@@ -338,10 +338,24 @@ public static int countQueryAgent(Configuration conf, String countQuery) throws
...
   public static String queryToString(Configuration conf, String query) throws Exception {
     return queryToString(conf, query, true);
   }
-  public static String queryToString(Configuration conf, String query, boolean includeHeader)
+
+  public static String queryToString(Configuration conf, String query, boolean includeHeader) throws Exception {
+    return queryToString(conf, query, includeHeader, "   ");
+  }
+
+  public static String queryToCsv(Configuration conf, String query) throws Exception {
```
abstractdog, line 350: what are the callers of this new queryToCsv method? I cannot see it used

also, basically, I like that you introduced new convenience util methods, however, I cannot see their value added to the original queryToString: what if I would like to create a CSV with ";" separator? currently, I should call queryToString with columnSeparator=";"
so I believe that queryToCsv doesn't add more value, it just express an assumption that CSV is always written with comma separators

InvisibleProgrammer: > what are the callers of this new queryToCsv method? I cannot see it used

As the summary says, the intention is to improve debugging experience. 
With the current queryToString method, if I want to analyze the output, my only choice if I want a header or not. But it separates the columns with 4 spaces. That is far from any usual. 
So in case I want to use it in a spreadsheet app, I need to manually convert 4 spaces to a separator character. After that, I can import it into a spreadsheet.
With that method, we can do it in one step. 

> also, basically, I like that you introduced new convenience util methods, however, I cannot see their value added to the original queryToString this time: what if I would like to create a CSV with ";" separator? currently, I should call queryToString with columnSeparator=";" despite the new util methods
so I believe that queryToCsv doesn't add more value, it just expresses an assumption that CSV is always written with comma separators (I know, CSV means comma separated, but still, I can see semicolon-separated stuff in many places)

In case you want to use a non-standard characther, like `;`, you can use the modified method that accepts the separator character. I choosed `,` as that is the original idea of CSV: to separate with colons.

PR #5086, opened 2024-02-14
HIVE-28076: Selecting data from a bucketed table with decimal column type throwing NPE.

mdayakar:
HIVE-28076: Selecting data from a bucketed table with decimal column type throwing NPE.

When _`hive.tez.bucket.pruning`_ is enabled, filters on bucket columns will be processed by filtering the splits against a bitset of included buckets. This will be done by `FixedBucketPruningOptimizer`. In the flow for finding the proper bucket number and enforcing precision&scale, getting the writable object inspector using `PrimitiveObjectInspectorFactory` by passing the primitive category which internally uses default typeinfo. For DECIMAL type the default decimal type info has 38,18 as precision and scale respectively which allows the integral part should have only 20 digits. If value contains more than 20 digits then its considered as OVERFLOW and returns null which resulting NPE.

To fix the issue, in `FixedBucketPruningOptimizer` while getting the writable object inspector using `PrimitiveObjectInspectorFactory`, passing typeinfo from the actual field object inspector which takes the decimal typeinfo from the actual field.

To fix the issue HIVE-28076.

No

No

q tests are added.
_mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=bucket_decimal_col_select.q -pl itests/qtest -Pitests_


kasakrisz (approved):
LGTM

PR #5087, opened 2024-02-15
HIVE-27405 - Throw out the detail error Invalid partition name to the…

tarak271:
… clients

When try to MSCK TABLE, if there is a directory that doesn't match the partition format, the query fails. However, it doesn't thrown out the detail error information to the client. Rethrowing exception coming from msck.repair call

Yes, change in error message thrown when there is an issue with msck repair command execution

No

Junit test case added

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
@@ -29,6 +29,7 @@ import org.apache.hadoop.hive.metastore.MsckInfo;
 import org.apache.hadoop.hive.metastore.PartitionManagementTask;
 import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.MetaException;
+import org.apache.hadoop.hive.metastore.api.MetastoreException;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.ql.ddl.DDLOperation;
@@ -51,7 +52,7 @@ public class MsckOperation extends DDLOperation<MsckDesc> {
   }
 
   @Override
-  public int execute() throws HiveException, IOException, TException {
+  public int execute() throws HiveException, IOException, TException, MetastoreException {
     try {
       Msck msck = new Msck(false, false);
       msck.init(Msck.getMsckConf(context.getDb().getConf()));
@@ -75,9 +76,9 @@ public class MsckOperation extends DDLOperation<MsckDesc> {
           desc.getFilterExp(), desc.getResFile(), desc.isRepairPartitions(),
           desc.isAddPartitions(), desc.isDropPartitions(), partitionExpirySeconds);
       return msck.repair(msckInfo);
-    } catch (MetaException e) {
+    } catch (MetaException | MetastoreException e) {
       LOG.error("Unable to create msck instance.", e);
-      return 1;
+      throw e;
     } catch (SemanticException e) {
       LOG.error("Msck failed.", e);
       return 1;
--- a/ql/src/test/results/clientnegative/msck_repair_1.q.out
+++ b/ql/src/test/results/clientnegative/msck_repair_1.q.out
@@ -21,4 +21,4 @@ POSTHOOK: Output: default@repairtable
 PREHOOK: query: MSCK TABLE default.repairtable
 PREHOOK: type: MSCK
 PREHOOK: Output: default@repairtable
-FAILED: Execution Error, return code 40000 from org.apache.hadoop.hive.ql.ddl.DDLTask. org.apache.hadoop.hive.metastore.api.MetastoreException: MSCK is missing partition columns under hdfs://### HDFS PATH ###
+#### A masked pattern was here ####
--- a/ql/src/test/results/clientnegative/msck_repair_1.q.out
+++ b/ql/src/test/results/clientnegative/msck_repair_1.q.out
@@ -21,4 +21,4 @@ POSTHOOK: Output: default@repairtable
 PREHOOK: query: MSCK TABLE default.repairtable
 PREHOOK: type: MSCK
 PREHOOK: Output: default@repairtable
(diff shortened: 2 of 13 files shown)


ql/src/test/results/clientnegative/table_nonprintable_negative.q.out
```diff
@@ -17,4 +17,4 @@ POSTHOOK: Output: default@table_external
 PREHOOK: query: msck repair table table_external
 PREHOOK: type: MSCK
 PREHOOK: Output: default@table_external
-FAILED: Execution Error, return code 1 from org.apache.hadoop.hive.ql.ddl.DDLTask
+FAILED: Execution Error, return code 40000 from org.apache.hadoop.hive.ql.ddl.DDLTask. Repair: Cannot add partition table_external:day=Foo due to invalid characters in the name
```
okumin, line 20: The new message is really helpful 👍

okumin (approved):

ql/src/test/results/clientnegative/msck_repair_1.q.out
```diff
@@ -21,4 +21,4 @@ POSTHOOK: Output: default@repairtable
 PREHOOK: query: MSCK TABLE default.repairtable
 PREHOOK: type: MSCK
 PREHOOK: Output: default@repairtable
-FAILED: Execution Error, return code 1 from org.apache.hadoop.hive.ql.ddl.DDLTask
+#### A masked pattern was here ####
```
zhangbutao, line 24: What's the real execption here?

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
```diff
@@ -75,9 +76,9 @@ public int execute() throws HiveException, IOException, TException {
           desc.getFilterExp(), desc.getResFile(), desc.isRepairPartitions(),
           desc.isAddPartitions(), desc.isDropPartitions(), partitionExpirySeconds);
       return msck.repair(msckInfo);
-    } catch (MetaException e) {
+    } catch (MetaException | MetastoreException e) {
       LOG.error("Unable to create msck instance.", e);
-      return 1;
+      throw e;
     } catch (SemanticException e) {
```
zhangbutao, line 82: Will client get the detail error information in case of `SemanticException`?

ql/src/test/results/clientnegative/msck_repair_1.q.out
tarak271: real exception is `FAILED: Execution Error, return code 40000 from org.apache.hadoop.hive.ql.ddl.DDLTask. org.apache.hadoop.hive.metastore.api.MetastoreException: MSCK is missing partition columns under file:/Users/tarak/IdeaProjects/tarak-Apache-Hive/itests/qtest/target/localfs/warehouse/repairtable/p1=c
`
since it contained file location details, it is getting masked in test case output as per [Masking rules](https://github.com/apache/hive/blob/5b76949da6fe65364a4e3766680871167131157f/itests/util/src/main/java/org/apache/hive/beeline/QFile.java#L336)

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
tarak271: Yes. Added test case msck_repair_8.q to see how semantic exception is handled

ql/src/test/results/clientnegative/msck_repair_2.q.out
```diff
@@ -21,4 +21,4 @@ POSTHOOK: Output: default@repairtable
 PREHOOK: query: MSCK TABLE default.repairtable
 PREHOOK: type: MSCK
 PREHOOK: Output: default@repairtable
-FAILED: Execution Error, return code 1 from org.apache.hadoop.hive.ql.ddl.DDLTask
+#### A masked pattern was here ####
```
tarak271, line 24: Real exception is 
`FAILED: Execution Error, return code 40000 from org.apache.hadoop.hive.ql.ddl.DDLTask. org.apache.hadoop.hive.metastore.api.MetastoreException: MSCK finds a file rather than a directory when it searches for file:/Users/tarak/IdeaProjects/tarak-Apache-Hive/itests/qtest/target/localfs/warehouse/repairtable/p1=c/datafile`

which is getting masked as it contained location of file

ql/src/test/results/clientnegative/msck_repair_3.q.out
```diff
@@ -21,4 +21,4 @@ POSTHOOK: Output: default@repairtable
 PREHOOK: query: MSCK TABLE default.repairtable
 PREHOOK: type: MSCK
 PREHOOK: Output: default@repairtable
-FAILED: Execution Error, return code 1 from org.apache.hadoop.hive.ql.ddl.DDLTask
+#### A masked pattern was here ####
```
tarak271, line 24: Real exception

`FAILED: Execution Error, return code 40000 from org.apache.hadoop.hive.ql.ddl.DDLTask. org.apache.hadoop.hive.metastore.api.MetastoreException: MSCK finds a file rather than a directory when it searches for file:/Users/tarak/IdeaProjects/tarak-Apache-Hive/itests/qtest/target/localfs/warehouse/repairtable/p1=c/datafile`

ql/src/test/results/clientnegative/msck_repair_4.q.out
```diff
@@ -21,4 +21,4 @@ POSTHOOK: Output: default@repairtable
 PREHOOK: query: MSCK REPAIR TABLE default.repairtable
 PREHOOK: type: MSCK
 PREHOOK: Output: default@repairtable
-FAILED: Execution Error, return code 1 from org.apache.hadoop.hive.ql.ddl.DDLTask
+#### A masked pattern was here ####
```
tarak271, line 24: Real exception is 
`FAILED: Execution Error, return code 40000 from org.apache.hadoop.hive.ql.ddl.DDLTask. org.apache.hadoop.hive.metastore.api.MetastoreException: Unexpected partition key p2 found at file:/Users/tarak/IdeaProjects/tarak-Apache-Hive/itests/qtest/target/localfs/warehouse/repairtable/p2=c
`

PR #5089, opened 2024-02-19
HIVE-28080: Propagate statistics from a source table to the materialized CTE

okumin:

Copy statistics of the source FileSinkOperator into the TableScanOperator, which scans a materialized CTE written by the FileSinkOperator.

Currently, Hive estimates the statistics of materialized CTEs are always empty. It causes OOM or significant slowdown.

No.

No.

I added a new test case. [The first commit explains the result with the current master branch](https://github.com/apache/hive/pull/5089/commits/ca6cae1ee2ecacd5fdf4291225edf1c79130a2b2).

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/Context.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/Context.java
@@ -49,7 +49,6 @@ import org.apache.hadoop.hive.common.TableName;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.ql.cleanup.CleanupService;
 import org.apache.hadoop.hive.ql.cleanup.SyncCleanupService;
-import org.apache.hadoop.hive.ql.exec.FileSinkOperator;
 import org.apache.hadoop.hive.ql.exec.TaskRunner;
 import org.apache.hadoop.hive.ql.exec.Utilities;
 import org.apache.hadoop.hive.ql.hooks.WriteEntity;
@@ -65,6 +64,7 @@ import org.apache.hadoop.hive.ql.parse.ExplainConfiguration.AnalyzeState;
 import org.apache.hadoop.hive.ql.parse.HiveParser;
 import org.apache.hadoop.hive.ql.parse.QB;
 import org.apache.hadoop.hive.ql.plan.LoadTableDesc;
+import org.apache.hadoop.hive.ql.plan.Statistics;
 import org.apache.hadoop.hive.ql.plan.mapper.EmptyStatsSource;
 import org.apache.hadoop.hive.ql.plan.mapper.PlanMapper;
 import org.apache.hadoop.hive.ql.plan.mapper.StatsSource;
@@ -142,7 +142,7 @@ public class Context {
   private AtomicInteger sequencer = new AtomicInteger();
 
   private final Map<String, Table> cteTables = new HashMap<String, Table>();
-  private final Map<TableName, FileSinkOperator> cteSources = new HashMap<>();
+  private final Map<TableName, Statistics> cteTableStats = new HashMap<>();
 
   // Keep track of the mapping from load table desc to the output and the lock
   private final Map<LoadTableDesc, WriteEntity> loadTableOutputMap =
@@ -1228,16 +1228,14 @@ public class Context {
     return cteTables.get(cteName);
   }
 
-  public void addMaterializedTable(String cteName, Table table) {
+  public void addMaterializedTable(String cteName, Table table, Statistics statistics) {
     cteTables.put(cteName, table);
+    cteTables.put(table.getFullyQualifiedName(), table);
+    cteTableStats.put(table.getFullTableName(), statistics);
   }
 
-  public FileSinkOperator getMaterializedTableSource(TableName tableName) {
-    return cteSources.get(tableName);
-  }
-
-  public void addMaterializedTableSource(TableName tableName, FileSinkOperator operator) {
-    cteSources.put(tableName, operator);
+  public Statistics getMaterializedTableStats(TableName tableName) {
+    return cteTableStats.get(tableName);
   }
 
   public AtomicInteger getSequencer() {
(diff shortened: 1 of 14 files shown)


okumin:
@kasakrisz I know you have expertise in this area. I'd be glad if you took a look. Thanks 👍

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -12143,6 +12148,27 @@ samplePredicate, true, new SampleDesc(ts.getNumerator(),
...
+            sourceColStatsList, colNames));
+      }
+      final List<ColStatistics> colStatsList = new ArrayList<>(sourceColStatsList.size());
+      for (int i = 0; i < sourceColStatsList.size(); i++) {
+        final ColStatistics colStats = sourceColStatsList.get(i).clone();
+        // FileSinkOperator stores column stats with internal names such as "_col1"
+        colStats.setColumnName(colNames.get(i));
+        colStatsList.add(colStats);
+      }
+      stats.setColumnStats(colStatsList);
```
kasakrisz, line 12168: What is the reason of cloning `ColStatistics` objects again?
`Statistics.clone()` does it too
https://github.com/apache/hive/blob/5b76949da6fe65364a4e3766680871167131157f/ql/src/java/org/apache/hadoop/hive/ql/plan/Statistics.java#L210-L216

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -8174,6 +8176,9 @@ protected Operator genFileSinkPlan(String dest, QB qb, Operator input)
 
     FileSinkOperator fso = (FileSinkOperator) output;
     fso.getConf().setTable(destinationTable);
+    if (destTableIsMaterialization) {
+      ctx.addMaterializedTableSource(destinationTable.getFullTableName(), fso);
```
kasakrisz, line 8180: Can this be merged with `Context.addMaterializedTable` and called in `materializeCTE`?
`SemanticAnalyzer` has `getSinkOp()` hence we can get the `FileSinkOperator` later from the analyzer instance.

Does the whole FileSinkOperator is needed?
```
Context.addMaterializedTable(String cteName, Table table, Statistics statistics);
```

kasakrisz (commented):
Thanks for the contribution, left comments about design.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
okumin: Just because `stats` could become temporarily inconsistent on `colStats.setColumnName`. Now, I think it is not a problem and I will remove the additional clone.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
okumin: That sounds cleaner!

> Does the whole FileSinkOperator is needed?

No. I stored the whole operator because stats of FSO are not computed yet at L8180. If we push it to the end of the `materializeCTE`, we can simply store only Statistics.

ql/src/java/org/apache/hadoop/hive/ql/Context.java
```diff
@@ -1225,8 +1228,14 @@ public Table getMaterializedTable(String cteName) {
     return cteTables.get(cteName);
   }
 
-  public void addMaterializedTable(String cteName, Table table) {
+  public void addMaterializedTable(String cteName, Table table, Statistics statistics) {
     cteTables.put(cteName, table);
+    cteTables.put(table.getFullyQualifiedName(), table);
```
okumin, line 1233: I moved this logic here for easiness, which means this pair is stored for SemanticAnalyzer. I would say the overhead is trivial and acceptable. Anyway, I hope most people use CalcitePlanner.

okumin:
@kasakrisz I modified a bit following the proposal. Please feel free to put additional comments if you have any points. I appreciate your quick review.

ql/src/test/queries/clientpositive/cte_mat_11.q
```diff
@@ -0,0 +1,15 @@
...
+  SELECT * FROM src
+),
+materialized_cte2 AS (
+  SELECT *
+  FROM materialized_cte1 a
+  JOIN materialized_cte1 b ON (a.key = b.key)
+)
+SELECT *
+FROM materialized_cte2 a
+JOIN materialized_cte2 b ON (a.key = b.key);
```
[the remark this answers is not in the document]
kasakrisz: I haven't dig in deeply what could be the cause of the failure. Maybe it can be a bug or it is something which Hive doesn't support and there are opportunities to make some improvement at this area or it is something which does not makes sense by SQL semantics. So it requires some investigation first. If it is a bug please file a new jira and include the repro steps.

About propagating stats: I think it is a good idea. I see some plan changes in existing tests so it seems that those are proving that the stats propagation happens. If you think we need some additional tests please feel free to add it. You can easily check whether CBO is passing or not by adding an `EXPLAIN CBO <query>` command, it should print the CBO plan.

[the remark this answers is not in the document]
okumin: @kasakrisz
You find very well! I didn't notice the possibility where my test case could be illegal. It will be probably invalid as far as I tested it with various parameters or other DBMS systems. I created the following ticket and am investigating more.
https://issues.apache.org/jira/browse/HIVE-28088

Additionally, I changed `cte_mat_11.q` so that it contains only valid SQLs. I am guessing HIVE-28088 is disjunctive with HIVE-28080, and we can review this PR. Please feel free to put any other concerns.
https://github.com/apache/hive/pull/5089/commits/961f27fce8c6be78bac8d44d70c25176a57a4fe1

Thanks a lot.

kasakrisz (approved):

okumin:
Thank you!

PR #5092, opened 2024-02-20
HIVE-28083: Enable HMS client cache and HMS query cache for Explain p…

soumyakanti3578:
…lans

Enable HMS caches for explain plans.

We use explain plans to measure compilation time sometimes.

No

No

Ran
`mvn test -Dtest=TestTezTPCDS30TBPerfCliDriver -Dtest.output.overwrite=true -Dqfile=cbo_query1.q`
and checked the logs to confirm that the Cache was getting used.

### DIFF
--- a/data/conf/perf/tpcds30tb/tez/hive-site.xml
+++ b/data/conf/perf/tpcds30tb/tez/hive-site.xml
@@ -18,6 +18,10 @@
 -->
 <configuration>
     <property>
+        <name>hive.metastore.client.cache.v2.recordStats</name>
+        <value>true</value>
+    </property>
+    <property>
         <name>hive.in.test</name>
         <value>true</value>
         <description>Internal marker for test. Used for masking env-dependent values</description>
--- a/data/conf/perf/tpcds30tb/tez/hive-site.xml
+++ b/data/conf/perf/tpcds30tb/tez/hive-site.xml
@@ -18,10 +18,6 @@
 -->
 <configuration>
     <property>
-        <name>hive.metastore.client.cache.v2.recordStats</name>
-        <value>true</value>
-    </property>
-    <property>
         <name>hive.in.test</name>
         <value>true</value>
         <description>Internal marker for test. Used for masking env-dependent values</description>
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMetaStoreClientWithLocalCache.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMetaStoreClientWithLocalCache.java
@@ -61,7 +61,6 @@ import org.apache.hadoop.hive.metastore.api.TableStatsResult;
 import org.apache.hadoop.hive.metastore.api.TableValidWriteIds;
 import org.apache.hadoop.hive.metastore.api.UniqueConstraintsRequest;
 import org.apache.hadoop.hive.metastore.api.UniqueConstraintsResponse;
-import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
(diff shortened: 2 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMetaStoreClientWithLocalCache.java
```diff
@@ -61,7 +61,6 @@
 import org.apache.hadoop.hive.metastore.api.TableValidWriteIds;
 import org.apache.hadoop.hive.metastore.api.UniqueConstraintsRequest;
 import org.apache.hadoop.hive.metastore.api.UniqueConstraintsResponse;
-import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
```
jfsii, line 64: Is this diff required?

ql/src/java/org/apache/hadoop/hive/ql/parse/ExplainSemanticAnalyzer.java
```diff
@@ -282,4 +282,11 @@ public boolean skipAuthorization() {
     }
     return super.skipAuthorization();
   }
+
```
jfsii, line 285: Does anything in the EXPLAIN output show if it is using cached metadata or not? If so, could we write a Q file test that ensures this continues to work?

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMetaStoreClientWithLocalCache.java
soumyakanti3578: Since `HiveMetaStoreClientWithLocalCache.java` is in HS2, I thought I might as well clean this up and use `HiveConf` instead of `MetastoreConf`. Let me know if this doesn't make a lot of sense, and I can revert it.

data/conf/perf/tpcds30tb/tez/hive-site.xml
```diff
@@ -17,6 +17,10 @@
    limitations under the License.
 -->
 <configuration>
+    <property>
```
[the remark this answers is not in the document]
soumyakanti3578: This is not strictly required. Adding this property enables logging whenever the cache is used. I can remove this.

ql/src/java/org/apache/hadoop/hive/ql/parse/ExplainSemanticAnalyzer.java
jfsii: Looking at the history of "HIVE_OPTIMIZE_HMS_QUERY_CACHE_ENABLED", I am not sure if there is a way to detect currently if a query's metadata is being served from the cache besides looking at the log file (and I would consider this outside of the scope of this PR but others might have differing opinion), so it likely would be difficult to do this via a qtest

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMetaStoreClientWithLocalCache.java
soumyakanti3578: This file was moved from `metastore-common` to `ql` in https://github.com/apache/hive/commit/f15de94c617c4566c87293479463cd90437beed5#diff-011304448070f76deed9090288839d16ee84ded1fb575a324f11a31bc6bab1e7, 
so I think it makes sense to use `HiveConf` instead of `MetastoreConf`

zabetak (approved):
LGTM! Will merge soon.

PR #5097, opened 2024-02-24
HIVE-28090: Correct desc of `hive.metastore.disallow.incompatible.col.type.changes` of `HiveConf.java`

zzzzming95:

correct desc of `hive.metastore.disallow.incompatible.col.type.changes`,  In fact the default value is true.

no

no need test


zhangbutao (approved):
LGTM +1

PR #5098, opened 2024-02-24
HIVE-28091: Remove invalid long datatype in ColumnStatsUpdateTask

zhangbutao:

Remove invalid long datatype in ColumnStatsUpdateTask::constructColumnStatsFromInput

IMO, Hive column does not support long data type. We should remove the incorrect data type in ColumnStatsUpdateTask.
https://github.com/apache/hive/blob/a348e5d86adc611e4f72d386815a21a96b667ab4/ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnStatsUpdateTask.java#L104-L106

In addition, the column stats related code blocks should be consistent with code in StatObjectConverter.java, which also does not have long type.

https://github.com/apache/hive/blob/a348e5d86adc611e4f72d386815a21a96b667ab4/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/StatObjectConverter.java#L378-L379

No

No

Existing UTs

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnStatsUpdateTask.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnStatsUpdateTask.java
@@ -102,8 +102,7 @@ public class ColumnStatsUpdateTask extends Task<ColumnStatsUpdateWork> {
 
     ColumnStatisticsData statsData = new ColumnStatisticsData();
 
-    if (columnType.equalsIgnoreCase("long")
-        || columnType.equalsIgnoreCase(serdeConstants.TINYINT_TYPE_NAME)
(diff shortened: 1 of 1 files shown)


deniskuzZ:
@zhangbutao , could we extract this check logic into some method to be consistent in multiple places where this check is used?

zhangbutao:
> @zhangbutao , could we extract this check logic into some method to be consistent in multiple places where this check is used?

It seems that it is not easy to extract this check logic.  `ColumnStatsUpdateTask::constructColumnStatsFromInput` is used for **update/write** table/partition stats, and `StatObjectConverter::getTableColumnStatisticsObj` is used for **get/read** table/partition stats. The two method has some different check & one is in `hive-exec` and another is in `standalone-metastore`.

deniskuzZ (approved):
+1

PR #5099, opened 2024-02-24
HIVE-28092: Clean up invalid exception thrown in MetaStoreClient

zhangbutao:

Clean up invalid exception thrown in MetaStoreClient

Lots of invalid exception were added in hms client methods, and this make the codes look like more redundant.
We can remove these invalid exception from method signature.

No

No

Existing UTs


deniskuzZ (approved):
LGTM. pending tests

PR #5100, opened 2024-02-25
HIVE-27850: Iceberg: Addendum 2: Set runAs user in CompactionInfo.

difin:

Setting CompactionInfo.runAs in Iceberg Compaction Service.

Without runAs set, execution fails during compilation phase on access authorization in internal environments. 
Besides that, if runAs is not set and Iceberg compaction throws any exception, without this fix there will be also another exception thrown in the catch block here:
https://github.com/apache/hive/blob/a348e5d86adc611e4f72d386815a21a96b667ab4/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Worker.java#L258

No

No

Testing with dev jars on local dev env and precommit tests.


deniskuzZ (approved):
LGTM +1

PR #5104, opened 2024-02-28
HIVE-28096: Fix the mismatched function names in HMSHandler audit log

wecharyu:
Normalize the `startFunction()` and `endFunction()` calls in `HMSHandler`:
1. Fix the mismatched function name.
2. Add the missed `startFunction()` or endFunction() calls.
3. Unify the function name, remove the `.(dot)` or additional suffix empty chars in some function names.

The mismatched function names would cause the counter and api cost metrics wrong.

No.

No.

Deploy and test the metrics.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -984,7 +984,7 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
 
   @Override
   public void alter_catalog(AlterCatalogRequest rqst) throws TException {
-    startFunction("alter_catalog " + rqst.getName());
+    startFunction("alter_catalog", ": " + rqst.getName());
     boolean success = false;
     Exception ex = null;
     RawStore ms = getMS();
@@ -1058,7 +1058,7 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
       ex = e;
       throw e;
     } finally {
-      endFunction("get_catalog", ret != null, ex);
+      endFunction("get_catalogs", ret != null, ex);
     }
     return new GetCatalogsResponse(ret == null ? Collections.emptyList() : ret);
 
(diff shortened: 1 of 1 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -1058,7 +1058,7 @@ public GetCatalogsResponse get_catalogs() throws MetaException {
       ex = e;
       throw e;
     } finally {
-      endFunction("get_catalog", ret != null, ex);
+      endFunction("get_catalogs", ret != null, ex);
```
zratkai, line 1061: This should be "alter_catalog" .

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -4392,19 +4392,17 @@ private List<Partition> createPartitionFolders(final List<Partition> partitionsT
...
     String catName = request.isSetCatName() ? request.getCatName() : getDefaultCatalog(conf);
     String dbName = request.getDbName();
     String tblName = request.getTblName();
     startTableFunction("add_partitions_req", catName, dbName, tblName);
 
+    AddPartitionsResult result = new AddPartitionsResult();
     Exception ex = null;
     try {
+      if (request.getParts().isEmpty()) {
+        return result;
```
zratkai, line 4404: This is not covered in the topic of thic PR: " Fix the mismatched function names".
Why do we need this here?

zratkai (commented):
Please review my finding.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: If we do not move this code into try clause, the `endFunction()` calls would be missed.

PR #5106, opened 2024-02-29
HIVE-28098: Fails to copy empty column statistics of materialized CTE

okumin:

Makes the stats population more defensive.
https://issues.apache.org/jira/browse/HIVE-28098

The current implementation fails when column stats are missing.

No, it doesn't.

No.

I added various test cases to cte_mat_11.q.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
@@ -1600,26 +1600,35 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
     LOG.info("{} will be materialized into {}", cteName, location);
     cte.source = analyzer;
 
-    ctx.addMaterializedTable(cteName, table, getMaterializedTableStats(analyzer.getSinkOp(), table));
+    ctx.addMaterializedTable(cteName, table, getMaterializedTableStats(analyzer.getSinkOp()));
 
     return table;
   }
 
-  static Statistics getMaterializedTableStats(Operator<?> sinkOp, Table table) {
+  protected Statistics getMaterializedTableStats(Operator<?> sinkOp) {
     final Statistics tableStats = sinkOp.getStatistics().clone();
-    final List<ColStatistics> sourceColStatsList = tableStats.getColumnStats();
-    final List<String> colNames = table.getCols().stream().map(FieldSchema::getName).collect(Collectors.toList());
-    if (sourceColStatsList.size() != colNames.size()) {
-      throw new IllegalStateException(String.format(
-          "The size of col stats must be equal to that of schema. Stats = %s, Schema = %s",
-          sourceColStatsList, colNames));
+    if (tableStats.getColumnStatsState() == Statistics.State.NONE || sinkOp.getNumParent() == 0) {
+      return tableStats;
     }
-    final List<ColStatistics> colStatsList = new ArrayList<>(sourceColStatsList.size());
-    for (int i = 0; i < sourceColStatsList.size(); i++) {
-      final ColStatistics colStats = sourceColStatsList.get(i);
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -1600,26 +1600,35 @@ Table materializeCTE(String cteName, CTEClause cte) throws HiveException {
...
-          sourceColStatsList, colNames));
-    }
-    final List<ColStatistics> colStatsList = new ArrayList<>(sourceColStatsList.size());
-    for (int i = 0; i < sourceColStatsList.size(); i++) {
-      final ColStatistics colStats = sourceColStatsList.get(i);
-      // FileSinkOperator stores column stats with internal names such as "_col1"
-      colStats.setColumnName(colNames.get(i));
-      colStatsList.add(colStats);
+    if (tableStats.getColumnStatsState() == Statistics.State.NONE || sinkOp.getNumParent() == 0) {
+      return tableStats;
```
okumin, line 1611: This is the main change.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -1600,26 +1600,35 @@ Table materializeCTE(String cteName, CTEClause cte) throws HiveException {
...
+    }
+
+    final List<String> parentColumnNames = sinkOp.getParentOperators().get(0).getSchema().getColumnNames();
```
okumin, line 1614: While I was debugging with a remote debugger, I found the order of `Statistics#getColumnStats` is non-deterministic as the underlying data structure is HashMap. That's why I modified how to map column names.
- https://github.com/apache/hive/blob/rel/release-4.0.0-beta-1/ql/src/java/org/apache/hadoop/hive/ql/plan/Statistics.java#L327C33-L327C44
- https://github.com/apache/hive/blob/rel/release-4.0.0-beta-1/ql/src/java/org/apache/hadoop/hive/ql/plan/Statistics.java#L211

okumin:
@kasakrisz I found the cases where HIVE-28080 doesn't work well in some complex patterns. Could you take a look when you have a chance?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -1600,26 +1600,35 @@ Table materializeCTE(String cteName, CTEClause cte) throws HiveException {
...
+      return tableStats;
+    }
+    final Map<String, String> mapping = new HashMap<>(parentColumnNames.size());
+    for (int i = 0; i < parentColumnNames.size(); i++) {
+      mapping.put(parentColumnNames.get(i), childColumnNames.get(i));
+    }
+
+    final List<ColStatistics> colStatsList = tableStats.getColumnStats();
+    for (ColStatistics colStats : colStatsList) {
+      if (mapping.containsKey(colStats.getColumnName())) {
```
kasakrisz, line 1629: Is it possible that we have a column statistic about a non existent column?
Maybe we should throw an exception or log it.

okumin: As far as I know, no, it isn't. The current implementation is a little defensive because there are countless methods which copy or mutate Statistics and it is unrealistic to verify all of them are correctly implemented.
I will add logging when we find unexpected states. Just a moment

okumin: I added a logging.
https://github.com/apache/hive/pull/5106/commits/111bcdf6e50b29ff1967c0a6e9a512dcf261c366

kasakrisz (approved):

PR #5112, opened 2024-03-02
HIVE-28099: Fix logging in HMS benchmarks

InvisibleProgrammer:
The logging is completely broken in HMS benchmarks. When we create a Log entry, it only outputs a format pattern string instead of the log message. 

Example output:

`%d [%thread] %-5level %logger - %msg`

TxnHandler refactor introduced Spring and it seems spring logging causes the issue. There are two changes in the PR: 
- exclude spring-jcl (https://stackoverflow.com/questions/50176843/log42-xml-is-not-picked-up-when-running-jmh-benchmark-from-test-directory)
- move from assembly to shading plugin to be able to run log4j2CacheTransformer (https://stackoverflow.com/questions/25065580/log4j2-custom-plugins-annotation-processing-with-maven-assembly-plugin)

I want to do some metastore performance benchmarks. It is pretty hard to do that without logging.

No

No

Manually test: `java    -jar ./target/hmsbench.jar -H localhost --savedata /tmp/benchdata --sanitize -N 100 -N 1000 -o bench_results_direct.csv -C -d testbench_http --params=100 -M "TestGetValidWriteIds" --runMode ACID`

### DIFF
--- a/standalone-metastore/metastore-tools/metastore-benchmarks/pom.xml
+++ b/standalone-metastore/metastore-tools/metastore-benchmarks/pom.xml
@@ -68,10 +68,10 @@
           <groupId>io.netty</groupId>
           <artifactId>*</artifactId>
         </exclusion>
-          <exclusion>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-jcl</artifactId>
-          </exclusion>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-jcl</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


zratkai:
LGTM.

standalone-metastore/metastore-tools/metastore-benchmarks/pom.xml
```diff
@@ -139,30 +143,46 @@
...
+                      </excludes>
+                    </filter>
+                  </filters>
+                </configuration>
               </execution>
             </executions>
+            <dependencies>
+              <dependency>
+                <groupId>com.github.edwgiz</groupId>
+                <artifactId>maven-shade-plugin.log4j2-cachefile-transformer</artifactId>
```
abstractdog, line 182: this is told to be part of official apache log4j2 instead of com.github.edwgiz, according to readme here: https://github.com/edwgiz/maven-shaded-log4j-transformer?tab=readme-ov-file
can we move to that?

abstractdog (approved):

PR #5115, opened 2024-03-04
HIVE-27172: (2.3) Add the HMS client connection timeout config

pan3793:

Backport HIVE-27172 to branch-2.3

Hive 2.3 is used in Apache Spark to communicate with HMS, while making the socket connect timeout configurable and independent with socket timeout is extremely useful.

In practice, we deploy multi HMS instances to achieve load balance and disaster recovery, we expect a short connect timeout in case the target HMS instance is under high load or has network issues, so that the HMS client could quickly failover to another one. While we expect a little bit longer socket timeout in cases some thrift invocations take a long time, e.g. a  HMS may take a dozen minutes to handle a table/partition request if the table is large or has million partitions.

Add a new configuration while the default value keeps the existing behavior.

No

New tests are added.

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/common/auth/HiveAuthUtils.java
+++ b/common/src/java/org/apache/hadoop/hive/common/auth/HiveAuthUtils.java
@@ -54,6 +54,11 @@ public class HiveAuthUtils {
     return getSSLSocketWithHttps(tSSLSocket);
   }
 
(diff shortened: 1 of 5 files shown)


pan3793:
cc @sunchao @wangyum

cxzl25 (approved):
LGTM

cxzl25:
cc @zhangbutao

pan3793:
Thanks, @zhangbutao and @cxzl25

PR #5117, opened 2024-03-05
HIVE-28105:Beeline: Show current database in prompt

tanishq-chugh:
In order to make the beeline better, this change suggests to show the current operating database name in the beeline prompt. It also provides the functionality to disable the same by using an argument ( --hideDbNameInPrompt ) while initialising beeline.

Why are the changes needed?
The changes are needed in order to make beeline better & more user-friendly.

Yes
Currently, Beeline prompt is as follows: 
            0: jdbc:hive2://link> use dbs;
            0: jdbc:hive2://link>

This change suggests prompt to be as follows:
            0: jdbc:hive2://link> use dbs;
            0: jdbc:hive2://link (dbs)> 

No

How was this patch tested?
Manual tests.
After making the necessary changes, we build the hive project again.
After manually testing some DDL and SQL commands, hive is working as expected.

### DIFF
--- a/beeline/src/java/org/apache/hive/beeline/BeeLine.java
+++ b/beeline/src/java/org/apache/hive/beeline/BeeLine.java
@@ -431,6 +431,13 @@ public class BeeLine implements Closeable {
         .withLongOpt("property-file")
         .withDescription("The file to read configuration properties from")
         .create());
+
+    // -hideDbNameInPrompt
+    options.addOption(OptionBuilder
+            .withLongOpt("hideDbNameInPrompt")
+            .withDescription("The flag to hide DB name in prompt")
+            .create());
+
   }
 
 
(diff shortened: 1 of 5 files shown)


InvisibleProgrammer:
Interesting idea. Personally, I like it. 
Also, I'm not sure if the change should be a new default.

tanishq-chugh:
@InvisibleProgrammer This functionality is targeted to be default in this PR because the aim is to make beeline usage more user-friendly especially for the new users, and they won't have enough knowledge to enable such things using arguments or any other way.
For Professionals, this might not be a necessity but will definitely help is minimising errors that sometimes arise due to wrong current database in their work.

ayushtkn:
this functionallity is already there just launch beeline like
```
beeline -u 'jdbc:hive2://localhost:10000/' --showDbInPrompt=true
```

```
ayushsaxena@ayushsaxena docker %  docker exec -it hiveserver2 beeline -u 'jdbc:hive2://localhost:10000/' --showDbInPrompt=true
Connecting to jdbc:hive2://localhost:10000/
Connected to: Apache Hive (version 4.1.0-SNAPSHOT)
Driver: Hive JDBC (version 4.1.0-SNAPSHOT)
Transaction isolation: TRANSACTION_REPEATABLE_READ
Beeline version 4.1.0-SNAPSHOT by Apache Hive
0: jdbc:hive2://localhost:10000/ (default)> use newDb;
No rows affected (0.076 seconds)
0: jdbc:hive2://localhost:10000/ (newDb)> 
```

tanishq-chugh:
@ayushtkn Then, Are there any reasons of why aren't we keeping this functionality active by default? Because i believe it would be better if we do so.

ayushtkn:
backward compatibility, so as to preserve old behaviour, that is how we operate generally while adding new stuff, those who want the new behaviour they can enable it via the params

PR #5118, opened 2024-03-05
HIVE-28037: Run multiple Qtest with Postgres

zratkai:
Running multiple qtests with Postgre will not fail with "Database does not exist: default". It keeps docker up and running between tests. 

Changes to be able to run multiple Qtests with Postgre.

Without this change only one Qtest can be run with Postgre.

No

No

Locally.

### DIFF
--- a/itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestMetaStoreHandler.java
+++ b/itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestMetaStoreHandler.java
@@ -98,14 +98,12 @@ public class QTestMetaStoreHandler {
   }
 
   public void beforeTest() throws Exception {
-    getRule().before();
-    if (!isDerby()) {// derby is handled with old QTestUtil logic (TxnDbUtil stuff)
-      getRule().install();
+    if (isDerby()) {
+      getRule().before();
     }
   }
 
   public void afterTest(QTestUtil qt) throws Exception {
-    getRule().after();
 
     // special qtest logic, which doesn't fit quite well into Derby.after()
     if (isDerby()) {
--- a/itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CoreCliDriver.java
+++ b/itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CoreCliDriver.java
@@ -92,6 +92,7 @@ public class CoreCliDriver extends CliAdapter {
   @AfterClass
   public void shutdown() throws Exception {
     qt.shutdown();
+    metaStoreHandler.getRule().after();
   }
 
   @Override


itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestMetaStoreHandler.java
```diff
@@ -98,14 +98,12 @@ private String getDbTypeConfString() {// "ORACLE", "MYSQL", "MSSQL", "POSTGRES"
   }
 
   public void beforeTest() throws Exception {
-    getRule().before();
-    if (!isDerby()) {// derby is handled with old QTestUtil logic (TxnDbUtil stuff)
-      getRule().install();
+    if (isDerby()) {
```
InvisibleProgrammer, line 101: If I understood it correctly, `DatabaseRule.before` initialises the docker instance. So that change says docker will initialised once for the whole test suite, not once per test case.
And we have to keep it for Derby as it sets the test directory (`test.tmp.dir`). 

Am I correct?

InvisibleProgrammer (commented):
Could you please add some explanation about the change? 
That description is a little bit confusing: 
```Change-Id: I2c37973af59f69d17de733d69f184fe83bcc3091```

File: itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestMetaStoreHandler.java [hunk 2]
```diff
@@ -98,14 +98,12 @@ private String getDbTypeConfString() {// "ORACLE", "MYSQL", "MSSQL", "POSTGRES"
   }
 
   public void beforeTest() throws Exception {
-    getRule().before();
```
InvisibleProgrammer, line 101: That method is used in other places. And at least at one place that doesn't run in the precommit tests suite. For example, I'm pretty sure `TestSchemaToolForMetastore` doesn't run with precommits. I'm not sure about the others.

zratkai: beforeTest method is only used in CliAdapter class, so I don't understand your concerns here.

itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestMetaStoreHandler.java
zratkai: Yes, since if the Postgre would be reinitialized ,then the tests would fail as it is mentioned in the ticket ("Database does not exist: default")

InvisibleProgrammer:
Short version: LGTM. 

Long version: I checked it out with a PR that uses postgresql driver as a default one to check what happens if we want to run all the qtestst with postgres. 
Test run: http://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5122/2/tests/
PR: https://github.com/apache/hive/pull/5122
Diff for the failing tests with only output diff (details below): https://github.com/apache/hive/compare/master...InvisibleProgrammer:hive:HIVE-28037 

The PR itself reached it's goal: it had no problem with running multiple qtests with postgres. 


## Different type of groups of test result

### Minor output differences
There were tests that failed but with minor output diff differences. Typically, with having columns in an output in different order or with whitespace differences. I would consider them as a pass. We cannot address those issues as the difference comes from the differences between derby and postgres engine. 
Some examples: 
- default_constraint - unique constraints are shown in different order
- create_with_constraints - key columns in different order
- drop_partitions_filter.q - Some columns in different order, duplicate key error has different output

### Cannot run the tests with explicitly defined postgres driver as the tests are designed to run with postgres: 
There are some tests with a comment like `--!qt:database:postgres:q_test_country_table.sql` 
Those tests are failing with `unable to start docker container` if we add the `postgres` db parameter to the test execution. I don't think it worth investigating as they are designed to run with postgres and no other driver. 
Some examples: 
- qt_database_postgres.q
- jdbc_partition_table_pruned_pcolumn.q

### Failed test cases that worth to fix: 
- https://issues.apache.org/jira/browse/HIVE-28150: Fix: cannot run Kudu and Blobstorage qtests with postgres driver as driver not found
- https://issues.apache.org/jira/browse/HIVE-28151: Fix escape1.q and escape2.q tests with postgres driver
- Fix partition_column_names_with_leading_and_trailing_spaces.q with postgres driver

InvisibleProgrammer (approved):

deniskuzZ (approved):
LGTM +1

deniskuzZ:
@zratkai please check jira id HIVE-28037, looks like it doesn't exist

deniskuzZ:
@zratkai, @InvisibleProgrammer could we please update instructions in https://cwiki.apache.org/confluence/display/Hive/HiveDeveloperFAQ#HiveDeveloperFAQ-HowdoIruntheclientpositive/clientnegativeunittests?

PR #5123, opened 2024-03-08
HIVE-28077: Iceberg: Major QB Compaction on partition level

difin:

Adding support for compacting a specific Iceberg partition

Until now it was possible to only compact whole the table at once. If a table is partitioned, it would compact all partitions. 
Now adding a new feature to compact a single specified partition.

Ability to compact Iceberg tables on partition level.

No

New q-test was added that compacts a partitioned table partition-by-partition.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -1928,18 +1928,18 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   }
 
   @Override
-  public org.apache.commons.lang3.tuple.Pair<Boolean, ErrorMsg> isEligibleForCompaction(
+  public Optional<ErrorMsg> isEligibleForCompaction(
       org.apache.hadoop.hive.ql.metadata.Table table, Map<String, String> partitionSpec) {
     if (partitionSpec != null) {
       Table icebergTable = IcebergTableUtil.getTable(conf, table.getTTable());
       if (hasUndergonePartitionEvolution(icebergTable)) {
-        return org.apache.commons.lang3.tuple.Pair.of(false, ErrorMsg.COMPACTION_PARTITION_EVOLUTION);
+        return Optional.of(ErrorMsg.COMPACTION_PARTITION_EVOLUTION);
       }
       if (!isIdentityPartitionTable(table)) {
-        return org.apache.commons.lang3.tuple.Pair.of(false, ErrorMsg.COMPACTION_NON_IDENTITY_PARTITION_SPEC);
+        return Optional.of(ErrorMsg.COMPACTION_NON_IDENTITY_PARTITION_SPEC);
       }
     }
-    return org.apache.commons.lang3.tuple.Pair.of(true, null);
+    return Optional.empty();
   }
 
   @Override
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -57,12 +57,14 @@ import org.apache.hadoop.hive.conf.Constants;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
 import org.apache.hadoop.hive.metastore.HiveMetaHook;
+import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.ColumnStatistics;
 import org.apache.hadoop.hive.metastore.api.ColumnStatisticsObj;
 import org.apache.hadoop.hive.metastore.api.EnvironmentContext;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.InvalidObjectException;
 import org.apache.hadoop.hive.metastore.api.LockType;
+import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
 import org.apache.hadoop.hive.metastore.utils.MetaStoreServerUtils;
 import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
@@ -1918,6 +1920,45 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
         table.currentSnapshot().allManifests(table.io()).parallelStream()
         .map(ManifestFile::partitionSpecId)
         .anyMatch(id -> id < table.spec().specId());
+  }
+
+  private boolean isIdentityPartitionTable(org.apache.hadoop.hive.ql.metadata.Table table) {
+    return getPartitionTransformSpec(table).stream().map(TransformSpec::getTransformType)
+        .allMatch(type -> type == TransformSpec.TransformType.IDENTITY);
+  }
+
+  @Override
+  public org.apache.commons.lang3.tuple.Pair<Boolean, ErrorMsg> isEligibleForCompaction(
+      org.apache.hadoop.hive.ql.metadata.Table table, Map<String, String> partitionSpec) {
+    if (partitionSpec != null) {
+      Table icebergTable = IcebergTableUtil.getTable(conf, table.getTTable());
+      if (hasUndergonePartitionEvolution(icebergTable)) {
+        return org.apache.commons.lang3.tuple.Pair.of(false, ErrorMsg.COMPACTION_PARTITION_EVOLUTION);
+      }
+      if (!isIdentityPartitionTable(table)) {
+        return org.apache.commons.lang3.tuple.Pair.of(false, ErrorMsg.COMPACTION_NON_IDENTITY_PARTITION_SPEC);
+      }
+    }
+    return org.apache.commons.lang3.tuple.Pair.of(true, null);
+  }
+
+  @Override
+  public List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table,
+      Map<String, String> partitionSpec) throws SemanticException {
+    return getPartitionNames(table, partitionSpec).stream()
+        .map(partName -> new DummyPartition(table, partName, partitionSpec)).collect(Collectors.toList());
+  }
+
+  @Override
+  public Partition getPartition(org.apache.hadoop.hive.ql.metadata.Table table,
+      Map<String, String> partitionSpec) throws SemanticException {
+    validatePartSpec(table, partitionSpec);
+    try {
+      String partName = Warehouse.makePartName(partitionSpec, false);
+      return new DummyPartition(table, partName, partitionSpec);
+    } catch (MetaException e) {
+      throw new SemanticException("Unable to construct name for dummy partition due to: ", e);
+    }
   }
 
   /**
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
@@ -20,15 +20,26 @@
 package org.apache.iceberg.mr.hive.compaction;
 
 import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
+import org.apache.hadoop.hive.metastore.Warehouse;
+import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.ql.Context.RewritePolicy;
 import org.apache.hadoop.hive.ql.DriverUtils;
+import org.apache.hadoop.hive.ql.metadata.Hive;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
+import org.apache.hadoop.hive.ql.metadata.HiveUtils;
 import org.apache.hadoop.hive.ql.session.SessionState;
 import org.apache.hadoop.hive.ql.txn.compactor.CompactorContext;
 import org.apache.hadoop.hive.ql.txn.compactor.QueryCompactor;
+import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
 import org.apache.hive.iceberg.org.apache.orc.storage.common.TableName;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -44,22 +55,51 @@ public class IcebergMajorQueryCompactor extends QueryCompactor  {
     Map<String, String> tblProperties = context.getTable().getParameters();
     LOG.debug("Initiating compaction for the {} table", compactTableName);
 
-    String compactionQuery = String.format("insert overwrite table %s select * from %<s",
-        compactTableName);
+    HiveConf conf = new HiveConf(context.getConf());
+    String partSpec = context.getCompactionInfo().partName;
+    String compactionQuery;
 
-    SessionState sessionState = setupQueryCompactionSession(context.getConf(),
-        context.getCompactionInfo(), tblProperties);
-    HiveConf.setVar(context.getConf(), ConfVars.REWRITE_POLICY, RewritePolicy.ALL_PARTITIONS.name());
+    if (partSpec == null) {
+      HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.ALL_PARTITIONS.name());
+      compactionQuery = String.format("insert overwrite table %s select * from %<s", compactTableName);
+    } else {
+      Hive db = Hive.get(conf);
+      org.apache.hadoop.hive.ql.metadata.Table table = db.getTable(context.getTable().getDbName(),
+          context.getTable().getTableName());
+      Map<String, String> partSpecMap = new LinkedHashMap<>();
+      Warehouse.makeSpecFromName(partSpecMap, new Path(partSpec), null);
+
+      List<FieldSchema> partitionKeys = table.getStorageHandler().getPartitionKeys(table);
+      List<String> partValues = partitionKeys.stream().map(
+          fs -> String.join("=", HiveUtils.unparseIdentifier(fs.getName()),
+              TypeInfoUtils.convertStringToLiteralForSQL(partSpecMap.get(fs.getName()),
+                  ((PrimitiveTypeInfo) TypeInfoUtils.getTypeInfoFromTypeString(fs.getType())).getPrimitiveCategory())
+          )
+      ).collect(Collectors.toList());
+
+      String queryFields = table.getCols().stream()
+          .map(FieldSchema::getName)
+          .filter(col -> !partSpecMap.containsKey(col))
+          .collect(Collectors.joining(","));
+
+      compactionQuery = String.format("insert overwrite table %1$s partition(%2$s) select %4$s from %1$s where %3$s",
+          compactTableName,
+          StringUtils.join(partValues, ","),
+          StringUtils.join(partValues, " and "),
+          queryFields);
+    }
+
+    SessionState sessionState = setupQueryCompactionSession(conf, context.getCompactionInfo(), tblProperties);
+
     try {
-      DriverUtils.runOnDriver(context.getConf(), sessionState, compactionQuery);
+      DriverUtils.runOnDriver(conf, sessionState, compactionQuery);
       LOG.info("Completed compaction for table {}", compactTableName);
+      return true;
     } catch (HiveException e) {
-      LOG.error("Error doing query based {} compaction", RewritePolicy.ALL_PARTITIONS.name(), e);
-      throw new RuntimeException(e);
+      LOG.error("Failed compacting table {}", compactTableName, e);
+      throw e;
     } finally {
       sessionState.setCompaction(false);
     }
-
-    return true;
   }
 }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
@@ -41,6 +41,7 @@ import org.apache.hadoop.hive.ql.txn.compactor.QueryCompactor;
 import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
 import org.apache.hive.iceberg.org.apache.orc.storage.common.TableName;
+import org.apache.iceberg.relocated.com.google.common.base.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -63,8 +64,7 @@ public class IcebergMajorQueryCompactor extends QueryCompactor  {
       HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.ALL_PARTITIONS.name());
       compactionQuery = String.format("insert overwrite table %s select * from %<s", compactTableName);
     } else {
-      Hive db = Hive.get(conf);
-      org.apache.hadoop.hive.ql.metadata.Table table = db.getTable(context.getTable().getDbName(),
+      org.apache.hadoop.hive.ql.metadata.Table table = Hive.get(conf).getTable(context.getTable().getDbName(),
           context.getTable().getTableName());
       Map<String, String> partSpecMap = new LinkedHashMap<>();
       Warehouse.makeSpecFromName(partSpecMap, new Path(partSpec), null);
@@ -90,13 +90,15 @@ public class IcebergMajorQueryCompactor extends QueryCompactor  {
     }
 
     SessionState sessionState = setupQueryCompactionSession(conf, context.getCompactionInfo(), tblProperties);
+    String compactionTarget = "table " + compactTableName +
+        (Strings.nullToEmpty(partSpec).isEmpty() ? "" : ", partition " + partSpec);
 
     try {
       DriverUtils.runOnDriver(conf, sessionState, compactionQuery);
-      LOG.info("Completed compaction for table {}", compactTableName);
+      LOG.info("Completed compaction for {}", compactionTarget);
       return true;
     } catch (HiveException e) {
-      LOG.error("Failed compacting table {}", compactTableName, e);
+      LOG.error("Failed compacting {}", compactionTarget, e);
       throw e;
     } finally {
       sessionState.setCompaction(false);
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
@@ -41,7 +41,6 @@ import org.apache.hadoop.hive.ql.txn.compactor.QueryCompactor;
 import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
 import org.apache.hive.iceberg.org.apache.orc.storage.common.TableName;
-import org.apache.iceberg.relocated.com.google.common.base.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -90,8 +89,8 @@ public class IcebergMajorQueryCompactor extends QueryCompactor  {
     }
 
     SessionState sessionState = setupQueryCompactionSession(conf, context.getCompactionInfo(), tblProperties);
-    String compactionTarget = "table " + compactTableName +
-        (Strings.nullToEmpty(partSpec).isEmpty() ? "" : ", partition " + partSpec);
+    String compactionTarget = "table " + HiveUtils.unparseIdentifier(compactTableName) +
+        (partSpec != null ? ", partition " + HiveUtils.unparseIdentifier(partSpec) : "");
 
     try {
       DriverUtils.runOnDriver(conf, sessionState, compactionQuery);
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
@@ -95,7 +95,7 @@ public final class PartitionUtils {
       throws SemanticException {
     Partition partition;
     try {
-      partition = db.getPartition(table, partitionSpec, false);
+      partition = db.getPartition(table, partitionSpec);
     } catch (Exception e) {
       throw new SemanticException(toMessage(ErrorMsg.INVALID_PARTITION, partitionSpec), e);
     }
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
@@ -39,6 +39,7 @@ import org.apache.hadoop.hive.ql.metadata.Partition;
 import org.apache.hadoop.hive.ql.metadata.Table;
 import org.apache.hadoop.hive.ql.txn.compactor.CompactorUtil;
 
+import org.apache.commons.lang3.tuple.Pair;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Map;
@@ -60,6 +61,13 @@ public class AlterTableCompactOperation extends DDLOperation<AlterTableCompactDe
     Table table = context.getDb().getTable(desc.getTableName());
     if (!AcidUtils.isTransactionalTable(table) && !AcidUtils.isNonNativeAcidTable(table)) {
       throw new HiveException(ErrorMsg.NONACID_COMPACTION_NOT_SUPPORTED, table.getDbName(), table.getTableName());
+    }
+    
+    if (table.getStorageHandler() != null) {
+      Pair<Boolean, ErrorMsg> result = table.getStorageHandler().isEligibleForCompaction(table, desc.getPartitionSpec());
+      if (!result.getLeft()) {
+        throw new HiveException(result.getRight(), table.getDbName(), table.getTableName());
+      }
     }
 
     Map<String, org.apache.hadoop.hive.metastore.api.Partition> partitionMap =
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
@@ -39,7 +39,6 @@ import org.apache.hadoop.hive.ql.metadata.Partition;
 import org.apache.hadoop.hive.ql.metadata.Table;
 import org.apache.hadoop.hive.ql.txn.compactor.CompactorUtil;
 
-import org.apache.commons.lang3.tuple.Pair;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Map;
@@ -64,9 +63,9 @@ public class AlterTableCompactOperation extends DDLOperation<AlterTableCompactDe
     }
     
     if (table.getStorageHandler() != null) {
-      Pair<Boolean, ErrorMsg> result = table.getStorageHandler().isEligibleForCompaction(table, desc.getPartitionSpec());
-      if (!result.getLeft()) {
-        throw new HiveException(result.getRight(), table.getDbName(), table.getTableName());
+      Optional<ErrorMsg> error = table.getStorageHandler().isEligibleForCompaction(table, desc.getPartitionSpec());
+      if (error.isPresent()) {
+        throw new HiveException(error.get(), table.getDbName(), table.getTableName());
       }
     }
 
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
@@ -751,6 +751,12 @@ public interface HiveStorageHandler extends Configurable {
     return true;
   }
 
+  /**
+   * Checks if a given table and partition specifications are eligible for compaction.
+   * @param table {@link org.apache.hadoop.hive.ql.metadata.Table} table metadata stored in Hive Metastore
+   * @param partitionSpec Map of Strings {@link java.util.Map} partition specification
+   * @return Optional of ErrorMsg {@link org.apache.hadoop.hive.ql.ErrorMsg}
+   */
   default Optional<ErrorMsg> isEligibleForCompaction(org.apache.hadoop.hive.ql.metadata.Table table,
       Map<String, String> partitionSpec) {
     throw new UnsupportedOperationException("Storage handler does not support validating eligibility for compaction");
@@ -783,16 +789,30 @@ public interface HiveStorageHandler extends Configurable {
             "for a table.");
   }
 
+  /**
+   * Returns partition based on table and partition specification.
+   * @param table {@link org.apache.hadoop.hive.ql.metadata.Table} table metadata stored in Hive Metastore
+   * @param partitionSpec Map of Strings {@link java.util.Map} partition specification
+   * @return Partition {@link org.apache.hadoop.hive.ql.metadata.Partition} 
+   * @throws SemanticException {@link org.apache.hadoop.hive.ql.parse.SemanticException} 
+   */
+  default Partition getPartition(org.apache.hadoop.hive.ql.metadata.Table table, Map<String, String> partitionSpec)
+      throws SemanticException {
+    throw new UnsupportedOperationException("Storage handler does not support getting partition for a table.");
+  }
+
+  /**
+   * Returns a list of partitions based on table and partial partition specification.
+   * @param table {@link org.apache.hadoop.hive.ql.metadata.Table} table metadata stored in Hive Metastore
+   * @param partitionSpec Map of Strings {@link java.util.Map} partition specification
+   * @return List of Partitions {@link org.apache.hadoop.hive.ql.metadata.Partition}
+   * @throws SemanticException {@link org.apache.hadoop.hive.ql.parse.SemanticException}
+   */
   default List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table, 
       Map<String, String> partitionSpec) throws SemanticException {
     throw new UnsupportedOperationException("Storage handler does not support getting partitions for a table.");
   }
 
-  default Partition getPartition(org.apache.hadoop.hive.ql.metadata.Table table, Map<String, String> partSpec) 
-      throws SemanticException {
-    throw new UnsupportedOperationException("Storage handler does not support getting partition for a table.");
-  }
-  
   default boolean supportsMergeFiles() {
     return false;
   }
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collections;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hadoop.conf.Configurable;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.hive.common.classification.InterfaceAudience;
@@ -750,6 +751,11 @@ public interface HiveStorageHandler extends Configurable {
     return true;
   }
 
+  default Pair<Boolean, ErrorMsg> isEligibleForCompaction(org.apache.hadoop.hive.ql.metadata.Table table,
+      Map<String, String> partitionSpec) {
+    throw new UnsupportedOperationException("Storage handler does not support validating eligibility for compaction");
+  }
+
   default List<String> getPartitionNames(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
       Map<String, String> partitionSpec) throws SemanticException {
     throw new UnsupportedOperationException("Storage handler does not support getting partitions " +
@@ -777,6 +783,16 @@ public interface HiveStorageHandler extends Configurable {
             "for a table.");
   }
 
+  default List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table, 
+      Map<String, String> partialPartSpec) throws SemanticException {
+    throw new UnsupportedOperationException("Storage handler does not support getting partitions for a table.");
+  }
+
+  default Partition getPartition(org.apache.hadoop.hive.ql.metadata.Table table, Map<String, String> partSpec) 
+      throws SemanticException {
+    throw new UnsupportedOperationException("Storage handler does not support getting partition for a table.");
+  }
+  
   default boolean supportsMergeFiles() {
     return false;
   }
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
@@ -24,7 +24,6 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collections;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hadoop.conf.Configurable;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.hive.common.classification.InterfaceAudience;
@@ -75,6 +74,7 @@ import org.apache.hadoop.mapred.OutputFormat;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 
 /**
@@ -751,7 +751,7 @@ public interface HiveStorageHandler extends Configurable {
     return true;
   }
 
-  default Pair<Boolean, ErrorMsg> isEligibleForCompaction(org.apache.hadoop.hive.ql.metadata.Table table,
+  default Optional<ErrorMsg> isEligibleForCompaction(org.apache.hadoop.hive.ql.metadata.Table table,
       Map<String, String> partitionSpec) {
     throw new UnsupportedOperationException("Storage handler does not support validating eligibility for compaction");
   }
@@ -784,7 +784,7 @@ public interface HiveStorageHandler extends Configurable {
   }
 
   default List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table, 
-      Map<String, String> partialPartSpec) throws SemanticException {
+      Map<String, String> partitionSpec) throws SemanticException {
     throw new UnsupportedOperationException("Storage handler does not support getting partitions for a table.");
   }
 
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
@@ -1187,6 +1187,11 @@ class MetaStoreDirectSql {
       MetastoreDirectSqlUtils.timingTrace(doTrace, queryText, start, queryTime);
     }
     String partIds = getIdListForIn(partitions.keySet());
+
+    if (partIds.isEmpty()) {
+      return orderedResult;
+    }
+
     // Now get all the one-to-many things. Start with partitions.
     MetastoreDirectSqlUtils
         .setPartitionParametersWithFilter(PARTITION_PARAMS, convertMapNullsToEmptyStrings, pm,
--- a/common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
+++ b/common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
@@ -480,6 +480,8 @@ public enum ErrorMsg {
   INVALID_METADATA_TABLE_NAME(10430, "Invalid metadata table name {0}.", true),
   TABLE_META_REF_NOT_SUPPORTED(10431, "Table Meta Ref extension is not supported for table {0}.", true),
   COMPACTION_REFUSED(10432, "Compaction request for {0}.{1}{2} is refused, details: {3}.", true),
+  COMPACTION_PARTITION_EVOLUTION(10438, "Compaction for {0}.{1} on partition level is not allowed on a table that has undergone partition evolution", true),
+  COMPACTION_NON_IDENTITY_PARTITION_SPEC(10439, "Compaction for {0}.{1} is not supported on the table with non-identity partition spec", true),
   CBO_IS_REQUIRED(10433,
           "The following functionality requires CBO (" + HiveConf.ConfVars.HIVE_CBO_ENABLED.varname + "): {0}", true),
   CTLF_UNSUPPORTED_FORMAT(10434, "CREATE TABLE LIKE FILE is not supported by the ''{0}'' file format", true),
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
@@ -85,7 +85,7 @@ public class IcebergTableUtil {
     return getTable(configuration, properties, skipCache);
   }
 
-  static Table getTable(Configuration configuration, org.apache.hadoop.hive.metastore.api.Table hmsTable) {
+  public static Table getTable(Configuration configuration, org.apache.hadoop.hive.metastore.api.Table hmsTable) {
     return getTable(configuration, hmsTable, false);
   }
 
--- a/iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition.q
@@ -1,0 +1,69 @@
+-- SORT_QUERY_RESULTS
+-- Mask neededVirtualColumns due to non-strict order
+--! qt:replace:/(\s+neededVirtualColumns:\s)(.*)/$1#Masked#/
+-- Mask the totalSize value as it can have slight variability, causing test flakiness
+--! qt:replace:/(\s+totalSize\s+)\S+(\s+)/$1#Masked#$2/
+-- Mask random uuid
+--! qt:replace:/(\s+uuid\s+)\S+(\s*)/$1#Masked#$2/
+-- Mask a random snapshot id
+--! qt:replace:/(\s+current-snapshot-id\s+)\S+(\s*)/$1#Masked#/
+-- Mask added file size
+--! qt:replace:/(\S\"added-files-size\\\":\\\")(\d+)(\\\")/$1#Masked#$3/
+-- Mask total file size
+--! qt:replace:/(\S\"total-files-size\\\":\\\")(\d+)(\\\")/$1#Masked#$3/
+-- Mask removed file size
+--! qt:replace:/(\S\"removed-files-size\\\":\\\")(\d+)(\\\")/$1#Masked#$3/
+-- Mask current-snapshot-timestamp-ms
+--! qt:replace:/(\s+current-snapshot-timestamp-ms\s+)\S+(\s*)/$1#Masked#$2/
+--! qt:replace:/(MAJOR\s+succeeded\s+)[a-zA-Z0-9\-\.\s+]+(\s+manual)/$1#Masked#$2/
+-- Mask compaction id as they will be allocated in parallel threads
+--! qt:replace:/^[0-9]/#Masked#/
+
+set hive.llap.io.enabled=true;
+set hive.vectorized.execution.enabled=true;
+set hive.optimize.shared.work.merge.ts.schema=true;
+
+create table ice_orc_wo_evo (
+    first_name string, 
+    last_name string
+ )
+partitioned by (dept_id bigint, 
+                city string, 
+                registration_date date)
+stored by iceberg stored as orc 
+tblproperties ('format-version'='2');
+
+insert into ice_orc_wo_evo VALUES ('fn1','ln1',1,'London','2024-03-11');
+insert into ice_orc_wo_evo VALUES ('fn2','ln2',1,'London','2024-03-11');
+insert into ice_orc_wo_evo VALUES ('fn3','ln3',1,'London','2024-03-11');
+insert into ice_orc_wo_evo VALUES ('fn4','ln4',1,'London','2024-03-11');
+insert into ice_orc_wo_evo VALUES ('fn5','ln5',2,'Paris','2024-02-16');
+insert into ice_orc_wo_evo VALUES ('fn6','ln6',2,'Paris','2024-02-16');
+insert into ice_orc_wo_evo VALUES ('fn7','ln7',2,'Paris','2024-02-16');
+
+update ice_orc_wo_evo set last_name = 'ln1a' where first_name='fn1';
+update ice_orc_wo_evo set last_name = 'ln2a' where first_name='fn2';
+update ice_orc_wo_evo set last_name = 'ln3a' where first_name='fn3';
+update ice_orc_wo_evo set last_name = 'ln4a' where first_name='fn4';
+update ice_orc_wo_evo set last_name = 'ln5a' where first_name='fn5';
+update ice_orc_wo_evo set last_name = 'ln6a' where first_name='fn6';
+update ice_orc_wo_evo set last_name = 'ln7a' where first_name='fn7';
+
+delete from ice_orc_wo_evo where last_name in ('ln1a', 'ln2a', 'ln7a');
+
+select * from ice_orc_wo_evo;
+describe formatted ice_orc_wo_evo;
+
+explain alter table ice_orc_wo_evo PARTITION (dept_id=1, city='London', registration_date='2024-03-11') COMPACT 'major' and wait;
+alter table ice_orc_wo_evo PARTITION (dept_id=1, city='London', registration_date='2024-03-11') COMPACT 'major' and wait;
+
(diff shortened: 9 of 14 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2232,7 +2232,9 @@ public static enum ConfVars {
         "If this is set to true the URI for auth will have the default location masked with DEFAULT_TABLE_LOCATION"),
     HIVE_ICEBERG_ALLOW_DATAFILES_IN_TABLE_LOCATION_ONLY("hive.iceberg.allow.datafiles.in.table.location.only", false,
         "If this is set to true, then all the data files being read should be withing the table location"),
-
+    HIVE_ICEBERG_COMPACTION_PARTITION("hive.iceberg.compaction.partition", "",
```
deniskuzZ, line 2235: why do we need HiveConf for that?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -544,14 +549,20 @@ private void commitWrite(Table table, String branchName, Long snapshotId, long s
...
-      RewritePolicy rewritePolicy) {
+      RewritePolicy rewritePolicy, String partitionSpec) {
     Preconditions.checkArgument(results.deleteFiles().isEmpty(), "Can not handle deletes with overwrite");
     if (!results.dataFiles().isEmpty()) {
       Transaction transaction = table.newTransaction();
       if (rewritePolicy == RewritePolicy.ALL_PARTITIONS) {
         DeleteFiles delete = transaction.newDelete();
         delete.deleteFromRowFilter(Expressions.alwaysTrue());
         delete.commit();
+      } else if (rewritePolicy == RewritePolicy.SINGLE_PARTITION) {
```
deniskuzZ, line 560: should it be `rewritePolicy != RewritePolicy.ALL_PARTITIONS`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -544,14 +549,20 @@ private void commitWrite(Table table, String branchName, Long snapshotId, long s
...
+        try {
+          HiveIcebergMetaHook.preTruncateTable(table, Arrays.asList(partitionSpec), transaction.newDelete());
```
deniskuzZ, line 562: should we rename to `truncateTable`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -2021,4 +2022,34 @@ public List<Partition> getPartitionsByExpr(org.apache.hadoop.hive.ql.metadata.Ta
...
+        .map(partName -> new DummyPartition(hmsTable, partName))
+        .collect(Collectors.toList());
+  }
+
+  private void validatePartitionName(org.apache.hadoop.hive.ql.metadata.Table hmsTable, Map<String, String> partSpec) {
+    try {
+      List<String> partNames = getPartitionNames(hmsTable, partSpec);
+
+      if (partNames.isEmpty()) {
+        throw new RuntimeException("Failed getting partitions from partition spec");
```
deniskuzZ, line 2049: do we have some more specific exceptions in Hive?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +52,68 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+      LOG.error("Error doing query based {} compaction", rewritePolicy.name(), e);
       throw new RuntimeException(e);
     } finally {
       sessionState.setCompaction(false);
     }
 
     return true;
   }
+
+  public String partDataToSQL(PartitionData partitionData, String partSpec, String delimiter) {
```
deniskuzZ, line 89: should it be public?

ql/src/java/org/apache/hadoop/hive/ql/Compiler.java
```diff
@@ -100,6 +100,7 @@ public QueryPlan compile(String rawCommand, boolean deferClose) throws CommandPr
     boolean parsed = false;
     QueryPlan plan = null;
     try {
+      SessionState.getSessionConf().set(Context.operationNameConf, org.apache.commons.lang3.StringUtils.EMPTY);
```
deniskuzZ, line 103: why do we need this?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/AbstractBaseAlterTableAnalyzer.java
```diff
@@ -98,6 +100,7 @@ protected void addInputsOutputsAlterTable(TableName tableName, Map<String, Strin
 
       if (AlterTableUtils.isFullPartitionSpec(table, partitionSpec)) {
         // Fully specified partition spec
+        SessionState.getSessionConf().set(Context.operationNameConf, op.getName());
```
deniskuzZ, line 103: can we get the operation from Context.getOperation?

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
```diff
@@ -99,7 +99,16 @@ private void getNextBatch() {
           batchCounter++;
         }
         try {
-          if (isAuthRequired) {
+          if (table.getStorageHandler() != null && table.getStorageHandler().alwaysUnpartitioned()) {
+            List<Partition> partitions = new ArrayList<>();
```
deniskuzZ, line 103: why do we need new List var?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
difin: Thanks, replaced with a constant in Context class.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: If it were `rewritePolicy != RewritePolicy.ALL_PARTITIONS`, it would also be true for the case `rewritePolicy != RewritePolicy.DEFAULT`, but the code in the else block should only be executed for compacting a specific partition, not in general case.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Changed to private.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java
```diff
@@ -932,7 +934,9 @@ public void setDbName(String databaseName) {
   }
 
   public List<FieldSchema> getPartitionKeys() {
-    return tTable.getPartitionKeys();
+    return getStorageHandler() != null && getStorageHandler().alwaysUnpartitioned() &&
+        SessionState.getSessionConf().get(Context.operationNameConf, StringUtils.EMPTY).equals(AlterTableType.COMPACT.getName()) ?
```
deniskuzZ, line 938: why do you need to check operationType?

ql/src/java/org/apache/hadoop/hive/ql/Compiler.java
difin: Done, removed checking of operation type as part of compact operation.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/AbstractBaseAlterTableAnalyzer.java
difin: Removed checking of operation type.

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
difin: I removed this change, it is not needed anymore.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java
difin: Removed this change, not needed anymore.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -640,7 +640,13 @@ public void preTruncateTable(org.apache.hadoop.hive.metastore.api.Table table, E
       throws MetaException {
     this.catalogProperties = getCatalogProperties(table);
     this.icebergTable = Catalogs.loadTable(conf, catalogProperties);
-    Map<String, PartitionField> partitionFieldMap = icebergTable.spec().fields().stream()
+    truncatePartitionBySpec(icebergTable, partNames, icebergTable.newDelete());
+    context.putToProperties("truncateSkipDataDeletion", "true");
```
deniskuzZ, line 644: could we please use `TRUNCATE_SKIP_DATA_DELETION` constant here.

ql/src/java/org/apache/hadoop/hive/ql/Context.java
```diff
@@ -117,6 +117,7 @@ public class Context {
   protected String optimizedSql;
   protected String calcitePlan;
   protected String cmd = "";
+  public static String compactPartition = "hive.iceberg.compaction.partition";
```
deniskuzZ, line 120: that shouldn't belong here, maybe AlterTableCompactOperation?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
deniskuzZ: can't we use the DEFAULT? rewrite policy enum is either all or not all (custom set)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -462,7 +463,12 @@ private void commitTable(FileIO io, ExecutorService executor, OutputTable output
           .map(x -> x.getJobConf().get(ConfVars.REWRITE_POLICY.varname))
           .orElse(RewritePolicy.DEFAULT.name()));
 
-      commitOverwrite(table, branchName, startTime, filesForCommit, rewritePolicy);
+      String partitionSpec = outputTable.jobContexts.stream()
```
deniskuzZ, line 466: that should be a List here instead of mapping in commitOverwrite

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +53,68 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
     } finally {
       sessionState.setCompaction(false);
     }
 
     return true;
   }
+
+  private String partDataToSQL(PartitionData partitionData, String partSpec, String delimiter) {
+    StringBuilder sb = new StringBuilder();
+    List<String> values = Arrays
```
deniskuzZ, line 92: should we use `Warehouse#makeSpecFromName(partName)`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +53,68 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+        .stream(partSpec.split("/"))
+        .map(x -> x.split("=")[1])
+        .collect(Collectors.toList());
+
+    for (int i = 0; i < partitionData.size(); ++i) {
```
deniskuzZ, line 97: why not use `Stream::collect(Collectors.joining())`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/AbstractBaseAlterTableAnalyzer.java
```diff
@@ -95,11 +98,16 @@ protected void addInputsOutputsAlterTable(TableName tableName, Map<String, Strin
       // ReadEntity as no lock.
       re.noLockNeeded();
       inputs.add(re);
+      Partition partition;
 
       if (AlterTableUtils.isFullPartitionSpec(table, partitionSpec)) {
         // Fully specified partition spec
-        Partition part = PartitionUtils.getPartition(db, table, partitionSpec, true);
-        outputs.add(new WriteEntity(part, writeType));
+        if (table.getStorageHandler() != null && table.getStorageHandler().alwaysUnpartitioned()) {
```
deniskuzZ, line 105: why not to make `PartitionUtils.getPartition(db, table, partitionSpec, true)` generic. it should return partitions based on table type

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -8810,7 +8810,8 @@ private Operator genConversionSelectOperatorByAddPartition(String dest, QB qb, O
     List<ExprNodeDesc> expressions = new ArrayList<>(outColumnCnt);
 
     AtomicBoolean convert = new AtomicBoolean(false);
-    if (inColumnCnt + staticPartitionColumnCnt != outColumnCnt) {
+    if (inColumnCnt + staticPartitionColumnCnt != outColumnCnt &&
```
deniskuzZ, line 8813: i don't think we should just ignore that check for iceberg, can we update the `staticPartitionColumnCnt`?
cc @kasakrisz

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
difin: Changed to the constant. I will check if that is used in case of Iceberg.

ql/src/java/org/apache/hadoop/hive/ql/Context.java
difin: Done, moved to AlterTableCompactOperation.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/AbstractBaseAlterTableAnalyzer.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Replaced the stream operation with a call to `Warehouse#makeSpecFromName(partName)`

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
difin: I modified the check for Iceberg case as following, is it fine now?

```
    if (table.getStorageHandler() != null && table.getStorageHandler().alwaysUnpartitioned()) {
      if (inColumnCnt != outColumnCnt || staticPartitionColumnCnt > inColumnCnt){
        String reason = "Always unpartitioned table " + dest + " has " + outColumnCnt
            + " columns, but query has " + inColumnCnt + " columns.";
        throw new SemanticException(ASTErrorUtils.getMsg(
            ErrorMsg.TARGET_TABLE_COLUMN_MISMATCH.getMsg(),
            qb.getParseInfo().getDestForClause(dest), reason));
      }
    }
```

deniskuzZ: previously this check was done regardless of `alwaysUnpartitioned` flag and now we moved it inside the if?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -21,45 +21,159 @@
...
+import org.apache.iceberg.mr.hive.IcebergBranchExec;
+import org.apache.iceberg.mr.hive.IcebergTableUtil;
+import org.apache.iceberg.types.Type;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class IcebergMajorQueryCompactor extends QueryCompactor  {
 
   private static final Logger LOG = LoggerFactory.getLogger(IcebergMajorQueryCompactor.class.getName());
+  private static final String branchName = "compaction";
```
deniskuzZ, line 51: finals should be in uppercase

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -21,45 +21,159 @@
...
+  private String compactTableName;
```
deniskuzZ, line 52: it shouldn't have a state

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -21,45 +21,159 @@
...
+        table.refresh();
+        Long createBranchSnapshotId = HiveIcebergOutputCommitter.getSnapshotId(table, "branch_" + branchName);
+
+        PartitionData partitionData = DataFiles.data(table.spec(), partSpec);
+        String truncateQuery = String.format("truncate table %s.branch_%s partition(%s)",
+            compactTableName, branchName, partDataToSQL(partitionData, partSpec, ","));
+        executeHiveQuery(context, truncateQuery, sessionState,
+            String.format("truncating partition %s of table %s.branch_%s", partSpec, compactTableName,
+                branchName));
+        table.refresh();
```
deniskuzZ, line 93: why is that needed?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -485,7 +485,7 @@ private void commitTable(FileIO io, ExecutorService executor, OutputTable output
     }
   }
 
-  private Long getSnapshotId(Table table, String branchName) {
+  public static Long getSnapshotId(Table table, String branchName) {
```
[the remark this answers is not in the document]
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -21,45 +21,159 @@
...
+
+        compactionQuery = String.format("insert overwrite table %s select * from %<s", compactTableName);
+
+        executeHiveQuery(context, compactionQuery, sessionState,
+            String.format("compacting table %s", compactTableName));
+      } else {
+        Table table = IcebergTableUtil.getTable(context.getConf(), context.getTable());
+        AlterTableSnapshotRefSpec.DropSnapshotRefSpec dropBranchSpec =
+            new AlterTableSnapshotRefSpec.DropSnapshotRefSpec(branchName, true);
+        IcebergBranchExec.dropBranch(table, dropBranchSpec);
```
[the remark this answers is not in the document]
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
deniskuzZ: i don't see constant used here

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -569,14 +575,25 @@ private void commitWrite(Table table, String branchName, Long snapshotId, long s
    * @param rewritePolicy The rewrite policy to use for the insert overwrite commit
    */
   private void commitOverwrite(Table table, String branchName, long startTime, FilesForCommit results,
-      RewritePolicy rewritePolicy) {
+      RewritePolicy rewritePolicy, String partitionSpec) {
     Preconditions.checkArgument(results.deleteFiles().isEmpty(), "Can not handle deletes with overwrite");
     if (!results.dataFiles().isEmpty()) {
       Transaction transaction = table.newTransaction();
       if (rewritePolicy == RewritePolicy.ALL_PARTITIONS) {
```
deniskuzZ, line 582: do we need a rewrite policy, since we can check for `partitionSpec.isEmpty()`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -569,14 +575,25 @@ private void commitWrite(Table table, String branchName, Long snapshotId, long s
...
         DeleteFiles delete = transaction.newDelete();
         delete.deleteFromRowFilter(Expressions.alwaysTrue());
         delete.commit();
+        LOG.info("Overwrite commit: truncated all partitions for table: {} with {} file(s)",
+            table, results.dataFiles().size());
+      } else if (!partitionSpec.isEmpty()) {
```
deniskuzZ, line 588: why do we need this? that won't help with partition evolution in any case

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +52,68 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
-      LOG.error("Error doing query based {} compaction", RewritePolicy.ALL_PARTITIONS.name(), e);
-      throw new RuntimeException(e);
+      LOG.error("Failed compacting table {}", compactTableName, e);
+      throw e;
     } finally {
       sessionState.setCompaction(false);
     }
+  }
+
+  private String partDataToSQL(PartitionData partitionData, String partSpec, String delimiter)
```
deniskuzZ, line 86: you should use logic from `truncatePartitionBySpec` as you can't override partition with non identity spec

if (!TransformSpec.TransformType.IDENTITY.equals(transformType)) {
              throw new MetaException(
                  String.format("Partition transforms are not supported in compact operation: %s", entry.getKey()));
            }

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +52,68 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+            partitionData.getType(i).typeId() == Type.TypeID.TIME ||
+            partitionData.getType(i).typeId() == Type.TypeID.TIMESTAMP ||
+            partitionData.getType(i).typeId() == Type.TypeID.BINARY) {
+          quoteOpt = "'";
+        }
+
+        sb.append(partitionData.getSchema().getFields().get(i).name())
+            .append("=")
+            .append(quoteOpt)
+            .append(partSpecMap.get(partitionData.getPartitionType().fields().get(i).name()))
```
deniskuzZ, line 110: should we use `partitionData.getSchema().getFields()` here as well

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: This is not for partition evolution case. This is for compaction on partition level without evolution. The partition needs to be truncated before adding compacted files.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +51,69 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+    } else {
+      Table table = IcebergTableUtil.getTable(conf, context.getTable());
+      PartitionData partitionData = DataFiles.data(table.spec(), partSpec);
+      compactionQuery = String.format("insert overwrite table %1$s partition(%2$s) select * from %1$s where %3$s",
+          compactTableName, partDataToSQL(partitionData, partSpec, ","),
+          partDataToSQL(partitionData, partSpec, " and "));
+    }
+
+    SessionState sessionState = setupQueryCompactionSession(conf, context.getCompactionInfo(), tblProperties);
+    sessionState.setConf(conf);
```
deniskuzZ, line 70: won't that override the conf setup in setupQueryCompactionSession
````
    Util.disableLlapCaching(conf);
    conf.set(HiveConf.ConfVars.HIVE_QUOTEDID_SUPPORT.varname, "column");
    conf.setBoolVar(HiveConf.ConfVars.HIVE_SERVER2_ENABLE_DOAS, true);
    conf.setBoolVar(HiveConf.ConfVars.HIVE_HDFS_ENCRYPTION_SHIM_CACHE_ON, false);
    Util.overrideConfProps(conf, compactionInfo, tblProperties);
````

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java:70]: check MajorQueryCompactor for example, it just clones the conf and passes that to session
````
 HiveConf conf = new HiveConf(hiveConf);
    /*
     * For now, we will group splits on tez so that we end up with all bucket files,
     * with same bucket number in one map task.
     */
    conf.set(HiveConf.ConfVars.SPLIT_GROUPING_MODE.varname, CompactorUtil.COMPACTOR);
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 27]
```diff
@@ -2127,4 +2127,26 @@ public List<FileStatus> getMergeTaskInputFiles(Properties properties) throws IOE
   public MergeTaskProperties getMergeTaskProperties(Properties properties) {
     return new IcebergMergeTaskProperties(properties);
   }
+
+  @Override
+  public boolean isIdentityPartitionSpec(org.apache.hadoop.hive.metastore.api.Table table,
```
deniskuzZ, line 2132: why should it be in API, could we move this to TableUtils?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: When the cloned conf is passed to the session, and the session is not null, the conf is not set in the session and any changes defined in it get lost. I found it during debugging when trying to understand why in a q-test that has 2 compaction calls the second compaction command is not working.

```
  public static SessionState setUpSessionState(HiveConf conf, String user, boolean doStart) {
    SessionState sessionState = SessionState.get();
    if (sessionState == null) {
      // Note: we assume that workers run on the same threads repeatedly, so we can set up
      //       the session here and it will be reused without explicitly storing in the worker.
      sessionState = new SessionState(conf, user);
      if (doStart) {
        // TODO: Required due to SessionState.getHDFSSessionPath. Why wasn't it required before?
        sessionState.setIsHiveServerQuery(true);
        SessionState.start(sessionState);
      }
      SessionState.setCurrentSessionState(sessionState);
    }
    return sessionState;
  }
```

deniskuzZ: how does it work then for ACID compactions? shouldn't we fix that in setUpSessionState and always update the conf?

difin: Yes, I think we should. By looking the at code, I think it behaves the same for ACID compactions, meaning that conf is not updates when using the same session..

deniskuzZ:
```` 
else {
      sessionState.setConf(conf);
    }
````

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java [hunk 28]
```diff
@@ -94,10 +97,21 @@ public static ExprNodeGenericFuncDesc makeUnaryPredicate(String fn, ExprNodeDesc
...
+    if (table.getStorageHandler() != null && table.getStorageHandler().alwaysUnpartitioned()) {
+      table.getStorageHandler().validatePartSpec(table, partitionSpec);
+      try {
+        String partName = Warehouse.makePartName(partitionSpec, false);
+        partition = new DummyPartition(table, partName, partitionSpec);
+      } catch (MetaException e) {
+        throw new SemanticException("Unable to construct name for dummy partition due to: ", e);
+      }
+    }
+    else {
```
deniskuzZ, line 109: please move it to prev line with `}`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
```diff
@@ -108,11 +122,17 @@ public static Partition getPartition(Hive db, Table table, Map<String, String> p
...
       boolean throwException) throws SemanticException {
     List<Partition> partitions;
-    try {
-      partitions = partitionSpec == null ? db.getPartitions(table) : db.getPartitions(table, partitionSpec);
-    } catch (Exception e) {
-      throw new SemanticException(toMessage(ErrorMsg.INVALID_PARTITION, partitionSpec), e);
-    }
+      if (table.getStorageHandler() != null && table.getStorageHandler().alwaysUnpartitioned()) {
+        partitions = new ArrayList<>();
+        partitions.add(getPartition(db, table, partitionSpec, throwException));
```
deniskuzZ, line 127: that is a diff API, it supposed to return all partitions matching spec, not just 1

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -133,7 +140,7 @@ private List<Partition> getPartitions(Table table, AlterTableCompactDesc desc, D
       }
     } else {
       Map<String, String> partitionSpec = desc.getPartitionSpec();
-      partitions = context.getDb().getPartitions(table, partitionSpec);
+      partitions.addAll(PartitionUtils.getPartitions(context.getDb(), table, partitionSpec, false));
```
deniskuzZ, line 143: why don't we use `partitionUtils.getPartition()` - it should return just 1 partition

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
deniskuzZ: we shouldn't overload existing code it's already complicated. is it possible to extract this column check in some validateColumns method?
cc @kasakrisz

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1187,6 +1187,11 @@ private List<Partition> getPartitionsByQuery(String catName, String dbName, Stri
       MetastoreDirectSqlUtils.timingTrace(doTrace, queryText, start, queryTime);
     }
     String partIds = getIdListForIn(partitions.keySet());
+
+    if (partIds.isEmpty()) {
```
deniskuzZ, line 1191: do we still need this after you've added iceberg partition list functionality in PartitionUtils?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
difin: Yes, it is used in case of compactions.
It is called from AlterTableCompactOperation.
It is used to parallelize compaction of partitioned table - when partition spec is not provided it creates one compaction request per partition. For now it is not done for Iceberg, but will be useful in next PRs in which we will implement it for Iceberg too.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: This is existing functionality which I didn't change.
I think it calls getPartitions on purpose to validate that only one partition is retrieved with the provided spec.
On the next lines it checks that only 1 partition was found for the partition spec and throws an exception otherwise.
```
      if (partitions.size() > 1) {
        throw new HiveException(ErrorMsg.TOO_MANY_COMPACTION_PARTITIONS);
      } else if (partitions.isEmpty()) {
        throw new HiveException(ErrorMsg.INVALID_PARTITION_SPEC);
      }
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
difin: Yes, because the issue with Metastore Direct SQL was encountered after I had added Iceberg partition list functionality in PartitionUtils and also I think PartitionUtils is not used by this metastore code because it is found in hive-ql and it's not a dependency for the metastore modules.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1110,8 +1110,9 @@ public void validateSinkDesc(FileSinkDesc sinkDesc) throws SemanticException {
         // If the table is empty we don't have any danger that some data can get lost.
         return;
       }
-      if (RewritePolicy.fromString(conf.get(ConfVars.REWRITE_POLICY.varname, RewritePolicy.DEFAULT.name())) ==
-          RewritePolicy.ALL_PARTITIONS) {
+      RewritePolicy rewritePolicy = RewritePolicy.fromString(conf.get(ConfVars.REWRITE_POLICY.varname,
+          RewritePolicy.DEFAULT.name()));
+      if (rewritePolicy == RewritePolicy.ALL_PARTITIONS || rewritePolicy == RewritePolicy.SINGLE_PARTITION) {
```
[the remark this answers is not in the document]
difin: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: not needed anymore, removed

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
difin: Moved this validation to a separate method.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -8815,6 +8815,23 @@ These props are now enabled elsewhere (see commit diffs).  It would be better in
...
     SessionState.get().getConf().set(AcidUtils.CONF_ACID_KEY, "true");
   }
+  
+  private void validateColumnCounts(String dest, Table table, Map<String, String> partitionSpec, 
+      int inColumnCnt, int outColumnCnt) throws SemanticException {
+
+    long staticPartitionColumnCnt = partitionSpec != null ?
+        partitionSpec.values().stream().filter(Objects::nonNull).count() : 0;
+
+    if (inColumnCnt + staticPartitionColumnCnt != outColumnCnt &&
```
deniskuzZ, line 8825: could we extract `table.getStorageHandler() != null && table.getStorageHandler().alwaysUnpartitioned()` into local var `alwaysUnpartitioned`?

deniskuzZ:
````
if (inColumnCnt + staticPartitionColumnCnt != outColumnCnt 
    && (!alwaysUnpartitioned || inColumnCnt != outColumnCnt) ) {
...
}
````

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java [hunk 34]
```diff
@@ -62,6 +65,12 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
       throw new HiveException(ErrorMsg.NONACID_COMPACTION_NOT_SUPPORTED, table.getDbName(), table.getTableName());
     }
 
+    if (desc.getPartitionSpec() != null && DDLUtils.isIcebergTable(table) &&
```
deniskuzZ, line 68: could we move this logic into AlterTableUtils?
````
isIdentityPartitionSpec(table) {
  return table.getStorageHandler() != null && table.getStorageHandler().getPartitionTransformSpec(table)
      .stream().map(TransformSpec::getTransformType)
      .allMatch(type -> type == TransformSpec.TransformType.IDENTITY);
````
}

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java:68]: we should also check if table `hasUndergonePartitionEvolution`, see method in HiveIcebergStorageHandler

File: ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java (hunk 33 above)
Comment (difin) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java:8825]: Extracted the code into a local var, but about the second part, unfortunately it won't work because I saw that for Iceberg it can be `inColumnCnt + staticPartitionColumnCnt == outColumnCnt` or` inColumnCnt == outColumnCnt` and the suggested expression won't cover the first case.

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java (hunk 34 above)
Comment (difin) [reply ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java:68]: Done

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java (hunk 30 above)
Comment (difin) [reply ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java:145]: Yes, this is used for scenarios like you said.
It is already supported for Iceberg with the change I did yesterday to PartitionUtils.getPartitions() method.
I verified it with a q-test and it throws the exception `TOO_MANY_COMPACTION_PARTITIONS` as expected.

File: ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java (hunk 33 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java:8827]: i had the same code at the beginning, but it looks a bit weird if you split it
````
!alwaysUnpartitioned && inColumnCnt + staticPartitionColumnCnt != outColumnCnt 
or
alwaysUnpartitioned && inColumnCnt + staticPartitionColumnCnt != outColumnCnt &&  inColumnCnt != outColumnCnt
````

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java:8827]: if it could be `inColumnCnt + staticPartitionColumnCnt == outColumnCnt where staticPartitionColumnCnt is > 0 and in other cases it can be  inColumnCnt == outColumnCnt` why do we even need to handle iceberg differently?

if(inColumnCnt + staticPartitionColumnCnt != outColumnCnt)

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java [hunk 35]
```diff
@@ -750,6 +750,10 @@ default boolean canUseTruncate(org.apache.hadoop.hive.ql.metadata.Table hmsTable
     return true;
   }
 
+  default boolean hasUndergonePartitionEvolution(org.apache.hadoop.hive.ql.metadata.Table table) {
```
deniskuzZ, line 753: don't expose it via API

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
difin: Changed IOW query in Iceberg compactor class and to not include partition keys in the select query and now this change in SemanticAnalyzer is not needed.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -62,6 +65,17 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
       throw new HiveException(ErrorMsg.NONACID_COMPACTION_NOT_SUPPORTED, table.getDbName(), table.getTableName());
     }
 
+    if (desc.getPartitionSpec() != null && !AlterTableUtils.isIdentityPartitionTable(table)) {
+      throw new HiveException("Compaction is not supported on the table " +
+          table.getDbName() + "." + table.getTableName() + " with non-identity partition spec");
+    }
+
+    if (desc.getPartitionSpec() != null && DDLUtils.isIcebergTable(table) &&
```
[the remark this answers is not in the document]
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -750,6 +750,11 @@ default boolean canUseTruncate(org.apache.hadoop.hive.ql.metadata.Table hmsTable
     return true;
   }
 
+  default boolean isEligibilityForCompaction(org.apache.hadoop.hive.ql.metadata.Table table,
```
deniskuzZ, line 753: typo: isEligibleForCompaction

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -133,7 +138,7 @@ private List<Partition> getPartitions(Table table, AlterTableCompactDesc desc, D
       }
     } else {
       Map<String, String> partitionSpec = desc.getPartitionSpec();
-      partitions = context.getDb().getPartitions(table, partitionSpec);
+      partitions.addAll(PartitionUtils.getPartitions(context.getDb(), table, partitionSpec, false));
```
deniskuzZ, line 141: you should amend `Hive.getPartitions`, similar to `Hive.getPartitionsByExpr`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -61,6 +62,10 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
     if (!AcidUtils.isTransactionalTable(table) && !AcidUtils.isNonNativeAcidTable(table)) {
       throw new HiveException(ErrorMsg.NONACID_COMPACTION_NOT_SUPPORTED, table.getDbName(), table.getTableName());
     }
+    
+    if (table.getStorageHandler() != null) {
+      table.getStorageHandler().isEligibilityForCompaction(table, desc.getPartitionSpec());
```
deniskuzZ, line 67: return if "false"

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
```diff
@@ -94,10 +98,20 @@ public static ExprNodeGenericFuncDesc makeUnaryPredicate(String fn, ExprNodeDesc
   public static Partition getPartition(Hive db, Table table, Map<String, String> partitionSpec, boolean throwException)
       throws SemanticException {
     Partition partition;
-    try {
-      partition = db.getPartition(table, partitionSpec, false);
-    } catch (Exception e) {
-      throw new SemanticException(toMessage(ErrorMsg.INVALID_PARTITION, partitionSpec), e);
+    if (table.getStorageHandler() != null && table.getStorageHandler().alwaysUnpartitioned()) {
```
deniskuzZ, line 101: that should be done in `Hive.getPartition()`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
```diff
@@ -108,10 +122,18 @@ public static Partition getPartition(Hive db, Table table, Map<String, String> p
   public static List<Partition> getPartitions(Hive db, Table table, Map<String, String> partitionSpec,
       boolean throwException) throws SemanticException {
     List<Partition> partitions;
-    try {
-      partitions = partitionSpec == null ? db.getPartitions(table) : db.getPartitions(table, partitionSpec);
-    } catch (Exception e) {
-      throw new SemanticException(toMessage(ErrorMsg.INVALID_PARTITION, partitionSpec), e);
+    if (table.getStorageHandler() != null && table.getStorageHandler().alwaysUnpartitioned()) {
```
deniskuzZ, line 125: same here, `Hive.getPartitions()`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +55,71 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+        }
 
-    return true;
+        String fieldName = partitionData.getSchema().getFields().get(i).name();
+
+        sb.append(fieldName)
+            .append("=")
+            .append(quoteOpt)
+            .append(partSpecMap.get(fieldName))
+             .append(quoteOpt);
```
deniskuzZ, line 117: nit: remove extra space
also, you could simplify the code with 
````
HiveUtils.unparseIdentifier(fieldName)
  .append("=")  
  .append(TypeInfoUtils.convertStringToLiteralForSQL(
    partSpecMap.get(fieldName), 
    ((PrimitiveTypeInfo) TypeInfoUtils.getTypeInfoFromTypeString(partitionData.getType(i).typeId())).getPrimitiveCategory() 
````
check `DDLPlanUtils.getPartitionActualName`

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java [hunk 43]
```diff
@@ -782,7 +782,7 @@ private void setCommonHmsTablePropertiesForIceberg(org.apache.hadoop.hive.metast
    * @param hmsTable Table for which we are calculating the properties
    * @return The properties we can provide for Iceberg functions, like {@link Catalogs}
    */
-  private static Properties getCatalogProperties(org.apache.hadoop.hive.metastore.api.Table hmsTable) {
+  public static Properties getCatalogProperties(org.apache.hadoop.hive.metastore.api.Table hmsTable) {
```
deniskuzZ, line 785: should it be public?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1915,6 +1915,28 @@ private boolean hasUndergonePartitionEvolution(Table table) {
...
+  @Override
+  public boolean isEligibilityForCompaction(org.apache.hadoop.hive.ql.metadata.Table table,
+      Map<String, String> partitionSpec) throws HiveException {
+    if (partitionSpec != null) {
+      Table icebergTable = IcebergTableUtil.getTable(conf, table.getTTable());
+      if (hasUndergonePartitionEvolution(icebergTable)) {
+        throw new HiveException("Compaction on partition level is not supported on the table " +
+            table.getDbName() + "." + table.getTableName() + " which has undergone partition evolution");
+      }
+      if (!isIdentityPartitionTable(table)) {
```
deniskuzZ, line 1932: only primitiveTypes

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
difin: done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
difin: Fixed. Was needed before, but not anymore.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
difin: Done. This is not used in this PR.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: `getTypeInfoFromTypeString()` didn't work with getting the type from partitionData, but I made it work a little differently just like in `TruncateTableAnalyzer.constructDeleteQuery()`.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1915,6 +1917,46 @@ private boolean hasUndergonePartitionEvolution(Table table) {
...
+        return org.apache.commons.lang3.tuple.Pair.of(false, ErrorMsg.COMPACTION_NON_IDENTITY_PARTITION_SPEC);
+      }
+    }
+    return org.apache.commons.lang3.tuple.Pair.of(true, null);
+  }
+
+  @Override
+  public List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table,
+      Map<String, String> partitionSpec, short limit) throws SemanticException {
+    return table.getStorageHandler().getPartitionNames(table, partitionSpec).stream()
```
deniskuzZ, line 1943: why do you need `table.getStorageHandler()`, call `getPartitionNames` directly

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1915,6 +1917,46 @@ private boolean hasUndergonePartitionEvolution(Table table) {
...
+        .limit(limit > 0 ? limit : Long.MAX_VALUE)
```
deniskuzZ, line 1944: `Math.min(limit, Long.MAX_VALUE)` ?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1915,6 +1917,46 @@ private boolean hasUndergonePartitionEvolution(Table table) {
...
+  }
+
+  @Override
+  public Partition getPartition(org.apache.hadoop.hive.ql.metadata.Table table,
+      Map<String, String> partitionSpec) throws SemanticException {
+    validatePartSpec(table, partitionSpec);
+    try {
+      String partName = Warehouse.makePartName(partitionSpec, false);
+      return new DummyPartition(table, partName, partitionSpec);
+    } catch (MetaException e) {
```
deniskuzZ, line 1955: you should throw `MetaException` and catch it in the orig place - Analyzer

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
```diff
@@ -95,7 +95,7 @@ public static Partition getPartition(Hive db, Table table, Map<String, String> p
       throws SemanticException {
     Partition partition;
     try {
-      partition = db.getPartition(table, partitionSpec, false);
+      partition = db.getPartition(table, partitionSpec);
```
deniskuzZ, line 98: 3rd param, `forceCreate` was omitted, why?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
deniskuzZ: how about 
````
PrimitiveObjectInspector primitiveObjInspector = IcebergObjectInspector.primitive(partitionData.getType(i))
TypeInfoUtils.convertStringToLiteralForSQL(value, primitiveObjInspector.getPrimitiveCategory())
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java (hunk 45 above)
Comment (difin) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java:1943]: Done

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 49]
```diff
@@ -1915,6 +1917,46 @@ private boolean hasUndergonePartitionEvolution(Table table) {
...
+      if (!isIdentityPartitionTable(table)) {
+        return org.apache.commons.lang3.tuple.Pair.of(false, ErrorMsg.COMPACTION_NON_IDENTITY_PARTITION_SPEC);
+      }
+    }
+    return org.apache.commons.lang3.tuple.Pair.of(true, null);
+  }
+
+  @Override
+  public List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table,
+      Map<String, String> partitionSpec, short limit) throws SemanticException {
```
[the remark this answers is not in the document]
difin: We don't need limit, fixed.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Thanks, done.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: The line above also throws SemanticException (as discussed)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +55,51 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
-        compactTableName);
+    HiveConf conf = new HiveConf(context.getConf());
+    String partSpec = context.getCompactionInfo().partName;
+    String compactionQuery;
+
+    if (partSpec == null) {
+      HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.ALL_PARTITIONS.name());
+      compactionQuery = String.format("insert overwrite table %s select * from %<s", compactTableName);
+    } else {
+      Hive db = Hive.get(conf);
```
deniskuzZ, line 66: can we inline where it's used

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +55,51 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
-    HiveConf.setVar(context.getConf(), ConfVars.REWRITE_POLICY, RewritePolicy.ALL_PARTITIONS.name());
     try {
-      DriverUtils.runOnDriver(context.getConf(), sessionState, compactionQuery);
+      DriverUtils.runOnDriver(conf, sessionState, compactionQuery);
       LOG.info("Completed compaction for table {}", compactTableName);
+      return true;
     } catch (HiveException e) {
-      LOG.error("Error doing query based {} compaction", RewritePolicy.ALL_PARTITIONS.name(), e);
-      throw new RuntimeException(e);
+      LOG.error("Failed compacting table {}", compactTableName, e);
```
deniskuzZ, line 99: "Failed to compact table {}, partition {} "

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -777,6 +783,16 @@ default List<Partition> getPartitionsByExpr(org.apache.hadoop.hive.ql.metadata.T
             "for a table.");
   }
 
+  default List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table, 
+      Map<String, String> partialPartSpec) throws SemanticException {
```
deniskuzZ, line 787: `partitionSpec`

difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -44,22 +56,52 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+      compactionQuery = String.format("insert overwrite table %1$s partition(%2$s) select %4$s from %1$s where %3$s",
+          compactTableName,
+          StringUtils.join(partValues, ","),
+          StringUtils.join(partValues, " and "),
+          queryFields);
+    }
+
+    SessionState sessionState = setupQueryCompactionSession(conf, context.getCompactionInfo(), tblProperties);
+    String compactionTarget = "table " + compactTableName +
+        (Strings.nullToEmpty(partSpec).isEmpty() ? "" : ", partition " + partSpec);
```
deniskuzZ, line 94: why no simply
````
"table " + HiveUtils.unparseIdentifier(compactTableName) +
    (partSpec != null ? ", partition " + HiveUtils.unparseIdentifier(partSpec) : "")
````

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java [hunk 54]
```diff
@@ -777,6 +783,16 @@ default List<Partition> getPartitionsByExpr(org.apache.hadoop.hive.ql.metadata.T
             "for a table.");
   }
 
+  default List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table, 
+      Map<String, String> partitionSpec) throws SemanticException {
+    throw new UnsupportedOperationException("Storage handler does not support getting partitions for a table.");
+  }
+
+  default Partition getPartition(org.apache.hadoop.hive.ql.metadata.Table table, Map<String, String> partSpec)
```
deniskuzZ, line 791: `partitionSpec`
please keep param names consistent, that is an interface

deniskuzZ (commented):
LGTM +1, minor code-style things

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
difin: Done

deniskuzZ (approved):

PR #5131, opened 2024-03-15
HIVE-28197: Add deserializer to convert JSON plans to RelNodes

soumyakanti3578:

Adds a deserializer that converts JSON plans to RelNodes

This is a new feature which can be used to get Calcite plans from JSON representation of the plan.

No

No

I have added the property `set hive.test.cbo.plan.serialization.deserialization.enabled=true;` to all the plans that had benign changes after deserialization. All other tests didn't have any changes.

Although I am hoping to get some feedback regarding what kind of tests to add. In one of the commits, I have tested by running all qtests through the deserializer - all of them pass and some of them have benign changes.


ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -1745,9 +1748,72 @@ public RelNode apply(RelOptCluster cluster, RelOptSchema relOptSchema, SchemaPlu
...
+        if (LOG.isDebugEnabled()) {
+          LOG.debug("Base plan: \n{}", RelOptUtil.toString(calcitePlan));
+          LOG.debug("Plan from JSON: \n{}", RelOptUtil.toString(fromJson));
+        }
+
+        calcitePlan = fromJson;
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+
```
kasakrisz, line 1792: I think test code should't be shipped to production. Could you please move this to a unit test, there are ways to build a test plan:
https://github.com/apache/hive/blob/master/ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/TestHiveRowIsDeletedPropagator.java

kasakrisz (changes requested):

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
soumyakanti3578: Only reason I have the method `testSerializationAndDeserialization` here instead of a unit test is to enable integration tests with just a property from the qfile. My idea was to enable `TEST_CBO_PLAN_SERIALIZATION_DESERIALIZATION_ENABLED` from either each individual qfile, or from `hive-site.xml` for a whole driver like `TestTezTPCDS30TBPerfCliDriver`.

That would ensure that we will test this for a larger number and more diverse set of queries, instead of testing a few scenarios with unit tests.

Let me know if you still want me to add a unit test for this and remove this from here and I will do it. :)

soumyakanti3578: Added a few unit tests, and removed this from here. Please review again.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/RelPlanParser.java
```diff
@@ -0,0 +1,477 @@
...
+  public RelNode read(String s) throws IOException {
+    JsonFactory jsonFactory = new JsonFactory();
+    Map<String, Object> o;
+    try (JsonParser parser = jsonFactory.createParser(s)) {
+      parser.nextToken();
+      assert parser.currentToken() == JsonToken.START_OBJECT;
+      parser.nextToken();
+      assert parser.currentToken() == JsonToken.FIELD_NAME;
+      parser.nextToken();
+      assert parser.currentToken() == JsonToken.VALUE_STRING;
```
kasakrisz, line 111: what happens in production code if one if these asserts fail? Should we throw exception instead?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/RelPlanParser.java
```diff
@@ -0,0 +1,477 @@
...
+
+      lastRel = null;
+      final ObjectMapper mapper = new ObjectMapper();
+      o = mapper.configure(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS, true)
+          .readValue(parser.getValueAsString(), TYPE_REF);
```
kasakrisz, line 116: What are the benefits of using Map<String, Object> as a json element representation (DTO) instead of annotated POJOs?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelJson.java
```diff
@@ -43,9 +127,170 @@ public Object toJson(Object value) {
...
+    if (SqlTypeFamily.TIMESTAMP == literal.getTypeName().getFamily()) {
+      // Had to do this to prevent millis or nanos from getting trimmed
+      value = literal.computeDigest(RexDigestIncludeType.NO_TYPE);
+    } else {
+      value = literal.getValue3();
+    }
+    map.put(LITERAL, RelEnumTypes.fromEnum(value));
+    map.put("type", toJson(literal.getType()));
+    return map;
+  }
```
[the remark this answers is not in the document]
soumyakanti3578: I have tried to use the Calcite code as much as possible, and only had to copy these when I needed to add something extra to the method or a part of the method and the super method is private (and thus not override-able). For example in this case, for TIMESTAMPs I needed to handle it differently than what was implemented in Calcite.

Also you can see that `toJsonRexCall` has extra things like [adding type to RexCalls](https://github.com/apache/hive/pull/5131/commits/ed787c28ba0708d04385fab5588296e1f1638284), support for MAPs and ARRAYs which is missing in the super method, etc. These are needed in the serialized plan to get all the information we need to deserialize it.

To see an example of where I was able to reuse super method, please checkout [this patch for AggregateCalls](https://github.com/apache/hive/pull/5131/commits/c82de5698e6a1b9c029a31034444d1991f0a0701#diff-29a8fea85c2750c60547a7b4d3088d3f704a48f77f6a6b9eb933f1b5b527e033R78).

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/RelPlanParser.java
soumyakanti3578: Adding a catch block for `AssertionError` and re-throwing it as an IOException.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/RelPlanParser.java
soumyakanti3578: I think we need the elements in `Map<String, Object>` to use Calcite's `RelJson` which expects elements to be in `Map<String, Object>`.

PR #5134, opened 2024-03-19
HIVE-28087: Iceberg: Timestamp partition columns with transforms are not correctly sorted during insert

SourabhBadhya:

HIVE-28087: Iceberg: Timestamp partition columns with transforms are not correctly sorted during insert

Timestamp partition columns with transforms are not correctly sorted during insert

No

No

QTest


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/udf/GenericUDFIcebergMonth.java
```diff
@@ -97,7 +97,7 @@ public ObjectInspector initialize(ObjectInspector[] arguments) throws UDFArgumen
         Function<Object, Integer> timestampLocalTzTransform = Transforms.month().bind(Types.TimestampType.withZone());
         evaluator = arg -> {
           TimestampLocalTZWritable val = (TimestampLocalTZWritable) converter.convert(arg.get());
-          result.set(timestampLocalTzTransform.apply(val.getNanos() / 1000L));
+          result.set(timestampLocalTzTransform.apply(val.getSeconds() * 1e6 + val.getNanos() / 1000L));
```
ayushtkn, line 100: this same logic is used multiple places, can explore extracting into a util to get microseconds

ayushtkn (approved):
If the tests come green. Changes LGTM.

If there is a possibility to reduce the size of the csv file used for testing, give it a shot

simhadri-g (approved):
I think we can reduce the number of rows in CSV. 

Rest LGTM +1.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/udf/GenericUDFIcebergMonth.java
SourabhBadhya: Added util methods in Timestamp and TimestampLocal code.

SourabhBadhya:
Reduced the number of rows from 1000 to 48 in the data file.

PR #5136, opened 2024-03-19
HIVE-28126: Use added record count in cost model when rebuilding materialized view stored by iceberg

kasakrisz:
1. Two plans are generated for rebuilding materialized view: incremental and full rebuild plan. The compiler chooses the cheaper plan based on cost. Instead of all records in a non native source table (Iceberg) the table scan operators should return only the number of records appended since the last MV rebuild when calculating the cost of the incremental rebuild plan. 
2. Deprecate `HiveStorageHandler.hasAppendsOnly()` and use the new `getSnapShots()` method.

Improve cost based decision heuristic.

No, but changes in `EXPLAIN` output.

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestIcebergCliDriver -Dqfile=mv_iceberg_orc.q,mv_iceberg_orc2.q,mv_iceberg_orc3.q,mv_iceberg_orc4.q,mv_iceberg_orc5.q,mv_iceberg_orc6.q,mv_iceberg_orc7.q,mv_iceberg_partitioned_orc.q,mv_iceberg_partitioned_orc2.q -pl itests/qtest-iceberg -Piceberg -Pitests
```

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -1647,7 +1647,7 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   }
 
   @Override
-  public Iterable<SnapshotContext> getSnapshots(
+  public Iterable<SnapshotContext> getSnapshotContexts(
       org.apache.hadoop.hive.ql.metadata.Table hmsTable, SnapshotContext since) {
 
     TableDesc tableDesc = Utilities.getTableDesc(hmsTable);
@@ -1747,7 +1747,7 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   /**
    * Check the operation type of all snapshots which are newer than the specified. The specified snapshot is excluded.
    * @deprecated
-   * <br>Use {@link HiveStorageHandler#getSnapshots(org.apache.hadoop.hive.ql.metadata.Table hmsTable, SnapshotContext since)}
+   * <br>Use {@link HiveStorageHandler#getSnapshotContexts(org.apache.hadoop.hive.ql.metadata.Table hmsTable, SnapshotContext since)}
    * and check {@link SnapshotContext.WriteOperationType#APPEND}.equals({@link SnapshotContext#getOperation()}).
    *
    * @param hmsTable table metadata stored in Hive Metastore
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -137,6 +137,7 @@ import org.apache.hadoop.mapred.Reporter;
 import org.apache.iceberg.BaseMetastoreTableOperations;
 import org.apache.iceberg.BaseTable;
 import org.apache.iceberg.DataFile;
+import org.apache.iceberg.DataOperations;
 import org.apache.iceberg.ExpireSnapshots;
 import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.FileScanTask;
@@ -1618,7 +1619,67 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
     if (current == null) {
       return null;
     }
-    return new SnapshotContext(current.snapshotId());
+    return toSnapshotContext(current);
+  }
+
+  private SnapshotContext toSnapshotContext(Snapshot snapshot) {
+    Map<String, String> summaryMap = snapshot.summary();
+    long addedRecords = getLongSummary(summaryMap, SnapshotSummary.ADDED_RECORDS_PROP);
+    long deletedRecords = getLongSummary(summaryMap, SnapshotSummary.DELETED_RECORDS_PROP);
+    return new SnapshotContext(
+        snapshot.snapshotId(), toWriteOperationType(snapshot.operation()), addedRecords, deletedRecords);
+  }
+
+  private SnapshotContext.WriteOperationType toWriteOperationType(String operation) {
+    if (DataOperations.APPEND.equals(operation)) {
+      return SnapshotContext.WriteOperationType.APPEND;
+    } else if (DataOperations.DELETE.equals(operation)) {
+      return SnapshotContext.WriteOperationType.DELETE;
+    } else if (DataOperations.OVERWRITE.equals(operation)) {
+      return SnapshotContext.WriteOperationType.OVERWRITE;
+    } else if (DataOperations.REPLACE.equals(operation)) {
+      return SnapshotContext.WriteOperationType.REPLACE;
+    } else {
+      return SnapshotContext.WriteOperationType.UNKNOWN;
+    }
+  }
+
+  private long getLongSummary(Map<String, String> summaryMap, String key) {
+    String textValue = summaryMap.get(key);
+    if (StringUtils.isBlank(textValue)) {
+      return 0;
+    }
+    return Long.parseLong(textValue);
+  }
+
+  public Iterable<SnapshotContext> getSnapshots(
+      org.apache.hadoop.hive.ql.metadata.Table hmsTable, SnapshotContext since) {
+
+    TableDesc tableDesc = Utilities.getTableDesc(hmsTable);
+    Table table = IcebergTableUtil.getTable(conf, tableDesc.getProperties());
+    return getSnapshots(table.snapshots(), since);
+  }
+
+  @VisibleForTesting
+  Iterable<SnapshotContext> getSnapshots(Iterable<Snapshot> snapshots, SnapshotContext since) {
+    List<SnapshotContext> result = Lists.newArrayList();
+
+    boolean foundSince = since == null;
+    for (Snapshot snapshot : snapshots) {
+      if (!foundSince) {
+        if (snapshot.snapshotId() == since.getSnapshotId()) {
+          foundSince = true;
+        }
+      } else {
+        result.add(toSnapshotContext(snapshot));
+      }
+    }
+
+    if (foundSince) {
+      return result;
+    }
+
+    return Collections.emptyList();
   }
 
   @Override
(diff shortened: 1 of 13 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1611,7 +1612,68 @@ public SnapshotContext getCurrentSnapshotContext(org.apache.hadoop.hive.ql.metad
...
+  private SnapshotContext toSnapshotContext(Snapshot snapshot) {
+    Map<String, String> summaryMap = snapshot.summary();
+    long addedRecords = getLongSummary(summaryMap, SnapshotSummary.ADDED_RECORDS_PROP);
+    long deletedRecords = getLongSummary(summaryMap, SnapshotSummary.DELETED_RECORDS_PROP);
+    return new SnapshotContext(
+        snapshot.snapshotId(), toWriteOperationType(snapshot.operation()), addedRecords, deletedRecords);
+  }
+
+  private SnapshotContext.WriteOperationType toWriteOperationType(String operation) {
+    if (DataOperations.APPEND.equals(operation)) {
```
deniskuzZ, line 1627: can't we do 
````
try {
  SnapshotContext.WriteOperationType.valueOf(DataOperations.APPEND.toUpperCase())
} catch(IllegalArgumentException ex) {...}
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 2]
```diff
@@ -1611,7 +1612,68 @@ public SnapshotContext getCurrentSnapshotContext(org.apache.hadoop.hive.ql.metad
     if (current == null) {
       return null;
     }
-    return new SnapshotContext(current.snapshotId());
+    return toSnapshotContext(current);
+  }
+
+  private SnapshotContext toSnapshotContext(Snapshot snapshot) {
```
deniskuzZ, line 1618: should we put helper private methods at the end of the class?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1611,7 +1612,68 @@ public SnapshotContext getCurrentSnapshotContext(org.apache.hadoop.hive.ql.metad
...
+  private long getLongSummary(Map<String, String> summaryMap, String key) {
+    String textValue = summaryMap.get(key);
+    if (StringUtils.isBlank(textValue)) {
+      return 0;
+    }
+    return Long.parseLong(textValue);
+  }
+
+  @Override
+  public Iterable<SnapshotContext> getSnapshots(
```
deniskuzZ, line 1649: maybe `getSnapshotContexts`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1611,7 +1612,68 @@ public SnapshotContext getCurrentSnapshotContext(org.apache.hadoop.hive.ql.metad
...
+    TableDesc tableDesc = Utilities.getTableDesc(hmsTable);
+    Table table = IcebergTableUtil.getTable(conf, tableDesc.getProperties());
+    return getSnapshots(table.snapshots(), since);
+  }
+
+  @VisibleForTesting
+  Iterable<SnapshotContext> getSnapshots(Iterable<Snapshot> snapshots, SnapshotContext since) {
+    List<SnapshotContext> result = Lists.newArrayList();
+
+    boolean foundSince = since == null;
```
deniskuzZ, line 1661: boolean foundSince = Objects.isNull(since)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1611,7 +1612,68 @@ public SnapshotContext getCurrentSnapshotContext(org.apache.hadoop.hive.ql.metad
...
+      if (!foundSince) {
+        if (snapshot.snapshotId() == since.getSnapshotId()) {
+          foundSince = true;
+        }
+      } else {
+        result.add(toSnapshotContext(snapshot));
+      }
+    }
+
+    if (foundSince) {
```
deniskuzZ, line 1672: maybe ternaty operator
````
return foundSince? result : Collections.emptyList()
````

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java [hunk 6]
```diff
@@ -2171,22 +2172,32 @@ private Materialization getMaterializationInvalidationInfo(MaterializedViewMetad
         materialization.setSourceTablesCompacted(true);
         return materialization;
       }
-      Boolean b = storageHandler.hasAppendsOnly(
-          table, mvSnapshot.getTableSnapshots().get(table.getFullyQualifiedName()));
-      if (b == null) {
+
+      boolean hasAppendsOnly = true;
+      boolean noSnapshots = true;
+      for (SnapshotContext snapshot : storageHandler.getSnapshots(
```
deniskuzZ, line 2178: should we check if snapshots are supported?
````
mvSnapshot.getTableSnapshots().get(table.getFullyQualifiedName())
````
won't create any issues without check?

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java [hunk 7]
```diff
@@ -2171,22 +2172,32 @@ private Materialization getMaterializationInvalidationInfo(MaterializedViewMetad
         materialization.setSourceTablesCompacted(true);
         return materialization;
       }
-      Boolean b = storageHandler.hasAppendsOnly(
-          table, mvSnapshot.getTableSnapshots().get(table.getFullyQualifiedName()));
-      if (b == null) {
+
+      boolean hasAppendsOnly = true;
```
deniskuzZ, line 2176: why do we need 1 more flag, can we leave with 1?

deniskuzZ (commented):
LGTM, some minor stuff

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1611,7 +1612,68 @@ public SnapshotContext getCurrentSnapshotContext(org.apache.hadoop.hive.ql.metad
...
+  }
+
+  @VisibleForTesting
+  Iterable<SnapshotContext> getSnapshots(Iterable<Snapshot> snapshots, SnapshotContext since) {
+    List<SnapshotContext> result = Lists.newArrayList();
+
+    boolean foundSince = since == null;
+    for (Snapshot snapshot : snapshots) {
+      if (!foundSince) {
+        if (snapshot.snapshotId() == since.getSnapshotId()) {
```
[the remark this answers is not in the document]
kasakrisz: Unfortunately `snapshotId` is not monotonically incremental. `sequence-number` would be better but it exist only from v2.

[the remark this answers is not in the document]
deniskuzZ: 👍

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
kasakrisz: changed

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
kasakrisz: I like them after the method where they called first to easily scroll back to the caller. 
IMHO if the class is too big like this one their location doesn't add any value.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
kasakrisz: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
kasakrisz: done

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
kasakrisz: Source table types are checked at MV creation. Also the default implementation of getSnapshots returns empty list which means we have to fall back to full rebuild. This is also the case when Iceberg table is empty so there are no snapshots. But I added a check It doesn't hurt.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
kasakrisz: Refactored to rely on only one flag

deniskuzZ: @kasakrisz, have you submitted the change?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: @kasakrisz, would be nice to have some consistency in API names

deniskuzZ:
@kasakrisz, please check if issues reported by sonar make sence

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
kasakrisz: renamed to getSnapshotContexts

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
kasakrisz: sorry, I just pushed the missing commit

deniskuzZ (approved):
LGTM +1

okumin:
@kasakrisz @deniskuzZ Maybe, this is not HIVE-27831 maybe but HIVE-28126. You may update the commit message on merging this PR.

kasakrisz:
@okumin 
Good catch. Thanks.

PR #5137, opened 2024-03-20
HIVE-28118: Hive Insert Into S3 with Viewfs overload scheme fails wit…

eshwitha:
…h MoveTask Error

During hive staging, isPathEncrypted of the source tmp path returns false, check for isPathEncrypted of destination tmp path leads to resolvePath which throws a FileNotFoundException leading to MoveTask error. We add a check for file existence before call to isPathEncrypted.

When ViewFs Overload scheme is enabled and HDFS encryption is enabled (hadoop.security.key.provider.path is not null), hive insert into fails in below cases without the fix: 
1. Encryption zone is not enabled for hive warehouse directory.
2. Insert Into is done for the partition present in S3 (mount link created using Viewfs overload).

No

Tests to verify the behaviour are added.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -5290,7 +5290,8 @@ private void constructOneLBLocationMap(FileStatus fSta,
     try {
       return srcHdfsEncryptionShim != null
           && destHdfsEncryptionShim != null
-          && (srcHdfsEncryptionShim.isPathEncrypted(srcf) || destHdfsEncryptionShim.isPathEncrypted(destf))
+          && (srcFs.exists(srcf) && srcHdfsEncryptionShim.isPathEncrypted(srcf)
+          || destFs.exists(destf) && destHdfsEncryptionShim.isPathEncrypted(destf))
           && !srcHdfsEncryptionShim.arePathsOnSameEncryptionZone(srcf, destf, destHdfsEncryptionShim);
     } catch (IOException e) {
       throw new HiveException(e);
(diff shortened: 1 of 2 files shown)


eshwitha:
@ayushtkn Can you please review the changes? Thanks.

eshwitha:
Screenshot 2024-04-02 at 2 33 28 PM
The test -> TestEncryptedHDFSCliDriver passes in my local branch.
Command used: mvn test -Dtest=TestEncryptedHDFSCliDriver -Dqfile=encryption_load_data_to_encrypted_tables.q -pl itests/qtest -Pitests

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -5266,7 +5266,8 @@ static private boolean needToCopy(final HiveConf conf, Path srcf, Path destf, Fi
     try {
       return srcHdfsEncryptionShim != null
           && destHdfsEncryptionShim != null
-          && (srcHdfsEncryptionShim.isPathEncrypted(srcf) || destHdfsEncryptionShim.isPathEncrypted(destf))
+          && ((srcFs.exists(srcf) && srcHdfsEncryptionShim.isPathEncrypted(srcf))
```
deniskuzZ, line 5269: why do you need extra brackets for `&&` operands?

deniskuzZ (commented):
LGTM, but I would wait for @ayushtkn

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
eshwitha: Thank you for the review @deniskuzZ, updated.

eshwitha:
Request to kindly review @ayushtkn @deniskuzZ

deniskuzZ (approved):
LGTM +1, pending green build

deniskuzZ:
@eshwitha, new tests fail with NullPointer https://ci.hive.apache.org/job/hive-precommit/job/PR-5137/10/

PR #5138, opened 2024-03-20
HIVE-28130: Iceberg: Add support for Rename Branch.

ayushtkn:

Add support for renaming an iceberg branch

Better Usability of the feature

Iceberg Branches can be renamed

No

UT


parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
```diff
@@ -503,6 +503,8 @@ alterStatementSuffixExecute
     -> ^(TOK_ALTERTABLE_EXECUTE KW_EXPIRE_SNAPSHOTS KW_RETAIN $numToRetain)
     | KW_EXECUTE KW_DELETE KW_ORPHAN_FILES (KW_OLDER KW_THAN LPAREN (timestamp=StringLiteral) RPAREN)?
     -> ^(TOK_ALTERTABLE_EXECUTE KW_ORPHAN_FILES $timestamp?)
+    | KW_EXECUTE KW_RENAME KW_BRANCH sourceBranch=StringLiteral KW_TO targetBranch=StringLiteral
```
deniskuzZ, line 506: why do we need `execute` here? simple rename should be enough
````
alter table ice01 rename branch 'source' to 'target';
````

Comment (zhangbutao) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g:506]: Agreed with @deniskuzZ .  
I think we need to reach an agreement on when to use this `execute` syntax.
IMHO, `execute` is good for some optimization & maintenance operation. I think TrinoDB also thought so:(https://trino.io/docs/current/connector/iceberg.html#alter-table-execute)

For basic operation, rename can use the same simple syntax like create/drop.

Comment (ayushtkn) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g]: thanx folks, I have changed the syntax & pushed

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/snapshotref/branch/rename/AlterTableRenameBranchAnalyzer.java [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.hadoop.hive.ql.parse.AlterTableSnapshotRefSpec;
+import org.apache.hadoop.hive.ql.parse.HiveParser;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+
+@DDLSemanticAnalyzerFactory.DDLType(types = HiveParser.TOK_ALTERTABLE_RENAME_BRANCH)
+public class AlterTableRenameBranchAnalyzer extends AbstractAlterTableAnalyzer {
+
+  protected AlterTableType alterTableType;
+
+  public AlterTableRenameBranchAnalyzer(QueryState queryState) throws SemanticException {
```
zhangbutao, line 44: Nit: Maybe use the name `AlterTableRenameSnapshotRefAnalyzer`? Just like `AlterTableCreateSnapshotRefAnalyzer`.

But i know iceberg tag can not be renamed, so this class can only be used for branch. I am not sure if we need to follow the previous naming rules.  cc @deniskuzZ

ql/src/java/org/apache/hadoop/hive/ql/parse/AlterTableSnapshotRefSpec.java
```diff
@@ -122,4 +122,29 @@ public String toString() {
       return MoreObjects.toStringHelper(this).add("refName", refName).add("ifExists", ifExists).toString();
     }
   }
+
+  public static class RenameBranchSpec {
```
zhangbutao, line 126: Also here, should we use `RenameSnapshotrefSpec`?

zhangbutao (commented):
Just left two nit comments, others looks good to me. Thanks.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/snapshotref/branch/rename/AlterTableRenameBranchAnalyzer.java
ayushtkn: I changed it, who knows those guys tmrw start supporting it for tags as well :-)

ql/src/java/org/apache/hadoop/hive/ql/parse/AlterTableSnapshotRefSpec.java
ayushtkn: Done

zhangbutao (approved):
LGTM + 1 Pending a green CI.

PR #5139, opened 2024-03-20
HIVE-28133: Log the original exception in HiveIOExceptionHandlerUtil#handleRecordReaderException

deniskuzZ:

Logging improvements

Troubleshooting

No

No

on a cluster

### DIFF
--- a/shims/common/src/main/java/org/apache/hadoop/hive/io/HiveIOExceptionHandlerUtil.java
+++ b/shims/common/src/main/java/org/apache/hadoop/hive/io/HiveIOExceptionHandlerUtil.java
@@ -21,8 +21,12 @@ import java.io.IOException;
 
 import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.mapred.RecordReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class HiveIOExceptionHandlerUtil {
+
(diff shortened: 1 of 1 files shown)


abstractdog:
LGTM, but created [HIVE-28135](https://issues.apache.org/jira/browse/HIVE-28135) as a followup

deniskuzZ:
> LGTM, but created [HIVE-28135](https://issues.apache.org/jira/browse/HIVE-28135) as a followup

it seems like a dead code with 0 implementations. should we then drop the whole stuff in 1 go?

abstractdog:
> > LGTM, but created [HIVE-28135](https://issues.apache.org/jira/browse/HIVE-28135) as a followup
> 
> it seems like a dead code with 0 implementations. should we then drop the whole stuff in 1 go? or maybe someone adds the handlers via aux jars?

even if they add it via aux jars, I still don't get the point of defining flexible exception handlers for record reader exceptions, not even a single google match that made sense, so I believe we can let this awesome handler chain go :)

deniskuzZ:
@abstractdog, should we merge this or wait for [HIVE-28135](https://issues.apache.org/jira/browse/HIVE-28135)?

abstractdog:
LGTM, we can merge until I finalize [HIVE-28135](https://issues.apache.org/jira/browse/HIVE-28135)

PR #5144, opened 2024-03-20
HIVE-28139: Upgrade to datasketches 1.2.0

zhangbutao:

https://datasketches.apache.org/docs/Community/Downloads.html shows that the latest version of datasketches-hive(Java Hive Adaptor) is 1.2.0, so we can upgrade it to from 1.1.0 to 1.2.0.

No

Yes
[mvn_den_log.txt](https://github.com/apache/hive/files/14668570/mvn_den_log.txt)

Existing UTs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -217,7 +217,7 @@
     <rs-api.version>2.0.1</rs-api.version>
     <json-path.version>2.9.0</json-path.version>
     <janino.version>3.0.11</janino.version>
-    <datasketches.version>1.1.0-incubating</datasketches.version>
+    <datasketches.version>1.2.0</datasketches.version>
     <spotbugs.version>4.0.3</spotbugs.version>
     <validation-api.version>1.1.0.Final</validation-api.version>
     <aws-secretsmanager-caching.version>1.0.1</aws-secretsmanager-caching.version>
--- a/standalone-metastore/pom.xml
+++ b/standalone-metastore/pom.xml
@@ -62,7 +62,7 @@
     <commons-lang3.version>3.12.0</commons-lang3.version>
     <commons-logging.version>1.1.3</commons-logging.version>
     <commons-dbcp2.version>2.9.0</commons-dbcp2.version>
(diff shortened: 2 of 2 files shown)


okumin (approved):

okumin:
We may ask the DataSketches community to cut a new tag someday. I recalled making changes to avoid WARNING when registering sketch UDFs.
https://github.com/apache/datasketches-hive/commits/master/

In another thought, we may port UDFs in datasketches-hive to hive someday to maintain them on our own.

zhangbutao:
> We may ask the DataSketches community to cut a new tag someday. I recalled making changes to avoid WARNING when registering sketch UDFs. https://github.com/apache/datasketches-hive/commits/master/

Good to know you fix this issue!

> In another thought, we may port UDFs in datasketches-hive to hive in the future to maintain them on our own.

I see datasketches-hive is still under development, though there is no commit for long time. I guess maybe it is stable.
BTW, we have inclued datasketches-hive jar in Hive's binary install package, so i think we can use these udf directly instead of porting them? But i am not sure if we can use these udf in Hive4 as datasketches-hive was built on Hive3 and Hive4 may have some compatibility issue with Hive3 on udf initialize.

But if we Hive community has strong usage about datasketches-hive udf, i think it is good to port them into Hive's repository. I would hear more folk's opinion about this and i think this part can be discussed more through Hive dev email.

okumin:
> But if we Hive community has strong usage about datasketches-hive udf, i think it is good to port them into Hive's repository. I would hear more folk's opinion about this and i think this part can be discussed more through Hive dev email.

Just a random thought but I agree as a future option.

> I just did take a look at the Hive4 code. I found https://issues.apache.org/jira/browse/HIVE-22940 has registered some datasketches-hive udf, could you please help to double check this? @okumin

That's true, and I remember I modified UDFs which are registered by FunctionRegistry so that I don't see WARNINGs. As far as I know, it is just a warning and no critical errors happen.

deniskuzZ (approved):

PR #5145, opened 2024-03-20
HIVE-28059: Iceberg REST Catalog

henrib:
This is a basic implementation of the Iceberg REST Catalog as a service embedded within HMS. 

For Hive users that have started using Iceberg to store data and want to expose those same tables to external query engines, implementing the Iceberg REST catalog embedded within HMS allows an easy upgrade and deployment path.
Although a dedicated service to host the REST catalog is certainly desirable, it also implies deploying and managing another service.

It does since HMS then 'hosts' a servlet that implements the Iceberg REST Catalog API.

No.

Unit tests.


deniskuzZ (changes requested):
-1, see dev thread

deniskuzZ:
Are you familiar with Hive HCatalog? 
````
HCatalog is a tool for accessing metadata that reside in Hive metastore. It acts as an API to expose the metastore as REST interface to external tools
````
https://www.linkedin.com/pulse/hive-metastore-hcatalog-hcat-haotian-zhang

If Hive already provides REST interface for HMS, we could just extend it with the Iceberg API, the same way DataBricks in UnityCatalog
````
sb.annotatedService(basePath + "iceberg",
      new IcebergRestCatalogService(catalogService, schemaService, tableService),
      icebergRequestConverter, icebergResponseConverter);
````
https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/UnityCatalogServer.java

Under hive-webhcat there is already HMS API REST server (probably API list needs to be updated), however, it could be extended with Iceberg API support

**Comment by henrib** [comment]
HCatalog seems almost abandoned. HIVE-28059 is just trying to expose the Iceberg API - nothing more - on an existing HMS instance. This patch allows Hive 4 users to consume their Iceberg data with other engines and at least 14 people seem to agree this would be/have been useful.

**Comment by deniskuzZ** [comment]
> HCatalog seems almost abandoned. [HIVE-28059](https://issues.apache.org/jira/browse/HIVE-28059) is just trying to expose the Iceberg API - nothing more - on an existing HMS instance. This patch allows Hive 4 users to consume their Iceberg data with other engines and at least 14 people seem to agree this would be/have been useful.

it is supposed to be a decoupled REST service solution, not a hacky enabler. You could check existing solutions to see how it's done: Gravitino REST Catalog, Polaris, Unity, etc

**Comment by henrib** [comment]
> it is supposed to be a decoupled REST service solution, not a hacky enabler. You could check existing projects to see how it's done: Gravitino REST Catalog, Polaris, Unity, etc

Iceberg tables still necessitate a Catalog - what HMS provides; it is sad that you can not see value in this practical approach that does not require depending on yet another service that ultimately will drive people away from Hive. Had you not embraced such a narrow view but played the community enabler role, Hive 4 would have had the 1st implementation of an Iceberg catalog 6 months ago... Since you are judge and jury, you should force close the JIRA and the PR.

**Comment by deniskuzZ** [comment]
> that does not require depending

1. I don't see any other community members supporting your "solution"; Did you get feedback or review comments for anyone else?
2. 6 months ago what we saw was a "broken hello world" nothing more. No doubt it improved after numerous review comments, but not all of them were resolved;
3. implemented here "rest catalog" solution is tightly coupled with HMS significantly limiting its usage patterns; 

PS: Failure to accept criticism does you no credit;

**Comment by henrib** [comment]
> 1. I don't see any other community members supporting your "solution"; Did you get feedback or review comments from anyone else?

10 thumbs up, 4 rockets on the PR - but I guess that is not "the process".

> 3. implemented here "rest catalog" solution is tightly coupled with HMS significantly limiting its usage patterns;

If you already have HMS... but this does not fit your dogma.

> PS: Failure to accept criticism does you no credit;

PS: Failure to accept other views does you no credit either;

**Comment by deniskuzZ** [comment]
> 10 thumbs up, 4 rockets on the PR - but I guess that is not "the process".
It seems like you can't comprehend the difference between an idea/feature and the way it is implemented. 

> If you already have HMS... but this does not fit your dogma.
WHY should it be coupled with HMS? 
Haven't we discussed this internally and reached a consensus to have a "decoupled REST service"? Why do you keep pushing your individual vision that is not supported by the majority, a vision that "only fits your dogma"?

1. Have you thought about the upgrade process? In order to release a new version of RestCatalog that would require HMS upgrade. 
2. What if someone decides to use a different Catalog implementation in Hive? 
3. What if we decide to use RestCatalog as central place for all metadata, serve it for multiple catalogs?
4. What if we need to scale them differently?

**Comment by henrib** [comment]
Dissing, ad-hominem attacks, soon you'll go (back) to insults; always delightful.
I will not talk about 'internal' products since this is not the place; I will just notice that your view of the majority is just claiming (first) you have it...
Since you obviously need to be right and can not accept the benefit of simplicity, I'll leave you be in your kingdom.

**Comment by deniskuzZ** [comment]
@henrib, i don't get you at all. I am trying to give you some suggestions, but you just make it personal.
Ok, I'll give it a last shot and leave it up to you:

If you think HMS users would benefit from the Rest API exposed directly from HMS Server, please move your RestCatalog implementation under the `standalone-metastore` (it shouldn't be part of the compute engine) and get reviews from the HMS folks: @nrg4878, @dengzhhu653, @saihemanth-cloudera

**Comment by vladislav-kuryata** [comment]
any updates here? This is a very useful implementation that will simplify access to metadata. and will simplify some external integration.

**Comment by henrib** [comment]
Work on it just resumed; need to move the whole under standalone-metastore and adapt to latest trunk (4.1) before presenting to new reviewers. Please add vote on the JIRA, this might help at review time.

**Review by saihemanth-cloudera — COMMENTED** [review]
First half of my review

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java [hunk 1]
```diff
@@ -732,10 +756,26 @@ public static void startMetaStore(int port, HadoopThriftAuthBridge bridge,
     }
     // optionally create and start the property server and servlet
     propertyServer = PropertyServlet.startServer(conf);
+    // optionally create and start the Iceberg REST server and servlet
+    icebergServer = startIcebergCatalog(conf);
```
dengzhhu653, line 760: Can we able to merge the `propertyServer` and the `icebergServer`, listening to the same port?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -1313,6 +1315,9 @@ private void create_database_core(RawStore ms, final Database db)
                   return null;
                 }
               });
+            } catch (SnapshotException e) {
```
dengzhhu653, line 1318: nit: looks this is unrelated to the change

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveActor.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.hadoop.hive.metastore.api.ShowLocksResponse;
+import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.iceberg.TableMetadata;
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.thrift.TException;
+
+/**
+ * Acts as the Hive client for the HiveCatalog benefit.
+ */
+public interface HiveActor {
```
dengzhhu653, line 37: To me, the `HiveActor` is not necessary. The embedded Hive can be implemented by setting `hive.metastore.uris` to an empty string during creating the `HiveCatalog`.
The new interface could introduce some burden if we want to port the Iceberg changes to Hive codebase in the future.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
```diff
@@ -0,0 +1,274 @@
...
+import org.apache.iceberg.catalog.Namespace;
+import org.apache.iceberg.hive.HiveActor;
+import org.apache.iceberg.hive.HiveLock;
+import org.apache.iceberg.hive.MetastoreLock;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.thrift.TException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HMSCatalogActor implements HiveActor {
```
dengzhhu653, line 60: This embedded `actor` doesn't support the multi-tenancy?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/Optimizer.java
```diff
@@ -192,7 +192,9 @@ public void initialize(HiveConf hiveConf) {
       transformations.add(new FixedBucketPruningOptimizer(compatMode));
     }
 
-    transformations.add(new BucketVersionPopulator());
+    if (!isTezExecEngine) {
```
dengzhhu653, line 195: nit: unrelated change?

standalone-metastore/metastore-catalog/data/conf/README.txt
```diff
@@ -0,0 +1 @@
+Need to force creation of a directory.
```
dengzhhu653, line 1: what's the new directory for?

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
```diff
@@ -3354,6 +3354,7 @@ const string PARTITION_TRANSFORM_SPEC = "partition_transform_spec",
 const string NO_CLEANUP = "no_cleanup",
 const string CTAS_LEGACY_CONFIG = "create_table_as_external",
 const string DEFAULT_TABLE_TYPE = "defaultTableType",
+const string DEFAULT_TABLE_LOCATION = "defaultLocation";
```
dengzhhu653, line 3357: what's this for? I cannot see where we use this variable.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/HiveCachingCatalog.java
```diff
@@ -0,0 +1,331 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(HiveCachingCatalog.class);
+  @SuppressWarnings("checkstyle:VisibilityModifier")
+  protected final long expirationIntervalMillis;
+  @SuppressWarnings("checkstyle:VisibilityModifier")
+  protected final Cache<TableIdentifier, Table> tableCache;
+  private final CATALOG catalog;
+  private final boolean caseSensitive;
+
+  @SuppressWarnings("checkstyle:VisibilityModifier")
+  protected HiveCachingCatalog(CATALOG catalog, long expirationIntervalMillis) {
```
dengzhhu653, line 66: seems like we only caching the metadata table, how do we benefit from this cache if we load a normal table?

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogServer.java
```diff
@@ -0,0 +1,139 @@
...
+  static Catalog getLastCatalog() {
+    return catalogRef != null ? catalogRef.get() :  null;
+  }
+
+  private HMSCatalogServer() {
+    // nothing
+  }
+
+  public static HttpServlet createServlet(SecureServletCaller security, Catalog catalog) throws IOException {
+    try (HMSCatalogAdapter adapter = new HMSCatalogAdapter(catalog)) {
```
dengzhhu653, line 62: nit: why do we need the try-with-resource, it sounds like the `adapter` is invalid after we create the `HMSCatalogServlet`.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogServer.java
```diff
@@ -0,0 +1,139 @@
...
+    String auth = MetastoreConf.getVar(configuration, MetastoreConf.ConfVars.CATALOG_SERVLET_AUTH);
+    boolean jwt = "jwt".equalsIgnoreCase(auth);
+    SecureServletCaller security = new ServletSecurity(configuration, jwt);
+    Catalog actualCatalog = catalog;
+    if (actualCatalog == null) {
+      actualCatalog = createCatalog(configuration);
+      actualCatalog.initialize("hive", Collections.emptyMap());
+    }
+    catalogRef = new SoftReference<>(actualCatalog);
+    return createServlet(security, actualCatalog);
```
dengzhhu653, line 98: should we put the `catalogRef` to `createServlet` instead of `actualCatalog`, so no strong reference to the `actualCatalog`?

common/pom.xml
```diff
@@ -39,10 +39,6 @@
       <artifactId>hive-shims</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <dependency>
```
dengzhhu653, line 42: nit: unrelate change?

dengzhhu653:
Just think aloud, how can we expand the capability of the Iceberg rest catalog? like supporting the Hive native tables, and Iceberg tables from different sources/places not only from the HMS.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/properties/PropertySchema.java
```diff
@@ -186,7 +186,7 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
    * @param version the schema version number
    * @param map        the map of properties and their types
    */
-  PropertySchema(String schemaName, int version, Map<String, PropertyType<?>> map) {
+  protected PropertySchema(String schemaName, int version, Map<String, PropertyType<?>> map) {
```
dengzhhu653, line 189: nit: unrelated change

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
henrib: For security reasons, it might be preferable to have 2 different ports so that routing can filter clients.
Besides, the property server can be configured not to start. (does not start by default I think).

dengzhhu653: Can this be achieved via `hive.metastore.catalog.servlet.path` and `hive.metastore.properties.servlet.path`?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2235,6 +2235,10 @@ public static enum ConfVars {
     HIVE_ICEBERG_ALLOW_DATAFILES_IN_TABLE_LOCATION_ONLY("hive.iceberg.allow.datafiles.in.table.location.only", false,
         "If this is set to true, then all the data files being read should be withing the table location"),
 
+    HIVE_ICEBERG_CATALOG_ACTOR_CLASS("hive.iceberg.catalog.actor.class", "org.apache.iceberg.rest.HMSCatalogActor",
+        "Catalog actor implementation class. Default value is an embedded (no-thrift) catalog actor." +
+            "For a thrift-based single-tenant actor, use \"org.apache.iceberg.hive.HiveCatalogActor\".."),
+
```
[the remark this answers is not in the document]
henrib: EmbeddedCatalogActor vs ThriftCatalogActor ?

hms-catalog/src/test/resources/scripts/metastore/upgrade/derby/hive-schema-4.1.0.derby.sql
```diff
@@ -0,0 +1,870 @@
...
+-- Source database is: /home/carl/Work/repos/hive1/metastore/scripts/upgrade/derby/mdb
+-- Connection URL is: jdbc:derby:/home/carl/Work/repos/hive1/metastore/scripts/upgrade/derby/mdb
+-- Specified schema is: APP
+-- appendLogs: false
+
+-- ----------------------------------------------
+-- DDL Statements for functions
+-- ----------------------------------------------
+
+CREATE FUNCTION "APP"."NUCLEUS_ASCII" (C CHAR(1)) RETURNS INTEGER LANGUAGE JAVA PARAMETER STYLE JAVA READS SQL DATA CALLED ON NULL INPUT EXTERNAL NAME 'org.datanucleus.store.rdbms.adapter.DerbySQLFunction.ascii' ;
```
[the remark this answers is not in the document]
henrib: Should not be there indeed.

iceberg/iceberg-catalog/pom.xml
```diff
@@ -32,6 +32,21 @@
...
+      <version>${iceberg.version}</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.iceberg</groupId>
+      <artifactId>iceberg-core</artifactId>
+      <version>${iceberg.version}</version>
+      <optional>true</optional>
+    </dependency>
+    -->
```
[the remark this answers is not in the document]
henrib: I must ask 'why'? It is very painful to debug when IntelliJ is 'unhappy' with shaded classes. I think this helps developers.

common/pom.xml
```diff
@@ -330,6 +330,10 @@
       <groupId>javolution</groupId>
       <artifactId>javolution</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-storage-api</artifactId>
```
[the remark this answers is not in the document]
henrib: By mistake...

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
henrib: It would make sense to host only one Jetty and create connectors/servlets per feature. Albeit this requires even more changes (and some logic to start the server if any of the feature servlets is required) that are not required for the task at hand. This seems like a 'lateral' move wrt handing Iceberg Catalog.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveActor.java
henrib: I beg to differ. My raw understanding of the code, starting with the HiveCatalog, requires a pool of IMetaStoreClient. The HiveMetaStore (the server part) itself can be instantiated as being local (isMetaStoreRemote=false) but this does not seem to make the client (IMetaStoreClient) any different; the method isMetaStoreRemote() is called only to help during autumn/authz from what I gather. This means that at least any call requires serializing/deserializing Thrift messages and go through the loopback interface (127.0.0.1) for no reason which is a cost we don't have to incur being embedded. Furthermore, this new interface does provide some isolation from HMS and instead of the whole IMetaStoreClient interface, we got a narrow set of identified behaviors required for the Iceberg Catalog only (not the whole legacy that HMS carries).

ql/src/java/org/apache/hadoop/hive/ql/optimizer/Optimizer.java
henrib: Probably.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -527,18 +507,26 @@ protected String defaultWarehouseLocation(TableIdentifier tableIdentifier) {
   }
 
   private String databaseLocation(String databaseName) {
-    String warehouseLocation = conf.get(HiveConf.ConfVars.METASTORE_WAREHOUSE.varname);
-    Preconditions.checkNotNull(
-            warehouseLocation, "Warehouse location is not set: hive.metastore.warehouse.dir=null");
+    String warehouseLocation = conf.get("metastore.warehouse.dir");
```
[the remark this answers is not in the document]
henrib: We indeed should check MetastoreConf.ConfVars.WAREHOUSE.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -240,19 +240,14 @@ private void renameTableOrView(
     String fromName = from.name();
 
     try {
-      Table table = clients.run(client -> client.getTable(fromDatabase, fromName));
-      validateTableIsIcebergTableOrView(contentType, table, CatalogUtil.fullTableName(name, from));
+      Table table = actor.getTable(fromDatabase, fromName);
+      validateTableIsIcebergTableOrView(contentType, table, TableIdentifier.of(from.namespace(), fromName).name());
```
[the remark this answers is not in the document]
henrib: We don't use it anywhere (else) but why not...

pom.xml
```diff
@@ -175,6 +175,10 @@
     <!-- Leaving libfb303 at 0.9.3 regardless of libthrift: As per THRIFT-4613 The Apache Thrift project does not publish items related to fb303 at this point -->
     <libfb303.version>0.9.3</libfb303.version>
     <libthrift.version>0.16.0</libthrift.version>
+    <!--
+    <thrift.home>/usr/local</thrift.home>
+    -->
```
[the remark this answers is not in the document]
henrib: Ok

serde/pom.xml
```diff
@@ -253,6 +258,12 @@
       <version>${mockito-core.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-classification</artifactId>
+      <version>4.1.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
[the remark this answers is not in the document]
henrib: Removed (dependency still exists, line 57).

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogServer.java
henrib: No, the servlet needs to hold a strong ref; the server holds it as a static soft ref for testing convenience.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogServer.java
henrib: Corrected.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveActor.java
henrib: Finally got what you said with some external help ;-) : if we are to set the THRIFT_URIS to blank in the config, the client created (HiveMetaStoreClient actor) checks MetastoreConf.isEmbeddedMetaStore(...) and calls callEmbeddedMetastore(...) which is not based on a Thrift client but a genuine IHMSHandler. By deriving HiveCatalog and overriding the method createPool() with a config modified to blank THRIFT_URIS, we now have a way to create embedded HiveCatalog. I wish I learnt that earlier...

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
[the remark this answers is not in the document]
zratkai: I could not easily distinguish between  EmbeddedCatalogActor and ThriftCatalogActor.
More meaningful names are the best choise. Since this is an Iceberg extension IcebergRestCatalogActor and HiveTriftCatalogActor sound good to me.

iceberg/iceberg-catalog/pom.xml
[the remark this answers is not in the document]
zratkai: It is never a recommended to store "garbage" and uncommented code in a repository. This is an antipattern. You can save it to yourself in your local repo and use it when needed.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -143,21 +149,21 @@ public FileIO io() {
   protected void doRefresh() {
     String metadataLocation = null;
     try {
-      Table table = metaClients.run(client -> client.getTable(database, tableName));
-      HiveOperationsBase.validateTableIsIceberg(table, fullName);
-
-      metadataLocation = table.getParameters().get(METADATA_LOCATION_PROP);
-
+      Table table = actor.getTable(database, tableName);
+      if (table != null) {
```
zratkai, line 153: How was this case handled before. when table is null?

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
```diff
@@ -0,0 +1,274 @@
...
+
+
+  @Override
+  public HiveActor initialize(Map<String, String> properties) {
+    return this;
+  }
+
+  @Override
+  public void alterDatabase(Namespace namespace, Database database) throws TException {
+    runVoid(h -> h.alter_database(namespace.level(0), database));
```
zratkai, line 152: Please use more meaningful names instead of just one character!

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
```diff
@@ -0,0 +1,274 @@
...
+
+  @FunctionalInterface
+  interface VoidAction {
+    void execute(IHMSHandler handler) throws TException;
+  }
+
+  private void runVoid(VoidAction action) throws TException {
+    IHMSHandler handler = null;
+    try {
+      try {
```
zratkai, line 127: This double try seems strange. Could you please use just one level of try-catch-finally here? It seems unnecessary to complicate.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  @Override
+  public void alterDatabase(Namespace namespace, Database database) throws TException {
+    runVoid(h -> h.alter_database(namespace.level(0), database));
+  }
+
+  @Override
+  public void alterTable(String databaseName, String tableName, Table table) throws TException {
+    runVoid(h -> h.alter_table(databaseName, tableName, table));
```
zratkai, line 157: Please use more meaningful names instead of just one character!

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
```diff
@@ -0,0 +1,274 @@
...
+
+  /** HiveTableOperations.NO_LOCK_EXPECTED_KEY */
+  static final String NO_LOCK_EXPECTED_KEY = "expected_parameter_key";
+  /** HiveTableOperations.NO_LOCK_EXPECTED_VALUE */
+  static final String NO_LOCK_EXPECTED_VALUE = "expected_parameter_value";
+
+  @Override
+  public void alterTable(String databaseName, String tableName, Table hmsTable, String metadataLocation)
+      throws TException {
+    runVoid(h -> h.alter_table_with_environment_context(
```
zratkai, line 168: Please use more meaningful names instead of just one character!

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
```diff
@@ -0,0 +1,274 @@
...
+
+  @FunctionalInterface
+  interface Action<R> {
+    R execute(IHMSHandler handler) throws TException;
+  }
+
+  private <R> R run(Action<R> action) throws TException {
+    IHMSHandler handler = null;
+    try {
+      try {
```
zratkai, line 102: This double try seems strange. Could you please use just one level of try-catch-finally here? It seems unnecessary to complicate.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
```diff
@@ -0,0 +1,622 @@
...
+    // update HMS catalog route counter metric
+    final String metricName = hmsCatalogMetricCount(route.name());
+    Counter counter = Metrics.getOrCreateCounter(metricName);
+    if (counter != null) {
+      counter.inc();
+    }
+    switch (route) {
+      case TOKENS: {
+        @SuppressWarnings("unchecked")
+        Map<String, String> request = (Map<String, String>) castRequest(Map.class, body);
```
zratkai, line 293: This switch contains 17 cases and even switch inside the case, which is really hard to follow what is happening here. Instead of adding the implementation inside the cases please transform it like:

switch (route) {
 case TOKENS:
  handleTokensRequest(...);
  case CONFIG:
   handleConfigRequest(...); 
  case LIST_NAMESPACES
   handleListNamespacesRequest(...);
  ...
}

This way it is much easier to understand, modify, maintain.

standalone-metastore/metastore-catalog/src/test/java/org/apache/iceberg/rest/HMSTestBase.java
```diff
@@ -0,0 +1,384 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.rest;
```
zratkai, line 20: This is not an Iceberg repository, so this is missleading. It must be something like: org.apache.hive.rest or org.apache.hive.iceberg.rest.

zratkai, line 20: This package name implies it is the iceberg repo itself. It should contain hive, like org.apache.hive.iceberg.rest.

standalone-metastore/metastore-catalog/pom.xml
```diff
@@ -0,0 +1,340 @@
...
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>hive-standalone-metastore</artifactId>
+    <groupId>org.apache.hive</groupId>
+    <version>4.1.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>hive-metastore-icecat</artifactId>
```
zratkai, line 20: I suggest to use more meaningful name like: hive-metastore-iceberg-catalog

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogServlet.java
```diff
@@ -0,0 +1,288 @@
...
+import org.apache.iceberg.relocated.com.google.common.io.CharStreams;
+import org.apache.iceberg.rest.HMSCatalogAdapter.HTTPMethod;
+import org.apache.iceberg.rest.HMSCatalogAdapter.Route;
+import org.apache.iceberg.rest.responses.ErrorResponse;
+import org.apache.iceberg.util.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Original @ https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/RESTCatalogServlet.java
```
zratkai, line 49: Is it a copy of a test?

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
```diff
@@ -0,0 +1,622 @@
...
+import org.apache.iceberg.rest.responses.ListNamespacesResponse;
+import org.apache.iceberg.rest.responses.ListTablesResponse;
+import org.apache.iceberg.rest.responses.LoadTableResponse;
+import org.apache.iceberg.rest.responses.OAuthTokenResponse;
+import org.apache.iceberg.rest.responses.UpdateNamespacePropertiesResponse;
+import org.apache.iceberg.util.Pair;
+import org.apache.iceberg.util.PropertyUtil;
+
+/**
+ * Original @ https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java
```
zratkai, line 77: Is it a copy of a test?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1825,6 +1825,24 @@ public enum ConfVars {
         "hive.metastore.properties.servlet.auth", "jwt",
         "Property-maps servlet authentication method (simple or jwt)."
     ),
+    PROPERTIES_NS("hive.metastore.properties.namespaces",
+        "hive.metastore.properties.namespaces", "",
+        "Property-maps namespace map; a JSON map where keys are namespace names, values namespace" +
+            "class and/or static factory method"),
+    CATALOG_SERVLET_PATH("hive.metastore.catalog.servlet.path",
```
zratkai, line 1832: Please use more appropriate name here. I could not find out neither from the name, nor from the description this property is for Iceberg Rest Server

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1825,6 +1825,24 @@ public enum ConfVars {
...
+        "hive.metastore.catalog.servlet.path", "icecli",
+        "Catalog servlet path component of URL endpoint."
+    ),
+    CATALOG_SERVLET_PORT("hive.metastore.catalog.servlet.port",
+        "hive.metastore.catalog.servlet.port", -1,
+        "Catalog servlet server port. Negative value disables the servlet," +
+            " 0 will let the system determine the catalog server port," +
+            " positive value will be used as-is."
+    ),
+    CATALOG_SERVLET_AUTH("hive.metastore.catalog.servlet.auth",
```
zratkai, line 1842: Please use more appropriate name here. I could not find out neither from the name, nor from the description this property is for Iceberg Rest Server

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogServer.java
```diff
@@ -0,0 +1,171 @@
...
+import org.eclipse.jetty.server.handler.gzip.GzipHandler;
+import org.eclipse.jetty.servlet.ServletContextHandler;
+import org.eclipse.jetty.servlet.ServletHolder;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.eclipse.jetty.util.thread.QueuedThreadPool;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class HMSCatalogServer {
```
zratkai, line 51: I think  HMSRESTCatalogServer would be better.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogServlet.java
```diff
@@ -0,0 +1,288 @@
...
+ * The RESTCatalogServlet provides a servlet implementation used in combination with a
+ * RESTCatalogAdaptor to proxy the REST Spec to any Catalog implementation.
+ */
+public class HMSCatalogServlet extends HttpServlet {
```
zratkai, line 53: I think HMSRESTCatalogServlet would be better.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
zratkai, line 60: I think HMSRESTCatalogActor would be better.

zratkai (changes requested):
Please check my comments!

henrib:
> Please check my comments!

Will do :-)

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
henrib: This code originates from the Iceberg test directory, yes.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogServlet.java
henrib: This code originates from the Iceberg test directory, yes.

standalone-metastore/metastore-catalog/pom.xml
henrib: This 'version' is hosted by the HMS2 thus having a "metastore" somewhere in the name; there will probably be different versions (standalone process).

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
henrib: Will do.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
henrib: Will do.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
henrib: This whole class will disappear.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
henrib: This whole class will disappear.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
henrib: This whole class will disappear.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
henrib: This whole class will disappear.

standalone-metastore/metastore-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogActor.java
henrib: This whole class will disappear.

PR #5151, opened 2024-03-21
HIVE-25054: (2.3) Drop vulnerable jodd-core dependency

pan3793:

This is an alternative to https://github.com/apache/hive/pull/4923 to address jodd CVE.

Hive uses jodd's JDateTime to handle Julian's calendar in processing Parquet timestamp, while the calendar is tricky and bugly stuff, considering that branch-2.3 has been in maintaining mode for a long time, I prefer to copy the corresponding code to **avoid involving any unexpected behavior change**.

To address CVE-2018-21234.

Jodd CVE is listed in https://issues.apache.org/jira/browse/SPARK-44757 top 1, with a score 9.8.

No.

No.

UT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,7 +165,6 @@
     <jline.version>2.12</jline.version>
     <jms.version>2.0.2</jms.version>
     <joda.version>2.8.1</joda.version>
-    <jodd.version>3.5.2</jodd.version>
     <json.version>1.8</json.version>
     <junit.version>4.11</junit.version>
     <kryo.version>3.0.3</kryo.version>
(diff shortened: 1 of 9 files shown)


pan3793:
cc @sunchao, this is an alternative to https://github.com/apache/hive/pull/4923, and I prefer this way for safety, please let me know what you think.

pom.xml
```diff
@@ -1119,6 +1118,7 @@
 	    <exclude>**/*.html</exclude>
 	    <exclude>**/sit</exclude>
       <exclude>**/test/queries/**/*.sql</exclude>
+        <exclude>**/ql/io/parquet/timestamp/datetime/**</exclude>
```
pan3793, line 1121: according to https://www.apache.org/legal/src-headers.html#3party, we should not add AL2 license header to the copied source files

sunchao (approved):
LGTM

pan3793:
Thanks for your approval, @sunchao

sunchao:
Thanks @pan3793 !

PR #5152, opened 2024-03-21
HIVE-28142: Fix flakyness in acid_bloom_filter_orc_file_dump

InvisibleProgrammer:
The test uses the PostExecOrcFileDump post exec hook on a limit 1 query. 
It has no order by and the test table contains two rows. Sometimes it picks up the other row than it is in the q.out file. 

Define explicit order

Test stability

No

No

`mvn test -Pitests -pl itests/qtest -Dtest=TestMiniLlapLocalCliDriver -Dqfile=acid_bloom_filter_orc_file_dump.q`


kuczoram (approved):
Looks good! Thanks for the fix!

PR #5153, opened 2024-03-21
HIVE-12930: Support SSL Shuffle for LLAP

ganeshashree:

Added support for enabling SSL for LLAP shuffle. Made changes in LLAP Shuffle handler similar to Tez shuffle handler. Added ssl-server.xml configuration file to LLAP daemon configs since LLAP daemon acts as both server and client for shuffle.

These changes are needed to enable SSL for LLAP shuffle. 

Yes, this change introduces _hive.llap.shuffle.ssl.enabled_ configuration property for enabling SSL for LLAP shuffle. Both _tez.runtime.shuffle.ssl.enable_ and _hive.llap.shuffle.ssl.enabled_ have to be enabled for enabling SSL for LLAP shuffle. This is because the shuffle fetcher implementation in Tez is dependent on _tez.runtime.shuffle.ssl.enable_  for enabling SSL while communicating with shuffle handler for fetching shuffle data. 

No

Manually verified that SSL is enabled for the LLAP shuffle.

### DIFF
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/shufflehandler/ShuffleHandler.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/shufflehandler/ShuffleHandler.java
@@ -204,6 +204,9 @@ public class ShuffleHandler implements AttemptRegistrationListener {
 
   public static final String CONNECTION_CLOSE = "close";
 
+  public static final String SHUFFLE_SSL_ENABLED_KEY = "hive.llap.shuffle.ssl.enabled";
+  public static final boolean SHUFFLE_SSL_ENABLED_DEFAULT = false;
+
   public static final String SUFFLE_SSL_FILE_BUFFER_SIZE_KEY =
     "llap.shuffle.ssl.file.buffer.size";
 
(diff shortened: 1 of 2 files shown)


llap-server/src/java/org/apache/hadoop/hive/llap/shufflehandler/ShuffleHandler.java
```diff
@@ -204,6 +204,9 @@ public class ShuffleHandler implements AttemptRegistrationListener {
 
   public static final String CONNECTION_CLOSE = "close";
 
+  public static final String SHUFFLE_SSL_ENABLED_KEY = "hive.llap.shuffle.ssl.enabled";
```
deniskuzZ, line 207: to be consistent with other configs, should we name it simply `llap.shuffle.ssl.enabled`?

deniskuzZ (commented):
LGTM, but I would ask for @abstractdog opinion

abstractdog:
> LGTM, but I would ask for @abstractdog opinion

1. can we fix the "SUFFLE" typos if we're already touching this part?
2. can you please share more details about the manual testing?
real cluster vs. some mini llap cluster?
Can this happen with a MiniLlapCluster as a test?

tez has TestSecureShuffle for this, which would be hard to adapt, but hive's own mini llap based tests make sense to me, as this is a new feature (at least in hive)
consider use tez unit test for guidance with setting up SSL:
https://github.com/apache/tez/blob/30e9cdbff7adb75c3553d5cbb8515761a7231e07/tez-tests/src/test/java/org/apache/tez/test/TestSecureShuffle.java#L241

PR #5155, opened 2024-03-22
HIVE-28129 Execute statement doesnot report the correct query string information

ramesh0201:
Execute Statement should report the correct query string. Because it inherits the conf from the prepare query's sem object, we need to update the query string.

Because currently execute queries report the prepare statement as their query string in all the logs.

No

No

mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=execute_query_string.q

### DIFF
--- a/ql/src/test/results/clientpositive/llap/prepare_plan.q.out
+++ b/ql/src/test/results/clientpositive/llap/prepare_plan.q.out
@@ -128,11 +128,11 @@ STAGE PLANS:
       Processor Tree:
         ListSink
 
-PREHOOK: query: prepare pcount from select count(*) from src where key > ?
+PREHOOK: query: explain execute pcount using '200'
 PREHOOK: type: EXECUTE QUERY
 PREHOOK: Input: default@src
 #### A masked pattern was here ####
-POSTHOOK: query: prepare pcount from select count(*) from src where key > ?
+POSTHOOK: query: explain execute pcount using '200'
 POSTHOOK: type: EXECUTE QUERY
 POSTHOOK: Input: default@src
 #### A masked pattern was here ####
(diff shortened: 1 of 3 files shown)


ql/src/test/queries/clientpositive/execute_query_string.q
```diff
@@ -0,0 +1,3 @@
+create table prepareexecute(a int, b int);
+prepare pcount from select count(*) from prepareexecute where a > ?;
+execute pcount using '200';
```
zabetak, line 3: The new test is redundant since the same query pattern appears in other .q files and the effects of the changes are clearly visible in the .q.out changes.

ql/src/test/results/clientpositive/llap/prepare_plan.q.out
```diff
@@ -128,11 +128,11 @@ STAGE PLANS:
       Processor Tree:
         ListSink
 
-PREHOOK: query: prepare pcount from select count(*) from src where key > ?
```
zabetak, line 131: The changes in the .q.out file are the expected ones however there seems to be something missing from the prepare statement implementation. After the changes introduced here the `query: prepare` pattern does not appear at all in the .q.out files. 

Is this normal? Should we log a follow-up ticket?

zabetak (approved):
LGTM, just some minor comments/questions but not blocking for the patch.

ql/src/test/queries/clientpositive/execute_query_string.q
ramesh0201: Yes you are right, I will remove the new q file

ql/src/test/results/clientpositive/llap/prepare_plan.q.out
ramesh0201: Prepare queries doesn't go through the execution phase as per current implementation https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/Driver.java#L166C30-L166C44
So the pre-execution and post-execution hooks are skipped as well, which is why we miss seeing those results in the q.out file. I feel it is ok. Just that we need to use compilation side hooks to test the prepare queries. explain..prepare still shows up in the q.out files.
So I am not sure if we have to create a follow up to test the prepare queries? what is your take?

zabetak: Thanks for the explanation. Let's leave it as is at the moment and if really needed we can create a JIRA later ;)

ramesh0201:
Thank you @zabetak for the review

PR #5160, opened 2024-03-25
HIVE-28110: MetastoreConf - String casting of default values breaks Hive

tanishq-chugh:
Modify the getVar(Configuration conf, ConfVars var) & getStringCollection(Configuration conf, ConfVars var) methods of the MetastoreConf class to use toString() method instead of type-casting.

Why are the changes needed?
When using the getVar(Configuration conf, ConfVars var) method of the MetastoreConf class, Apache breaks when e.g. trying to retrieve the environment variable "USE_SSL" and it isn't set in the system. The method then tries to cast the default value, which is the boolean false for USE_SSL, to a String which can't work. Also in the getStringCollection(Configuration conf, ConfVars var) method it tries to cast any default values to a String.

No

No

How was this patch tested?
Manual tests.
After making the necessary changes, we build the hive project again.
After manually testing some DDL and SQL commands, hive is working as expected.

### DIFF
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/conf/TestMetastoreConf.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/conf/TestMetastoreConf.java
@@ -510,4 +510,29 @@ public class TestMetastoreConf {
     Assert.assertEquals(MetastoreConf.ACID_OPEN_TXNS_COUNTER_SERVICE_CLASS,
         AcidOpenTxnsCounterService.class.getName());
   }
+
+  /**
+   * Verify that getVar(Configuration conf, ConfVars var) method of MetaStoreConf.java runs as expected in case of unset environment variables
+   * or any other variables that don't have default Value type of String
+   */
+  @Test
+  public void testGetVarForUnsetSystemProperties(){
+    conf = MetastoreConf.newMetastoreConf();
+    System.clearProperty(ConfVars.USE_SSL.getVarname());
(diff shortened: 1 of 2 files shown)


tanishq-chugh:
Added the Unit tests corresponding to the changes for the case specified in the JIRA. 
For other regular cases using the changes methods, the test named "default" in TestMetastoreConf.java suffices.  

CC: @ayushtkn

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/conf/TestMetastoreConf.java
```diff
@@ -510,4 +510,29 @@ public void testClassNames() {
...
   }
+
+  /**
+   * Verify that getVar(Configuration conf, ConfVars var) method of MetaStoreConf.java runs as expected in case of unset environment variables
+   * or any other variables that don't have default Value type of String
+   */
+  @Test
+  public void testGetVarForUnsetSystemProperties(){
+    conf = MetastoreConf.newMetastoreConf();
+    System.clearProperty(ConfVars.USE_SSL.getVarname());
```
ayushtkn, line 521: Why are you doing this `` System.clearProperty``, we don't do that, play with the conf, maybe do a conf.set(ConfVars.USE_SSL.getVarname(), null); or something like that

tanishq-chugh: I was using `System.clearProperty()` to replicate the required use case considering it being used in many other unit tests also, but now have changed it to `conf.unset()` .

PR #5162, opened 2024-03-26
HIVE-28069: Addendum: Iceberg: Implement Merge task functionality for Iceberg tables - Add serde properties for CTAS and CMV

SourabhBadhya:

HIVE-28069: Addendum: Iceberg: Implement Merge task functionality for Iceberg tables - Add serde properties for CTAS and CMV

If file format is missing, it uses Parquet which leads to incorrect file format exceptions.

No

No

QTest

### DIFF
--- a/iceberg/iceberg-handler/src/test/queries/positive/iceberg_merge_files.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/iceberg_merge_files.q
@@ -57,7 +57,7 @@ select count(distinct(file_path)) from default.parquet_part_source.files;
 
 -- Insert into the tables both for unpartitioned and partitioned cases for Parquet formats.
 insert into table parquet_source select * from parquet_source;
-insert into table parquet_part_source select * from parquet_part_source where ds = 102 union all select * from orc_part_source where ds = 103;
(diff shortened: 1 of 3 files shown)


iceberg/iceberg-handler/src/test/queries/positive/iceberg_merge_files.q
```diff
@@ -57,7 +57,7 @@ select count(distinct(file_path)) from default.parquet_part_source.files;
 
 -- Insert into the tables both for unpartitioned and partitioned cases for Parquet formats.
 insert into table parquet_source select * from parquet_source;
-insert into table parquet_part_source select * from parquet_part_source where ds = 102 union all select * from orc_part_source where ds = 103;
+insert into table parquet_part_source select * from parquet_part_source where ds = 102 union all select * from parquet_part_source where ds = 103;
```
deniskuzZ, line 60: why is this change? i think idea was to support multiple source formats

SourabhBadhya: @deniskuzz That is just keeping in line with other queries, it was a mistake while copying in the previous PR even though it works.
The previous PR works with multiple source formats. So its not a problem.

deniskuzZ (approved):
LGTM +1

PR #5164, opened 2024-03-26
HIVE-26926: SHOW PARTITIONS for a non partitioned table should just throw execution error instead of full stack trace.

tanishq-chugh:
SHOW PARTITIONS command implemented on a  non partitioned table should just throw execution error instead of full stack trace.

Why are the changes needed?
It is an improvement to already existing flow where a huge stack trace is shown to user wherein only a small prompt is useful.

Yes,
Earlier: 
Screenshot 2024-03-26 at 5 05 02 PM

Screenshot 2024-03-26 at 5 05 22 PM

This PR proposes the following change:
Screenshot 2024-03-26 at 5 05 58 PM

No

How was this patch tested?
Manual tests.
After making the necessary changes, we build hive again, and then ran some manual commands to make sure hive is working as expected as well as the required change is working fine.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
@@ -30,6 +30,7 @@ import java.util.Set;
 import org.apache.hadoop.hive.ql.ddl.table.create.CreateTableDesc;
 import org.apache.hadoop.hive.ql.exec.Task;
 import org.apache.hadoop.hive.ql.exec.repl.util.ReplUtils;
+import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.parse.ExplainConfiguration.AnalyzeState;
 import org.apache.hadoop.hive.ql.plan.api.StageType;
 import org.reflections.Reflections;
@@ -85,7 +86,24 @@ public final class DDLTask extends Task<DDLWork> implements Serializable {
       } else {
         throw new IllegalArgumentException("Unknown DDL request: " + ddlDesc.getClass());
       }
-    } catch (Throwable e) {
+    } catch(HiveException e){
+      int excpErrorCode = ReplUtils.handleException(work.isReplication(), e, work.getDumpDirectory(),
+              work.getMetricCollector(), getName(), conf);
+      if(excpErrorCode == 10241){
+        return excpErrorCode;
+      }else{
+        if(work.isReplication() && ReplUtils.shouldIgnoreOnError(ddlOperation, e)) {
+          LOG.warn("Error while table creation: ", e);
+          return 0;
+        }
+        failed(e);
+        if(ddlOperation != null) {
+          LOG.error("DDLTask failed, DDL Operation: " + ddlOperation.getClass().toString(), e);
+        }
+        return excpErrorCode;
+      }
+    }
+    catch (Throwable e) {
       if(work.isReplication() && ReplUtils.shouldIgnoreOnError(ddlOperation, e)) {
         LOG.warn("Error while table creation: ", e);
         return 0;
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
@@ -27,7 +27,6 @@ import java.util.Map;
 import java.util.Random;
 import java.util.Set;
 
-import org.apache.hadoop.hive.ql.ErrorMsg;
 import org.apache.hadoop.hive.ql.ddl.table.create.CreateTableDesc;
 import org.apache.hadoop.hive.ql.exec.Task;
 import org.apache.hadoop.hive.ql.exec.repl.util.ReplUtils;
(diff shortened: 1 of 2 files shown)


ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
```diff
@@ -85,7 +86,24 @@ public int execute() {
       } else {
         throw new IllegalArgumentException("Unknown DDL request: " + ddlDesc.getClass());
       }
-    } catch (Throwable e) {
+    } catch(HiveException e){
+      int excpErrorCode = ReplUtils.handleException(work.isReplication(), e, work.getDumpDirectory(),
```
deniskuzZ, line 90: is that a generic `HiveException` handling logic?

tanishq-chugh: Yes it is a generic logic. Just to make sure we don't show the stack trace only for this particular case, i have added conditional statement based on the error code.

ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
```diff
@@ -85,7 +86,24 @@ public int execute() {
...
+              work.getMetricCollector(), getName(), conf);
+      if(excpErrorCode == 10241){
```
deniskuzZ, line 92: what is this magic number, do we have constants for that?

ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
```diff
@@ -85,7 +86,24 @@ public int execute() {
...
+        return excpErrorCode;
+      }else{
```
deniskuzZ, line 94: that is the same as Throwable handling

ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
deniskuzZ: when `work.isReplication() && ReplUtils.shouldIgnoreOnError` you don't need to call handleException

ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
tanishq-chugh: image

Yes, it the Error code corresponding to "Table Not Partitioned" error we are handling, you can find the same in [https://github.com/apache/hive/blob/master/common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java](url)

deniskuzZ: then use it: `ErrorMsg.TABLE_NOT_PARTITIONED.getErrorCode()`

ql/src/java/org/apache/hadoop/hive/ql/ddl/DDLTask.java
```diff
@@ -90,12 +92,16 @@ public int execute() {
         LOG.warn("Error while table creation: ", e);
         return 0;
       }
+      int excpErrorCode = ReplUtils.handleException(work.isReplication(), e, work.getDumpDirectory(),
```
deniskuzZ, line 95: could you please rename var to simply `errorCode` and format the code: there should be space between `if` and opening `(`

deniskuzZ (commented):
LGTM, minor refactor ask

deniskuzZ (approved):
LGTM +1, pending tests

ayushtkn:
In your above image
This PR proposes the following change:
Screenshot 2024-03-26 at 5 05 58 PM

The error message has {0} instead of the table name

btw. did you test with the current changes?

deniskuzZ:
btw, I think it's better to change the relevant `ShowPartitionsOperation` execute method instead of base one:
````
context.getTask().setException(
    new HiveException(ErrorMsg.TABLE_NOT_PARTITIONED, desc.getTabName())
);
return 1;
````

**Comment by tanishq-chugh** [comment]
@ayushtkn yes, all the changes were tested locally, but didn't realise that table index was being shown instead of table name. With the latest commit, i have moved the logic to ShowPartitionOperation which is a better approach as suggested by @deniskuzZ , and with the latest change, output is as follows:

final_output

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/show/ShowPartitionsOperation.java [hunk 5]
```diff
@@ -58,7 +58,8 @@ public int execute() throws HiveException {
     if (tbl.isNonNative() && tbl.getStorageHandler().supportsPartitionTransform()) {
       parts = tbl.getStorageHandler().showPartitions(context, tbl);
     } else if (!tbl.isPartitioned()) {
-      throw new HiveException(ErrorMsg.TABLE_NOT_PARTITIONED, desc.getTabName());
+      context.getTask().setException(new HiveException(ErrorMsg.TABLE_NOT_PARTITIONED, desc.getTabName()));
+      return ErrorMsg.TABLE_NOT_PARTITIONED.getErrorCode();
```
deniskuzZ, line 62: return 1;

tanishq-chugh: @deniskuzZ , With your suggested change, the output is as follows:

output

`return code 1` is printed on the console differing from the expected output of `return code 10241` as per [HIVE-26926](https://issues.apache.org/jira/browse/HIVE-26926). 

Should the change be made?

deniskuzZ: no, please don't

PR #5166, opened 2024-03-27
HIVE-28154: Throw friendly exception if the table does not support partition transform

zhangbutao:

At present, create basic tables such as orc/parquet with partition transform will succeed. However, we only support iceberg tables for partition transform now, so i think we need to throw a friendly exception if the table does not support partition transform, to avoid user misunderstanding that the basic tables have such ability.

We can throw friendly exception to client if the table does not support partition transform, such as the following exception:
```
0: jdbc:hive2://127.0.0.1:10000/default> create table customer(customer_id int,first_name string,last_name string) partitioned by spec(TRUNCATE(2, last_name)) stored as orc;
Error: Error while compiling statement: FAILED: SemanticException Partition transforms are only supported by Iceberg storage handler (state=42000,code=40000)
```

Throw friendly exception if the table does not support partition transform to avoid user misunderstanding.

No

No

Added UT.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
@@ -14496,8 +14496,7 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
         }
       }
     } else if (partitionTransformSpecExists) {
-      throw new SemanticException("Partition transform is only supported for specific storage handler tables. e.g. " +
-              "Iceberg tables");
+      throw new SemanticException("Partition transforms are only supported by Iceberg storage handler");
     }
   }
 
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
@@ -14495,6 +14495,9 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
                           + "'='" + fileFormat + "')");
         }
       }
+    } else if (partitionTransformSpecExists) {
+      throw new SemanticException("Partition transform is only supported for specific storage handler tables. e.g. " +
+              "Iceberg tables");
     }
   }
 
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -14495,6 +14495,9 @@ private void validateStorageFormat(
                           + "'='" + fileFormat + "')");
         }
       }
+    } else {
+      throw new SemanticException("Partition transform is only supported for specific storage handler tables. e.g. " +
```
zhangbutao, line 14499: If you think the exception hint is not good, please feel free to give your opinion here.
Thanks.

deniskuzZ: make sense, I think we could be more specific "Partition transforms are only supported by Iceberg storage handler"

zhangbutao: Changed.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -14495,6 +14495,8 @@ private void validateStorageFormat(
                           + "'='" + fileFormat + "')");
         }
       }
+    } else if (partitionTransformSpecExists) {
+      throw new SemanticException("Partition transforms are only supported by Iceberg storage handler");
```
okumin, line 14499: Hopefully, we will have a generic message like "The default Hive storage handler doesn't support partition transforms" so that we can add a new one that supportsPartitionTransform.

okumin (approved):

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
okumin: I found we already had a discussion and intentionally chosen this message. So, it sounds ok as it is.
https://github.com/apache/hive/pull/5166#discussion_r1540632711

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
deniskuzZ: @zhangbutao, should we add this to ErrorMsg?

zhangbutao: Done. Added exception to ErrorMsg.

deniskuzZ (approved):
+1

zhangbutao:
http://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5166/6/tests/ A known flaky  test  `TestConflictingDataFiles.testMultiFiltersUpdate`

PR #5169, opened 2024-03-28
HIVE-28161: Incorrect Copyright years in META-INF/NOTICE files

zabetak:
Set the necessary properties in the pom file to generate an appropriate Copyright timespan (used by maven-remote-resources-plugin).

The generated META-INF/NOTICE file which resides inside each jar produced by Hive has incorrect copyright years.

Inside all jars the NOTICE file has the following incorrect content:

```
Copyright 2020 The Apache Software Foundation
```
The Copyright statement should include the timespan from the inception of the project to now.
```
Copyright 2008-2024 The Apache Software Foundation
```
The problem can be easily seen by inspecting the jar content after building any module or checking the previously published jars in Maven central.

No

No

Inspecting various jars before/after the changes
```
mvn clean install -DskipTests -pl common/
jar xf common/target/hive-common-4.1.0-SNAPSHOT.jar META-INF
cat META-INF/NOTICE 
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,7 @@
   <packaging>pom</packaging>
   <name>Hive</name>
   <url>https://hive.apache.org</url>
+  <inceptionYear>2008</inceptionYear>
   <modules>
     <module>storage-api</module>
     <module>accumulo-handler</module>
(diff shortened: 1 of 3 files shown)


deniskuzZ (approved):
+1

dengzhhu653:
Hi @zabetak, hive-standalone-metastore-server-4.1.0-SNAPSHOT.jar this jar still has the incorrect content, cloud you please check that? thank you!
```
Copyright 2020 The Apache Software Foundation

dengzhhu653 (approved):
Thank you @zabetak! should we fix the same on storage-api as well, though it's released  on a different line, but I find the jar in the packaged tarball.

zabetak:
@dengzhhu653 You are right, we should also add it in storage-api module. I took care of it in https://github.com/apache/hive/pull/5169/commits/8b0469c4a8859f0fff224a5fd3444a011b152964

I keep forgetting that we have these two modules disconnected from the main pom.

PR #5172, opened 2024-03-29
HIVE-28164: Remove log4j:log4j transitive dependency

Aggarwal-Raghav:
There are dependency like accumulo and slf4j bringing log4j vulnerable jars in dependency tree. There are few dependencies also which don't appear in dependecy tree but are bringing old and vulnerable log4j. This can be observed in local m2 repo cache.

For CVE's and security reasons.

NO

Yes, here is new dependency tree on 771b0038d84758897c80d6931f21ff7073f4da31 (master): 
[dependency_tree.txt](https://github.com/user-attachments/files/17363530/dependency_tree.txt)

Will see the UT from CI

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
     <maven.cyclonedx.plugin.version>2.7.10</maven.cyclonedx.plugin.version>
     <maven.license.plugin.version>2.3.0</maven.license.plugin.version>
     <!-- Library Dependency Versions -->
-    <accumulo.version>1.10.1</accumulo.version>
+    <accumulo.version>1.10.4</accumulo.version>
     <ant.version>1.10.13</ant.version>
     <antlr.version>3.5.2</antlr.version>
     <!-- Make sure to sync it with standalone-metastore/pom.xml -->
@@ -1607,11 +1607,17 @@
                   <excludes>
                     <!-- Move to SLF4J -->
                     <exclude>commons-logging:commons-logging</exclude>
-                    <exclude>log4j:log4j</exclude>
                     <exclude>ch.qos.reload4j:reload4j</exclude>
                   </excludes>
                   <searchTransitive>false</searchTransitive>
                   <message>A banned logging dependency was found!</message>
+                </bannedDependencies>
+                <bannedDependencies>
+                  <excludes>
+                    <exclude>log4j:log4j</exclude>
+                  </excludes>
+                  <searchTransitive>true</searchTransitive>
+                  <message>Banned log4j:log4j dependency/transitive dependency was found!</message>
                 </bannedDependencies>
               </rules>
               <fail>true</fail>
--- a/standalone-metastore/metastore-tools/metastore-benchmarks/pom.xml
+++ b/standalone-metastore/metastore-tools/metastore-benchmarks/pom.xml
@@ -84,11 +84,6 @@
       <artifactId>jmh-generator-annprocess</artifactId>
       <version>${jmh.version}</version>
     </dependency>
-    <!-- https://mvnrepository.com/artifact/org.slf4j/slf4j-log4j12 -->
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
-    </dependency>
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-slf4j-impl</artifactId>


Aggarwal-Raghav:
@zabetak, if you have bandwidth can you please review and provide your inputs on this?

standalone-metastore/metastore-server/pom.xml
```diff
@@ -720,6 +720,14 @@
         <groupId>org.datanucleus</groupId>
         <artifactId>datanucleus-maven-plugin</artifactId>
         <version>5.2.1</version>
+        <!-- The plexus-contianer-default can be removed once we move to datanucleus-maven-plugin version 6.0.0-release-->
+        <dependencies>
+          <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-container-default</artifactId>
+            <version>1.5.6</version>
```
zabetak, line 728: Use properties instead of hardcoded versions.

zabetak (changes requested):
If we want to eliminate log4j completely from the project then we should also tune the [respective enforcer rules](https://github.com/apache/hive/blob/855e4055675e3c993a61f59501f783e641abaaa6/pom.xml#L1652) to prevent this from reappearing.

Aggarwal-Raghav:
Apologies for the late response, Have updated the PR with suggested changes. I have created a new enforcer execution step to search for transitive dependencies of log4j. The scope of this PR is to only exclude log4j.

Aggarwal-Raghav:
@zabetak , can you please help with review?

standalone-metastore/metastore-tools/metastore-benchmarks/pom.xml
```diff
@@ -88,6 +88,12 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-log4j12</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>log4j</groupId>
+          <artifactId>log4j</artifactId>
+        </exclusion>
+      </exclusions>
```
zabetak, line 96: The slf4j-log4j12 is an explicit binding of slf4j with log4j 1.2 thus this exclusion here doesn't make sense. The whole dependency must be dropped.

Just below we have log4j-slf4j-impl which is an SLF4J binding with Log4j 2 so I think that's all we need.

beeline/pom.xml
```diff
@@ -276,6 +276,11 @@
             <artifactId>maven-shade-plugin.log4j2-cachefile-transformer</artifactId>
             <version>2.1</version>
           </dependency>
+          <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>${log4j2.version}</version>
+          </dependency>
```
zabetak, line 283: Why do we need to add this? Did we use log4j transitively?

standalone-metastore/metastore-tools/metastore-benchmarks/pom.xml
Aggarwal-Raghav: understood.

beeline/pom.xml
Aggarwal-Raghav: it is bringing log4j-core 2.1 version which is also has lot of CVE's (https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-core/2.1)
Screenshot 2024-05-29 at 4 45 22 PM

zabetak: Then the scope of this PR is wider than what is currently described in the ticket. Please adapt the description accordingly. 

It seems that the PR wants to avoid also using log4j 2 versions lower than X. In that case, I think we will need to revisit the enforcement rules to cover that case as well otherwise we cannot tell if it is handled properly in all modules.

Aggarwal-Raghav: Hi @zabetak, you are correct, I want to exclude log4j 2 versions lower than X.

I tried modifying the enforcement rules to prevent log4j-core less than 2.17.1 (based on https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-core) like this
`<exclude>org.apache.logging.log4j:log4j-core:(,2.17.1)</exclude>`

But there are few problems:
	1. Druid-handler still requires log4j 2.8.2 (coming transitively from druid-server).
	2. Even with the enforcement rule, the log4j-core 2.1 coming in beeline (attached in previous comment snapshot) is not getting captured with this and still passing. It seems the rule are enforced on maven dependency tree output and as this beeline log4j-core is not present in dependency tree, it is getting enforced.
	
What's next:
	1. I can remove this beeline change and update the PR. Then I think the scope of the PR and work done in PR is aligned for log4j:log4j
	2. In description explicitly mention only this beeline pom change.

Let me know if you have any better way to resolve this.

Aggarwal-Raghav:
@zabetak , can you please re-review it, if you have the bandwidth

zabetak:
I forgot about this issue. I will take a a look now.

beeline/pom.xml
zabetak: The `log4j-core:2.1` version is a dependency of https://github.com/edwgiz/maven-shaded-log4j-transformer plugin. There seems to be a 1-to-1 mapping between the plugin and the log4j-core version thus it doesn't make sense to declare an explicit dependency to log4j-core. If we want to avoid the usage of the old log4j version we should consider changing the version of the transformer. Nevertheless this is out of the scope of the current PR so let's track it under another JIRA if needed.

The `log4j-core:2.1` version is used by the shade plugin during the Hive build but it is not a Hive dependency thus its not used by the product. Since it is not used by the product it is not captured by the enforcer rule. I don't know of a good way to ban plugin dependencies and not sure if it is necessary or worth the effort investing in such tooling. If a plugin is deemed unsafe then we should either ban the plugin usage or bump its version.

standalone-metastore/metastore-server/pom.xml
```diff
@@ -720,6 +720,14 @@
...
         <artifactId>datanucleus-maven-plugin</artifactId>
         <version>5.2.1</version>
+        <!-- The plexus-contianer-default can be removed once we move to datanucleus-maven-plugin version >= 6.0.0-release-->
+        <dependencies>
+          <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-container-default</artifactId>
+            <version>${plexus.version}</version>
+          </dependency>
+        </dependencies>
```
zabetak, line 730: The same argument that holds for the shade plugin and `log4j-core:2.1` also applies here. The fact that datanucleus plugin has a dependency to log4j is not very problematic since it is not a Hive product/module dependency. log4j will only be used during the build and this usage is rather fine.

Nevertheless, if you are confident that upgrading the plexus `1.5.6` version is fully compatible with `5.2.1` I am OK to make this change although I don't find it necessary. I checked the enhanced classes before and after the change and they seem identical so I am ok to leave this as is.

pom.xml
```diff
@@ -1617,6 +1616,24 @@
...
+            </goals>
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <excludes>
+                    <exclude>log4j:log4j</exclude>
+                  </excludes>
+                  <searchTransitive>true</searchTransitive>
+                  <message>A banned transitive logging dependency was found!</message>
+                </bannedDependencies>
```
zabetak, line 1632: This code snippet is already a rule so you don't need to copy-paste the entire execution block. You can just put this snippet in the existing `enforce-banned-dependencies-logging` execution. You could also use https://maven.apache.org/enforcer/enforcer-rules/banTransitiveDependencies.html rule to make it slightly less verbose.

zabetak (commented):
@Aggarwal-Raghav I left some small comments and then we can merge this. Let me know if you have time to address them otherwise I can commit the final touches.

Aggarwal-Raghav:
Thanks @zabetak , for the review. your insights are always appreciated. I will address the review comments :-)

zabetak (approved):
LGTM, will merge soon!

PR #5173, opened 2024-03-30
HIVE-28166: Iceberg: Truncate on branch operates on the main table.

ayushtkn:

Truncate on iceberg branch should work on that branch not on the main table itself

Else truncate operations on the branch wipes out the original table.

Yes, If you call truncate on an iceberg branch it will get truncated instead of the main table

No

UT

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -1772,6 +1772,10 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   public void validatePartSpec(org.apache.hadoop.hive.ql.metadata.Table hmsTable, Map<String, String> partitionSpec)
       throws SemanticException {
     Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
+    if (hmsTable.getSnapshotRef() != null && hasUndergonePartitionEvolution(table)) {
+      // for this case we rewrite the query as delete query, so validations would be done as part of delete.
+      return;
+    }
 
     if (table.spec().isUnpartitioned() && MapUtils.isNotEmpty(partitionSpec)) {
       throw new SemanticException("Writing data into a partition fails when the Iceberg table is unpartitioned.");
@@ -1817,6 +1821,8 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
     Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
     if (MapUtils.isEmpty(partitionSpec) || !hasUndergonePartitionEvolution(table)) {
       return true;
+    } else if (hmsTable.getSnapshotRef() != null) {
+      return false;
     }
 
     Expression finalExp = generateExpressionFromPartitionSpec(table, partitionSpec);
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -1530,7 +1530,7 @@ public class Hive {
       List<String> partNames = ((null == partSpec)
               ? null : getPartitionNames(table.getDbName(), table.getTableName(), partSpec, (short) -1));
       if (snapshot == null) {
-        getMSC().truncateTable(table.getDbName(), table.getTableName(), partNames);
+        getMSC().truncateTable(table.getFullTableName(), partNames);
       } else {
         boolean truncateUseBase = HiveConf.getBoolVar(conf, HiveConf.ConfVars.HIVE_ACID_TRUNCATE_USE_BASE)
           || HiveConf.getBoolVar(conf, HiveConf.ConfVars.HIVE_ACID_LOCKLESS_READS_ENABLED);
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
@@ -2098,7 +2098,7 @@ public class HiveMetaStoreClient implements IMetaStoreClient, AutoCloseable {
 
   @Override
   public void truncateTable(TableName table, List<String> partNames) throws TException {
-    truncateTableInternal(getDefaultCatalog(conf), table.getDb(), table.getTable(), table.getTableMetaRef(), partNames,
+    truncateTableInternal(table.getCat(), table.getDb(), table.getTable(), table.getTableMetaRef(), partNames,
         null, -1, true);
   }
 
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
@@ -121,6 +121,8 @@ public class HiveMetaStoreClient implements IMetaStoreClient, AutoCloseable {
   public static final String MANUALLY_INITIATED_COMPACTION = "manual";
   public static final String TRUNCATE_SKIP_DATA_DELETION = "truncateSkipDataDeletion";
   public static final String SKIP_DROP_PARTITION = "dropPartitionSkip";
+
+  public static final String SNAPSHOT_REF = "snapshot_ref";
   public static final String RENAME_PARTITION_MAKE_COPY = "renamePartitionMakeCopy";
 
   /**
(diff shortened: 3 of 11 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -2104,9 +2106,14 @@ public void truncateTable(String catName, String dbName, String tableName, List<
   private void truncateTableInternal(String catName, String dbName, String tableName,
       List<String> partNames, String validWriteIds, long writeId, boolean deleteData)
           throws TException {
+    EnvironmentContext context = new EnvironmentContext();
+    if (tableName.contains(".")) {
```
deniskuzZ, line 2110: should we use `HiveTableName.of(tableName)` instead?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -1530,7 +1530,11 @@ public void truncateTable(String dbDotTableName, Map<String, String> partSpec, L
       List<String> partNames = ((null == partSpec)
               ? null : getPartitionNames(table.getDbName(), table.getTableName(), partSpec, (short) -1));
       if (snapshot == null) {
-        getMSC().truncateTable(table.getDbName(), table.getTableName(), partNames);
+        if (table.getSnapshotRef() != null && table.getStorageHandler().isTableMetaRefSupported()) {
```
deniskuzZ, line 1533:
````
String fullTableName = table.getTableName();
if (table.getSnapshotRef() != null && table.getStorageHandler().isTableMetaRefSupported())  {
 fullTableName += "." + table.getSnapshotRef();
}
````
can we move this into table.getFullTableName?

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/misc/truncate/TruncateTableAnalyzer.java [hunk 3]
```diff
@@ -87,7 +87,7 @@ public void analyzeInternal(ASTNode root) throws SemanticException {
     ASTNode tableNode = (ASTNode) root.getChild(0); // TOK_TABLE_PARTITION
     String tableNameString = getUnescapedName((ASTNode) tableNode.getChild(0));
     Table table = getTable(tableNameString, true);
-    TableName tableName = HiveTableName.of(table);
+    TableName tableName = HiveTableName.ofNullable(table.getTableName(), table.getDbName(), table.getSnapshotRef());
```
deniskuzZ, line 90: i think we should add snapshot ref to HiveTableName.of

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1817,6 +1821,8 @@ public boolean canUseTruncate(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
     Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
     if (MapUtils.isEmpty(partitionSpec) || !hasUndergonePartitionEvolution(table)) {
       return true;
+    } else if (hmsTable.getSnapshotRef() != null) {
```
deniskuzZ, line 1824: why can't we use truncate when branch is provided?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1772,6 +1772,10 @@ private List<String> getPartitions(DDLOperationContext context, Configuration jo
   public void validatePartSpec(org.apache.hadoop.hive.ql.metadata.Table hmsTable, Map<String, String> partitionSpec)
       throws SemanticException {
     Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
+    if (hmsTable.getSnapshotRef() != null && hasUndergonePartitionEvolution(table)) {
```
deniskuzZ, line 1775: would this even be called if canUseTruncate returns false?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: this is branch is involved & the table has undergone partition evolution, if it didn't went through partition evolution the above if will return true for branch as well

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: it will be called before can use truncate, it is called as part of ``addTruncateTableOutputs``

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
ayushtkn: ``HiveTableName`` is part of hive-exec, that ain't a dependency here :-(

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
ayushtkn: we can do but we already have on getFullTableName
```
  public TableName getFullTableName() {
```
this will create a ``TableName`` object & return, from which we get the concatenated table name, so object creation will redundant & Java won't let us create another ``getFullTableName`` with just different return type, will any other name work?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: oh, but why do we need to check the branch name if table has undergone partition evolution, shouldn't we fallback right away?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
deniskuzZ: hm, we already have `getFullTableName()` and deprecated `getFullyQualifiedName()`
maybe let's keep in inline?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
deniskuzZ: :(

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: below there is some more optimisation for normal table, I don't catch it well but it is checking if the table has undergone partition evolution, but there is no datafile outside the new partition or so, so in that case even in case of partition evolution you can go with truncate but that would throw you an NPE, in case of branches if the partition column is removed in the current table spec.

deniskuzZ (approved):
LGTM +1

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
deniskuzZ, line 2110: wait, are we doing this because we concatenate in API? can we extend API with `String tableMetaRef`

deniskuzZ (changes requested):

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
ayushtkn: As discussed offline, resloving

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -2079,34 +2081,40 @@ public void dropTable(String catName, String dbname, String name, boolean delete
...
+    truncateTableInternal(getDefaultCatalog(conf), dbName, tableName, null, partNames, validWriteIds, writeId,
+        deleteData);
   }
 
   @Override
   public void truncateTable(String dbName, String tableName, List<String> partNames,
       String validWriteIds, long writeId) throws TException {
-    truncateTableInternal(getDefaultCatalog(conf),
-        dbName, tableName, partNames, validWriteIds, writeId, true);
+    truncateTableInternal(getDefaultCatalog(conf), dbName, tableName, null, partNames, validWriteIds, writeId, true);
```
deniskuzZ, line 2091:
````
truncateTable(dbName, tableName, partNames, validWriteIds, writeId, true)
````

File: standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java [hunk 7]
```diff
@@ -2079,34 +2081,40 @@ public void dropTable(String catName, String dbname, String name, boolean delete
...
+  public void truncateTable(TableName table, List<String> partNames) throws TException {
+    truncateTableInternal(getDefaultCatalog(conf), table.getDb(), table.getTable(), table.getTableMetaRef(), partNames,
+        null, -1, true);
   }
 
   @Override
   public void truncateTable(String catName, String dbName, String tableName, List<String> partNames)
       throws TException {
-    truncateTableInternal(catName, dbName, tableName, partNames, null, -1, true);
+    truncateTableInternal(catName, dbName, tableName, null, partNames, null, -1, true);
```
deniskuzZ, line 2108:
````
truncateTable(TableName.fromString(tableName, catName, dbName), partNames)
````

File: standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java [hunk 8]
```diff
@@ -2079,34 +2081,40 @@ public void dropTable(String catName, String dbname, String name, boolean delete
...
   }
 
-  private void truncateTableInternal(String catName, String dbName, String tableName,
+  private void truncateTableInternal(String catName, String dbName, String tableName, String ref,
       List<String> partNames, String validWriteIds, long writeId, boolean deleteData)
           throws TException {
+    EnvironmentContext context = new EnvironmentContext();
```
deniskuzZ, line 2114: don't move the EnvironmentContext, just add the `tableMetaRef`

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
ayushtkn: done

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
ayushtkn: done

deniskuzZ (approved):
LGTM +1

PR #5174, opened 2024-03-31
HIVE-28165 HiveSplitGenerator: send splits through filesystem instead of RPC in case of big payload

abstractdog:
NOTE: this patch won't compile until TEZ-4548 is not released in Tez 0.10.4, because it uses new API InputDataInformationEvent.createWithSerializedPath(...) and InputInitializerContext.getVertexId().
I ran a full recommit test against a downstream tez containing this change, that's how I got qtest and unit test results.

Serialize splits to filesystem according to their size.

To avoid holding too much split payload in memory, as they're are collected in a loop and returned as a list.

No.

No.

Cluster testing:
- logs sanity-checked
- OOM issue was resolved
Unit testing:
- hive unit tests
- tez unit tests


ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,90 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+   * becoming the bottleneck (as write() is called from a loop currently).
+   */
+  class SplitSerializer {
+    // fields needed for filepath
+    private String queryId;
+    private String inputName;
+    private int vertexId;
+    private Path appStagingPath;
+    // metrics
+    private AtomicInteger timeSpentWithSplitWriteMs = new AtomicInteger(0);
```
InvisibleProgrammer, line 181: What is the reason why all the private fields are initialized in `lazyInit` but the AtomicInteger fields are not?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,90 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+     * No need to synchronize, this is called from a loop.
+     */
+    private void lazyInit() throws IOException {
+      if (fs != null) {
+        return;
+      }
+      queryId = jobConf.get(HiveConf.ConfVars.HIVEQUERYID.varname);
+      inputName = getContext().getInputName();
+      vertexId = getContext().getVertexId();
+      appStagingPath = TezCommonUtils.getTezSystemStagingPath(conf, getContext().getApplicationId().toString());
```
InvisibleProgrammer, line 198: Are you sure that `getContext()` cannot give back null? 

The reason why I ask is I already found multiple code path in HiveSplitGenerator that assumes it gives back a null value. Example: https://github.com/apache/hive/blob/0bc624ade3020826b4e7101b7ea4789be7e8585b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java#L261

abstractdog: no reason, I can initialize them there also

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: no reason, I can initialize them there also

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: btw, this is not the only path to be refactored in HiveSplitGenerator: we have a conf and jobConf field as well which is confusing, I'll create a follow-up ticket to cleanup things here

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,90 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+      lazyInit();
+
+      InputDataInformationEvent diEvent;
+      Path filePath = getSerializedFilePath(count);
+
+      // parallel writes for better performance (this is called from a loop)
+      executor.submit(() -> {
+        try {
+          long now = Time.monotonicNow();
+          try (FSDataOutputStream out = fs.create(filePath, false)) {
```
InvisibleProgrammer, line 216: out is Closeable. Are you sure it is necessary to explicitly `call out.close()`?

abstractdog: oh right, that's why I used try with resource clause, so out.close is not expected

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,90 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+      // e.g. staging_dir/events/queryid/inputtable_InputDataInformationEvent_0
+      return new Path(String.format("%s/events/%s/%d_%s_InputDataInformationEvent_%d", appStagingPath, queryId,
+          vertexId, inputName, index));
+    }
+
+    private void close() throws IOException {
+      if (fs != null) {
+        executor.shutdown();
+        try {
+          executor.awaitTermination(5, TimeUnit.MINUTES);
```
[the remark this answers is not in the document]
abstractdog: yes, I think so, when we can tell how much time is enough to do certain operations
this particularly should end within a minute (few seconds), so 5 minutes is the worst case: if it fails to finish in 5 minutes, the split generation is fine to die
regarding config: it would make no sense, awaitTermination consumes a timeout, so in development time I had the chance to do a reasonable worst case, and I don't want users to tune it

ql/src/test/results/clientpositive/llap/tez_split_serialization.q.out
```diff
@@ -0,0 +1,44 @@
...
+PREHOOK: type: QUERY
+PREHOOK: Input: default@test_table
+#### A masked pattern was here ####
+POSTHOOK: query: select * from test_table
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@test_table
+#### A masked pattern was here ####
+ABC	1
+DEF	2
+PREHOOK: query: select * from test_table
```
InvisibleProgrammer, line 35: I see no difference between the two select * output. Is the explain analyze provides better output to show what happens here?

abstractdog: it's not supposed to have any changes, it should just work with the new serialized split the same way
this detail is not exposed via explain also

zratkai:
As you mentioned: "NOTE: this patch won't compile until TEZ-4548 is not released in Tez 0.10.4", I guess a pom.xml change would be necessary as well, but I don't see in the list of modified files.

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,90 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+      if (fs != null) {
+        return;
+      }
+      queryId = jobConf.get(HiveConf.ConfVars.HIVEQUERYID.varname);
+      inputName = getContext().getInputName();
+      vertexId = getContext().getVertexId();
+      appStagingPath = TezCommonUtils.getTezSystemStagingPath(conf, getContext().getApplicationId().toString());
+
+      fs = appStagingPath.getFileSystem(jobConf);
+      executor = Executors.newFixedThreadPool(8,
```
InvisibleProgrammer, line 201: Why 8 threads?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,90 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+          long now = Time.monotonicNow();
+          try (FSDataOutputStream out = fs.create(filePath, false)) {
+            mrSplit.writeTo(out);
+            out.close();
+          }
+          splitsWritten.getAndIncrement();
+          long elapsed = Time.monotonicNow() - now;
+          timeSpentWithSplitWriteMs.getAndAdd((int) elapsed);
+          LOG.debug("Split #{} event to output path: {} written in {} ms", count, filePath, elapsed);
+        } catch (IOException e) {
```
InvisibleProgrammer, line 224: The `write` method throws IOException. And now we catch IOException and throw RuntimeException instead. But I don't see when the RuntimeException is handled.

InvisibleProgrammer (approved):
I had only minor questions, on overall, LGTM.

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: throwing RuntimeException actually means that I have to catch IOException because it's a checked exception, but I cannot help with that, so I don't really care who catches it and when: it's most probably bubble up to the split generation root, causing a DAG initialization failure

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: no specific reason, 8 threads led to satisfying performance improvement by ruling out the bottleneck for writing the events serially, but without having too many threads (which is also suboptimal, as every thread maintains its own stack)

abstractdog:
> As you mentioned: "NOTE: this patch won't compile until [TEZ-4548](https://issues.apache.org/jira/browse/TEZ-4548) is not released in Tez 0.10.4", I guess a pom.xml change would be necessary as well, but I don't see in the list of modified files.

Tez 0.10.4 has not yet been released, not even planned yet, so this PR is blocked until that
so I could just expect a formal code review here

abstractdog:
> I had only minor questions, on overall, LGTM.

thanks a lot, I addressed what was possible and replied about the rest, please let me know if you can see anything else

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
InvisibleProgrammer: Could you please share the follow-up tickets with me? I would be happy to help

abstractdog: https://issues.apache.org/jira/browse/HIVE-28178

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -4756,6 +4756,10 @@ public static enum ConfVars {
         "Class to use for calculating available slots during split generation"),
     HIVE_TEZ_GENERATE_CONSISTENT_SPLITS("hive.tez.input.generate.consistent.splits", true,
         "Whether to generate consistent split locations when generating splits in the AM"),
+    HIVE_TEZ_SPLIT_FS_SERIALIZATION_THRESHOLD("hive.tez.split.fs.serialization.threshold", 524288,
```
deniskuzZ, line 4759: should this be split size config or global input config? we might have huge number of split causing OOM

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -444,4 +572,12 @@ private AvailableSlotsCalculator getAvailableSlotsCalculator() throws Exception
     slotsCalculator.initialize(conf, this);
     return slotsCalculator;
   }
+
+  /**
+   * Convenience method for callers that want to disable the serialization of splits to filesystem.
+   */
+  public HiveSplitGenerator splitFsSerialization(boolean splitFsSerialization) {
```
deniskuzZ, line 579: wait, input param is not used at all

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
deniskuzZ: numThreads should be configurable

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: LOL, right, I just used this for disabling :D
I'll fix this to:
{code}
  public HiveSplitGenerator splitFsSerialization(boolean splitFsSerialization) {
    HiveConf.setIntVar(jobConf, HiveConf.ConfVars.HIVE_TEZ_SPLIT_FS_SERIALIZATION_THRESHOLD,
        splitFsSerialization ? HiveConf.ConfVars.HIVE_TEZ_SPLIT_FS_SERIALIZATION_THRESHOLD.getDefaultValue() : -1);
    return this;
  }
{code}

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,92 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+      // parallel writes for better performance (this is called from a loop)
+      executor.submit(() -> {
+        try {
+          long now = Time.monotonicNow();
+          try (FSDataOutputStream out = fs.create(filePath, false)) {
+            mrSplit.writeTo(out);
+          }
+          splitsWritten.getAndIncrement();
+          long elapsed = Time.monotonicNow() - now;
+          timeSpentWithSplitWriteMs.getAndAdd((int) elapsed);
```
deniskuzZ, line 224: that won't be accurate if write is done in parallel

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
deniskuzZ: we should log and rethrow the IO exception

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,92 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+          throw new RuntimeException(e);
+        }
+      });
+      // return the event now, we'll wait for the actual file write in close() if needed
+      return InputDataInformationEvent.createWithSerializedPath(count, filePath.toString());
+    }
+
+    private Path getSerializedFilePath(int index) {
+      // e.g. staging_dir/events/queryid/inputtable_InputDataInformationEvent_0
+      return new Path(String.format("%s/events/%s/%d_%s_InputDataInformationEvent_%d", appStagingPath, queryId,
```
deniskuzZ, line 236: format should go to constants

abstractdog: ack

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: right, I was thinking about the same

in case of a global threshold, we'll serialize every split to fs once we reach that regardless of the size, so we might end up serializing smaller splits also (20-30Kb), but that's fine as we remain stable...we need to figure out a reasonable default, e.g. 256MB?

-1: disabled
0: serialize everything to fs
n: global threshold

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,92 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+      splitsWritten = new AtomicInteger(0);
+
+      fs = appStagingPath.getFileSystem(jobConf);
+      executor = Executors.newFixedThreadPool(8,
+          new ThreadFactoryBuilder().setDaemon(true)
+              .setNameFormat("HiveSplitGenerator.SplitSerializer Thread - " + "#%d").build());
+    }
+
+    private InputDataInformationEvent write(int count, MRSplitProto mrSplit) throws IOException {
+      lazyInit();
```
deniskuzZ, line 210: lazyInit is initialiizing shared variables that are used inside the executor job, that could lead to race condition

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: that would be a "log and throw" antipattern, which I don't really have strong opinion about, however I believe it will be logged on the top of the AM stack

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
deniskuzZ, line 210: why do you need lazyInit inside of the write method, can't you just do it in the constructor?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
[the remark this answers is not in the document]
deniskuzZ: you should check `executor != null` + close the fs as well

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -362,13 +461,42 @@ private List<Event> createEventList(boolean sendSerializedEvents, InputSplitInfo
...
+      long inMemoryPayloadSize = 0;
+      long serializedPayloadSize = 0;
+
+      int payloadSerializationThresholdBytes =
+          HiveConf.getIntVar(jobConf, HiveConf.ConfVars.HIVE_TEZ_SPLIT_FS_SERIALIZATION_THRESHOLD);
+      SplitSerializer splitSerializer = new SplitSerializer();
+
+      List<MRSplitProto> splits = inputSplitInfo.getSplitsProto().getSplitsList();
+
+      LOG.info("Start creating events for {} splits", splits.size());
```
deniskuzZ, line 474: should it be under debug?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -362,13 +461,42 @@ private List<Event> createEventList(boolean sendSerializedEvents, InputSplitInfo
...
+
+      for (MRSplitProto mrSplit : splits) {
+        ByteBuffer payloadBuffer = mrSplit.toByteString().asReadOnlyByteBuffer();
+        int payloadSize = payloadBuffer.limit();
+        boolean shouldSerializeEventToFile =
+            payloadSerializationThresholdBytes != -1 && payloadSize > payloadSerializationThresholdBytes;
+        LOG.debug("Split #{} ByteBuffer size: {} bytes, serialize to file: {} (threshold: {} bytes)", count, payloadSize,
```
deniskuzZ, line 481: Split #{}, byteBuffer size: {} bytes

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +165,92 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+      inputName = getContext().getInputName();
+      vertexId = getContext().getVertexId();
+      appStagingPath = TezCommonUtils.getTezSystemStagingPath(conf, getContext().getApplicationId().toString());
+
+      timeSpentWithSplitWriteMs = new AtomicInteger(0);
+      splitsWritten = new AtomicInteger(0);
+
+      fs = appStagingPath.getFileSystem(jobConf);
+      executor = Executors.newFixedThreadPool(8,
+          new ThreadFactoryBuilder().setDaemon(true)
```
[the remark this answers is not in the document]
abstractdog: ack

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -362,13 +461,42 @@ private List<Event> createEventList(boolean sendSerializedEvents, InputSplitInfo
...
+          serializedPayloadSize += payloadSize;
+          diEvent = splitSerializer.write(count, mrSplit);
+        } else {
+          inMemoryPayloadSize += payloadSize;
+          diEvent = InputDataInformationEvent.createWithSerializedPayload(count, payloadBuffer);
+        }
         events.add(diEvent);
+        count += 1;
       }
+      splitSerializer.close();
```
deniskuzZ, line 496: probably should be in finally block

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
deniskuzZ: what I don't understand, is why can't we transfer over the wire incrementaly as well

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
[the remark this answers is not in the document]
deniskuzZ: if they are pooled, shouldn't you release it back to the pool?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: ack

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
deniskuzZ: you are throwing an exception inside of the executor. what is the strategy here, should we fail fast if any of jobs fail?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: it's written once per split generation, I thought it was more like an INFO, but I'm fine with having this DEBUG but keep the finish on INFO (which actually shows useful information)
```
      LOG.info(
          "Finished creating events ({} splits), size of payloads: in memory: {} bytes, serialized to fs: {} bytes",
          splits.size(), inMemoryPayloadSize, serializedPayloadSize);
```

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
deniskuzZ: have no idea about the default size, you have more experience here

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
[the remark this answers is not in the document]
abstractdog: right, those might be in the FileSystem CACHE or not, so we are supposed to call .close() on that without taking care of the hadoop caching details

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: hm, right, we're supposed to fail fast, but this way we're going to lose the exception thrown in the loop

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
abstractdog: okay, I'm setting it to 256MB, which is reasonable for every AM sizes between the typical 2-4GB range

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: regarding:
```
also, why do you need this method at all, since you pass the jobConf to HiveSplitGenerator constructor? pass conf with disabled serialization
```
because I was assuming this cleaner and more convenient, so the caller doesn't have to be aware of the config details

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: I ended up with:
```
      try (SplitSerializer splitSerializer = new SplitSerializer()) {
```
with the lazyInit() logic

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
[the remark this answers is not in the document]
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
deniskuzZ: maybe try
````
 Tasks.range(numTasks)
        .throwFailureWhenFinished(throwOnFailure)
        .executeWith(executor)
        .retry(3)
        .run(taskId -> {
        })
....

````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java (hunk 12 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java:210]: if you use auto-clousable, why don't we init in constructor, it doesn't seem expensive, however, code looks cleaner, WDYT?

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java (hunk 14 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java:481]: oh, ok, thanks for the clarification!

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java (hunk 12 above)
Comment (abstractdog) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java:210]: ack, let it be, I guess lazy initialization was also a workaround for an NPE in some unit test scenarios (maybe jobConf was null or some craziness), if I face it again, I'm going to fix

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java (hunk 7 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java:226]:
````
      for (final I item : items) {
        // submit a task for each item that will either run or abort the task
        futures.add(
            service.submit(
                new Runnable() {
                  @Override
                  public void run() {
                    if (!(stopOnFailure && taskFailed.get())) {
                      // run the task with retries
                      boolean threw = true;
                      try {
                        task.run(item);
                        succeeded.add(item);
                        threw = false;
                      } catch (Exception e) {
                        taskFailed.set(true);
                        exceptions.add(e);
                      } finally {
                        if (threw) {
                          taskFailed.set(true);
                        }
                      }

                    }
/ let the above tasks complete (or abort)
      exceptions.addAll(waitFor(futures));
      futures.clear();

      if (throwFailureWhenFinished && !exceptions.isEmpty()) {
        Tasks.throwOne(exceptions, exceptionClass);
      } else if (throwFailureWhenFinished && taskFailed.get()) {
        throw new RuntimeException("Task set failed with an uncaught throwable");
      }

      return !taskFailed.get();
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java (hunk 2 above)
Comment (abstractdog) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java:198]: ops I just realized InputInitializerContext is null when the split generation happens outside the AM, e.g.:
```
2024-04-05T00:57:35,207 ERROR [HiveServer2-Background-Pool: Thread-1193] tez.HiveSplitGenerator: An exception was caught while running split generation, this is not recoverable
java.lang.NullPointerException: null
	at org.apache.hadoop.hive.ql.exec.tez.HiveSplitGenerator$SplitSerializer.<init>(HiveSplitGenerator.java:193) ~[hive-exec-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.hadoop.hive.ql.exec.tez.HiveSplitGenerator.getSplitSerializer(HiveSplitGenerator.java:521) ~[hive-exec-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.hadoop.hive.ql.exec.tez.HiveSplitGenerator.createEventList(HiveSplitGenerator.java:475) ~[hive-exec-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.hadoop.hive.ql.exec.tez.HiveSplitGenerator.initialize(HiveSplitGenerator.java:425) ~[hive-exec-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.hadoop.hive.ql.udf.generic.GenericUDTFGetSplits.getSplits(GenericUDTFGetSplits.java:475) ~[hive-exec-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.hadoop.hive.ql.udf.generic.GenericUDTFGetSplits.getSplitResult(GenericUDTFGetSplits.java:254) ~[hive-exec-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.hadoop.hive.ql.udf.generic.GenericUDTFGetSplits2.process(GenericUDTFGetSplits2.java:78) ~[hive-exec-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
```
I'll handle it in the code

**Comment by abstractdog** [comment]
@deniskuzZ : addressed all comments in https://github.com/apache/hive/pull/5174/commits/cc0f6e2029f73c4226a874490be95ee4bc571ceb
unit tests added for the failure scenarios

File: common/src/java/org/apache/hadoop/hive/common/ExecutorUtils.java [hunk 17]
```diff
@@ -0,0 +1,101 @@
...
+          } catch (InterruptedException e) {
+            LOG.warn("Interrupted while getting future results", e);
+            for (Throwable t : uncaught) {
+              e.addSuppressed(t);
+            }
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+
+          } catch (CancellationException e) {
+            // ignore cancellations
```
deniskuzZ, line 65: shouldn't we interrupt the thread?

common/src/java/org/apache/hadoop/hive/common/ExecutorUtils.java
```diff
@@ -0,0 +1,101 @@
...
+   * This implementation is borrowed from iceberg's Tasks.
+   * @param futures
+   * @return
+   */
+  public static Collection<Throwable> waitFor(Collection<Future<?>> futures) {
+    while (true) {
+      int numFinished = 0;
+      for (Future<?> future : futures) {
+        if (future.isDone()) {
+          numFinished += 1;
```
deniskuzZ, line 45: on a 2nd iteration, you'll add same completed futures to the counter?

common/src/java/org/apache/hadoop/hive/common/ExecutorUtils.java
```diff
@@ -0,0 +1,101 @@
...
+public class ExecutorUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(ExecutorUtils.class);
+
+  /**
+   * Waiting for futures to finish and return any exceptions caught.
+   * This implementation is borrowed from iceberg's Tasks.
+   * @param futures
+   * @return
+   */
+  public static Collection<Throwable> waitFor(Collection<Future<?>> futures) {
```
deniskuzZ, line 40: that could be replaced with CompletableFuture API
````
            try {
              List<CompletableFuture<Void>> asyncTasks = new ArrayList<>();
              for (Runnable task : tasks) {
                CompletableFuture<Void> asyncTask = CompletableFuture.runAsync(
                                task, executor)
                        .exceptionally(t -> {
                          LOG.error("Error due to :", t);
                          return null;
                        });
                asyncTasks.add(asyncTask);
              }
              //Use get instead of join, so we can receive InterruptedException and shutdown gracefully
              CompletableFuture.allOf(asyncTasks.toArray(new CompletableFuture[0])).get();
              Thread.currentThread().interrupt();
              return;
            } catch (Throwable t) {
              LOG.error("Caught an exception while executing {}",
                        t.getMessage());
              throw t;
            }
````

File: common/src/java/org/apache/hadoop/hive/conf/HiveConf.java [hunk 20]
```diff
@@ -4756,6 +4756,13 @@ public static enum ConfVars {
         "Class to use for calculating available slots during split generation"),
     HIVE_TEZ_GENERATE_CONSISTENT_SPLITS("hive.tez.input.generate.consistent.splits", true,
         "Whether to generate consistent split locations when generating splits in the AM"),
+    HIVE_TEZ_SPLIT_FS_SERIALIZATION_THRESHOLD("hive.tez.split.fs.serialization.threshold", 268435456,
```
deniskuzZ, line 4759: should it be `hive.tez.input.fs.serialization.threshold` similar to HIVE_TEZ_GENERATE_CONSISTENT_SPLITS("hive.tez.input.generate.consistent.splits"?

abstractdog: make sense, let me change

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
abstractdog: CompletableFuture API looks cleaner and it works, implemented in:
https://github.com/apache/hive/pull/5174/commits/54e95ec655fbcc58f67dbd3beed78624820a19d7

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +173,107 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
...
+      try {
+        CompletableFuture.allOf(asyncTasks.toArray(new CompletableFuture[0])).get();
+      } catch (InterruptedException e) {
+        LOG.info("Interrupted while generating splits", e);
+        Thread.currentThread().interrupt();
+      } catch (ExecutionException e) {// ExecutionException wraps the original exception
+        LOG.error("Exception while generating splits", e.getCause());
+        throw new RuntimeException(e.getCause());
+      } finally {
+        if (executor == null) {
```
deniskuzZ, line 263: != null

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
```diff
@@ -154,6 +173,107 @@ private void prepare(InputInitializerContext initializerContext) throws IOExcept
     LOG.info("SplitLocationProvider: " + splitLocationProvider);
   }
 
+  /**
+   * SplitSerializer is a helper class for taking care of serializing splits to the tez scratch dir
+   * when a size criteria defined by "hive.tez.input.fs.serialization.threshold" is met.
+   * It utilizes an ExecutorService for parallel writes to prevent a single split write operation
+   * becoming the bottleneck (as write() is called from a loop currently).
+   */
+  class SplitSerializer implements AutoCloseable {
```
deniskuzZ, line 182: private static final

abstractdog: this could not be either private or static or final
1. testing it from TestHiveSplitGenerator (private)
2. jobConf cannot be accessed from static context (static)
3. it's inherited in unit test (final)

deniskuzZ:
@abstractdog, are we still waiting for the Tez changes to land?

abstractdog:
> @abstractdog, are we still waiting for the Tez changes to land?

yes, unfortunately, we cannot merge this until tez 0.10.4 is released, that's why I set it blocked by:
https://issues.apache.org/jira/browse/TEZ-4556

abstractdog:
reopened as https://github.com/apache/hive/pull/5449

PR #5175, opened 2024-04-01
HIVE-28167: Full table deletion fails when converting to truncate for Iceberg and ACID tables

SourabhBadhya:

HIVE-28167: Full table deletion fails when converting to truncate for Iceberg and ACID tables

Full table deletion fails when converting to truncate for Iceberg and ACID tables

No

No

QTest


deniskuzZ (approved):
LGTM +1

PR #5177, opened 2024-04-02
HIVE-28171 Active queries API does not provide query information if they are still pending

ramesh0201:


ayushtkn (commented):
Thanx @ramesh0201 for the fix. Can we extend a test as well for this? maybe in TestSqlOperation or so?

PR #5179, opened 2024-04-03
HIVE-28172: Iceberg: Truncate should work for V1 tables with partition evolution.

ayushtkn:

Truncate on v1 Iceberg tables should be supported 

Better usability

Yes, Truncate on v1 tables with partition evolution can work

No

UT


SourabhBadhya (approved):
LGTM +1

PR #5180, opened 2024-04-03
HIVE-28173: Fixed staging dir issues on materialized views on HDFS en…

scarlin-cloudera:
…crypted tables

There are 2 issues involving staging directories on hdfs encrypted tables

1) The staging directory is created at compile time.  For non hdfs encrypted tables, the "mkdir" flag is set to false.
There is no such flag for hdfs encrypted tables.

2) The "FileSystem.deleteOnFileExit()" method is not called from the HiveMaterializedViewRegistry thread.

For the first issue, the fix has only been implemented for the materialized view registry thread. It is a bit more risky to implement this for all compiled queries in the SemanticAnalyzer, but perhaps it is correct. However, the implemented fix in this commit is similar to previous implemented logic and a customer has this specific problem on their production machines so this fix is important.

The second issue is also specifically fixed for the materialized view registry thread. A small refactoring was done since the "isLoadingMaterializedView" flag should not be set within ParseUtils

Mentioned above

Fixes an issue in a production environment.  Staging directories are being left behind after a materialized view update on an hdfs encrypted table

No

No

Patch was tested in my local environment. Existing materialized view tests should test some of the changes. However, a huge framework change would have to be created to test things are working properly for the materialized view registry test. Given that this is a small change and given that this is a problem on a production environment, I think it's a bit hard to implement this kind of testing in a customer expected amount of time.

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/parse/TestParseUtilsStagingDir.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/parse/TestParseUtilsStagingDir.java
@@ -1,0 +1,150 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


deniskuzZ (commented):
LGTM, @scarlin-cloudera could we add tests for MV?

scarlin-cloudera:
> LGTM, @scarlin-cloudera could we add tests for MV? Maybe @kasakrisz can suggest if there are some similar tests that. we could extend

Thanks for the review! 

Yeah, I'm looking to add a unit test.  I've been trying to add for MV, but the real change involves the MV daemon thread and there are no current integrated or unit tests for that. 

Having said that, it looks like there might be something I can add and I'll hopefully be adding that within the next day or two.

kasakrisz (approved):

deniskuzZ (approved):
+1

PR #5182, opened 2024-04-05
HIVE-28143: After HIVE-27492 fix, some HPLSQL built-in functions like trim, lower are not working when used in insert statement.

mdayakar:
HIVE-28143: After HIVE-27492 fix, some HPLSQL built-in functions like trim, lower are not working when used in insert statement.

As a part of  HIVE-27492 issue fix, removed the UDFs which are present in Hive as it may become duplicate but HPLSQL can support local and offline mode which requires the HPLSQL supported UDFs must present in HPLSQL itself so added them back and also fixed issues related to those UDFs.

To fix HIVE-28143 as all UDFs which are supported by HPLSQL are not working when they used other select statements.

No

No

_mvn test -Dtest=TestHplsqlOffline
mvn  test -Dtest=TestHplsqlLocal 
mvn test -Dtest=TestHplSqlViaBeeLine -pl itests/hive-unit -Pitests_


hplsql/src/main/java/org/apache/hive/hplsql/Expression.java
```diff
@@ -565,7 +575,11 @@ public void operatorConcatSql(HplsqlParser.Expr_concatContext ctx) {
     sql.append("CONCAT(");
     int cnt = ctx.expr_concat_item().size();
     for (int i = 0; i < cnt; i++) {
-      sql.append(evalPop(ctx.expr_concat_item(i)).toString());
+      String concatStr = evalPop(ctx.expr_concat_item(i)).toString();
+      if (!concatStr.startsWith("'")) {
```
kasakrisz, line 579: AFAIK `concat` expects string parameters. The actual parameters can be a
* string literals which needs quotation 
* or an string typed expressions which should not be quoted
Could you please share an example when quotation is missing but required at this point?

hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
```diff
@@ -787,8 +787,19 @@ public Integer insertValues(HplsqlParser.Insert_stmtContext ctx) {
...
+      for (int j = 0; j < cols; j++) {
+        Var var = evalPop(row.expr(j));
+        String value =  null;
+        if (var.type == Type.TIMESTAMP) {
+          value = String.format("TIMESTAMP '%s'", var.toString());
+        } else if (var.type == Type.DATE) {
+          value = String.format("DATE '%s'", var.toString());
+        } else {
+          value = var.toSqlString();
+        }
```
kasakrisz, line 799: How about `var.toString()`?

hplsql/src/main/java/org/apache/hive/hplsql/Var.java
```diff
@@ -601,7 +602,7 @@ else if (type == Type.STRING) {
       return (String)value;
     }
     else if (type == Type.DATE) {
-      return String.format("DATE '%s'", value);
+      return ((Date)value).toString();
```
kasakrisz, line 605: Is it safe to remove the `DATE` keyword? Let's say we have `2024-04-09`. Is it a string or a date literal?

hplsql/src/main/java/org/apache/hive/hplsql/Var.java
```diff
@@ -612,7 +613,7 @@ else if (type == Type.TIMESTAMP) {
       if (t.length() > len) {
         t = t.substring(0, len);
       }
-      return String.format("TIMESTAMP '%s'", t);
+      return t;
```
kasakrisz, line 616: Is it safe to remove the `TIMESTAMP`?
A few lines above a comment says the value can be `.0`. Is it a float or a timestamp?

https://github.com/apache/hive/blob/be857bf83c00e41a315a7cc8e013832cf169ef18/hplsql/src/main/java/org/apache/hive/hplsql/Var.java#L608C54-L608C56

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionDatetime.java
```diff
@@ -162,6 +174,35 @@ void toTimestamp(HplsqlParser.Expr_func_paramsContext ctx) {
...
+    if (value.type != Var.Type.BIGINT) {
+      Var newVar = new Var(Var.Type.BIGINT);
+      value = newVar.cast(value);
+    }
+    long epoch = value.longValue();
+    String format = "yyyy-MM-dd HH:mm:ss";
+    if (cnt > 1) {
+      format = Utils.unquoteString(evalPop(ctx.func_param(1).expr()).toString());
+    }
+    evalString(Utils.quoteString(new SimpleDateFormat(format).format(new Date(epoch * 1000))));
```
kasakrisz, line 196: Can the newer Java Date Time API be used? `java.time.*`

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionMisc.java
```diff
@@ -59,6 +64,43 @@ public void register(BuiltinFunctions f) {
...
   }
+
+  /**
+   * CAST function
+   */
+  void cast(HplsqlParser.Expr_spec_funcContext ctx) {
+    if (ctx.expr().size() != 1) {
+      evalNull();
+      return;
+    }
```
kasakrisz, line 75: Does it mean that the user specified wrong parameters? Should we throw exception?

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionMisc.java
```diff
@@ -113,7 +155,9 @@ void currentUser(HplsqlParser.Expr_spec_funcContext ctx) {
   }
   
   public static Var currentUser() {
-    return new Var("CURRENT_USER()");
+    String currentUser = System.getProperty("user.name");
+    currentUser = Utils.quoteString(currentUser);
+    return new Var(currentUser);
```
kasakrisz, line 160: Why do we need quotes here? The variable has string type:
https://github.com/apache/hive/blob/be857bf83c00e41a315a7cc8e013832cf169ef18/hplsql/src/main/java/org/apache/hive/hplsql/Var.java#L77-L80

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionMisc.java
```diff
@@ -211,6 +269,16 @@ public void partCount(HplsqlParser.Expr_spec_funcContext ctx) {
     query.close();
   }
 
+  public void modulo(HplsqlParser.Expr_func_paramsContext ctx) {
+    if (ctx.func_param().size() == 2) {
+      int a = evalPop(ctx.func_param(0).expr()).intValue();
+      int b = evalPop(ctx.func_param(1).expr()).intValue();
+      evalInt(a % b);
+    } else {
+      evalNull();
```
kasakrisz, line 278: Should exception be thrown?

hplsql/src/main/antlr4/org/apache/hive/hplsql/Hplsql.g4
```diff
@@ -1563,7 +1563,7 @@ non_reserved_words :                      // Tokens that are not reserved words
      | T_TO     
      | T_TOP
      | T_TRANSACTION
-     | T_TRIM
+     // | T_TRIM
```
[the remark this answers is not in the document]
mdayakar: No. Now uncommented it.

hplsql/src/main/java/org/apache/hive/hplsql/Expression.java
```diff
@@ -110,7 +110,17 @@ else if (ctx.interval_item() != null) {
...
+      if (value.type == Type.NULL && sql.toString().length() == 0) {
+        exec.stackPush(new Var());
+        return;
+      } else if (exec.buildSql && value.type == Type.DATE) {
+        sql.append(String.format("DATE '%s'", value.toString()));
+      } else if (exec.buildSql && value.type == Type.TIMESTAMP) {
+        sql.append(String.format("TIMESTAMP '%s'", value.toString()));
+      } else {
+        sql.append(value.toString());
+      }
```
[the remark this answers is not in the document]
mdayakar: As a part of HIVE-27492 fix, added these format in Var.toString() API itself but this format is required only if the UDF used in as a select statement (DATE '2024-04-18' is valid where as 2024-04-18 is not a valid value in selection) so changed accordingly(now added only for select flow).

hplsql/src/main/java/org/apache/hive/hplsql/Expression.java
mdayakar: For example below example
`SELECT '#' || TRIM(' Hello ') || '#';`
Here HPLSQL internally calls concat and when `TRIM(' Hello ')` is evaluated it will become `Hello` without quotes so here concat will throw error. To avoid it if the concat item is not starting with single quote then enclosing it with single quotes.

hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
mdayakar: As explained at https://github.com/apache/hive/pull/5182#discussion_r1570337451 comment, this format is required for insert also otherwise it throws exception. Example: insert into test values('Bob', 2024-04-17 10:20:30) throws exception here even the the type of second column is timestamp.

hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
```diff
@@ -787,8 +787,19 @@ public Integer insertValues(HplsqlParser.Insert_stmtContext ctx) {
...
+        if (value.startsWith("''")) {
```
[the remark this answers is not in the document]
mdayakar: if the `value` has already a quote then `var.toSqlString();` add another one so remove that extra quote added this check. Anyway I changed the code to add the quote if it is not present.

hplsql/src/main/java/org/apache/hive/hplsql/Var.java
mdayakar: As explained in https://github.com/apache/hive/pull/5182#discussion_r1570337451 comment, this was added by me only just reverted back the change done as a part of [HIVE-27492](https://issues.apache.org/jira/browse/HIVE-27492) fix.

hplsql/src/main/java/org/apache/hive/hplsql/Var.java
mdayakar: As explained in https://github.com/apache/hive/pull/5182#discussion_r1570337451 comment, this was added by me only just reverted back the change done as a part of [HIVE-27492](https://issues.apache.org/jira/browse/HIVE-27492) fix.

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionDatetime.java
mdayakar: These are the APIs removed as a part of [HIVE-27492](https://issues.apache.org/jira/browse/HIVE-27492) fix. So just added back them again.Didn't do any code changes.

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionMisc.java
mdayakar: These are the APIs removed as a part of [HIVE-27492](https://issues.apache.org/jira/browse/HIVE-27492) fix. So just added back them again.Didn't do any code changes.

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionMisc.java
mdayakar: if the UDF is getting called as a part of select then without quotes it will throw error. Ex: `SELECT Bob ` where as `SELECT 'Bob'` is. a valid statement. So adding single quote.

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionMisc.java
mdayakar: These are the APIs removed as a part of [HIVE-27492](https://issues.apache.org/jira/browse/HIVE-27492) fix. So just added back them again.Didn't do any code changes.

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionMisc.java
kasakrisz: Sorry, I'm missing something. IIUC instances of `Var` are the internal representations of an SQL literal. In case of string the single quote characters in the beginning and the end should be removed at parse time and also single quote characters part of the text should be unescaped.
If the original SQL form of the literal is needed there is  https://github.com/apache/hive/blob/1a969f6642d4081027040f97e6a689bf2e687db3/hplsql/src/main/java/org/apache/hive/hplsql/Var.java#L620-L631

hplsql/src/main/java/org/apache/hive/hplsql/Expression.java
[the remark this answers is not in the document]
mdayakar: Added Var.toSqlString(boolean isBuildSql) API and using the same.

hplsql/src/main/java/org/apache/hive/hplsql/Expression.java
mdayakar: Fixed

hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
mdayakar: Added Var.toSqlString(boolean isBuildSql) API and using the same.

hplsql/src/main/java/org/apache/hive/hplsql/functions/FunctionMisc.java
mdayakar: As a part [HIVE-27492](https://issues.apache.org/jira/browse/HIVE-27492) fix, we fixed some issues like UDF using in select statement, that fix should be present so not reverting back entire [HIVE-27492](https://issues.apache.org/jira/browse/HIVE-27492) fix only the removed UDF cde is getting added and corrected for the functionality.

kasakrisz (changes requested):

kasakrisz (approved):

PR #5186, opened 2024-04-09
HIVE-28190: Fix heartbeatLockMaterializationRebuild

InvisibleProgrammer:
It fails with the following error: 

```
org.springframework.dao.InvalidDataAccessApiUsageException: SQL [UPDATE "MATERIALIZATION_REBUILD_LOCKS" SET "MRL_LAST_HEARTBEAT" = 1712571919559 WHERE "MRL_TXN_ID" = 2297 AND "MRL_DB_NAME" = ? AND "MRL_TBL_NAME" = ?]: given 2 parameters but expected 0 
```

We didn't spot it so far as when the heartbeat of materialized view fails with an error, it doesn't affect the rebuild query run. So that it can be only spotted by actively watching the logs.

Fix the query parameterisation

Because it is wrong

No

No

Manually, by modifying an existing qtest materialized_view_create_rewrite_6.q and checking the logs.


InvisibleProgrammer:
@kasakrisz , what do you think, what should be the expected behaviour when the heartbeat fails: 
- should it cancel the query and Hive give an exception, so that the user will know about the issue (I would prefer that way)
- or we can still ignore it. In that case, only a pure luck can help spotting that kind of issues.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
```diff
@@ -705,12 +706,11 @@ public LockResponse lockMaterializationRebuild(String dbName, String tableName,
   public boolean heartbeatLockMaterializationRebuild(String dbName, String tableName, long txnId) throws MetaException {
     int result = jdbcResource.execute(
         "UPDATE \"MATERIALIZATION_REBUILD_LOCKS\"" +
-            " SET \"MRL_LAST_HEARTBEAT\" = " + Instant.now().toEpochMilli() +
-            " WHERE \"MRL_TXN_ID\" = " + txnId +
-            " AND \"MRL_DB_NAME\" = ?" +
-            " AND \"MRL_TBL_NAME\" = ?",
+            " SET \"MRL_LAST_HEARTBEAT\" = " + getEpochFn(jdbcResource.getDatabaseProduct()) +
```
aturoczy, line 709: Just for curiasity what the difference between the Intant.now and the getEpochFn?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/HeartbeatTxnRangeFunction.java
```diff
@@ -82,7 +82,7 @@ public HeartbeatTxnRangeResponse execute(MultiDataSourceJdbcResource jdbcResourc
     }
     if (updateCnt == numTxnsToHeartbeat) {
       //fast pass worked, i.e. all txns we were asked to heartbeat were Open as expected
-      context.rollbackToSavepoint(savePoint);
+      context.createSavepoint();
```
aturoczy, line 85: Dont you need to roll it back?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
InvisibleProgrammer: It provides different methods to get the timestamp, based on the different database providers. 
The original code, `Instant.now()` generates the timestamp in Java side. This method says the db engine will handle that.

```java
      new EnumMap<DatabaseProduct.DbType, String>(DatabaseProduct.DbType.class) {{
        put(DbType.DERBY, "{ fn timestampdiff(sql_tsi_frac_second, timestamp('" + new Timestamp(0) +
            "'), current_timestamp) } / 1000000");
        put(DbType.CUSTOM, "{ fn timestampdiff(sql_tsi_frac_second, timestamp('" + new Timestamp(0) +
            "'), current_timestamp) } / 1000000");
        put(DbType.MYSQL, "round(unix_timestamp(now(3)) * 1000)");
        put(DbType.POSTGRES, "round(extract(epoch from current_timestamp) * 1000)");
        put(DbType.ORACLE, "(cast(systimestamp at time zone 'UTC' as date) - date '1970-01-01')*24*60*60*1000 " +
            "+ cast(mod( extract( second from systimestamp ), 1 ) * 1000 as int)");
        put(DbType.SQLSERVER, "datediff_big(millisecond, '19700101', sysutcdatetime())");
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/HeartbeatTxnRangeFunction.java
deniskuzZ: savepoint is kinda commit, same thing is done after refactor in heartbeatTxn/Locks

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
deniskuzZ: since we have `LockMaterializationRebuildFunction` and separate functions for heartbeat `HeartbeatLockFunction`, `HeartbeatTxnFunction`, maybe we should extract this code into a function as well

kasakrisz:
> @kasakrisz , what do you think, what should be the expected behaviour when the heartbeat fails:
> 
> * should it cancel the query and Hive give an exception, so that the user will know about the issue (I would prefer that way)
> * or we can still ignore it. In that case, only a pure luck can help spotting that kind of issues.


IMHO It should work like in case of other DML statements (`INSERT SELECT`, `INSERT OVERWRITE SELECT`) and it can be a separate jira/pr

deniskuzZ:
please don't merge without the tests

deniskuzZ (changes requested):
please don't merge without the tests

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
zratkai, line 709: Instead of concatenating the timestamp, I would prefer to use it as parametrised value like txnId, dbName.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/HeartbeatTxnRangeFunction.java
InvisibleProgrammer: FYI: I removed using `getEpochFn(jdbcResource.getDatabaseProduct())` from this change and switched back to the original way of calculating the timestamp, for two reasons: 
- jdbc paramters are typed. And the framework knows the type of the columns. And I cannot use `getEpochFn(jdbcResource.getDatabaseProduct())` in a parmeterized query as for that column, it expects a bigint value (long in java side). And it doesn't run at all. The possible way to introduce it would be to use it as string, not a paramterized query. 
- the other reason is that I checked other places when we use that column and I wanted to keep the code consistent: Change it at all places or not change at all. 

I would consider doing it at an other, refactor ticket that I mentioned in my previous comment.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
InvisibleProgrammer: I reverted that part to use a parameterzied query.

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
```diff
@@ -0,0 +1,114 @@
...
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class TestHeartbeatTxnRangeFunction {
+  static final private String CLASS_NAME = TxnHandler.class.getName();
+
+  private final HiveConf conf = new HiveConf();
+  private TxnStore txnHandler;
+
+  public TestHeartbeatTxnRangeFunction() throws Exception {
```
deniskuzZ, line 52: it shouldn't be in `txn` package, not `materializedview`

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
```diff
@@ -0,0 +1,114 @@
...
+  public void testHeartbeatTxnRangeFunction_NoSuchTxn() throws MetaException {
+    HeartbeatTxnRangeRequest request = new HeartbeatTxnRangeRequest(1L, 1L);
+
+    HeartbeatTxnRangeResponse response = txnHandler.heartbeatTxnRange(request);
+
+    assertEquals(1, response.getNosuchSize());
+  }
+
+  @Test
+  public void testHeartbeatTxnRangeFunction_AbortedTxn() throws MetaException, TxnAbortedException, NoSuchTxnException {
```
deniskuzZ, line 82: TxnHandler already has the following tests, why we duplicate them, if you need extra check just extend
````
heartbeatTxnRangeOneAborted
heartbeatTxnRangeOneCommitted
heartbeatTxnRange
````

File: ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java [hunk 5]
```diff
@@ -0,0 +1,114 @@
...
+
+  @Test
+  public void testHeartbeatTxnRangeFunction_Success() throws Exception {
+    long txnId = openTxn();
+
+    String firstHeartbeat = TestTxnDbUtil.queryToString(conf, "select \"TXN_LAST_HEARTBEAT\" from \"TXNS\" where \"TXN_ID\" = " + txnId, false);
+    HeartbeatTxnRangeRequest request = new HeartbeatTxnRangeRequest(txnId, txnId);
+
+    HeartbeatTxnRangeResponse response = txnHandler.heartbeatTxnRange(request);
+    String updatedHeartbeat = TestTxnDbUtil.queryToString(conf, "select \"TXN_LAST_HEARTBEAT\" from \"TXNS\" where \"TXN_ID\" = " + txnId, false);
```
deniskuzZ, line 100: what is the purpose of `updatedHeartbeat` query, what does it test?

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.Test;
+
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class TestHeartbeatTxnRangeFunction {
+  static final private String CLASS_NAME = TxnHandler.class.getName();
```
deniskuzZ, line 47: copy-paste

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
```diff
@@ -0,0 +1,114 @@
...
+  private TxnStore txnHandler;
+
+  public TestHeartbeatTxnRangeFunction() throws Exception {
+    TestTxnDbUtil.setConfValues(conf);
+    TestTxnDbUtil.prepDb(conf);
+    LoggerContext ctx = (LoggerContext) LogManager.getContext(false);
+    Configuration conf = ctx.getConfiguration();
+    conf.getLoggerConfig(CLASS_NAME).setLevel(Level.DEBUG);
+    ctx.updateLoggers(conf);
+    tearDown();
```
deniskuzZ, line 59: why do we call `cleanDb` right after the `prepDb`?

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @Before
+  public void setUp() throws Exception {
+    txnHandler = TxnUtils.getTxnStore(conf);
```
deniskuzZ, line 64: why do you need a new instance of txnHandler for every test?

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
InvisibleProgrammer: It is a good question. I got the inspiration from `TestTxnHandler`. TBH, I had the exact same question. But decided to keep it here assuming there is something that I should know about. 
See: https://github.com/apache/hive/blob/6b48995442a514cd946663b2c0343e772430761e/ql/src/test/org/apache/hadoop/hive/metastore/txn/TestTxnHandler.java#L127

Anyway, let me experiment with it. I want to see what happens if I remove it.

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
InvisibleProgrammer: It is a good question. I got the inspiration from TestTxnHandler. TBH, I had the exact same question. But decided to keep it here assuming there is something that I should know about.

See: https://github.com/apache/hive/blob/6b48995442a514cd946663b2c0343e772430761e/ql/src/test/org/apache/hadoop/hive/metastore/txn/TestTxnHandler.java#L1860

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
InvisibleProgrammer: Thank you for spotting it. I will fix it.

InvisibleProgrammer: Hi, at the end, I removed it completely. The new class doesn't require that part.

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
InvisibleProgrammer: I tested and we don't need that. Removed

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
InvisibleProgrammer: I checked it. Our tests can run with static txnHandler so I refactored it. 
TBH, my first thought was that it can lead to flaky test. Usually, sharing any kind of state between tests is a bad pattern as test won't be isolated.

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
InvisibleProgrammer: I removed/embedded them those tests into TestHeartbeatTxnRangeFunction.

deniskuzZ (approved):
LGTM +1

ql/src/test/org/apache/hadoop/hive/ql/materializedview/TestHeartbeatTxnRangeFunction.java
deniskuzZ: txnHandler is just a client here

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
deniskuzZ: HIVE-6965: Transaction manager should use RDBMS time instead of machine time. Determine the current time, using the RDBMS as a source of truth.

However, since it's only used for MV rebuild that's probably ok

InvisibleProgrammer:
@deniskuzZ , thank you for your help.

PR #5188, opened 2024-04-10
HIVE-28192 Iceberg: Fix thread safety issue with PositionDeleteInfo i…

simhadri-g:
…n IOContext

### What changes were proposed in this pull request?

Make PositionDeleteInfo in IOContext Thread Safe

### Why are the changes needed?

Assume:
Query in thread 1:  UPDATE customers SET first_name='Changed' WHERE  last_name='Henderson' OR last_name='Johnson'

Query in thread 2:  UPDATE customers SET first_name='Changed' WHERE  last_name='Taylor' AND customer_id=1

When there are concurrent update queries, PositionDeleteInfo is not thread safe, as a result the delete file of thread 1 end up with pointing to the delete record of query in thread 2  leading to a corrupted delete file. 

In this example, The delete file for "henderson" in thread 1 is pointing to the record "taylor"   in thread 2. As a result we are ending up with 2 entries for henderson
```
   ~  $HIVE_HOME/bin/hive --orcfiledump -d -p  /Users/simhadri.govindappa/Documents/apache/hive/iceberg/iceberg-handler/target/tmp/hive3325124320669640146/external/customers/data/last_name=Henderson/customer_id_bucket=3/00000-0-delete-simhadri.govindappa_20240328120113_3f3c2165-9f33-4238-bc22-9095399c3b9e-job_17116524743461_0001-4-00002.orc

Processing data file /Users/simhadri.govindappa/Documents/apache/hive/iceberg/iceberg-handler/target/tmp/hive3325124320669640146/external/customers/data/last_name=Henderson/customer_id_bucket=3/00000-0-delete-simhadri.govindappa_20240328120113_3f3c2165-9f33-4238-bc22-9095399c3b9e-job_17116524743461_0001-4-00002.orc [length: 1300]
{"file_path":"file:/Users/simhadri.govindappa/Documents/apache/hive/iceberg/iceberg-handler/target/tmp/hive3325124320669640146/external/customers/data/last_name=Taylor/customer_id_bucket=4/00000-0-data-simhadri.govindappa_20240328120056_632c8bb7-b3af-42af-a931-2c1bd9094414-job_17116524611241_0001-1-00008.orc","pos":0}
```

writer.HiveIcebergDeleteWriter PositionDeleteInfo
```
[...]

PR #5189, opened 2024-04-10
HIVE-28193: CBO fallback can overwrite the real error message when EXCEPT or INTERSECT are used

okumin:

Ignore `hive.cbo.fallback.strategy` if EXCEPT or INTERSECT are used.

I have considered some alternatives.
- Always show the error message of the first attempt with CBO
    - This option is not trivial because the re-execution plugin handles not only compile errors but also runtime errors such as OOM
- Make `CBOFallbackStrategy#isFatal` configurable, e.g. accept regular expressions to test error messages
    - It sounds hard to maintain fatal error messages so that it is consistent with the actual message in an Exception
- Add a new field like `isRetryableWithCBOFallback` to SemanticException
    - This could be an option. I prefer the approach of this PR because SemanticException is very common and I don't want to add an ad-hoc attribute. The fallback mode is going to be deprecated in the future

As explained in HIVE-28193, the fallback attempt can fail with an inconsistent reason when the operators are used.

Yes. Users would see a different but better error message when they hit the issue.

No.

I added an integration test.


okumin:
CI failed but I feel the test case is failing also on the master branch

simhadri-g:
Yes, I have raised a PR: https://github.com/apache/hive/pull/5188 to fix failing test

okumin:
I found QUALIFY also shares the same problem. I will address it together.

okumin:
I created a new PR.
https://github.com/apache/hive/pull/5312

PR #5191, opened 2024-04-11
HIVE-28196: Preserve column stats when applying UDF upper/lower

ngsg:

This patch preserves column stats when applying UDF `upper`/`lower` on a column.

Current Hive re-estimates column stats when it encounters UDF. UDF `upper` and `lower` do not change column stats, but current Hive resets `avgColLen` to `hive.stats.max.variable.length`. Keeping the input data's column stats, including `avgColLen`,  will give more accurate stats to later query optimizers and make a better query plan.

No

No

Added a qfile for testing column stats computation on upper/lower UDF. Current Hive chooses MergeJoin due to column stats over-estimation, while this patch chooses MapJoin.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
@@ -82,6 +82,8 @@ import org.apache.hadoop.hive.ql.stats.estimator.StatEstimatorProvider;
 import org.apache.hadoop.hive.ql.udf.generic.GenericUDAFSum;
 import org.apache.hadoop.hive.ql.udf.generic.GenericUDF;
 import org.apache.hadoop.hive.ql.udf.generic.GenericUDFBridge;
+import org.apache.hadoop.hive.ql.udf.generic.GenericUDFLower;
+import org.apache.hadoop.hive.ql.udf.generic.GenericUDFUpper;
 import org.apache.hadoop.hive.ql.udf.generic.NDV;
 import org.apache.hadoop.hive.ql.util.JavaDataModel;
 import org.apache.hadoop.hive.ql.util.NamedForkJoinWorkerThreadFactory;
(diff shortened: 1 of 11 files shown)


ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
```diff
@@ -1736,6 +1749,13 @@ private static boolean isWideningCast(ExprNodeGenericFuncDesc engfd) {
             engfd.getTypeInfo());
   }
 
+  private static boolean isColStatsReusableFunc(ExprNodeGenericFuncDesc engfd) {
+    GenericUDF udf = engfd.getGenericUDF();
+    Class udfClass =
+        (udf instanceof GenericUDFBridge) ? ((GenericUDFBridge)udf).getUdfClass() : udf.getClass();
+    return udfClass == GenericUDFLower.class || udfClass == GenericUDFUpper.class;
```
okumin, line 1756: I like this PR. I have also observed that an estimation was wrong, and Map Join was disabled.
We implement some estimation APIs on the UDF side. I'll share examples with you. This is not a strong suggestion, and the current implementation seems to be a good starting point. We may finally need a pluggable I/F to empower custom UDFs more.
- https://github.com/apache/hive/blob/rel/release-4.0.0/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/NDV.java
- https://github.com/apache/hive/blob/rel/release-4.0.0/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFEvaluator.java#L63-L81

okumin (commented):
Looks good, but we have to update some `.q.out` of failing test cases.

ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
ngsg: Thanks for sharing the examples. I moved the estimation logic to UDF side by using `StatEstimatorProvider` interface.

okumin: I didn't know the interface and looks super nice!

okumin (approved):
`mv_iceberg_orc5` failed, but the PR looks mostly good to me

deniskuzZ (approved):
LGTM +1, pending tests
note: mv_iceberg_orc5 fails in master as well, looks like flaky

zhangbutao:
> LGTM +1, pending tests note: mv_iceberg_orc5 fails in master as well, looks like flaky

`mv_iceberg_orc5.q` is flaky. Please see https://issues.apache.org/jira/browse/HIVE-28203

PR #5195, opened 2024-04-15
HIVE-28202: Incorrect projected column size after ORC upgrade to v1.6.7

deniskuzZ:

Fixed projected column size logic

Incorrect projected column size leads to wrong split estimates and task overprovision.

No

No

qtests + TestInputOutputFormat#testProjectedColumnSize

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java
@@ -1727,31 +1727,21 @@ public class OrcInputFormat implements InputFormat<NullWritable, OrcStruct>,
     private long computeProjectionSize(List<OrcProto.Type> fileTypes,
         List<OrcProto.ColumnStatistics> stats, boolean[] fileIncluded) throws FileFormatException {
       List<Integer> internalColIds = Lists.newArrayList();
+      int rootColumn = 0;
       if (fileIncluded == null) {
         // Add all.
-        for (int i = 0; i < fileTypes.size(); i++) {
+        for (int i = rootColumn + 1; i < fileTypes.size(); i++) {
           internalColIds.add(i);
         }
       } else {
-        for (int i = 0; i < fileIncluded.length; i++) {
-          if (fileIncluded[i]) {
+        for (int i = rootColumn + 1; i < fileIncluded.length; i++) {
+          if (fileIncluded[i] && (isOriginal || i != OrcRecordUpdater.ROW + 1)) {
             internalColIds.add(i);
           }
         }
       }
       return ReaderImpl.getRawDataSizeFromColIndices(internalColIds, fileTypes, stats);
     }
-  }
-
-  public static boolean[] shiftReaderIncludedForAcid(boolean[] included) {
-    // We always need the base row
-    included[0] = true;
-    boolean[] newIncluded = new boolean[included.length + OrcRecordUpdater.FIELDS];
-    Arrays.fill(newIncluded, 0, OrcRecordUpdater.FIELDS, true);
-    for (int i = 0; i < included.length; ++i) {
-      newIncluded[i + OrcRecordUpdater.FIELDS] = included[i];
-    }
-    return newIncluded;
   }
 
   /** Class intended to update two values from methods... Java-related cruft. */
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java
```diff
@@ -1727,14 +1727,15 @@ private void populateAndCacheStripeDetails() throws IOException {
...
         // Add all.
-        for (int i = 0; i < fileTypes.size(); i++) {
+        for (int i = rootColumn + 1; i < fileTypes.size(); i++) {
           internalColIds.add(i);
         }
       } else {
-        for (int i = 0; i < fileIncluded.length; i++) {
-          if (fileIncluded[i]) {
+        for (int i = rootColumn + 1; i < fileIncluded.length; i++) {
+          if (fileIncluded[i] && (isOriginal || i != OrcRecordUpdater.ROW + 1)) {
```
abstractdog, line 1738: this condition is very vague and hard to understand for an average code reader, please include comments for every parts:
- isOriginal
- OrcRecordUpdater.ROW + 1

ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java
```diff
@@ -1727,14 +1727,15 @@ private void populateAndCacheStripeDetails() throws IOException {
     private long computeProjectionSize(List<OrcProto.Type> fileTypes,
         List<OrcProto.ColumnStatistics> stats, boolean[] fileIncluded) throws FileFormatException {
       List<Integer> internalColIds = Lists.newArrayList();
+      int rootColumn = 0;
```
abstractdog, line 1730: rootColumn doesn't seem to change, which is strange, I guess this is more like a constant instead

deniskuzZ: it's not supposed to, it's an index of the root struct element that should be skipped

ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java
deniskuzZ: added. new test

ql/src/test/org/apache/hadoop/hive/ql/io/orc/TestInputOutputFormat.java
```diff
@@ -1837,6 +1836,26 @@ public void testProjectedColumnSize() throws Exception {
...
+    OrcInputFormat.SplitGenerator splitter = new OrcInputFormat.SplitGenerator(splitInfo, null, true, true);
+    
+    List<OrcSplit> results = splitter.call();
+    assertEquals(1, results.size());
+    
+    OrcSplit result = results.get(0);
+    assertEquals(3, result.getStart());
+    assertEquals(900, result.getLength());
+    assertEquals(1, results.size());
+    assertEquals(1246255309, results.get(0).getProjectedColumnsUncompressedSize());
```
abstractdog, line 1856: nit: reuse result instead of results.get(0)

ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java
```diff
@@ -1725,35 +1725,22 @@ private void populateAndCacheStripeDetails() throws IOException {
...
+          .boxed().collect(Collectors.toList());
       } else {
-        for (int i = 0; i < fileIncluded.length; i++) {
-          if (fileIncluded[i]) {
-            internalColIds.add(i);
-          }
-        }
+        internalColIds = IntStream.range(1, fileTypes.size())
+          .filter(i -> fileIncluded[i])
+          .filter(i -> isOriginal || i != OrcRecordUpdater.ROW + 1)
```
abstractdog, line 1737: is it possible to reuse getRootColumn somehow? it would make more sense
```
  public static int getRootColumn(boolean isOriginal) {
    return isOriginal ? 0 : (OrcRecordUpdater.ROW + 1);
  }
```
so maybe:
```
...
filter(i -> i != getRootColumn(isOriginal))
...
```

deniskuzZ: this is not a root column, that Is an ACID <row> column. we should exclude `0` always + <row> struct id in case of non original ACID

ql/src/test/org/apache/hadoop/hive/ql/io/orc/TestInputOutputFormat.java
deniskuzZ: removed result

PR #5198, opened 2024-04-16
HIVE-28200: Improve get_partitions_by_filter/expr when partition limit enabled

wecharyu:

When `hive.metastore.limit.partition.request` is configured, HMS would get the matched partition counts before getting the real partition objects. The count operation could be a slow query if the input filter or expr is too complex, and such slow filter will be executed twice.
We can make an improvement by the new steps:
1. get matched partition names firstly, 
2. then check limit through the size of partition names,
3.  finally get the partitions by the partition names.

To reduce the slow filter execute times.

No.

No.

Add unit test and benchmark test.

```bash
java -jar hmsbench.jar -H localhost --savedata /tmp/benchdata --sanitize -N 100 -N 10000 -o bench_results.csv -C -d testbench_http --params=100 -M 'getPartitionsByFilter.*'
```
- Before this patch
```bash
getPartitionsByFilter   15.5682 15.1223 14.1135 20.1627 8.36433
getPartitionsByFilter.100       15.0892 14.9975 14.4175 17.5468 3.42779
getPartitionsByFilter.10000     36.3477 36.1636 35.1360 40.1805 2.22496
```
- After this patch
```bash
getPartitionsByFilter   10.5924 10.4109 9.78332 21.0283 12.3698
getPartitionsByFilter.100       10.5717 10.4636 9.89701 18.8140 8.36692
getPartitionsByFilter.10000     22.5289 21.8280 20.8911 30.0108 7.28651
```

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestMetaStoreLimitPartitionRequest.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestMetaStoreLimitPartitionRequest.java
@@ -160,7 +160,7 @@ public class TestMetaStoreLimitPartitionRequest {
   @Test
   public void testSimpleQueryWithDirectSqlTooManyPartitions() throws Exception {
     String queryString = "select value from %s where ds>'2008-04-20'";
-    executeQueryExceedPartitionLimit(queryString, 8);
+    executeQueryExceedPartitionLimit(queryString, 5);
   }
 
   @Test
@@ -244,20 +244,20 @@ public class TestMetaStoreLimitPartitionRequest {
   public void testQueryWithInWithFallbackToORMTooManyPartitions() throws Exception {
     setupNumTmpTable();
     String queryString = "select value from %s a where a.num in (select value from " + TABLE_NAME + "_num_tmp)";
-    executeQueryExceedPartitionLimit(queryString, 12);
+    executeQueryExceedPartitionLimit(queryString, 5);
   }
 
   @Test
   public void testQueryWithInWithFallbackToORMTooManyPartitions2() throws Exception {
     setupNumTmpTable();
     String queryString = "select value from %s a where a.num in (select value from " + TABLE_NAME + "_num_tmp where value='25')";
-    executeQueryExceedPartitionLimit(queryString, 12);
+    executeQueryExceedPartitionLimit(queryString, 5);
   }
 
   @Test
   public void testQueryWithLikeWithFallbackToORMTooManyPartitions() throws Exception {
     String queryString = "select value from %s where num like '3%%'";
-    executeQueryExceedPartitionLimit(queryString, 6);
+    executeQueryExceedPartitionLimit(queryString, 5);
   }
 
   private void setupNumTmpTable() throws SQLException {
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -1052,7 +1052,7 @@ public class MetastoreConf {
                 "get_partitions_by_names, \n" +
                 "get_partitions_with_auth, \n" +
                 "get_partitions_by_filter, \n" +
-                "get_partitions_spec_by_filter, \n" +
+                "get_partitions_spec_by_expr, \n" +
                 "get_partitions_by_expr,\n" +
                 "get_partitions_ps,\n" +
                 "get_partitions_ps_with_auth.\n" +
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -7227,10 +7227,11 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
       authorizeTableForPartitionMetadata(catName, dbName, tblName);
       if (needCheckPartitionLimit(args.getMax())) {
         // Since partition limit is configured, we need fetch at most (limit + 1) partition names
+        int requestMax = args.getMax();
         int max = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST) + 1;
         args = new GetPartitionsArgs.GetPartitionsArgsBuilder(args).max(max).build();
         List<String> partNames = rs.listPartitionNamesByFilter(catName, dbName, tblName, args);
-        checkLimitNumberOfPartitions(tblName, partNames.size(), args.getMax());
+        checkLimitNumberOfPartitions(tblName, partNames.size(), requestMax);
         ret = rs.getPartitionsByNames(catName, dbName, tblName,
             new GetPartitionsArgs.GetPartitionsArgsBuilder(args).partNames(partNames).build());
       } else {
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -5569,15 +5569,6 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     }
   }
 
-  private void checkLimitNumberOfPartitionsByExpr(String catName, String dbName, String tblName,
-                                                  byte[] filterExpr, int maxParts)
-      throws TException {
-    if (isPartitionLimitEnabled()) {
-      checkLimitNumberOfPartitions(tblName, get_num_partitions_by_expr(catName, dbName, tblName,
-          filterExpr), maxParts);
-    }
-  }
-
   private void checkLimitNumberOfPartitionsByPs(String catName, String dbName, String tblName,
                                                 List<String> partVals, int maxParts)
           throws TException {
@@ -7223,13 +7214,21 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     fireReadTablePreEvent(catName, dbName, tblName);
     List<Partition> ret = null;
     Exception ex = null;
+    RawStore rs = getMS();
     try {
-      checkLimitNumberOfPartitionsByFilter(catName, dbName,
-          tblName, args.getFilter(), args.getMax());
-
       authorizeTableForPartitionMetadata(catName, dbName, tblName);
+      if (isPartitionLimitEnabled()) {
+        // Since partition limit is configured, we need fetch at most (limit + 1) partition names
+        int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
+        int max = args.getMax() < 0 ? partitionLimit + 1 : Math.min(args.getMax(), partitionLimit + 1);
+        args = new GetPartitionsArgs.GetPartitionsArgsBuilder(args).max(max).build();
+        List<String> partNames = rs.listPartitionNamesByFilter(catName, dbName, tblName, args);
+        checkLimitNumberOfPartitions(TableName.getQualified(catName, dbName, tblName), partNames.size(), -1);
+        ret = rs.getPartitionsByNames(catName, dbName, tblName, partNames);
+      } else {
+        ret = rs.getPartitionsByFilter(catName, dbName, tblName, args);
+      }
 
-      ret = getMS().getPartitionsByFilter(catName, dbName, tblName, args);
       ret = FilterUtils.filterPartitionsIfEnabled(isServerFilterEnabled, filterHook, ret);
     } catch (Exception e) {
       ex = e;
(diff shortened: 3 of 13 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1052,7 +1052,7 @@ public enum ConfVars {
                 "get_partitions_by_names, \n" +
                 "get_partitions_with_auth, \n" +
                 "get_partitions_by_filter, \n" +
-                "get_partitions_spec_by_filter, \n" +
```
wecharyu, line 1055: Minor typo fix.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -3466,11 +3467,49 @@ protected List<String> getJdoResult(
     }.run(true);
   }
 
-  private List<String> getPartitionNamesViaOrm(Table table, ExpressionTree tree, String order,
```
wecharyu, line 3469: Refactor this method because it's used in partition level, and do not needs the cd field of table, but the original parameter `Table table`  retrieved via ORM must include cd information.

wecharyu:
@deniskuzZ @zhangbutao @saihemanth-cloudera : could you help review this patch?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7223,13 +7214,23 @@ private List<Partition> get_partitions_by_filter_internal(final String catName,
...
     try {
-      checkLimitNumberOfPartitionsByFilter(catName, dbName,
-          tblName, args.getFilter(), args.getMax());
-
       authorizeTableForPartitionMetadata(catName, dbName, tblName);
+      if (isPartitionLimitEnabled()) {
+        // Since partition limit is configured, we need fetch at most (limit + 1) partition names
+        int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
+        int requestMax = args.getMax();
+        int max = requestMax < 0 ? partitionLimit + 1 : Math.min(requestMax, partitionLimit + 1);
```
dengzhhu653, line 7224: if requestMax > 0 && requestMax <= partitionLimit, should we ignore the limit check and fall through to`rs.getPartitionsByFilter` directly?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7350,6 +7346,25 @@ public PartitionsByExprResult get_partitions_by_expr(
...
+  private PartitionsByExprResult get_partitions_by_expr_internal(
+      String catName, String dbName, String tblName, GetPartitionsArgs args) throws TException {
+    List<Partition> partitions = new LinkedList<>();
+    boolean hasUnknownPartitions = false;
+    RawStore rs = getMS();
+    if (isPartitionLimitEnabled()) {
+      // Since partition limit is configured, we need fetch at most (limit + 1) partition names
+      int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
+      int max = args.getMax() < 0 ? partitionLimit + 1 : Math.min(args.getMax(), partitionLimit + 1);
+      List<String> partNames = rs.listPartitionNames(catName, dbName, tblName, args.getDefaultPartName(), args.getExpr(), null, (short) max);
```
dengzhhu653, line 7358: I think in this case, we shouldn't cast the `max` to short, maybe we should change the type of `max` to int in the `listPartitionNames` method, what do you think?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1046,6 +1030,20 @@ private List<Partition> getPartitionsByNames(String catName, String dbName,
     return getPartitionsByQuery(catName, dbName, tblName, queryText, params, isAcidTable, args);
   }
 
+  public List<Partition> getPartitionsByPartitionIdsInBatch(String catName, String dbName,
```
dengzhhu653, line 1033: nit: can we declare this method as `private`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
```diff
@@ -855,6 +868,20 @@ boolean getPartitionsByExpr(String catName, String dbName, String tblName,
...
+   * Prune the input partition names that match the expressions.
+   * @param catName catalog name
+   * @param dbName database name
+   * @param tblName table name
+   * @param result input list of partition names
+   * @param args additional arguments for pruning
+   * @return true if the result contains unknown partition names.
+   * @throws MetaException error executing the expression
+   */
+  boolean prunePartitionNamesByExpr(String catName, String dbName, String tblName,
```
dengzhhu653, line 881: why we need this method in `RawStore`, I don't see it being called in the `HMSHandler`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
```diff
@@ -868,19 +895,6 @@ boolean getPartitionsByExpr(String catName, String dbName, String tblName,
...
-   * Get the number of partitions that match an already parsed expression.
-   * @param catName catalog name.
-   * @param dbName database name.
-   * @param tblName table name.
-   * @param expr an already parsed Hive expression
-   * @return number of matching partitions.
-   * @throws MetaException error accessing the RDBMS or working with the expression.
-   * @throws NoSuchObjectException no such table.
-   */
-  int getNumPartitionsByExpr(String catName, String dbName, String tblName, byte[] expr)
```
dengzhhu653, line 881: Since this method is introduced in 3.0.0, can we mark this method as `Deprecated` instead of removing it, we can delete it in the next major release.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -4154,18 +4209,14 @@ private String getDefaultPartitionName(String inputDefaultPartName) {
    * @param result The resulting names.
    * @return Whether the result contains any unknown partitions.
    */
-  private boolean getPartitionNamesPrunedByExprNoTxn(String catName, String dbName, String tblName, List<FieldSchema> partColumns, byte[] expr,
```
dengzhhu653, line 4157: nit: though it looks a bit un-formatted, I would prefer to keep it, this cloud help code review and easy to track via git history.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -3466,11 +3467,49 @@ protected List<String> getJdoResult(
...
+      @Override
+      protected List<String> getSqlResult(GetHelper<List<String>> ctx) throws MetaException {
+        return directSql.getPartitionNamesViaSql(filter, partitionKeys,
+            getDefaultPartitionName(args.getDefaultPartName()), null, args.getMax());
+      }
+
+      @Override
+      protected List<String> getJdoResult(GetHelper<List<String>> ctx)
+          throws MetaException, NoSuchObjectException, InvalidObjectException {
+        return getPartitionNamesViaOrm(catName, dbName, tblName, tree, null,
```
dengzhhu653, line 3501: what if the `getPartitionNamesViaOrm` returns null, should we throw a message to indicate there is an error in generating the filter?

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestMetaStoreLimitPartitionRequest.java
```diff
@@ -160,7 +160,7 @@ public void testMoreComplexQueryWithDirectSql() throws Exception {
   @Test
   public void testSimpleQueryWithDirectSqlTooManyPartitions() throws Exception {
     String queryString = "select value from %s where ds>'2008-04-20'";
-    executeQueryExceedPartitionLimit(queryString, 8);
+    executeQueryExceedPartitionLimit(queryString, 5);
```
dengzhhu653, line 163: why do we change the tests?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: Nice catch, it should be.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: Make sense, and it's a compatible change.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
wecharyu: Done.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
wecharyu: change to `private` in `ObjectStore`.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
wecharyu: Reverted.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
wecharyu: Throw exception in `ObjectStore#getPartitionNamesViaOrm`, because the rest api `get_partition_names_req` also use this method, we can not return a null value in the rest response.
https://github.com/apache/hive/blob/b970c980f187124ea850b8233fa4434288319254/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L3492-L3499

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestMetaStoreLimitPartitionRequest.java
wecharyu: because we will try to get at most `limit + 1` (i.e 4 + 1 = 5) partitions after this patch.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: This logic can also be used in other list partitions api, and now it actually causes a bug, create a follow-up ticket [HIVE-28239](https://issues.apache.org/jira/browse/HIVE-28239).

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -4156,16 +4210,10 @@ private String getDefaultPartitionName(String inputDefaultPartName) {
...
-            catName,
-            dbName,
-            tblName,
-            maxParts));
-    return expressionProxy.filterPartitionsByExpr(
-            partColumns,
-            expr,
-            getDefaultPartitionName(defaultPartName),
-            result);
+    result.addAll(getPartitionNamesNoTxn(catName, dbName, tblName, (short) -1));
```
dengzhhu653, line 4213: nit: do we need to consider the `maxParts` instead of `-1`?

dengzhhu653:
Thanks @wecharyu, there are some test failures seems related to the fix, cloud you please fix that?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
wecharyu: We'd better get all partition names here because `prunePartitionNamesByExpr()` may filter out some partition names.

dengzhhu653 (approved):

wecharyu:
@chenwyi2 do you mean using partition id to check the size and fetch partition objects? We did not use it in this PR because the `ObjectStore` has no API related to partition ID, it's indeed an improvement in the plan.

PR #5201, opened 2024-04-18
HIVE-28204: Remove some HMS obsolete scripts

dengzhhu653:


dengzhhu653:
Under standalone-metastore/metastore-server/src/main/sql, there are scripts for 1.2.0, mainly for TestSchemaToolForMetastore to test the upgrade from 1.2.0.

InvisibleProgrammer:
Hi, is that some legacy way how we handled the metastore upgrade scripts at an old Hive version?

dengzhhu653:
> Hi, is that some legacy way how we handled the metastore upgrade scripts at an old Hive version?

The upgrade story is almost the same, while these deleted scripts are for Hive version <= 2.3(including the upgrade scripts).
For the scripts target at 1.x or 0.x, I think we can remove them safety.
For the scripts target at 2.x, the branch 2.x already contains the scripts for initializing, we don't need to keep them on the 4.x branch, 
the scripts under https://github.com/apache/hive/tree/master/standalone-metastore/metastore-server/src/main/sql/postgres  can upgrade to the latest version from any releases that newer than 1.2.0, so I remove the upgrade scripts from the directories as well.

deniskuzZ:
> > Hi, is that some legacy way how we handled the metastore upgrade scripts at an old Hive version?
> 
> The upgrade story is almost the same, while these deleted scripts are for Hive version <= 2.3(including the upgrade scripts). For the scripts target at 1.x or 0.x, I think we can remove them safety. For the scripts target at 2.x, the branch 2.x already contains the scripts for initializing, we don't need to keep them on the 4.x branch, the scripts under https://github.com/apache/hive/tree/master/standalone-metastore/metastore-server/src/main/sql/postgres can upgrade to the latest version from any releases that newer than 1.2.0, so I remove the upgrade scripts from the directories as well.

what if customer decided to upgrade from 2.0 to 4.0, won't we miss some of the upgrade script?

dengzhhu653:
> > > Hi, is that some legacy way how we handled the metastore upgrade scripts at an old Hive version?
> > 
> > 
> > The upgrade story is almost the same, while these deleted scripts are for Hive version <= 2.3(including the upgrade scripts). For the scripts target at 1.x or 0.x, I think we can remove them safety. For the scripts target at 2.x, the branch 2.x already contains the scripts for initializing, we don't need to keep them on the 4.x branch, the scripts under https://github.com/apache/hive/tree/master/standalone-metastore/metastore-server/src/main/sql/postgres can upgrade to the latest version from any releases that newer than 1.2.0, so I remove the upgrade scripts from the directories as well.
> 
> what if customer decided to upgrade from 2.0 to 4.0, won't we miss some of the upgrade script?

https://github.com/apache/hive/tree/master/standalone-metastore/metastore-server/src/main/sql/postgres this directory contains the scripts for upgrading to 4.0, 

https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/sql/postgres/upgrade.order.postgres

dengzhhu653:
Hello @nrg4878, @saihemanth-cloudera cloud you please take a look as well if have cycles?

nrg4878:
@dengzhhu653 Given there is a RC for 2.3.10 release, should we retain the upgrade scripts from 2.3 to the latest version?

dengzhhu653:
> @dengzhhu653 Given there is a RC for 2.3.10 release, should we retain the upgrade scripts from 2.3 to the latest version?
This https://github.com/apache/hive/tree/master/standalone-metastore/metastore-server/src/main/sql/postgres directory contains the scripts for upgrading to 4.0 from 2.3 or 1.2. The PR removes the scripts in metastore module

henrib (commented):
Non binding, but LGTM :-)

dengzhhu653:
@kuczoram would you please take a look if have secs? Thank you in advance!

saihemanth-cloudera (approved):
LGTM +1

PR #5129, opened 2024-03-15
HIVE-28121: Use direct SQL for transactional altering table parameter

lirui-apache:
### What changes were proposed in this pull request?
Use direct SQL for transactional update table parameter and check the number of affected rows to detect concurrent writes.

### Why are the changes needed?
Maintain consistency in case of concurrent writes.

### Does this PR introduce _any_ user-facing change?
No

### Is the change a dependency upgrade?
No

### How was this patch tested?
Covered by existing tests

PR #4892, opened 2023-11-22
HIVE-7145: [2.3] Remove dependence on apache commons-lang

pan3793:
### What changes were proposed in this pull request?

Backport HIVE-7145 to branch-2.3

### Why are the changes needed?

The downstream projects like Spark would benefit from that.

### Does this PR introduce _any_ user-facing change?

No.

### Is the change a dependency upgrade?

Maybe? Actually, it's a migration.

### How was this patch tested?

Pass CI

PR #5204, opened 2024-04-20
HIVE-28121. (2.3) Use direct SQL for transactional altering table parameter

pan3793:

Port https://github.com/apache/hive/pull/5129 to branch-2.3

This is a follow-up of HIVE-26882, including correctness fixes for some RDBMS.

No

No

UT

### DIFF
--- a/metastore/src/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
+++ b/metastore/src/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
@@ -78,7 +78,6 @@ import org.datanucleus.store.rdbms.query.ForwardQueryResult;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 
 /**
--- a/metastore/src/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
+++ b/metastore/src/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
@@ -78,6 +78,7 @@ import org.datanucleus.store.rdbms.query.ForwardQueryResult;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 
 /**
@@ -1963,5 +1964,14 @@ class MetaStoreDirectSql {
       }
     }
     return ret;
+  }
+
+  long updateTableParam(Table table, String key, String expectedValue, String newValue) {
+    String queryText = String.format("UPDATE \"TABLE_PARAMS\" SET \"PARAM_VALUE\" = ? " +
+            "WHERE \"PARAM_KEY\" = ? AND \"PARAM_VALUE\" = ? AND \"TBL_ID\" IN " +
+            "(SELECT \"TBL_ID\" FROM \"TBLS\" JOIN \"DBS\" ON \"TBLS\".\"DB_ID\" = \"DBS\".\"DB_ID\" WHERE \"TBL_NAME\" = '%s' AND \"NAME\" = '%s')",
+        table.getTableName(), table.getDbName());
+    Query query = pm.newQuery("javax.jdo.query.SQL", queryText);
+    return (long) query.executeWithArray(newValue, key, expectedValue);
   }
 }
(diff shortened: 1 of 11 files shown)


metastore/src/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -608,6 +608,33 @@ public boolean openTransaction(String isolationLevel) {
     return result;
   }
 
+  @Override
+  public long updateParameterWithExpectedValue(Table table, String key, String expectedValue, String newValue)
+      throws MetaException, NoSuchObjectException {
+    final Table _table = table;
```
pan3793, line 614: this might be limitation of Java 7

lirui-apache (approved):
+1, thanks @pan3793 for taking care of this

pan3793:
@pvary Could you please take a look at this one? AFAIK it's the last one before the next RC for 2.3.10

metastore/src/java/org/apache/hadoop/hive/metastore/ObjectStore.java
pvary: This is ok

pvary (approved):
The backport seems ok.
I am a bit concerned about the failing tests, but if @sunchao says, that this is the current state of the tests on this branch, then we are ok to go

pan3793:
kindly ping @sunchao, are we good to go?

sunchao:
Thanks @pan3793 @lirui-apache and @pvary . It looks like this PR has more tests failing than the previous PR:

Screenshot 2024-04-23 at 8 57 41 PM

In the previous PR: #4892, there were only 27 tests failing (normally there would only be 26 tests but 1 was flaky in that PR).

Could you double check whether the test failures are related?

sunchao:
> that this is the current state of the tests on this branch

Yes that's unfortunately the current state of branch-2.3. We do know a fixed set of tests are failing though and are using that as the baseline for testing against PRs.

pan3793:
@sunchao based on the CI results of this PR branch, I think the failure tests should be flaky cases
http://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/activity?branch=PR-5204

sunchao:
OK, just re-triggered CI to confirm.

sunchao:
Hmm @pan3793 I tried the tests again for 2 times, and each time there were over 40 tests failing, which is quite more than the normal 26. Could you double-check whether there is anything we should fix here? Thanks.

pan3793:
@sunchao the failed tests are not caused by this patch, I opened another dummy PR based on the latest branch-2.3, and got the same failures https://github.com/apache/hive/pull/5225

sunchao:
oops, thanks @pan3793  - let me check the commits in branch-2.3 and see which commit caused the regression

sunchao (approved):

sunchao:
Merged, thanks

pan3793:
@sunchao can we have the next RC for 2.3.10?

sunchao:
Sorry for the delay @pan3793 . I just sent out RC1.

metastore/src/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1964,4 +1964,13 @@ public List<SQLPrimaryKey> getPrimaryKeys(String db_name, String tbl_name) throw
     }
     return ret;
   }
+
+  long updateTableParam(Table table, String key, String expectedValue, String newValue) {
+    String queryText = String.format("UPDATE \"TABLE_PARAMS\" SET \"PARAM_VALUE\" = ? " +
+            "WHERE \"PARAM_KEY\" = ? AND \"PARAM_VALUE\" = ? AND \"TBL_ID\" IN " +
```
pan3793, line 1970: Just for reference, we hit performance issues in our production environment on executing this SQL. Changing the `IN` to `=` would solve this issue, at least for MySQL 5.7

image
image

lirui-apache: Thanks @pan3793 for reporting the issue. I agree it's better to use scalar subquery here. But since both 2.x and 3.x lines are EOL, and subquery is not used in 4.0, so I guess we won't have to do anything at the moment, right?

pan3793: Yes, the Hive 4.0 doesn't have such an issue. This is just a tip for users who want to port this patch to their internal legacy Hive version (2.3 or 3.x)

PR #5206, opened 2024-04-21
HIVE-28205: Implement direct sql for get_partitions_ps_with_auth api

wecharyu:
Implement direct sql for `getNumPartitionsByPs` and `listPartitionsPsWithAuth`.

To improve performance.

No.

No.

Passing existing tests and add benchmark test `benchmarkGetPartitionsByPs`:
```bash
java -jar hmsbench.jar -H localhost --savedata /tmp/benchdata --sanitize -N 100 -N 1000 -o bench_results.csv -C -d testbench_http --params=100 -M 'getPartitionsByPs.*'
```

- Before this patch
```bash
getPartitionsByPs       14.0807 13.9823 12.8813 19.2509 5.63364
getPartitionsByPs.100   276.291 275.259 267.642 301.320 2.33622
getPartitionsByPs.1000  2631.50 2631.01 2551.76 2991.55 1.86420
```
- After this patch
```bash
getPartitionsByPs       11.6247 11.2780 10.7851 23.8686 12.7184
getPartitionsByPs.100   18.3550 18.0011 17.1338 36.5164 10.8893
getPartitionsByPs.1000  61.5760 60.6764 59.3697 78.2133 5.14500
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
@@ -22,7 +22,6 @@ import org.apache.hadoop.hive.metastore.api.InvalidOperationException;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
 import org.apache.hadoop.hive.metastore.api.Partition;
-import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -36,6 +35,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import static org.apache.hadoop.hive.metastore.Warehouse.LOG;
 import static org.apache.hadoop.hive.metastore.Warehouse.makePartName;
 import static org.apache.hadoop.hive.metastore.utils.MetaStoreUtils.makePartNameMatcher;
 
@@ -121,8 +121,8 @@ final class PartitionTree {
    * {@link MetaStoreUtils#getPvals(List, Map)}
    */
   List<Partition> getPartitionsByPartitionVals(List<String> partialPartVals) throws MetaException {
-    if (MetaStoreUtils.arePartValsEmpty(partialPartVals)) {
-      return new ArrayList<>(parts.values());
+    if (partialPartVals == null || partialPartVals.isEmpty()) {
+      throw new MetaException("Partition partial vals cannot be null or empty");
     }
     String partNameMatcher = makePartNameMatcher(tTable, partialPartVals, ".*");
     List<Partition> matchedPartitions = new ArrayList<>();
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
@@ -122,8 +122,8 @@ final class PartitionTree {
    * {@link MetaStoreUtils#getPvals(List, Map)}
    */
   List<Partition> getPartitionsByPartitionVals(List<String> partialPartVals) throws MetaException {
-    if (MetaStoreUtils.arePartValsEmpty(partialPartVals)) {
-      return new ArrayList<>(parts.values());
+    if (partialPartVals == null || partialPartVals.isEmpty()) {
+      throw new MetaException("Partition partial vals cannot be null or empty");
     }
     String partNameMatcher = makePartNameMatcher(tTable, partialPartVals, ".*");
     List<Partition> matchedPartitions = new ArrayList<>();
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.hive.metastore.api.InvalidOperationException;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
 import org.apache.hadoop.hive.metastore.api.Partition;
+import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -121,8 +122,8 @@ final class PartitionTree {
    * {@link MetaStoreUtils#getPvals(List, Map)}
    */
   List<Partition> getPartitionsByPartitionVals(List<String> partialPartVals) throws MetaException {
-    if (partialPartVals == null || partialPartVals.isEmpty()) {
-      throw new MetaException("Partition partial vals cannot be null or empty");
+    if (MetaStoreUtils.arePartValsEmpty(partialPartVals)) {
+      return new ArrayList<>(parts.values());
     }
     String partNameMatcher = makePartNameMatcher(tTable, partialPartVals, ".*");
     List<Partition> matchedPartitions = new ArrayList<>();
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
@@ -36,7 +36,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
-import static org.apache.hadoop.hive.metastore.Warehouse.LOG;
 import static org.apache.hadoop.hive.metastore.Warehouse.makePartName;
 import static org.apache.hadoop.hive.metastore.utils.MetaStoreUtils.makePartNameMatcher;
 
@@ -122,8 +121,8 @@ final class PartitionTree {
    * {@link MetaStoreUtils#getPvals(List, Map)}
    */
   List<Partition> getPartitionsByPartitionVals(List<String> partialPartVals) throws MetaException {
-    if (partialPartVals == null || partialPartVals.isEmpty()) {
-      throw new MetaException("Partition partial vals cannot be null or empty");
+    if (MetaStoreUtils.arePartValsEmpty(partialPartVals)) {
+      return new ArrayList<>(parts.values());
     }
     String partNameMatcher = makePartNameMatcher(tTable, partialPartVals, ".*");
     List<Partition> matchedPartitions = new ArrayList<>();
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
@@ -1118,7 +1118,7 @@ public class SessionHiveMetaStoreClient extends HiveMetaStoreClientWithLocalCach
       String tableName, List<String> partialPvals, int maxParts, String userName,
       List<String> groupNames) throws TException {
     org.apache.hadoop.hive.metastore.api.Table table = getTempTable(dbName, tableName);
-    if (table == null || partialPvals == null) {
+    if (table == null) {
       //(assume) not a temp table - Try underlying client
       return super.listPartitionsWithAuthInfo(catName, dbName, tableName, partialPvals, maxParts, userName,
           groupNames);
@@ -1179,7 +1179,7 @@ public class SessionHiveMetaStoreClient extends HiveMetaStoreClientWithLocalCach
   public List<String> listPartitionNames(String catName, String dbName, String tblName,
       List<String> partVals, int maxParts) throws TException {
     org.apache.hadoop.hive.metastore.api.Table table = getTempTable(dbName, tblName);
-    if (table == null || partVals == null) {
+    if (table == null) {
       return super.listPartitionNames(catName, dbName, tblName, partVals, maxParts);
     }
     TempTable tt = getPartitionedTempTable(table);
@@ -1295,7 +1295,7 @@ public class SessionHiveMetaStoreClient extends HiveMetaStoreClientWithLocalCach
   public List<Partition> listPartitions(String catName, String dbName, String tblName,
       List<String> partVals, int maxParts) throws TException {
     org.apache.hadoop.hive.metastore.api.Table table = getTempTable(dbName, tblName);
-    if (table == null || partVals == null) {
+    if (table == null) {
       return super.listPartitions(catName, dbName, tblName, partVals, maxParts);
     }
     TempTable tt = getPartitionedTempTable(table);
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
@@ -1118,7 +1118,7 @@ public class SessionHiveMetaStoreClient extends HiveMetaStoreClientWithLocalCach
       String tableName, List<String> partialPvals, int maxParts, String userName,
       List<String> groupNames) throws TException {
     org.apache.hadoop.hive.metastore.api.Table table = getTempTable(dbName, tableName);
-    if (table == null) {
+    if (table == null || partialPvals == null) {
       //(assume) not a temp table - Try underlying client
       return super.listPartitionsWithAuthInfo(catName, dbName, tableName, partialPvals, maxParts, userName,
           groupNames);
@@ -1179,7 +1179,7 @@ public class SessionHiveMetaStoreClient extends HiveMetaStoreClientWithLocalCach
   public List<String> listPartitionNames(String catName, String dbName, String tblName,
       List<String> partVals, int maxParts) throws TException {
     org.apache.hadoop.hive.metastore.api.Table table = getTempTable(dbName, tblName);
-    if (table == null) {
+    if (table == null || partVals == null) {
       return super.listPartitionNames(catName, dbName, tblName, partVals, maxParts);
     }
     TempTable tt = getPartitionedTempTable(table);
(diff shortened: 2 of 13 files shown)


standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/client/TestListPartitions.java
```diff
@@ -409,10 +409,13 @@ public void testListPartitionsByValues() throws Exception {
     assertEquals(0, partitions.size());
   }
 
-  @Test(expected = MetaException.class)
```
wecharyu, line 412: Empty part_vals should not throw `MetaException` because we support it in `ObjectStore` and regard it as fetching all partitions:
https://github.com/apache/hive/blob/2d855b27d31db6476f18870651db6987816bb5e3/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L3890-L3893

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -769,6 +769,28 @@ public List<Partition> run(List<Long> input) throws MetaException {
     });
   }
 
+  public List<Partition> getPartitionsViaSqlPs(Table table, GetPartitionsArgs args) throws MetaException {
+    String catName = table.getCatName();
+    String dbName = table.getDbName();
+    String tblName = table.getTableName();
+
+    String sqlFilter = "" + PARTITIONS + ".\"PART_NAME\" like ? ";
+    String partialName = MetaStoreUtils.makePartNameMatcher(table, args.getPart_vals(), "_%");
```
dengzhhu653, line 778: have you tested the matcher against other db? such as Postgres, MySQL, etc.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1270,6 +1291,31 @@ public int getNumPartitionsViaSqlFilter(SqlFilterForPushdown filter) throws Meta
...
+    Object[] params = new Object[4];
+    params[0] = table.getTableName();
+    params[1] = table.getDbName();
+    params[2] = table.getCatName();
+    params[3] = partialName;
+
+    return getNumPartitionsViaSql(queryText, params);
+  }
+
+  private int getNumPartitionsViaSql(String queryText, Object[] params) {
```
dengzhhu653, line 1317: can we rename this method to `getCountOfQuery` and move it to `MetastoreDirectSqlUtils`?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -2245,8 +2245,8 @@ public List<Partition> listPartitions(String db_name, String tbl_name,
   public List<Partition> listPartitions(String catName, String db_name, String tbl_name,
                                         List<String> part_vals, int max_parts) throws TException {
     // TODO should we add capabilities here as well as it returns Partition objects
-    if (db_name == null || tbl_name == null || part_vals == null) {
-      throw new MetaException("Database name/Table name/partition values should not be null");
+    if (db_name == null || tbl_name == null) {
```
dengzhhu653, line 2248: nit: if the client invokes this method specifically, I would like to add a check `part_vals == null`

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
```diff
@@ -121,8 +122,8 @@ List<Partition> addPartitions(List<Partition> partitions, boolean ifNotExists)
    * {@link MetaStoreUtils#getPvals(List, Map)}
    */
   List<Partition> getPartitionsByPartitionVals(List<String> partialPartVals) throws MetaException {
-    if (partialPartVals == null || partialPartVals.isEmpty()) {
-      throw new MetaException("Partition partial vals cannot be null or empty");
+    if (MetaStoreUtils.arePartValsEmpty(partialPartVals)) {
```
dengzhhu653, line 125: same as above

standalone-metastore/metastore-tools/tools-common/src/main/java/org/apache/hadoop/hive/metastore/tools/Util.java
```diff
@@ -648,6 +682,15 @@ static void addManyPartitionsNoException(@NotNull HMSClient client,
             addManyPartitions(client, dbName, tableName, parameters, arguments, npartitions));
   }
 
+  static void addManyPartitionsNoException(@NotNull HMSClient client,
+                                           @NotNull String dbName,
+                                           @NotNull String tableName,
+                                           @Nullable Map<String, String> parameters,
+                                           List<List<String>> values) {
+    throwingSupplierWrapper(() ->
+            addManyPartitions(client, dbName, tableName, parameters, values));
```
dengzhhu653, line 691: why not merge these two methods? `addManyPartitionsNoException` and `addManyPartitions`?

standalone-metastore/metastore-tools/tools-common/src/main/java/org/apache/hadoop/hive/metastore/tools/Util.java
```diff
@@ -584,15 +584,30 @@ static List<Partition> createManyPartitions(@NotNull Table table,
                                               @Nullable Map<String, String> parameters,
                                               @NotNull List<String> arguments,
                                               int npartitions) {
-    return IntStream.range(0, npartitions)
```
dengzhhu653, line 587: why split this method?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
wecharyu: Addressed.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
wecharyu: Since metastore server supports null and empty partial values, I think client can also support it.

standalone-metastore/metastore-tools/tools-common/src/main/java/org/apache/hadoop/hive/metastore/tools/Util.java
wecharyu: Done.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
wecharyu: it has been used in: https://github.com/apache/hive/blob/978194b7739e53b4b0e555a8b616a1ccf7d8d437/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L825-L828

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
dengzhhu653: I think the `part_vals` shouldn't be null according to the comment indicates: `Get a list of partitions based on a (possibly partial) list of partition values`.
Another point is what would happen if the new client connects to the old server which hasn't upgraded?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreDirectSqlUtils.java
```diff
@@ -609,4 +609,16 @@ public static void throwMetaOrRuntimeException(Exception e) throws MetaException
       throw new RuntimeException(e);
     }
   }
+
+  static int getCountOfQuery(PersistenceManager pm, String queryText, Object[] params) {
```
dengzhhu653, line 613: nit: how about declaring this method as `getCountOfQuery(Query query, Object[] params)`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -3960,9 +3926,57 @@ public List<Partition> listPartitionsPsWithAuth(String catName, String db_name,
     return partitions;
   }
 
+  private Integer getNumPartitionsViaOrmPs(Table table, List<String> partVals) throws MetaException {
+    // size is known since it contains dbName, catName, tblName and partialRegex pattern
+    Map<String, String> params = new HashMap<>(4);
```
dengzhhu653, line 3931: if this method isn't reused elsewhere, I would prefer put this codes into `getJdoResult`.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
wecharyu: Although old server supports null part_vals, let's revert to the original client behavior.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreDirectSqlUtils.java
wecharyu: Done.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
wecharyu: Done.

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
 import org.apache.hadoop.hive.metastore.api.Partition;
+import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
```
dengzhhu653, line 25: nit: can we revert the import as well?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreDirectSqlUtils.java
```diff
@@ -609,4 +609,10 @@ public static void throwMetaOrRuntimeException(Exception e) throws MetaException
       throw new RuntimeException(e);
     }
   }
+
+  static int getCountOfQuery(Query query, Object[] params) {
```
dengzhhu653, line 613: nit: how about move the QueryWrapper here as well, like:
```
long start = doTrace ? System.nanoTime() : 0;
try (QueryWrapper query = new QueryWrapper(query)) {
     query.setUnique(true);
     int sqlResult = MetastoreDirectSqlUtils.extractSqlInt(query.executeWithArray(params));
     long queryTime = doTrace ? System.nanoTime() : 0;
     MetastoreDirectSqlUtils.timingTrace(doTrace, queryText, start, queryTime);
     return sqlResult;
   }
```

wecharyu: Reuse `executeWithArray()` to replace this method.

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
wecharyu: Add null check in `SessionHiveMetaStoreClient`, here we can follow the server side behavior to support empty values.

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
```diff
@@ -1201,7 +1201,7 @@ public List<String> listPartitionNames(String catName, String dbName, String tbl
   public List<String> listPartitionNames(String catName, String dbName, String tblName,
       List<String> partVals, int maxParts) throws TException {
     org.apache.hadoop.hive.metastore.api.Table table = getTempTable(dbName, tblName);
-    if (table == null) {
+    if (table == null || partVals == null) {
```
dengzhhu653, line 1204: why if the table is a temporary table while the `partVals` is null?

wecharyu: It would throw MetaException like `HiveMetaStoreClient`.

dengzhhu653: Why not revert back the check on [PartitionTree.java](https://github.com/apache/hive/pull/5206/files#diff-f83827d7c6ec0a73bbc558101e64145afb555daff397c999c7491373253168c7), it's a bit confusing a temporary table just for performing a null check in `HiveMetaStoreClient`

wecharyu: Because `SessionHiveMetaStoreClient` extends `HiveMetaStoreClient`, it should also check null for part_vals like `HiveMetaStoreClient`, and `PartitionTree` should support empty or even null part_vals like metastore server in my opinion.

Otherwise it would cause the inconsistent test results in `TestListPartitions` for `HiveMetaStoreClient` and `SessionHiveMetaStoreClient`, for example `testListPartitionsByValuesNoVals()`. The master branch mark this test expect `MetaException`, but actually it should return all partitions for empty part_vals. It could pass the test in master because it throws `MetaException` when making part name for empty partVals in `getNumPartitionsByPs`, which should be a bug in implement this method.

dengzhhu653: > Because SessionHiveMetaStoreClient extends HiveMetaStoreClient, it should also check null for part_vals like HiveMetaStoreClient

I think you cannot just push the null check of a temporary table in `HiveMetaStoreClient`,  and the `PartitionTree` is for the temporary table to process the partition request, it works at the client side, why not put the check here?

wecharyu: > the PartitionTree is for the temporary table to process the partition request, it works at the client side, why not put the check here?

Because `get_partitions_ps_with_auth` support null and empty part_vals, the null check is added in client side. So in `PartitionTree` we can also follow the server behavior and push null check in `SessionHiveMetaStoreClient`. This change seems to have no side effects, and in turn supports empty part_vals like `HiveMetaStoreClient`.

dengzhhu653: I don't like this idea: null check for temporary table in HiveMetaStoreClient, let check others' opinion. cc @deniskuzZ @zhangbutao

zhangbutao: I tried to understand you folk's talk:

@wecharyu thinks that some hms partition api support `null ` & `empty `, such as `get_partitions_ps_with_auth`, but some hms partition api don't support `null ` & `empty`. So, for temp table, check `null `& `empry` and always throw exception in `PartitionTree ` in not inappropriate. 
I think in this case, @wecharyu you need to make sure all the api's behaviors which invoke the `PartitionTree ::getPartitionsByPartitionVals`  are unaffected. For example, `TempTable::getPartitionsByPartitionVals`  is invoked in four api, but i see that two of them are not changed after you change `PartitionTree ::getPartitionsByPartitionVals`, is that safe?


@dengzhhu653 I am not very familair with temp table's ability.  `PartitionTree` is only used for temp table, if temp table has same ability with common table, especially partition ability, and  meanwhile some hms api like `get_partitions_ps_with_auth` can allow `null `& `empty `, i think we can check in `HiveMetaStoreClient` side, and it is good for me to unify the behavior of common table and temp table.

wecharyu: > but i see that two of them are not changed

One is that `SessionHiveMetaStoreClient#istPartitionNamesRequest` does not check null for part_vals, I think it's acceptable because `HiveMetaStoreClient#listPartitionNamesRequest` also not check it. 

The other is `SessionHiveMetaStoreClient#exchangePartitions` where part_vals can not be null.

I can not find out the side effect on enable null or empty part_vals in these calls if we do regard null or empty part_vals is used to fetch all parts or part names.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java
```diff
@@ -460,6 +460,26 @@ public static List<String> getPvals(List<FieldSchema> partCols,
...
+   * @param partVals The partitions values used to filter out the partitions.
+   * @return true if partVals is empty or if all the values in partVals is empty strings.
+   * other wise false.
+   */
+  public static boolean arePartValsEmpty(List<String> partVals) {
+    if (partVals == null || partVals.isEmpty()) {
+      return true;
+    }
+    for (String val : partVals) {
+      if (val != null && !val.isEmpty()) {
```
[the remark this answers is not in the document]
wecharyu: It's existing code, the val is determined in client side.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java
```diff
@@ -460,6 +460,26 @@ public static List<String> getPvals(List<FieldSchema> partCols,
...
+        return false;
+      }
+    }
+    return true;
+  }
```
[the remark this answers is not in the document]
wecharyu: Good suggestion, will update together with the final discussion result of part_val null check.

wecharyu:
@dengzhhu653 @zhangbutao pls take a look again, I have reverted all client-related changes.

dengzhhu653 (approved):

PR #5209, opened 2024-04-23
HIVE-28211: Restore hive-exec-core jar

simhadri-g:

The hive-exec-core jar is used by spark, oozie, hudi and many other pojects. Removal of the hive-exec-core jar has caused the following issues. 

Spark : https://lists.apache.org/list?dev@hive.apache.org:lte=1M:joda 
Oozie: https://lists.apache.org/thread/yld75ltf9y8d9q3cow3xqlg0fqyj6mkg 
Hudi: https://github.com/apache/hudi/issues/8147 

Until the we  shade & relocate dependencies in hive-exec, we should restore the hive-exec core jar .

### DIFF
--- a/ql/pom.xml
+++ b/ql/pom.xml
@@ -1001,6 +1001,28 @@
             </goals>
           </execution>
           <execution>
+            <!-- Restoring Hive exec core jar: HIVE-26220. The core jar is being used by Spark, oozie and trino
+            projects as a result we will need to restore this to prevent downstream projects from breaking.
+            This 'execution' will produce a hive-exec-$version-core.jar which has
+            all classes from ql module, i.e. the same contents as hive-exec-$version.jar would
+            have had if there was shade plugin below.  The way shade was configured since 0.13, is
+            to override the default jar for ql module with the saded one but keep the same name.
+            In other words, ql modules producesd a single uber jar under the module's coordinates,
+            i.e. w/o a classifier.  We also need a slim jar that only has classes from ql.  This
+            jar will now be with a 'core' classifier to ensure that prior behavior is preserved.
+            It would have been better to have a slim jar called hive-exec-$version.jar and a fat
+            jar called hive-exec-$version-shaded.jar but doing this now will cause backwards compat.
+            issues-->
+            <id>core-jar</id>
+            <phase>package</phase>
+            <goals>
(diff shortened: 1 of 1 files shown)


zabetak (commented):
Hey @simhadri-g, thanks for taking the initiative. 

I would prefer before moving forward with this change to reach consensus in the community. Maybe its removal was a mistake but it did have some advantages. Please start or comment on an existing thread in the dev list and we can continue the discussion there.

Moreover, if we decide to restore the core jar it should be done under a separate JIRA ticket. HIVE-26220 is about shading problems so this is not strictly related. The shading problems will have to be resolved anyways.

simhadri-g:
Sure @zabetak , i will start or continue an existing mail thread shortly.
I have created a new jira for this as well: https://issues.apache.org/jira/browse/HIVE-28211 

The intention for this jira was to provide a smoother transition from the hive-exec-core to hive-exec jar for down stream projects  .

Thanks!

pan3793:
+1 for restoring. `hive-exec-core`, currently `hive-exec` shaded classes but not relocated, which can not be consumed by lots of downstream projects, choosing `hive-exec-core` and handling transitive deps manually is the only choice.

Handling hive transitive deps is always a challenge for downstream projects.

forward my previous comments on https://github.com/google/guava/issues/6666#issuecomment-1688687762
> `hive-exec-2.3.7.jar` is actually a standalone jar, and `hive-exec-2.3.7-core.jar` is actually the original normal jar. Unfortunately, `hive-exec-2.3.7.jar` seems to be abused as a normal dependency widely.

and https://github.com/apache/hive/pull/4888#issuecomment-1833000636
> Seems reducing dependencies is not a mission in the Hive project

simhadri-g:
Thanks @zabetak  and @pan3793 

I have started a mail thread on this here: https://lists.apache.org/thread/t2vzf1mydnzsckctcvfl93srrod4w76j 

Thanks!

dengzhhu653:
+1 from my side, as relocating the class seems not an easy task, it requires careful tests, which may in turn causing some inconvenience and delay of upgrading to the new major release for downstream project.

ayushtkn (approved):
LGTM

ayushtkn:
We have enough agreement on the ML as well, merging.

PR #5212, opened 2024-04-24
HIVE-28212: MiniHS2: use a base folder which is more likely writable on the local FS

abstractdog:
Use a different root folder in MiniHS2.

Describe on jira in detail, but TLDR SCRATCH_DIR makes much more sense to me.

No, MiniHS2 is a testing dev tool.

No.

Locally tested with MiniHS2.


simhadri-g (approved):
LGTM +1

aturoczy (approved):

kokila-19 (approved):

PR #5213, opened 2024-04-24
HIVE-28216:Upgrade Commons-Configuration to 2.10.1

devaspatikrishnatri:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,7 @@
     <commons-collections.version>3.2.2</commons-collections.version>
     <commons-collections4.version>4.1</commons-collections4.version>
     <commons-compress.version>1.26.0</commons-compress.version>
-    <commons-configuration.version>1.10</commons-configuration.version>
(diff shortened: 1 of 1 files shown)


ql/pom.xml
```diff
@@ -78,8 +78,8 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>commons-configuration</groupId>
-      <artifactId>commons-configuration</artifactId>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-configuration2</artifactId>
       <version>${commons-configuration.version}</version>
```
ayushtkn, line 83: Shouldn't there be changes in imports as well?
like change from ``org.apache.commons.configuration.`` to org.apache.commons.configuration2.``

You should even check the maven dependency tree, if anyone else is pulling ``commons-configuration`` and exclude that & add ``org.apache.commons.configuration.`` in restricted imports

PR #5214, opened 2024-04-25
HIVE-28217: Make default port effective

wForget:

Make default port effective.

Default port is not set effectively.

No

No

added unit test

### DIFF
--- a/jdbc/src/java/org/apache/hive/jdbc/Utils.java
+++ b/jdbc/src/java/org/apache/hive/jdbc/Utils.java
@@ -617,7 +617,7 @@ public class Utils {
           if (port <= 0) {
             port = Integer.parseInt(Utils.DEFAULT_PORT);
           }
-          connParams.setHost(jdbcBaseURI.getHost());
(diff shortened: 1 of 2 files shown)


jdbc/src/java/org/apache/hive/jdbc/Utils.java
```diff
@@ -618,7 +618,7 @@ public static JdbcConnectionParams extractURLComponents(String uri, Properties i
             port = Integer.parseInt(Utils.DEFAULT_PORT);
           }
           connParams.setHost(jdbcBaseURI.getHost());
```
ayushtkn, line 620: can we use ``host`` variable here to match the code style

ayushtkn (approved):
LGTM

jdbc/src/java/org/apache/hive/jdbc/Utils.java
wForget: > can we use `host` variable here to match the code style

sure, changed

PR #5217, opened 2024-04-26
HIVE-28220: Reset sessionVars before updating connection params from zookeeper

wForget:

Reset sessionVars before updating connection params from zookeeper.

When jdbc connection is retried, the principal of the next server is ignored, causing kerberos authentication of retried connection to always fail.

No

No

Manual test successful

### DIFF
--- a/jdbc/src/java/org/apache/hive/jdbc/Utils.java
+++ b/jdbc/src/java/org/apache/hive/jdbc/Utils.java
@@ -725,15 +725,21 @@ public class Utils {
    * explored. Also update the host, port, jdbcUriString and other configs published by the server.
    *
    * @param connParams
+   * @param info
    * @return true if new server info is retrieved successfully
    */
-  static boolean updateConnParamsFromZooKeeper(JdbcConnectionParams connParams) {
+  static boolean updateConnParamsFromZooKeeper(JdbcConnectionParams connParams, Properties info) throws JdbcUriParseException {
     // Add current host to the rejected list
     connParams.getRejectedHostZnodePaths().add(connParams.getCurrentHostZnodePath());
     String oldServerHost = connParams.getHost();
     int oldServerPort = connParams.getPort();
     // Update connection params (including host, port) from ZooKeeper
     try {
+      // We should reset sessionVars to the original client configurations to
+      // avoid the rejected server-side sessionVars affecting next server.
+      JdbcConnectionParams originalConnParams = extractURLComponents(connParams.getJdbcUriString(), info);
+      connParams.getSessionVars().clear();
+      connParams.getSessionVars().putAll(originalConnParams.getSessionVars());
       ZooKeeperHiveClientHelper.configureConnParams(connParams);
       connParams.setJdbcUriString(connParams.getJdbcUriString().replace(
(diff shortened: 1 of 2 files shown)


jdbc/src/java/org/apache/hive/jdbc/Utils.java
```diff
@@ -725,15 +725,21 @@ private static String getAuthorityFromJdbcURL(String uri) throws JdbcUriParseExc
...
     // Add current host to the rejected list
     connParams.getRejectedHostZnodePaths().add(connParams.getCurrentHostZnodePath());
     String oldServerHost = connParams.getHost();
     int oldServerPort = connParams.getPort();
     // Update connection params (including host, port) from ZooKeeper
     try {
+      // We should reset sessionVars to the original client configurations to
+      // avoid the rejected server-side sessionVars affecting next server.
+      JdbcConnectionParams originalConnParams = extractURLComponents(connParams.getJdbcUriString(), info);
+      connParams.getSessionVars().clear();
```
dengzhhu653, line 741: can you please elaborate a bit? looks we have already changed the host:
https://github.com/apache/hive/blob/4614ce72a7f366674d89a3a78f687e419400cb89/jdbc/src/java/org/apache/hive/jdbc/ZooKeeperHiveClientHelper.java#L185
or should we just remove the `AUTH_PRINCIPAL` from the `sessionVars`, so it will be changed accordingly?
https://github.com/apache/hive/blob/4614ce72a7f366674d89a3a78f687e419400cb89/jdbc/src/java/org/apache/hive/jdbc/ZooKeeperHiveClientHelper.java#L373

wForget: Yes, we changed the `host` but not `AUTH_PRINCIPAL`.

> or should we just remove the AUTH_PRINCIPAL from the sessionVars, so it will be changed accordingly?

In my scenario it is enough because all other `sessionVars` are the same. But I think it might be better to reset all `sessionVars`, what do you think?

dengzhhu653: I think all other vars remain the same, so instead of parsing the connection url again, I mean cloud we just `connParams.getSessionVars().remove(JdbcConnectionParams.AUTH_PRINCIPAL)` so that the `ZooKeeperHiveClientHelper` can reset the `AUTH_PRINCIPAL`?

wForget: Thanks,  makes sense to me, I will do that.

wForget: Sorry, I rethought it, it might not be good to remove `AUTH_PRINCIPAL` directly, we need to follow user configuration priority. It is possible that we configured `AUTH_PRINCIPAL` in the url instead of getting it from previous server zk data.

wForget: > can you please elaborate a bit? looks we have already changed the host:

I see what you mean, I used server principal with actual hostname instead of `_HOST`. But I think this is still an improvement.

PR #5218, opened 2024-04-26
HIVE-28224: Upgrade Orc version in Hive to 1.9.3

difin:

Upgrading Orc version in Hive to the latest version that supports Java 8 - 1.9.4

To keep the Orc version in Hive updated.

No

Pre-commit testing


ql/src/test/queries/clientpositive/insert_values_orig_table_use_metadata.q
```diff
@@ -1,5 +1,6 @@
 --! qt:dataset:src
 --! qt:dataset:alltypesorc
+--! qt:replace:/(\s+totalSize\s+)\S+(\s+)/$1#Masked#$2/
```
ayushtkn, line 3: I don't think these lines are required now, this masking is automatically done now

ql/src/test/results/clientpositive/llap/materialized_view_create_rewrite_10.q.out
```diff
@@ -79,17 +79,17 @@ STAGE PLANS:
                 TableScan
                   alias: t1
                   filterExpr: (b = 1) (type: boolean)
-                  Statistics: Num rows: 69 Data size: 13710 Basic stats: COMPLETE Column stats: NONE
+                  Statistics: Num rows: 70 Data size: #Masked# Basic stats: COMPLETE Column stats: NONE
```
ayushtkn, line 82: How did number of rows changed with ORC upgrade?

ql/src/test/results/clientpositive/llap/vector_decimal_aggregate.q.out
```diff
@@ -403,7 +403,7 @@ POSTHOOK: Input: default@decimal_vgby
...
 6981	3	5831542.2692483780	-515.6210729730	5830511.0271024320	1943503.67570081066667	2749258.455012492	3367140.1929065133	3	6984454.21109769200000	-617.56077692307690	6983219.08954384584620	2327739.696514615282066667	3292794.4113115156	4032833.0678006653
 762	2	5831542.2692483780	1531.2194054054	5833073.4886537834	2916536.74432689170000	2915005.5249214866	4122440.3477364695	2	6984454.21109769200000	1833.94569230769250	6986288.15678999969250	3493144.078394999846250000	3491310.1327026924	4937458.140118757
-NULL	3072	9318.4351351351	-4298.1513513514	5018444.1081079808	1633.60810810806667	5695.483082135323	5696.410307714464	3072	11160.71538461538500	-5147.90769230769300	6010604.30769230735360	1956.576923076922966667	6821.495748565151	6822.606289190915
+NULL	3072	9318.4351351351	-4298.1513513514	5018444.1081079808	1633.60810810806667	5695.483082135325	5696.4103077144655	3072	11160.71538461538500	-5147.90769230769300	6010604.30769230735360	1956.576923076922966667	6821.4957485651385	6822.606289190904
```
ayushtkn, line 406: query result is changing here?

ql/src/test/results/clientpositive/llap/vector_decimal_aggregate.q.out
```diff
@@ -847,4 +847,4 @@ FROM (SELECT cint,
 POSTHOOK: type: QUERY
 POSTHOOK: Input: default@decimal_vgby_small
 #### A masked pattern was here ####
-95165244160
+95767761728
```
ayushtkn, line 850: here also the result of the query is changing

ayushtkn (commented):
in general looks good, but I am not sure why the query results are changing, same query returning different results pre & post upgrade doesn't look f9 to me in general

ql/src/test/queries/clientpositive/insert_values_orig_table_use_metadata.q
difin: Done

ql/src/test/results/clientpositive/llap/materialized_view_create_rewrite_10.q.out
difin: That's number of stats rows that changed. Not sure how. The number of data rows is 2.

ql/src/test/results/clientpositive/llap/vector_decimal_aggregate.q.out
difin: This difference is caused by the previous differences in STDDEV_POP(cdecimal1) and STDDEV_SAMP(cdecimal1) fields

pom.xml
```diff
@@ -180,8 +180,8 @@
     <mysql.version>8.0.31</mysql.version>
     <postgres.version>42.7.3</postgres.version>
     <oracle.version>21.3.0.0</oracle.version>
-    <opencsv.version>2.3</opencsv.version>
-    <orc.version>1.8.5</orc.version>
+    <opencsv.version>5.9</opencsv.version>
```
dongjoon-hyun, line 183: May I ask why do you change `opencsv` under the PR title, `HIVE-28224: Upgrade Orc version in Hive to 1.9.3`, @difin ?

Could you proceed to upgrade `opencsv` library independently before upgrading ORC?

difin: I added opencsv changes because Hive wasn't able to compile without them. Hive has a code that depends on opencsv version that was coming from Orc 1.8.5 and it is not in 1.9.3.

I created a separate PR for upgrading opencsv https://github.com/apache/hive/pull/5240
Once it gets merged, I will update this PR and remove opencsv changes from it.

dongjoon-hyun: Thank you, @difin .

difin: You are welcome, @dongjoon-hyun

ql/src/test/queries/clientpositive/insert_values_orig_table_use_metadata.q
zhangbutao: `totalSize` is not needed. We can remove it.

ql/src/test/queries/clientpositive/materialized_view_create_rewrite_5.q
```diff
@@ -1,3 +1,4 @@
+--! qt:replace:/(\s+totalSize\s+)\S+(\s+)/$1#Masked#$2/
```
zhangbutao, line 1: `totalSize` is not needed.

ql/src/test/queries/clientpositive/stats_part.q
```diff
@@ -1,3 +1,5 @@
+-- Mask the totalSize value as it can have slight variability, causing test flakiness
+--! qt:replace:/(\s+totalSize\s+)\S+(\s+)/$1#Masked#$2/
```
zhangbutao, line 2: `totalSize` is not needed.

ql/src/test/results/clientpositive/llap/stats_part.q.out
```diff
@@ -161,7 +161,7 @@ Table Parameters:
 	numPartitions       	0                   
 	numRows             	0                   
 	rawDataSize         	0                   
-	totalSize           	#Masked#
+	totalSize           	#Masked#
```
zhangbutao, line 164: Need to regenerate this out once remove the `totalSize`.

ql/src/test/results/clientpositive/llap/materialized_view_create_rewrite_5.q.out
```diff
@@ -404,7 +404,7 @@ Table Parameters:
 	numFiles            	2                   
 	numRows             	10                  
 	rawDataSize         	0                   
-	totalSize           	#Masked#
+	totalSize           	#Masked#
```
zhangbutao, line 407: Need to regenerate this out once remove the `totalSize`.

ql/src/test/results/clientpositive/llap/insert_values_orig_table_use_metadata.q.out
```diff
@@ -108,7 +108,7 @@ Table Parameters:
 	numFiles            	0                   
 	numRows             	0                   
 	rawDataSize         	0                   
-	totalSize           	#Masked#
+	totalSize           	#Masked#
```
zhangbutao, line 111: Need to regenerate this out once remove the `totalSize`.

ql/src/test/queries/clientpositive/materialized_view_create_rewrite_5.q
difin: Done, thanks.

ql/src/test/results/clientpositive/llap/materialized_view_create_rewrite_5.q.out
difin: Done, thanks

ql/src/test/results/clientpositive/llap/stats_part.q.out
difin: Done, thanks

ql/src/test/results/clientpositive/llap/insert_values_orig_table_use_metadata.q.out
difin: Done, thanks

ql/src/test/queries/clientpositive/stats_part.q
difin: Done, thanks

ql/src/test/queries/clientpositive/insert_values_orig_table_use_metadata.q
difin: Done, thanks

pom.xml
difin: Hi @dongjoon-hyun, opencsv changes were removed from this PR and were done separately. The differences in the results of tests are still present. 
I did a few additional checks: 
- The same differences are present when Hive statistics are enabled and when they are disabled. 
- With the `decimal_vgby` table used as a source in the test in Parquet format instead of Orc, the differences are absent (so it is an additional confirmation that it is something related to Orc).
- I dumped the content of the `decimal_vgby` table with and without Orc upgrade and they were the same, which is odd

difin: Hi @dongjoon-hyun,
I found that this Hive behavior changed as a result of [ORC-1266](https://issues.apache.org/jira/browse/ORC-1266).
It looks like this commit causes data correctness issues in Orc.
Can you please help?

cxzl25: ORC-1266, at first, we call reset method every time we finish using `DecimalColumnVector`, to avoid inaccurate isRepeating and incorrect data reading, then we change to set `isRepeating=true` first, and then gradually traverse the data, if not repeated, then set `isRepeating=false`.
Later changed to first set to `isRepeating=true`, and then gradually traverse the data, if not repeated and then set `isRepeating=false`.

--- 

I did some validation testing on this vector_decimal_aggregate.q query and it has three values ​​that appear 1024 times, which is the same as the batch size read.
```
-119.4594594595
9318.4351351351
-4298.1513513514
```

Before ORC-1266, read 1024 identical values `-119.4594594595`, DecimalColumnVector's isRepeating is false, and vector has 1024 values.

After ORC-1266, because the value of the whole batch is the same, DecimalColumnVector's isRepeating is true, and vector has 1024 values.

This change should not be affected, because the read value has not changed.

--- 

Or I could submit a patch in ORC and not change it to true when isRepeating is set to false first.

deniskuzZ: @cxzl25, how do we handle repeated nulls? it seems that `setIsRepeatingIfNeeded` is only called when `result.noNulls || !result.isNull[0]`?

cxzl25: > Or I could submit a patch in ORC and not change it to true when isRepeating is set to false first.

See https://github.com/apache/orc/pull/1960

difin: @cxzl25 I did a test on my local env: backported your Orc patch into Orc branch 1.9 and run the Hive query test that had been failing previously. Now the test passed. Can you create a new Orc release from branch 1.9 so that we could use it in Hive? This is because we can't use Orc 2.0 or later because it requires Java 17+ and Hive still support Java 8.

cxzl25: > how do we handle repeated nulls in ORC?

`isRepeating=true, noNulls=false, isNull[]=true`

image


> Also why logic is different from, for example, `DoubleColumnVector`, where we set `result.isRepeating = false;` at the beginning of `nextVector`

Because the formats for storing data are different, DecimalTreeReader and DoubleTreeReader behave differently.

org.apache.orc.impl.TreeReaderFactory.DecimalTreeReader#nextVector
```java
      // read the scales
      scaleReader.nextVector(result, scratchScaleVector, batchSize);
```

dongjoon-hyun:
FYI, Apache ORC community is preparing a new 1.9.4 release . Please let us know if something is required for this Apache Hive community activity.
- https://github.com/apache/orc/issues/1969 (Release Manager: @williamhyun)

ql/src/java/org/apache/hadoop/hive/ql/io/orc/encoded/EncodedTreeReaderFactory.java
```diff
@@ -1078,6 +1078,7 @@ public void nextVector(ColumnVector previousVector, boolean[] isNull, final int
         FilterContext filterContext, ReadPhase readPhase) throws IOException {
       if (vectors == null) {
         super.nextVector(previousVector, isNull, batchSize, filterContext, readPhase);
+        previousVector.isRepeating = false;
```
ayushtkn, line 1081: @dongjoon-hyun /  @cxzl25  can you help confirm, hardcoding `isRepeating` here would be correct workaround for [ORC-1741](https://issues.apache.org/jira/browse/ORC-1741) & it won't lead to any other side affects?

PR #5219, opened 2024-04-27
HIVE-28225: Iceberg: Delete on entire table fails on COW mode.

ayushtkn:

Fix delete on entire iceberg table with COW mode, with truncate disabled

To avoid NPE & make the command work

Yes, instead of NPE, the query succeeds

No

UT


zabetak (approved):
LGTM, although it seems that this code path may be removed in the near future given that `hive.optimize.delete.all` is deprecated.

PR #5223, opened 2024-04-29
HIVE-28219: Support drop partitions by names in IMetaStoreClient

wecharyu:
In thrift api definition, HMS support to drop partitions by both partition names and expressions. But current api in `IMetaStoreClient` only support drop partitions by expressions, we should add new api to support drop partitions by names.
```java
  // IMetaStoreClient.java
  List<Partition> dropPartitions(String catName, String dbName, String tblName,
                                 RequestPartsSpec partsSpec, PartitionDropOptions options)
      throws NoSuchObjectException, MetaException, TException;
```

To improve the `dropPartitions` function.

Yes, user can use the new api to drop partitions by names.

No.

Add unit test:
```bash
mvn test -Dtest=org.apache.hadoop.hive.ql.metadata.TestHive -pl :hive-exec
```

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -54,7 +54,7 @@ import org.apache.hadoop.hive.metastore.api.SQLPrimaryKey;
 import org.apache.hadoop.hive.metastore.api.SerDeInfo;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
-import org.apache.hadoop.hive.metastore.client.BaseMetaStoreClient;
+import org.apache.hadoop.hive.metastore.client.ThriftHiveMetaStoreClient;
 import org.apache.hadoop.hive.metastore.partition.spec.PartitionSpecProxy;
 import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
 import org.apache.hadoop.hive.ql.QueryState;
@@ -1242,7 +1242,7 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
     } else if (partsSpec.isSetNames()) {
       preTruncateTable(hmsTable, context, partsSpec.getNames());
     }
-    context.putToProperties(BaseMetaStoreClient.SKIP_DROP_PARTITION, "true");
+    context.putToProperties(ThriftHiveMetaStoreClient.SKIP_DROP_PARTITION, "true");
   }
 
   private static void validatePartitionSpec(SearchArgument sarg, PartitionSpec partitionSpec) {
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -1225,7 +1225,6 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
     } catch (IOException e) {
       throw new MetaException(String.format("Error while fetching the partitions due to: %s", e));
     }
-    context.putToProperties(BaseMetaStoreClient.SKIP_DROP_PARTITION, "true");
   }
 
   @Override
@@ -1242,8 +1241,8 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
       preDropPartitions(hmsTable, context, partExprs);
     } else if (partsSpec.isSetNames()) {
       preTruncateTable(hmsTable, context, partsSpec.getNames());
-      context.putToProperties(BaseMetaStoreClient.SKIP_DROP_PARTITION, "true");
     }
+    context.putToProperties(BaseMetaStoreClient.SKIP_DROP_PARTITION, "true");
   }
 
   private static void validatePartitionSpec(SearchArgument sarg, PartitionSpec partitionSpec) {
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -45,14 +45,16 @@ import org.apache.hadoop.hive.metastore.HiveMetaHook;
 import org.apache.hadoop.hive.metastore.PartitionDropOptions;
 import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.CreateTableRequest;
+import org.apache.hadoop.hive.metastore.api.DropPartitionsExpr;
 import org.apache.hadoop.hive.metastore.api.EnvironmentContext;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.MetaException;
+import org.apache.hadoop.hive.metastore.api.RequestPartsSpec;
 import org.apache.hadoop.hive.metastore.api.SQLPrimaryKey;
 import org.apache.hadoop.hive.metastore.api.SerDeInfo;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
-import org.apache.hadoop.hive.metastore.client.ThriftHiveMetaStoreClient;
+import org.apache.hadoop.hive.metastore.client.BaseMetaStoreClient;
 import org.apache.hadoop.hive.metastore.partition.spec.PartitionSpecProxy;
 import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
 import org.apache.hadoop.hive.ql.QueryState;
@@ -1223,7 +1225,25 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
     } catch (IOException e) {
       throw new MetaException(String.format("Error while fetching the partitions due to: %s", e));
     }
-    context.putToProperties(ThriftHiveMetaStoreClient.SKIP_DROP_PARTITION, "true");
+    context.putToProperties(BaseMetaStoreClient.SKIP_DROP_PARTITION, "true");
+  }
+
+  @Override
+  public void preDropPartitions(org.apache.hadoop.hive.metastore.api.Table hmsTable,
+                                EnvironmentContext context,
+                                RequestPartsSpec partsSpec) throws MetaException {
+    if (partsSpec.isSetExprs()) {
+      List<DropPartitionsExpr> exprs = partsSpec.getExprs();
+      List<org.apache.commons.lang3.tuple.Pair<Integer, byte[]>> partExprs = Lists.newArrayList();
+      for (DropPartitionsExpr expr : exprs) {
+        partExprs.add(
+            org.apache.commons.lang3.tuple.Pair.of(expr.getPartArchiveLevel(), expr.getExpr()));
+      }
+      preDropPartitions(hmsTable, context, partExprs);
+    } else if (partsSpec.isSetNames()) {
+      preTruncateTable(hmsTable, context, partsSpec.getNames());
+      context.putToProperties(BaseMetaStoreClient.SKIP_DROP_PARTITION, "true");
+    }
   }
 
   private static void validatePartitionSpec(SearchArgument sarg, PartitionSpec partitionSpec) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -4081,13 +4081,13 @@ private void constructOneLBLocationMap(FileStatus fSta,
       exprs.add(dpe);
     }
     rps.setExprs(exprs);
-    return dropPartitions(getDefaultCatalog(conf), dbName, tableName, rps, dropOptions);
+    return dropPartitions(new TableName(getDefaultCatalog(conf), dbName, tableName), rps, dropOptions);
   }
 
-  public List<Partition> dropPartitions(String catName, String dbName, String tableName,
+  public List<Partition> dropPartitions(TableName tableName,
       RequestPartsSpec partsSpec, PartitionDropOptions dropOptions) throws HiveException {
     try {
-      Table table = getTable(dbName, tableName);
+      Table table = getTable(tableName.getDb(), tableName.getTable());
       if (!dropOptions.deleteData) {
         AcidUtils.TableSnapshot snapshot = AcidUtils.getTableSnapshot(conf, table, true);
         if (snapshot != null) {
@@ -4098,7 +4098,7 @@ private void constructOneLBLocationMap(FileStatus fSta,
         dropOptions.setTxnId(txnId);
       }
       List<org.apache.hadoop.hive.metastore.api.Partition> partitions = getMSC().dropPartitions(
-          catName, dbName, tableName, partsSpec, dropOptions, null);
+          tableName, partsSpec, dropOptions, null);
       return convertFromMetastore(table, partitions);
     } catch (NoSuchObjectException e) {
       throw new HiveException("Partition or table doesn't exist.", e);
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -154,6 +154,7 @@ import org.apache.hadoop.hive.metastore.api.DataConnector;
 import org.apache.hadoop.hive.metastore.api.Database;
 import org.apache.hadoop.hive.metastore.api.DefaultConstraintsRequest;
 import org.apache.hadoop.hive.metastore.api.DropDatabaseRequest;
+import org.apache.hadoop.hive.metastore.api.DropPartitionsExpr;
 import org.apache.hadoop.hive.metastore.api.EnvironmentContext;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.FireEventRequest;
@@ -188,6 +189,7 @@ import org.apache.hadoop.hive.metastore.api.PrimaryKeysRequest;
 import org.apache.hadoop.hive.metastore.api.PrincipalPrivilegeSet;
 import org.apache.hadoop.hive.metastore.api.PrincipalType;
 import org.apache.hadoop.hive.metastore.api.PrivilegeBag;
+import org.apache.hadoop.hive.metastore.api.RequestPartsSpec;
 import org.apache.hadoop.hive.metastore.api.Role;
 import org.apache.hadoop.hive.metastore.api.RolePrincipalGrant;
 import org.apache.hadoop.hive.metastore.api.SQLAllTableConstraints;
@@ -4069,6 +4071,21 @@ private void constructOneLBLocationMap(FileStatus fSta,
   public List<Partition> dropPartitions(String dbName, String tableName,
       List<Pair<Integer, byte[]>> partitionExpressions,
       PartitionDropOptions dropOptions) throws HiveException {
+    RequestPartsSpec rps = new RequestPartsSpec();
+    List<DropPartitionsExpr> exprs = new ArrayList<>(partitionExpressions.size());
+
+    for (Pair<Integer, byte[]> partExpr : partitionExpressions) {
+      DropPartitionsExpr dpe = new DropPartitionsExpr();
+      dpe.setExpr(partExpr.getRight());
+      dpe.setPartArchiveLevel(partExpr.getLeft());
+      exprs.add(dpe);
+    }
+    rps.setExprs(exprs);
+    return dropPartitions(getDefaultCatalog(conf), dbName, tableName, rps, dropOptions);
+  }
+
+  public List<Partition> dropPartitions(String catName, String dbName, String tableName,
+      RequestPartsSpec partsSpec, PartitionDropOptions dropOptions) throws HiveException {
     try {
       Table table = getTable(dbName, tableName);
       if (!dropOptions.deleteData) {
@@ -4080,8 +4097,8 @@ private void constructOneLBLocationMap(FileStatus fSta,
           .map(ss -> ss.getTxnMgr().getCurrentTxnId()).orElse(0L);
         dropOptions.setTxnId(txnId);
       }
-      List<org.apache.hadoop.hive.metastore.api.Partition> partitions = getMSC().dropPartitions(dbName, tableName,
-          partitionExpressions, dropOptions);
+      List<org.apache.hadoop.hive.metastore.api.Partition> partitions = getMSC().dropPartitions(
+          catName, dbName, tableName, partsSpec, dropOptions, null);
       return convertFromMetastore(table, partitions);
     } catch (NoSuchObjectException e) {
       throw new HiveException("Partition or table doesn't exist.", e);
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
@@ -1562,11 +1562,11 @@ public class SessionHiveMetaStoreClient extends MetaStoreClientWrapper {
   }
 
   @Override
-  public List<Partition> dropPartitions(String catName, String dbName, String tblName,
+  public List<Partition> dropPartitions(TableName tableName,
       RequestPartsSpec partsSpec, PartitionDropOptions options, EnvironmentContext context)
       throws TException {
-    if (isDefaultCatalog(catName)) {
-      Table table = getTempTable(dbName, tblName);
+    if (isDefaultCatalog(tableName.getCat())) {
+      Table table = getTempTable(tableName.getDb(), tableName.getTable());
       if (table != null) {
         TempTable tt = getPartitionedTempTable(table);
             List<List<String>> partValues = new ArrayList<>();
@@ -1602,7 +1602,7 @@ public class SessionHiveMetaStoreClient extends MetaStoreClientWrapper {
       }
     }
 
-    return delegate.dropPartitions(catName, dbName, tblName, partsSpec, options, context);
+    return delegate.dropPartitions(tableName, partsSpec, options, context);
   }
 
   @Override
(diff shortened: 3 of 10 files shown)


wecharyu:
This PR is followed https://github.com/apache/hive/pull/4905#discussion_r1464194112, @zhangbutao @dengzhhu653 @deniskuzZ pls take a look in your convenience.

zhangbutao:
Does `drop partitions by names` have any advantages over `drop partitions by expressions`?

My concern is that we already have lots of redundant api and other components(Spark/Trino/Impala) may not need the new added api.

But if the new added api has some advantages over old api, i think we can move on.

I also want to listen to other folks's opinions. :)

wecharyu:
> Does drop partitions by names have any advantages over drop partitions by expressions?

drop partitions by names is not aimed to replace drop by expressions, they each serve distinct purpose. 
And it's more easy to use and do not need deserialization in server side compared with expressions.

dengzhhu653:
> Does `drop partitions by names` have any advantages over `drop partitions by expressions`?
> 
> My concern is that we already have lots of redundant api and other components(Spark/Trino/Impala) may not need the new added api.
> 
> But if the new added api has some advantages over old api, i think we can move on.
> 
> I also want to listen to other folks's opinions. :)

+1

This change looks like a variant of `dropPartitions` based on one current implementation, it's not a good choice to declare it as a first class API.

If we really need to add it in the future, I would prefer making it an util method instead in `MetaStoreUtils`

wecharyu:
> This change looks like a variant of dropPartitions based on one current implementation, it's not a good choice to declare it as a first class API.

It's more like a function complement, before this PR, hive client can only drop partitions by exprs, but actually the server side supports drop partitions by names.

wecharyu:
@zhangbutao @dengzhhu653 @deniskuzZ  could you take a look again on this PR?
The HMS rest API `drop_partitions_req` supports both partition `names` and `expressions`:
```bash
union RequestPartsSpec {
  1: list<string> names;
  2: list<DropPartitionsExpr> exprs;
}
```
But the current client implementation only exposes methods that accept partition `expressions`, I think it's necessary to add support for specifying partition `names` on client side as well.

deniskuzZ:
hi @wecharyu, could you please create a thread on a dev list to restart the discussion? 
I think your point is valid. 
Btw, I found the client implementation exposing the `dropPartitions` by names API in `metastore.tools.HMSClient`:
````
  DropPartitionsResult dropPartitions(@NotNull String dbName, @NotNull String tableName,
                                      @Nullable List<String> partNames) throws TException {
    if (partNames == null) {
      return dropPartitions(dbName, tableName, getPartitionNames(dbName, tableName));
    }
    if (partNames.isEmpty()) {
      return null;
    }
    return client.drop_partitions_req(new DropPartitionsRequest(dbName,
        tableName, RequestPartsSpec.names(partNames)));
  }
````

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java [hunk 1]
```diff
@@ -4031,6 +4032,29 @@ public List<Partition> dropPartitions(String dbName, String tableName,
     }
   }
 
+  public List<Partition> dropPartitions(String dbName, String tableName,
```
deniskuzZ: I don't think we need it here, though. Only expose via HMSClient

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -2179,6 +2179,37 @@ List<Partition> dropPartitions(String catName, String dbName, String tblName,
...
+   * @param dbName Name of the database
+   * @param tblName Name of the table
+   * @param partsSpec request partition specification
+   * @param options Boolean options for dropping partitions
+   * @return List of Partitions dropped
+   * @throws NoSuchObjectException No partition matches the expression(s), and ifExists was false.
+   * @throws MetaException error access the RDBMS or storage.
+   * @throws TException On failure
+   */
+  List<Partition> dropPartitions(String dbName, String tblName,
```
deniskuzZ: do we really need this one?

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
```diff
@@ -1449,26 +1452,39 @@ public boolean dropPartition(String catName, String dbName, String tableName, St
 
   @Override
   public List<Partition> dropPartitions(String catName, String dbName, String tblName,
-      List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options) throws TException {
+                                        RequestPartsSpec partsSpec, PartitionDropOptions options)
```
deniskuzZ: It seems like a breaking change. Spark, for instance, uses hive-exec-core and might not handle the API change well.

deniskuzZ:
hi @wecharyu, please take a look when you have time. I've added a few comments. Thanks!

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
wecharyu: Some engines like Spark implement shims for `Hive.java`, It seems better to expose it here too.
https://github.com/apache/spark/blob/6c6ca1f0b6f43a2034f4d6bcf1af1f07e6b7ccc7/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveShim.scala#L266-L277

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
wecharyu: Currently Spark only support drop single partition in a call, if the `dropPartitionByNames` are exposed in client, spark could use it to speed up the drop partitions ddl.

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
wecharyu: Not really, the existing function `dropPartitions(..., List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options)` will call the new function `dropPartitions(..., RequestPartsSpec partsSpec, PartitionDropOptions options)`. We does not change the existing function.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
deniskuzZ: any clue why don't they use the public API and shade the hive-exec-core instead?

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
deniskuzZ: i don't see the old signature anymore. didn't you change it here? `List<Pair<Integer, byte[]>> partExprs` -> `RequestPartsSpec partsSpec`

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
deniskuzZ: can't we reuse existing method to avoid code duplication?
````
 public List<Partition> dropPartitions(String dbName, String tableName,
      List<Pair<Integer, byte[]>> partitionExpressions,
      PartitionDropOptions dropOptions) 
````

File: standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java [hunk 4]
```diff
@@ -1937,24 +1937,38 @@ public List<Partition> dropPartitions(String catName, String dbName, String tblN
                                         PartitionDropOptions options) throws TException {
     RequestPartsSpec rps = new RequestPartsSpec();
     List<DropPartitionsExpr> exprs = new ArrayList<>(partExprs.size());
+    for (Pair<Integer, byte[]> partExpr : partExprs) {
```
deniskuzZ: This was moved to `ThriftHiveMetaStoreClient`, no wonder i wasn't able to find link with SessionHiveMetaStoreClient.
This PR needs rebase.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
deniskuzZ: @wecharyu, could we add a generic signature with TableName instead of existing 2?

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
deniskuzZ: https://github.com/apache/hive/pull/5223#discussion_r2206838985

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
wecharyu: Addressed.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -2092,6 +2092,10 @@ List<Partition> dropPartitions(String catName, String dbName, String tblName,
       List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options, EnvironmentContext context)
       throws NoSuchObjectException, MetaException, TException;
 
+  List<Partition> dropPartitions(String catName, String dbName, String tblName,
```
deniskuzZ: @wecharyu, could we replace the triplet (catName, dbName, tblName) with TableName? Please see [HIVE-28166](https://issues.apache.org/jira/browse/HIVE-28166) as a reference

wecharyu: Addressed.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaHook.java
```diff
@@ -226,4 +227,16 @@ default void preDropPartitions(Table table,
...
   }
+
+ /**
+   * Called before dropping the partitions from the table in the metastore during ALTER TABLE DROP PARTITION.
+   * @param table table whose partition needs to be dropped
+   * @param context context of the  operation
+   * @param partsSpec request partition specification
+   * @throws MetaException
+   */
+  default void preDropPartitions(Table table,
```
deniskuzZ, line 241: should we deprecate the other `preDropPartitions` with `List<Pair<Integer, byte[]>> partExprs`?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -2092,6 +2092,10 @@ List<Partition> dropPartitions(String catName, String dbName, String tblName,
       List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options, EnvironmentContext context)
       throws NoSuchObjectException, MetaException, TException;
 
+  List<Partition> dropPartitions(TableName tableName,
```
deniskuzZ, line 2124: should we deprecate `dropPartitions` with `List<Pair<Integer, byte[]>> partExprs`?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/BaseMetaStoreClient.java
```diff
@@ -557,6 +561,32 @@ public final List<Partition> dropPartitions(String catName, String dbName, Strin
     return dropPartitions(catName, dbName, tblName, partExprs, options, null);
   }
 
+  @Override
+  public List<Partition> dropPartitions(String catName, String dbName, String tblName,
```
deniskuzZ, line 564: why the `SessionHiveMetaStoreClient#dropPartitions` migrated here?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/ThriftHiveMetaStoreClient.java
```diff
@@ -1588,34 +1588,18 @@ public boolean dropPartition(String catName, String db_name, String tbl_name,
...
-  public List<Partition> dropPartitions(String catName, String dbName, String tblName,
-      List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options, EnvironmentContext context)
+  public List<Partition> dropPartitions(TableName tableName,
+      RequestPartsSpec partsSpec, PartitionDropOptions options, EnvironmentContext context)
       throws NoSuchObjectException, MetaException, TException {
-    if (context == null) {
-      context = new EnvironmentContext();
-    }
-
-    if (context.getProperties() != null &&
```
deniskuzZ, line 1598: shoud we keep `SKIP_DROP_PARTITION` flag here?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -2092,6 +2092,10 @@ List<Partition> dropPartitions(String catName, String dbName, String tblName,
...
+      RequestPartsSpec partsSpec, PartitionDropOptions options, EnvironmentContext context)
+      throws NoSuchObjectException, MetaException, TException;
```
deniskuzZ: NoSuchObjectException, MetaException are not thrown

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/ThriftHiveMetaStoreClient.java
```diff
@@ -1588,34 +1588,18 @@ public boolean dropPartition(String catName, String db_name, String tbl_name,
...
-    req.setCatName(catName);
+    DropPartitionsRequest req = new DropPartitionsRequest(tableName.getDb(), tableName.getTable(), partsSpec);
+    req.setCatName(tableName.getCat());
     req.setDeleteData(options.deleteData);
     req.setNeedResult(options.returnResults);
     req.setIfExists(options.ifExists);
 
+    if (context == null) {
+      context = new EnvironmentContext();
+    }
```
deniskuzZ, line 1603: drop the redundant `context = Optional.ofNullable(context).orElse(new EnvironmentContext());` below

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
```diff
@@ -1559,36 +1562,47 @@ public boolean dropPartition(String catName, String dbName, String tableName, St
   }
 
   @Override
-  public List<Partition> dropPartitions(String catName, String dbName, String tblName,
-      List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options, EnvironmentContext context)
+  public List<Partition> dropPartitions(TableName tableName,
```
deniskuzZ, line 1565: what happens if someone invokes the legacy signature? it's gonna call the delegate directly and would ignore the tempTable special handling?
````
  public boolean dropPartition(String catName, String db_name, String tbl_name, List<String> part_vals,
      PartitionDropOptions options) throws NoSuchObjectException, MetaException, TException {
    return delegate.dropPartition(catName, db_name, tbl_name, part_vals, options);
  }
````

File: standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaHook.java (hunk 6 above)
Comment (wecharyu) [reply standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaHook.java:241]: Make sense.

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java (hunk 7 above)
Comment (wecharyu) [reply standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java:2124]: Make sense.

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/BaseMetaStoreClient.java (hunk 8 above)
Comment (wecharyu) [reply standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/BaseMetaStoreClient.java:564]: All derived client class could call `dropPartitions(... RequestPartsSpec...)` api for `dropPartitions(...List<Pair<> partExprs...)`, so I move it to the top `BaseMetaStoreClient`.

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java (hunk 12 above)
Comment (wecharyu) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java:1565]: All the `dropPartitions` APIs will call the new `dropPartitions(TableName...)` api, ideally it will not miss the tempTable logic for `SessionHiveMetaStoreClient`.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java:1565]: but in this case shouldn't we disable the old API as it could lead to incorrect behaviour?

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java [hunk 13]
```diff
@@ -1223,7 +1225,25 @@ public void preDropPartitions(org.apache.hadoop.hive.metastore.api.Table hmsTabl
...
+      List<DropPartitionsExpr> exprs = partsSpec.getExprs();
+      List<org.apache.commons.lang3.tuple.Pair<Integer, byte[]>> partExprs = Lists.newArrayList();
+      for (DropPartitionsExpr expr : exprs) {
+        partExprs.add(
+            org.apache.commons.lang3.tuple.Pair.of(expr.getPartArchiveLevel(), expr.getExpr()));
+      }
+      preDropPartitions(hmsTable, context, partExprs);
+    } else if (partsSpec.isSetNames()) {
+      preTruncateTable(hmsTable, context, partsSpec.getNames());
+      context.putToProperties(BaseMetaStoreClient.SKIP_DROP_PARTITION, "true");
```
deniskuzZ: should we set `SKIP_DROP_PARTITION` outside of if block?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4080,8 +4097,8 @@ public List<Partition> dropPartitions(String dbName, String tableName,
           .map(ss -> ss.getTxnMgr().getCurrentTxnId()).orElse(0L);
         dropOptions.setTxnId(txnId);
       }
-      List<org.apache.hadoop.hive.metastore.api.Partition> partitions = getMSC().dropPartitions(dbName, tableName,
-          partitionExpressions, dropOptions);
+      List<org.apache.hadoop.hive.metastore.api.Partition> partitions = getMSC().dropPartitions(
+          tableName, partsSpec, dropOptions, null);
```
deniskuzZ, line 4101: can we have a default method in API to avoid passing nulls?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -1980,6 +1980,7 @@ boolean dropPartition(String catName, String db_name, String tbl_name, List<Stri
 
   /**
    * Drop partitions based on an expression.
+   * @deprecated Use {@link #dropPartitions(TableName, RequestPartsSpec, PartitionDropOptions, EnvironmentContext)} instead
```
deniskuzZ: please add for all new deprecations:
````
@deprecated since 4.1.0, will be removed in 5.0.0
use {@link #dropPartitions(TableName, RequestPartsSpec, PartitionDropOptions, EnvironmentContext)} instead
````

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java [hunk 16]
```diff
@@ -2083,15 +2092,24 @@ List<Partition> dropPartitions(String dbName, String tblName,
...
 
+  /**
+   * @deprecated Use {@link #dropPartitions(TableName, RequestPartsSpec, PartitionDropOptions, EnvironmentContext)} instead
+   */
+  @Deprecated
   List<Partition> dropPartitions(String catName, String dbName, String tblName,
       List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options, EnvironmentContext context)
       throws NoSuchObjectException, MetaException, TException;
 
+  List<Partition> dropPartitions(TableName tableName,
```
deniskuzZ, line 2124: please add java doc

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/HookEnabledMetaStoreClient.java
```diff
@@ -325,6 +327,21 @@ public List<Partition> dropPartitions(String catName, String dbName, String tblN
     return delegate.dropPartitions(catName, dbName, tblName, partExprs, options, context);
   }
 
+  @Override
```
deniskuzZ: you've dropped the old signature in `SessionHiveMetaStoreClient`, but not here, why?
````
public List<Partition> dropPartitions(String catName, String dbName, String tblName,
      List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options, EnvironmentContext context)
````

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java (hunk 12 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java:1565]: 👍 now is good

Comment (deniskuzZ) [inline ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java:1565]: should we check the `SKIP_DROP_PARTITION` in all the clients, not only `ThriftHiveMetastoreClient`?

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/BaseMetaStoreClient.java [hunk 18]
```diff
@@ -557,6 +561,32 @@ public final List<Partition> dropPartitions(String catName, String dbName, Strin
...
+      List<Pair<Integer, byte[]>> partExprs, PartitionDropOptions options, EnvironmentContext context)
+      throws NoSuchObjectException, MetaException, TException {
+    if (context == null) {
+      context = new EnvironmentContext();
+    }
+
+    if (context.getProperties() != null &&
+        Boolean.parseBoolean(context.getProperties().get(SKIP_DROP_PARTITION))) {
```
deniskuzZ: Should we check the `SKIP_DROP_PARTITION` in all the clients NEW signature, not only `ThriftHiveMetastoreClient`?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaHook.java
```diff
@@ -217,13 +218,27 @@ default void postGetTable(Table table) {
 
   /**
    * Called before dropping the partitions from the table in the metastore during ALTER TABLE DROP PARTITION.
+   * @deprecated Use {@link #preDropPartitions(Table, EnvironmentContext, RequestPartsSpec)} instead.
```
deniskuzZ: same, see https://github.com/apache/hive/pull/5223/files#r2293388430

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
wecharyu: The original `preDropPartition()` for `partExprs` will also set `SKIP_DROP_PARTITION`, it's OK to move it out and remove the set operation in deprecated original `preDropPartition()`.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
wecharyu: Do we need a new API for it? There seems already so many APIs for `dropPartitions`.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/HookEnabledMetaStoreClient.java
wecharyu: Nice catch, I want to preserve the use of `hook.preDropPartitions(table, context, partExprs);` at the beginning. Actually it could be removed here too.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/BaseMetaStoreClient.java
wecharyu: Currently all clients will use the `ThriftHiveMetastoreClient` as the delegate, so it's OK to just check `SKIP_DROP_PARTITION` in `ThriftHiveMetastoreClient`.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
deniskuzZ: +1

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
deniskuzZ: not sure, however, passing NULLs isn't great either

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/BaseMetaStoreClient.java
deniskuzZ: @wecharyu,  in that case, should we drop this check from the `BaseMetastoreClient` and keep it just in `ThriftHiveMetastoreClient` to be consistent with the old API?

wecharyu: Addressed.

PR #5224, opened 2024-04-29
HIVE-28228: Remove redundant/unused sources maven profile

zabetak:
Simplify build and reduce maintenance overhead

The sources maven profile was introduced in HIVE-5717 among others to package source code into a .jar and publish it to maven. However, after adding Apache pom as parent in HIVE-6608 the sources profile is now redundant; it is subsumed by the apache-release profile.

In some cases, the sources profile allows to package the test sources in a jar file (HIVE-13206) but this profile is not used during the release so the generated test-sources.jars are not consumed by anyone. This seems like an unused dev only feature thus it can be removed.

Devs will no longer have the option to generate test-sources.jar using the `sources` profile.

No

N/A

### DIFF
--- a/beeline/pom.xml
+++ b/beeline/pom.xml
@@ -197,27 +197,6 @@
       <optional>true</optional>
     </dependency>
   </dependencies>
-  <profiles>
-    <profile>
-      <id>sources</id>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-source-plugin</artifactId>
-            <executions>
-              <execution>
-                <id>attach-sources</id>
-                <goals>
-                  <goal>test-jar</goal>
-                </goals>
-              </execution>
-            </executions>
-          </plugin>
-        </plugins>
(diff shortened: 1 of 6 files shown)


ayushtkn:
there is a use case mentioned:
https://issues.apache.org/jira/browse/HIVE-13206?focusedCommentId=15183873&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15183873

that will get messed up, we do release the src tar, so somebody can build and get the test jar for their use case

zabetak:
@ayushtkn I consider the use-case in HIVE-13206 as a "dev only" feature and not a very useful one. On the other hand this "feature" adds confusion and maintenance overhead (see exchanges in HIVE-28024).

zabetak:
Hey @ayushtkn , did you see my latest comment? If you feel that this profile is useful let me know and I will close the JIRA and PR.

ayushtkn:
hey @zabetak 
no strong opinions here, if others are convinced, I don't pose any objections

PR #5226, opened 2024-04-30
HIVE-28235 : Hive should honour heapsize set in hive-env

AnmolSun:
Hive processes do not honour the heap size set in hive-env. Right now, they only consider the value set in hadoop-env and ignore those set in hive-env.  

Bug - because values related to Heap set in hive-env is ignored. 

No

No

With the updated script, heap values set in hive-env is honoured over hadoop-env.


Aggarwal-Raghav:
+1 (non-binding), have reproduced the issue in local setup and tested the fix as well.

PR #5228, opened 2024-04-30
HIVE-28237: Add a default tez-site.xml to data/conf

abstractdog:
Added a tez-site.xml file to data/conf: copied from data/conf/tez and added a typical test scenario setting as below:
```
  <property>
    <name>tez.local.mode</name>
    <value>true</value>
  </property>
```

As described on jira, users of StartMiniHS2Cluster might find it confusing where to set tez configs, not even defined without this patch.

No.

No.

With StartMiniHS2Cluster, tez.local.mode=true was kicking in with the patch, this log line below was found in the log file of StartMiniHS2Cluster, so it created a LocalClient as expected
```
2024-04-30T02:55:14,168  INFO [DAGAppMaster Thread] client.LocalClient: Using working directory: /tmp/hive/scratch/laszlobodor/_tez_session_dir/7c9b847c-9d48-469a-a1e7-fa4d29b3148f/.tez/application_1714470913855_0001_wd
```

### DIFF
--- a/data/conf/tez-site.xml
+++ b/data/conf/tez-site.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0"?>
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


aturoczy:
Is it a copy or a move?

abstractdog:
> Is it a copy or a move? As a copy won't be better to reference to the original path? Just for the sake of file duplication.

copy + added an extra opt, I can experiment with adding it to the original file and making a symlink
in the mean I have to handle test failures

abstractdog:
cannot reopnen, moving to https://github.com/apache/hive/pull/5348

PR #5229, opened 2024-04-30
HIVE-28214: HPLSQL not using the hive variables passed through beeline using --hivevar option

mdayakar:
HIVE-28214: HPLSQL not using the hive variables passed through beeline using --hivevar option

Hive variables passed using --hivevar option through beeline are not getting used in HPLSQL procedures/cursors. So changed code in such a way that they uses the hive variables passed through beeline.

To support hive variables passed through the beeline in HPLSQL.

No

No


kasakrisz (approved):
LGTM

PR #5230, opened 2024-04-30
HIVE-28240: Compaction requests remain in Working state after Timeout…

difin:
…Exception

When a compaction worker receive TimeoutException during execution of a compaction request, it attempts to cancel a running compaction by calling cancel on a compaction's Future object and by killing and recreating a compaction executor. However, we got a case on a downstream code where calling the cancel method on the compaction's future object didn't stop the compaction requests and they remained in the Working status. These changes propose marking compaction request as Failed to ensure that it doesn't stay in Working status.

To ensure that a compaction request doesn't stay in Working status after TimeoutException.

No

Pre-commit tests.


InvisibleProgrammer:
Hi, could you please cover that with a test?

deniskuzZ:
sorry, I didn't notice the backoff and retry. we shouldn't just mark as aborted

PR #5231, opened 2024-05-02
HIVE-28215: Signalling CONDITION HANDLER is not working in HPLSQL.

mdayakar:
HIVE-28215: Signalling CONDITION HANDLER is not working in HPLSQL.

Currently the CONDITION HANDLERs defined by user are not getting invoked when the signal is given to the corresponding condition. As a part of processing exception/error handling conditions, user defined conditions also processed (just looging into the log file) so when it has to invoke user defined condition handlers there are no condition handlers in the stack.

To fix the issue of invoking user defined condition handlers properly.

No

No

### DIFF
--- a/hplsql/src/main/java/org/apache/hive/hplsql/Exec.java
+++ b/hplsql/src/main/java/org/apache/hive/hplsql/Exec.java
@@ -1049,10 +1049,8 @@ public class Exec extends HplsqlBaseVisitor<Integer> implements Closeable {
       }
     }
     // if there are any user defined signals then push them back to signals stack to handle them later.
-    if (userDefinedSignals.size() > 0) {
-      for (int i = userDefinedSignals.size() - 1; i >= 0; i--) {
-        exec.signals.push(userDefinedSignals.get(i));
-      }
+    for (int i = userDefinedSignals.size() - 1; i >= 0; i--) {
+      exec.signals.push(userDefinedSignals.get(i));
     }
   } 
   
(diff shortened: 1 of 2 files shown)


hplsql/src/main/java/org/apache/hive/hplsql/Exec.java
```diff
@@ -1044,6 +1048,12 @@ public void printExceptions() {
         trace(null, "Signal: " + sig.type);
       }
     }
+    // if there are any user defined signals then push them back to signals stack to handle them later.
+    if (userDefinedSignals.size() > 0) {
+      for (int i = userDefinedSignals.size() - 1; i >= 0; i--) {
```
kasakrisz, line 1053: Is this check necessary? If `userDefinedSignals` is empty the loop doesn't executed.

hplsql/src/main/java/org/apache/hive/hplsql/Exec.java
```diff
@@ -1028,9 +1028,13 @@ void cleanup() {
    * Output information about unhandled exceptions
    */
   public void printExceptions() {
+    List<Signal> userDefinedSignals = new ArrayList<>();
```
kasakrisz, line 1031: Is it allowed to add the same Signal more than one times?
Does the order of `Signals` matter?

hplsql/src/main/java/org/apache/hive/hplsql/Exec.java
mdayakar: Fixed

hplsql/src/main/java/org/apache/hive/hplsql/Exec.java
mdayakar: > Is it allowed to add the same Signal more than one times?
As per code, same signal is not getting added more than once.

> Does the order of Signals matter?
Here to store the signals Stack data structure is getting used so might be the order matters thats why pushing the userdefined signals in the same order which they were present in Stack before processing other signals.

PR #5232, opened 2024-05-02
HIVE-28208: WITH column list doesn't work with CTE materialization

okumin:

ANSI SQL supports column aliases for CTEs. Hive also supports it but it fails when CTEs are materialized.

https://issues.apache.org/jira/browse/HIVE-28208

This patch allows materialized CTEs to obey the standard.

No. The affected queries never succeed now.

No.

I added test queries to `cte_mat_1.q`.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
@@ -32,6 +32,7 @@ import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.DEFAU
 import static org.apache.hadoop.hive.ql.ddl.view.create.AbstractCreateViewAnalyzer.validateTablesUsed;
 import static org.apache.hadoop.hive.ql.optimizer.calcite.translator.ASTConverter.NON_FK_FILTERED;
 
+import com.google.common.base.Preconditions;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.security.AccessControlException;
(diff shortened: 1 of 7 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -8379,7 +8402,9 @@ private ColsAndTypes deriveFileSinkColTypes(
       String colName = colInfo.getInternalName();  //default column name
       if (columns != null) {
         FieldSchema col = new FieldSchema();
-        if (!("".equals(nm[0])) && nm[1] != null) {
+        if (i < withColList.size()) {
```
okumin, line 8405: Using a random name when <with column list> is shorter than the input. This follows the behavior of non-materialized CTEs.
If I understand correctly, with-column-list is supported only when the number of columns is exactly equal to that of the query expression. So, it could be an option to make a query fail if `colInfos.size() != withColList.size()`.

ql/src/test/queries/clientpositive/cte_mat_1.q
```diff
@@ -1,10 +1,37 @@
...
 explain
 with q1(srcKey, srcValue) as (select * from src where key= '5')
+select a.srcKey, b.srcValue
+from q1 a join q1 b
+on a.srcKey=b.srcKey;
+
+set hive.optimize.cte.materialize.threshold=2;
+set hive.optimize.cte.materialize.full.aggregate.only=false;
+-- Use a format that retains column names
+set hive.default.fileformat=parquet;
```
okumin, line 14: The default one positionally accesses physical columns, and we can't find some kinds of errors.

deniskuzZ:
cc @zabetak

PR #5233, opened 2024-05-02
HIVE-28209: Allow the use of a proxy user

snmvaughan:

Allow delegation for HiveMetaStore connections

The log message indicates that `HADOOP_PROXY_USER` is set, but then doesn't allow its use by calling `getRealUser()` which will be `null`.  This change allows delegation for HiveMetaStore connections by using either the real user or self.  The code for this is coming from Hadoop 3.3.6.

Allows proxying instead of throwing an exception.

No

Local builds which have removed `getRealUser()` to allow the use of delegation work as expected.  Running the existing code always fail because `getRealUser()` will always return `null`.


prasanthj:
+1

prasanthj (approved):

prasanthj:
@snmvaughan Thanks for the contrib! Merged to master.

PR #5234, opened 2024-05-03
HIVE-28244: Add SBOM for storage-api and standalone-metastore modules

Aggarwal-Raghav:
-Pdist profile present in parent pom.xml doesn't come into effect for storage-api module and standalone-metastore module.

Publish SBOM for standalone-metastore and storage-api module

NO

NO

Tested via my local .m2 repository cache. After the fix, it is creating the "-cyclonedx.json" and "-cyclonedx.xml" for storage-api and standalone-metastore.

### DIFF
--- a/standalone-metastore/pom.xml
+++ b/standalone-metastore/pom.xml
@@ -40,6 +40,7 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <maven.compiler.useIncrementalCompilation>false</maven.compiler.useIncrementalCompilation>
+    <maven.cyclonedx.plugin.version>2.7.10</maven.cyclonedx.plugin.version>
     <maven.repo.local>${settings.localRepository}</maven.repo.local>
     <maven.exec.plugin.version>3.1.0</maven.exec.plugin.version>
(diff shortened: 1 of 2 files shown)


Aggarwal-Raghav:
@zabetak, can you please help with review?

Aggarwal-Raghav:
@deniskuzZ thanks for the review,

In branch-3, the version of storage-api is 2.6.1 (in my opinion, it is wrong version, it should be 3.2.0 now) and,
In branch-3.1, the storage api module is not getting build as it not present in the module list in pom.xml (not sure about the reason, why it was removed)https://github.com/apache/hive/blob/bdf3285bc250609c1b31a52e7732be8c043b8817/pom.xml#L34, instead we are using 2.7.0 version and getting from artifactory, https://github.com/apache/hive/blob/bdf3285bc250609c1b31a52e7732be8c043b8817/pom.xml#L198

That's why storage-api jars for 3.x are missing in [artifactory](https://repo1.maven.org/maven2/org/apache/hive/hive-storage-api/) 
**For hive 4.x, this has been fixed and it is generating the artifacts.**

Not entirely sure about the usage of storage-api module, please refer to this [HIVE-10171](https://issues.apache.org/jira/browse/HIVE-10171)

PR #5236, opened 2024-05-05
[hotfix][test] Initializing MiniDFSCluster avoid using deprecated methods.

slfan1989:

I have been recently tracking the issue of upgrading Hadoop from 3.3.6 to 3.4.0. During this process, I identified an improvement point: we are using deprecated methods while initializing MiniDFSCluster. Currently, four unit tests utilize deprecated methods, while others use the recommended ones. Instead of creating a JIRA ticket, I am attempting to fix it directly using a hotfix approach.

It is better to use the recommended method.

No.

No.

This pr is upgrading the code for unit tests, and the upgraded code won't introduce any other issues.

### DIFF
--- a/hbase-handler/src/test/org/apache/hadoop/hive/hbase/TestHBaseSerDe.java
+++ b/hbase-handler/src/test/org/apache/hadoop/hive/hbase/TestHBaseSerDe.java
@@ -1273,7 +1273,8 @@ public class TestHBaseSerDe {
 
     try {
       // MiniDFSCluster litters files and folders all over the place.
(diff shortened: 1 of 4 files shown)


zhangbutao:
@slfan1989 IMO, though this is a hotfix , we still need a JIRA to track this change. :)

slfan1989:
> @slfan1989 IMO, though this is a hotfix , we still need a JIRA to track this change. :)

@zhangbutao  Thanks for your suggestion! I will create a JIRA to follow this change.

PR #5239, opened 2024-05-06
HIVE-28247: Execute immediate `select count(*) from tbl` throwing ClassCastException in hplsql mode.

mdayakar:
HIVE-28247: Execute immediate `select count(*) from tbl` throwing ClassCastException in hplsql mode.

select count(*) query returns a long value but hplsql is expecting a string type value as result ClassCastException is coming. If expected type is String then converting result value to string and returning now.

To support select count(*) query using execute immediate statement.

No

No


kasakrisz:
> select count(*) query returns a long value but hplsql is expecting a string type value

Why is it expecting a string type value ?

mdayakar:
> > select count(*) query returns a long value but hplsql is expecting a string type value
> 
> Why is it expecting a string type value ?

In HPLSQL, when the query gets executed through execute/execute immediate statement then the query directly will gets executed by HS2 and the results will be processed in HPLSQL. I think here they are converting to string to display/print to the client in string format (though not exactly sure the exact reason). This issue is not present when the query select count(*) is executed directly, issue is there only with execute/execute immediate.

kasakrisz (commented):
+1

PR #5243, opened 2024-05-10
Hive-28251: HiveSessionImpl init ReaderStream set Charset with UTF-8

xuzifu666:
Fix some StreamReader not set with UTF8，if we actually default charset not support Chinese chars such as latin and conf contain Chinese chars，it would not resolve success，so we need set it as utf8 in StreamReader，we can find all StreamReader with utf8 charset in other compute framework，such as Calcite、Hudi and so on.

HiveSessionImpl init ReaderStream set Charset with UTF-8

No

not need

### DIFF
--- a/service/src/java/org/apache/hive/service/cli/session/HiveSessionImpl.java
+++ b/service/src/java/org/apache/hive/service/cli/session/HiveSessionImpl.java
@@ -18,6 +18,7 @@
 
 package org.apache.hive.service.cli.session;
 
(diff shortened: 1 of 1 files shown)


xuzifu666:
@ayushtkn Could you give a review or close the pr please? Thanks

ayushtkn (approved):
LGTM

xuzifu666:
@ayushtkn Thanks，could you help to merge it？

PR #5244, opened 2024-05-10
HIVE-28252: AssertionError when using HiveTableScan with a HepPlanner cluster

zabetak:
1. Relax the assertion to not throw an error when the cluster does not contain a `ConventionTraitDef` (`HepPlanner`).
2. Add tests to ensure that it is possible to create a `HiveTableScan` with `VolcanoPlanner` and `HepPlanner` and the expected convention is returned.

The HepPlanner is heavily used by the optimizer so it makes to be able to create a HiveTableScan with its respective cluster without residing in workarounds (such as requiring the creation of a VolcanoPlanner and obtaining the cluster from there).

More details under HIVE-28252.

No, anyways assertions are not enabled in production code.

No

```
mvn test -Dtest=TestHiveTableScan
```


kasakrisz (approved):
+1, LGTM

PR #5245, opened 2024-05-12
HIVE-28254: CBO (Calcite Return Path): Multiple DISTINCT leads to wrong results

okumin:

This PR would resolve the issue that causes a UDAF accesses a wrong key.

https://issues.apache.org/jira/browse/HIVE-28254

Bug fix. Without this fix, #5091 would generate wrong results.

Literally no. Affected queries are not generating correct results.

No.

Added a qtest.


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/opconventer/HiveGBOpConvUtil.java
```diff
@@ -1028,32 +1028,65 @@ private static OpAttr genReduceSideGB1NoMapGB(OpAttr inputOpAf, GBInfo gbInfo,
...
+
+        final String rsUDAFParamName;
+        if (distinctColumnNameTable.get(i).get(j) != null) {
+          rsUDAFParamName = distinctColumnNameTable.get(i).get(j);
+        } else if (distinctColumnMapping.get(rsUDAFParamColInfo) != null) {
+          // This UDAF is not labeled with DISTINCT, but it refers to a DISTINCT key.
+          // The original internal name is already obsolete as any DISTINCT keys are renamed.
+          rsUDAFParamName = distinctColumnMapping.get(rsUDAFParamColInfo);
+        } else {
+          rsUDAFParamName = rsUDAFParamColInfo.getInternalName();
```
okumin, line 1085: As far as I tested, `getInternalName` retains `KEY.col0:{index of RS keys}.col0`. It conflicts with the name generated by the following logic, and a non-distinct UDAF can refer to a wrong column.

```
          distinctColumnName = Utilities.ReduceField.KEY.name() + "." + lastReduceKeyColName
                  + ":" + numDistinctUDFs + "." + SemanticAnalyzer.getColumnInternalName(j);
```

Non-Return Path resolves the correct column name by letting RowResolver resolve the correct name before creating `ExprNodeColumnDesc`. This PR would implement the logic by traversing params with two paths.
Why do we need the two paths? That's because the non-distinct UDAF may refer to the input of DISTINCT UDAF which comes later like below.

```
SELECT
  SUM(DISTINCT col1),
  COUNT(DISTINCT col1),
  SUM(col2), -- This has to refer to the key for `SUM(DISTINCT col2)`
  MAX(DISTINCT col1),
  SUM(DISTINCT col2),
  MIN(DISTINCT col1)
FROM test;
```

Note that this method is used only when `hive.map.aggr` is disabled. With it enabled, non-distinct UDAFs don't refer to DISTINCT keys as they are deduplicated.

ql/src/test/results/clientpositive/llap/cbo_rp_groupby3_noskew_multi_distinct.q.out
```diff
@@ -185,3 +185,118 @@ POSTHOOK: type: QUERY
...
+                      null sort order: zz
+                      sort order: ++
+                      Statistics: Num rows: 4 Data size: 32 Basic stats: COMPLETE Column stats: COMPLETE
+            Execution mode: vectorized, llap
+            LLAP IO: all inputs
+        Reducer 2 
+            Execution mode: llap
+            Reduce Operator Tree:
+              Group By Operator
+                aggregations: sum(DISTINCT KEY._col0:0._col0), count(DISTINCT KEY._col0:1._col0), sum(KEY._col0:3._col0), max(DISTINCT KEY._col0:2._col0), sum(DISTINCT KEY._col0:3._col0), min(DISTINCT KEY._col0:4._col0)
```
okumin, line 262: Without this patch, the third UDAF is not `sum(KEY._col0:3._col0)` but `sum(KEY._col0:1._col0)`, which means it reads a different column.
https://github.com/apache/hive/pull/5245/commits/3681775c9e183208fcde5d1d429fb5cf31fe81be#diff-c0641a94a4e5e0c28830fcdccd252506a0dc7842a24aa414bf096a229e8e4235R262

With `set hive.cbo.returnpath.hiveop=false;`, Hive generates the same plan as that in q.out.

ql/src/test/results/clientpositive/llap/count.q.out
```diff
@@ -811,7 +811,7 @@ STAGE PLANS:
             Execution mode: llap
             Reduce Operator Tree:
               Group By Operator
-                aggregations: count(DISTINCT KEY._col1:0._col0), count(DISTINCT KEY._col1:1._col0), sum(VALUE._col0), sum(VALUE._col1), sum(VALUE._col2), sum(KEY._col1:0._col0), sum(KEY._col1:1._col0), sum(KEY._col0), sum(DISTINCT KEY._col1:2._col0), sum(DISTINCT KEY._col1:3._col0)
+                aggregations: count(DISTINCT KEY._col1:0._col0), count(DISTINCT KEY._col1:1._col0), sum(VALUE._col0), sum(VALUE._col1), sum(VALUE._col2), sum(KEY._col1:0._col0), sum(KEY._col1:1._col0), sum(KEY._col1:2._col0), sum(DISTINCT KEY._col1:2._col0), sum(DISTINCT KEY._col1:3._col0)
```
okumin, line 814: In this case, the first `KEY._col0` doesn't conflict generated columns for DISTINCT UDAFs. So, technically, the problem would not happen without the change of this PR, which mean either seems to be acceptable.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/opconventer/HiveGBOpConvUtil.java
kasakrisz: @okumin 
I'm not familiar with the logic of distinct key generation. IIUC in this case some of these keys are conflicting with non-distinct keys.

Could you please elaborate distinct key generation logic?
Why can we construct keys like this
```
distinctColumnName = Utilities.ReduceField.KEY.name() + "." + lastReduceKeyColName + ":" + numDistinctUDFs + "." + SemanticAnalyzer.getColumnInternalName(j);
```
I was checking the legacy non-cbo return patch code, it also has this
https://github.com/apache/hive/blob/a1420ed816c315d98be7ebf05cdc3ba139a68643/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java#L5372-L5379
but I still don't understand why can we reference columns which are not coming any schema or what is the schema which contains these generated keys?

ql/src/test/results/clientpositive/llap/count.q.out
okumin: `KEY._col0` had a special name because it is not only a distinct param but also an aggregation param. So, it is accessed with two types of names in GroupByOperator.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/opconventer/HiveGBOpConvUtil.java
kasakrisz: Thanks for explanation.

okumin: BTW, HIVE-474 and HIVE-537 explains why RSO + GBO use UNION to express multiple DISTINCTs.

> Why GroupByOperator needs to rename the source column for UNION?

[Reading the primitive change](https://github.com/apache/hive/commit/ed3f0b1983edcdfdc2dbd2d3862d385cbb5c04ae#diff-7d3ff1b241592b075dc3b4afcc01d068e58799c810fc5085815ad6bf8ef5e6a9R609), I guess the operator uses `position of UDAF` of `KEY.{the final output col name}:{position of UDAF}.{position of arguments of the UDAF}` in order to identify if it needs to be DISTINCTed or not while being processed.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/opconventer/HiveGBOpConvUtil.java
```diff
@@ -1028,32 +1028,65 @@ private static OpAttr genReduceSideGB1NoMapGB(OpAttr inputOpAf, GBInfo gbInfo,
...
+    for (int i = 0; i < paramColInfoTable.size(); i++) {
+      final ArrayList<ExprNodeDesc> aggParameters = new ArrayList<>();
+
+      for (int j = 0; j < paramColInfoTable.get(i).size(); j++) {
+        final ColumnInfo rsUDAFParamColInfo = paramColInfoTable.get(i).get(j);
+
+        final String rsUDAFParamName;
+        if (distinctColumnNameTable.get(i).get(j) != null) {
+          rsUDAFParamName = distinctColumnNameTable.get(i).get(j);
+        } else if (distinctColumnMapping.get(rsUDAFParamColInfo) != null) {
```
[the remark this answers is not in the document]
okumin: Done. I merged the latest master and triggered CI once more.
https://github.com/apache/hive/pull/5245/commits/902e65824a414867cb77171b0e2f36fdc8fe180a

PR #5246, opened 2024-05-12
HIVE-28239: Fix bug on HMSHandler#checkLimitNumberOfPartitions

wecharyu:
Fix the bug on `HMSHandler#checkLimitNumberOfPartitions`:
1. use maxParts to determine if need check partition limit
2. remove maxParts on `checkLimitNumberOfPartitions`

Fix the bug, assume that HMS configure `metastore.limit.partition.request` as 100, the client calls `get_partitions_by_filter` with maxParts as 101, and the matching partition number is 50, in this case the HMS server should not throw `MetaException` by partition limit check.

No.

No.

Fix some unit tests.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -5563,7 +5563,7 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
   private void checkLimitNumberOfPartitionsByFilter(String catName, String dbName,
                                                     String tblName, String filterString,
                                                     int maxParts) throws TException {
-    if (needCheckPartitionLimit(maxParts)) {
+    if (exceedsPartitionLimit(maxParts)) {
       checkLimitNumberOfPartitions(tblName, get_num_partitions_by_filter(prependCatalogToDbName(
           catName, dbName, conf), tblName, filterString));
     }
@@ -5572,23 +5572,23 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
   private void checkLimitNumberOfPartitionsByPs(String catName, String dbName, String tblName,
                                                 List<String> partVals, int maxParts)
           throws TException {
-    if (needCheckPartitionLimit(maxParts)) {
+    if (exceedsPartitionLimit(maxParts)) {
       checkLimitNumberOfPartitions(tblName, getNumPartitionsByPs(catName, dbName, tblName,
               partVals));
     }
   }
 
-  // Check request partition limit iff:
+  // Check input size exceeding partition limit iff:
   //  1. partition limit is enabled.
-  //  2. request size is greater than the limit.
-  private boolean needCheckPartitionLimit(int requestSize) {
+  //  2. input size is greater than the limit.
+  private boolean exceedsPartitionLimit(int size) {
     int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
-    return partitionLimit > -1 && (requestSize < 0 || requestSize > partitionLimit);
+    return partitionLimit > -1 && (size < 0 || size > partitionLimit);
   }
 
   private void checkLimitNumberOfPartitions(String tblName, int numPartitions) throws MetaException {
-    int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
-    if (partitionLimit > -1 && numPartitions > partitionLimit) {
+    if (exceedsPartitionLimit(numPartitions)) {
+      int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
       String configName = ConfVars.LIMIT_PARTITION_REQUEST.toString();
       throw new MetaException(String.format(PARTITION_NUMBER_EXCEED_LIMIT_MSG, numPartitions,
           tblName, partitionLimit, configName));
@@ -7217,7 +7217,7 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     RawStore rs = getMS();
     try {
       authorizeTableForPartitionMetadata(catName, dbName, tblName);
-      if (needCheckPartitionLimit(args.getMax())) {
+      if (exceedsPartitionLimit(args.getMax())) {
         // Since partition limit is configured, we need fetch at most (limit + 1) partition names
         int max = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST) + 1;
         args = new GetPartitionsArgs.GetPartitionsArgsBuilder(args).max(max).build();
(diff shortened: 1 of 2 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -5591,15 +5586,12 @@ private boolean needCheckPartitionLimit(int requestSize) {
...
-      int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
-      int partitionRequest = (maxToFetch < 0) ? numPartitions : maxToFetch;
-      if (partitionRequest > partitionLimit) {
-        String configName = ConfVars.LIMIT_PARTITION_REQUEST.toString();
-        throw new MetaException(String.format(PARTITION_NUMBER_EXCEED_LIMIT_MSG, partitionRequest,
-            tblName, partitionLimit, configName));
-      }
+  private void checkLimitNumberOfPartitions(String tblName, int numPartitions) throws MetaException {
+    int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
+    if (partitionLimit > -1 && numPartitions > partitionLimit) {
```
deniskuzZ, line 5591: " > 0" ??

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -5563,26 +5563,21 @@ public List<Partition> get_partitions_with_auth(final String dbName,
   private void checkLimitNumberOfPartitionsByFilter(String catName, String dbName,
                                                     String tblName, String filterString,
                                                     int maxParts) throws TException {
-    if (isPartitionLimitEnabled()) {
+    if (needCheckPartitionLimit(maxParts)) {
```
deniskuzZ, line 5566: why do we need `needCheckPartitionLimit` since we have the same checks in `checkLimitNumberOfPartitions`

wecharyu: It's different. `needCheckPartitionLimit` use maxParts to determine if need run the `checkLimitNumberOfPartitions`, while `checkLimitNumberOfPartitions` checks if  the matching `numPartitions` is greater than limit.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: Here we followed the existing code in `isPartitionLimitEnabled()`, do we need change it?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
deniskuzZ: 1. we execute the same function before and after db fetch - 'needCheckPartitionLimit', maybe we could rename and reuse 
2. also why we check count and fail post db call and not limit cound inside of it?

deniskuzZ:
````
  private void checkLimitNumberOfPartitionsByPs(String catName, String dbName, String tblName,
                                                List<String> partVals, int maxParts)
          throws TException {
    if (needCheckPartitionLimit(maxParts)) {
      checkLimitNumberOfPartitions(tblName, getNumPartitionsByPs(catName, dbName, tblName,
              partVals));
    }
  }

  private boolean needCheckPartitionLimit(int requestSize) {
    int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
    return partitionLimit > -1 && (requestSize < 0 || requestSize > partitionLimit);
  }

  private void checkLimitNumberOfPartitions(String tblName, int numPartitions) throws MetaException {
    int partitionLimit = MetastoreConf.getIntVar(conf, ConfVars.LIMIT_PARTITION_REQUEST);
    if (partitionLimit > -1 && numPartitions > partitionLimit) {
      String configName = ConfVars.LIMIT_PARTITION_REQUEST.toString();
      throw new MetaException(String.format(PARTITION_NUMBER_EXCEED_LIMIT_MSG, numPartitions,
          tblName, partitionLimit, configName));
    }
  }
````
where do we limit the fetch size? we throws exception if returned `numPartitions` > `partitionLimit`?

Comment (wecharyu) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java:5566]: 1. `needCheckPartitionLimit` is used to determine if needs to check liimt.
let's say that a request size is 100, LIMIT is 50, then the returned partition size may hit limit, so we need check limit;
Another request size is 30, LIMIT is 50, then returned partition size is never greater than limit, so we do not need check limit.
Let's rename it to `exceedsPartitionLimit(int size)` to reuse it.
2. The existing behavior is throw exception when hit partition limit, I think it's reasonable because returning the limit size partitions is confused whether there are other partitions matching my condition and it maybe not the expected result to users.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: 1. sounds good
2. makes sense, but do we return an error back to the client saying that the configured partition fetch limit was exceeded or just log an exception?

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java [hunk 3]
```diff
@@ -5563,43 +5563,35 @@ public List<Partition> get_partitions_with_auth(final String dbName,
...
-          catName, dbName, conf), tblName, filterString), maxParts);
+          catName, dbName, conf), tblName, filterString));
     }
   }
 
   private void checkLimitNumberOfPartitionsByPs(String catName, String dbName, String tblName,
                                                 List<String> partVals, int maxParts)
           throws TException {
-    if (isPartitionLimitEnabled()) {
+    if (exceedsPartitionLimit(maxParts)) {
```
deniskuzZ, line 5575: why it's called `maxPart` and not 'reqParts' (requested partitions count)?

deniskuzZ (dismissed):
LGTM +1

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -5563,43 +5563,35 @@ public List<Partition> get_partitions_with_auth(final String dbName,
...
-    return partitionLimit > -1;
-  }
-
-  // Check request partition limit iff:
+  // Check input size exceeding partition limit iff:
   //  1. partition limit is enabled.
-  //  2. request size is greater than the limit.
-  private boolean needCheckPartitionLimit(int requestSize) {
+  //  2. input size is greater than the limit.
+  private boolean exceedsPartitionLimit(int size) {
```
deniskuzZ, line 5584: maybe. `exceedsPartitionFetchLimit`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -5563,43 +5563,35 @@ public List<Partition> get_partitions_with_auth(final String dbName,
   private void checkLimitNumberOfPartitionsByFilter(String catName, String dbName,
                                                     String tblName, String filterString,
                                                     int maxParts) throws TException {
-    if (isPartitionLimitEnabled()) {
+    if (exceedsPartitionLimit(maxParts)) {
```
deniskuzZ, line 5566: should it be `requestMax`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
deniskuzZ, line 5575: same `requestMax`

wecharyu: Addressed.

deniskuzZ (approved):
LGTM +1, pending tests

PR #5250, opened 2024-05-14
HIVE-28260: CreateTableEvent wrongly skips authorizing DFS_URI for managed table.

ayushtkn:

Correct the check to restore proper behaviour 

The check is wrong, the string needs to be compared using equals

Yes, the URI will be validated during create for managed table

No

👀


aturoczy (approved):

SourabhBadhya (approved):
LGTM +1

PR #5215, opened 2024-04-25
HIVE-28268: Iceberg: Retrieve row count from iceberg SnapshotSummary in case of iceberg.hive.keep.stats=false

zhangbutao:
### What changes were proposed in this pull request?

At present, in case of  `iceberg.hive.keep.stats=true` & `hive.compute.query.using.stats=true`, HS2 will do a fetch task to get iceberg table's `numRows `property from HMS  to optimize `count `query.
If `iceberg.hive.keep.stats=false`, HS2 will always launch tez task to compute table's row count when filing a `count `query.

However, as we know, iceberg table's metadata has some stats information, we can also just start a fetch task to retrieve the row count from iceberg's snapshot summary when `iceberg.hive.keep.stats=false` or no stats stored in hms.  This can avoid launching tez task to compute the table's row count.

BTW, timetravel or branch/tag has different stats from current snapshot, so we need to get the specified snapshotid based on the different iceberg version. Otherwise, we will get the wrong stats when querying the time travel/branch/tag.

### Does this PR introduce _any_ user-facing change?

No
### Is the change a dependency upgrade?

No

### How was this patch tested?

Qtest

PR #5253, opened 2024-05-16
HIVE-28266: Iceberg: select count(*) from data_files metadata tables …

difin:
…gives wrong result

Modified Iceberg method "canComputeQueryUsingStats" to return false for queries over metadata tables to make Hive execute query over metadata table instead of getting the result from statistics.

Presently, when running a SELECT COUNT(*) query over an Iceberg table X.data_files where X is a data table, the result returns number of records in X rather than in X.data_files.

No

No

New query test added

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -570,7 +570,7 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
 
   @Override
   public boolean canComputeQueryUsingStats(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
-    if (getStatsSource().equals(HiveMetaHook.ICEBERG)) {
+    if (getStatsSource().equals(HiveMetaHook.ICEBERG) && hmsTable.getMetaTable() == null) {
       Table table = getTable(hmsTable);
       if (table.currentSnapshot() != null) {
         Map<String, String> summary = table.currentSnapshot().summary();
@@ -1032,7 +1032,7 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
 
   @Override
   public boolean isValidMetadataTable(String metaTableName) {
-    return IcebergMetadataTables.isValidMetaTable(metaTableName);
+    return metaTableName != null && IcebergMetadataTables.isValidMetaTable(metaTableName);
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -570,7 +570,7 @@ private ColumnStatistics readColStats(Table table, Path statsPath) {
 
   @Override
   public boolean canComputeQueryUsingStats(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
-    if (getStatsSource().equals(HiveMetaHook.ICEBERG)) {
+    if (getStatsSource().equals(HiveMetaHook.ICEBERG) && hmsTable.getProperty("metaTable") != null) {
```
deniskuzZ, line 573: do we have constant for "metaTable"

difin: Created a constant and refactored other places to use it instead of the hardcoded string.

zhangbutao (approved):
+1 LGTM. Pending tests.

I also think there are some other similar places which query iceberg metadata tables but using the data table's statistics wrongly. We can fix them incrementally. Like https://github.com/apache/hive/pull/5215 which i am doing.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1512,7 +1512,8 @@ private String collectColumnAndReplaceDummyValues(ExprNodeDesc node, String foun
   private void fallbackToNonVectorizedModeBasedOnProperties(Properties tableProps) {
     Schema tableSchema = SchemaParser.fromJson(tableProps.getProperty(InputFormatConfig.TABLE_SCHEMA));
     if (FileFormat.AVRO.name().equalsIgnoreCase(tableProps.getProperty(TableProperties.DEFAULT_FILE_FORMAT)) ||
-        (tableProps.containsKey("metaTable") && isValidMetadataTable(tableProps.getProperty("metaTable"))) ||
+        (tableProps.containsKey(IcebergAcidUtil.META_TABLE_PROPERTY) &&
+            isValidMetadataTable(tableProps.getProperty(IcebergAcidUtil.META_TABLE_PROPERTY))) ||
```
deniskuzZ, line 1516: can't we simplify to `isValidMetadataTable(tableProps)`?

difin: Done

PR #5255, opened 2024-05-21
HIVE-28270: Fix missing partition paths bug on drop_database

wecharyu:
In `HMSHandler#drop_database_core`, it needs to collect all partition paths that were not in the subdirectory of the table path, but now it only fetch the last batch of paths.
1. Collect all customized partition paths while dropping tables.
2. Add a unit test to cover this case.

Fix the bug, otherwise some partition paths that were not in the subdirectory of table path would not be deleted correctly.

No.

No.

Add a unit test:
```bash
mvn test -Dtest.groups= -Dtest=org.apache.hadoop.hive.metastore.TestRemoteHiveMetaStore#testDropDatabase -pl :hive-standalone-metastore-server
```

### DIFF
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
@@ -148,7 +148,7 @@ public abstract class TestHiveMetaStore {
     MetastoreConf.setVar(conf, ConfVars.METASTORE_METADATA_TRANSFORMER_CLASS, " ");
 
     MetaStoreTestUtils.setConfForStandloneMode(conf);
-    MetastoreConf.setLongVar(conf, ConfVars.BATCH_RETRIEVE_MAX, 2);
+    MetastoreConf.setLongVar(conf, ConfVars.BATCH_RETRIEVE_MAX, 1);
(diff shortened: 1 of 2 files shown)


standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
```diff
@@ -148,7 +148,7 @@ public void setUp() throws Exception {
     MetastoreConf.setVar(conf, ConfVars.METASTORE_METADATA_TRANSFORMER_CLASS, " ");
 
     MetaStoreTestUtils.setConfForStandloneMode(conf);
-    MetastoreConf.setLongVar(conf, ConfVars.BATCH_RETRIEVE_MAX, 2);
+    MetastoreConf.setLongVar(conf, ConfVars.BATCH_RETRIEVE_MAX, 1);
```
ayushtkn, line 151: Why do we need to change this in this scope?

ayushtkn (approved):
LGTM apart from the conf change in the test

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
wecharyu: Reverted, and turns to create 3 tables in the new test.

PR #5256, opened 2024-05-21
HIVE-28271: DirectSql fails for AlterPartitions.

ayushtkn:

Fix DirectSql for AlterPartitions for some DB

To fix directSql for alter partions

No

No

Checked Logs to see no direct sql related exception where they were present earlier

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
@@ -548,6 +548,16 @@ class MetaStoreDirectSql {
    */
   public List<Partition> alterPartitions(MTable table, List<String> partNames,
                                          List<Partition> newParts, String queryWriteIdList) throws MetaException {
+
+    for (Partition tmpPart : newParts) {
+      if (!tmpPart.getDbName().equalsIgnoreCase(table.getDatabase().getName())) {
+        throw new MetaException("Invalid DB name : " + tmpPart.getDbName());
+      }
+
+      if (!tmpPart.getTableName().equalsIgnoreCase(table.getTableName())) {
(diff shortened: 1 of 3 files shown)


ayushtkn:
cc. @wecharyu

ayushtkn:
the test failures seems related:
http://ci.hive.apache.org/job/hive-precommit/job/PR-5256/1/testReport/junit/org.apache.hadoop.hive.metastore.client/TestAlterPartitions/

I think earlier directSql wasn't running so they slipped through, not sure if it is a test issue or actual prod issue, the tests passes individually but fail when run as entire class

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -548,6 +548,16 @@ public void addPartitions(List<MPartition> parts, List<List<MPartitionPrivilege>
    */
   public List<Partition> alterPartitions(MTable table, List<String> partNames,
                                          List<Partition> newParts, String queryWriteIdList) throws MetaException {
+
+    for (Partition tmpPart : newParts) {
```
wecharyu, line 552: Can we move this common check into `ObjectStore#alterPartitions`?

wecharyu:
LGTM overall, only a nit code change request. BTW the error mainly comes from Derby DB right? @ayushtkn

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ayushtkn: done

ayushtkn:
Should be Oracle DB as well, they also return clob

wecharyu (approved):

zhangbutao (approved):
LGTM +1

PR #5257, opened 2024-05-22
HIVE-28272: Support setting per-session S3 credentials in Warehouse

zhangbutao:

At present, `Warehouse ` use a fixed `conf ` and we can not change it in the session connection. But acutully we have supported set session-level conf in HMSHandler, see `HMSHandler::setMetaConf`.
Some other info can be found in this previous ticket HIVE-16913.

This PR tries to support set seesion-level conf in `Warehouse `, and then we can unblock the usefull case like per-session S3 credentials.
1. HMS use `Warehouse `to delete/create hdfs dir, so we should pass s3 credentials to `Warehouse `conf. Adding threadlocal conf in `Warehouse `so that we can pass session-level conf to `Warehouse `synchronously.  e.g. `set metaconf:fs.s3a.secret.key=my_secret_key;`
2. Make S3 related credentials param valid in` MetastoreConf.ConfVars`

Currently, users can only set S3 credentials on the HMS server side and users cannot set different S3 credentials for multiple S3 buckets. With this change, users can set per-session S3 credentials in client side.
e.g, we can set S3 credentials like this from beeline client:

```
set fs.s3a.secret.key=my_secret_key;
set fs.s3a.access.key=my_access.key;
set fs.s3a.endpoint=my_endpoint;
set metaconf:fs.s3a.secret.key=my_secret_key;
set metaconf:fs.s3a.access.key=my_access_key; 
set metaconf:fs.s3a.endpoint=my_endpoint;
```

And you can try to create/insert a s3a table:
`create table testtbls3(id int) location 's3a://bucket/testdb/testtbls3';`

No

No

Local test.

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
@@ -100,6 +100,28 @@ public class Warehouse {
     isTenantBasedStorage = MetastoreConf.getBoolVar(conf, ConfVars.ALLOW_TENANT_BASED_STORAGE);
   }
 
+  // Thread local configuration is needed as many threads could make changes
+  // to the conf using the connection hook
+  private static final ThreadLocal<Configuration> threadLocalWhConf =
+          ThreadLocal.withInitial(() -> null);
+
+  public void setWhConf(Configuration conf) {
+    threadLocalWhConf.set(conf);
+  }
+
+  public Configuration getWhConf() {
+    Configuration conf = threadLocalWhConf.get();
+    if (conf == null) {
+      conf = new Configuration(this.conf);
+      threadLocalWhConf.set(conf);
+    }
+    return conf;
+  }
+
+  public static void removeWhThreadConf() {
+    threadLocalWhConf.remove();
+  }
+
   private MetaStoreFS getMetaStoreFsHandler(Configuration conf)
       throws MetaException {
     String handlerClassStr = MetastoreConf.getVar(conf, ConfVars.FS_HANDLER_CLS);
@@ -128,7 +150,7 @@ public class Warehouse {
   }
 
   public FileSystem getFs(Path f) throws MetaException {
-    return getFs(f, conf);
+    return getFs(f, getWhConf());
   }
 
 
@@ -188,7 +210,7 @@ public class Warehouse {
   }
 
   public Path getDnsPath(Path path) throws MetaException {
-    return getDnsPath(path, conf);
+    return getDnsPath(path, getWhConf());
   }
 
   /**
@@ -457,7 +479,7 @@ public class Warehouse {
       FileSystem srcFs = getFs(sourcePath);
       FileSystem destFs = getFs(destPath);
       // TODO: this operation can be expensive depending on the size of data. 
-      return FileUtils.copy(srcFs, sourcePath, destFs, destPath, false, false, conf);
+      return FileUtils.copy(srcFs, sourcePath, destFs, destPath, false, false, getWhConf());
     } catch (Exception ex) {
       MetaStoreUtils.throwMetaException(ex);
     }
@@ -489,7 +511,7 @@ public class Warehouse {
       }
     }
     FileSystem fs = getFs(f);
-    return fsHandler.deleteDir(fs, f, recursive, ifPurge, conf);
+    return fsHandler.deleteDir(fs, f, recursive, ifPurge, getWhConf());
   }
 
   public void recycleDirToCmPath(Path f, boolean ifPurge) throws MetaException {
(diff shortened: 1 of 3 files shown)


Aggarwal-Raghav:
@zhangbutao, Thanks for quickly re-creating this PR.

Aggarwal-Raghav:
@zhangbutao , i think the configs needs to be added to whitelist https://github.com/apache/hive/blob/33cadc5b498b57f779cddc9bf4e3f8aef2d9f6dc/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java#L6741 otherwise we cannot modify it at runtime.
Runtime_conf

zhangbutao:
> @zhangbutao , i think the configs needs to be added to whitelist
> 
> https://github.com/apache/hive/blob/33cadc5b498b57f779cddc9bf4e3f8aef2d9f6dc/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java#L6741
> 
> otherwise we cannot modify it at runtime.
> Runtime_conf

Yes, you are right. need to add these conf to whitelist.
Will optimize this if any other folks have the futher comments.

Aggarwal-Raghav:
@zhangbutao, 1 question here, when I try to insert the data using tez, the query fails with NoAuthWithAWSException on staging dir and when I debugged it, I found that the conf is not passed to tez (on yarn) container because of https://github.com/apache/hive/blob/33cadc5b498b57f779cddc9bf4e3f8aef2d9f6dc/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java#L1478
As fs.s3a.secret.key and fs.s3a.access.key exist in default hidden list.
I have tried it on hive3 as I don't have a setup of hive4 with S3. **_I understand that because of patch differences, issue might not occur with hive4 but in hive4 also, this DAGUtils is removing those s3 credentials from jobconf (tested on local setup with hdfs based table)_**. 

If you have any insights, can you help me understand if I am missing anything? The same happens in select query as well. In hive3 I have tested by removing the fs.s3a* configs from hidden list, then the insert/select query is passing.

Also, do we want to add metaconf:fs.s3a.* configs to hidden list? As in the above attached screenshot, I am able to see the credential value.

zhangbutao:
> @zhangbutao, 1 question here, when I try to insert the data using tez, the query fails with NoAuthWithAWSException on staging dir and when I debugged it, I found that the conf is not passed to tez (on yarn) container because of
> 
> https://github.com/apache/hive/blob/33cadc5b498b57f779cddc9bf4e3f8aef2d9f6dc/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java#L1478
> 
> 
> As fs.s3a.secret.key and fs.s3a.access.key exist in default hidden list.
> I have tried it on hive3 as I don't have a setup of hive4 with S3. **_I understand that because of patch differences, issue might not occur with hive4 but in hive4 also, this DAGUtils is removing those s3 credentials from jobconf (tested on local setup with hdfs based table)_**.
> If you have any insights, can you help me understand if I am missing anything? The same happens in select query as well. In hive3 I have tested by removing the fs.s3a* configs from hidden list, then the insert/select query is passing.
> 
> Also, do we want to add metaconf:fs.s3a.* configs to hidden list? As in the above attached screenshot, I am able to see the credential value.

@Aggarwal-Raghav I think this is a good question. See previous ticket https://issues.apache.org/jira/browse/HIVE-25929. At present, if we want to propagate the s3 related properttity to tez, we need to remove the fs.s3a* configs from hidden list, but this maybe insecurely.

I just did a magic way which only set a specific hidden property to allow the  fs.s3a*  configs to be propagated to tez. Like this:
```
   <property>
    <name>hive.conf.hidden.list</name>
    <value>javax.jdo.option.ConnectionPassword</value>
  </property>

```

Aggarwal-Raghav:
@zhangbutao, I have created this doc for Hive on S3, summarizing the problem, few limitation of current implementation and possible approach. Let me know if there are any gaps in my understanding and looking forward for your comments on this.
[Hive_on_S3A.docx](https://github.com/user-attachments/files/16086447/Hive_on_S3A.docx)

@abstractdog, if possible can you also review this as you might have some insights on this because of [HIVE-25929](https://issues.apache.org/jira/browse/HIVE-25929)
Thanks.

zhangbutao:
@Aggarwal-Raghav Good Job! I think the doc is good enough,
I only have one question about the way you mentioned to securely pass the credentials to TEZ job. If i understand correctly, this way need namenode to create credentials file. But in fact, some users don't want to deploy the hadoop cluster(namenode & datanode), and they just want to use hadoop-client to acess the s3 storage. So we may need another way to avoid deploying hadoop cluster.

Aggarwal-Raghav:
> @Aggarwal-Raghav Good Job! I think the doc is good enough, I only have one question about the way you mentioned to securely pass the credentials to TEZ job. If i understand correctly, this way need namenode to create credentials file. But in fact, some users don't want to deploy the hadoop cluster(namenode & datanode), and they just want to use hadoop-client to acess the s3 storage. So we may need another way to avoid deploying hadoop cluster.

@zhangbutao, yes you are right, I am looking at this problem from the perspective where both hdfs and s3 is present in the cluster. Our use case is to use s3 for archival purpose on-prem.

Aggarwal-Raghav:
@deniskuzZ, @dengzhhu653  can you please help with the review on this. I still think this is very good patch to have. We can expose `fs.s3a.credential.provider.path` in _MetastoreConf_ to securly pass S3A bucket credentials in session level.

PR #5259, opened 2024-05-22
HIVE-28274: Iceberg: Add support for 'If Not Exists' and 'or Replace' for Create Branch.

ayushtkn:

Add support for if not exists and OR Replace while creating Branches

Better Usability 

Yes, Queries with If not exists and OR Replace with create branch can be executed

No

UT

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergSnapshotRefExec.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergSnapshotRefExec.java
@@ -22,6 +22,7 @@ package org.apache.iceberg.mr.hive;
 import java.util.Optional;
 import org.apache.hadoop.hive.ql.parse.AlterTableSnapshotRefSpec;
 import org.apache.iceberg.ManageSnapshots;
+import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.SnapshotRef;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.util.SnapshotUtil;
@@ -42,6 +43,11 @@ public class IcebergSnapshotRefExec {
    */
   public static void createBranch(Table table, AlterTableSnapshotRefSpec.CreateSnapshotRefSpec createBranchSpec) {
     String branchName = createBranchSpec.getRefName();
+    boolean refExistsAsBranch = refExistsAsBranch(table, branchName);
+    if (createBranchSpec.isIfNotExists() && refExistsAsBranch) {
+      return;
+    }
+    boolean isReplace = createBranchSpec.isReplace() && refExistsAsBranch;
     Long snapshotId = null;
     if (createBranchSpec.getSnapshotId() != null) {
       snapshotId = createBranchSpec.getSnapshotId();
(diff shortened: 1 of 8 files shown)


ayushtkn:
* Had to refactor a bit 'unnecessarily' because checkstyle was complaining of cyclomatic complexity :-( 
* The empty iceberg table replace was bit tricky, I didn't find any other alt than dropping & recreating the branch under a transaction, will double check in the iceberg code, if there in any API which can do that magic

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergSnapshotRefExec.java
```diff
@@ -56,16 +59,35 @@ public static void createBranch(Table table, AlterTableSnapshotRefSpec.CreateSna
...
+    ManageSnapshots manageSnapshots;
     if (snapshotId != null) {
-      LOG.info("Creating a branch {} on an iceberg table {} with snapshotId {}", branchName, table.name(), snapshotId);
-      manageSnapshots.createBranch(branchName, snapshotId);
+      manageSnapshots = createBranch(table, isReplace, branchName, snapshotId);
     } else {
-      LOG.info("Creating a branch {} on an empty iceberg table {}", branchName, table.name());
-      manageSnapshots.createBranch(branchName);
+      if (isReplace) {
+        LOG.info("Replacing a branch {} on an empty iceberg table {}", branchName, table.name());
```
zhangbutao, line 70: > The empty iceberg table replace was bit tricky, I didn't find any other alt than dropping & recreating the branch under a transaction, will double check in the iceberg code, if there in any API which can do that magic, thought of truncate but it isn't replace, the same branch stays

I  checked Spark-iceberg code, and it will throw exception if replacing the branch on a empty table :
https://github.com/apache/iceberg/blob/dd2197f835aa3f180c0f49246055ce665f9fef59/spark/v3.5/spark-extensions/src/main/scala/org/apache/spark/sql/execution/datasources/v2/CreateOrReplaceBranchExec.scala#L62-L68

So i think we can do like spark. it's more simpler.
But if you think we need to support to replace branch on empty. I also have no objection. :)

ayushtkn: I don't have a usecase as such, it is just, If I call create branch on an empty iceberg table it will succeed but it won't when there is existing branch & some one uses replace, though he can himself achieve that by first dropping the branch & then calling create again.

if there are even some corner cases where we can be screwed by the current impl, we should definitely fallback to throwing exception.

Let me know, what you think, I will update the PR accordingly :-)

zhangbutao: ok,Let us follow spark-iceberg behavior, just throw exception on empty table when replacing branch.

ayushtkn: I have updated the PR to throw exception in this case, added a test as well to validate the case

zhangbutao (approved):
LGTM  +1

PR #5260, opened 2024-05-23
HIVE-28276: Iceberg:Make Iceberg split threads configurable when table scanning

zhangbutao:

I have noticed that if a iceberg table has lots of metadata/data files, iceberg will use many system cores to do scan planning, which may put some pressure on Tez AM memory/cpu.
We can try to make the thread pool size configurable, to avoid concurrency pressure on Tez AM.

Other OSS also did similar optimization, like FIink:
https://github.com/apache/iceberg/blob/cbb853073e681b4075d7c8707610dceecbee3a82/flink/v1.18/flink/src/main/java/org/apache/iceberg/flink/source/FlinkInputFormat.java#L92-L100

This can configure the icbeberg planning thread size, to reduce Tez AM memory/cpu pressure as well as reduce some pressure on operating system.
If we don't limit the thread pool size, it wii use a default value which is equal system cores when doing `scan.planTasks()`:, 
https://github.com/apache/iceberg/blob/9a5d24fee239352021a9a73f6a4cad8ecf464f01/core/src/main/java/org/apache/iceberg/SystemConfigs.java#L38-L43

No

No

Existing test.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/InputFormatConfig.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/InputFormatConfig.java
@@ -87,6 +87,8 @@ public class InputFormatConfig {
   public static final String CATALOG_DEFAULT_CONFIG_PREFIX = "iceberg.catalog-default.";
   public static final String QUERY_FILTERS = "iceberg.query.filters";
 
+  public static final String TABLE_PLAN_WORKER_POOL_SIZE = "iceberg.worker.num-threads";
+
   public enum InMemoryDataModel {
     PIG,
     HIVE,
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/InputFormatConfig.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/InputFormatConfig.java
@@ -87,8 +87,6 @@ public class InputFormatConfig {
   public static final String CATALOG_DEFAULT_CONFIG_PREFIX = "iceberg.catalog-default.";
   public static final String QUERY_FILTERS = "iceberg.query.filters";
 
-  public static final String TABLE_PLAN_WORKER_POOL_SIZE = "iceberg.worker.num-threads";
-
   public enum InMemoryDataModel {
     PIG,
     HIVE,
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -97,6 +98,7 @@ import org.apache.iceberg.types.TypeUtil;
 import org.apache.iceberg.types.Types;
 import org.apache.iceberg.util.PartitionUtil;
 import org.apache.iceberg.util.SerializationUtil;
+import org.apache.iceberg.util.ThreadPools;
 
 /**
  * Generic Mrv2 InputFormat API for Iceberg.
@@ -215,10 +217,13 @@ public class IcebergInputFormat<T> extends InputFormat<Void, T> {
 
     long fromVersion = conf.getLong(InputFormatConfig.SNAPSHOT_ID_INTERVAL_FROM, -1);
     Scan<?, FileScanTask, CombinedScanTask> scan;
+    final ExecutorService workerPool =
+            ThreadPools.newWorkerPool("iceberg-plan-worker-pool",
+                    conf.getInt(InputFormatConfig.TABLE_PLAN_WORKER_POOL_SIZE, ThreadPools.WORKER_THREAD_POOL_SIZE));
     if (fromVersion != -1) {
-      scan = applyConfig(conf, createIncrementalAppendScan(table, conf));
+      scan = applyConfig(conf, createIncrementalAppendScan(table, conf)).planWith(workerPool);
     } else {
-      scan = applyConfig(conf, createTableScan(table, conf));
+      scan = applyConfig(conf, createTableScan(table, conf)).planWith(workerPool);
     }
 
     boolean allowDataFilesWithinTableLocationOnly =
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
@@ -63,6 +63,7 @@ import org.apache.iceberg.Schema;
 import org.apache.iceberg.SchemaParser;
 import org.apache.iceberg.SnapshotRef;
 import org.apache.iceberg.StructLike;
+import org.apache.iceberg.SystemConfigs;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.TableProperties;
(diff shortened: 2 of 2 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
```diff
@@ -215,10 +217,13 @@ public List<InputSplit> getSplits(JobContext context) {
 
     long fromVersion = conf.getLong(InputFormatConfig.SNAPSHOT_ID_INTERVAL_FROM, -1);
     Scan<?, FileScanTask, CombinedScanTask> scan;
+    final ExecutorService workerPool =
+            ThreadPools.newWorkerPool("iceberg-plan-worker-pool",
+                    conf.getInt(InputFormatConfig.TABLE_PLAN_WORKER_POOL_SIZE, ThreadPools.WORKER_THREAD_POOL_SIZE));
```
ayushtkn, line 222: Does this work
```
    final ExecutorService workerPool =
        ThreadPools.newWorkerPool("iceberg-plan-worker-pool", SystemConfigs.WORKER_THREAD_POOL_SIZE.value());
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
```diff
@@ -215,10 +217,13 @@ public List<InputSplit> getSplits(JobContext context) {
...
     if (fromVersion != -1) {
-      scan = applyConfig(conf, createIncrementalAppendScan(table, conf));
+      scan = applyConfig(conf, createIncrementalAppendScan(table, conf)).planWith(workerPool);
```
ayushtkn, line 224: Do we need to shutdown this ``workerPool`` in the ``finally`` block

ayushtkn (commented):
maybe if the value configured is 1, we might not need to initialize this executor service and go with normal flow

zhangbutao:
@ayushtkn I think I got mixed up.  :(  The scan is in Tez AM side not HS2 side. I will double check this PR later. Make it WIP...

zhangbutao:
> maybe if the value configured is 1, we might not need to initialize this executor service and go with normal flow

The iceberg api side will always initialize the executor service,  so we can not do other things.




> @ayushtkn I think I got mixed up. :( The scan is in Tez AM side not HS2 side. I will double check this PR later. Make it WIP...
This PR is aimed for Tez AM, as the scan logic is on Tez AM side. This PR can let hive users to configure the Tez AM  scan thread size.
And i think we also have some iceberg table scan on HS2 side, especially iceberg maintenance feature.  I think we can also do similar optimization in the future.

ayushtkn (approved):
LGTM

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
ayushtkn: Just double check once, that iceberg takes care of that & there is no leak here before committing

zhangbutao: Make sense. Let me double check again.

zhangbutao: I checked iceberg code, and didn't find any code snippet about `workPool `shutdown. So i think you are right, shutdown this workerPool in the finally block is a more better way. Like flink iceberg side:
https://github.com/apache/iceberg/blob/cbb853073e681b4075d7c8707610dceecbee3a82/flink/v1.18/flink/src/main/java/org/apache/iceberg/flink/source/FlinkInputFormat.java#L92-L101

Change in this commit https://github.com/apache/hive/pull/5260/commits/ef6d6840a02c0a8e96cb959f61cca23da5878845


Thanks.

ayushtkn: thnx, looks good

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/InputFormatConfig.java
```diff
@@ -87,6 +87,8 @@ private InputFormatConfig() {
   public static final String CATALOG_DEFAULT_CONFIG_PREFIX = "iceberg.catalog-default.";
   public static final String QUERY_FILTERS = "iceberg.query.filters";
 
+  public static final String TABLE_PLAN_WORKER_POOL_SIZE = "iceberg.worker.num-threads";
```
deniskuzZ, line 90: can we use iceberg WORKER_POOL config for that: 'iceberg-worker-pool'?

zhangbutao: If users don't set the `iceberg.worker.num-threads` in beeline console, we will use & follow the pool size which equals iceberg `WORKER_POOL`'s pool size, and `WORKER_POOL`'s pool size is also the default thread pool size in current iceberg design.

deniskuzZ: can we ref the iceberg conf name like `SystemConfigs.WORKER_THREAD_POOL_SIZE.propertyKey()`

zhangbutao: Sure, we can. Fixed.

deniskuzZ: 👍

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
```diff
@@ -216,9 +230,9 @@ public List<InputSplit> getSplits(JobContext context) {
     long fromVersion = conf.getLong(InputFormatConfig.SNAPSHOT_ID_INTERVAL_FROM, -1);
     Scan<?, FileScanTask, CombinedScanTask> scan;
     if (fromVersion != -1) {
-      scan = applyConfig(conf, createIncrementalAppendScan(table, conf));
+      scan = applyConfig(conf, createIncrementalAppendScan(table, conf)).planWith(workerPool);
```
deniskuzZ, line 233: can we add 
````
plan = plan.planWith(workerPool)
````
after the `if-else`

Comment (zhangbutao) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java]: Done. :)

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java [hunk 5]
```diff
@@ -220,6 +233,7 @@ public List<InputSplit> getSplits(JobContext context) {
     } else {
       scan = applyConfig(conf, createTableScan(table, conf));
     }
+    scan = (Scan<?, FileScanTask, CombinedScanTask>) scan.planWith(workerPool);
```
deniskuzZ, line 236:
````
Scan<? extends Scan, FileScanTask, CombinedScanTask> scan;
scan = scan.planWith(workerPool);
````

deniskuzZ (commented):
no need for casting

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
zhangbutao: Fixed. Thanks!

deniskuzZ (approved):
+1, pending tests

PR #5261, opened 2024-05-23
HIVE-28278: Iceberg: Stats: IllegalStateException Invalid file: file length 0

deniskuzZ:

BugFix, can happen when the stats file was already created but stats object has not yet been written, and someone tried to read it. 

````
ERROR : FAILED: IllegalStateException Invalid file: file length 0 is less tha minimal length of the footer tail 12
java.lang.IllegalStateException: Invalid file: file length 0 is less tha minimal length of the footer tail 12
````

No

No

Locally

### CODE REVIEW

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 1]
```diff
@@ -534,10 +565,13 @@ public boolean canProvideColStatistics(org.apache.hadoop.hive.ql.metadata.Table
...
   private boolean canProvideColStats(Table table, long snapshotId) {
-    Path statsPath = getColStatsPath(table, snapshotId);
+    Optional<Path> statsPath = IcebergTableUtil.getColStatsPath(table, snapshotId);
+    if (!statsPath.isPresent()) {
+      return false;
+    }
     try {
-      FileSystem fs = statsPath.getFileSystem(conf);
-      return  fs.exists(statsPath);
+      FileSystem fs = statsPath.get().getFileSystem(conf);
```
zhangbutao, line 573: Do we still need the check filesystem path to confirm the stats file?

I think the result of `IcebergTableUtil::getColStatsPath` is enough.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -498,30 +501,58 @@ public boolean canSetColStatistics(org.apache.hadoop.hive.ql.metadata.Table hmsT
...
-      boolean rewriteStats = removeColStatsIfExists(tbl);
-      if (!rewriteStats) {
+      if (!shouldRewriteColStats(tbl)) {
         checkAndMergeColStats(tableColStats, tbl);
       }
       // Currently, we are only serializing table level stats.
       byte[] serializeColStats = SerializationUtils.serialize(tableColStats);
-      try (PuffinWriter writer = Puffin.write(tbl.io().newOutputFile(getColStatsPath(tbl).toString()))
+      StatisticsFile statisticsFile;
+      String statsPath = tbl.location() + STATS + UUID.randomUUID();
```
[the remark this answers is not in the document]
deniskuzZ: we can, however, none of the other data files provide this information

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -498,30 +501,58 @@ public boolean canSetColStatistics(org.apache.hadoop.hive.ql.metadata.Table hmsT
...
+        long snapshotId = tbl.currentSnapshot().snapshotId();
+        long snapshotSequenceNumber = tbl.currentSnapshot().sequenceNumber();
+        puffinWriter.add(
+            new Blob(
+              ColumnStatisticsObj.class.getSimpleName(),
+              ImmutableList.of(1),
+              snapshotId,
+              snapshotSequenceNumber,
+              ByteBuffer.wrap(serializeColStats),
+              PuffinCompressionCodec.NONE,
```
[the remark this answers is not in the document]
deniskuzZ: i can try to change in this PR, hopefully that won't require much changes on reader side

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: nice catch!

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -498,30 +501,58 @@ public boolean canSetColStatistics(org.apache.hadoop.hive.ql.metadata.Table hmsT
...
-    } catch (InvalidObjectException | IOException e) {
+      Transaction transaction = tbl.newTransaction();
+      transaction
+          .updateStatistics()
+          .setStatistics(statisticsFile.snapshotId(), statisticsFile)
+          .commit();
+      transaction.commitTransaction();
+      return true;
+
+    } catch (InvalidObjectException e) {
```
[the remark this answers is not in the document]
deniskuzZ: 👍

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -498,30 +500,58 @@ public boolean canSetColStatistics(org.apache.hadoop.hive.ql.metadata.Table hmsT
...
         LOG.warn("Unable to write stats to puffin file {}", e.getMessage());
         return false;
       }
-    } catch (InvalidObjectException | IOException e) {
+      Transaction transaction = tbl.newTransaction();
+      transaction
+          .updateStatistics()
+          .setStatistics(statisticsFile.snapshotId(), statisticsFile)
+          .commit();
+      transaction.commitTransaction();
```
ayushtkn, line 551: Why do we need a transaction here, why don't we directly update the statistics?
```
      tbl.updateStatistics().setStatistics(statisticsFile.snapshotId(), statisticsFile).commit();
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -133,6 +135,20 @@ static Table getTable(Configuration configuration, Properties properties) {
...
 
+  static Optional<Path> getColStatsPath(Table table) {
+    return getColStatsPath(table, table.currentSnapshot().snapshotId());
+  }
+
+  static Optional<Path> getColStatsPath(Table table, long snapshotId) {
+    return table.statisticsFiles().stream()
+      .filter(stats -> stats.snapshotId() == snapshotId)
+      .filter(stats -> stats.blobMetadata().stream()
+        .anyMatch(metadata -> ColumnStatisticsObj.class.getSimpleName().equals(metadata.type()))
```
ayushtkn, line 146: We did set created-by Hive while creating the Puffin file(` .createdBy(Constants.HIVE_ENGINE).build()``), is there a possibility to check that here as well, that the Stats file that we are consuming is Hive written only?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -498,30 +500,58 @@ public boolean canSetColStatistics(org.apache.hadoop.hive.ql.metadata.Table hmsT
...
-      boolean rewriteStats = removeColStatsIfExists(tbl);
-      if (!rewriteStats) {
+      if (!shouldRewriteColStats(tbl)) {
         checkAndMergeColStats(tableColStats, tbl);
       }
       // Currently, we are only serializing table level stats.
       byte[] serializeColStats = SerializationUtils.serialize(tableColStats);
-      try (PuffinWriter writer = Puffin.write(tbl.io().newOutputFile(getColStatsPath(tbl).toString()))
+      StatisticsFile statisticsFile;
+      String statsPath = tbl.location() + STATS + UUID.randomUUID();
```
ayushtkn, line 514: Maybe overthinking but if there is a clash with ``UUID.randomUUID()``, means for some other stats file, ``UUID.randomUUID()`` returns the same id then?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -498,30 +500,58 @@ public boolean canSetColStatistics(org.apache.hadoop.hive.ql.metadata.Table hmsT
...
+        long snapshotId = tbl.currentSnapshot().snapshotId();
+        long snapshotSequenceNumber = tbl.currentSnapshot().sequenceNumber();
+        puffinWriter.add(
+            new Blob(
+              ColumnStatisticsObj.class.getSimpleName(),
+              ImmutableList.of(1),
+              snapshotId,
+              snapshotSequenceNumber,
+              ByteBuffer.wrap(serializeColStats),
+              PuffinCompressionCodec.ZSTD,
```
ayushtkn, line 527: earlier it was `none`, Are we sure we want to go with ZSTD or we should preserve `none` here like before? I think the iceberg Puffin default is `none`

https://github.com/apache/iceberg/blob/311dbbb10dfa41fd22966ec87a03d449d0d84f96/core/src/main/java/org/apache/iceberg/puffin/Puffin.java#L39

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -534,34 +564,24 @@ public boolean canProvideColStatistics(org.apache.hadoop.hive.ql.metadata.Table
...
-    Path statsPath = getColStatsPath(table, snapshotId);
-    try {
-      FileSystem fs = statsPath.getFileSystem(conf);
-      return  fs.exists(statsPath);
-    } catch (Exception e) {
-      LOG.warn("Exception when trying to find Iceberg column stats for table:{} , snapshot:{} , " +
-          "statsPath: {} , stack trace: {}", table.name(), table.currentSnapshot(), statsPath, e);
-    }
-    return false;
+    return IcebergTableUtil.getColStatsPath(table, snapshotId).isPresent();
```
ayushtkn, line 567: Do we know in one query how many times do we call this ``getColStatsPath()``? I think atleast twice, once for `canProvideColStats` & second would be when we do `getColStatistics`, if more or not that cheap then may be we should explore caching it or some other way to avoid calculating multiple times

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -534,34 +564,24 @@ public boolean canProvideColStatistics(org.apache.hadoop.hive.ql.metadata.Table
...
   private ColumnStatistics readColStats(Table table, Path statsPath) {
     try (PuffinReader reader = Puffin.read(table.io().newInputFile(statsPath.toString())).build()) {
       List<BlobMetadata> blobMetadata = reader.fileMetadata().blobs();
-      Map<BlobMetadata, ColumnStatistics> collect = Streams.stream(reader.readAll(blobMetadata)).collect(
-          Collectors.toMap(Pair::first, blobMetadataByteBufferPair -> SerializationUtils.deserialize(
-              ByteBuffers.toByteArray(blobMetadataByteBufferPair.second()))));
-      return collect.get(blobMetadata.get(0));
-    } catch (IOException | IndexOutOfBoundsException e) {
-      LOG.warn(" Unable to read iceberg col stats from puffin files: ", e);
+      byte[] byteBuffer = ByteBuffers.toByteArray(reader.readAll(blobMetadata).iterator().next().second());
```
ayushtkn, line 580: we didn't do hasNext() here? Can this lead to an exception in case someone tries to read the file before it was completely written as the case mentioned in the description?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
deniskuzZ: nope, it's not available in metadata

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: it's not doing FS listing, just iceberg stats metadata filter

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: file won't be registered in snapshot until the write is done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: are we writing data files with `none`? i think we should use same codec

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: I don't think we should worry about this, it's used for all Iceberg files;
According to [wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Random_UUID_probability_of_duplicates), regarding the probability of duplicates in random UUIDs:
````
Only after generating 1 billion UUIDs every second for the next 100 years, the probability of creating just one duplicate would be about 50%. Or, to put it another way, the probability of one duplicate would be about 50% if every person on earth owned 600 million UUIDs.
````

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: no reason,  i'll remove

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: That doesn't read the Stats file? If not thats cool, it should be cheap if no FS call is involved

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: It should be defined by the config I believe "write.parquet.compression-codec" "write.avro.compression-codec" & they are different I believe & even different for us downstream :-)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: no, just checks metadata if it has stats refs there

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
ayushtkn: Should be fine, I don't plan to live for 100 years, thats a cool fact though :-)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: changed back to NONE

ayushtkn (approved):
If the build is happy
Changes LGTM

zhangbutao (approved):
+1

PR #5262, opened 2024-05-23
HIVE-25351: stddev(), stddev_pop() return NaN when CBO is enabled

yangjiandan:
when cbo is enable 
```
STDDEV_SAMP(x) → SQRT( CASE WHEN ((SUM(x * x) - SUM(x) * SUM(x) / COUNT(x)) < 0) THEN 0 THEN (SUM(x * x) - SUM(x) * SUM(x) / COUNT(x)) / CASE COUNT(x) WHEN 1 THEN NULL ELSE COUNT(x) - 1 END)
```

When the column type is double or decimal then SUM(x * x) and SUM(x) * SUM(x) / COUNT(x) values difference is coming negative value (very very small difference in fractional value) which is leading to NaN when applied SQRT(For more detailed info, see https://github.com/apache/hive/pull/4812)


yangjiandan:
Hi, @vineetgarg02, could you please review my merge request? I would appreciate your feedback on the change. thank you!

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAggregateReduceFunctionsRule.java
```diff
@@ -540,7 +546,7 @@ private RexNode reduceStddev(
 
     final RexNode div =
         rexBuilder.makeCall(
-            SqlStdOperatorTable.DIVIDE, diff, denominator);
+            SqlStdOperatorTable.DIVIDE, correctedDiff, denominator);
```
deniskuzZ, line 549: are we dividing on 0 here when `diff<0` ?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAggregateReduceFunctionsRule.java
```diff
@@ -517,6 +517,12 @@ private RexNode reduceStddev(
         rexBuilder.makeCall(
             SqlStdOperatorTable.MINUS,
             sumArgSquared, avgSumSquaredArg);
+    final RexNode zeroLiteral =
+            rexBuilder.makeExactLiteral(BigDecimal.ZERO);
+    final RexNode correctedDiff = rexBuilder.makeCall(SqlStdOperatorTable.CASE,
+            rexBuilder.makeCall(SqlStdOperatorTable.LESS_THAN, diff, zeroLiteral),
```
deniskuzZ, line 523: should we check that it's not only less than zero, but within defined epsilon?

deniskuzZ:
hi @yangjiandan, it seems that this PR went under the radar. 
Could you please rebase and check the comment

deniskuzZ (dismissed):
+1

deniskuzZ:
STDDEV_SAMP(x) Screenshot 2025-02-01 at 17 56 57

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAggregateReduceFunctionsRule.java
```diff
@@ -517,6 +517,12 @@ private RexNode reduceStddev(
         rexBuilder.makeCall(
             SqlStdOperatorTable.MINUS,
             sumArgSquared, avgSumSquaredArg);
+    final RexNode zeroLiteral =
```
deniskuzZ: could be simplified with `greatest` (remove final)
````
diff = rexBuilder.makeCall(
    SqlLibraryOperators.GREATEST, diff, rexBuilder.makeExactLiteral(BigDecimal.ZERO));
````

**Comment by deniskuzZ** [comment]
hi @yangjiandan, do you plan to continue with the PR?

**Comment by deniskuzZ** [comment]
hi @yangjiandan, sorry I missed the notification. I think the issue is related to your local changes since master is green. Could you please check under debug if we generate a valid SQL for STDDEV?

File: ql/src/test/results/clientpositive/llap/stddev_variance_with_double_decimal_test.q.out [hunk 4]
```diff
@@ -0,0 +1,196 @@
...
+#### A masked pattern was here ####
+      Edges:
+        Reducer 2 <- Map 1 (CUSTOM_SIMPLE_EDGE)
+#### A masked pattern was here ####
+      Vertices:
+        Map 1
+            Map Operator Tree:
+                TableScan
+                  alias: cbo_test
+                  Statistics: Num rows: 6 Data size: 744 Basic stats: COMPLETE Column stats: COMPLETE
```
deniskuzZ: @okumin, @kasakrisz I think we have a `good` regression. 
On branch-4.1 Column stats: **NONE**
`set hive.stats.column.autogather=false;` is a 2nd line a q.file, so it's good, right?

kasakrisz: It is `NONE` on the master branch too and it makes sense.

deniskuzZ: Thanks for the confirmation, Krisztian! I am trying to resurrect this PR and get the fix for NaN issue

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAggregateReduceFunctionsRule.java
deniskuzZ: tried it, but got an exception

deniskuzZ (approved):
+1, pending tests

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAggregateReduceFunctionsRule.java
deniskuzZ: updated PR with above change

zhangbutao:
Is this PR ready to be merged in branch-4.1&master branch?

deniskuzZ:
> Is this PR ready to be merged in branch-4.1&master branch?

@zhangbutao, first we need to merge it in master, and since I refactored the PR, we need another +1. 
Similar change was already merged into Calcite:  https://github.com/apache/calcite/pull/4448

deniskuzZ:
hi @kasakrisz, could you please +1? 
https://github.com/apache/calcite/pull/4448 is already merged. Once it is released, we could try replacing the whole `HiveAggregateReduceFunctionsRule`

kasakrisz:
> hi @kasakrisz, could you please +1? [apache/calcite#4448](https://github.com/apache/calcite/pull/4448) is already merged. Once it is released, we could try replacing the whole `HiveAggregateReduceFunctionsRule`

Currently we have vectorized implementation of the `Sum` function only but not the `Sum0` function.
https://github.com/apache/hive/blob/master/ql/src/gen/vectorization/UDAFTemplates/VectorUDAFSum.txt

`VectorUDAFSum0` is required to use `AggregateReduceFunctionsRule` shipped with Calcite
https://github.com/apache/calcite/blob/e536d3949674cbbc0fd1162b3fccb211ca75789b/core/src/main/java/org/apache/calcite/rel/rules/AggregateReduceFunctionsRule.java#L289-L292

In `HiveAggregateReduceFunctionsRule` we actually do the opposite: convert `sum0` to `sum` in order to exploit vectorization
https://github.com/apache/hive/blob/1a7fbd7183554f9ace10fbf7d066677eff0447a7/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAggregateReduceFunctionsRule.java#L213-L215

PR #5264, opened 2024-05-23
HIVE-28279: Output the database name for SHOW EXTENDED TABLES statement

wecharyu:
[HIVE-21301](https://issues.apache.org/jira/browse/HIVE-21301) introduced SHOW EXTENDED TABLES statement which output table name and table type while listing tables in a database.

In this patch, we aim to add a new output filed for database name.

1. database name in `SHOW EXTENDED TABLES` statement is optional, output the database is informational in this case.
2. when statistic table names and database names by this statement for list of databases, the output result including database name is much more helpful.

Yes, `SHOW EXTENDED TABLES` will show database name now.

No

Test in beeline:
```bash
0: jdbc:hive2://> show extended tables;
+----------+-----------+-----------------+
| db_name  | tab_name  |   table_type    |
+----------+-----------+-----------------+
| default  | tbl       | EXTERNAL_TABLE  |
| default  | tbl2      | EXTERNAL_TABLE  |
+----------+-----------+-----------------+
2 rows selected (1.519 seconds)
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesDesc.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesDesc.java
@@ -34,7 +34,7 @@ public class ShowTablesDesc implements DDLDesc, Serializable {
   private static final long serialVersionUID = 1L;
 
   private static final String TABLES_VIEWS_SCHEMA = "tab_name#string";
-  private static final String EXTENDED_TABLES_SCHEMA = "tab_name,table_type#string,string";
+  private static final String EXTENDED_TABLES_SCHEMA = "db_name,tab_name,table_type#string,string,string";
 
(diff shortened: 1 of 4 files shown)


wecharyu:
@deniskuzZ @zhangbutao @ayushtkn : could you take a look?

deniskuzZ:
@wecharyu, why do we need this output format, all tables would be for the same db specified in show query?

wecharyu:
@deniskuzZ Yes, it would get the same db in the result of single query. The output with db_name is useful in collecting the results of multiple show tables queries.

ayushtkn:
I don't think it add any value, if someone is collecting the output per databse, they can add one line in their notes for the database name & put the output.

I don't think for such use case we need to do code changes

wecharyu:
Most engines and dbms do not output db_name for `SHOW TABLES` statement indeed, will close this PR.

PR #5267, opened 2024-05-29
HIVE-28282: Merging into iceberg table fails with copy on write when values clause has a function call

kasakrisz:
Hive rewrites merge statements to insert overwrite statements. When the target table is Iceberg with table property
```
'write.merge.mode'='copy-on-write'
```
The rewrite of
```
explain
merge into target_ice as t using source src ON t.a = src.a
when matched and t.a > 100 THEN DELETE
when not matched then insert (a, b) values (src.a, concat(src.b, '-merge new 2'));
```
should be
```
WITH `src` AS ( SELECT * FROM
(SELECT `PARTITION__SPEC__ID` AS `t__PARTITION__SPEC__ID`,`PARTITION__HASH` AS `t__PARTITION__HASH`,`FILE__PATH` AS `t__FILE__PATH`,`ROW__POSITION` AS `t__ROW__POSITION`,`a` AS `t__a`, `b` AS `t__b`, `c` AS `t__c` FROM `default`.`target_ice`) `t`
  FULL OUTER JOIN
  `default`.`source` `src`
  ON `t__a` = `src`.`a`
),
t AS (
  select `t__PARTITION__SPEC__ID`,`t__PARTITION__HASH`,`t__FILE__PATH`,-1,`t__a`,`t__b`,`t__c` from (
    select `t__PARTITION__SPEC__ID`,`t__PARTITION__HASH`,`t__FILE__PATH`,-1,`t__a`,`t__b`,`t__c`, row_number() OVER (partition by `t__FILE__PATH`) rn from `src`
    where `t__a` = `src`.`a` AND `t__a` > 100
  ) q
  where rn=1
)
INSERT INTO `default`.`target_ice`
    -- insert clause
SELECT `t__PARTITION__SPEC__ID`,`t__PARTITION__HASH`,`t__FILE__PATH`,`t__ROW__POSITION`,`src`.`a`,concat(`src`.`b`, '-merge new 2'),null
FROM `src`
   WHERE `t__a` IS NULL
union all
    -- delete clause
SELECT `t__PARTITION__SPEC__ID`,`t__PARTITION__HASH`,`t__FILE__PATH`,`t__ROW__POSITION`,`t__a`,`t__b`,`t__c`
FROM `src`  WHERE 
  ( NOT(`t__a` = `src`.`a` AND `t__a` > 100 OR `t__a` IS NULL) OR (`t__a` = `src`.`a` AND `t__a` > 100 OR `t__a` IS NULL) IS NULL )
  AND `t__FILE__PATH` IN ( select `t__FILE__PATH` from t )
union all
select * from t
```
however when the values clause in when not matched insert branch has a function call with more than one parameter the expressions in the values clause are malformed.
The values clause is treated as one string which contains all the value expressions delimited by `,` and it is passed to  `CopyOnWriteMergeRewriter` which generates the insert statement. The value expressions are extracted from the string by splitting it 
https://github.com/apache/hive/blob/3be197e2ca8ad819fd41c507936ef9327571f855/ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/CopyOnWriteMergeRewriter.java#L160
In our example
```
src.a, concat(src.b, '-merge new 2')
```
the function call
```
concat(src.b, '-merge new 2')
```
also contains a `,` and it is split. So we end up with the following value expressions
```
`src`.`a`
`concat`(`src`.`b`
'-merge new 2')
```
where
```
`concat`(`src`.`b`
'-merge new 2')
```
should remain
```
concat(`src`.`b`, '-merge new 2')
```
Splitting value expressions leads to invalid value expression list hence malformed rewritten statement.

No.

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestIcebergCliDriver -Dqfile=merge_iceberg_copy_on_write_unpartitioned.q -pl itests/qtest-iceberg/ -Pitests,iceberg
mvn test -Dtest.output.overwrite -Dtest=TestMiniLlapLocalCliDriver -Dqfile=insert_into_default_keyword.q,insert_into_default_keyword_2.q -pl itests/qtest -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
@@ -19,12 +19,27 @@
 package org.apache.hadoop.hive.ql.metadata;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
+import java.util.Stack;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.SetMultimap;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.ql.lib.CostLessRuleDispatcher;
+import org.apache.hadoop.hive.ql.lib.ExpressionWalker;
+import org.apache.hadoop.hive.ql.lib.Node;
+import org.apache.hadoop.hive.ql.lib.NodeProcessorCtx;
+import org.apache.hadoop.hive.ql.lib.SemanticGraphWalker;
+import org.apache.hadoop.hive.ql.lib.SemanticNodeProcessor;
+import org.apache.hadoop.hive.ql.parse.ASTNode;
+import org.apache.hadoop.hive.ql.parse.HiveParser;
 import org.apache.hadoop.hive.ql.parse.Quotation;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+import org.apache.hadoop.hive.ql.parse.UnparseTranslator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.conf.Configuration;
@@ -300,6 +315,52 @@ public final class HiveUtils {
 
   public static String unparseIdentifier(String identifier) {
     return unparseIdentifier(identifier, Quotation.BACKTICKS);
+  }
+
+  public static UnparseTranslator collectUnescapeIdentifierTranslations(ASTNode node)
+      throws SemanticException {
+    UnparseTranslator unparseTranslator = new UnparseTranslator(Quotation.BACKTICKS);
+    unparseTranslator.enable();
+
+    SetMultimap<Integer, SemanticNodeProcessor> astNodeToProcessor = HashMultimap.create();
+    astNodeToProcessor.put(HiveParser.TOK_TABLE_OR_COL, new ColumnExprProcessor());
+    astNodeToProcessor.put(HiveParser.DOT, new ColumnExprProcessor());
+    NodeProcessorCtx nodeProcessorCtx = new QuotedIdExpressionContext(unparseTranslator);
+
+    CostLessRuleDispatcher costLessRuleDispatcher = new CostLessRuleDispatcher(
+        (nd, stack, procCtx, nodeOutputs) -> null, astNodeToProcessor, nodeProcessorCtx);
+    SemanticGraphWalker walker = new ExpressionWalker(costLessRuleDispatcher);
+    walker.startWalking(Collections.singletonList(node), null);
+    return unparseTranslator;
+  }
+
+  static class ColumnExprProcessor implements SemanticNodeProcessor {
+
+    @Override
+    public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx, Object... nodeOutputs)
+        throws SemanticException {
+      UnparseTranslator unparseTranslator = ((QuotedIdExpressionContext)procCtx).getUnparseTranslator();
+      ASTNode tokTableOrColNode = (ASTNode) nd;
+      for (int i = 0; i < tokTableOrColNode.getChildCount(); ++i) {
+        ASTNode child = (ASTNode) tokTableOrColNode.getChild(i);
+        if (child.getType() == HiveParser.Identifier) {
+          unparseTranslator.addIdentifierTranslation(child);
+        }
+      }
+      return null;
+    }
+  }
+
+  static class QuotedIdExpressionContext implements NodeProcessorCtx {
+    private final UnparseTranslator unparseTranslator;
+
+    public QuotedIdExpressionContext(UnparseTranslator unparseTranslator) {
+      this.unparseTranslator = unparseTranslator;
+    }
+
+    public UnparseTranslator getUnparseTranslator() {
+      return unparseTranslator;
+    }
   }
 
   public static HiveStorageHandler getStorageHandler(
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
@@ -18,6 +18,7 @@
 package org.apache.hadoop.hive.ql.parse;
 
 import org.antlr.runtime.TokenRewriteStream;
+import org.apache.commons.lang3.ObjectUtils;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.ql.ErrorMsg;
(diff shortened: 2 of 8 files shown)


ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
```diff
@@ -302,6 +317,52 @@ public static String unparseIdentifier(String identifier) {
     return unparseIdentifier(identifier, Quotation.BACKTICKS);
   }
 
+  public static UnparseTranslator collectUnescapeIdentifierTranslations(ASTNode node)
```
deniskuzZ, line 320: no change here, just the class move, right?

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
```diff
@@ -178,8 +178,8 @@ protected static class MergeWhenClauseSqlGenerator implements MergeStatement.Mer
     @Override
     public void appendWhenNotMatchedInsertClause(MergeStatement.InsertClause insertClause) {
       sqlGenerator.append("INSERT INTO ").append(mergeStatement.getTargetName());
-      if (insertClause.getColumnListText() != null) {
-        sqlGenerator.append(' ').append(insertClause.getColumnListText());
+      if (insertClause.getColumnList() != null) {
+        sqlGenerator.append(' ').append(String.join(",",insertClause.getColumnList()));
```
deniskuzZ, line 182: nit, space `String.join(", ", insertClause.getColumnList())`

deniskuzZ: should we create a helper method 'toString(List<String> value)'?

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -350,26 +352,51 @@ private MergeStatement.InsertClause handleInsert(ASTNode whenNotMatchedClause, S
...
+    if (columnListNode != null) {
+      columnNames = new ArrayList<>(valuesNode.getChildCount());
+      for (int i = 0; i < columnListNode.getChildCount(); ++i) {
+        ASTNode columnNameNode = (ASTNode) columnListNode.getChild(i);
+        String columnName = ctx.getTokenRewriteStream().toString(columnNameNode.getTokenStartIndex(),
+            columnNameNode.getTokenStopIndex()).trim();
+        columnNames.add(columnName);
+      }
+    } else {
+      columnNames = null;
```
deniskuzZ, line 366: should it be an empty list? do we handle nulls later on

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -350,26 +352,51 @@ private MergeStatement.InsertClause handleInsert(ASTNode whenNotMatchedClause, S
...
-    defaultValuesTranslator.applyTranslations(ctx.getTokenRewriteStream());
-    String valuesClause = getMatchedText(valuesNode);
-    valuesClause = valuesClause.substring(1, valuesClause.length() - 1); //strip '(' and ')'
 
-    String extraPredicate = getWhenClausePredicate(whenNotMatchedClause);
-    return new MergeStatement.InsertClause(
-        getMatchedText(columnListNode), valuesClause, onClausePredicate, extraPredicate);
-  }
+    List<String> columnNames;
+    if (columnListNode != null) {
```
deniskuzZ, line 357: can it be null, there is no check above for null - columnListNode.getChildCount()?

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -350,26 +352,51 @@ private MergeStatement.InsertClause handleInsert(ASTNode whenNotMatchedClause, S
...
-          ASTNode valueClause, Table targetTable, List<String> targetSchema, UnparseTranslator unparseTranslator)
-          throws SemanticException {
+    List<String> values = new ArrayList<>(valuesNode.getChildCount());
+    UnparseTranslator unparseTranslator = HiveUtils.collectUnescapeIdentifierTranslations(valuesNode);
+    unparseTranslator.applyTranslations(ctx.getTokenRewriteStream(), MERGE_INSERT_VALUES_PROGRAM);
+    List<String> targetSchema = processTableColumnNames(columnListNode, targetTable.getFullyQualifiedName());
     List<String> defaultConstraints = getDefaultConstraints(targetTable, targetSchema);
-    for (int j = 0; j < defaultConstraints.size(); j++) {
-      unparseTranslator.addDefaultValueTranslation((ASTNode) valueClause.getChild(j + 1), defaultConstraints.get(j));
+    // First child is 'struct' the rest are the value expressions
```
deniskuzZ, line 374: comma before `the rest`

deniskuzZ (commented):
LGTM, some minor comments

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
kasakrisz: Added space. 
I think the helper method is `String.join` here. Column list is joined only once. I don't have strong opinion about this.

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: When column names are specified additional steps must be done in both Rewriter implementations so the null check is required. If the lack of column names is represented by an empty list `List.isEmpty()` would be used.

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: refactored this part

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
kasakrisz: added

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
```diff
@@ -350,26 +352,51 @@ private MergeStatement.InsertClause handleInsert(ASTNode whenNotMatchedClause, S
...
+    //       any_column_name
+    //    3
+    for (int i = 1; i < valuesNode.getChildCount(); ++i) {
+      ASTNode valueNode = (ASTNode) valuesNode.getChild(i);
+      String value;
+      if (valueNode.getType() == HiveParser.TOK_TABLE_OR_COL
+          && valueNode.getChild(0).getType() == HiveParser.TOK_DEFAULT_VALUE) {
+        value = defaultConstraints.get(i - 1);
+        if (value == null) {
+          value = "NULL";
```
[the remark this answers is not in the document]
kasakrisz: changed

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/MergeRewriter.java
deniskuzZ: ok

ql/src/java/org/apache/hadoop/hive/ql/parse/MergeSemanticAnalyzer.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
kasakrisz: yes

PR #5268, opened 2024-05-29
HIVE-28287: Attempt make the scratch directory writable before failing.

ayushtkn:

Attempt making the scratch directory writable before failing

Better User/Admin experiance, in cloud situations where managing 

Yes, The Env setup won't fail if it is able to make the directory writable

No

In an actual Cloudera Azure Env


simhadri-g (approved):
LGTM +1

PR #5269, opened 2024-05-29
HIVE-28286: Add filtering support for get_table_metas (Naveen Gangam)

nrg4878:
Currently the filterTableMetas() method in the HiveMetastoreAuthorizer is a no-op. It returns the full set of tables that are passed into it. This prevents any filtering that is to be done.

This PR adds logic to be able to filter out the ones that the user does not have access to.

To enhance security in HMS.

No

No

Tested manually in a Ranger enabled environment.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
@@ -65,8 +65,11 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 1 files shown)


nrg4878:
@jfsii Could you review this change please?

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
```diff
@@ -213,15 +214,46 @@ public final List<String> filterCatalogs(List<String> catalogs) throws MetaExcep
 
   @Override
   @Deprecated
-  public List<TableMeta> filterTableMetas(String catName, String dbName,List<TableMeta> tableMetas)
+  public List<TableMeta> filterTableMetas(String catName, String dbName, List<TableMeta> tableMetas)
```
mayankkunwar, line 217: In this method, I see that "catName" is never used, so what would be the reason to have this parameter?

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
```diff
@@ -213,15 +214,46 @@ public final List<String> filterCatalogs(List<String> catalogs) throws MetaExcep
...
+    return filteredTableMetas;
   }
 
   @Override
   public final List<TableMeta> filterTableMetas(List<TableMeta> tableMetas)
       throws MetaException {
-    return tableMetas;
+    String catName = null;
+    String dbName = null;
+    if (tableMetas != null) {
```
mayankkunwar, line 252: If "tableMetas" is empty, can it throw ArrayOutOfBound exception?

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
```diff
@@ -213,15 +214,46 @@ public final List<String> filterCatalogs(List<String> catalogs) throws MetaExcep
...
+      catName = tableMetas.get(0).getCatName();
+      dbName = tableMetas.get(0).getDbName();
```
mayankkunwar, line 254: I am not sure about this, but will the "tableMetas" have only single DbName? Can't it have, multiple DbNames?

dengzhhu653: sounds like we might get multiple dbs in this call, as the param `dbnames` in `get_table_meta` is a pattern.

mayankkunwar: So, I think in the other method, sending same dbName with the list of tables in "TableFilterContext" would not work as expected.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
```diff
@@ -213,15 +214,46 @@ public final List<String> filterCatalogs(List<String> catalogs) throws MetaExcep
...
-  public List<TableMeta> filterTableMetas(String catName, String dbName,List<TableMeta> tableMetas)
+  public List<TableMeta> filterTableMetas(String catName, String dbName, List<TableMeta> tableMetas)
       throws MetaException {
-    return filterTableMetas(tableMetas);
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> HiveMetaStoreAuthorizer.filterTableMetas()");
+    }
+    List<String> tableNames = new ArrayList<>();
+    List<TableMeta> filteredTableMetas = new ArrayList<>();
+    if (tableMetas != null) {
```
mayankkunwar, line 224: "&& !tableMetas.isEmpty()" in IF clause can help

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
```diff
@@ -213,15 +214,46 @@ public final List<String> filterCatalogs(List<String> catalogs) throws MetaExcep
 
   @Override
   @Deprecated
-  public List<TableMeta> filterTableMetas(String catName, String dbName,List<TableMeta> tableMetas)
+  public List<TableMeta> filterTableMetas(String catName, String dbName, List<TableMeta> tableMetas)
       throws MetaException {
-    return filterTableMetas(tableMetas);
+    if (LOG.isDebugEnabled()) {
```
zratkai, line 219: No need to check LOG.isDebugEnabled(), since LOG.debug will only output, when debug level logging is enabled.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
```diff
@@ -213,15 +214,46 @@ public final List<String> filterCatalogs(List<String> catalogs) throws MetaExcep
...
+      TableFilterContext     tableFilterContext     = new TableFilterContext(dbName, tableNames);
+      HiveMetaStoreAuthzInfo hiveMetaStoreAuthzInfo = tableFilterContext.getAuthzContext();
+      final List<String>  filteredTableNames = filterTableNames(hiveMetaStoreAuthzInfo, dbName, tableNames);
+      if (CollectionUtils.isEmpty(filteredTableNames)) {
+        filteredTableMetas = Collections.emptyList();
+        if (LOG.isInfoEnabled()) {
+          LOG.info("<== HiveMetaStoreAuthorizer.filterTableMetas() : returning empty set");
+        }
+      } else {
+        if (LOG.isDebugEnabled()) {
```
zratkai, line 237: No need to check LOG.isDebugEnabled(), since LOG.debug will only output, when debug level logging is enabled.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
jfsii: It is a deprecated method from the interface MetaStoreFilterHook, it has to be implemented. As for the usage of it, I would say removing the parameter would be outside of the scope of this review (looking around the codebase, most people pass null to the method, so I'm going to guess it is not really used anywhere but fixing it should be a different PR).

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
```diff
@@ -213,15 +214,46 @@ public final List<String> filterCatalogs(List<String> catalogs) throws MetaExcep
...
+    List<String> tableNames = new ArrayList<>();
+    List<TableMeta> filteredTableMetas = new ArrayList<>();
+    if (tableMetas != null) {
+      for (TableMeta tableMeta : tableMetas) {
+        tableNames.add(tableMeta.getTableName());
+      }
+      TableFilterContext     tableFilterContext     = new TableFilterContext(dbName, tableNames);
+      HiveMetaStoreAuthzInfo hiveMetaStoreAuthzInfo = tableFilterContext.getAuthzContext();
+      final List<String>  filteredTableNames = filterTableNames(hiveMetaStoreAuthzInfo, dbName, tableNames);
+      if (CollectionUtils.isEmpty(filteredTableNames)) {
```
jfsii, line 231: Do you need the if/else?
It could be re-written as (since filteredTableMetas is initialized empty):
```
if (CollectionUtils.isNotEmpty(filteredTableNames)) {
   filteredTableMetas = tableMetas.stream().filter(tblMeta -> filteredTableNames.stream()
            .anyMatch(tblName -> tblName.equals(tblMeta.getTableName()))).collect(Collectors.toList());
}
```
Actually I think you can even remove the check for empty too if you wanted:
```
filteredTableMetas = tableMetas.stream().filter(tblMeta -> filteredTableNames.stream()
            .anyMatch(tblName -> tblName.equals(tblMeta.getTableName()))).collect(Collectors.toList());
```
which might do a bit more work than needed (if there are no filteredTableNames, I assume collect(Collectors.toList()) would return an empty list.

Now that I think about it the streaming line will be N^2 if nothing is filtered, it might be a bit better to convert fileteredTableNames to a set, so you can simply do an existence check in the filter portion rather than another stream.

jfsii (commented):
Left a few comments on structure and code, I didn't see any incorrect functionality - I agree with the other comments around the log lines.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
mayankkunwar: Got it. Thanks

dengzhhu653 (approved):
I think we have addressed the comments, it would be nice if someone can take an extra eye, cc @jfsii @mayankkunwar @zratkai. Thank you in advance!

mayankkunwar (approved):
LGTM

PR #5270, opened 2024-05-30
HIVE-26018: Produce the same result for UNIQUEJOIN regardless of execution engine

ngsg:

Change CommonJoinOperator's logic to treat a single dummy row as a null entry. The new logic is already used when it processes OuterJoin.
 
https://github.com/apache/hive/blob/e7de9164557e6029fd845b8da0f78a6021e965bb/ql/src/java/org/apache/hadoop/hive/ql/exec/CommonJoinOperator.java#L989-L996
https://github.com/apache/hive/blob/e7de9164557e6029fd845b8da0f78a6021e965bb/ql/src/java/org/apache/hadoop/hive/ql/exec/CommonJoinOperator.java#L856-L859

As HIVE-26018 explained, Hive on Tez produces incorrect results for UniqueJoin queries.

No

No

I added a new qfile and checked that both TestMiniLlapLocalCliDriver(Tez engine) and TestCliDriver(MR engine) produce the same result for it.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/CommonJoinOperator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/CommonJoinOperator.java
@@ -949,6 +949,8 @@ public abstract class CommonJoinOperator<T extends JoinDesc> extends
         if (!alw.hasRows()) {
           alw.addRow(dummyObj[i]);
           hasNulls = true;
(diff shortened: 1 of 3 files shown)


deniskuzZ (approved):
LGTM +1

kasakrisz (approved):

PR #4326, opened 2023-05-16
HIVE-27346: Getting exception for wildcard (*) search for database and table name

DanielZhu58:
### What changes were proposed in this pull request?
Changed the function when user search database and table name in some certain pattern.

### Why are the changes needed?
To make sure that if * are used in searching for database and table name, there is no exception.

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
Unit tests.

PR #5273, opened 2024-05-31
HIVE-28292: Optimize SHOW TABLES|VIEWS statements

wecharyu:
Make some improvements in `SHOW TABLES` statement:
1. Push down table pattern to metastore
2. Request table names instead of table objects
3. Change index `UNIQUETABLE(TBL_NAME, DB_ID)` to `UNIQUETABLE (DB_ID, TBL_NAME)` and drop index `TBLS_N49(DB_ID)`

Improve the performance of `SHOW TABLES|VIEWS` statements.

No.

No.

Passing existing tests and do benchmark tests.
```bash
 java -jar hmsbench.jar -H localhost --savedata /tmp/benchdata --sanitize -N 100 -N 10000 -o bench_results.csv -C -d testbench_http --params=100 -M 'listTables.*'
```
- Before this patch
```bash
listTables      1.52074 1.50483 1.39134 1.90837 4.07209
listTables.100  1.68746 1.68214 1.57137 1.87610 2.93073
listTables.10000        16.9628 16.9202 16.4026 19.1795 1.77736
```
- After this patch
```bash
listTables      1.50856 1.49130 1.39420 1.76155 4.03957
listTables.100  1.66239 1.65599 1.52580 1.96438 3.59206
listTables.10000        10.5922 10.2885 9.74368 15.3253 9.76028
```


standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/client/TestTablesGetExists.java
```diff
@@ -286,13 +286,6 @@ public void testGetTables() throws Exception {
     Assert.assertTrue("Comparing tablenames", tables.contains(testTables[5].getTableName()));
     Assert.assertTrue("Comparing tablenames", tables.contains(testTables[6].getTableName()));
 
-    tables = client.getTables("*", "*");
-    Assert.assertEquals("All tables size", 7, tables.size());
-    tables = client.getTables("d*", "*");
-    Assert.assertEquals("All tables size", 7, tables.size());
-    tables = client.getTables("def*", "*");
-    Assert.assertEquals("All tables size", 5, tables.size());
```
wecharyu, line 294: Remove these tests, discussed in https://github.com/apache/hive/pull/4326#issuecomment-2143551034.

wecharyu:
@dengzhhu653 @deniskuzZ @zhangbutao Could you please help review this patch?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesFormatter.java
```diff
@@ -51,7 +52,7 @@ public static ShowTablesFormatter getFormatter(HiveConf conf) {
 
   public abstract void showTables(DataOutputStream out, List<String> tables) throws HiveException;
 
-  abstract void showTablesExtended(DataOutputStream out, List<Table> tables) throws HiveException;
+  abstract void showTablesExtended(DataOutputStream out, List<Pair<String, String>> tables) throws HiveException;
```
deniskuzZ, line 55: wouldn't it be better to use `TableName`?

deniskuzZ, line 55: why is that change? if you'd like to avoid using fat class, maybe try org.apache.hadoop.hive.metastore.api.Table?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java
```diff
@@ -1315,4 +1316,35 @@ public static String getHttpPath(String httpPath) {
     }
     return httpPath;
   }
+
+  public static String convertSqlPatternToRegExp(String likePattern) {
```
deniskuzZ, line 1320: it looks like a copy&paste from `UDFLike.likePatternToRegExp`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesOperation.java
```diff
@@ -78,20 +75,22 @@ private void showTables() throws HiveException {
...
 
   private void showTablesExtended() throws HiveException {
-    List<Table> tableObjects = new ArrayList<>();
-    tableObjects.addAll(context.getDb().getTableObjects(desc.getDbName(), null, desc.getTypeFilter()));
-    if (desc.getPattern() != null) {
-      Pattern pattern = Pattern.compile(UDFLike.likePatternToRegExp(desc.getPattern()), Pattern.CASE_INSENSITIVE);
-      tableObjects = tableObjects.stream()
-          .filter(object -> pattern.matcher(object.getTableName()).matches())
-          .collect(Collectors.toList());
+    List<Pair<String, String>> tableTypePairs = new ArrayList<>();
```
deniskuzZ, line 78: can't you use Map here?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesOperation.java
```diff
@@ -78,20 +75,22 @@ private void showTables() throws HiveException {
...
+    String pattern = MetaStoreUtils.convertSqlPatternToRegExp(desc.getPattern());
+    TableType typeFilter = desc.getTypeFilter();
+    TableType[] tableTypes = typeFilter == null ? TableType.values() : new TableType[]{typeFilter};
```
deniskuzZ, line 81: why is this magic with typeFilter?

ql/src/test/results/clientpositive/llap/authorization_privilege_objects.q.out
```diff
@@ -42,8 +42,8 @@ PREHOOK: query: SHOW TABLES IN test_auth_obj_db
 PREHOOK: type: SHOWTABLES
 PREHOOK: Input: database:test_auth_obj_db
 filterListCmdObjects
-HIVE PRIVILEGE OBJECT { objectName: test_privs type: TABLE_OR_VIEW actionType: OTHER dbName: test_auth_obj_db OWNER: testuser OWNERTYPE: USER}
-HIVE PRIVILEGE OBJECT { objectName: test_privs2 type: TABLE_OR_VIEW actionType: OTHER dbName: test_auth_obj_db OWNER: testuser2 OWNERTYPE: USER}
+HIVE PRIVILEGE OBJECT { objectName: test_privs type: TABLE_OR_VIEW actionType: OTHER dbName: test_auth_obj_db}
+HIVE PRIVILEGE OBJECT { objectName: test_privs2 type: TABLE_OR_VIEW actionType: OTHER dbName: test_auth_obj_db}
```
deniskuzZ, line 46: why owner information is missing?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -2897,26 +2897,44 @@ public List<String> getTables(String dbname, String tablePattern) throws MetaExc
...
-      req.setProcessorIdentifier(processorIdentifier);
-    req.setProjectionSpec(projectionsSpec);
-    List<Table> tableObjects = client.get_table_objects_by_name_req(req).getTables();
-    tableObjects = deepCopyTables(FilterUtils.filterTablesIfEnabled(isClientFilterEnabled, filterHook, tableObjects));
-    for (Table tbl : tableObjects) {
-      tables.add(tbl.getTableName());
+    List<String> tables = new ArrayList<>();
+    Database db = null;
+    try {
+      db = getDatabase(catName, dbName);
```
deniskuzZ, line 2906: is that an extra db call? what do we save here?

standalone-metastore/metastore-server/src/main/sql/derby/hive-schema-4.1.0.derby.sql
```diff
@@ -269,7 +269,7 @@ CREATE UNIQUE INDEX "APP"."ROLEENTITYINDEX" ON "APP"."ROLES" ("ROLE_NAME");
 
 CREATE INDEX "APP"."TABLEPRIVILEGEINDEX" ON "APP"."TBL_PRIVS" ("AUTHORIZER", "TBL_ID", "PRINCIPAL_NAME", "PRINCIPAL_TYPE", "TBL_PRIV", "GRANTOR", "GRANTOR_TYPE");
 
-CREATE UNIQUE INDEX "APP"."UNIQUETABLE" ON "APP"."TBLS" ("TBL_NAME", "DB_ID");
+CREATE UNIQUE INDEX "APP"."UNIQUETABLE" ON "APP"."TBLS" ("DB_ID", "TBL_NAME");
```
deniskuzZ, line 272: what does it change?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -2897,26 +2897,44 @@ public List<String> getTables(String dbname, String tablePattern) throws MetaExc
...
+      // This branch can be removed once it's supported.
+      GetProjectionsSpec projectionsSpec = new GetProjectionsSpec();
+      projectionsSpec.setFieldList(Arrays.asList("dbName", "tableName", "owner", "ownerType"));
+      GetTablesRequest req = new GetTablesRequest(dbName);
+      req.setCatName(catName);
+      req.setCapabilities(version);
+      req.setTblNames(null);
+      req.setTablesPattern(tablePattern);
+      if (processorCapabilities != null)
+        req.setProcessorCapabilities(new ArrayList<String>(Arrays.asList(processorCapabilities)));
```
deniskuzZ, line 2920: what is that double ArrayList?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesFormatter.java
wecharyu: `SHOW TABLES EXTENDED` only outputs table name and table type, so we do not need the `Table` objects here.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesOperation.java
wecharyu: It means to request all types table names if not input type filter in statement.

ql/src/test/results/clientpositive/llap/authorization_privilege_objects.q.out
wecharyu: since we turns to request only table names instead of table objects, so here auth check will not need owner info.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
wecharyu: Yes, we need it to determine if this is a remote database which does not support table pattern yet.

standalone-metastore/metastore-server/src/main/sql/derby/hive-schema-4.1.0.derby.sql
wecharyu: We change the order of this composite index to improve performance. Because we always get the`DB_ID` first while doing any query for hive table, so the first level (DB_ID) of the index alway take effect after this change.

It's a similar change like https://github.com/apache/hive/pull/4831#pullrequestreview-1814098075

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
wecharyu: It's copied from source, removed redundant ArrayList.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesOperation.java
wecharyu: Good idea, addressed.

wecharyu:
@deniskuzZ could you take a look again?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesOperation.java
```diff
@@ -78,20 +74,20 @@ private void showTables() throws HiveException {
...
-          .filter(object -> pattern.matcher(object.getTableName()).matches())
-          .collect(Collectors.toList());
+    TreeMap<String, String> tableNameToType = new TreeMap<>();
+    String pattern = MetaStoreUtils.convertSqlPatternToRegExp(desc.getPattern());
+    TableType typeFilter = desc.getTypeFilter();
+    TableType[] tableTypes = typeFilter == null ? TableType.values() : new TableType[]{typeFilter};
+    for (TableType tableType : tableTypes) {
+      String typeString = tableType.toString();
+      List<String> tables = context.getDb().getTablesByType(desc.getDbName(), pattern, tableType);
+      tables.forEach(name -> tableNameToType.put(name, typeString));
```
deniskuzZ, line 84: minor: could we inline local var: `tableNameToType.put(name, tableType.toString())` ?

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
```diff
@@ -321,7 +321,7 @@ public List<String> getTables(String dbName, String tablePattern) throws MetaExc
     if (tables == null || tables.size() == 0) {
       return tableNames;
     }
-    tablePattern = tablePattern.replaceAll("\\*", ".*");
+    tablePattern = tablePattern.replaceAll("\\.\\*", "\\*").replaceAll("\\*", ".*");
```
deniskuzZ, line 324: i don't follow, you replace `.*` with `*` and then `*` with `.*` ?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -2897,26 +2897,44 @@ public List<String> getTables(String dbname, String tablePattern) throws MetaExc
...
+      tableObjects = deepCopyTables(FilterUtils.filterTablesIfEnabled(isClientFilterEnabled, filterHook, tableObjects));
+      for (Table tbl : tableObjects) {
+        tables.add(tbl.getTableName());
+      }
+    } else {
+      // This trick handles pattern for both string regex and wildcards ('*' and '|').
+      // We need unify the pattern definition, see HIVE-28297 for details.
+      String[] patterns = tablePattern.split("\\|");
+      for (String pattern : patterns) {
+        pattern = "(?i)" + pattern.replaceAll("\\.\\*", "\\*").replaceAll("\\*", ".*");
```
deniskuzZ, line 2934: what is `(?i)` used for and 2 replaceAll?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -2897,26 +2897,44 @@ public List<String> getTables(String dbname, String tablePattern) throws MetaExc
...
+        String filter = hive_metastoreConstants.HIVE_FILTER_FIELD_TABLE_NAME + " like \"" + pattern + "\"";
```
deniskuzZ, line 2935: why not use String.format("like \"%s\"", hive_metastoreConstants.HIVE_FILTER_FIELD_TABLE_NAM)

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java
deniskuzZ: ok, why not add param `bool quote` and reuse the method

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -4941,6 +4929,14 @@ private Table ensureGetTable(String catName, String dbName, String tblName)
     return convertToTable(ensureGetMTable(catName, dbName, tblName));
   }
 
+  private Database ensureGetDatabase(String catName, String dbName) throws UnknownDBException {
```
deniskuzZ, line 4932: is that used in multiple places? if not just inline into method where it's used

standalone-metastore/metastore-server/src/main/sql/derby/hive-schema-4.1.0.derby.sql
deniskuzZ: are you saying we won't hit the index if the order of columns is reversed?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
wecharyu: `(?i)` means case-insensitive in regex matching.

The 2 replaceAll is a bit confused. It's needed because in some unit tests they use wildcards (`*` and `|`) for table pattern and regard `*` as all (which should be converted to `.*` in string regex.

standalone-metastore/metastore-server/src/main/sql/derby/hive-schema-4.1.0.derby.sql
wecharyu: It could only hit the first part (`TBL_NAME`) of `UNIQUETABLE` index currently if we query tables with table pattern because it will use `like clause` for `TBL_NAME`. If we switch the order, both `DB_ID` and `TBL_NAME` can be hit.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java
wecharyu: Addressed, need make some minor change on this method.

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
deniskuzZ: @wecharyu, could we replace it with `tablePattern.replaceAll("(?<!\\.)\\*", ".*")` ?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
deniskuzZ: could we use negative lookbehind here as well?

deniskuzZ (commented):
LGTM, sorry for a few month delay. 1 suggestion on regex simplification

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java
```diff
@@ -77,9 +80,9 @@ public static String likePatternToRegExp(String likePattern) {
       if (n == '_') {
         sb.append(".");
       } else if (n == '%') {
-        sb.append(".*?");
+        sb.append(".*");
```
deniskuzZ, line 83: I am not sure about this change, should we keep the matching lazy?

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java
```diff
@@ -61,7 +61,10 @@ private enum PatternType {
   public UDFLike() {
   }
 
-  public static String likePatternToRegExp(String likePattern) {
+  public static String likePatternToRegExp(String likePattern, boolean quote) {
```
deniskuzZ, line 64: since it's used in multiple places maybe you can keep the default impl with 1 param
````
public static String likePatternToRegExp(likePattern){
  return likePatternToRegExp(likePattern, true);
}
````

Comment (deniskuzZ) [inline ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java:64]: maybe `literalize` instead of `quote`

File: ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java (hunk 15 above)
Comment (wecharyu) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java:83]: Reverted. We only use the pattern for string matching, is reasonable to use the non-greedy matching.

File: ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java (hunk 16 above)
Comment (wecharyu) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java:64]: Addressed.

**Review by deniskuzZ — APPROVED** [review]
LGTM +1, pending tests

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesFormatter.java [hunk 17]
```diff
@@ -51,7 +52,7 @@ public static ShowTablesFormatter getFormatter(HiveConf conf) {
 
   public abstract void showTables(DataOutputStream out, List<String> tables) throws HiveException;
 
-  abstract void showTablesExtended(DataOutputStream out, List<Table> tables) throws HiveException;
+  abstract void showTablesExtended(DataOutputStream out, TreeMap<String, String> tableNameToType) throws HiveException;
```
deniskuzZ, line 55: please code to an interface, use `Map`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesFormatter.java
```diff
@@ -62,16 +63,16 @@ public void showTables(DataOutputStream out, List<String> tables) throws HiveExc
     }
 
     @Override
-    void showTablesExtended(DataOutputStream out, List<Table> tables) throws HiveException {
-      if (tables.isEmpty()) {
+    void showTablesExtended(DataOutputStream out, TreeMap<String, String> tableNameToType) throws HiveException {
```
deniskuzZ, line 66: `Map`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesFormatter.java
```diff
@@ -96,8 +97,8 @@ public void showTables(DataOutputStream out, List<String> tables) throws HiveExc
...
```
deniskuzZ, line 100: `Map`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesOperation.java
```diff
@@ -78,20 +73,19 @@ private void showTables() throws HiveException {
...
 
   private void showTablesExtended() throws HiveException {
-    List<Table> tableObjects = new ArrayList<>();
-    tableObjects.addAll(context.getDb().getTableObjects(desc.getDbName(), null, desc.getTypeFilter()));
-    if (desc.getPattern() != null) {
-      Pattern pattern = Pattern.compile(UDFLike.likePatternToRegExp(desc.getPattern()), Pattern.CASE_INSENSITIVE);
-      tableObjects = tableObjects.stream()
-          .filter(object -> pattern.matcher(object.getTableName()).matches())
-          .collect(Collectors.toList());
+    TreeMap<String, String> tableNameToType = new TreeMap<>();
```
deniskuzZ, line 76: `Map`

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFLike.java
wecharyu: JDO does not support non-greedy matching :(

Add a new parameter `greedyMatch` for this option.

PR #5275, opened 2024-06-03
HIVE-28294: drop database cascade operation can skip client side filt…

saihemanth-cloudera:
…ering while fetching tables in db

Skip client-side filtering for drop database operation, so that all the tables can be authorized for drop database cascade operation.

some tables in db if filtered out can be missed while checking for drop privileges.

No

No

Manually verified that tables are not getting filtered out.


ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java
```diff
@@ -61,7 +62,7 @@ public void analyzeInternal(ASTNode root) throws SemanticException {
     boolean isDbLevelLock = true;
     if (cascade) {
       try {
-        List<Table> tables = db.getAllTableObjects(databaseName);
+        List<Table> tables = db.getAllTableObjects(databaseName, true);
```
dengzhhu653, line 65: There is a `hive.metastore.client.filter.enabled` to disable the client filter, we can use it instead of depend on a new method, e.g
```java
try {
   HiveConf hiveConf = new HiveConf(conf);
   conf.set("hive.metastore.client.filter.enabled", "false");
   Hive db = Hive.get(hiveConf);
   ....
} finally {
   // restore
   db = Hive.get(conf);
 }

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java [hunk 2]
```diff
@@ -73,6 +74,11 @@ public void analyzeInternal(ASTNode root) throws SemanticException {
           }
           outputs.add(new WriteEntity(table, lockType));
         }
+        // fetch all the functions in the database
+        List<Function> functions = db.getFunctionsInDb(databaseName, ".*");
```
dengzhhu653, line 78: Given in reality, the number of permanent functions stored in remote db should be smaller compared to the other meta, I think we can use `db.getAllFunctions` then select the functions belonging to this database.

dengzhhu653:
I'm trying to understand the problem, why we need "all the tables can be authorized for drop database cascade", given that we have the drop privilege on the database?

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java
```diff
@@ -60,7 +62,13 @@ public void analyzeInternal(ASTNode root) throws SemanticException {
     // if cascade=true, then we need to authorize the drop table action as well, and add the tables to the outputs
     boolean isDbLevelLock = true;
     if (cascade) {
+      SessionState sessionState = SessionState.get();
       try {
+        HiveConf hiveConf = new HiveConf(conf);
+        hiveConf.set("hive.metastore.client.filter.enabled", "false");
+        if (sessionState != null) {
```
dengzhhu653, line 69: Not sure if the current `db` would take this property, as this `db` is created by the original `conf` which doesn't have this property.

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java
dengzhhu653: On HS2 restart, we load all functions from HMS, 
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L364
I think this call is much light weight than the `getAllTableObjects` in this method.
Another point is the `drop database` is a long consuming call as expected and less frequent, I'm a little worry about if it's worth to add an extra API just to fetch all the functions in this database.

saihemanth-cloudera: I still think this method would be useful if we decide to enforce client-side filtering on functions in the future. @nrg4878 -- Can you please share your suggestion here on whether to add a new API for fetching functions in a database or not?

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java
saihemanth-cloudera: I'll implement this suggestion as setting config in session is not taken into effect. we don't need a finally block since we are creating a new variable 'db' instead of overriding the already existing one. Also, the existing variable 'db' is a final variable so we cannot override it.

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java
saihemanth-cloudera: Ack.

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java
dengzhhu653: The new created `db` will overwrite the thread-local hive, and the underlying client disables the client filtering, if we use this client for other calls in the future, it may cause problems.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -3008,6 +3007,15 @@ Function getFunction(String catName, String dbName, String funcName)
...
       throws MetaException, TException;
 
+  /**
+   * Get all functions matching a pattern
+   * @param dbName database name.
+   * @param pattern to match.  This is a java regex pattern.
+   * @throws MetaException error accessing the RDBMS
+   * @throws TException thrift transport error
+   */
+  List<Function> getFunctionsInDb(String dbName, String pattern)
```
dengzhhu653, line 3017: how about change it to `GetFunctionsResponse getFunctions(..` , and mark the `List<String> getFunctions` as deprecated?

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
```diff
@@ -3028,6 +3040,8 @@ PartitionsResponse get_partitions_req(1:PartitionsRequest req)
 
   list<string> get_functions(1:string dbName, 2:string pattern)
       throws (1:MetaException o1)
+  GetFunctionsResponse get_functions_in_db(1:GetFunctionsRequest request)
```
dengzhhu653, line 3043: `get_functions_in_db` -> `get_functions_req`?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -4727,6 +4727,16 @@ public List<String> getFunctions(String catName, String dbName, String pattern)
...
   }
 
+  @Override
+  public List<Function> getFunctionsInDb(String dbName, String pattern) throws TException {
+    GetFunctionsRequest request = new GetFunctionsRequest(dbName);
+    request.setCatalogName(getDefaultCatalog(conf));
+    request.setPattern(pattern);
+    request.setReturnNames(false);
+    return client.get_functions_in_db(request).getFunctions();
+//    return client.get_functions_in_db(prependCatalogToDbName(getDefaultCatalog(conf), dbName, conf), pattern);
```
dengzhhu653, line 4737: nit: remove this comment?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
```diff
@@ -1623,6 +1623,15 @@ void dropFunction(String catName, String dbName, String funcName)
...
   List<String> getFunctions(String catName, String dbName, String pattern) throws MetaException;
 
+  /**
+   * Retrieve list of function names based on name pattern.
+   * @param dbName database name
+   * @param pattern pattern to match
+   * @return functions that match the pattern
+   * @throws MetaException incorrectly specified function
+   */
+  List<Function> getFunctionsInDb(String catName, String dbName, String pattern) throws MetaException;
```
dengzhhu653, line 1633: I was thinking how we can merge `List<String> getFunctions` and `List<Function> getFunctionsInDb`, can we change it to ` <T> void getFunctions(String catName, String dbName, String pattern, List<T> result) throws MetaException;` and mark the original `getFunctions` as deprecated?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -8884,6 +8884,30 @@ public List<String> get_functions(String dbName, String pattern)
...
 
+  @Override
+  public GetFunctionsResponse get_functions_in_db(GetFunctionsRequest req)
+      throws MetaException {
+    startFunction("get_functions", ": db=" + req.getDbName()
+        + " pat=" + req.getPattern());
+
+    RawStore ms = getMS();
+    Exception ex = null;
+    List<Function> funcNames = null;
```
dengzhhu653, line 8895: nit: funcNames -> funcs

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -8884,6 +8884,30 @@ public List<String> get_functions(String dbName, String pattern)
...
+
+    try {
+      funcNames = ms.getFunctionsInDb(req.getCatalogName(), req.getDbName(), req.getPattern());
```
dengzhhu653, line 8898: how about if `returnNames=false`?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
saihemanth-cloudera: Ack

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
saihemanth-cloudera: ack

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
saihemanth-cloudera: ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
saihemanth-cloudera: ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: ack

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -3005,9 +3004,19 @@ Function getFunction(String catName, String dbName, String funcName)
...
       throws MetaException, TException;
 
+  /**
+   * Get all functions matching a pattern
+   * @param dbName database name.
+   * @param pattern to match.  This is a java regex pattern.
+   * @throws MetaException error accessing the RDBMS
+   * @throws TException thrift transport error
+   */
+  GetFunctionsResponse getFunctionsRequest(String dbName, String pattern)
```
dengzhhu653, line 3018: Change the args to `GetFunctionsRequest`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -8884,6 +8884,30 @@ public List<String> get_functions(String dbName, String pattern)
...
+  public GetFunctionsResponse get_functions_req(GetFunctionsRequest req)
+      throws MetaException {
+    startFunction("get_functions", ": db=" + req.getDbName()
+        + " pat=" + req.getPattern());
+
+    RawStore ms = getMS();
+    Exception ex = null;
+    List<Function> funcs = null;
+    try {
+      funcs = ms.getFunctionsRequest(req.getCatalogName(), req.getDbName(),
```
dengzhhu653, line 8897: the `catalogName` might be null

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -8884,6 +8884,30 @@ public List<String> get_functions(String dbName, String pattern)
     return funcNames;
   }
 
+  @Override
+  public GetFunctionsResponse get_functions_req(GetFunctionsRequest req)
+      throws MetaException {
+    startFunction("get_functions", ": db=" + req.getDbName()
```
dengzhhu653, line 8890: nit: `get_functions` -> `get_functions_req`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -11878,6 +11878,45 @@ public List<String> getFunctions(String catName, String dbName, String pattern)
...
+      appendSimpleCondition(filterBuilder, "database.catalogName", new String[] {catName}, parameterVals);
+      if(pattern != null) {
+        appendPatternCondition(filterBuilder, "functionName", pattern, parameterVals);
+      }
+      query = pm.newQuery(MFunction.class, filterBuilder.toString());
+      if (isReturnNames) {
+        query.setResult("functionName");
+      }
+      query.setOrdering("functionName ascending");
+      List<MFunction> functionList = (List<MFunction>) query.executeWithArray(parameterVals.toArray(new String[0]));
```
dengzhhu653, line 11903: if the `isReturnNames` is true, the result should be `List<String>`?

saihemanth-cloudera: Yeah, we fetch Mfunction in case of function names: https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L11866-L11868 and get function name from it.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: ack

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
saihemanth-cloudera: ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
dengzhhu653: I mean if the `isReturnNames` is true, the result of `query.executeWithArray(parameterVals.toArray(new String[0]));` should be a `List<String>`?

saihemanth-cloudera: Ack

dengzhhu653 (approved):

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -8884,6 +8884,31 @@ public List<String> get_functions(String dbName, String pattern)
...
+      throws MetaException {
+    startFunction("get_functions_req", ": db=" + req.getDbName()
+        + " pat=" + req.getPattern());
+
+    RawStore ms = getMS();
+    Exception ex = null;
+    List<Function> funcs = null;
+    String catName = req.isSetCatalogName() ? req.getCatalogName() : getDefaultCatalog(conf);
+    try {
+      funcs = ms.getFunctionsRequest(catName, req.getDbName(),
```
dengzhhu653, line 8898: if the `req.isReturnNames()` is true, you need to return the function names

PR #5276, opened 2024-06-04
HIVE-28299: Iceberg: Optimize show partitions through column projection

zhangbutao:

In the current show partitions implementation, we need to fetch all columns data(11 columns), but in fact we only need two columns data, `partition `& `spec_id`.
https://github.com/apache/hive/blob/45867be6cb5308566e4cf16c7b4cf8081085b58c/common/src/java/org/apache/hadoop/hive/conf/Constants.java#L108-L110

We can only fetch the two columns through column project, and this can improve the performance in case of big iceberg partition table.

This change can make the `show partitions` more faster.

No

No

Existing Tests.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -118,6 +118,7 @@ import org.apache.hadoop.hive.ql.session.SessionState;
 import org.apache.hadoop.hive.ql.session.SessionStateUtil;
 import org.apache.hadoop.hive.ql.stats.Partish;
 import org.apache.hadoop.hive.serde2.AbstractSerDe;
+import org.apache.hadoop.hive.serde2.ColumnProjectionUtils;
 import org.apache.hadoop.hive.serde2.DefaultFetchFormatter;
 import org.apache.hadoop.hive.serde2.Deserializer;
 import org.apache.hadoop.hive.serde2.FetchFormatter;
@@ -1804,7 +1805,9 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
       fetcher.initialize(job, HiveTableUtil.getSerializationProps());
       org.apache.hadoop.hive.ql.metadata.Table metaDataPartTable =
           context.getDb().getTable(hmstbl.getDbName(), hmstbl.getTableName(), "partitions", true);
+      SessionState.getSessionConf().set(ColumnProjectionUtils.READ_COLUMN_NAMES_CONF_STR, "partition,spec_id");
       Deserializer currSerDe = metaDataPartTable.getDeserializer();
+      SessionState.getSessionConf().unset(ColumnProjectionUtils.READ_COLUMN_NAMES_CONF_STR);
       ObjectMapper mapper = new ObjectMapper();
       Table tbl = getTable(hmstbl);
       while (reader.next(key, value)) {
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -58,6 +58,7 @@ import org.apache.hadoop.hive.conf.Constants;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
 import org.apache.hadoop.hive.metastore.HiveMetaHook;
+import org.apache.hadoop.hive.metastore.HiveMetaStoreUtils;
 import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.ColumnStatistics;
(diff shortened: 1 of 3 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1840,7 +1841,9 @@ private List<String> getPartitions(DDLOperationContext context, Configuration jo
       fetcher.initialize(job, HiveTableUtil.getSerializationProps());
       org.apache.hadoop.hive.ql.metadata.Table metaDataPartTable =
           context.getDb().getTable(hmstbl.getDbName(), hmstbl.getTableName(), "partitions", true);
+      SessionState.getSessionConf().set(ColumnProjectionUtils.READ_COLUMN_NAMES_CONF_STR, "partition,spec_id");
```
zhangbutao, line 1844: https://github.com/apache/hive/blob/45867be6cb5308566e4cf16c7b4cf8081085b58c/ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java#L343-L344
`HiveIcebergSerDe `can only use the `session conf`, so here i set the `ColumnProjectionUtils.READ_COLUMN_NAMES_CONF_STR` in the session conf, and then unset this propertity to avoid affecting other sql queries in the current seesion.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java
```diff
@@ -302,7 +302,7 @@ static String getParseData(String parseData, String specId, ObjectMapper mapper,
 
   static JobConf getPartJobConf(Configuration confs, org.apache.hadoop.hive.ql.metadata.Table tbl) {
     JobConf job = new JobConf(confs);
-    job.set(ColumnProjectionUtils.READ_COLUMN_NAMES_CONF_STR, Constants.ICEBERG_PARTITION_TABLE_SCHEMA);
+    job.set(ColumnProjectionUtils.READ_COLUMN_NAMES_CONF_STR, "partition,spec_id");
```
zhangbutao, line 305: This can be used by `scan ` & `getSplits `stage, to fetch the related two columns:
https://github.com/apache/hive/blob/4bc36eb40ebd16fb0b79f19ec4471549cb42a177/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergInputFormat.java#L152-L153

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1840,7 +1841,8 @@ private List<String> getPartitions(DDLOperationContext context, Configuration jo
       fetcher.initialize(job, HiveTableUtil.getSerializationProps());
       org.apache.hadoop.hive.ql.metadata.Table metaDataPartTable =
           context.getDb().getTable(hmstbl.getDbName(), hmstbl.getTableName(), "partitions", true);
-      Deserializer currSerDe = metaDataPartTable.getDeserializer();
+      Deserializer currSerDe = HiveMetaStoreUtils.getDeserializer(job, metaDataPartTable.getTTable(),
```
zhangbutao, line 1844: Use job conf to get the iceberg Deserializer so that we can propagate `ColumnProjectionUtils.READ_COLUMN_NAMES_CONF_STR` to `HiveIcebergSerDe`
https://github.com/apache/hive/blob/e570e65a397aae47c62472f12acaaed06b68cf54/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergSerDe.java#L163

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1840,7 +1841,8 @@ private List<String> getPartitions(DDLOperationContext context, Configuration jo
...
+          metaDataPartTable.getMetaTable(), false);
```
deniskuzZ, line 1845: why do we need metatable?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java
deniskuzZ, line 305: `ImmutableList.of("partition", "spec_id")` could we move to constants `PARTITION_COLUMNS`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
zhangbutao: iceberg `show partitons` will use the iceberg metadata table `partitions`.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java
zhangbutao: Fixed. Thanks.

PR #5277, opened 2024-06-04
HIVE-28300: Fix AlterTableConcatenate when using Hive-Tez

ngsg:

Modify AlterTableConcatenateOperation to refer to MergeFileWork.getInputPaths().

When using a List Bucketing table, FileMergeOperator expects only 1 subdirectory as its input. To satisfy this requirement, AlterTableConcatenateOperation changes input paths by calling MergeFileWork.resolveConcatenateMerge(). But Hive-Tez does not properly propagate the changed input paths to SplitGenerator, which causes the problem

No

No

Using list_bucket_dml_8.q.
This PR changes the test driver for list_bucket_dml_8.q from TestCliDriver to TestMiniLlapLocalCliDriver, as all the other list_bucket_dml_*.q are already run by TestMiniLlapLocalCliDriver. Also MR specific config keys are replaced by corresponding Tez specific keys.
I changed the file format of table from RCFile to ORC because ORC is more popular than RCFile and the issue comes from their common class: AbstractFileMergeOperator.

### DIFF
--- a/itests/src/test/resources/testconfiguration.properties
+++ b/itests/src/test/resources/testconfiguration.properties
@@ -237,7 +237,6 @@ mr.query.files=\
   inputwherefalse.q,\
   join_map_ppr.q,\
   join_vc.q,\
-  list_bucket_dml_8.q,\
   localtimezone.q,\
   manyViewJoin.q,\
   mapjoin1.q,\
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/concatenate/AlterTableConcatenateOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/concatenate/AlterTableConcatenateOperation.java
@@ -76,7 +76,9 @@ public class AlterTableConcatenateOperation extends DDLOperation<AlterTableConca
 
     Map<Path, List<String>> pathToAliases = new LinkedHashMap<>();
     List<String> inputDirStr = Lists.newArrayList(inputDirList.toString());
-    pathToAliases.put(desc.getInputDir(), inputDirStr);
+    for (Path path: mergeWork.getInputPaths()) {
+      pathToAliases.put(path, inputDirStr);
+    }
     mergeWork.setPathToAliases(pathToAliases);
 
     FileMergeDesc fmd = getFileMergeDesc();
--- a/ql/src/test/queries/clientpositive/list_bucket_dml_8.q
+++ b/ql/src/test/queries/clientpositive/list_bucket_dml_8.q
@@ -1,11 +1,10 @@
 --! qt:dataset:srcpart
 set hive.mapred.mode=nonstrict;
 set hive.exec.dynamic.partition=true;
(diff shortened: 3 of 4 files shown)


itests/src/test/resources/testconfiguration.properties
```diff
@@ -237,7 +237,6 @@ mr.query.files=\
   inputwherefalse.q,\
   join_map_ppr.q,\
   join_vc.q,\
-  list_bucket_dml_8.q,\
```
SourabhBadhya, line 240: Why remove this?

ql/src/test/queries/clientpositive/list_bucket_dml_8.q
```diff
@@ -1,11 +1,10 @@
 --! qt:dataset:srcpart
 set hive.mapred.mode=nonstrict;
 set hive.exec.dynamic.partition=true;
-set hive.input.format=org.apache.hadoop.hive.ql.io.BucketizedHiveInputFormat;
+set hive.tez.input.format=org.apache.hadoop.hive.ql.io.BucketizedHiveInputFormat;
 set hive.merge.smallfiles.avgsize=200;
 set mapred.input.dir.recursive=true;
-set hive.merge.mapfiles=false;	
-set hive.merge.mapredfiles=false;
+set hive.merge.tezfiles=false;
```
SourabhBadhya, line 7: So if we use MR it wont work?

SourabhBadhya (commented):
The patch looks good. Minor comments on MR.

itests/src/test/resources/testconfiguration.properties
ngsg: If a qfile is not contained in this list, then TestMiniLlapLocalCliDriver(the default dirver of qtest) pickes it.

ql/src/test/queries/clientpositive/list_bucket_dml_8.q
ngsg: Yes. This configuration lets Hive create multiple small files and prevents merging them until `ALTER ... CONCATENATE` command. So running the modified qfile with MR produces wrong result.

deniskuzZ (commented):
LGTM +1

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/concatenate/AlterTableConcatenateOperation.java
```diff
@@ -76,7 +76,9 @@ private MergeFileWork getMergeFileWork(CompilationOpContext opContext) throws Hi
 
     Map<Path, List<String>> pathToAliases = new LinkedHashMap<>();
     List<String> inputDirStr = Lists.newArrayList(inputDirList.toString());
-    pathToAliases.put(desc.getInputDir(), inputDirStr);
+    for (Path path: mergeWork.getInputPaths()) {
```
deniskuzZ, line 79: @ngsg, `mergeWork.getInputPaths()` and `desc.getInputDir()` are the same, isn't it?
````
mergeWork = new MergeFileWork(Lists.newArrayList(desc.getInputDir())
````

PR #5280, opened 2024-06-05
HIVE-28304: Replace org.apache.log4j.Logger with org.slf4j.Logger

ngsg:

 Replace org.apache.log4j.Logger with org.slf4j.Logger.

Logger should be created via facade(SLF4J), but some classes still directly access to Logger implementation. This patch fixes this issue.

No

No

This patch was not tested as it just changes logger instantiation code.

### DIFF
--- a/accumulo-handler/src/test/org/apache/hadoop/hive/accumulo/TestHiveAccumuloHelper.java
+++ b/accumulo-handler/src/test/org/apache/hadoop/hive/accumulo/TestHiveAccumuloHelper.java
@@ -31,14 +31,15 @@ import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.token.Token;
 import org.apache.hadoop.security.token.TokenIdentifier;
-import org.apache.log4j.Logger;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
+import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


okumin (approved):

zhangbutao (commented):
Are there other places where `org.apache.log4j.Logger` is being used?
If not, can we remove the `log4j-1.2-api ` dependency?  For example, I see that hive install package has the jar `log4j-1.2-api-2.18.0.jar`.

ngsg:
@zhangbutao, there is no `org.apache.log4j.Logger` import in Hive project.
Let me test Hive if we could remove `log4j-1.2-api` dependency or not.

ngsg:
I tried compiling Hive without `log4j-1.2-api`, but the compilation failed as some code still rely on it. So we can not remove `log4j-1.2-api` dependency as of now.

maven failure message:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project hive-common: Compilation failure
[ERROR] /data2/hive-lngsg/common/src/java/org/apache/hadoop/hive/common/JvmMetrics.java:[185,20] cannot access org.apache.log4j.AppenderSkeleton
[ERROR]   class file for org.apache.log4j.AppenderSkeleton not found
```

zhangbutao (approved):
LGTM +1

zhangbutao:
@okumin Sorry for forgetting add your name into the commit message. :(

PR #5281, opened 2024-06-05
HIVE-28305: ProfileServlet: add html to output formats and prepare for profiler 3.0

abstractdog:
Added HTML format support to ProfileServlet.

Because HTML is the default format for creating flamegraphs.

No, profile servlet is more like a dev tool.

No.

set web port for hs2 in hive-site.xml
```
+<property>
+  <name>hive.server2.webui.port</name>
+  <value>10003</value>
+</property>
```

```
#tried with both 2.9 and 3.0
export ASYNC_PROFILER_HOME=/Users/laszlobodor/Downloads/async-profiler-2.9-macos
export ASYNC_PROFILER_HOME=/Users/laszlobodor/Downloads/async-profiler-3.0-macos

 mvn install -Dtest=StartMiniHS2Cluster -DminiHS2.clusterType=llap -DminiHS2.conf="target/testconf/llap/hive-site.xml"  -DminiHS2.run=true -DminiHS2.usePortsFromConf=true -Dpackaging.minimizeJar=false -T 1C -DskipShade -Dremoteresources.skip=true -Dmaven.javadoc.skip=true -Denforcer.skip=true -pl itests/hive-unit -Pitests -nsu
```

called localhost:10003/prof and get the flamegraph successfully
https://issues.apache.org/jira/secure/attachment/13069348/async-prof-pid-21327-cpu-2.html

### DIFF
--- a/common/src/java/org/apache/hive/http/ProfileServlet.java
+++ b/common/src/java/org/apache/hive/http/ProfileServlet.java
@@ -17,6 +17,9 @@ package org.apache.hive.http;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


common/src/java/org/apache/hive/http/ProfileServlet.java
```diff
@@ -45,19 +47,19 @@
  * //  -m method         fully qualified method name: 'ClassName.methodName'
  * //  -t                profile different threads separately
  * //  -s                simple class names instead of FQN
- * //  -o fmt[,fmt...]   output format: summary|traces|flat|collapsed|svg|tree|jfr
+ * //  -o fmt[,fmt...]   output format: summary|traces|flat|collapsed|svg|tree|jfr|html
```
deniskuzZ, line 50: should we remove `svg` if that is not supported anymore?

abstractdog: it's supported in older versions of the profiler, so the user might want to use that

common/src/java/org/apache/hive/http/ProfileServlet.java
```diff
@@ -293,6 +295,16 @@ protected void doGet(final HttpServletRequest req, final HttpServletResponse res
     }
   }
 
+  /**
+   * Get the path of the profiler script to be executed.
+   * Before async-profiler 3.0, the script was named profiler.sh, and after 3.0 it's bin/asprof
+   * @return
+   */
+  private String getProfilerScriptPath() {
+    return Files.exists(Paths.get(asyncProfilerHome + "/bin/asprof")) ? (asyncProfilerHome + "/bin/asprof")
```
deniskuzZ, line 304:
````
Path defaultPath = Paths.get(asyncProfilerHome + "/bin/asprof");
return Files.exists(defaultPath)? defaultPath.toString() : asyncProfilerHome + "/profiler.sh"
````

File: common/src/java/org/apache/hive/http/ProfileServlet.java [hunk 3]
```diff
@@ -141,7 +142,8 @@ enum Output {
     COLLAPSED,
     SVG,
```
deniskuzZ, line 143: remove if not supported

deniskuzZ (approved):
LGTM, minor comment

kokila-19 (approved):

common/src/java/org/apache/hive/http/ProfileServlet.java
```diff
@@ -344,10 +357,11 @@ private Output getOutput(final HttpServletRequest req) {
       try {
         return Output.valueOf(outputArg.trim().toUpperCase());
       } catch (IllegalArgumentException e) {
-        return Output.SVG;
+        LOG.warn("Output format value is invalid, returning with default HTML");
```
ayushtkn, line 360: Can you add what value is actually configured in the log as well
```
        LOG.warn("Output format value {} is invalid, returning with default HTML", outputArg);
```

ayushtkn (approved):
LGTM

common/src/java/org/apache/hive/http/ProfileServlet.java
abstractdog: makes sense!

abstractdog:
all tests passed, merging
https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5281/3/pipeline

PR #5282, opened 2024-06-05
HIVE-28306: Iceberg: Return new scan after applying column project parameter

zhangbutao:

1. Shoud return a new scan after applying column project parameter, Otherwise it doesn't make any sense

2. `scan.project()` and `scan.select()` can not be specified at the same time
Check this:
https://github.com/apache/iceberg/pull/1353/files#diff-a94dee2a22a2890689a84529a978a769bdc3b084f35d79caf9608d49b2e1cc1b

No

No

Existing tests.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
@@ -174,6 +174,11 @@ public class IcebergInputFormat<T> extends InputFormat<Void, T> {
       Long openFileCost = splitSize > 0 ? splitSize : TableProperties.SPLIT_SIZE_DEFAULT;
       scan = scan.option(TableProperties.SPLIT_OPEN_FILE_COST, String.valueOf(openFileCost));
     }
+    //  TODO: Currently, this projection optimization stored on scan is not being used effectively on Hive side, as
+    //   Hive actually uses conf to propagate the projected columns to let the final reader to read the only
+    //   projected columns data. See IcebergInputFormat::readSchema(Configuration conf, Table table, boolean
+    //   caseSensitive). But we can consider using this projection optimization stored on scan in the future when
+    //   needed.
     Schema readSchema = InputFormatConfig.readSchema(conf);
     if (readSchema != null) {
       scan = scan.project(readSchema);
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
@@ -177,12 +177,9 @@ public class IcebergInputFormat<T> extends InputFormat<Void, T> {
     }
     String schemaStr = conf.get(InputFormatConfig.READ_SCHEMA);
     if (schemaStr != null) {
-      scan.project(SchemaParser.fromJson(schemaStr));
-    }
-
-    String[] selectedColumns = conf.getStrings(InputFormatConfig.SELECTED_COLUMNS);
-    if (selectedColumns != null) {
-      scan.select(selectedColumns);
+      scan = scan.project(SchemaParser.fromJson(schemaStr));
+    } else if (conf.getStrings(InputFormatConfig.SELECTED_COLUMNS) != null) {
+      scan = scan.select(conf.getStrings(InputFormatConfig.SELECTED_COLUMNS));
     }
 
     // TODO add a filter parser to get rid of Serialization
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
@@ -59,7 +59,6 @@ import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Partitioning;
 import org.apache.iceberg.Scan;
 import org.apache.iceberg.Schema;
-import org.apache.iceberg.SchemaParser;
 import org.apache.iceberg.SnapshotRef;
 import org.apache.iceberg.StructLike;
 import org.apache.iceberg.Table;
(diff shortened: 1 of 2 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInternalRecordWrapper.java
```diff
@@ -57,7 +57,7 @@ public IcebergInternalRecordWrapper(StructType tableSchema, StructType readSchem
   public IcebergInternalRecordWrapper wrap(StructLike record) {
     int idx = 0;
     for (Types.NestedField field : readSchema.fields()) {
-      int position = fieldToPositionInTableSchema.get(field.name());
+      int position = fieldToPositionInReadSchema.get(field.name());
```
zhangbutao, line 60: Fixed the failed metadata qtest: http://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5282/1/tests/

zhangbutao:
Submitted a same PR in Apache Iceberg: https://github.com/apache/iceberg/pull/10449

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java
```diff
@@ -177,12 +177,9 @@ CombinedScanTask> applyConfig(
     }
     String schemaStr = conf.get(InputFormatConfig.READ_SCHEMA);
     if (schemaStr != null) {
-      scan.project(SchemaParser.fromJson(schemaStr));
-    }
-
-    String[] selectedColumns = conf.getStrings(InputFormatConfig.SELECTED_COLUMNS);
-    if (selectedColumns != null) {
-      scan.select(selectedColumns);
+      scan = scan.project(SchemaParser.fromJson(schemaStr));
```
deniskuzZ, line 180: why don't we use helper methods:
````
Schema readSchema = InputFormatConfig.readSchema(conf);
if (readSchema != null) {
  scan = scan.project(readSchema)
} else {
  String[] selectedColumns = InputFormatConfig.selectedColumns(conf);
  if (selectedColumns != null) {
    scan = scan.select(selectedColumns)
  }
}
````

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java:180]: btw, there is `expectedSchema = readSchema(conf, table, caseSensitive)`
but it creates a Hive schema with virtual columns, maybe that could be reused

**Comment by deniskuzZ** [comment]
@zhangbutao, does this mean we always read the complete schema from iceberg and then applied project on Hive side?

**Comment by zhangbutao** [comment]
> @zhangbutao, does this mean we always read the complete schema from iceberg and then applied project on Hive side? that should reduce number of bytes read, right?

@deniskuzZ Not completely right. I think the original intention `project()` & `select()`  stored on `scan `is to reduce the number of bytes read.
But **on Hive side**, i found that we didn't take full advantage of the `column projection` propertity stored on `scan`. After some debugging, i think the main use `project()` & `select()` on scan is here,: just to do some scan metrics **but we don't use on Hive**:
https://github.com/apache/iceberg/blob/c7de6cb345995cb47312edbef6edae2f17fb8aba/core/src/main/java/org/apache/iceberg/SnapshotScan.java#L143-L157

In fact, on Hive side, we propagate the `projected column` through `conf` not the `scan` class, and we get this `projected column` through conf where we need it, some code snippet like this:
https://github.com/apache/hive/blob/98d9d22398370f817fe64449368671b978fff096/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java#L342

the final place in hive which use `projected column` is here(Parquet iceberg table):
https://github.com/apache/hive/blob/98d9d22398370f817fe64449368671b978fff096/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java#L510


In one word, with or without this my change, we both can optimize the `project column`, as on Hive we don't rely on the  `project()` & `select()` stored on `scan`, and we just use conf to propagate the `projected column`.
But we need make the code correct & return the new scan, so that we can do some stuff in the future by using the `project()` & `select()` stored on `scan`.
`

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java (hunk 2 above)
Comment (zhangbutao) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java]: > btw, there is `expectedSchema = readSchema(conf, table, caseSensitive)` but it creates a Hive schema with virtual columns, maybe that could be reused

I didn't find a  better way to reuse this method. :(

**Review by deniskuzZ — APPROVED** [review]
+1, thanks @zhangbutao for the fix!

**Comment by deniskuzZ** [comment]
> > @zhangbutao, does this mean we always read the complete schema from iceberg and then applied project on Hive side? that should reduce number of bytes read, right?
> 
> @deniskuzZ Not completely right. I think the original intention `project()` & `select()` stored on `scan `is to reduce the number of bytes read. But **on Hive side**, i found that we didn't take full advantage of the `column projection` propertity stored on `scan`. After some debugging, i think the main use `project()` & `select()` on scan is here,: just to do some scan metrics **but we don't use on Hive**: https://github.com/apache/iceberg/blob/c7de6cb345995cb47312edbef6edae2f17fb8aba/core/src/main/java/org/apache/iceberg/SnapshotScan.java#L143-L157
> 
> In fact, on Hive side, we propagate the `projected column` through `conf` not the `scan` class, and we get this `projected column` through conf where we need it, some code snippet like this:
> 
> https://github.com/apache/hive/blob/98d9d22398370f817fe64449368671b978fff096/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java#L342
> 
> the final place in hive which use `projected column` is here(Parquet iceberg table):
> 
> https://github.com/apache/hive/blob/98d9d22398370f817fe64449368671b978fff096/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java#L510
> 
> In one word, with or without this my change, we both can optimize the `project column`, as on Hive we don't rely on the `project()` & `select()` stored on `scan`, and we just use conf to propagate the `projected column`. But we need make the code correct & return the new scan, so that we can do some stuff in the future by using the `project()` & `select()` stored on `scan`. `

since it's not used now, could we add TODO otherwise that would be just a dead code

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergInputFormat.java [hunk 3]
```diff
@@ -175,14 +174,19 @@ CombinedScanTask> applyConfig(
...
     }
-    String schemaStr = conf.get(InputFormatConfig.READ_SCHEMA);
-    if (schemaStr != null) {
-      scan.project(SchemaParser.fromJson(schemaStr));
-    }
-
-    String[] selectedColumns = conf.getStrings(InputFormatConfig.SELECTED_COLUMNS);
-    if (selectedColumns != null) {
-      scan.select(selectedColumns);
+    //  TODO: Currently, this projection optimization stored on scan is not being used effectively on Hive side, as
```
zhangbutao, line 177: @deniskuzZ Added TODO here, Please help to check if it is ok.
Thanks.

PR #5283, opened 2024-06-06
HIVE-28302: Let SUM UDF return NULL when all rows have non-numeric texts

okumin:

Make SUM return not 0.0 but NULL when both of the following conditions are satisfied.
- The input type is a STRING like type
- All values are non-numeric and it is impossible to successfully cast to DOUBLE

https://issues.apache.org/jira/browse/HIVE-28302

I understand the SQL standard requires a UDAF to return NULL when all rows are NULL. Actually, other UDAFs such as AVG behave like that.

We can see some more discussions in https://github.com/apache/hive/pull/5091

Yes. But I believe the current behavior is not an intentional one but a kind of bug.

No.

Updated integration tests.


ql/src/test/results/clientpositive/llap/udaf_number_format.q.out
```diff
@@ -92,4 +92,4 @@ FROM src
 POSTHOOK: type: QUERY
 POSTHOOK: Input: default@src
 #### A masked pattern was here ####
-0.0	NULL	NULL	NULL
+NULL	NULL	NULL	NULL
```
okumin, line 95: We may remove this test case as `cbo_aggregate_reduce_functions_rule.q` covers the case.
https://github.com/apache/hive/blob/master/ql/src/test/queries/clientpositive/udaf_number_format.q

kasakrisz:
In the test [cbo_aggregate_reduce_functions_rule.q](https://github.com/apache/hive/blob/b15ce9126dcc596c953fe202e6834cebdfca6080/ql/src/test/queries/clientpositive/cbo_aggregate_reduce_functions_rule.q#L57-L62) aggregate functions under test (sum, count, stddev*, etc) are wrapped into a `round` function call. I understand the reason however I think it can cause some noise: if an expression like
```
ROUND(SUM(c_numeric), 3)
```
returns invalid result it can have multiple reasons: `sum` is wrong or `round` is wrong or both. 
In theory if both are wrong as a side effect the expression itself can return a good result in some cases hence a potential bug in both function implementations remains hidden.

WDYT?

okumin:
@kasakrisz I don't have any objections. The original intention is human-readability and floating-point error mitigation. [I tried to remove ROUND](https://github.com/apache/hive/pull/5283/commits/8b8136e4c856c276778874710d8eccb06a278152).

To be fair, we observe difference between CBO and non-CBO.

```
< +228.0  228.0   0.0     0.0     193.0   193.0   28.5    28.5    NULL    NULL    32.166666666666664      32.166666666666664      47.34448225506326       47.34448225506326       NULL    NULL    53.52387836802893       53.52387836802893       50.61338050075578  50.61338050075578       NULL    NULL    58.63247109466449       58.63247109466449       2241.5  2241.5  NULL    NULL    2864.805555555555       2864.805555555555       2561.714285714286       2561.714285714286       NULL    NULL    3437.7666666666664 3437.7666666666664      228.0   228.0   NULL    NULL    193.0   193.0   8       8       8       0       8       6
---
> +228.0  228.0   0.0     0.0     193.0   193.0   28.5    28.5    NULL    NULL    32.166666666666664      32.166666666666664      47.34448225506326       47.34448225506326       NULL    NULL    53.52387836802894       53.52387836802894       50.61338050075578  50.61338050075578       NULL    NULL    58.632471094664496      58.632471094664496      2241.5  2241.5  NULL    NULL    2864.805555555556       2864.805555555556       2561.714285714286       2561.714285714286       NULL    NULL    3437.7666666666673 3437.7666666666673      228.0   228.0   NULL    NULL    193.0   193.0   8       8       8       0       8       6
```

For better visibility,

```
< 53.52387836802893
< 53.52387836802893
---
> 53.52387836802894
> 53.52387836802894
71,72c71,72
< 58.63247109466449
< 58.63247109466449
---
> 58.632471094664496
> 58.632471094664496
77,78c77,78
< 2864.805555555555
< 2864.805555555555
---
> 2864.805555555556
> 2864.805555555556
83,84c83,84
< 3437.7666666666664
< 3437.7666666666664
---
> 3437.7666666666673
> 3437.7666666666673
```

I presume they are acceptable as floating-point errors.

kasakrisz (approved):

okumin:
Thanks!

PR #4996, opened 2024-01-10
HIVE-21520: Query 'Submit plan' time reported is incorrect

zhangbutao:
### Why are the changes needed?

Please check HIVE-21520. This PR is based on @rbalamohan 's patch and this try to fix incorrect display time(Compile Query & Prepare Plan & Submit Plan) in Beeline console.
**Before this PR:**
INFO  : Query Execution Summary
INFO  : ----------------------------------------------------------------------------------------------
INFO  : OPERATION                            DURATION
INFO  : ----------------------------------------------------------------------------------------------
INFO  : Compile Query                           0.00s
INFO  : Prepare Plan                            0.00s
INFO  : Get Query Coordinator (AM)              0.00s
INFO  : Submit Plan                         1704899467.78s
INFO  : Start DAG                               0.05s
INFO  : Run DAG                                 7.91s
INFO  : --`

**After this PR:**
INFO  : Query Execution Summary
INFO  : ----------------------------------------------------------------------------------------------
INFO  : OPERATION                            DURATION
INFO  : ----------------------------------------------------------------------------------------------
INFO  : Compile Query                           6.63s
INFO  : Prepare Plan                            0.77s
INFO  : Get Query Coordinator (AM)              0.02s
INFO  : Submit Plan                             0.54s
INFO  : Start DAG                               0.14s
INFO  : Run DAG                                 8.87s
INFO  : --

### Does this PR introduce _any_ user-facing change?

No
### Is the change a dependency upgrade?

No

### How was this patch tested?

Manual test in local hive cluster.

PR #5285, opened 2024-06-07
HIVE-28309: fixed Task Execution Summary duration is not displayed

ryukobayashi:

Fixed an issue where the `Task Execution Summary` duration is not displayed.

`Task Execution Summary` duration is not displayed and 0 is always displayed.

No

Confirmed that `Task Execution Summary` duration is displayed after run.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java
@@ -124,17 +124,17 @@ public class TezJobMonitor {
     this.dag = dag;
     this.context = ctx;
     console = SessionState.getConsole();
+    this.perfLogger = perfLogger;
     updateFunction = updateFunction();
     this.counters = counters;
-    this.perfLogger = perfLogger;
   }
 
   private RenderStrategy.UpdateFunction updateFunction() {
     return InPlaceUpdate.canRenderInPlace(hiveConf)
         && !SessionState.getConsole().getIsSilent()
         && !SessionState.get().isHiveServerQuery()
-        ? new RenderStrategy.InPlaceUpdateFunction(this)
-        : new RenderStrategy.LogToFileFunction(this);
+        ? new RenderStrategy.InPlaceUpdateFunction(this, perfLogger)
+        : new RenderStrategy.LogToFileFunction(this, perfLogger);
   }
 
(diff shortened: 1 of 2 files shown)


zhangbutao:
Oh, This issue shoud be caused by [HIVE-21520](https://issues.apache.org/jira/browse/HIVE-21520) https://github.com/apache/hive/pull/4996. :(

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java
```diff
@@ -124,17 +124,17 @@ public TezJobMonitor(List<BaseWork> topSortedWorks, final DAGClient dagClient, H
     this.dag = dag;
     this.context = ctx;
     console = SessionState.getConsole();
+    this.perfLogger = perfLogger;
```
zhangbutao, line 127: Why move this line of code?

ryukobayashi: Since it is using here, it needs to be set before the `updateFunction` is called.
https://github.com/apache/hive/blob/82099b8fbf7e7ec3f3aaa778ddd70b8f959c1ad2/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java#L136-L137

https://github.com/apache/hive/blob/82099b8fbf7e7ec3f3aaa778ddd70b8f959c1ad2/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java#L128

zhangbutao (approved):
Tested the patch in local cluster, and it worked as expected. 

Based on my current understanding, +1
It would be good if @abstractdog can take a look. Thanks.

abstractdog:
looks good to me, assuming that it's been tested by @zhangbutao 
can someone please describe here and/or on the jira what the root cause was? that would be crucial for later reference

ryukobayashi:
@zhangbutao @abstractdog 
I added the root cause to JIRA ticket comment: https://issues.apache.org/jira/browse/HIVE-28309?focusedCommentId=17853884&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17853884

abstractdog (approved):
+1

ryukobayashi:
Thanks both.

PR #5286, opened 2024-06-07
HIVE-28310: Disable hive.optimize.join.disjunctive.transitive.predicates.pushdown by default

zabetak:
Pushing dis-junction and can lead to OOM and HS2 crashes. Turning off this feature would make HS2 more stable at the expense of some less efficient plans.

Plan and potentially performance changes.

No

Existing tests


asolimando (approved):
I agree that, given the number of escalations and issues we have seen over time due to disjunctive filters, it's safer to disable it by default.

Queries experiencing performance regressions can always turn this property on.

Approval modulo pending tests.

PR #5287, opened 2024-06-07
HIVE-28311: Backward compatibility of java.sql.Date and java.sql.Timestamp in hive-serde

wecharyu:
Support `java.sql.Date` and `java.sql.Timestamp` objects in JavaXXXObjectInspector.

The old clients before HIVE-20007 convert date value to `java.sql.Date` and timestamp value to `java.sql.Timestamp`, but the new HMS only accept internal `Date` and `Timestamp` while deserialize the expressions, which would cause `ClassCastException`.

No.

No.

Add new unit test:
```bash
mvn test -Dtest.groups= -Dtest=org.apache.hadoop.hive.serde2.objectinspector.primitive.TestPrimitiveObjectInspectorUtils -pl :hive-serde
```

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/common/type/Date.java
+++ b/common/src/java/org/apache/hadoop/hive/common/type/Date.java
@@ -190,6 +190,10 @@ public class Date implements Comparable<Date> {
     }
   }
 
+  public static Date valueOf(java.sql.Date d) {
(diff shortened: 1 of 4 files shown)


wecharyu:
@zabetak @dengzhhu653 @zhangbutao could you pls review this PR?

PR #5289, opened 2024-06-07
HIVE-28312: Save query error message on Driver level

abstractdog:
Store the query error message in a variable in DriverContext.

Because a simple failure reason message should be available for later analysis or query-tracking purposes.

No.

No.

New unit test class with related methods is included: TestDriver.


simhadri-g (approved):
LGTM +1

PR #5290, opened 2024-06-07
HIVE-28313: Support ambiguous column references when CBO fallback is …

soumyakanti3578:
…disabled

Fix test failures for ambiguous_col.q when CBO is enabled.

Refer to https://issues.apache.org/jira/browse/HIVE-27831

No

No

`mvn test -pl itests/qtest -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile="ambiguous_col.q"`
`mvn test -pl itests/qtest -Pitests -Dtest=TestNegativeLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile="ambiguous_col.q"`


ramesh0201:
LGTM, Did you already check if there are more q files with CBO disabled that can cause the ambiguous column errors? If the check is done then please free to merge this change

soumyakanti3578:
Thanks for the review @ramesh0201. Yes, this was the only qtest which was causing this error.

PR #5292, opened 2024-06-09
HIVE-28307: Fast return when database exists and IF NOT EXISTS clause…

wecharyu:
… is enabled

Check if database exists when `IF NOT EXISTS` clause is used.

1. It could avoid the HMS calls if unnecessary, and will not print confused HMS error log.
2. Align the behavior as `create table if not exists`.

No.

No.

Test locally.


ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseOperation.java
```diff
@@ -44,6 +44,9 @@ public CreateDatabaseOperation(DDLOperationContext context, CreateDatabaseDesc d
 
   @Override
   public int execute() throws HiveException {
+    if (desc.getIfNotExists() && context.getDb().databaseExists(desc.getName())) {
```
zhangbutao, line 47: Should we also follow `table if not exist` check pattern, just do a check in https://github.com/apache/hive/blob/3eb56ecae056d36370ff0ee9a0936c28b431724a/ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseAnalyzer.java#L47-L50

wecharyu: Addressed.

zhangbutao (approved):
LGTM +1 Pending a green CI.

PR #5293, opened 2024-06-12
HIVE-28323: Iceberg: Allow reading tables irrespective whether they were created with hive engined enabled or not.

ayushtkn:

Better interpretability with other engines

To make iceberg table readable from Hive, created by engines with hive engine disabled (which is by default in iceberg)

Iceberg tables would be readable from Hive irrespective if the other engine sets the correct the SerDe or StorageHandler

No

UT

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -1072,9 +1072,9 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
           hmsTable.getParameters().put(TableProperties.FORMAT_VERSION, formatVersion);
         }
         // Set the serde info
-        hmsTable.getSd().setInputFormat("org.apache.iceberg.mr.hive.HiveIcebergInputFormat");
-        hmsTable.getSd().setOutputFormat("org.apache.iceberg.mr.hive.HiveIcebergOutputFormat");
-        hmsTable.getSd().getSerdeInfo().setSerializationLib("org.apache.iceberg.mr.hive.HiveIcebergSerDe");
+        hmsTable.getSd().setInputFormat(HiveIcebergInputFormat.class.getName());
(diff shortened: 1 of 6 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -1071,6 +1071,16 @@ public void postGetTable(org.apache.hadoop.hive.metastore.api.Table hmsTable) {
         if (!"1".equals(formatVersion)) {
           hmsTable.getParameters().put(TableProperties.FORMAT_VERSION, formatVersion);
         }
+        // Set the serde info
+        hmsTable.getSd().setInputFormat("org.apache.iceberg.mr.hive.HiveIcebergInputFormat");
```
simhadri-g, line 1075: nit:  Can we replace the strings with `getClass().getName()` ?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -5860,6 +5860,16 @@ private HiveStorageHandler createStorageHandler(org.apache.hadoop.hive.metastore
     }
   }
 
+  private static String getStorageHandlerClassName(org.apache.hadoop.hive.metastore.api.Table tbl) {
+    String tableType = tbl.getParameters().get(HiveMetaHook.TABLE_TYPE);
+    String metaTableStorage = tbl.getParameters().get(META_TABLE_STORAGE);
+    if (HiveMetaHook.ICEBERG.equalsIgnoreCase(tableType) && metaTableStorage == null) {
+      return "org.apache.iceberg.mr.hive.HiveIcebergStorageHandler";
```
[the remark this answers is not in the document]
ayushtkn: changed to use constant, as discussed

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
ayushtkn: Done

simhadri-g (approved):
LGTM +1

PR #5294, opened 2024-06-12
HIVE-28321 Support select alias in the having clause for CBO

ramesh0201:
Having clause CBO path of the queries should have visibility from the alias of the select expressions, as this is the case in the non-CBO path.

In order to make the CBO path and non-CBO work the same way in terms of syntax and semantics of having clause

No

No

Unit tests. Specifically mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=limit_pushdown_negative.q

### DIFF
--- a/itests/hive-blobstore/src/test/queries/clientpositive/having.q
+++ b/itests/hive-blobstore/src/test/queries/clientpositive/having.q
@@ -1,4 +1,5 @@
 -- Test HAVING clause
+set hive.cbo.fallback.strategy=NEVER;
 
 DROP TABLE having_blobstore_test;
 CREATE TABLE having_blobstore_test(a int, b int, c string)
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
@@ -5161,7 +5161,10 @@ public class CalcitePlanner extends SemanticAnalyzer {
               UnsupportedFeature.Having_clause_without_any_groupby);
         }
         ASTNode targetNode = (ASTNode) havingClause.getChild(0);
-        validateNoHavingReferenceToAlias(qb, targetNode);
+        QBParseInfo qbPI = qb.getParseInfo();
+        Map<ASTNode, String> exprToAlias = qbPI.getAllExprToColumnAlias();
+        RowResolver inputRR = relToHiveRR.get(srcRel);
+        inputRR.putAll(exprToAlias);
         if (!qbp.getDestToGroupBy().isEmpty()) {
           final boolean cubeRollupGrpSetPresent = (!qbp.getDestRollups().isEmpty()
                   || !qbp.getDestGroupingSets().isEmpty() || !qbp.getDestCubes().isEmpty());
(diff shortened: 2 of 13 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -5167,7 +5167,14 @@ private RelNode genGBHavingLogicalPlan(QB qb, RelNode srcRel) throws SemanticExc
               UnsupportedFeature.Having_clause_without_any_groupby);
         }
         ASTNode targetNode = (ASTNode) havingClause.getChild(0);
-        validateNoHavingReferenceToAlias(qb, targetNode);
+        QBParseInfo qbPI = qb.getParseInfo();
+        Map<ASTNode, String> exprToAlias = qbPI.getAllExprToColumnAlias();
+        RowResolver inputRR = relToHiveRR.get(srcRel);
+        for (ASTNode astNode : exprToAlias.keySet()) {
+          if (inputRR.getExpression(astNode) != null) {
+            inputRR.put("", exprToAlias.get(astNode), inputRR.getExpression(astNode));
```
soumyakanti3578, line 5175: I see that `tableAlias` is an empty string here, and even in non-cbo path it is an empty string, but not sure if there is a better way to deal with this. I see several instances in the code base where we use empty string for this method.

itests/hive-blobstore/src/test/queries/clientpositive/having.q
```diff
@@ -1,4 +1,5 @@
 -- Test HAVING clause
```
soumyakanti3578, line 1: Since we are supporting this through CBO, I think it'd be nice to include a CBO plan for one of the queries below.
Also, it would be good to add a more complex test with multiple column aliases in the having clause.

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -5167,7 +5167,14 @@ private RelNode genGBHavingLogicalPlan(QB qb, RelNode srcRel) throws SemanticExc
...
+          }
+        }
```
zabetak, line 5177: This part is basically modifying the `RowResolver` adding some kind of reverse mappings. Moreover, this logic is duplicated in `SemanticAnalyzer#genHavingPlan`. I think it makes sense to refactor this snippet and create a new method in the `RowResolver` class with a proper name (and javadoc if necessary) that clarifies what is the code supposed to do. The only parameter to the method should be a `Map<ASTNode, String>`.

itests/hive-blobstore/src/test/queries/clientpositive/having.q
zabetak: Adding more tests and EXPLAIN CBO plans is a good idea.

zabetak (commented):
LGTM, let's add the extra tests and small suggested improvements and then we can get this in.

ql/src/java/org/apache/hadoop/hive/ql/parse/RowResolver.java
```diff
@@ -125,6 +125,22 @@ private boolean putInternal(String tabAlias, String colAlias, ColumnInfo colInfo
...
 
+  /**
+   * Puts a resolver entry corresponding to a aggregate expression and its
+   * column alias which is used for alias recognition in the having clause.
+   * @param exprToColumnAlias
+   *          The map containing the expression to alias mapping
+   * @return
+   * @throws SemanticException
+   */
+  public void putAggregateAlias(Map<ASTNode, String> exprToColumnAlias) throws SemanticException {
```
zabetak, line 136: The name of the method is unnecessarily tight to aggregates and aliases while the implementation seems much more generic than that. Given that we just call `put` for every entry in the map maybe a better name would be `putAll` or maybe `replaceAll` since we only perform the action if an entry is already there.

Will this logic really add an entry in the resolver or it just changes the alias for ColumnInfo that is registered previously? Depending on the answer another potential name would be `replaceAliases`.

A well chosen name will better indicate what the method really does without requiring checking the implementation.

ql/src/test/queries/clientpositive/having.q
```diff
@@ -18,4 +19,7 @@ EXPLAIN SELECT key, max(value) FROM src GROUP BY key HAVING max(value) > "val_25
 SELECT key, max(value) FROM src GROUP BY key HAVING max(value) > "val_255";
 
 EXPLAIN SELECT key, COUNT(value) FROM src GROUP BY key HAVING count(value) >= 4;
-SELECT key, COUNT(value) FROM src GROUP BY key HAVING count(value) >= 4;
\ No newline at end of file
+SELECT key, COUNT(value) FROM src GROUP BY key HAVING count(value) >= 4;
+
+EXPLAIN CBO SELECT count(value) as c, max(key) as m from src GROUP BY key HAVING c > 3 and m > 0;
+SELECT count(value) as c, max(key) as m from src GROUP BY key HAVING c > 3 and m > 0;
```
zabetak, line 25: Since we are talking about `max(key)` does the `m > 0` condition filter anything? If not then maybe changing to `m > 400` would be a safer choice.

zabetak (approved):

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
ramesh0201: I agree with you. I think the better way to handle is to pass the tableAlias field instead of the "" and then let the field fallback to the empty string if it is empty. Given that it is a much broader scope than this PR, I would prefer to address this issue separately. Let me know what you think?

soumyakanti3578: That is fine with me :)

ramesh0201:
Thank you @soumyakanti3578 and @zabetak for the review

PR #5298, opened 2024-06-13
HIVE-28325: Slow compilation without owner information with Ranger au…

scarlin-cloudera:
…thorization

Added owner information to the HivePrivilegeObject created in the SemanticAnalyzer. The owner information is used by Ranger for filtering of objects. If the owner is not provided, it can slow down compilation.

This is related to HIVE-27285 and RANGER-3593.

Described in commit message

Compilation becomes noticeably quicker when Ranger is provided owner information for the table 

No

No

Unit test changed and now shows owner information in the hook.


jfsii (approved):
It looks good to me. Thanks for making this code path a bit better, Ranger calling HMS with each check does kill perf.

saihemanth-cloudera (approved):
LGTM +1, Pending green Jenkins tests.

PR #5299, opened 2024-06-14
HIVE-28326: Fix StageIDsRearranger

ngsg:

StageIDsRearranger now returns PhysicalContext instead of null.

The return value of a PhysicalResolver is passed to another PhysicalResolver. So returning null makes problem in the following PhysicalResolver.

No

No

Using manyViewJoin.q. Also I changed the test driver to TestMiniLlapLocalCliDriver, which becomes the default test driver since HIVE-23510.

### DIFF
--- a/itests/src/test/resources/testconfiguration.properties
+++ b/itests/src/test/resources/testconfiguration.properties
@@ -241,7 +241,6 @@ mr.query.files=\
   join_vc.q,\
   list_bucket_dml_8.q,\
   localtimezone.q,\
-  manyViewJoin.q,\
   mapjoin1.q,\
   mapjoin47.q,\
   mapjoin_distinct.q,\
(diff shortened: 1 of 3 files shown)


itests/src/test/resources/testconfiguration.properties
```diff
@@ -241,7 +241,6 @@ mr.query.files=\
   join_vc.q,\
   list_bucket_dml_8.q,\
   localtimezone.q,\
-  manyViewJoin.q,\
```
okumin, line 244: Why can we test HIVE-28326 using `manyViewJoin.q`? I applied the change of `StageIDsRearranger.java`, built it, and run `mvn test -Dtest=TestCliDriver -Dtest.output.overwrite=false -Dqfile=manyViewJoin.q`. The test succeeded on my machine.

okumin: I understand the test case with the original revision fails with NPE.

```
See ./ql/target/tmp/log/hive.log or ./itests/qtest/target/tmp/log/hive.log, or check ./ql/target/surefire-reports or ./itests/qtest/target/surefire-reports/ for specific test cases logs.
 java.lang.NullPointerException
	at org.apache.hadoop.hive.ql.optimizer.physical.LlapDecider.resolve(LlapDecider.java:530)
	at org.apache.hadoop.hive.ql.parse.TezCompiler.optimizeTaskPlan(TezCompiler.java:834)
	at org.apache.hadoop.hive.ql.parse.TaskCompiler.compile(TaskCompiler.java:340)
```

okumin (approved):
I have confirmed all PhysicalPlanResolvers used in TezCompiler returns a context now

deniskuzZ (approved):
LGTM +1

PR #5308, opened 2024-06-18
HIVE-28334 Support queries with function expression in the prepare execute workflow

ramesh0201:
Queries with function expression throw error as their ObjectInspector is not initialized. As part of prepare execute execute workflow, we serialize and deserialize plan during execute analyzer. During this step, ExprNodeGenericFuncDesc's ObjectInspector field is not initialized as it is a transient field and is never part of the serialized byte array. Assuming it can be null, we try to populate the field when required.

To make any queries with function expression work as part of the prepare execute workflow.

No

No

mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=prepare_plan_func_expr.q

### DIFF
--- a/ql/src/test/queries/clientpositive/prepare_plan_func_expr.q
+++ b/ql/src/test/queries/clientpositive/prepare_plan_func_expr.q
@@ -1,0 +1,9 @@
+--! qt:dataset:src
+--! qt:dataset:alltypesorc
+
+explain prepare pint from select sum(ctinyint)/count(ctinyint) as ag from alltypesorc where cint = ? group by cint;
+explain cbo prepare pint from select sum(ctinyint)/count(ctinyint) as ag from alltypesorc where cint = ? group by cint;
+prepare pint from select sum(ctinyint)/count(ctinyint) as ag from alltypesorc where cint = ? group by cint;
+
+execute pint using 1000828;
+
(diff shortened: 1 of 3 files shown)


ql/src/test/queries/clientpositive/prepare_plan_func_expr.q
```diff
@@ -0,0 +1,12 @@
...
+--SORT_QUERY_RESULTS
+
+explain prepare pint from select sum(ctinyint)/count(ctinyint) as ag from alltypesorc where cint <= ?  and cbigint <= ? and cfloat != ? group by cint;
+
+prepare pint from select sum(ctinyint)/count(ctinyint) as ag from alltypesorc where cint <= ?  and cbigint <= ? and cfloat != ? group by cint;
+
+explain execute pint using 100, 5000000, 0.023;
+
+execute pint using 100, 5000000, 0.023;
+
```
soumyakanti3578, line 12: Small nits:
1. Maybe add `explain cbo` plans too.
2. Since we want to just test if execute statements run, we don't need a lot of output rows anyway. So, maybe reduce the number of output rows by only using equality or IN operators. Also include those columns (cint, cbigint, cfloat) in the output.
3. If you want to test !=, <=, >= operators, add more tests with counts.

These are not required btw, and the change looks good to me.

ramesh0201: Thanks for the review @soumyakanti3578 , Can you please look at the changes in the test file and verify?

soumyakanti3578: Thanks, LGTM!

PR #5004, opened 2024-01-15
HIVE-27999: Upgrade to Java 17 to run sonar analysis

wecharyu:
### What changes were proposed in this pull request?
Upgrade to java 17 in CI to run sonar analysis.

### Why are the changes needed?
To make CI pass.

### Does this PR introduce _any_ user-facing change?
No.

### Is the change a dependency upgrade?
No.

### How was this patch tested?
Pass CI.

PR #5309, opened 2024-06-18
HIVE-28337: Process timestamps at UTC timezone instead of local timezone in MetaStoreUtils

KiranVelumuri:

Timestamps are processed using LocalDateTime at local time zone. Change this to use Instant instead to process timestamps at UTC time zone.

https://issues.apache.org/jira/browse/HIVE-28337

No

No

mvn test -Dtest=TestMetaStoreUtils

Before the patch:
with Random(seed: 23) in line 65(default case)
image

with Random(seed: 24) in line 65
image

After the patch:
image

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java
@@ -25,8 +25,8 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.sql.Date;
 import java.sql.Timestamp;
+import java.time.Instant;
 import java.time.LocalDate;
-import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.ResolverStyle;
 import java.util.ArrayList;
@@ -124,7 +124,7 @@ public class MetaStoreUtils {
    * @return Timestamp in string format.
    */
   public static String convertTimestampToString(Timestamp timestamp) {
-    return TIMESTAMP_FORMATTER.format(timestamp.toLocalDateTime());
+    return TIMESTAMP_FORMATTER.format(timestamp.toInstant());
   }
 
   /**
@@ -133,8 +133,8 @@ public class MetaStoreUtils {
    * @return java.sql.Timestamp object.
    */
   public static Timestamp convertStringToTimestamp(String timestamp) {
-    LocalDateTime val = LocalDateTime.from(TIMESTAMP_FORMATTER.parse(timestamp));
-    return Timestamp.valueOf(val);
+    Instant instant = Instant.from(TIMESTAMP_FORMATTER.parse(timestamp));
+    return Timestamp.from(instant);
   }
 
   // Indicates a type was derived from the deserializer rather than Hive's metadata.
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreUtils.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreUtils.java
@@ -47,16 +47,18 @@ public class TestMetaStoreUtils {
   private static final TimeZone DEFAULT = TimeZone.getDefault();
   private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss");
   private final TimeZone timezone;
+  private final Timestamp timestamp;
   private final String date;
-  private final String timestamp;
+  private final String timestampString;
 
-  public TestMetaStoreUtils(String zoneId, LocalDateTime timestamp) {
+  public TestMetaStoreUtils(String zoneId, LocalDateTime localDateTime) {
     this.timezone = TimeZone.getTimeZone(zoneId);
-    this.timestamp = timestamp.format(FORMATTER);
-    this.date = timestamp.toLocalDate().format(DateTimeFormatter.ISO_LOCAL_DATE);
+    this.timestamp = Timestamp.from(localDateTime.toInstant(ZoneOffset.UTC));
+    this.timestampString = localDateTime.format(FORMATTER);
+    this.date = timestamp.toLocalDateTime().format(DateTimeFormatter.ISO_LOCAL_DATE);
   }
 
-  @Parameterized.Parameters(name = "zoneId={0}, timestamp={1}")
+  @Parameterized.Parameters(name = "zoneId={0}, localDateTime={1}")
   public static Collection<Object[]> generateZoneTimestampPairs() {
     List<Object[]> params = new ArrayList<>();
     long minDate = LocalDate.of(0, 1, 1).atStartOfDay().toEpochSecond(ZoneOffset.UTC);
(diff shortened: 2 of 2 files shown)


standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreUtils.java
```diff
@@ -50,7 +51,7 @@ public class TestMetaStoreUtils {
   private final String date;
   private final String timestamp;
 
-  public TestMetaStoreUtils(String zoneId, LocalDateTime timestamp) {
+  public TestMetaStoreUtils(String zoneId, ZonedDateTime timestamp) {
     this.timezone = TimeZone.getTimeZone(zoneId);
     this.timestamp = timestamp.format(FORMATTER);
```
wecharyu, line 56: How about keeping `LocalDateTime` in constructor parameter, and generate the timeZone related timestamp string here:
```java
this.timestamp = ZonedDateTime.of(timestamp, ZoneId.of(zoneId)).format(FORMATTER);
```

wecharyu:
Nice catch~ This failed test was also found in https://github.com/apache/hive/pull/5004 and we can revert https://github.com/kgyrtkirk/hive-dev-box/pull/16 once this patch is merged. 
@zabetak could you also take a look?

zabetak (changes requested):
The intention of the test is to ensure that conversions from/to string do not alter the value no matter the timezone because the DATE/TIMESTAMP data types [should be timezone agnostic|https://cwiki.apache.org/confluence/display/Hive/Different+TIMESTAMP+types].

Changing LocalDateTime to ZonedDateTime defeats the purpose of the tests and actually hides a bug that affects the MetaStoreUtils APIs.

I left a similar comment under the JIRA ticket.

KiranVelumuri:
The 2 test cases mentioned in https://issues.apache.org/jira/browse/HIVE-28337 are failing for me locally in my machine with Java version 1.8.0_331. I tried running these test cases in https://github.com/apache/hive/pull/5004, but the DST related one was successful. I was investigating as to why this was happening, and now I get it seeing the previous comments that this is occurring due to JDK bug.

KiranVelumuri:
> The intention of the test is to ensure that conversions from/to string do not alter the value no matter the timezone because the DATE/TIMESTAMP data types [should be timezone agnostic](https://cwiki.apache.org/confluence/display/Hive/Different+TIMESTAMP+types).
> 
> Changing LocalDateTime to ZonedDateTime defeats the purpose of the tests and actually hides a bug that affects the MetaStoreUtils APIs.
> 
> I left a similar comment under the JIRA ticket.

Yes, I agree with you regarding the intention and purpose of the test. I am looking into the issue and would re-work this PR taking the MetaStoreUtils APIs into account.

KiranVelumuri:
I have updated the tests and the MetaStoreUtils APIs to always have Timestamp created in UTC timezone, since the timestamps during the daylight savings transition could not be represented in the local time zone.


However, there are 2 known issues for the timestamps mentioned below but it is the expected behaviour from Java's Timestamp:
1. Timestamps with year 0000 would be converted to year 0001.
2. Timestamps in the range [1582-10-05 00:00:00, 1582-10-14 23:59:59] (both inclusive) would have 10 days added due to the shift from Julian calendar to Gregorian calendar. (The next day after 1582-10-04 is 1582-10-15)

References:
[1] https://www.ibm.com/docs/en/db2/11.5?topic=dttmddtija-date-time-timestamp-values-that-can-cause-problems-in-jdbc-sqlj-applications#imjcc_r0053436__title__5
[2] https://strasbourg.eleusal.com/en/what-happened-between-5-and-14-october-1582/

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreUtils.java
KiranVelumuri: From @zabetak's comment, I have reverted to LocalDateTime.

zabetak (changes requested):
I understand that due to bugs the tests may fail in certain environments.

If the main motivation of this PR is to avoid the failures when this happens it seems more appropriate to add conditions for detecting skipping the specific test with appropriate bug reference. 

If the intention is to fix the problems revealed by the tests then more work is needed since the proposed changes cannot land as they are.

KiranVelumuri:
g

KiranVelumuri:
I have updated the test to not hit the failing cases. Will raise a separate Jira for the rework that is needed around timestamp. @wecharyu @zabetak Please review. Thank you.

KiranVelumuri:
> I understand that due to bugs the tests may fail in certain environments.
> 
> If the main motivation of this PR is to avoid the failures when this happens it seems more appropriate to add conditions for detecting skipping the specific test with appropriate bug reference.
> 
> If the intention is to fix the problems revealed by the tests then more work is needed since the proposed changes cannot land as they are.

The initial motivation was to fix the test failures, which later changed to solve the problem revealed in MetaStoreUtils. Upon your comments, I have reworked my PR which now solves the problem in MetaStoreUtils and accordingly modified TestMetaStoreUtils to validate the tests properly.

@zabetak @wecharyu Please review. Thank you.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreUtils.java
```diff
@@ -134,7 +135,7 @@ public static String convertTimestampToString(Timestamp timestamp) {
    */
   public static Timestamp convertStringToTimestamp(String timestamp) {
     LocalDateTime val = LocalDateTime.from(TIMESTAMP_FORMATTER.parse(timestamp));
```
zabetak, line 137: Since now we are converting to `Instant` do we need to pass from `LocalDateTime`? Can we use directly `Instant.from` here?

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreUtils.java
```diff
@@ -67,9 +69,31 @@ public static Collection<Object[]> generateZoneTimestampPairs() {
...
     return params;
   }
 
+  public static void generateDaylightSavingTimestampPairs(List<Object[]> params) {
+    // For U.S timezones, the timestamp 2024-03-10 02:01:00 falls between the transition from 02:00 AM to 03:00 AM for daylight savings
+    params.add(new Object[] {"America/Anchorage", LocalDateTime.ofEpochSecond(1710036060L, 0, ZoneOffset.UTC)});
+    params.add(new Object[] {"America/St_Johns", LocalDateTime.ofEpochSecond(1710036060L, 0, ZoneOffset.UTC)});
+    params.add(new Object[] {"America/Chicago", LocalDateTime.ofEpochSecond(1710036060L, 0, ZoneOffset.UTC)});
+    params.add(new Object[] {"America/Indiana/Indianapolis", LocalDateTime.ofEpochSecond(1710036060L, 0, ZoneOffset.UTC)});
+    params.add(new Object[] {"America/Los_Angeles", LocalDateTime.ofEpochSecond(1710036060L, 0, ZoneOffset.UTC)});
```
zabetak, line 85: These 5 tests seem kinda identical maybe we just need one of them.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreUtils.java
```diff
@@ -67,9 +69,31 @@ public static Collection<Object[]> generateZoneTimestampPairs() {
...
+    // Timestamp and Date do not have the year 0000. So, the year 0000 gets converted to year 0001.
+    // This is one such test case with year as 0000: 0000-01-07 22:44:36
+    params.add(new Object[] {"Asia/Kolkata", LocalDateTime.ofEpochSecond(-62166618924L, 0, ZoneOffset.UTC)});
+    generateDaylightSavingTimestampPairs(params);
     return params;
   }
 
+  public static void generateDaylightSavingTimestampPairs(List<Object[]> params) {
+    // For U.S timezones, the timestamp 2024-03-10 02:01:00 falls between the transition from 02:00 AM to 03:00 AM for daylight savings
+    params.add(new Object[] {"America/Anchorage", LocalDateTime.ofEpochSecond(1710036060L, 0, ZoneOffset.UTC)});
```
zabetak, line 81: Instead of using the `LocalDateTime.ofEpochSecond` constructor you could use the basic constructor which allows you to specify all fields explicitly. By doing this the comment above becomes redundant since it is clear from the code what is the date.

zabetak (commented):
LGTM. I left some small comments but hopefully nothing major. Let's see how the tests go.

@KiranVelumuri Please update the PR/JIRA summary and description to focus on the changes that were done in the production (and why) rather than the test failures.

@mdayakar @dengzhhu653 This is related to code that you touched recently so you may want to take a look.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreUtils.java
```diff
@@ -92,7 +118,7 @@ public void testStringToDate() {
 
   @Test
   public void testStringToTimestamp() {
-    assertEquals(Timestamp.valueOf(timestamp), MetaStoreUtils.convertStringToTimestamp(timestamp));
+    assertEquals(timestamp, MetaStoreUtils.convertStringToTimestamp(localTimestampString));
```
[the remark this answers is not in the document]
zabetak: This comment is not relevant in the latest commit. Please ignore.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreUtils.java
KiranVelumuri: They are in different time zones with DST, so included them

zabetak (approved):
LGTM, will merge soon.

PR #1655, opened 2020-11-09
HIVE-24349: Client connection count is not printed correctly in HiveM…

ArkoSharma:

PR #5310, opened 2024-06-18
HIVE-28338: Client connection count is not correct in HiveMetaStore#close

wecharyu:
1. Revert [#1655](https://github.com/apache/hive/pull/1655).
2. Add a test class for `HiveMetaStoreClient` connections.

Bug fix, connection counts is used in remote metastore with `TTransport`, so we cannot decrease counts when transport is null.

No.

No.

New test class:
```bash
mvn test -Dtest.groups= -Dtest=org.apache.hadoop.hive.metastore.client.TestHiveMetaStoreClient  -pl :hive-standalone-metastore-server
```

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
@@ -445,6 +445,11 @@ public class HiveMetaStoreClient implements IMetaStoreClient, AutoCloseable {
     return transport;
   }
 
+  @VisibleForTesting
+  public static AtomicInteger getConnCount() {
+    return connCount;
+  }
+
   @Override
   public boolean isLocalMetaStore() {
     return localMetaStore;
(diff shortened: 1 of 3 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -1007,17 +1012,10 @@ public void close() {
...
-        if ((transport == null) || !transport.isOpen()) {
-          final int newCount = connCount.decrementAndGet();
-          LOG.info("Closed a connection to metastore, current connections: {}",
-                  newCount);
-        }
       }
     } catch (TException e) {
       LOG.debug("Unable to shutdown metastore client. Will try closing transport directly.", e);
     }
-    // Transport would have got closed via client.shutdown(), so we dont need this, but
```
wecharyu, line 1019: `client.shutdown()` calls would not close transport in client side, we must close transport explicitly.

dengzhhu653 (approved):

saihemanth-cloudera (approved):
LGTM +1

PR #5307, opened 2024-06-17
HIVE-28333: Skip CBO for UNIQUEJOIN and TABLESAMPLE

soumyakanti3578:
### What changes were proposed in this pull request?

Skip CBO for UNIQUEJOIN and TABLESAMPLE

### Why are the changes needed?

Refer https://issues.apache.org/jira/browse/HIVE-27831

### Does this PR introduce _any_ user-facing change?

No

### Is the change a dependency upgrade?

No

### How was this patch tested?

By running various Ptests

PR #5312, opened 2024-06-18
HIVE-28193: Skip CBO fallback when EXCEPT/INTERSECT/QUALIFY are used

okumin:

Skip `hive.cbo.fallback.strategy=CONSERVATIVE|ALWAYS` when EXCEPT, INTERSECT, or QUALIFY are used.

https://issues.apache.org/jira/browse/HIVE-28193

Users see a confusing error message on hitting this problem. They are asked to enable CBO even though they enable CBO.

Practically, no.

No.

I added negative tests to check error messages.


deniskuzZ (commented):
LGTM, but I'll leave it to @kasakrisz, @zabetak  
just 1 thing, maybe a new check should be moved to `HiveCalciteUtil`

okumin:
> just 1 thing, maybe a new check should be moved to HiveCalciteUtil

Thanks. I looked at the file and I am sure that it is the right place

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
```diff
@@ -109,6 +110,25 @@ public static boolean validateASTForUnsupportedTokens(ASTNode ast) {
...
+        return true;
+      default:
+        for (int i = 0; i < node.getChildCount(); i++) {
+          if (requiresCBO(node.getChild(i))) {
+            return true;
+          }
+        }
+    }
+    return false;
+  }
```
zabetak, line 130: This method traverses the entire tree to find certain tokens something which was already done before at least once (during parse) or twice (during analyze). This method seems to incur some unnecessary overhead. There are some other classes such as `QueryProperties` which could help us avoid the extra traversals.

Moreover, we already have utility methods such as `ParseUtils.containsTokenOfType` which can detect if certain tokens exist by traversing the tree so it seems we don't need the new method.

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -688,7 +688,7 @@ Operator genOPTree(ASTNode ast, PlannerContext plannerCtx) throws SemanticExcept
           this.ctx.setCboInfo(cboMsg);
 
           // Determine if we should re-throw the exception OR if we try to mark the query to retry as non-CBO.
-          if (fallbackStrategy.isFatal(e)) {
+          if (fallbackStrategy.isFatal(e) || HiveCalciteUtil.requiresCBO(ast)) {
```
zabetak, line 691: The additional disjunctive clause changes a bit the behavior of the fallback strategies notably when the `ALWAYS/CONSERVATIVE` option is set. 

Nevertheless, it makes sense to avoid recompiling a query when we know that it will fail so the check fits here. Let's just check if we really need a new method and a tree traversal as I mentioned in the other comment.

deniskuzZ:
@okumin FYI, something similar is done in https://github.com/apache/hive/pull/5307

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
okumin: Thanks for the advice. I checked the potential where we can skip the extra traversal, and `QueryProperties` could be an option. I attached a debugger and confirmed it sounds like an up-and-coming option if we can accept a new attribute like `QueryProperties#requiresCBO.`
I will check the feasibility more in exceptional cases with sub-queries or CTEs. If the approach turns out to have another problem, I will try it using ParseUtils.
Thanks again.

zabetak: For the properties suggestion, I was thinking something similar to the existing design:
```
QueryProperties#hasIntersect
QueryProperties#hasExcept
QueryProperties#hasQualify
```

okumin:
I think @zabetak 's proposal is further better, and the last revision follows it. I tested some cases by hand and verified the approach would likely work in all known cases. I didn't add them to `qtest` because it sounds overkilling to test all the cases.

```
0: jdbc:hive2://hive-hiveserver2:10000/defaul> select cast(0 as bigint) = '1' except select cast(1 as bigint) = '1';
Error: Error while compiling statement: FAILED: SemanticException Line 0:-1 Wrong arguments ''1'': Unsafe compares between different types are disabled for safety reasons. If you know what you are doing, please set hive.strict.checks.type.safety to false and make sure that hive.mapred.mode is not set to 'strict' to proceed. Note that you may get errors or incorrect results if you make a mistake while using some of the unsafe features. (state=42000,code=40000)

0: jdbc:hive2://hive-hiveserver2:10000/defaul> select count(*) from (select cast(0 as bigint) = '1' except select cast(1 as bigint) = '1') x;
Error: Error while compiling statement: FAILED: SemanticException Line 0:-1 Wrong arguments ''1'': Unsafe compares between different types are disabled for safety reasons. If you know what you are doing, please set hive.strict.checks.type.safety to false and make sure that hive.mapred.mode is not set to 'strict' to proceed. Note that you may get errors or incorrect results if you make a mistake while using some of the unsafe features. (state=42000,code=40000)

0: jdbc:hive2://hive-hiveserver2:10000/defaul> with x as (select cast(0 as bigint) = '1' except select cast(1 as bigint) = '1') select count(*) from x;
Error: Error while compiling statement: FAILED: SemanticException Line 0:-1 Wrong arguments ''1'': Unsafe compares between different types are disabled for safety reasons. If you know what you are doing, please set hive.strict.checks.type.safety to false and make sure that hive.mapred.mode is not set to 'strict' to proceed. Note that you may get errors or incorrect results if you make a mistake while using some of the unsafe features. (state=42000,code=40000)

0: jdbc:hive2://hive-hiveserver2:10000/defaul> set hive.optimize.cte.materialize.threshold=1;
No rows affected (0.005 seconds)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> set hive.optimize.cte.materialize.full.aggregate.only=false;
No rows affected (0.006 seconds)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> with x as (select cast(0 as bigint) = '1' except select cast(1 as bigint) = '1') select count(*) from x;
Error: Error while compiling statement: FAILED: SemanticException Line 0:-1 Wrong arguments ''1'': Unsafe compares between different types are disabled for safety reasons. If you know what you are doing, please set hive.strict.checks.type.safety to false and make sure that hive.mapred.mode is not set to 'strict' to proceed. Note that you may get errors or incorrect results if you make a mistake while using some of the unsafe features. (state=42000,code=40000)
```

```
0: jdbc:hive2://hive-hiveserver2:10000/defaul> select cast(key as bigint) = '1' from src qualify row_number() over (partition by key order by value) = 1;
Error: Error while compiling statement: FAILED: SemanticException Line 0:-1 Wrong arguments ''1'': Unsafe compares between different types are disabled for safety reasons. If you know what you are doing, please set hive.strict.checks.type.safety to false and make sure that hive.mapred.mode is not set to 'strict' to proceed. Note that you may get errors or incorrect results if you make a mistake while using some of the unsafe features. (state=42000,code=40000)

0: jdbc:hive2://hive-hiveserver2:10000/defaul> select count(*) from (select cast(key as bigint) = '1' from src qualify row_number() over (partition by key order by value) = 1) x;
Error: Error while compiling statement: FAILED: SemanticException Line 0:-1 Wrong arguments ''1'': Unsafe compares between different types are disabled for safety reasons. If you know what you are doing, please set hive.strict.checks.type.safety to false and make sure that hive.mapred.mode is not set to 'strict' to proceed. Note that you may get errors or incorrect results if you make a mistake while using some of the unsafe features. (state=42000,code=40000)

0: jdbc:hive2://hive-hiveserver2:10000/defaul> with x as (select cast(key as bigint) = '1' from src qualify row_number() over (partition by key order by value) = 1) select count(*) from x;
Error: Error while compiling statement: FAILED: SemanticException Line 0:-1 Wrong arguments ''1'': Unsafe compares between different types are disabled for safety reasons. If you know what you are doing, please set hive.strict.checks.type.safety to false and make sure that hive.mapred.mode is not set to 'strict' to proceed. Note that you may get errors or incorrect results if you make a mistake while using some of the unsafe features. (state=42000,code=40000)

0: jdbc:hive2://hive-hiveserver2:10000/defaul> set hive.optimize.cte.materialize.threshold=1;
No rows affected (0.006 seconds)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> set hive.optimize.cte.materialize.full.aggregate.only=false;
No rows affected (0.006 seconds)
0: jdbc:hive2://hive-hiveserver2:10000/defaul> with x as (select cast(key as bigint) = '1' from src qualify row_number() over (partition by key order by value) = 1) select count(*) from x;
Error: Error while compiling statement: FAILED: SemanticException Line 0:-1 Wrong arguments ''1'': Unsafe compares between different types are disabled for safety reasons. If you know what you are doing, please set hive.strict.checks.type.safety to false and make sure that hive.mapred.mode is not set to 'strict' to proceed. Note that you may get errors or incorrect results if you make a mistake while using some of the unsafe features. (state=42000,code=40000)
```

ql/src/java/org/apache/hadoop/hive/ql/QueryProperties.java
```diff
@@ -49,6 +49,10 @@ public class QueryProperties {
   boolean hasJoinFollowedByGroupBy = false;
   boolean hasPTF = false;
   boolean hasWindowing = false;
+  boolean hasQualify = false;
+
```
zabetak, line 53: nit: space not necessary

ql/src/java/org/apache/hadoop/hive/ql/QueryProperties.java
```diff
@@ -302,6 +330,10 @@ public void clear() {
     hasJoinFollowedByGroupBy = false;
     hasPTF = false;
     hasWindowing = false;
+    hasQualify = false;
+
```
zabetak, line 334: nit: space not necessary

ql/src/test/queries/clientnegative/cbo_fallback_skip_with_except.q
```diff
@@ -0,0 +1,2 @@
+set hive.cbo.fallback.strategy=CONSERVATIVE;
+select cast(0 as bigint) = '1' except select cast(1 as bigint) = '1';
```
zabetak, line 2: Let's just keep this q test and remove the qualify which is more involved.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -683,15 +683,19 @@ void doPhase1QBExpr(ASTNode ast, QBExpr qbexpr, String id, String alias, boolean
...
       case HiveParser.TOK_INTERSECTDISTINCT:
+        queryProperties.setHasIntersect(true);
         qbexpr.setOpcode(QBExpr.Opcode.INTERSECT);
         break;
       case HiveParser.TOK_EXCEPTALL:
+        queryProperties.setHasExcept(true);
         qbexpr.setOpcode(QBExpr.Opcode.EXCEPTALL);
         break;
       case HiveParser.TOK_EXCEPTDISTINCT:
+        queryProperties.setHasExcept(true);
```
zabetak, line 698: Can you please add a few unit tests for the analyzer to test that the query properties are set correctly for the different types of queries. You can get some inspiration from `TestSemanticAnalyzer` and similar classes.

zabetak (approved):
LGTM. Some extra unit tests would be even better but don't feel very strongly about it.

ql/src/test/queries/clientnegative/cbo_fallback_skip_with_except.q
okumin: Removed the one with qualify
https://github.com/apache/hive/pull/5312/commits/1ad217d7bfb4147e265ea6f43914afc293b3cbbb

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
okumin: Thanks for the suggestion. I didn't have that idea. I tried to implement one, but I am not confident that this is the way. Please feel free to add any comments or suggestions to the new file.
https://github.com/apache/hive/pull/5312/commits/62e44adb4b202966479d4764e749340e939507a5

okumin:
Thanks a lot!

PR #5313, opened 2024-06-19
HIVE-28315: Missing classes while using hive jdbc standalone jar

dengzhhu653:
Add classes/trim dependencies to/from the jdbc standalone jar

- Solve the missing classes
- Reduce the jar size from 98M to 42M

No

No

  - Tested create/show/select/load queries, fetching the result;
  - Tested DatabaseMetaData;
  - Tested Kerberos & Delegation token & Simple authentication;
  - Tested using ZK for service discovery;
  - Inspected the classes inside the standalone jar.

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -240,7 +240,6 @@
                         <include>org/apache/hadoop/metrics2/**</include>
                         <include>org/apache/hadoop/security/token/**</include>
                         <include>org/apache/hadoop/util/*</include>
-                        <include>org/apache/hadoop/io/Writable.class</include>
                         <include>org/apache/hadoop/net/*</include>
                         <include>org/apache/hadoop/io/*</include>
                         <include>org/apache/hadoop/HadoopIllegalArgumentException.class</include>
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -34,11 +34,23 @@
       <groupId>org.apache.hive</groupId>
       <artifactId>hive-common</artifactId>
       <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.hive</groupId>
-      <artifactId>hive-llap-client</artifactId>
+      <artifactId>hive-standalone-metastore-common</artifactId>
       <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.hive</groupId>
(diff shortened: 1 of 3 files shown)


jdbc/pom.xml
```diff
@@ -189,17 +235,24 @@
...
+                        <include>org/apache/hadoop/fs/impl/*</include>
+                        <include>org/apache/hadoop/fs/statistics/*</include>
+                        <include>org/apache/hadoop/metrics2/**</include>
+                        <include>org/apache/hadoop/security/token/**</include>
+                        <include>org/apache/hadoop/util/*</include>
                         <include>org/apache/hadoop/io/Writable.class</include>
-                        <include>org/apache/hadoop/io/DataOutputBuffer.class</include>
-                        <include>org/apache/hadoop/io/DataInputBuffer.class</include>
+                        <include>org/apache/hadoop/net/*</include>
+                        <include>org/apache/hadoop/io/*</include>
```
deniskuzZ, line 245: should we remove `org/apache/hadoop/io/Writable.class` above?

jdbc/pom.xml
```diff
@@ -308,21 +369,27 @@
...
                       <exclude>org.codehaus.jettison:*</exclude>
                       <exclude>org.datanucleus:*</exclude>
+                      <exclude>org.eclipse.jetty:*</exclude>
                       <exclude>org.eclipse.jetty.aggregate:*</exclude>
+                      <exclude>org.ehcache:*</exclude>
                       <exclude>org.fusesource.leveldbjni:*</exclude>
                       <exclude>org.glassfish.jersey*:*</exclude>
                       <exclude>org.hamcrest:*</exclude>
                       <exclude>org.htrace:*</exclude>
+                      <exclude>org.jetbrains.kotlin:*</exclude>
```
deniskuzZ, line 392: what is this `jetbrains.kotlin`?

jdbc/pom.xml
```diff
@@ -189,17 +235,24 @@
...
                         <include>org/apache/hadoop/io/Writable.class</include>
-                        <include>org/apache/hadoop/io/DataOutputBuffer.class</include>
-                        <include>org/apache/hadoop/io/DataInputBuffer.class</include>
+                        <include>org/apache/hadoop/net/*</include>
+                        <include>org/apache/hadoop/io/*</include>
                         <include>org/apache/hadoop/HadoopIllegalArgumentException.class</include>
                       </includes>
                     </filter>
+                    <filter>
+                      <artifact>org.apache.hadoop:hadoop-auth</artifact>
```
deniskuzZ, line 250: previously it was excluded completely, why it that change?

dengzhhu653: This is for UGI to login the user from keytab, 
```
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/util/PlatformName
	at org.apache.hadoop.security.UserGroupInformation.getOSLoginModuleName(UserGroupInformation.java:434)
	at org.apache.hadoop.security.UserGroupInformation.<clinit>(UserGroupInformation.java:462)
	at org.example.App.main(App.java:45)

File: jdbc/pom.xml (hunk 2 above)
Comment (dengzhhu653) [reply jdbc/pom.xml:392]: It's introduced by hadoop-hdfs-client, 
```
[INFO] |  |  +- org.apache.hadoop:hadoop-hdfs-client:jar:3.3.6:compile
[INFO] |  |  |  +- com.squareup.okhttp3:okhttp:jar:4.9.3:compile
[INFO] |  |  |  |  \- com.squareup.okio:okio:jar:2.8.0:compile
[INFO] |  |  |  +- org.jetbrains.kotlin:kotlin-stdlib:jar:1.4.10:compile
[INFO] |  |  |  \- org.jetbrains.kotlin:kotlin-stdlib-common:jar:1.4.10:compile

deniskuzZ (approved):
LGTM +1

PR #5315, opened 2024-06-20
HIVE-28340: Test concurrent JDBC connections with Kerberized cluster, impersonation, and HTTP transport

zabetak:
1. New test case simulating JDBC clients concurrently accessing a Kerberized cluster, with impersonation and HTTP transport.
2. Set transport mode based on configuration when creating MiniHS2 from `MiniHiveKdc#getMiniHS2WithKerbWithRemoteHMSWithKerb`.

Increase test coverage and guard against regressions. More info under HIVE-28340.

No

No

```
mvn test -Dtest=TestJdbcWithMiniKdcDoAsHttp -pl itests/hive-minikdc -Pitests
```

### DIFF
--- a/itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/MiniHiveKdc.java
+++ b/itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/MiniHiveKdc.java
@@ -194,9 +194,6 @@ public class MiniHiveKdc {
                                               .withConf(hiveConf)
                                               .withMiniKdc(hivePrincipal, hiveKeytab)
                                               .withAuthenticationType(authType);
-    if (HiveServer2.isHttpTransportMode(hiveConf)) {
-      miniHS2Builder.withHTTPTransport();
-    }
     return miniHS2Builder.build();
  }
 
--- a/itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/MiniHiveKdc.java
+++ b/itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/MiniHiveKdc.java
@@ -194,6 +194,9 @@ public class MiniHiveKdc {
                                               .withConf(hiveConf)
                                               .withMiniKdc(hivePrincipal, hiveKeytab)
                                               .withAuthenticationType(authType);
+    if (HiveServer2.isHttpTransportMode(hiveConf)) {
+      miniHS2Builder.withHTTPTransport();
+    }
     return miniHS2Builder.build();
  }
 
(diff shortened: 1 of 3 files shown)


itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/TestJdbcWithMiniKdcDoAsHttp.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Test
+  public void testConcurrentConnectionsWithSessionOverlapNoSaslException() throws Exception {
+    ExecutorService service = Executors.newFixedThreadPool(2);
+    // Open, close, client connections repeatedly till interrupted; essentially leading to many HS2 sessions.
+    // Simulate a client application that uses many short-lived connections.
+    Future<Boolean> openCloseConnectionTask = service.submit(() -> {
+      do {
+        try (Connection c = DriverManager.getConnection(hs2.getJdbcURL())) {
```
dengzhhu653, line 69: I'm confusing, the `hs2.getJdbcURL()` should be a Kerberized connection, right?

itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/MiniHiveKdc.java
```diff
@@ -194,9 +194,6 @@ public static MiniHS2 getMiniHS2WithKerb(MiniHiveKdc miniHiveKdc, HiveConf hiveC
                                               .withConf(hiveConf)
                                               .withMiniKdc(hivePrincipal, hiveKeytab)
                                               .withAuthenticationType(authType);
-    if (HiveServer2.isHttpTransportMode(hiveConf)) {
```
dengzhhu653, line 197: why do we remove it here? what if the HS2 reads this property and opens the http server, while the client is trying to establish the connection with the binary port?

zabetak: The JDBC URL is usually obtained from `MiniHS2#getJdbcURL()` so the server/client configs should be aligned. 

Nevertheless to simplify the review of this PR I reverted the changes here and in MiniHS2 and added the same `isHttpTransportMode` logic in the the other static factory method that is used in this test (https://github.com/apache/hive/pull/5315/commits/add40a8b80d2be8ce5f03022c4206259bdb421de ).

zabetak:
Tests are green can you have another look @dengzhhu653 ? Thanks!

dengzhhu653 (approved):

PR #5318, opened 2024-06-24
HIVE-28345: Avoid redundant HiveConf creation in MiniHS2.Builder

zabetak:
Every creation of a MiniHS2.Builder object triggers the creation  of a HiveConf object. In many cases this new configuration object is thrown away and replaced by another conf object via the withConf method.

Creating a HiveConf object is computationally heavy so for performance reasons its best to avoid it when possible.

No

No

Existing tests


okumin (approved):

simhadri-g (approved):
LGTM +1

PR #5322, opened 2024-06-25
HIVE-28350: Drop remote database should succeed deleting data under d…

saihemanth-cloudera:
…b location

Setting remote database location and managed location

Otherwise drop remote database is not successful

No

No

Local Cluster


zhangbutao (approved):
+1 LGTM  Pending a green CI.
BTW, this case only failed when disabling direct sql `hive.metastore.try.direct.sql.ddl=false`

dengzhhu653:
I'm not sure how the fix can solve the problem, might the `mdb.getLocationUri()` or `mdb.getManagedLocationUri()` is null for remote database?

saihemanth-cloudera:
> I'm not sure how the fix can solve the problem, might the `mdb.getLocationUri()` or `mdb.getManagedLocationUri()` is null for remote database?

Yeah, because the location URI is null, we would hit null pointer exception here: https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java#L1803-L1804

PR #5323, opened 2024-06-26
HIVE-28347: Make a UDAF 'collect_set' work with complex types, even when map-side aggregation is disabled.

JeongDaeKim:

UDAFs, collect_set and collect_list, won't work with complex types, if map-side aggregation is  disabled. This PR will fix this bug.

When users disable the map-side aggregation feature(`"hive.map.aggr`), collect_set doesn't work with complex types, such as struct.

No.

No.

Tested with qtest (TestMiniTezCliDriver)

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
@@ -63,7 +63,6 @@ public class GenericUDAFMkCollectionEvaluator extends GenericUDAFEvaluator
     // init output object inspectors
     // Mode.PARTIAL1 or Mode.COMPLETE: T => List[T]
     // Mode.PARTIAL2 or Mode.FINAL: List[T] => List[T]
-    // The output of a partial aggregation is a list, output of COMPLETE is List as well
     if (mode == Mode.PARTIAL1 || mode == Mode.COMPLETE) {
       inputOI = parameters[0];
       return ObjectInspectorFactory.getStandardListObjectInspector(
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
@@ -61,13 +61,13 @@ public class GenericUDAFMkCollectionEvaluator extends GenericUDAFEvaluator
       throws HiveException {
     super.init(m, parameters);
     // init output object inspectors
-    // Mode.PARTIAL1 or Mode.COMPLETE: T => List[T]
-    // Mode.PARTIAL2 or Mode.FINAL: List[T] => List[T]
     if (mode == Mode.PARTIAL1 || mode == Mode.COMPLETE) {
+      // T => List[T]
       inputOI = parameters[0];
       return ObjectInspectorFactory.getStandardListObjectInspector(
           ObjectInspectorUtils.getStandardObjectInspector(inputOI));
     } else {
+      // List[T] => List[T]
       internalMergeOI = (ListObjectInspector) parameters[0];
       inputOI = internalMergeOI.getListElementObjectInspector();
       loi = (StandardListObjectInspector)
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
@@ -61,7 +61,9 @@ public class GenericUDAFMkCollectionEvaluator extends GenericUDAFEvaluator
       throws HiveException {
     super.init(m, parameters);
     // init output object inspectors
-    // The output of a partial aggregation is a list
+    // Mode.PARTIAL1 or Mode.COMPLETE: T => List[T]
+    // Mode.PARTIAL2 or Mode.FINAL: List[T] => List[T]
+    // The output of a partial aggregation is a list, output of COMPLETE is List as well
     if (mode == Mode.PARTIAL1 || mode == Mode.COMPLETE) {
       inputOI = parameters[0];
       return ObjectInspectorFactory.getStandardListObjectInspector(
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
```diff
@@ -66,18 +66,17 @@ public ObjectInspector init(Mode m, ObjectInspector[] parameters)
...
-        internalMergeOI = (ListObjectInspector) parameters[0];
-        inputOI = internalMergeOI.getListElementObjectInspector();
-        loi = (StandardListObjectInspector)
-            ObjectInspectorUtils.getStandardObjectInspector(internalMergeOI);
-        return loi;
-      }
+      //no map aggregation.
+      inputOI = parameters[0];
+      return ObjectInspectorFactory.getStandardListObjectInspector(
+          ObjectInspectorUtils.getStandardObjectInspector(inputOI));
```
okumin, line 79: Can we merge this case with PARTIAL1?

```
if (m == Mode.PARTIAL1 || m == Mode.COMPLETE) {
  ...
```

okumin:
@JeongDaeKim Taking a glance, your finding and fix seems to be very reasonable. Could you please regenerate the result file of `udaf_collect_set_2.q`?

JeongDaeKim:
@okumin Thank you for a quick review! 👍

okumin (approved):
I have done three things.
First, I checked that the evaluator is likely correctly implemented now.
Second, I checked that we have new regrettion tests. Nice work.
Third, I ran the original `udaf_collect_set_2.q` with `set hive.map.aggr = true or false`. Both generated the same result set.

Then, the changes here look to me. Note that you also need +1 from a committer.

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
```diff
@@ -62,22 +62,16 @@ public ObjectInspector init(Mode m, ObjectInspector[] parameters)
...
-        inputOI = ObjectInspectorUtils.getStandardObjectInspector(parameters[0]);
-        return ObjectInspectorFactory.getStandardListObjectInspector(inputOI);
-      } else {
-        internalMergeOI = (ListObjectInspector) parameters[0];
-        inputOI = internalMergeOI.getListElementObjectInspector();
-        loi = (StandardListObjectInspector)
-            ObjectInspectorUtils.getStandardObjectInspector(internalMergeOI);
-        return loi;
-      }
+      internalMergeOI = (ListObjectInspector) parameters[0];
```
deniskuzZ, line 70: is it safe to remove
`!(parameters[0] instanceof ListObjectInspector)`
condition?

JeongDaeKim: Yes, I think that condition was just to check "no map aggregation" or not, not for safe type casting.
As comments said above, "The output of a partial aggregation is a list"

https://github.com/apache/hive/blob/588c2ad9bd343b0ec1e6fa761bc36c0e00682ee6/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java#L43-L45

okumin: I also guess so. `COLLECT_LIST|SET` processes objects of a type T like that.
- `Mode.PARTIAL1` or `Mode.COMPLETE`: T => List[T]
- `Mode.PARTIAL2` or `Mode.FINAL`: List[T] => List[T]

The original one tried to check the type of inputs, T or List[T], to conclude the current mode. But it can lead to a wrong conclusion when T is List[X] because both `List[X]` and `List[List[X]]` hit the condition.

deniskuzZ: 👍, thanks for the explanation!

deniskuzZ (approved):
LGTM +1

ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java
deniskuzZ: could we update the comment with what @okumin wrote:
````
Mode.PARTIAL1 or Mode.COMPLETE: T => List[T]
Mode.PARTIAL2 or Mode.FINAL: List[T] => List[T]
````
`The output of a partial aggregation is a list` is kinda confusing, PARTIAL2 is also partial

see comments at the top
// For PARTIAL1 and COMPLETE: ObjectInspectors for original data
// For PARTIAL2 and FINAL: ObjectInspectors for partial aggregations (list of objs)

Comment (JeongDaeKim) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java:72]: https://github.com/apache/hive/pull/5323/commits/74e7ce7087682c2a9c23a25b9c1423840baf3b0a

Comments updated. (I don't know this comments are necessary though.)

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java:72]: @JeongDaeKim , sorry fo the confusion, no need for `// The output of a partial aggregation is a list, output of COMPLETE is List as well`. Just keep the mapping explanation

File: ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDAFMkCollectionEvaluator.java [hunk 3]
```diff
@@ -61,23 +61,19 @@ public ObjectInspector init(Mode m, ObjectInspector[] parameters)
       throws HiveException {
     super.init(m, parameters);
     // init output object inspectors
-    // The output of a partial aggregation is a list
-    if (m == Mode.PARTIAL1) {
+    // Mode.PARTIAL1 or Mode.COMPLETE: T => List[T]
+    // Mode.PARTIAL2 or Mode.FINAL: List[T] => List[T]
+    // The output of a partial aggregation is a list, output of COMPLETE is List as well
```
deniskuzZ, line 66: could we please inline the comments
````
if (mode == Mode.PARTIAL1 || mode == Mode.COMPLETE) {
      // T => List[T]
      inputOI = parameters[0];
      return ObjectInspectorFactory.getStandardListObjectInspector(
          ObjectInspectorUtils.getStandardObjectInspector(inputOI));
    } else {
      // List[T] => List[T]
      ....
}
````

deniskuzZ:
thanks @JeongDaeKim! I'll merge it with the next green build

PR #5325, opened 2024-06-27
HIVE-28353: Iceberg: Reading *Files Metadata table files if the column is of TIMESTAMP type.

ayushtkn:

Use the correct type while fetching Metadata Table, with tables having TIMESTAMP

To fix reading Metadata Tables where the main table has a column of TIMESTAMP type

Yes, If the tables of TIMESTAMP data type, then *FILES Metadata tables can be read

No

UT


simhadri-g (approved):
LGTM +1

PR #5327, opened 2024-06-27
HIVE-28355: Fix intermittent failure of TestHplSqlViaBeeLine#testUNIX_TIMESTAMPHplSQLFunction

KiranVelumuri:

Compare absolute UNIX timestamps value instead of using regex.

https://issues.apache.org/jira/browse/HIVE-28355

No

mvn test

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
+++ b/itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
@@ -34,6 +34,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.apache.hadoop.hive.UtilsForTest;
@@ -45,7 +46,7 @@ import org.junit.Test;
 
 public class TestHplSqlViaBeeLine {
   private static MiniHS2 miniHS2;
-  private static final String userName = System.getProperty("user.name");
+  private static final String USER_NAME = System.getProperty("user.name");
 
   /**
    * Start up a local Hive Server 2 for these tests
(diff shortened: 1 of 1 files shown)


KiranVelumuri:
There were Sonarcloud issues for variable names not complying with the naming convention for some variables. So, I made the appropriate refactoring changes for these variables.

KiranVelumuri:
@okumin Please review

okumin:
I might need some more time since I am not familiar with the issue, maybe. Also, please note that you finally need to get +1 from a committer

okumin (approved):
The description of HIVE-28355 is super great. Thanks for the detailed explanation. This change would resolve the problem.

For reference, I am putting alternatives that came to my mind. I don't have preference and the current PR looks fine to me.
- Just to remove the non-deterministic test case
- To change the expected text to `{System.currentTimeMillis()/10000}|{System.currentTimeMillis()/10000 + 1}`

shivjha30:
Have you worked on it?

KiranVelumuri:
> The description of [HIVE-28355](https://issues.apache.org/jira/browse/HIVE-28355) is super great. Thanks for the detailed explanation. This change would resolve the problem.
> 
> For reference, I am putting alternatives that came to my mind. I don't have preference and the current PR looks fine to me.
> 
> * Just to remove the non-deterministic test case
> * To change the expected text to `{System.currentTimeMillis()/10000}|{System.currentTimeMillis()/10000 + 1}`
> * Control the output using `hive.test.currenttimestamp`

Thank you for the quick review. The first alternative would not work as the test result depends on the time of execution of test. For the latter one, the test would fail when the timestamp differs other than ten's digit. Could you please tag a committer to get +1? Thank you.

KiranVelumuri:
@zhangbutao Could you also please review this? Thank you.

This issue is occuring intermittently depending on the time of execution of the test. 
Screenshot from recent occurence:
image

zhangbutao (approved):
+1 LGTM
Thanks for the fix!

KiranVelumuri:
Thank you @okumin and @zhangbutao for the review!

PR #5328, opened 2024-06-27
HIVE-28341: Iceberg: Change Major QB Full Table Compaction to compact…

difin:
… partition by partition

Change Major QB Full Table Compaction to compact partitions in parallel.

Currently, Iceberg Major compaction compacts a whole table in one step. If a table is partitioned and has a lot of data this operation can take a lot of time and it risks getting write conflicts at the commit stage. This PR proposes to improve it to compacts in parallel. Also, for each partition it will create one snapshot instead of 2 snapshots (truncate+IOW) created now when compacting the whole table in one step.

No

No

New query test added and updated existing iceberg compaction query tests with the new correct expected results.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -604,7 +604,7 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
    */
   private void commitCompaction(Table table, long startTime, FilesForCommit results,
       RewritePolicy rewritePolicy, Integer partitionSpecId, String partitionPath) {
-    if (rewritePolicy == RewritePolicy.ALL_PARTITIONS) {
+    if (rewritePolicy == RewritePolicy.FULL_TABLE) {
       // Full table compaction
       Transaction transaction = table.newTransaction();
       DeleteFiles delete = transaction.newDelete();
@@ -617,13 +617,10 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
       LOG.debug("Compacted full table with files {}", results);
     } else {
       // Single partition compaction
-      List<DataFile> existingDataFiles = partitionPath != null ?
-          IcebergTableUtil.getDataFiles(table, partitionSpecId, partitionPath) :
-          IcebergTableUtil.getDataFilesNotInSpec(table, partitionSpecId);
-
-      List<DeleteFile> existingDeleteFiles = partitionPath != null ?
-          IcebergTableUtil.getDeleteFiles(table, partitionSpecId, partitionPath) :
-          IcebergTableUtil.getDeleteFilesNotInSpec(table, partitionSpecId);
+      List<DataFile> existingDataFiles =
+          IcebergTableUtil.getDataFiles(table, partitionSpecId, partitionPath, partitionPath != null);
+      List<DeleteFile> existingDeleteFiles =
+          IcebergTableUtil.getDeleteFiles(table, partitionSpecId, partitionPath, partitionPath != null);
 
       RewriteFiles rewriteFiles = table.newRewrite();
       rewriteFiles.validateFromSnapshot(table.currentSnapshot().snapshotId());
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -604,10 +604,6 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
    */
   private void commitCompaction(Table table, long startTime, FilesForCommit results,
       RewritePolicy rewritePolicy, Integer partitionSpecId, String partitionPath) {
-    if (results.dataFiles().isEmpty()) {
-      LOG.info("Empty compaction commit, took {} ms for table: {}", System.currentTimeMillis() - startTime, table);
-      return;
-    }
     if (rewritePolicy == RewritePolicy.ALL_PARTITIONS) {
       // Full table compaction
       Transaction transaction = table.newTransaction();
@@ -621,8 +617,13 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
       LOG.debug("Compacted full table with files {}", results);
     } else {
       // Single partition compaction
-      List<DataFile> existingDataFiles = IcebergTableUtil.getDataFiles(table, partitionSpecId, partitionPath);
-      List<DeleteFile> existingDeleteFiles = IcebergTableUtil.getDeleteFiles(table, partitionSpecId, partitionPath);
+      List<DataFile> existingDataFiles = partitionPath != null ?
+          IcebergTableUtil.getDataFiles(table, partitionSpecId, partitionPath) :
+          IcebergTableUtil.getDataFilesNotInSpec(table, partitionSpecId);
+
+      List<DeleteFile> existingDeleteFiles = partitionPath != null ?
+          IcebergTableUtil.getDeleteFiles(table, partitionSpecId, partitionPath) :
+          IcebergTableUtil.getDeleteFilesNotInSpec(table, partitionSpecId);
 
       RewriteFiles rewriteFiles = table.newRewrite();
       rewriteFiles.validateFromSnapshot(table.currentSnapshot().snapshotId());
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -35,6 +35,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
@@ -618,9 +619,11 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
     } else {
       // Single partition compaction
       List<DataFile> existingDataFiles =
-          IcebergTableUtil.getDataFiles(table, partitionSpecId, partitionPath, partitionPath != null);
+          IcebergTableUtil.getDataFiles(table, partitionSpecId, partitionPath,
+              partitionPath == null ? Predicate.isEqual(partitionSpecId).negate() : Predicate.isEqual(partitionSpecId));
       List<DeleteFile> existingDeleteFiles =
-          IcebergTableUtil.getDeleteFiles(table, partitionSpecId, partitionPath, partitionPath != null);
+          IcebergTableUtil.getDeleteFiles(table, partitionSpecId, partitionPath,
+              partitionPath == null ? Predicate.isEqual(partitionSpecId).negate() : Predicate.isEqual(partitionSpecId));
 
       RewriteFiles rewriteFiles = table.newRewrite();
       rewriteFiles.validateFromSnapshot(table.currentSnapshot().snapshotId());
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -1913,10 +1913,12 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
 
   private boolean hasUndergonePartitionEvolution(Table table) {
     // If it is a table which has undergone partition evolution, return true.
+    // if a table has undergone partition evolution, the current spec is not necessary the latest which can happen
+    // if partition spec was changed to one of table's past specs.
     return table.currentSnapshot() != null &&
         table.currentSnapshot().allManifests(table.io()).parallelStream()
         .map(ManifestFile::partitionSpecId)
-        .anyMatch(id -> id < table.spec().specId());
+        .anyMatch(id -> id != table.spec().specId());
   }
 
   private boolean isIdentityPartitionTable(org.apache.hadoop.hive.ql.metadata.Table table) {
@@ -1946,7 +1948,7 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
 
   public boolean isPartitioned(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
     Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
-    return IcebergTableUtil.isPartitioned(table);
+    return table.spec().isPartitioned();
   }
 
   @Override
@@ -2140,8 +2142,8 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   }
 
   @Override
-  public boolean isUndergonePartitionEvolution(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
+  public boolean hasUndergonePartitionEvolution(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
     Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
-    return table.specs().size() > 1;
+    return hasUndergonePartitionEvolution(table);
   }
 }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -1937,7 +1937,8 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   @Override
   public List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table,
       Map<String, String> partitionSpec, boolean latestSpecOnly) throws SemanticException {
-    return getPartitionNames(table, partitionSpec, latestSpecOnly).stream()
+    Table icebergTable = IcebergTableUtil.getTable(conf, table.getTTable());
+    return IcebergTableUtil.getPartitionNames(icebergTable, partitionSpec, latestSpecOnly).stream()
         .map(partName -> {
           Map<String, String> partSpecMap = Maps.newLinkedHashMap();
           Warehouse.makeSpecFromName(partSpecMap, new Path(partName), null);
@@ -1974,22 +1975,10 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
    * @return A list of partition values which satisfies the partition spec provided corresponding to the table.
    * @throws SemanticException Exception raised when there is an issue performing a scan on the partitions table.
    */
-  @Override
   public List<String> getPartitionNames(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
-      Map<String, String> partitionSpec, boolean latestSpecOnly) throws SemanticException {
+      Map<String, String> partitionSpec) throws SemanticException {
     Table icebergTable = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
-
-    try {
-      return IcebergTableUtil
-          .getPartitionInfo(icebergTable, partitionSpec, true, latestSpecOnly).entrySet().stream()
-          .map(e -> {
-            PartitionData partitionData = e.getKey();
-            int specId = e.getValue();
-            return icebergTable.specs().get(specId).partitionToPath(partitionData);
-          }).collect(Collectors.toList());
-    } catch (IOException e) {
-      throw new SemanticException(String.format("Error while fetching the partitions due to: %s", e));
-    }
+    return IcebergTableUtil.getPartitionNames(icebergTable, partitionSpec, true);
   }
 
   /**
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -1912,9 +1912,8 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
   }
 
   private boolean hasUndergonePartitionEvolution(Table table) {
-    // If it is a table which has undergone partition evolution, return true.
-    // if a table has undergone partition evolution, the current spec is not necessary the latest which can happen
-    // if partition spec was changed to one of table's past specs.
+    // The current spec is not necessary the latest which can happen when partition spec was changed to one of
+    // table's past specs.
     return table.currentSnapshot() != null &&
         table.currentSnapshot().allManifests(table.io()).parallelStream()
         .map(ManifestFile::partitionSpecId)
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -1935,13 +1935,18 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
 
   @Override
   public List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table,
-      Map<String, String> partitionSpec) throws SemanticException {
-    return getPartitionNames(table, partitionSpec).stream()
+      Map<String, String> partitionSpec, boolean latestSpecOnly) throws SemanticException {
+    return getPartitionNames(table, partitionSpec, latestSpecOnly).stream()
         .map(partName -> {
           Map<String, String> partSpecMap = Maps.newLinkedHashMap();
           Warehouse.makeSpecFromName(partSpecMap, new Path(partName), null);
           return new DummyPartition(table, partName, partSpecMap);
         }).collect(Collectors.toList());
+  }
+
+  public boolean isPartitioned(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
+    Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
+    return IcebergTableUtil.isPartitioned(table);
   }
 
   @Override
@@ -1970,12 +1975,13 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
    */
   @Override
   public List<String> getPartitionNames(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
-      Map<String, String> partitionSpec) throws SemanticException {
+      Map<String, String> partitionSpec, boolean latestSpecOnly) throws SemanticException {
     Table icebergTable = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
 
     try {
       return IcebergTableUtil
-          .getPartitionInfo(icebergTable, partitionSpec, true).entrySet().stream().map(e -> {
+          .getPartitionInfo(icebergTable, partitionSpec, true, latestSpecOnly).entrySet().stream()
+          .map(e -> {
             PartitionData partitionData = e.getKey();
             int specId = e.getValue();
             return icebergTable.specs().get(specId).partitionToPath(partitionData);
@@ -2131,5 +2137,11 @@ public class HiveIcebergStorageHandler implements HiveStoragePredicateHandler, H
             Operation.DELETE.name());
     tableDesc.setProperty(HiveCustomStorageHandlerUtils.WRITE_OPERATION_CONFIG_PREFIX +
             tableDesc.getTableName(), Operation.DELETE.name());
+  }
+
+  @Override
+  public boolean isUndergonePartitionEvolution(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
+    Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
+    return table.specs().size() > 1;
   }
 }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
@@ -511,4 +511,19 @@ public class IcebergTableUtil {
 
     return result;
   }
+
+  public static List<String> getPartitionNames(Table icebergTable, Map<String, String> partitionSpec,
+      boolean latestSpecOnly) throws SemanticException {
+    try {
+      return IcebergTableUtil
+          .getPartitionInfo(icebergTable, partitionSpec, true, latestSpecOnly).entrySet().stream()
+          .map(e -> {
+            PartitionData partitionData = e.getKey();
+            int specId = e.getValue();
+            return icebergTable.specs().get(specId).partitionToPath(partitionData);
+          }).collect(Collectors.toList());
+    } catch (IOException e) {
+      throw new SemanticException(String.format("Error while fetching the partitions due to: %s", e));
+    }
+  }
 }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.function.BinaryOperator;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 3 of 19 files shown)


ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -493,7 +493,8 @@ public static CompactionResponse scheduleCompactionIfRequired(CompactionInfo ci,
 
       checkInterrupt(Initiator.class.getName());
 
-      CompactionType type = checkForCompaction(ci, validWriteIds, sd, t.getParameters(), runAs, txnHandler, conf);
+      CompactionType type = MetaStoreUtils.isIcebergTable(t.getParameters()) ? ci.type :
```
deniskuzZ, line 496: should we move iceberg check inside of `checkForCompaction`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -63,20 +63,33 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+        HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.PARTITION.name());
+        conf.set(IcebergCompactionService.PARTITION_SPEC_ID, String.valueOf(icebergTable.spec().specId()));
+        // A single filter on a virtual column causes errors during compilation,
+        // added another filter on file_path as a workaround.
+        compactionQuery = String.format("insert overwrite table %1$s select * from %1$s " +
+                "where %2$s != %3$d and %4$s is not null",
+            compactTableName, VirtualColumn.PARTITION_SPEC_ID.getName(), icebergTable.spec().specId(),
+            VirtualColumn.FILE_PATH.getName());
+      } else {
+        return true;
```
deniskuzZ, line 86: what is this branch for? should we log something here?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
```diff
@@ -63,20 +63,33 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+    org.apache.hadoop.hive.ql.metadata.Table table = Hive.get(conf).getTable(context.getTable().getDbName(),
+        context.getTable().getTableName());
+    Table icebergTable = IcebergTableUtil.getTable(conf, table.getTTable());
     String compactionQuery;
 
     if (partSpec == null) {
-      HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.ALL_PARTITIONS.name());
-      compactionQuery = String.format("insert overwrite table %s select * from %<s", compactTableName);
+      if (!IcebergTableUtil.isPartitioned(icebergTable)) {
+        HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.ALL_PARTITIONS.name());
```
deniskuzZ, line 73: if table is unpartitioned, why policy is RewritePolicy.ALL_PARTITIONS?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -484,10 +507,16 @@ public static Map<PartitionData, Integer> getPartitionInfo(Table icebergTable, M
...
+                    (entry.getKey().size() == partSpecMap.size() || allowPartialSpec) &&
+                    (entry.getValue() == icebergTable.spec().specId() || !latestSpecOnly);
               }).forEach(entry -> result.put(entry.getKey(), entry.getValue())));
     }
 
     return result;
   }
+
+  public static boolean isPartitioned(Table table) {
+    return IcebergTableUtil.getPartitionFields(table).size() > 0;
```
deniskuzZ, line 519: do we need this method? can we use `table.spec().isUnpartitioned()`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -2112,4 +2118,10 @@ public List<FileStatus> getMergeTaskInputFiles(Properties properties) throws IOE
   public MergeTaskProperties getMergeTaskProperties(Properties properties) {
     return new IcebergMergeTaskProperties(properties);
   }
+
+  @Override
+  public boolean isUndergonePartitionEvolution(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
+    Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
+    return table.specs().size() > 1;
```
deniskuzZ, line 2125: `HiveIcebergStorageHandler#hasUndergonePartitionEvolution` has similar logic, can we reuse?
number of specs doesn't mean that all of them are referenced

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1923,15 +1923,20 @@ public Optional<ErrorMsg> isEligibleForCompaction(
...
+      Map<String, String> partitionSpec, boolean latestSpecOnly) throws SemanticException {
+    return getPartitionNames(table, partitionSpec, latestSpecOnly).stream()
         .map(partName -> {
           Map<String, String> partSpecMap = Maps.newLinkedHashMap();
           Warehouse.makeSpecFromName(partSpecMap, new Path(partName), null);
           return new DummyPartition(table, partName, partSpecMap);
         }).collect(Collectors.toList());
   }
 
+  public boolean isPartitioned(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
```
deniskuzZ, line 1935: can we keep 1 copy of this check

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -621,8 +617,13 @@ private void commitCompaction(Table table, long startTime, FilesForCommit result
       LOG.debug("Compacted full table with files {}", results);
     } else {
       // Single partition compaction
-      List<DataFile> existingDataFiles = IcebergTableUtil.getDataFiles(table, partitionSpecId, partitionPath);
-      List<DeleteFile> existingDeleteFiles = IcebergTableUtil.getDeleteFiles(table, partitionSpecId, partitionPath);
+      List<DataFile> existingDataFiles = partitionPath != null ?
```
deniskuzZ, line 620: i feel that this became overloaded, could we extract this into `CompactionOperation` 
see examples in iceberg: 
1. CopyOnWriteOperation extends BaseBatchWrite
2. DynamicOverwrite extends BaseBatchWrite

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -621,8 +617,13 @@ private void commitCompaction(Table table, long startTime, FilesForCommit result
...
+          IcebergTableUtil.getDataFiles(table, partitionSpecId, partitionPath) :
+          IcebergTableUtil.getDataFilesNotInSpec(table, partitionSpecId);
+
+      List<DeleteFile> existingDeleteFiles = partitionPath != null ?
+          IcebergTableUtil.getDeleteFiles(table, partitionSpecId, partitionPath) :
```
deniskuzZ, line 625: should we add extra param like `latestSpecOnly` in IcebergTableUtil.getDataFiles instead. of creating multiple methods?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -104,6 +106,14 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
                 compactionRequest, ServerUtils.hostname(), txnHandler, context.getConf());
         parseCompactionResponse(compactionResponse, table, partitionMapEntry.getKey());
       }
+      // If Iceberg table had partition evolution, it will create compaction request without partition specification,
+      // and it will compact all files from old partition specs, besides compacting partitions of current spec in parallel.
+      if (DDLUtils.isIcebergTable(table) && table.getStorageHandler().isPartitioned(table) &&
+          table.getStorageHandler().isUndergonePartitionEvolution(table)) {
+        compactionRequest.setPartitionname(null);
```
deniskuzZ, line 113: isn't that a default , null?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -89,7 +90,8 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
     }
 
     //Will directly initiate compaction if an un-partitioned table/a partition is specified in the request
-    if (desc.getPartitionSpec() != null || !table.isPartitioned()) {
+    if (desc.getPartitionSpec() != null || (!table.isPartitioned() && !DDLUtils.isIcebergTable(table) ) ||
```
deniskuzZ, line 93: could we simplify 
````
!(table.isPartitioned() || table.getStorageHandler().isPartitioned(table))
````

Comment (deniskuzZ) [inline ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java:93]: could we simplify 
````
!(table.isPartitioned() || table.getStorageHandler().isPartitioned(table))
````

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java [hunk 11]
```diff
@@ -135,7 +145,7 @@ private List<Partition> getPartitions(Table table, AlterTableCompactDesc desc, D
     List<Partition> partitions = new ArrayList<>();
 
     if (desc.getPartitionSpec() == null) {
-      if (table.isPartitioned()) {
+      if (table.isPartitioned() || (DDLUtils.isIcebergTable(table) && table.getStorageHandler().isPartitioned(table))) {
```
deniskuzZ, line 148: why check for iceberg?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -104,6 +106,14 @@ public AlterTableCompactOperation(DDLOperationContext context, AlterTableCompact
                 compactionRequest, ServerUtils.hostname(), txnHandler, context.getConf());
         parseCompactionResponse(compactionResponse, table, partitionMapEntry.getKey());
       }
+      // If Iceberg table had partition evolution, it will create compaction request without partition specification,
+      // and it will compact all files from old partition specs, besides compacting partitions of current spec in parallel.
+      if (DDLUtils.isIcebergTable(table) && table.getStorageHandler().isPartitioned(table) &&
```
deniskuzZ, line 111: why check for iceberg?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Yes, agreed, added a comment and an exception

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: For partitioned tables, starting with this PR, RewritePolicy.ALL_PARTITIONS won’t be used anymore, because if table is partitioned it will always create compaction request per partition and will use RewritePolicy.PARTITION for each partition.
RewritePolicy.ALL_PARTITIONS is used now in case the table is unpartitioned which means it compacts entire table.
Should we rename it to a proper name, like RewritePolicy.UNPARTITIONED or RewritePolicy.FULL_TABLE?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Yes, I didn't notice hasUndergonePartitionEvolution method

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: Yes, but it reuses the same compaction request object when submitting compaction requests for different partitions modifying the compaction name. Existing ACID compaction functionality. Setting partition name to null because in that code block it is not null, but may have the value of the previous partition that was requested.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: I don't think it can be simplified like that..
Since `org.apache.hadoop.hive.ql.metadata.table.isPartitioned()` will always return false for Iceberg tables, it is not a suitable check for Iceberg, so I am checking `(!table.isPartitioned() && !DDLUtils.isIcebergTable(table))`, which check that the table is unpartitioned and is not Iceberg.

In another part of the condition I am checking if the table is Iceberg and if it is unpartitioned:
`(DDLUtils.isIcebergTable(table) && !table.getStorageHandler().isPartitioned(table))`
I think we have to check if the table is Iceberg because for native tables table.getStorageHandler() will be null.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: because `table.isPartitioned()` does not work for Iceberg. For Iceberg it will always return False. That's why I added  a call to getStorageHandler().isPartitioned(table).

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: I added a comment in the code to explain why I added that.
```
// If Iceberg table had partition evolution, it will create compaction request without partition specification,
// and it will compact all files from old partition specs, besides compacting partitions of current spec in parallel.
```
It check for Iceberg because it is needed only for Iceberg.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: Yes, I added one more parameter to the function and got rid of the overloaded functions.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
difin: I added one more parameter to the function and got rid of the overloaded functions, is it fine now?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1901,10 +1901,12 @@ public boolean canUseTruncate(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
 
   private boolean hasUndergonePartitionEvolution(Table table) {
     // If it is a table which has undergone partition evolution, return true.
+    // if a table has undergone partition evolution, the current spec is not necessary the latest which can happen
+    // if partition spec was changed to one of table's past specs.
```
deniskuzZ, line 1905: which can happen ` when` partition spec

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1901,10 +1901,12 @@ public boolean canUseTruncate(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
 
   private boolean hasUndergonePartitionEvolution(Table table) {
     // If it is a table which has undergone partition evolution, return true.
```
deniskuzZ, line 1903: remove this

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1901,10 +1901,12 @@ public boolean canUseTruncate(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
...
+    // if a table has undergone partition evolution, the current spec is not necessary the latest which can happen
```
deniskuzZ, line 1904: just keep the part about the current spec, cause it's valid not only for partition evolution case
`the current spec is not necessary the latest which can happen when partition spec was changed to one of table's past specs.`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: wouldn't it be better to move this into `IcebergTableUtil`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -393,28 +393,47 @@ public static PartitionData toPartitionData(StructLike sourceKey, Types.StructTy
...
+  /**
+   * Returns list of data files filtered by specId and partitionPath as following:
+   *  1. If matchBySpecId is true, then filters files by specId == file's specId, else by specId != file's specId
+   *  2. If partitionPath is not null, then also filters files where partitionPath == file's partition path
+   * @param table the iceberg table
+   * @param specId partition spec id
+   * @param partitionPath partition path
+   * @param matchBySpecId If true, searches for files matching the specId, else searches for ones that don't match
+   */
+  public static List<DataFile> getDataFiles(Table table, int specId, String partitionPath, boolean matchBySpecId) {
```
deniskuzZ, line 405: wouldn't it be better to pass a filter
````
public static List<DataFile> getDataFiles(Table table, Predicate<DataFile> filter);
````
getDataFiles(table, file -> file.specId() == specId)
getDataFiles(table, file -> file.specId() != specId)

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java [hunk 17]
```diff
@@ -63,20 +63,35 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+        // A single filter on a virtual column causes errors during compilation,
+        // added another filter on file_path as a workaround.
+        compactionQuery = String.format("insert overwrite table %1$s select * from %1$s " +
+                "where %2$s != %3$d and %4$s is not null",
+            compactTableName, VirtualColumn.PARTITION_SPEC_ID.getName(), icebergTable.spec().specId(),
+            VirtualColumn.FILE_PATH.getName());
+      } else {
+        // Partitioned table without partition evolution with partition spec as null in the compaction request - this
+        // code branch is not supposed to be reachable
+        throw new HiveException(ErrorMsg.NO_COMPACTION_PARTITION);
```
deniskuzZ, line 88: maybe `COMPACTION_NO_PARTITION`

ql/src/java/org/apache/hadoop/hive/ql/Context.java
```diff
@@ -255,7 +255,7 @@ public String toString() {
   public enum RewritePolicy {
 
     DEFAULT,
-    ALL_PARTITIONS,
+    FULL_TABLE,
```
deniskuzZ, line 258: could we please make it 3rd in a list

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -2112,4 +2120,10 @@ public List<FileStatus> getMergeTaskInputFiles(Properties properties) throws IOE
   public MergeTaskProperties getMergeTaskProperties(Properties properties) {
     return new IcebergMergeTaskProperties(properties);
   }
+
+  @Override
+  public boolean hasUndergonePartitionEvolution(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
```
deniskuzZ, line 2125: wouldn't it be better to move this into IcebergTableUtil?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4055,7 +4055,7 @@ public List<String> getPartitionNames(String dbName, String tblName,
     List<String> names = null;
     Table t = getTable(dbName, tblName);
     if (t.getStorageHandler() != null && t.getStorageHandler().alwaysUnpartitioned()) {
-      return t.getStorageHandler().getPartitionNames(t, partSpec);
+      return t.getStorageHandler().getPartitionNames(t, partSpec, false);
```
deniskuzZ, line 4058: should we add the default method for simplicity, to avoid passing false

deniskuzZ, line 4058: should it be false?

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -778,10 +778,28 @@ default Optional<ErrorMsg> isEligibleForCompaction(org.apache.hadoop.hive.ql.met
     throw new UnsupportedOperationException("Storage handler does not support validating eligibility for compaction");
   }
 
+  /**
+   * Returns partitions names for the current table spec that correspond to the provided partition spec.
+   * @param hmsTable {@link org.apache.hadoop.hive.ql.metadata.Table} table metadata stored in Hive Metastore
+   * @param partitionSpec Map of Strings {@link java.util.Map} partition specification
+   * @return Optional of ErrorMsg {@link org.apache.hadoop.hive.ql.ErrorMsg}
```
deniskuzZ, line 785: is that copy-paste? it returns a list

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -778,10 +778,28 @@ default Optional<ErrorMsg> isEligibleForCompaction(org.apache.hadoop.hive.ql.met
...
       Map<String, String> partitionSpec) throws SemanticException {
+    return getPartitionNames(hmsTable, partitionSpec, true);
+  }
+
+  /**
+   * Returns partitions names that correspond to the provided partition spec.
+   * @param hmsTable {@link org.apache.hadoop.hive.ql.metadata.Table} table metadata stored in Hive Metastore
+   * @param partitionSpec Map of Strings {@link java.util.Map} partition specification
+   * @param latestSpecOnly Tells whether to return partition names for the latest spec only or for past specs too
+   * @return Optional of ErrorMsg {@link org.apache.hadoop.hive.ql.ErrorMsg}
```
deniskuzZ, line 797: same

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -839,9 +857,31 @@ default Partition getPartition(org.apache.hadoop.hive.ql.metadata.Table table, M
...
+   * @param latestSpecOnly Specifies if to return only partitions for the latest partition spec
+   * @return List of Partitions {@link org.apache.hadoop.hive.ql.metadata.Partition}
+   * @throws SemanticException {@link org.apache.hadoop.hive.ql.parse.SemanticException}
+   */
+  default List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table table,
+      Map<String, String> partitionSpec, boolean latestSpecOnly) throws SemanticException {
     throw new UnsupportedOperationException("Storage handler does not support getting partitions for a table.");
   }
 
+  default boolean isPartitioned(org.apache.hadoop.hive.ql.metadata.Table table) {
```
deniskuzZ, line 876: i don't think it should be here -> IcebergTableUtil

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -839,9 +857,31 @@ default Partition getPartition(org.apache.hadoop.hive.ql.metadata.Table table, M
...
+    throw new UnsupportedOperationException("Storage handler does not support checking if table is partitioned.");
+  }
+
+  default boolean hasUndergonePartitionEvolution(org.apache.hadoop.hive.ql.metadata.Table table) {
```
deniskuzZ, line 880: same -> IcebergTableUtil

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Nice idea, done.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -393,28 +393,47 @@ public static PartitionData toPartitionData(StructLike sourceKey, Types.StructTy
...
+  /**
+   * Returns list of delete files filtered by specId and partitionPath as following:
+   *  1. If matchBySpecId is true, then filters files by specId == file's specId, else by specId != file's specId
+   *  2. If partitionPath is not null, then also filters files where partitionPath == file's partition path
+   * @param table the iceberg table
+   * @param specId partition spec id
+   * @param partitionPath partition path
+   * @param matchBySpecId If true, searches for files matching the specId, else searches for ones that don't match
+   */
+  public static List<DeleteFile> getDeleteFiles(Table table, int specId, String partitionPath, boolean matchBySpecId) {
```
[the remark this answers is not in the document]
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergMajorQueryCompactor.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/Context.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: I added it here because I had to add it to Hive storage handler API.
It is called from Hive ql module

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
difin: It is called from Hive ql module, how can it be in IcebergTableUtil and not in storage handler API?

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
difin: This is also called from Hive ql module, can it be in IcebergTableUtil and not in storage handler API?

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
difin: Yes, done.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4298,7 +4298,7 @@ private List<Partition> getPartitionsWithAuth(Table tbl, Map<String, String> par
   public List<Partition> getPartitions(Table tbl, Map<String, String> partialPartSpec)
   throws HiveException {
     if (tbl.getStorageHandler() != null && tbl.getStorageHandler().alwaysUnpartitioned()) {
-      return tbl.getStorageHandler().getPartitions(tbl, partialPartSpec);
+      return tbl.getStorageHandler().getPartitions(tbl, partialPartSpec, false);
```
deniskuzZ, line 4301: same question, should it be false? you changed the other one

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
deniskuzZ: 👍

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
difin: No, because it is called from compaction flow when compaction request is called with a specific partition which can be partition of an old partition spec: https://github.com/apache/hive/blob/76a03b8173e36681d8824bd30920cadff0bfb4fa/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java#L143-L147

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -778,10 +778,28 @@ default Optional<ErrorMsg> isEligibleForCompaction(org.apache.hadoop.hive.ql.met
...
+  }
+
+  /**
+   * Returns partitions names that correspond to the provided partition spec.
+   * @param hmsTable {@link org.apache.hadoop.hive.ql.metadata.Table} table metadata stored in Hive Metastore
+   * @param partitionSpec Map of Strings {@link java.util.Map} partition specification
+   * @param latestSpecOnly Tells whether to return partition names for the latest spec only or for past specs too
+   * @return List of partition names
+   */
+  default List<String> getPartitionNames(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
```
deniskuzZ, line 799: why do we need to expose it in API if it's not used outside of StorageHandler?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
difin: Done

deniskuzZ (approved):
LGTM +1, pending tests

PR #5333, opened 2024-07-03
HIVE-28360: Upgrade jersey to 1.19.4

lvyankui:
Upgrade jersey to 1.19.4

[HIVE-28360](https://issues.apache.org/jira/browse/HIVE-28360)
After upgrading to Hadoop 3.3.5, the Hive WebHCat server fails to start because of inconsistent versions of the Jersey JAR package. Hive HCat lacks the jersey-server-1.19 jar.

None, that I am aware of.

CI

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -214,6 +214,14 @@
           <groupId>org.codehaus.jettison</groupId>
           <artifactId>jettison</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>com.sun.jersey</groupId>
+          <artifactId>jersey-json</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.sun.jersey</groupId>
+          <artifactId>jersey-client</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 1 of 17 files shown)


dengzhhu653 (approved):

devaspatikrishnatri:
[mvn_dependency_tree.txt](https://github.com/user-attachments/files/16466175/mvn_dependency_tree.txt)
Hi @lvyankui This is the dependency tree with this change and it would be a bit cleaner practice to exclude 1.19 versions which are shown in dependency tree as a part of org.apache.tez:tez-api:jar:0.10.3 and org.apache.atlas:atlas-client-common:jar:2.3.0.Otherwise this change looks good.

lvyankui:
> [mvn_dependency_tree.txt](https://github.com/user-attachments/files/16466175/mvn_dependency_tree.txt) Hi @lvyankui This is the dependency tree with this change and it would be a bit cleaner practice to exclude 1.19 versions which are shown in dependency tree as a part of org.apache.tez:tez-api:jar:0.10.3 and org.apache.atlas:atlas-client-common:jar:2.3.0.Otherwise this change looks good.

Thank you for your feedback. I have already taken care of it

dengzhhu653:
@lvyankui have you verified the hcatalog ok with the latest change? seems like the https://github.com/apache/hive/blob/master/common/pom.xml#L214 still has direct reference to the old jersey

lvyankui:
> @lvyankui have you verified the hcatalog ok with the latest change? seems like the https://github.com/apache/hive/blob/master/common/pom.xml#L214 still has direct reference to the old jersey

@dengzhhu653 I have processed all the references. Please help review again, thanks!

PR #5335, opened 2024-07-05
HIVE-28362: Fail to materialize a CTE with VOID

okumin:

Allow materialized CTEs to hold NULL literals.
https://issues.apache.org/jira/browse/HIVE-28362

Materialized CTEs should keep the original semantics of the CTEs. Regular CTEs allow a VOID type, then materialized CTEs also allow it.

No

No

Tested by the new qtest.


okumin:
Based on the discussion in JIRA, it is not a good idea to accept VOID because some data formats like Parquet or ORC don't support it.

PR #5341, opened 2024-07-11
HIVE-28367: Bump org.xerial.snappy:snappy-java from 1.1.10.4 to 1.1.10.5

tanishq-chugh:
Upgrade snappy-java from version 1.1.10.4 to 1.1.10.5
Have also made changes to make snappy-java version consistent throughout hive, which varied across modules earlier.

[https://github.com/xerial/snappy-java/releases/tag/v1.1.10.5](https://github.com/xerial/snappy-java/releases/tag/v1.1.10.5)

No

Yes
Dependency Tree: [dpn.txt](https://github.com/user-attachments/files/16175382/dpn.txt)

Manually tested by triggering basic queries in hive after making the change.

### DIFF
--- a/standalone-metastore/pom.xml
+++ b/standalone-metastore/pom.xml
@@ -95,6 +95,7 @@
     <protobuf.version>3.24.4</protobuf.version>
     <io.grpc.version>1.51.0</io.grpc.version>
     <sqlline.version>1.9.0</sqlline.version>
+    <snappy.version>1.1.10.5</snappy.version>
     <jline.version>2.14.6</jline.version>
     <ST4.version>4.0.4</ST4.version>
     <storage-api.version>4.1.0-SNAPSHOT</storage-api.version>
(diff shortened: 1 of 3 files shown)


standalone-metastore/pom.xml
```diff
@@ -365,6 +366,11 @@
         <artifactId>spring-core</artifactId>
         <version>${spring.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.xerial.snappy</groupId>
+        <artifactId>snappy-java</artifactId>
+        <version>${snappy.version}</version>
+      </dependency>
```
ayushtkn, line 373: Why are we defining here? We haven't excluded snappy in this pom either?

tanishq-chugh: snappy is coming in only transitively from hadoop-common inside sub-modules of standalone-metastore. 
Changing in the parent pom's dependency management does not effect on this inclusion. So have added in here.

PR #5345, opened 2024-07-12
HIVE-28369: Fix LLAP cache proactive eviction

ngsg:

This patch adds configuration/null checks to prevent NullPointerException.

As I explained the JIRA ticket(HIVE-28369), Hive throws NPE during proactive eviction when `hive.llap.io.encode.enabled` is false. 

No

No

Tested manually.
This cannot be tested using qtest because we cannot set up LLAP configuration in qfile as LLAP environment is initialized before test driver reads qfiles.

### DIFF
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/io/api/impl/LlapIoImpl.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/io/api/impl/LlapIoImpl.java
@@ -308,9 +308,14 @@ public class LlapIoImpl implements LlapIo<VectorizedRowBatch>, LlapIoDebugDump {
     LOG.debug("Starting proactive eviction.");
     long time = System.currentTimeMillis();
 
-    long markedBytes = dataCache.markBuffersForProactiveEviction(predicate, isInstantDeallocation);
-    markedBytes += fileMetadataCache.markBuffersForProactiveEviction(predicate, isInstantDeallocation);
-    markedBytes += serdeCache.markBuffersForProactiveEviction(predicate, isInstantDeallocation);
+    long markedBytes = 0;
+    if (useLowLevelCache) {
+      markedBytes += dataCache.markBuffersForProactiveEviction(predicate, isInstantDeallocation);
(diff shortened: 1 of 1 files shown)


kokila-19 (approved):

llap-server/src/java/org/apache/hadoop/hive/llap/io/api/impl/LlapIoImpl.java
```diff
@@ -308,9 +308,14 @@ public long evictEntity(LlapDaemonProtocolProtos.EvictEntityRequestProto protoRe
     LOG.debug("Starting proactive eviction.");
     long time = System.currentTimeMillis();
 
-    long markedBytes = dataCache.markBuffersForProactiveEviction(predicate, isInstantDeallocation);
-    markedBytes += fileMetadataCache.markBuffersForProactiveEviction(predicate, isInstantDeallocation);
-    markedBytes += serdeCache.markBuffersForProactiveEviction(predicate, isInstantDeallocation);
+    long markedBytes = 0;
+    if (useLowLevelCache) {
```
deniskuzZ, line 312: `SimpleBufferManager` is used for `dataCache` when `useLowLevelCache` is `false`, so we should evict it as well

deniskuzZ (changes requested):
`SimpleBufferManager` is used for `dataCache` when `useLowLevelCache` is `false`, so we should evict it as well

ngsg:
> `SimpleBufferManager` is used for `dataCache` when `useLowLevelCache` is `false`, so we should evict it as well

@deniskuzZ,  `SimpleBufferManager#markBuffersForProactiveEviction` always throws `UnsupportedOperationException`, so I excluded it when `useLowLevelCache == false`.

In fact, we don't have to check `dataCache` and `fileMetadataCache` because the first if-clause (`memoryManager == null`) ensures that `dataCache instanceof LowLevelCacheImpl` and `fileMetadataCache != null`. I didn't notice it and fixed my patch to check only `serdeCache`.

deniskuzZ (approved):
LGTM +1

PR #5351, opened 2024-07-17
HIVE-28374: Iceberg: Handle change of default format-version.

ayushtkn:

Handle the change of default version to 2 in the iceberg lib

To Avoid behavioural changes, when format version is explicitly defined as 2 vs it is being used as default

Yes, Sync behaviour of iceberg table, when format-version is explicitly defined as 2 vs it is being used as default

No

UT


simhadri-g (approved):
LGTM +1

PR #5354, opened 2024-07-17
HIVE-28376: Remove unused Hive object from RelOptHiveTable

zabetak:
The Hive object is not used inside RelOptHiveTable so keeping a reference to it is wasting memory and also complicates creation of RelOptHiveTable objects (constructor parameter).

Moreover, the Hive objects have thread local scope so in general they shouldn't be passed around cause their lifecycle becomes harder to manage.

No

No

Existing tests


simhadri-g (approved):
LGTM +1

kokila-19 (approved):

PR #5355, opened 2024-07-17
HIVE-28377: Add support for hive.output.file.extension to HCatStorer

VenkatSNarayanan:
HCatStorer now respects hive.file.output.extension for the output files it writes.

Brings HCatStorer's feature set more in line with Hive's.

Adds support for a property to HCatStorer that would previously have been ignored.

No

TestHCatExtension was added.

### DIFF
--- a/hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/FileOutputFormatContainer.java
+++ b/hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/FileOutputFormatContainer.java
@@ -97,8 +97,9 @@ class FileOutputFormatContainer extends OutputFormatContainer {
           (org.apache.hadoop.mapred.RecordWriter)null, context);
     } else {
       Path parentDir = new Path(context.getConfiguration().get("mapred.work.output.dir"));
+      String extension = HiveConf.getVar(context.getConfiguration(), HiveConf.ConfVars.OUTPUT_FILE_EXTENSION,"");
       Path childPath = new Path(parentDir,FileOutputFormat.getUniqueName(new JobConf(context.getConfiguration()),
-               context.getConfiguration().get("mapreduce.output.basename", "part")));
+               context.getConfiguration().get("mapreduce.output.basename", "part")) + extension);
 
       rw = new StaticPartitionFileRecordWriterContainer(
           getBaseOutputFormat().getRecordWriter(
--- a/hcatalog/core/src/test/java/org/apache/hive/hcatalog/mapreduce/HCatMapReduceTest.java
+++ b/hcatalog/core/src/test/java/org/apache/hive/hcatalog/mapreduce/HCatMapReduceTest.java
@@ -334,9 +334,6 @@ public abstract class HCatMapReduceTest extends HCatBaseTest {
     job.setOutputFormatClass(HCatOutputFormat.class);
 
     OutputJobInfo outputJobInfo = OutputJobInfo.create(dbName, tableName, partitionValues);
-    /*if (customDynamicPathPattern != null) {
-      job.getConfiguration().set(HCatConstants.HCAT_DYNAMIC_CUSTOM_PATTERN, customDynamicPathPattern);
-    }*/
     HCatOutputFormat.setOutput(job, outputJobInfo);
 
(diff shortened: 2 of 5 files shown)


yigress:
can the file extension be detected instead of user setting it explicitly? this may not be a good idea if a single job writes to different file formats. for example user set hive.output.file.extension to be '.parquet' but then same data is loaded into both parquet and orc tables.

hcatalog/core/src/test/java/org/apache/hive/hcatalog/mapreduce/TestHCatExtension.java
```diff
@@ -0,0 +1,111 @@
...
+    protected List<FieldSchema> getTableColumns() {
+        List<FieldSchema> fields = new ArrayList<FieldSchema>();
+        fields.add(new FieldSchema("c1", serdeConstants.INT_TYPE_NAME, ""));
+        fields.add(new FieldSchema("c2", serdeConstants.STRING_TYPE_NAME, ""));
+        return fields;
+    }
+
+    @Test
+    public void testHCatExtension() throws Exception {
+        HashMap<String, String> properties = new HashMap<String, String>();
```
difin, line 96: Can be simplified as `Map<String, String> properties = new HashMap<>();`

hcatalog/core/src/test/java/org/apache/hive/hcatalog/mapreduce/HCatMapReduceTest.java
```diff
@@ -282,7 +282,7 @@ public void map(WritableComparable key, HCatRecord value, Context context)
   Job runMRCreate(Map<String, String> partitionValues, List<HCatFieldSchema> partitionColumns,
       List<HCatRecord> records, int writeCount, boolean assertWrite) throws Exception {
     return runMRCreate(partitionValues, partitionColumns, records, writeCount, assertWrite,
-        true, null);
+        true, new HashMap<String, String>());
```
difin, line 285: Can `Collections.newMap()` be used instead?

VenkatSNarayanan: Changed.

hcatalog/core/src/test/java/org/apache/hive/hcatalog/mapreduce/TestHCatExtension.java
VenkatSNarayanan: Changed.

VenkatSNarayanan:
@yigress So for my implementation I just tried to match what Hive does in its existing implementation. If a sequential job wants to have different extensions for different tables, it can simply adjust the setting between HCat queries. Does Hive do something other than that to support that case that I missed?

yigress:
> @yigress So for my implementation I just tried to match what Hive does in its existing implementation. If a sequential job wants to have different extensions for different tables, it can simply adjust the setting between HCat queries. Does Hive do something other than that to support that case that I missed?

in hive the property is only used for text file format https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/Utilities.java#L916

+1 for this useful feature, if a single job only loads into one table and the table format is known user can switch the values for the property between jobs.  I am not expert in pig, somehow I have the impression that one pig job can load into multiple tables simultaneously, then user needs to be careful about setting

hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/FileOutputFormatContainer.java
```diff
@@ -97,8 +97,9 @@ public RecordWriter<WritableComparable<?>, HCatRecord> getRecordWriter(TaskAttem
           (org.apache.hadoop.mapred.RecordWriter)null, context);
     } else {
       Path parentDir = new Path(context.getConfiguration().get("mapred.work.output.dir"));
+      String extension = HiveConf.getVar(context.getConfiguration(), HiveConf.ConfVars.OUTPUT_FILE_EXTENSION,"");
```
deniskuzZ, line 100: nit, space

deniskuzZ (approved):
LGTM +1

PR #5358, opened 2024-07-18
HIVE-28042: DigestMD5 token expired or does not exist issue while opening connection to HMS

vikramahuja1001:
Adding two changes to fix this issue:
1. Rework expiry thread to not remove token after renewal time has passed for that particular token. It will actually try to renew the token in this case.
2. Individual calls to retrievePassword during the TSaslClientTransport auth will also try to renew the token if required before retrieving the password.

Facing DigestMD5 token expiry issue in a session which has been open since a long time when a new new connection is opened to HMS using TSaslClientTransport with DigestMD5 based auth. This issue is happening due to the fact that the new connection is trying to authenticate using the token identifier which is removed by the expiry thread in the background.

No

No

Added a test case to check the expiry thread renewing the token automatically after some time and removing a token automatically after the token has expired.
Tested the scenario on a machine with dedicated HMS, HS2 with Sasl enabled.
Added a test case to test the complete scenario of creating a new HMSClient after another.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/security/TokenStoreDelegationTokenSecretManager.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/security/TokenStoreDelegationTokenSecretManager.java
@@ -31,14 +31,16 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.Writable;
+import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.token.Token;
-import org.apache.hadoop.security.token.delegation.AbstractDelegationTokenIdentifier;
 import org.apache.hadoop.security.token.delegation.AbstractDelegationTokenSecretManager;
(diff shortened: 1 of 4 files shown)


chinnaraolalam:
LGTM +1

vikramahuja1001:
Thanks for the review

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/security/TokenStoreDelegationTokenSecretManager.java
```diff
@@ -236,22 +246,40 @@ public synchronized void stopThreads() {
...
+        }
+      }
+    }
+  }
+
+  private void renewIfRequired(long currentTime, DelegationTokenIdentifier id, DelegationTokenInformation tokenInfo)
+          throws InvalidToken {
+    if (tokenInfo != null) {
+      if (currentTime > tokenInfo.getRenewDate() && currentTime < id.getMaxDate()) {
+        // This will be the case when now > tokenInfo.getRenewDate() but less than the token expiration/max time.
```
pan3793, line 270: This seems to violate the design purpose of "renew date" - renewals are only allowed before this instant.

PR #5360, opened 2024-07-23
HIVE-25952: Drop HiveRelMdPredicates::getPredicates(Project) in favor of RelMdPredicates

zabetak:
1. Avoid potentially wrong results due to HIVE-26733
2. Facilitate code maintenance by removing duplication

Query plans may change in terms of performance and semantics but to the better.

No

Existing tests


zabetak:
Hey bot please keep it alive for a bit longer.

PR #5362, opened 2024-07-24
HIVE-28399: Improve the fetch size in HiveConnection

dengzhhu653:
If the 4.x Hive Jdbc client connects to an older HS2 or other thrift implementations, it might throw the IllegalStateException: https://github.com/apache/hive/blob/master/jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java#L1253-L1258, as the remote might haven't set the property:  
hive.server2.thrift.resultset.default.fetch.size back to the response of OpenSession request. It also introduces the confusing on what the real fetch size the connection is, as we have both initFetchSize and defaultFetchSize in this HiveConnection, the HiveStatement checks the initFetchSize, defaultFetchSize and 
HIVE_SERVER2_THRIFT_RESULTSET_DEFAULT_FETCH_SIZE.defaultIntVal to obtain the real fetch size, we can make them one in HiveConnection, so every statement created from the connection uses this new fetch size.

### DIFF
--- a/jdbc/src/java/org/apache/hive/jdbc/HiveStatement.java
+++ b/jdbc/src/java/org/apache/hive/jdbc/HiveStatement.java
@@ -20,7 +20,6 @@ package org.apache.hive.jdbc;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hive.common.classification.InterfaceAudience.LimitedPrivate;
-import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hive.jdbc.logs.InPlaceUpdateStream;
 import org.apache.hive.service.cli.RowSet;
 import org.apache.hive.service.cli.RowSetFactory;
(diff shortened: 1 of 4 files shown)


jdbc/src/java/org/apache/hive/jdbc/HiveStatement.java
```diff
@@ -130,22 +126,20 @@ public class HiveStatement implements java.sql.Statement {
 
   public HiveStatement(HiveConnection connection, TCLIService.Iface client,
       TSessionHandle sessHandle) {
-    this(connection, client, sessHandle, false, 0, DEFAULT_FETCH_SIZE);
+    this(connection, client, sessHandle, false, connection.fetchSize);
   }
 
   public HiveStatement(HiveConnection connection, TCLIService.Iface client, TSessionHandle sessHandle,
-      boolean isScrollableResultset, int initFetchSize, int defaultFetchSize) {
+      boolean isScrollableResultset, int fetchSize) {
```
zhangbutao, line 133: Does this change cause backward compatibility issues?

dengzhhu653: It shouldn't as I know, the original constructor is introduced since Hive 4.0, the HiveConnection and HiveStatement are in a single jar, so they are supposed to be loaded together by the same class loader, I'm not sure how the old `HiveConnection` can point to the new `HiveStatement` in the runtime.

zhangbutao (approved):
+1 LGTM

PR #5364, opened 2024-07-24
HIVE-28401: Drop redundant XML test report post-processing from CI pipeline

zabetak:
The Maven Surefire Plugin (3.0.0-M4) does not store system-out and system-err in the XML report so the xmlstarlet post-processing step is redundant and can be removed.

1. Avoid `Huge input lookup` error thrown by `xmlstarlet` when XML input is large:
a. allow subsequent post-processing actions to complete (save disk space)
b. avoid Jenkins pipeline build failures confusing developers
2. Simplify the pipeline script and minor perf improvement

No

No

Investigate the build artifacts generated by this PR and ensure that there is no system-out/system-err entries in XML files.


asolimando (approved):
LGTM, the `xmlstartlet` choking affected me many times in the past!

PR #5365, opened 2024-07-25
HIVE-28403: Delete redundant Javadoc for Hive

caicancai:
https://issues.apache.org/jira/browse/HIVE-28403


caicancai:
@zabetak @asolimando PATL，thanks

PR #5368, opened 2024-07-25
HIVE-28405: Changed the timeunit to DAYS

srbiswal:
Add a validator to interpret the time unit passed in **hive.repl.cm.retain** correctly in metastoreConf.

The value of "10" is different when it is set in HS2 and in Metastore. Adding a timevalidator allows us to set the values '10d' in both. 

No

No

NA


ayushtkn (changes requested):
this is an incompatible change, if someone has configured say 100, post this patch that 100 hrs will get changed to 100 days

srbiswal:
@ayushtkn updated the code to correctly interpret and handle values like 10d when set by users.

PR #5370, opened 2024-07-26
HIVE-28409: Column lineage when creating view is missing if atlas HiveHook is set

kasakrisz:

* Remove hardcoded Atlas HiveHook related filters from lineage information Generator and introduce an independent setting to filter based on statement/operation type.
* Add `isView` property to `QueryProperties` to indicate that this statement is a logical view creation.

Lineage information was not generated in case of create view operation. See jira for details: https://issues.apache.org/jira/browse/HIVE-28409

No.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=lineage6.q -pl itests/qtest -Pitests
mvn test -Dtest=TestGenerator.java -pl ql
```

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -3869,9 +3869,10 @@ public class HiveConf extends Configuration {
             "Deprecated: use hive.lineage.statement.filter instead."),
     HIVE_LINEAGE_STATEMENT_FILTER("hive.lineage.statement.filter", "ALL",
         "Whether Hive provides lineage information to hooks for the specified statements only, " +
-            "the value is a comma-separated list of HiveOperation (ex.: CREATEVIEW,CREATE_MATERIALIZED_VIEW," +
-            "CREATETABLE,CREATETABLE_AS_SELECT). There are two special values: ALL means lineage information is always " +
-            "provided, NONE means never."),
+            "the value is a comma-separated list (ex.: CREATE_MATERIALIZED_VIEW," +
+            "CREATE_TABLE,CREATE_TABLE_AS_SELECT). Possible values are: CREATE_TABLE, CREATE_TABLE_AS_SELECT, " +
+            "CREATE_VIEW, CREATE_MATERIALIZED_VIEW, LOAD, QUERY, ALL, NONE." +
+            " ALL means lineage information is always provided, NONE and empty string means never."),
 
     HIVE_SSL_PROTOCOL_BLACKLIST("hive.ssl.protocol.blacklist", "SSLv2,SSLv3",
         "SSL Versions to disable for all Hive Servers"),
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -3865,6 +3865,8 @@ public class HiveConf extends Configuration {
         "get old behavior, if desired. See, test-case in patch for HIVE-6689."),
     HIVE_LINEAGE_INFO("hive.lineage.hook.info.enabled", false,
         "Whether Hive provides lineage information to hooks."),
+    HIVE_LINEAGE_STATEMENT_FILTER("hive.lineage.statement.filter", "All",
+        ""),
 
     HIVE_SSL_PROTOCOL_BLACKLIST("hive.ssl.protocol.blacklist", "SSLv2,SSLv3",
         "SSL Versions to disable for all Hive Servers"),
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -3863,10 +3863,15 @@ public class HiveConf extends Configuration {
         "partition columns or non-partition columns while displaying columns in describe\n" +
         "table. From 0.12 onwards, they are displayed separately. This flag will let you\n" +
         "get old behavior, if desired. See, test-case in patch for HIVE-6689."),
+    @Deprecated
     HIVE_LINEAGE_INFO("hive.lineage.hook.info.enabled", false,
-        "Whether Hive provides lineage information to hooks."),
-    HIVE_LINEAGE_STATEMENT_FILTER("hive.lineage.statement.filter", "All",
-        ""),
+        "Whether Hive provides lineage information to hooks." +
+            "Deprecated: use hive.lineage.statement.filter instead."),
+    HIVE_LINEAGE_STATEMENT_FILTER("hive.lineage.statement.filter", "ALL",
+        "Whether Hive provides lineage information to hooks for the specified statements only, " +
+            "the value is a comma-separated list of HiveOperation (ex.: CREATEVIEW,CREATE_MATERIALIZED_VIEW," +
+            "CREATETABLE,CREATETABLE_AS_SELECT). There are two special values: ALL means lineage information is always " +
+            "provided, NONE means never."),
 
     HIVE_SSL_PROTOCOL_BLACKLIST("hive.ssl.protocol.blacklist", "SSLv2,SSLv3",
         "SSL Versions to disable for all Hive Servers"),
--- a/ql/src/java/org/apache/hadoop/hive/ql/hooks/LineageLogger.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/hooks/LineageLogger.java
@@ -6,7 +6,7 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *f
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
--- a/ql/src/java/org/apache/hadoop/hive/ql/hooks/LineageLogger.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/hooks/LineageLogger.java
@@ -6,7 +6,7 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
+ *f
  *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/Optimizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/Optimizer.java
@@ -79,7 +79,7 @@ public class Optimizer {
         || postExecHooks.contains("org.apache.hadoop.hive.ql.hooks.PostExecutePrinter")
         || postExecHooks.contains("org.apache.hadoop.hive.ql.hooks.LineageLogger")
         || postExecHooks.contains("org.apache.atlas.hive.hook.HiveHook")) {
-      transformations.add(new Generator(postExecHooks));
+      transformations.add(Generator.fromConf(hiveConf));
     }
 
     // Try to transform OR predicates in Filter into simpler IN clauses first
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SimpleFetchOptimizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SimpleFetchOptimizer.java
@@ -105,7 +105,8 @@ public class SimpleFetchOptimizer extends Transform {
   @Override
   public ParseContext transform(ParseContext pctx) throws SemanticException {
     Map<String, TableScanOperator> topOps = pctx.getTopOps();
-    if (pctx.getQueryProperties().isQuery() && !pctx.getQueryProperties().isAnalyzeCommand()
+    if ((pctx.getQueryProperties().isQuery() || pctx.getQueryProperties().isView())
+        && !pctx.getQueryProperties().isAnalyzeCommand()
         && topOps.size() == 1) {
       // no join, no groupby, no distinct, no lateral view, no subq,
       // no CTAS or insert, not analyze command, and single sourced.
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
@@ -67,14 +67,11 @@ import org.slf4j.LoggerFactory;
 public class Generator extends Transform {
 
   private static final Logger LOG = LoggerFactory.getLogger(Generator.class);
-
-  public static Generator fromConf(HiveConf conf) {
-    return new Generator(createFilterPredicateFromConf(conf));
-  }
-
   private static final String ALL = "ALL";
   private static final String NONE = "NONE";
   private static final Map<HiveOperation, Function<ParseContext, Boolean>> filterMap;
+
+  private final Predicate<ParseContext> statementFilter;
 
   static {
     Map<HiveOperation, Function<ParseContext, Boolean>> map = new EnumMap<>(HiveOperation.class);
@@ -87,6 +84,10 @@ public class Generator extends Transform {
         parseContext -> !(parseContext.getLoadTableWork() == null || parseContext.getLoadTableWork().isEmpty()));
     map.put(HiveOperation.QUERY, parseContext -> parseContext.getQueryProperties().isQuery());
     filterMap = Collections.unmodifiableMap(map);
+  }
+
+  public static Generator fromConf(HiveConf conf) {
+    return new Generator(createFilterPredicateFromConf(conf));
   }
 
   static Predicate<ParseContext> createFilterPredicateFromConf(Configuration conf) {
@@ -120,8 +121,6 @@ public class Generator extends Transform {
     return parseContext ->
         operations.stream().anyMatch(hiveOperation -> filterMap.get(hiveOperation).apply(parseContext));
   }
-
-  private final Predicate<ParseContext> statementFilter;
 
   public Generator(Predicate<ParseContext> statementFilter) {
     this.statementFilter = statementFilter;
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
@@ -19,10 +19,18 @@
 package org.apache.hadoop.hive.ql.optimizer.lineage;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
+import org.apache.commons.lang3.EnumUtils;
+import org.apache.hadoop.conf.Configuration;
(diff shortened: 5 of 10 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -3865,6 +3865,8 @@ public static enum ConfVars {
         "get old behavior, if desired. See, test-case in patch for HIVE-6689."),
     HIVE_LINEAGE_INFO("hive.lineage.hook.info.enabled", false,
         "Whether Hive provides lineage information to hooks."),
+    HIVE_LINEAGE_STATEMENT_FILTER("hive.lineage.statement.filter", "All",
```
InvisibleProgrammer, line 3868: I don't know if we have a rule about it or not. But can the default value be lowercase? 
I would love to avoid future confusion about if is that All, all or ALL. 

Also, could you please put some description and and example values there that can help us to understand what that config value exactly does?

ql/src/java/org/apache/hadoop/hive/ql/hooks/LineageLogger.java
```diff
@@ -6,7 +6,7 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
+ *f
```
InvisibleProgrammer, line 9: accident typo

ql/src/java/org/apache/hadoop/hive/ql/optimizer/Optimizer.java
```diff
@@ -79,7 +79,7 @@ public void initialize(HiveConf hiveConf) {
         || postExecHooks.contains("org.apache.hadoop.hive.ql.hooks.PostExecutePrinter")
         || postExecHooks.contains("org.apache.hadoop.hive.ql.hooks.LineageLogger")
         || postExecHooks.contains("org.apache.atlas.hive.hook.HiveHook")) {
-      transformations.add(new Generator(postExecHooks));
+      transformations.add(Generator.fromConf(hiveConf));
```
InvisibleProgrammer, line 82: I assume that before that change, we read the transformations from the postExecHooks that is as far I remember, calculated once. And from now, we read those again at every single query execution. 
Am I correct?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
```diff
@@ -59,11 +68,63 @@ public class Generator extends Transform {
 
   private static final Logger LOG = LoggerFactory.getLogger(Generator.class);
 
-  private final Set<String> hooks;
-  private static final String ATLAS_HOOK_CLASSNAME = "org.apache.atlas.hive.hook.HiveHook";
+  public static Generator fromConf(HiveConf conf) {
+    return new Generator(createFilterPredicateFromConf(conf));
+  }
+
+  private static final String ALL = "ALL";
```
InvisibleProgrammer, line 75: it relates to my previous comment at HiveConf about the possible values. At HiveConf, it is All, not ALL.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
```diff
@@ -59,11 +68,63 @@ public class Generator extends Transform {
...
+        return parseContext -> true;
+      }
+      if (NONE.equalsIgnoreCase(valueText)) {
+        noneSpecified = true;
+        continue;
+      }
+
+      HiveOperation enumValue = EnumUtils.getEnumIgnoreCase(HiveOperation.class, valueText);
+      if (enumValue == null) {
+        throw new EnumConstantNotPresentException(HiveOperation.class, valueText);
```
InvisibleProgrammer, line 106: I'm not familiar with that part: what will be the user experience if they add this value to the conf: "Hello" ?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
```diff
@@ -59,11 +68,63 @@ public class Generator extends Transform {
...
+      else {
+        return parseContext -> false;
+      }
+    }
+
+    return parseContext ->
+        operations.stream().anyMatch(hiveOperation -> filterMap.get(hiveOperation).apply(parseContext));
+  }
+
+  private final Predicate<ParseContext> statementFilter;
```
InvisibleProgrammer, line 124: Can you move that declaration and the constructor up so that the class can keep that pattern: fields -> constructors -> methods?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
```diff
@@ -72,18 +133,11 @@ public Generator(Set<String> hooks) {
...
-      if (!pctx.getQueryProperties().isCTAS()
-          && !pctx.getQueryProperties().isMaterializedView()
-          && pctx.getQueryProperties().isQuery()
-          && pctx.getCreateTable() == null
-          && pctx.getCreateViewDesc() == null
-          && (pctx.getLoadTableWork() == null || pctx.getLoadTableWork().isEmpty())) {
-        LOG.debug("Not evaluating lineage");
-        return pctx;
-      }
+    if (!statementFilter.test(pctx)) {
```
InvisibleProgrammer, line 136: I have a fear that this change will break existing configs: Previously, we said that if the Atlas HiveHook is set, we will run that transformation. Now we are saying that if `hive.lineage.statement.filter` is set, we do the transformation.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/Optimizer.java
kasakrisz: `postExecHooks` also calculated at every query execution:
https://github.com/apache/hive/blob/09553fca66ff69ff870c8a181750b70d81a8640e/ql/src/java/org/apache/hadoop/hive/ql/optimizer/Optimizer.java#L75-L77

I think this should be addressed in a follow-up only if this is bottleneck.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
kasakrisz: The comparison is not case sensitive to make this more felxible
```
if (ALL.equalsIgnoreCase(valueText)) {
```
```
if (NONE.equalsIgnoreCase(valueText)) {
```
```
HiveOperation enumValue = EnumUtils.getEnumIgnoreCase(HiveOperation.class, valueText);
```

I added an upper case default value because other values coming from HiveOperation are in upper case. There are other configs with upper case values so I think it is not an exception.
I also added the missing description.

ql/src/java/org/apache/hadoop/hive/ql/hooks/LineageLogger.java
kasakrisz: removed

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
kasakrisz: Exception is thrown at compile time.
Example:
```
java.lang.EnumConstantNotPresentException: org.apache.hadoop.hive.ql.plan.HiveOperation.Hello
	at org.apache.hadoop.hive.ql.optimizer.lineage.Generator.createFilterPredicateFromConf(Generator.java:106)
	at org.apache.hadoop.hive.ql.optimizer.lineage.Generator.fromConf(Generator.java:72)
	at org.apache.hadoop.hive.ql.optimizer.Optimizer.initialize(Optimizer.java:82)
	at org.apache.hadoop.hive.ql.parse.SemanticAnalyzer.analyzeInternal(SemanticAnalyzer.java:13274)
...
```

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
kasakrisz: Everything before `statementFilter` is static. Where should those static fields and methods go?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
kasakrisz: Yes, it breaks existing configs but previously Atlas HiveHook-related filters were hardcoded and impossible to test. Now `hive.lineage.statement.filter` is independent of hooks and hence applies for any hook making it possible to test using `LineageLogger`.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -3863,8 +3863,15 @@ public static enum ConfVars {
         "partition columns or non-partition columns while displaying columns in describe\n" +
         "table. From 0.12 onwards, they are displayed separately. This flag will let you\n" +
         "get old behavior, if desired. See, test-case in patch for HIVE-6689."),
+    @Deprecated
     HIVE_LINEAGE_INFO("hive.lineage.hook.info.enabled", false,
-        "Whether Hive provides lineage information to hooks."),
+        "Whether Hive provides lineage information to hooks." +
+            "Deprecated: use hive.lineage.statement.filter instead."),
+    HIVE_LINEAGE_STATEMENT_FILTER("hive.lineage.statement.filter", "ALL",
```
InvisibleProgrammer, line 3870: I would suggest using NONE as default.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/Optimizer.java
InvisibleProgrammer: thx

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
InvisibleProgrammer: thx

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
InvisibleProgrammer: I would really not add method declarations before field ones. 
What about static fields, instance fields, static initialization block, constructor, public methods, private methods?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
kasakrisz: NONE means that this functionality is never triggered. IMHO all functionality should be tested otherwise follow-up changes in the project may make it impossible to turn it on in the future. We have experienced this in the past a few times.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
kasakrisz: I usually move static members to the beginning of the file but I don't have string opinion about this.
I changed the order as you suggested.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SimpleFetchOptimizer.java
```diff
@@ -105,7 +105,8 @@ public class SimpleFetchOptimizer extends Transform {
   @Override
   public ParseContext transform(ParseContext pctx) throws SemanticException {
     Map<String, TableScanOperator> topOps = pctx.getTopOps();
-    if (pctx.getQueryProperties().isQuery() && !pctx.getQueryProperties().isAnalyzeCommand()
+    if ((pctx.getQueryProperties().isQuery() || pctx.getQueryProperties().isView())
```
deniskuzZ, line 108: does that cover create table?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
```diff
@@ -58,12 +67,63 @@
 public class Generator extends Transform {
 
   private static final Logger LOG = LoggerFactory.getLogger(Generator.class);
+  private static final String ALL = "ALL";
```
deniskuzZ, line 70: wouldn't it better to create LiniageFilter enum?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
```diff
@@ -58,12 +67,63 @@
...
+
+  static Predicate<ParseContext> createFilterPredicateFromConf(Configuration conf) {
+    Set<HiveOperation> operations = new HashSet<>();
+    boolean noneSpecified = false;
+    for (String valueText : conf.getStringCollection(HiveConf.ConfVars.HIVE_LINEAGE_STATEMENT_FILTER.varname)) {
+      if (ALL.equalsIgnoreCase(valueText)) {
+        return parseContext -> true;
+      }
+      if (NONE.equalsIgnoreCase(valueText)) {
+        noneSpecified = true;
```
deniskuzZ, line 101: why not just check size of valueText end exit?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
```diff
@@ -72,18 +132,11 @@ public Generator(Set<String> hooks) {
...
-          && pctx.getQueryProperties().isQuery()
-          && pctx.getCreateTable() == null
-          && pctx.getCreateViewDesc() == null
-          && (pctx.getLoadTableWork() == null || pctx.getLoadTableWork().isEmpty())) {
-        LOG.debug("Not evaluating lineage");
-        return pctx;
-      }
+    if (!statementFilter.test(pctx)) {
+      LOG.debug("Not evaluating lineage");
+      return pctx;
```
deniskuzZ, line 137: i think prev filter was much simpler

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
kasakrisz: In that case the values in `HiveOperation` would be listed again.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
deniskuzZ: ok, but instead of `NONE` option, can't we just check if it's Collections.emptySet()?

kasakrisz: If `Configuration.java.getStringCollection` returns empty set the loop does zero iterations, this `if` statement doesn't called and we end up with an always false lambda expression here:
```
return parseContext ->
        operations.stream().anyMatch(hiveOperation -> filterMap.get(hiveOperation).apply(parseContext));
```

deniskuzZ: 👍 do we need then NONE? can't we simply set `hive.lineage.statement.filter` to empty string?

kasakrisz: IMHO having a `NONE` setting is more user friendly but empty string is also allowed and means the same.
```
set hive.lineage.statement.filter=NONE;
```
vs
```
set hive.lineage.statement.filter=;
```

deniskuzZ: exactly, why do we need NONE then
take a look at other configs like `TASK_THREADS_REMOTE_ONLY`, we supply the complete list
'ALL' is not declarative as well

deniskuzZ: ````hive.lineage.statement.filter=CREATETABLE, CREATETABLE_AS_SELECT, CREATEVIEW, CREATE_MATERIALIZED_VIEW, LOAD, QUERY````
i think it's more user-friendly and gives an understanding of what are the filters if one wants to disable some

kasakrisz: TASK_THREADS_REMOTE_ONLY has 6 values vs HiveOperation has ~40-50. Listing them all is not user friendly that is why I introduced the `ALL` value. `NONE` is the opposite of `ALL`.

deniskuzZ: oh, sorry, I thought it was only above 6

deniskuzZ (approved):
LGTM +1

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
deniskuzZ:
````
if (!pctx.getQueryProperties().isCTAS()
          && !pctx.getQueryProperties().isMaterializedView()
          && pctx.getQueryProperties().isQuery()
          && pctx.getCreateTable() == null
          && pctx.getCreateViewDesc() == null
          && (pctx.getLoadTableWork() == null || pctx.getLoadTableWork().isEmpty())) {
        LOG.debug("Not evaluating lineage");
        return pctx;
      }
````
@kasakrisz, we didn't log all operations before, why should we now?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java:101]: filterMap contains just a few operations, what happens in operation is not in a filter list?
````
filterMap.get(hiveOperation).apply(parseContext));
````
could there be NPE?

Comment (kasakrisz) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java:100]: > we didn't log all HiveOperation before, why should we now

Only if `org.apache.atlas.hive.hook.HiveHook` present:
https://github.com/apache/hive/blob/b775c708f7dd4cd2088e68f994460b42a22189a2/ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java#L75
Which never happens in any of our tests.

> filterMap contains just a few operations, what happens in operation is not in a filter list?

Good point. I found that we can not check every HiveOperation based on the parse context so I introduced a new enum only with the supported operation types.

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java [hunk 13]
```diff
@@ -67,59 +63,59 @@
...
-    filterMap = Collections.unmodifiableMap(map);
+  enum LineageInfoFilter {
+    CREATE_TABLE(parseContext -> parseContext.getCreateTable() != null),
+    CREATE_TABLE_AS_SELECT(parseContext -> parseContext.getQueryProperties().isCTAS()),
+    CREATE_VIEW(parseContext -> parseContext.getQueryProperties().isView()),
+    CREATE_MATERIALIZED_VIEW(parseContext -> parseContext.getQueryProperties().isMaterializedView()),
+    LOAD(parseContext -> !(parseContext.getLoadTableWork() == null || parseContext.getLoadTableWork().isEmpty())),
+    QUERY(parseContext -> parseContext.getQueryProperties().isQuery()),
+    ALL(parseContext -> true),
+    NONE(parseContext -> false);
```
deniskuzZ, line 77: i still think that NONE is redundant, but I am to keep it

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
deniskuzZ: 👍

PR #5372, opened 2024-07-28
HIVE-28402: Precommit tests fail with OOM when running split-19

dengzhhu653:
The oom on split-19 is gone after applying the fix, while the output size of test-results.tgz almost remains the same.

### DIFF
--- a/.mvn/develocity.xml
+++ b/.mvn/develocity.xml
@@ -37,6 +37,9 @@
         <![CDATA[authenticated]]>
       </onlyIf>
     </publishing>
+    <capture>
+      <testLogging>false</testLogging>
+    </capture>
   </buildScan>
   <buildCache>
     <local>


kasakrisz:
@simhadri-g 
Could you please review this. it seems to be related to

simhadri-g (approved):
LGTM +1

zabetak (approved):
I reproduced locally and was checking the heap dump which is full of https://docs.gradle.com/enterprise/event-model-javadoc/com/gradle/scan/eventmodel/maven/MvnTestOutput_1_0.html events.

Please enrich the JIRA ticket with details about the root cause and solution and other than that it looks good to me. 

We could also consider disabling build logs since I don't think anyone is using those at the moment.

```
<buildLogging>false</buildLogging>
```

zabetak:
For the record, the Jenkins build artifact `test-results.tgz` do not seem related to this configuration at all.

dengzhhu653:
> Please enrich the JIRA ticket with details about the root cause and solution and other than that it looks good to me.

Sure. let me update it

zabetak:
We have the root cause and we have a green run so I would suggest to merge this ASAP to unblock precommits.

PR #5375, opened 2024-07-31
HIVE-28417: Bump Log4j2 to 2.24.1 to facilitate compilation of GraalVM Native Image

linghengqian:

- Bump Log4j2 to 2.24.1 to facilitate compilation of GraalVM Native Image. 

- This PR is aimed at ensuring that HiveServer2 JDBC Driver is compatible with future versions of LOG4J as much as possible. I created a unit test at https://github.com/linghengqian/log4j2-v2231-graalvm-native-image-test to prove that the HiveServer2 JDBC Driver using `apache/logging-log4j2:2.23.1` can be used under the GraalVM Native Image compiled by GraalVM CE For JDK 22.0.2.
```shell
sdk install java 8.0.422-tem
sdk install java 22.0.2-graalce
sdk use java 8.0.422-tem
sdk install maven
git clone git@github.com:linghengqian/hive.git -b log4j-bump
cd ./hive/
mvn clean install -DskipTests
mvn clean package -pl packaging -DskipTests -Pdocker
cd ../

git clone git@github.com:linghengqian/log4j2-v2231-graalvm-native-image-test.git
cd ./log4j2-v2231-graalvm-native-image-test/
sdk use java 22.0.2-graalce
./mvnw -PgenerateMetadata -DskipNativeTests -e -T1C clean test native:metadata-copy
./mvnw -PnativeTestInHive -T1C -e clean test
```

- See also https://github.com/apache/logging-log4j2/blob/rel/2.23.1/src/changelog/2.20.0/LOG4J2-3650_Move_tests.xml .

``` 
Moved `log4j-api` and `log4j-core` artifacts with classifier `tests` to `log4j-api-test` and `log4j-core-test` respectively. 
```

- See also https://github.com/apache/logging-log4j2/issues/1291#issuecomment-1450793036 .

```
I took care of the [release notes](https://logging.apache.org/log4j/2.x/release-notes/2.20.0.html).

The difference in dependencies is caused by Maven:

for the log4j-core artifact with tests classifier Maven pulled the wrong dependencies (compile scope instead of test scope),
for the log4j-core-test artifact the real dependencies are no longer hidden: the artifact mostly contains JUnit 4 and JUnit 5 extensions, so it depends on all those libraries.
```

- See also https://github.com/apache/logging-log4j2/pull/1032 . The constructor of `org.apache.logging.slf4j.Log4jMarker` has changed.
- See also https://github.com/apache/logging-log4j2/issues/2791 and https://github.com/apache/logging-log4j2/issues/2794 . The Log4j configuration file used by Hive contains deprecated properties .

No.

- Below is the output of `mvn dependency:tree > ./dependency-tree.txt`.

- [dependency-tree.txt](https://github.com/user-attachments/files/17350631/dependency-tree.txt)

```shell
sdk install java 8.0.422-tem
sdk use java 8.0.422-tem
sdk install maven
mvn clean install -DskipTests
mvn test -Dtest=TestHplsqlLocal -pl hplsql
mvn test -Dtest=TestHplsqlOffline -pl hplsql
```

### DIFF
--- a/data/conf/hive-log4j2.properties
+++ b/data/conf/hive-log4j2.properties
@@ -14,9 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-status = INFO
 name = HiveLog4j2Test
-packages = org.apache.hadoop.hive.ql.log
 
 # list of properties
 property.hive.log.level = DEBUG
@@ -24,9 +22,6 @@ property.hive.root.logger = DRFA
 property.hive.log.dir = ${sys:test.tmp.dir}/log
 property.hive.log.file = hive.log
 property.hive.test.console.log.level = INFO
-
-# list of all appenders
-appenders = console, DRFA
 
 # console appender
 appender.console.type = Console
@@ -48,9 +43,6 @@ appender.DRFA.policies.time.interval = 1
 appender.DRFA.policies.time.modulate = true
 appender.DRFA.strategy.type = DefaultRolloverStrategy
 appender.DRFA.strategy.max = 30
-
-# list of all loggers
-loggers = HadoopIPC, HadoopSecurity, Hdfs, HdfsServer, HadoopMetrics2, Mortbay, Yarn, YarnServer, Tez, HadoopConf, Zookeeper, ServerCnxn, NIOServerCnxn, ClientCnxn, ClientCnxnSocket, ClientCnxnSocketNIO, DataNucleus, Datastore, JPOX, Operator, Serde2Lazy, ObjectStore, CalcitePlanner, AmazonAws, ApacheHttp, Thrift, Jetty, BlockStateChange, swo, CBORuleLogger
 
 logger.HadoopIPC.name = org.apache.hadoop.ipc
 logger.HadoopIPC.level = WARN
@@ -128,7 +120,7 @@ logger.CBORuleLogger.filter.marker.type = MarkerFilter
 logger.CBORuleLogger.filter.marker.marker = FULL_PLAN
 # Change filter to ACCEPT, to see the produced plan after every rule invocation using the EXPLAIN CBO format
 logger.CBORuleLogger.filter.marker.onMatch = DENY
-logger.CBORuleLogger.filter.marker.onMisMatch = NEUTRAL
+logger.CBORuleLogger.filter.marker.onMismatch = NEUTRAL
 
 logger.AmazonAws.name=com.amazonaws
 logger.AmazonAws.level = INFO
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/QueryTracker.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/QueryTracker.java
@@ -48,8 +48,9 @@ import org.apache.tez.common.security.JobTokenIdentifier;
 import org.apache.tez.common.security.TokenCache;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.slf4j.Marker;
 import org.slf4j.MDC;
+import org.slf4j.Marker;
+import org.slf4j.impl.StaticMarkerBinder;
 
 import java.io.IOException;
 import java.util.Collections;
@@ -70,7 +71,7 @@ public class QueryTracker extends AbstractService {
 
   private static final Logger LOG = LoggerFactory.getLogger(QueryTracker.class);
   private static final Marker QUERY_COMPLETE_MARKER =
-      new Log4jMarker(new Log4jQueryCompleteMarker());
+      new Log4jMarker(StaticMarkerBinder.getSingleton().getMarkerFactory(), new Log4jQueryCompleteMarker());
 
   /// Shared singleton MetricsSource instance for all DAG locks
   private static final MetricsSource LOCK_METRICS;
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@
     <!-- Leaving libfb303 at 0.9.3 regardless of libthrift: As per THRIFT-4613 The Apache Thrift project does not publish items related to fb303 at this point -->
     <libfb303.version>0.9.3</libfb303.version>
     <libthrift.version>0.16.0</libthrift.version>
-    <log4j2.version>2.18.0</log4j2.version>
+    <log4j2.version>2.24.1</log4j2.version>
     <mariadb.version>2.5.0</mariadb.version>
     <mssql.version>6.2.1.jre8</mssql.version>
     <mysql.version>8.0.31</mysql.version>
@@ -1791,7 +1791,6 @@
             <log4j.configurationFile>${test.log4j.scheme}${test.conf.dir}/hive-log4j2.properties</log4j.configurationFile>
             <hive.test.console.log.level>${test.console.log.level}</hive.test.console.log.level>
             <hive.cluster.id>hive-test-cluster-id-cli</hive.cluster.id>
-            <log4j.debug>true</log4j.debug>
             <!-- don't dirty up /tmp -->
             <java.io.tmpdir>${test.tmp.dir}</java.io.tmpdir>
             <!-- Hadoop's minidfs class uses this -->
--- a/llap-server/pom.xml
+++ b/llap-server/pom.xml
@@ -367,9 +367,8 @@
     </dependency>
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-core</artifactId>
+      <artifactId>log4j-core-test</artifactId>
       <version>${log4j2.version}</version>
-      <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/llap-server/src/test/org/apache/hadoop/hive/llap/daemon/impl/TestLlapDaemonLogging.java
+++ b/llap-server/src/test/org/apache/hadoop/hive/llap/daemon/impl/TestLlapDaemonLogging.java
@@ -25,7 +25,7 @@ import org.apache.hadoop.security.Credentials;
 import org.apache.hadoop.security.token.Token;
 import org.apache.hive.testutils.junit.extensions.DoNothingTCPServer;
 import org.apache.hive.testutils.junit.extensions.DoNothingTCPServerExtension;
-import org.apache.logging.log4j.junit.LoggerContextSource;
+import org.apache.logging.log4j.core.test.junit.LoggerContextSource;
 import org.apache.tez.common.security.TokenCache;
 
 import org.junit.jupiter.api.Test;
--- a/standalone-metastore/metastore-server/pom.xml
+++ b/standalone-metastore/metastore-server/pom.xml
@@ -418,9 +418,13 @@
     </dependency>
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-core</artifactId>
+      <artifactId>log4j-core-test</artifactId>
       <version>${log4j2.version}</version>
-      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreServerUtils.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreServerUtils.java
@@ -44,7 +44,7 @@ import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.config.LoggerConfig;
-import org.apache.logging.log4j.test.appender.ListAppender;
+import org.apache.logging.log4j.core.test.appender.ListAppender;
 import org.apache.thrift.TException;
 import org.hamcrest.core.IsNot;
 import org.junit.After;
--- a/standalone-metastore/pom.xml
+++ b/standalone-metastore/pom.xml
@@ -89,7 +89,7 @@
     <junit.vintage.version>5.6.3</junit.vintage.version>
     <libfb303.version>0.9.3</libfb303.version>
     <libthrift.version>0.16.0</libthrift.version>
-    <log4j2.version>2.18.0</log4j2.version>
+    <log4j2.version>2.24.1</log4j2.version>
     <mockito-core.version>3.4.4</mockito-core.version>
     <orc.version>1.9.4</orc.version>
     <protobuf.version>3.24.4</protobuf.version>


linghengqian (commented):
- @pgaref I don't understand why the bot didn't update the github tag after I put the PR back for review. Do you know how to make the bot update the github tag again?
- update 1st: To be honest, I can't understand the error log of https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5375/5/tests . 
```shell
Unexpected exception java.lang.RuntimeException: Process docker failed to run in 30 seconds
 at org.apache.hadoop.hive.ql.externalDB.AbstractExternalDB.runCmd(AbstractExternalDB.java:91)
 at org.apache.hadoop.hive.ql.externalDB.AbstractExternalDB.runCmdAndPrintStreams(AbstractExternalDB.java:106)
```
- update 2nd: I took a moment to get to know the Jenkins UI... Looks like it should work, right?

Aggarwal-Raghav:
Had 1 question, added it as comment. Otherwise java and pom changes LGTM (+1 non-binding).

data/conf/hive-log4j2.properties
```diff
@@ -49,9 +44,6 @@ appender.DRFA.policies.time.modulate = true
 appender.DRFA.strategy.type = DefaultRolloverStrategy
 appender.DRFA.strategy.max = 30
 
-# list of all loggers
-loggers = HadoopIPC, HadoopSecurity, Hdfs, HdfsServer, HadoopMetrics2, Mortbay, Yarn, YarnServer, Tez, HadoopConf, Zookeeper, ServerCnxn, NIOServerCnxn, ClientCnxn, ClientCnxnSocket, ClientCnxnSocketNIO, DataNucleus, Datastore, JPOX, Operator, Serde2Lazy, ObjectStore, CalcitePlanner, AmazonAws, ApacheHttp, Thrift, Jetty, BlockStateChange, swo, CBORuleLogger
```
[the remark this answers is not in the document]
linghengqian: - Refer to https://github.com/apache/logging-log4j2/issues/2791#issuecomment-2265017589 and https://logging.apache.org/log4j/2.x/manual/configuration.html#Properties mentioned in my PR description. According to the log4j2 committer, defining a `logger` no longer requires listing a `logger` root node, and this `logger` root node will be removed in log4j2 version 3.0.

- Or we can keep the root `logger` node until log4j2 removes it in version 3.0.

> Prior to version 2.6, the properties configuration requires that you list the identifiers of the appenders, filters and loggers, in a comma separated list in properties with those names. Each of those components will then be expected to be defined in sets of properties that begin with component.<.identifier>.. The identifier does not have to match the name of the component being defined but must uniquely identify all the attributes and subcomponents that are part of the component. If the list of identifiers is not present the identifier must not contain a '.'. Each individual component MUST have a "type" attribute specified that identifies the component's Plugin type.
> As of version 2.6, this list of identifiers is no longer required as names are inferred upon first usage, however if you wish to use more complex identifies you must still use the list. If the list is present it will be used.

linghengqian (commented):
Due to the bugs https://issues.apache.org/jira/browse/HIVE-28559 and https://issues.apache.org/jira/browse/HIVE-28552, I can't actually merge the master branch directly.

pom.xml
```diff
@@ -175,7 +175,7 @@
     <!-- Leaving libfb303 at 0.9.3 regardless of libthrift: As per THRIFT-4613 The Apache Thrift project does not publish items related to fb303 at this point -->
     <libfb303.version>0.9.3</libfb303.version>
     <libthrift.version>0.16.0</libthrift.version>
-    <log4j2.version>2.18.0</log4j2.version>
+    <log4j2.version>2.23.1</log4j2.version>
```
zhangbutao, line 178: Can we use the lateset log4j2 version? 2.24.1
https://github.com/apache/logging-log4j2/releases/tag/rel%2F2.24.1

data/conf/hive-log4j2.properties
```diff
@@ -14,9 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-status = INFO
```
zhangbutao, line 17: Why this changed?

data/conf/hive-log4j2.properties
```diff
@@ -14,9 +14,7 @@
...
 name = HiveLog4j2Test
-packages = org.apache.hadoop.hive.ql.log
```
zhangbutao, line 19: Why this changed?

data/conf/hive-log4j2.properties
zhangbutao: How do you test this change is ok? Do we have unit tests to verify this?

linghengqian: - `status` is only used to set the output of log4j2's internal log. Its default value is `WARN`. Setting `status` to `INFO` actually masks the deprecated property warning of log4j2. When I looked at `mvn test -Dtest=TestHplsqlLocal`, I found that the assertion of the unit test is essentially matching the predefined output log file. In this case, there is actually no need to change the value of `status`. And starting from log4j2 2.24.0, the status attribute in the Configuration element has been deprecated, refer to https://logging.apache.org/log4j/2.x/manual/status-logger.html#config .
- The `log4j.debug` system property directly overrides changes to the `status` property.

linghengqian: > How do you test this change is ok? Do we have unit tests to verify this?

- More discussion is at https://github.com/apache/logging-log4j2/issues/2794 . More discussion is at https://github.com/apache/logging-log4j2/issues/2794 . There is even a spinoff PR https://github.com/apache/hive/pull/5381 to improve unit testing.

data/conf/hive-log4j2.properties
linghengqian: - According to https://github.com/apache/logging-log4j2/issues/2791, we no longer need such list definition, log4j2 will scan it automatically. Not surprisingly, retaining this attribute causes the warning.
```shell
2024-08-02T10:55:16.634Z main WARN The use of package scanning to locate plugins is deprecated and will be removed in a future release
2024-08-02T10:55:16.650Z main WARN The use of package scanning to locate plugins is deprecated and will be removed in a future release
2024-08-02T10:55:16.670Z main WARN The use of package scanning to locate plugins is deprecated and will be removed in a future release
2024-08-02T10:55:16.686Z main WARN The use of package scanning to locate plugins is deprecated and will be removed in a future release
```

pom.xml
linghengqian: - I have submitted new commits, just waiting for CI to complete first. 
- Of course, I will say that hive's PR template requires that PRs that update dependencies need to submit a separate `dependency-tree.txt` file, which is a pain. Because my IDE and shell are installed under WSLg, and I need to use Windows to search for files in the remote file system to upload `dependency-tree.txt`. If hive can integrate with ASF's https://ge.apache.org/, I think uploading a suspicious `dependency-tree.txt` will no longer be necessary. Because there are more straightforward interfaces like https://ge.apache.org/s/sztfzljasgv52/dependencies that output dependency trees.

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/QueryTracker.java
```diff
@@ -70,7 +71,7 @@ public class QueryTracker extends AbstractService {
 
   private static final Logger LOG = LoggerFactory.getLogger(QueryTracker.class);
   private static final Marker QUERY_COMPLETE_MARKER =
-      new Log4jMarker(new Log4jQueryCompleteMarker());
+      new Log4jMarker(StaticMarkerBinder.getSingleton().getMarkerFactory(), new Log4jQueryCompleteMarker());
```
zhangbutao, line 74: `StaticMarkerBinder.getSingleton().getMarkerFactory()` is the correct way to initialize the `Maker`?

pom.xml
```diff
@@ -1791,7 +1791,6 @@
             <log4j.configurationFile>${test.log4j.scheme}${test.conf.dir}/hive-log4j2.properties</log4j.configurationFile>
             <hive.test.console.log.level>${test.console.log.level}</hive.test.console.log.level>
             <hive.cluster.id>hive-test-cluster-id-cli</hive.cluster.id>
-            <log4j.debug>true</log4j.debug>
```
zhangbutao, line 1794: Why the change?

linghengqian: log4j.debug actually has no effect in previous versions of log4j2. Setting it to `true` will also make it impossible to override changes to log4j2's internal behavior through the command line. In log4j 2.24.0, which strengthens the role of log4j.debug, setting log4j.debug will cause the configuration of log4j2's internal log in the log4j2 properties file to become invalid. log4j.debug should usually only be called from the command line.

llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/QueryTracker.java
linghengqian: - This is how old versions behave. The corresponding function signature was changed 2 years ago at https://github.com/apache/logging-log4j2/pull/1032 .

pom.xml
zhangbutao: Did you see any issue with `log4j.debug` in your change?

how we use `log4j.debug` from command?  -Dlog4j.debug=true ?

linghengqian: > Did you see any issue with `log4j.debug` in your change?

- The full error stack is at https://github.com/apache/logging-log4j2/issues/2794 .

> how we use `log4j.debug` from command?  -Dlog4j.debug=true ?

- Yes.

data/conf/hive-log4j2.properties
zhangbutao: Can we replace the deprecated attribute  `status ` to `statusLoggerLevel`?

https://logging.apache.org/log4j/2.x/manual/status-logger.html#config
`Since version 2.24.0, status attribute in the Configuration element is deprecated and should be replaced with the log4j2.statusLoggerLevel configuration property.`

https://logging.apache.org/log4j/2.x/manual/status-logger.html#log4j2.statusLoggerLevel


It will be convinent for people to change the log status.

data/conf/hive-log4j2.properties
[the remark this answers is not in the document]
zhangbutao: I prefer not to change the code here for now. We can remove this the `loggers`  in log4j 3.0.

data/conf/hive-log4j2.properties
linghengqian: @zhangbutao So, do you want to put `<log4j2.statusLoggerLevel>INFO</log4j2.statusLoggerLevel>` in the top-level `pom.xml` to apply globally? Otherwise, a separate `log4j2.StatusLogger.properties` file is required.

data/conf/hive-log4j2.properties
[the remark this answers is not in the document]
linghengqian: > I prefer not to change the code here for now. We can remove this the `loggers`  in log4j 3.0.

I remember 2 months ago, keeping this property would cause additional warnings and unit tests to fail. What do you think about https://github.com/apache/hive/pull/5375#discussion_r1800890436 ?

data/conf/hive-log4j2.properties
```diff
@@ -25,9 +23,6 @@ property.hive.log.dir = ${sys:test.tmp.dir}/log
 property.hive.log.file = hive.log
 property.hive.test.console.log.level = INFO
 
-# list of all appenders
-appenders = console, DRFA
```
zhangbutao, line 29: Why change this? I didn't see any information from https://github.com/apache/logging-log4j2/issues/2791. Could you please give a explanation?

data/conf/hive-log4j2.properties
zhangbutao: https://github.com/apache/logging-log4j2/issues/2794#issuecomment-2267564572 

`<log4j.debug>true` line in `/pom.xml`  and the `status` should be changed at the same?

Can we only keep the `status`? (can replace with the new property)

data/conf/hive-log4j2.properties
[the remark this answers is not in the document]
zhangbutao: I have tested that deleting this line have nothing bad effect. So i am ok for the removal.

data/conf/hive-log4j2.properties
linghengqian: @zhangbutao 
- This involves a historical issue that has nothing to do with `apache/hive`. In the past two months, the log4j2 website documentation has been completely refactored, which has resulted in https://logging.apache.org/log4j/2.x/manual/configuration.html and https://logging.apache.org/log4j/2.x/manual/plugins.html no longer recording the deprecation of appenders and loggers. These deprecation information can be traced back to log4j2 2.6. Refer to https://github.com/apache/logging-log4j2/issues/2791#issuecomment-2265017589 pointed out by Log4j Committer.

> I noticed that Apache Hive uses a pre-2.6 properties configuration format (the `loggers` is not required since version 2.6).
Would you consider switching to a more easily readable configuration format? Version 3.x will introduce breaking changes to the properties format, but not the hierarchical ones.

- I have to resort to the `web.archive.org` to show the original document directly, please open https://web.archive.org/web/20240823014436/https://logging.apache.org/log4j/2.x/manual/configuration.html#configuration-with-properties directly.

> Prior to version **2.6**, the properties configuration requires that you list the identifiers of the **appenders**, **filters** and **loggers**, in a comma separated list in properties with those names. Each of those components will then be expected to be defined in sets of properties that begin with component.<.identifier>.. The identifier does not have to match the name of the component being defined but must uniquely identify all the attributes and subcomponents that are part of the component. If the list of identifiers is not present the identifier must not contain a '.'. Each individual component MUST have a "type" attribute specified that identifies the component's Plugin type.
>
> As of version 2.6, this list of identifiers is no longer required as names are inferred upon first usage, however if you wish to use more complex identifies you must still use the list. If the list is present it will be used.

- image

- `appenders`, `filters` and `loggers` do not actually have any effect, as log4j2 will infer them automatically.

data/conf/hive-log4j2.properties
zhangbutao: https://web.archive.org/web/20240823014436/https://logging.apache.org/log4j/2.x/manual/configuration.html#configuration-syntax

`Use of the packages attribute is deprecated and will be removed in Log4j 3.0. Plugins should be processed with the Log4j annotation processor. A comma separated list of package names to search for plugins. Plugins are only loaded once per classloader so changing this value may not have any effect upon reconfiguration.`

How we replace `packages `with the new way? I think removing `package `will cause some unkown issue.

data/conf/hive-log4j2.properties
zhangbutao: OK. i see. Thanks.

data/conf/hive-log4j2.properties
linghengqian: > How we replace `packages` with the new way? I think removing `package` will cause some unkown issue.

- I would say that even without merging this PR, `packages` never worked in Hive. Unless Hive has always used log4j1 and never used log4j2. Because the class `org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor` exists in the JAR of the final binary of Hive, `org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor` generates `Log4j2Plugins.dat`, which directly covers the operation of `packages`. It is located at `/hive/ql/target/hive-exec-4.1.0-SNAPSHOT.jar!/META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat`. `packages` is actually a `log4j1` thing.
- image
- You can determine from a comment by a log4j2 committer that this property does not actually take effect. See https://github.com/apache/logging-log4j2/issues/2791#issuecomment-2267168356 .

> There is no problem in Apache Hive: I checked that the appropriate annotation processor is configured in the `hive-exec` project and there is a `Log4j2Plugins.dat` file in the JAR. You can safely remove the packages configuration attribute.

- `org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor` is available since log4j2 2.0. The processor will generate the necessary `META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat` descriptor that replaces classpath scanning. See https://github.com/apache/logging-log4j2/issues/2434#issuecomment-2037064399 . There is also an intuitive PR at https://github.com/apache/jmeter/pull/6326 .

- Since Hive has neither CI for JDK23 and JDK24 nor CI for GraalVM Native Image, we actually don't need to register a `org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor` in `pom.xml` as https://logging.apache.org/log4j/2.x/manual/plugins.html#plugin-registry .

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-compiler-plugin</artifactId>
  <version>${maven-compiler-plugin.version}</version>
  <executions>
    <execution>
      <id>generate-log4j-plugin-descriptor</id>
      <goals>
        <goal>compile</goal>
      </goals>
      <phase>process-classes</phase>
      <configuration>
        <proc>only</proc>
        <annotationProcessorPaths>
          <path>
            <groupId>org.apache.logging.log4j</groupId>
            <artifactId>log4j-core</artifactId>
            <version>2.24.1</version>
          </path>
        </annotationProcessorPaths>
        <annotationProcessors>
          <processor>org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor</processor>
        </annotationProcessors>
      </configuration>
    </execution>
  </executions>
</plugin>
```
- You may ask why. The reason is simple, because for JDK8-JDK22, the default value of the JVM parameter `-proc` is `full`. The JVM parameter `-proc` did not become `only` until JDK23. `-proc` of `full` means that all processors are included by default. See https://github.com/apache/shardingsphere/pull/33025 and OpenJDK Committer’s Blog https://inside.java/2024/06/18/quality-heads-up/ . I also suspect that the ASF Jenkins node has an outdated JDK8 version and does not actually recognize the `-proc` JVM parameter, just like the Temurin JDK 8 on the Github Actions Runner also does not recognize the `-proc` JVM parameter.
- Are there unit tests for Hive Exec under GraalVM Native Image compiled by GraalVM CE For JDK22 and GraalVM CE For JDK23? Yes, they are at https://github.com/linghengqian/hive-server2-jdbc-driver and https://github.com/apache/shardingsphere . Due to the ongoing dispute of https://issues.apache.org/jira/browse/HIVE-28445, I have to maintain Hive's Thin JAR outside of Hive to provide it for unit testing of ShardingSphere under GraalVM Native Image.
- Does `packages = org.apache.hadoop.hive.ql.log` affect things like `logger.Operator.name = org.apache.hadoop.hive.ql.exec.Operator`? Apparently not. Most people use `packages` because they are actually using log4j1.

zhangbutao: Good explanation!

> Since Hive has neither CI for JDK23 and JDK24 nor CI for GraalVM Native Image, we actually don't need to register a org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor in pom.xml as https://logging.apache.org/log4j/2.x/manual/plugins.html#plugin-registry .

BTW, if Hive wants to use JDK23 or JDK24, should we need to register the `PluginProcessor `?

linghengqian: > BTW, if Hive wants to use JDK23 or JDK24, should we need to register the PluginProcessor ?

- The answer is yes. But I think most people will actually choose to set `-proc:full` for JDK `23` to make JDK `23` behave the same as JDK8. `-proc:full` is only available on `11.0.23`, `17.0.11`, JDK `21` - `24`. `-proc:full` is not available on any OpenJDK 8u, `-proc:full` is only available on Oracle's closed source JDK `8u411`.
- For JDK23, we can set `<maven.compiler.proc>full</maven.compiler.proc>` in a separate Maven Profile to make JDK23 behave the same as JDK 8. Refer to https://github.com/apache/shardingsphere/pull/33224 for the solution.
- I would say setting `<maven.compiler.proc>full</maven.compiler.proc>` is not useful for Hive, as I found that the master branch of Hive cannot be compiled with JDK22. This means that none of the unit tests will be affected.

zhangbutao (approved):
+1 LGTM

Thanks for the patch!

PR #5377, opened 2024-08-01
HIVE-28422: Iceberg: Added missing awaitility dependency in iceberg-catalog

zhangbutao:

[HIVE-28364](https://issues.apache.org/jira/browse/HIVE-28364) only added the org.awaitility dependency in iceberg root pom, but this dep is used by iceberg-catalog, so we need to add it in iceberg-catalog moudle. Otherwise, the Intellij idea will regard **Awaitility** as unrecognized code in **TestHiveTableConcurrency** which may make people confused, though we can build iceberg project successfully.

No

No

Run the existing test TestHiveTableConcurrency.

### DIFF
--- a/iceberg/iceberg-catalog/pom.xml
+++ b/iceberg/iceberg-catalog/pom.xml
@@ -72,6 +72,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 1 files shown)


zhangbutao:
cc @deniskuzZ

deniskuzZ:
@zhangbutao it's already there, now declared twice

zhangbutao:
> @zhangbutao it's already there, now declared twice https://github.com/apache/hive/blob/master/iceberg/iceberg-catalog/pom.xml#L100-L104

My bad.. I didn't reload my project after rebasing the lastest code.
Closing this ticket.

PR #5378, opened 2024-08-01
[WIP] HIVE-28426: mysql schema upgrade fails

pkumarsinha:


dengzhhu653:
@pkumarsinha it won't block the upgrade right? on my side it just shows some warn message and the upgrade continues on

standalone-metastore/metastore-server/src/main/sql/mysql/upgrade-3.2.0-to-4.0.0-alpha-1.mysql.sql
```diff
@@ -104,7 +104,7 @@ CREATE TABLE IF NOT EXISTS REPLICATION_METRICS (
   PRIMARY KEY(RM_SCHEDULED_EXECUTION_ID)
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
---Increase the size of RM_PROGRESS to accomodate the replication statistics
+-- Increase the size of RM_PROGRESS to accomodate the replication statistics
 ALTER TABLE REPLICATION_METRICS MODIFY RM_PROGRESS varchar(24000);
 
 ALTER TABLE REPLICATION_METRICS MODIFY RM_PROGRESS varchar(10000);
```
dengzhhu653, line 110: could we remove the redundant alter query? in the postgres upgrade script, seems we have the same problem

PR #5379, opened 2024-08-01
HIVE-28427: HivePreFilteringRule gets applied multiple times

soumyakanti3578:

1. Change the key of the Map in `HiveRulesRegistry` to `String` from `RelNode` to get consistent hash values
2. Add a test to ensure that HivePreFilteringRule is only applied once
3. Move `TestRuleBase.java` to use its mocked tables

With `RelNode` as the key of a Map, we were getting different hash values for similar nodes, which resulted in multiple application of the rule

No

No

Added unit test to ensure that the rule is applied only once
```
mvn test -Dtest=org.apache.hadoop.hive.ql.optimizer.calcite.TestCBORuleFiredOnlyOnce#testHivePreFilteringRuleFiredOnlyOnce
```


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveRulesRegistry.java
```diff
@@ -27,23 +27,35 @@
...
+    return this.registryVisited.get(rule).contains(getRelNodeIdentifier(node));
+  }
+
+  private String getRelNodeIdentifier(RelNode node) {
+    if (node instanceof Filter) {
+      String inputClass = ((Filter) node).getInput().getClass().getSimpleName();
+      String condition = ((Filter) node).getCondition().toString();
+      String rowType = node.getRowType().toString();
+      return "input=" + inputClass + " condition=" + condition + " rowtype=" + rowType;
+    }
```
kasakrisz, line 56: I found that Hive defines a `HiveRelNode` interface which is implemented by all Hive operators. How about moving the `getRelNodeIdentifier` method to this interface. The default implementation can return the digest and `HiveFilter` can override it. Besides that `registerVisited` can take `HiveRelNode`
```
registerVisited(RelOptRule rule, HiveRelNode operator)
```
so we can avoid `instanceof`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveRulesRegistry.java
```diff
@@ -27,23 +27,35 @@
...
-    return this.registryVisited.get(rule);
+  public boolean hasBeenVisitedBy(RelOptRule rule, RelNode node) {
+    return this.registryVisited.get(rule).contains(getRelNodeIdentifier(node));
+  }
+
+  private String getRelNodeIdentifier(RelNode node) {
+    if (node instanceof Filter) {
+      String inputClass = ((Filter) node).getInput().getClass().getSimpleName();
+      String condition = ((Filter) node).getCondition().toString();
+      String rowType = node.getRowType().toString();
```
zabetak, line 54: I think there is an edge case where this check can prevent some optimizations to happen. If you have the same `Filter` in two different places in the tree (e.g., CTEs) then from now on the `HivePreFilteringRule` will trigger only on one of those. I am not sure if this should be a show stopper for merging the PR. I am trying to think of a better way to handle this but don't have many ideas at the moment.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/TestRuleHelper.java
```diff
@@ -88,16 +83,16 @@ public static RelBuilder buildRelBuilder(AbstractRelOptPlanner planner,
...
         relBuilder.field(field), relBuilder.literal(value));
   }
 
-  static RexNode or(RelBuilder relBuilder, RexNode... args) {
+  public static RexNode or(RelBuilder relBuilder, RexNode... args) {
     return relBuilder.call(SqlStdOperatorTable.OR, args);
   }
 
-  static RexNode and(RelBuilder relBuilder, RexNode... args) {
+  public static RexNode and(RelBuilder relBuilder, RexNode... args) {
```
zabetak, line 95: The `RelBuilder` API has methods for doing this operations so we don't need to use these utilities.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/TestHiveAugmentMaterializationRule.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.hadoop.hive.common.ValidReaderWriteIdList;
 import org.apache.hadoop.hive.common.ValidTxnWriteIdList;
 import org.apache.hadoop.hive.common.ValidWriteIdList;
+import org.apache.hadoop.hive.ql.optimizer.calcite.TestRuleBase;
```
zabetak, line 25: Unrelated changes; do we need them?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveRulesRegistry.java
soumyakanti3578: Thanks! This is a good idea!

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/views/TestHiveAugmentMaterializationRule.java
soumyakanti3578: These were needed because I moved TestRuleBase to a different package. I will revert it and these changes too.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/TestRuleHelper.java
soumyakanti3578: Thanks, I will use the APIs from RelBuilder instead.

okumin:
Just curious. Do we have a real example(i.e., SQL) causing significant problems?

soumyakanti3578:
@okumin Yes, we saw the issue while trying to upgrade to Calcite 1.33. Here's @kasakrisz's investigation: https://github.com/apache/hive/pull/5196#discussion_r1660939490

@kasakrisz has already fixed the issue in Calcite (1.38): https://issues.apache.org/jira/browse/CALCITE-6453

This PR tries to fix the RelNode matching issue in the `onMatch` method of HivePreFilteringRule present in the current version. After we merge this PR the test failing test in the 1.33 PR should pass.

soumyakanti3578:
@kasakrisz @zabetak 
Please review this when you find some time.

PR #5380, opened 2024-08-02
HIVE-28428 Map hash aggregation performance degradation

ryukobayashi:

Fix performance regression in map-hash aggregation. The following log will be output in large volume, causing performance degradation.
https://github.com/apache/hive/blob/rel/release-4.0.0/ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java#L967
https://github.com/apache/hive/blob/rel/release-4.0.0/ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java#L978

And, currently the flush size is fixed, but performance can be improved by changing it depending on the data

Prevents performance degradation.

Yes, we can improve performance by changing the flash size depending on your data.

Performance was checked in a local environment.


ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java
```diff
@@ -964,7 +963,8 @@ private void flushHashTable(boolean complete) throws HiveException {
     }
 
     int oldSize = hashAggregations.size();
-    LOG.info("Hash Tbl flush: #hash table = {}", oldSize);
+    int flushSize = (int) (oldSize * hashAggrFlushPercent);
+    LOG.trace("Hash Tbl flush: #hash table = {}", oldSize);
```
deniskuzZ, line 967: should we log the flush size as well?

deniskuzZ:
@ryukobayashi is it possible to create some test for that? maybe something similar to `tpcds30tb` tests

ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java
ryukobayashi: OK, I will log it.

ryukobayashi:
@deniskuzZ Since I'm not familiar with `tpcds30tb`, I tried adding a test in the same way. Please let me know if there's anything wrong.

deniskuzZ:
> @deniskuzZ Since I'm not familiar with `tpcds30tb`, I tried adding a test in the same way. Please let me know if there's anything wrong.

yes, it was added in the right place. tpcds30tb doesn't have the 30tb data but it simulates it via stats. Does it cover you test scenario, is the output expected?

ryukobayashi:
@deniskuzZ Yes, I have created and checked a test scenario for this issue.

deniskuzZ:
> @deniskuzZ Yes, I have created and checked a test scenario for this issue.

how so, I don't see you override `hive.map.aggr.hash.fulush.size.percent` in a test, so it's using the default 0.1

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2012,6 +2012,9 @@ public static enum ConfVars {
         "Set to 1 to make sure hash aggregation is never turned off."),
     HIVE_MAP_AGGR_HASH_MIN_REDUCTION_LOWER_BOUND("hive.map.aggr.hash.min.reduction.lower.bound", (float) 0.4,
         "Lower bound of Hash aggregate reduction filter. See also: hive.map.aggr.hash.min.reduction"),
+    HIVE_MAP_AGGR_HASH_FLUSH_SIZE_PERCENT("hive.map.aggr.hash.fulush.size.percent", (float) 0.1,
```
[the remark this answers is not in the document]
ryukobayashi: fixed.

ryukobayashi:
@deniskuzZ Right, it does not directly affect the plan.

deniskuzZ (approved):
LGTM +1

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2012,6 +2012,9 @@ public static enum ConfVars {
         "Set to 1 to make sure hash aggregation is never turned off."),
     HIVE_MAP_AGGR_HASH_MIN_REDUCTION_LOWER_BOUND("hive.map.aggr.hash.min.reduction.lower.bound", (float) 0.4,
         "Lower bound of Hash aggregate reduction filter. See also: hive.map.aggr.hash.min.reduction"),
+    HIVE_MAP_AGGR_HASH_FLUSH_SIZE_PERCENT("hive.map.aggr.hash.flush.size.percent", (float) 0.1,
+        "Whether percentage of the map side group aggregation hash tables to flush.　\n" +
```
deniskuzZ, line 2016: @ryukobayashi, could you please update the description before the merge
`Percentage of entries to flush in the map-side group aggregation hash table`

deniskuzZ:
> @deniskuzZ Right, it does not directly affect the plan.

in that case, should we keep the added q test?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
ryukobayashi: Got it. will do.

ryukobayashi:
@deniskuzZ I think it is not necessary so I will remove it.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2012,6 +2012,9 @@ public static enum ConfVars {
         "Set to 1 to make sure hash aggregation is never turned off."),
     HIVE_MAP_AGGR_HASH_MIN_REDUCTION_LOWER_BOUND("hive.map.aggr.hash.min.reduction.lower.bound", (float) 0.4,
         "Lower bound of Hash aggregate reduction filter. See also: hive.map.aggr.hash.min.reduction"),
+    HIVE_MAP_AGGR_HASH_FLUSH_SIZE_PERCENT("hive.map.aggr.hash.flush.size.percent", (float) 0.1,
+        "Percentage of hash table entries to flush in map-side group aggregation.\n" +
+        "Flushes the specified percentage in map-side group aggregation hash table."),
```
deniskuzZ, line 2017: @ryukobayashi, please remove that, it just repeats the above statement

ryukobayashi: Oh OK, done.

PR #5384, opened 2024-08-05
HIVE-28431 Fix RexLiteral to ExprNode conversion if the literal is an empty string

ramesh0201:
An empty literal in the filter should still produce a valid RexNode.
Currently conversion from RexLiteral to ExprNode fail if the literal is an empty string. This causes the CBO to fail for such queries with this filter.

No

No

mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=filter_emptyliteral.q

### DIFF
--- a/ql/src/test/queries/clientpositive/filter_emptyliteral.q
+++ b/ql/src/test/queries/clientpositive/filter_emptyliteral.q
@@ -1,0 +1,4 @@
+create table employee(firstname varchar(255), lastname char(25), id int);
+
+explain cbo SELECT id FROM employee where firstname = '';
+explain cbo SELECT id FROM employee where lastname = '';
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ExprNodeConverter.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ExprNodeConverter.java
@@ -361,6 +361,9 @@ public class ExprNodeConverter extends RexVisitorImpl<ExprNodeDesc> {
             "char values must use NlsString for correctness");
         int precision = lType.getPrecision();
         HiveChar value = new HiveChar((String) literal.getValue3(), precision);
+        if (value.getCharacterLength() == 0) {
+          return new ExprNodeConstantDesc(TypeInfoFactory.stringTypeInfo, null);
+        }
         return new ExprNodeConstantDesc(new CharTypeInfo(precision), value);
       }
       case VARCHAR: {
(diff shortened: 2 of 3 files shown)


ql/src/test/queries/clientpositive/filter_emptyliteral.q
```diff
@@ -0,0 +1,3 @@
+create table employee(name varchar(255), id int);
+
+explain cbo SELECT id FROM employee where name = '';
```
soumyakanti3578, line 3: Please add another test for `char` for completeness. Otherwise, looks good to me!

ramesh0201: I have added this. Also I have removed the redundant piece of code for CHAR and VARCHAR sql type.

ramesh0201:
@soumyakanti3578 
Thanks for the initial review. I have tested the patch with empty literal filters on char and varchar column.
Empty literal filter on char column appends space on the empty literal before RexNode to ExprNode conversion. So by the time of conversion we are not dealing with a empty string.
In the case of empty literal filter on varchar column, the literal is identified as a char and then has cast to varchar on top of it. Like cast(''(char type) as varchar). So the switch case in the toExprNodeConstantDesc for the varchar will never receive a empty literal and hence the fix for the char case alone will suffice.

Please let me know your thoughts and also re-review the latest patch. Thanks!

soumyakanti3578:
@ramesh0201 From what I understand, we are never going into `case VARCHAR` when dealing with a literal. In that case, do we even need a case for varchar? Just trying to understand if it's dead code or if there are scenarios where we go into the varchar case for literals.

Other than this, the change and the tests look good to me :)

ramesh0201:
> @ramesh0201 From what I understand, we are never going into `case VARCHAR` when dealing with a literal. In that case, do we even need a case for varchar? Just trying to understand if it's dead code or if there are scenarios where we go into the varchar case for literals.
> 
> Other than this, the change and the tests look good to me :)

We still need that code, it is not dead code.
1. cast(''(identified as char) as varchar) -> converting '' goes to the char case of the switch, whereas {'' as varchar} goes to the varchar case when converting the input to the cast(but it goes to the null block on the if condition) https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ExprNodeConverter.java#L290
2. cast('test'(identified as char) as varchar) -> converting 'test' goes to the char case of the switch, whereas {'test' as varchar} goes to the varchar case when converting the input to the cast(this time it goes to the else block) https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ExprNodeConverter.java#L367

soumyakanti3578:
Thanks for clarifying @ramesh0201! I think this can be merged as all the tests are passing now.

ramesh0201:
Thank you! for the review @soumyakanti3578

PR #3472, opened 2022-07-22
HIVE-26423: Make commons-pool2 an explicit dependency (and upgrade to 2.11.1)

jfsii:
### What changes were proposed in this pull request?
Makes commons-pool2 an explicit dependency

### Why are the changes needed?
To prevent runtime issues and to make a dependency that we directly use explicit.

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
Through this PRs test framework and verification that the specified version of commons-pool2 is getting put in the packaging directory.

PR #5385, opened 2024-08-06
HIVE-28438: Remove commons-pool2 as an explicit dependency & upgrade commons-dbcp2 to 2.12.0

tanishq-chugh:
Upgrade commons-dbcp2 and commons-pool2 to 2.12.0

To fix multiple CVEs and make commons-pool2 version consistent across complete project.

No

Yes
New dependency tree: 
[dpn.txt](https://github.com/user-attachments/files/16570391/dpn.txt)

Manually triggered some CRUD operations after dependency upgrade. Rest relying on precommit tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,10 +126,9 @@
     <commons-configuration.version>1.10</commons-configuration.version>
     <commons-exec.version>1.1</commons-exec.version>
     <commons-io.version>2.12.0</commons-io.version>
-    <commons-pool2.version>2.11.1</commons-pool2.version>
     <commons-lang3.version>3.12.0</commons-lang3.version>
     <commons-math3.version>3.6.1</commons-math3.version>
-    <commons-dbcp2.version>2.9.0</commons-dbcp2.version>
+    <commons-dbcp2.version>2.12.0</commons-dbcp2.version>
     <commons-text.version>1.10.0</commons-text.version>
     <derby.version>10.14.2.0</derby.version>
     <dropwizard.version>3.1.0</dropwizard.version>
--- a/ql/pom.xml
+++ b/ql/pom.xml
@@ -91,11 +91,6 @@
       <artifactId>commons-math3</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-pool2</artifactId>
-      <version>${commons-pool2.version}</version>
-    </dependency>
-    <dependency>
       <groupId>org.apache.hive</groupId>
       <artifactId>hive-vector-code-gen</artifactId>
       <version>${project.version}</version>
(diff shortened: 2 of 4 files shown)


Aggarwal-Raghav:
LGTM +1

kokila-19 (approved):

pom.xml
```diff
@@ -126,10 +126,10 @@
     <commons-configuration.version>1.10</commons-configuration.version>
     <commons-exec.version>1.1</commons-exec.version>
     <commons-io.version>2.12.0</commons-io.version>
-    <commons-pool2.version>2.11.1</commons-pool2.version>
+    <commons-pool2.version>2.12.0</commons-pool2.version>
```
deniskuzZ, line 129: this is a transitive dependency of commons-dbcp2. do we really need to explicitly declare it?

tanishq-chugh: commons-pool2 was made to be declared explicitly as part of : [HIVE-26423](https://issues.apache.org/jira/browse/HIVE-26423) & [PR-3472](https://github.com/apache/hive/pull/3472)
where the reasoning was that hive uses commons-pool2 directly and to prevent any kind of runtime/packaging issues, we need to declare it explicitly.

deniskuzZ: if you are trying to `make commons-pool2 version consistent` why do you explicitly define the dependency version in ql module and instead of dependency-management in root pom?

tanishq-chugh: commons-pool2 was defined in ql module in the same PR where it was declared explicitly - [PR-3472](https://github.com/apache/hive/pull/3472) . But, i believe that would be because mentioning in dependency-management in root pom would only make sure that version of commons-pool2 coming in transitively is kept consistent and not explicitly make commons-pool2 as direct dependency of hive which was required then. 
But even i am not sure that why was it declared only in ql and not other modules such as standalone-metastore where commons-dbcp2 is being pulled in.

deniskuzZ: What I meant was to declare the version in dependency management, so you won't get different dependency versions in multiple Hive modules, same as it's done for `commons-dbcp2`

deniskuzZ: `to prevent any potential runtime issues` - how could it cause runtime issues, scope of transitive dependency is compile, not provided. I don't get what [HIVE-26423](https://issues.apache.org/jira/browse/HIVE-26423) is supposed to fix.

jfsii: Yeah, it is safe to remove as far as I can tell. It was a downstream issue in which Compaction was failing at runtime because "commons-pool2" was not getting packaged. I have no idea why I decided to put this upstream rather than just keep it downstream.

tanishq-chugh: Thank you @jfsii for the information. I'll make the changes to remove commons-pool2 as an explicit dependency & upgrade only commons-dbcp2 in the same PR and also update the JIRA to reflect the same.

tanishq-chugh:
Hi @deniskuzZ , attaching the new dependency tree after the latest changes for reference: 
[dpn.txt](https://github.com/user-attachments/files/16580309/dpn.txt)

deniskuzZ (approved):
LGTM +1, thanks @tanishq-chugh!

PR #5386, opened 2024-08-06
HIVE-28366: Iceberg: Concurrent Insert and IOW produce incorrect result

deniskuzZ:

Introduces the locking to serialize some operation execution

Fixes the data correctness in case of concurrent Insert and IOW

No

No

ITests

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -888,7 +888,6 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
                 LOG.info("Cleaning job for jobID: {}, table: {}", jobContext.getJobID(), output);
 
                 Table table = output.getKey();
-                FileSystem fileSystem = new Path(table.location()).getFileSystem(jobConf);
                 String jobLocation = generateJobLocation(table.location(), jobConf, jobContext.getJobID());
                 // list jobLocation to get number of forCommit files
                 // we do this because map/reduce num in jobConf is unreliable
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -526,6 +526,7 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
    * Creates and commits an Iceberg change with the provided data and delete files.
    * If there are no delete files then an Iceberg 'append' is created, otherwise Iceberg 'overwrite' is created.
    * @param table      The table we are changing
+   * @param snapshotId The snapshot id of the table to use for validation
    * @param startTime  The start time of the commit - used only for logging
    * @param results    The object containing the new files we would like to add to the table
    * @param filterExpr Filter expression for conflict detection filter
@@ -590,11 +591,11 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
    * Creates and commits an Iceberg compaction change with the provided data files.
    * Either full table or a selected partition contents is replaced with compacted files.
    *
-   * @param table             The table we are changing
-   * @param snapshotId        The snapshot id of the table to use for validation
-   * @param startTime         The start time of the commit - used only for logging
-   * @param results           The object containing the new files
-   * @param partitionPath     The path of the compacted partition
+   * @param table         The table we are changing
+   * @param snapshotId    The snapshot id of the table to use for validation
+   * @param startTime     The start time of the commit - used only for logging
+   * @param results       The object containing the new files
+   * @param partitionPath The path of the compacted partition
    */
   private void commitCompaction(Table table, Long snapshotId, long startTime, FilesForCommit results,
       String partitionPath) {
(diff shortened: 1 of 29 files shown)


ayushtkn (commented):
@deniskuzZ  I am not sure if I am catching this right, isn't the problem just that for IOW we used an older snapshotId, why are we going into locking and stuff,
Can we just not do some validations during the commit that there is no new snapshot, rather than locking and going sequential.

Any pointers how other engines like spark handle concurrency in such cases?

deniskuzZ:
> @deniskuzZ I am not sure if I am catching this right, isn't the problem just that for IOW we used an older snapshotId, why are we going into locking and stuff, Can we just not do some validations during the commit that there is no new snapshot, rather than locking and going sequential.
> 
> Any pointers how other engines like spark handle concurrency in such cases?

Impala doesn't handle this. there is no validation in Iceberg API to prevent lost-update in case of IOW and Insert

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestAcidOnTez.java
```diff
@@ -544,7 +545,7 @@ public void testCtasTezUnion() throws Exception {
     //check we have right delete delta files after minor compaction
     status = fs.listStatus(new Path(TEST_WAREHOUSE_DIR + "/" +
       (Table.ACIDNOBUCKET).toString().toLowerCase()), FileUtils.STAGING_DIR_PATH_FILTER);
-    String[] expectedDelDelta2 = { "delete_delta_0000002_0000002_0000", "delete_delta_0000003_0000003_0000", "delete_delta_0000001_0000003_v0000015"};
```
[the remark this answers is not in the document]
deniskuzZ: cause there was a bug and we opened redundant txn for `alter table compact` command

ql/src/java/org/apache/hadoop/hive/ql/Driver.java
```diff
@@ -184,7 +186,17 @@ private void runInternal(String command, boolean alreadyCompiled) throws Command
...
+      try {
+        context.getLoadTableOutputMap().forEach(
+          (ltd, we) -> {
+            HiveStorageHandler handler = we.getTable().getStorageHandler();
+            if (handler != null) {
+              handler.validateCurrentSnapshot(ltd.getTable());
+            }
+          });
+      } catch (ReCompileException ex) {
+        compileInternal(context.getCmd(), true);
```
ayushtkn, line 198: Question -> Like If my validateCurrentSnapshot() fails, we will be compiling again & under the hood fetching the snapshot again, does ``handler.validateCurrentSnapshot(ltd.getTable());`` gets called again?

Just thinking the snapshot gets outdated in the another attempt as well, what happens

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1443,6 +1449,25 @@ static void overlayTableProperties(Configuration configuration, TableDesc tableD
     map.remove("columns.comments");
   }
 
+  @Override
+  public void validateCurrentSnapshot(TableDesc tableDesc) {
+    if (conf.getBoolean(ConfigProperties.LOCK_HIVE_ENABLED, TableProperties.HIVE_LOCK_ENABLED_DEFAULT) ||
+        !HiveConf.getBoolVar(conf, ConfVars.HIVE_TXN_EXT_LOCKING_ENABLED)) {
+      return;
+    }
+    Table table = IcebergTableUtil.getTable(conf, tableDesc.getProperties());
```
ayushtkn, line 1458: This would be fetching the table from cache, by default `skipCache` is `false`, can that lead to getting an outdated snapshot id in any case?

ayushtkn (commented):
minor doubts, rest looks good

ql/src/java/org/apache/hadoop/hive/ql/Driver.java
deniskuzZ: at this step we already acquired the locks, so we just need to update the snapshot

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: we compare the state of the table in a cache (populated at compile time) with the latest one
````
((BaseTable) table).operations().refresh() 
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java [hunk 4]
```diff
@@ -601,46 +602,52 @@ private void commitCompaction(Table table, Long snapshotId, long startTime, File
...
+   * if no data files are provided.
+   * For partitioned tables the relevant partitions are replaced with the new data files. Table remains unchanged
+   * unless data files are provided.
    *
-   * @param table                   The table we are changing
-   * @param startTime               The start time of the commit - used only for logging
-   * @param results                 The object containing the new files
+   * @param table         The table we are changing
+   * @param startTime     The start time of the commit - used only for logging
+   * @param results       The object containing the new files
```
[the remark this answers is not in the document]
deniskuzZ: added

ql/src/java/org/apache/hadoop/hive/ql/Driver.java
deniskuzZ: refactored the logic in the last commit

ayushtkn (approved):
LGTM

PR #5387, opened 2024-08-07
HIVE-28439: Iceberg: Bucket partition transform with DECIMAL can throw NPE

okumin:

Fix a bug in the iceberg_bucket UDF.

A bucket partition transform with DECIMAL doesn't correctly work when the source column type has extreme configurations. Simply, `GenericUDFIcebergBucket` always uses the default precision and scale regardless of the precise source type.

No

No

Update a qtest which fails in the current master


deniskuzZ (approved):
LGTM +1

okumin:
Thanks!

PR #5390, opened 2024-08-07
HIVE-28440: unblock hcatalog parquet project pushdown

yigress:
to unblock hcatloader parquet table project pushdown. HCatLoader sets the needed columns intending to be used for project/filter pushdown, however, org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport#init ReadContext skips project/filter pushdown when job properties 'columns' is not set. This change sets the 'columns' property.

to unblock hcatloader parquet table project pushdown

No

No

tested locally


daijyc:
LGTM

PR #5391, opened 2024-08-08
HIVE-28441: NPE in ORC tables when hive.orc.splits.include.file.footer is enabled

Aggarwal-Raghav:
Check [HIVE-28441](https://issues.apache.org/jira/browse/HIVE-28441) for steps to reproduce this issue and stacktrace

NullPointerException is thrown when hive.orc.splits.include.file.footer is enabled in ORC tables

NO

NO

Using a q file present in the commits.
```
mvn clean test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=orc_footer_enabled.q -pl itests/qtest -Pitests -Dtest.output.overwrite=true
mvn clean test -Dtest=TestMiniTezCliDriver -Dqfile=orc_footer_enabled.q -pl itests/qtest -Pitests -Dtest.output.overwrite=true
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcNewSplit.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcNewSplit.java
@@ -27,6 +27,7 @@ import org.apache.hadoop.hive.ql.io.AcidInputFormat;
 import org.apache.hadoop.io.WritableUtils;
 import org.apache.hadoop.mapreduce.lib.input.FileSplit;
 import org.apache.orc.OrcProto;
+import org.apache.orc.impl.BufferChunk;
 import org.apache.orc.impl.OrcTail;
 
 /**
@@ -101,7 +102,7 @@ public class OrcNewSplit extends FileSplit {
       byte[] tailBuffer = new byte[tailLen];
       in.readFully(tailBuffer);
       OrcProto.FileTail fileTail = OrcProto.FileTail.parseFrom(tailBuffer);
-      orcTail = new OrcTail(fileTail, null);
+      orcTail = new OrcTail(fileTail, new BufferChunk(0, 0), -1);
     }
   }
 
--- a/itests/src/test/resources/testconfiguration.properties
+++ b/itests/src/test/resources/testconfiguration.properties
@@ -11,6 +11,7 @@ minitez.query.files.shared=\
   empty_skip_header_footer_aggr.q,\
   hybridgrace_hashjoin_1.q,\
   hybridgrace_hashjoin_2.q,\
+  orc_footer_enabled.q,\
   substr_estimate_range.q
 
 # NOTE: Add tests to minitez only if it is very
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
@@ -546,13 +546,6 @@ public class OrcEncodedDataReader extends CallableWithNdc<Void>
     LlapIoImpl.ORC_LOGGER.trace("Creating reader for {} ({})", path, split.getPath());
     long startTime = counters.startTimeCounter();
     ReaderOptions opts = EncodedOrcFile.readerOptions(jobConf).filesystem(fsSupplier).fileMetadata(fileMetadata);
-    if (split instanceof OrcSplit) {
-      OrcTail orcTail = ((OrcSplit) split).getOrcTail();
-      if (orcTail != null) {
-        LlapIoImpl.ORC_LOGGER.debug("Setting OrcTail. path={}", path);
-        opts.orcTail(orcTail);
-      }
-    }
     orcReader = EncodedOrcFile.createReader(path, opts);
     counters.incrWallClockCounter(LlapIOCounters.HDFS_TIME_NS, startTime);
   }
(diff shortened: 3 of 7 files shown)


Aggarwal-Raghav:
As per my understanding:

1. One of the benefit of enabling _hive.orc.splits.include.file.foote_r is to reduce fs calls as explained in HIVE-15038. In ORC code, _extractFileTail_  https://github.com/apache/orc/blob/7878691befc66ecc372ff41715cbdff97ec7aafd/java/core/src/java/org/apache/orc/impl/ReaderImpl.java#L569 make a fs call for creating OrcTail but with the config enabled, it was optimized and we were creating OrcTail object in OrcSplit.java https://github.com/apache/hive/blob/d0d5d6d7d11b3eece0d0bc17b429cb30dec5dc79/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcSplit.java#L230

3. In HIVE-15665 with hive.orc.splits.include.file.footer enabled, it requires the OrcTail to have serializedTail present (passing null or empty BufferChunk won't help as it will throw NPE) https://github.com/apache/hive/blob/d0d5d6d7d11b3eece0d0bc17b429cb30dec5dc79/llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java#L669
	
4. Possible fix is while creating OrcTail in OrcSplit.java, we "**_somehow_**" get the serializedTail without making additional fs call or we need to revert HIVE-15038, doing so will force the orcReader in OrcEncodedDataReader.java to get perform _extractFileTail_ which will have the serializedTail.
	
5. I have gone with reverting the HIVE-15038. Looking forward for suggestions on this.

Aggarwal-Raghav:
@pgaref, can you please provide your insights on this?

deniskuzZ:
@Aggarwal-Raghav, is there still some benefit of `hive.orc.splits.include.file.footer` without [HIVE-15038](https://issues.apache.org/jira/browse/HIVE-15038)?

Aggarwal-Raghav:
@deniskuzZ, thanks for looking into this.
I think in Tez on Yarn, we can still prevent an additional fs call with this config.

ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcNewSplit.java
```diff
@@ -101,7 +102,7 @@ public void readFields(DataInput in) throws IOException {
       byte[] tailBuffer = new byte[tailLen];
       in.readFully(tailBuffer);
       OrcProto.FileTail fileTail = OrcProto.FileTail.parseFrom(tailBuffer);
-      orcTail = new OrcTail(fileTail, null);
+      orcTail = new OrcTail(fileTail, new BufferChunk(0, 0), -1);
```
zhangbutao, line 105: I want to know that if this issue is a regression which caused by orc version upgrading?
The related orc code snippet was first introduced by ORC 1.6.0 &  [ORC-520](https://issues.apache.org/jira/browse/ORC-520).
So, can i assume that we wont' meet this issue  before ORC 1.6.0?

If this issuce was caused by the newer orc version(>1.6.0), should we give a fix for orc to resolve the issue at its root?

Aggarwal-Raghav: @zhangbutao, In my opinion it is not because of orc version upgrade.
In tez on yarn flow, this issue is surfaced after orc version upgrade. 
in tez on llap, i think this issue is since [HIVE-15665](https://issues.apache.org/jira/browse/HIVE-15665  as https://github.com/apache/hive/blob/d0d5d6d7d11b3eece0d0bc17b429cb30dec5dc79/llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java#L669 requires serialized footer which with _`hive.orc.splits.include.file.footer enabled`_ is null (earlier) and (empty buffer) in this PR, both won't help. 

This code requires actual serialized buffer which I think can only be obtained by _`extractFileTail`_ function call. because if we pass empty buffer also, when I debugged, it is checking for last byte in the buffer that represents postscript length and as in empty buffer case it is 0, so something related to **_malformed ORC error is thrown_**

zhangbutao:
@Aggarwal-Raghav I did some codes debug. Found that the Tez Application Master has already initialize the `OrcTail`  when creating orc splits. I want to know if we can pass the `OrcTail`  from Tez AM to Tez Task? If ok, we can solve this issue by the way.
Here i provided some related codes, maybe we can try to do some code debug to explore a better way to fix the issue?

- Tez AM related codes: **create orc split with the orctail**

https://github.com/apache/hive/blob/13dfae1c0a7d4540f4bc5edc50bc922f0cfc83e8/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java#L1672-L1673

https://github.com/apache/hive/blob/13dfae1c0a7d4540f4bc5edc50bc922f0cfc83e8/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java#L1497-L1498


- Tez Task related code: **get orc split**(But it can not get the orctail now, we can think about how to get it here?)

https://github.com/apache/hive/blob/13dfae1c0a7d4540f4bc5edc50bc922f0cfc83e8/ql/src/java/org/apache/hadoop/hive/ql/io/HiveInputFormat.java#L216-L223

https://github.com/apache/hive/blob/13dfae1c0a7d4540f4bc5edc50bc922f0cfc83e8/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcSplit.java#L204-L205

PR #5397, opened 2024-08-19
HIVE-28452: Iceberg: Cache delete files on executors

deniskuzZ:

Cache delete files on executors to save on multiple reads

Performance improvement

No

No

Itest + regression test on a cluster


zhangbutao (approved):
+1 LGTM
btw, i would suggest to add more context about this change. For example, the change mainly benefits LLAP daemon.

deniskuzZ:
> +1 LGTM btw, i would suggest to add more context about this change. For example, the change mainly benefits LLAP daemon.

thanks for checking, @zhangbutao. In a container mode (no LLAP) Tez-object-cache is used, see ObjectCacheFactory#getCache

PR #5398, opened 2024-08-20
HIVE-28456: ObjectStore updatePartitionColumnStatisticsInBatch can ca…

dengzhhu653:
…use connection starvation

Make the time consuming API request the connection from primary connection pool

Since [HIVE-26419](https://issues.apache.org/jira/browse/HIVE-26419), we have a secondary connection pool for schema generation, and for value generation operations, the size of this pool is 2. However, based on DataNucleus documentation on datanucleus.ConnectionFactory2, link:
https://www.datanucleus.org/products/accessplatform_5_0/jdo/persistence.html
the secondary pool also serves for nontransactional connections(effectively "auto-commit" mode), which makes the ObjectStore updatePartitionColumnStatisticsInBatch request the connection from this pool, as it doesn't open a transaction explicitly. If there is a slow on inserting or updating the column statistics, the pool will become unavailable quickly(the pool reaches its maximum size), the ObjectStore cloud see the "Connection is not available, request timed out" under such a situation.

No

No

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
@@ -51,6 +51,7 @@ import org.slf4j.LoggerFactory;
 import org.apache.hadoop.conf.Configuration;
 
 import javax.jdo.PersistenceManager;
+import javax.jdo.Transaction;
 import javax.jdo.datastore.JDOConnection;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -99,14 +100,6 @@ class DirectSqlUpdatePart {
     this.dbType = dbType;
     this.maxBatchSize = batchSize;
     sqlGenerator = new SQLGenerator(dbType, conf);
-  }
-
-  void rollbackDBConn(Connection dbConn) {
-    try {
-      if (dbConn != null && !dbConn.isClosed()) dbConn.rollback();
-    } catch (SQLException e) {
-      LOG.warn("Failed to rollback db connection ", e);
-    }
   }
 
   void closeDbConn(JDOConnection jdoConn) {
@@ -484,14 +477,14 @@ class DirectSqlUpdatePart {
                                                       String validWriteIds, long writeId,
                                                       List<TransactionalMetaStoreEventListener> transactionalListeners)
           throws MetaException {
+
     JDOConnection jdoConn = null;
-    Connection dbConn = null;
-    boolean committed = false;
+    Transaction tx = pm.currentTransaction();
     try {
       dbType.lockInternal();
+      tx.begin();
       jdoConn = pm.getDataStoreConnection();
-      dbConn = (Connection) (jdoConn.getNativeConnection());
-
+      Connection dbConn = (Connection) jdoConn.getNativeConnection();
       setAnsiQuotes(dbConn);
 
       Map<PartitionInfo, ColumnStatistics> partitionInfoMap = getPartitionInfo(dbConn, tbl.getId(), partColStatsMap);
@@ -526,18 +519,18 @@ class DirectSqlUpdatePart {
         MetaStoreListenerNotifier.notifyEventWithDirectSql(transactionalListeners,
                 EventMessage.EventType.UPDATE_PARTITION_COLUMN_STAT_BATCH, eventBatch, dbConn, sqlGenerator);
       }
-      dbConn.commit();
-      committed = true;
+      closeDbConn(jdoConn);
+      tx.commit();
       return result;
     } catch (Exception e) {
       LOG.error("Unable to update Column stats for  " + tbl.getTableName(), e);
       throw new MetaException("Unable to update Column stats for  " + tbl.getTableName()
               + " due to: "  + e.getMessage());
     } finally {
-      if (!committed) {
-        rollbackDBConn(dbConn);
+      if (tx.isActive()) {
+        closeDbConn(jdoConn);
+        tx.rollback();
       }
-      closeDbConn(jdoConn);
       dbType.unlockInternal();
     }
   }
@@ -548,14 +541,13 @@ class DirectSqlUpdatePart {
    */
   public long getNextCSIdForMPartitionColumnStatistics(long numStats) throws MetaException {
     long maxCsId = 0;
-    boolean committed = false;
-    Connection dbConn = null;
     JDOConnection jdoConn = null;
-
+    Transaction tx = pm.currentTransaction();
     try {
       dbType.lockInternal();
+      tx.begin();
       jdoConn = pm.getDataStoreConnection();
-      dbConn = (Connection) (jdoConn.getNativeConnection());
+      Connection dbConn = (Connection) jdoConn.getNativeConnection();
 
       setAnsiQuotes(dbConn);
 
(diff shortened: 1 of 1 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -526,18 +519,18 @@ public Map<String, Map<String, String>> updatePartitionColumnStatistics(Map<Stri
         MetaStoreListenerNotifier.notifyEventWithDirectSql(transactionalListeners,
                 EventMessage.EventType.UPDATE_PARTITION_COLUMN_STAT_BATCH, eventBatch, dbConn, sqlGenerator);
       }
-      dbConn.commit();
-      committed = true;
+      closeDbConn(jdoConn);
```
dengzhhu653, line 522: this won't close the underlying connection because the current active is active, see 
https://github.com/datanucleus/datanucleus-core/blob/efba3bd0e8345e1171d5fe529cae693293f178ae/src/main/java/org/datanucleus/store/AbstractStoreManager.java#L378-L392
It tells we have reached the end of `transaction` and is able to commit now, until the `tx.commit()` the underlying connection will return back to the connection pool.

dengzhhu653: If the `tx.commit()` happens before the `closeDbConn(jdoConn)`, a JDOUserException cloud see when commit the change: 
`The Connection was acquired by the developer and must be closed before using the persistence API`.

https://www.datanucleus.org/products/accessplatform_6_0/jdo/persistence.html#_user_connection
> You must return the connection back to the PersistenceManager before performing any PersistenceManager operation. You do this by calling conn.close(). If you don’t return the connection and try to perform an PersistenceManager operation which requires the connection then a JDOUserException is thrown.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -602,18 +594,18 @@ public long getNextCSIdForMPartitionColumnStatistics(long numStats) throws MetaE
...
       return maxCsId;
     } catch (Exception e) {
       LOG.error("Unable to getNextCSIdForMPartitionColumnStatistics", e);
       throw new MetaException("Unable to getNextCSIdForMPartitionColumnStatistics  "
               + " due to: " + e.getMessage());
     } finally {
-      if (!committed) {
-        rollbackDBConn(dbConn);
+      if (tx.isActive()) {
+        closeDbConn(jdoConn);
```
deniskuzZ, line 606: should it be
````
closeDbConn(jdoConn);
if (tx.isActive()) {
  tx.rollback();
}
````

Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java:606]: We have a `closeDbConn(jdoConn)` before tx commits, if the `jdoConn` is being closed twice, a NucleusUserException is thrown from https://github.com/datanucleus/datanucleus-core/blob/master/src/main/java/org/datanucleus/store/NucleusConnectionImpl.java#L54-L62, then we log the exception at`closeDbConn(jdoConn);`, it could be misleading however.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java:606]: can we check in `closeDbConn` if jdoConn is closed and if not attempt to close it?

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java:606]: ````!dbConn.isClosed()````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java [hunk 3]
```diff
@@ -484,14 +477,14 @@ public Map<String, Map<String, String>> updatePartitionColumnStatistics(Map<Stri
                                                       String validWriteIds, long writeId,
                                                       List<TransactionalMetaStoreEventListener> transactionalListeners)
           throws MetaException {
+
     JDOConnection jdoConn = null;
-    Connection dbConn = null;
-    boolean committed = false;
+    Transaction tx = pm.currentTransaction();
```
deniskuzZ, line 482: why do we need txn here?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
deniskuzZ: why not wrap query execution in try-finally and close db connection there?
````
      try (Statement statement = dbConn.createStatement()) {
        statement.executeUpdate(query);
      } finally {
        closeDbConn(jdoConn);
      }
      tx.commit();
````

deniskuzZ:
````
 } finally {
            if (tx.isActive()) {
                tx.rollback();
            }
````

dengzhhu653: If the exception throws from https://github.com/apache/hive/blob/c6965423da67515c4fc8ba70a1d10ba15624789d/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java#L563-L586 , we still need to close the `jdoConn` at the outer finally,

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
dengzhhu653: It's used to mark we have opened a transaction explicitly, when `jdoConn = pm.getDataStoreConnection()`, it will get the connection from the primary connection pool, https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java#L307

https://github.com/datanucleus/datanucleus-core/blob/master/src/main/java/org/datanucleus/store/AbstractStoreManager.java#L373

https://github.com/datanucleus/datanucleus-core/blob/master/src/main/java/org/datanucleus/store/connection/ConnectionManager.java#L92

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
deniskuzZ: ok, back to the prev approach, can't we check if jdoConn is not already closed?
````
closeDbConn(jdoConn); 
if (tx.isActive()) { 
    tx.rollback(); 
}
...
void closeDbConn(JDOConnection jdoConn) {
    try {
      if (dbConn != null && !dbConn.isClosed()) {
        jdoConn.close();
      }
````

dengzhhu653: sounds the `JDOConnection` doesn't provide the method `isClosed()` to check the state...

deniskuzZ: what do you mean? check the rollback method that was removed

deniskuzZ: how about a nested try, since it doesn't implement auto-closeble
````
try {
    jdoConn = pm.getDataStoreConnection();
....
} finally {
    closeDbConn(jdoConn);
}
tx.commit();
````

dengzhhu653: ack, thank you for the suggestion!

deniskuzZ (approved):
LGTM +1

PR #5400, opened 2024-08-21
[WIP] Iceberg: Do not use HMS stats when statsSource is Iceberg

deniskuzZ:

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -118,7 +118,6 @@ import org.apache.hadoop.hive.ql.security.authorization.HiveAuthorizationProvide
 import org.apache.hadoop.hive.ql.security.authorization.HiveCustomStorageHandlerUtils;
 import org.apache.hadoop.hive.ql.session.SessionState;
 import org.apache.hadoop.hive.ql.session.SessionStateUtil;
-import org.apache.hadoop.hive.ql.stats.Partish;
 import org.apache.hadoop.hive.serde2.AbstractSerDe;
 import org.apache.hadoop.hive.serde2.DefaultFetchFormatter;
 import org.apache.hadoop.hive.serde2.Deserializer;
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/StatsOptimizer.java
```diff
@@ -945,7 +945,10 @@ private Long getRowCnt(
         if (!StatsUtils.areBasicStatsUptoDateForQueryAnswering(tbl, tbl.getParameters())) {
           return null;
         }
-        rowCnt = Long.valueOf(tbl.getProperty(StatsSetupConst.ROW_COUNT));
+        Map<String, String> basicStats = MetaStoreUtils.isNonNativeTable(tbl.getTTable()) ?
+          tbl.getStorageHandler().getBasicStatistics(tbl) : tbl.getParameters();
```
zhangbutao, line 949: Discuss:
Shoule we regard the stats is always accurate when statsSource is iceberg?
 If so, we need always to keep the configuration `iceberg.hive.keep.stats` true when statsSource is iceberg, so that we can optimization the count(*) when statsSource is iceberg by **StatsOptimizer**.  Same idea i wanted to do is https://github.com/apache/hive/pull/5215
https://github.com/apache/hive/blob/4f7200dc14acf71e189db05376baf4280742ff94/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L174

https://github.com/apache/hive/blob/4f7200dc14acf71e189db05376baf4280742ff94/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L225-L227

https://github.com/apache/hive/blob/4f7200dc14acf71e189db05376baf4280742ff94/ql/src/java/org/apache/hadoop/hive/ql/optimizer/StatsOptimizer.java#L945

deniskuzZ: Oh sorry, I missed the https://github.com/apache/hive/pull/5215 
I am not very certain how accurate is the `TOTAL_RECORDS_PROP` from the snapshot summary especially when there are deletes. 
Since we have a statsSource flag I just wanted to be consistent where we take a stats.

deniskuzZ: yes, i think `iceberg.hive.keep.stats` should be enabled when stats source is not iceberg

deniskuzZ: @zhangbutao, could you please check the comments in https://github.com/apache/hive/pull/5215 and maybe incorporate changes from this PR into yours

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -613,7 +616,7 @@ public boolean canComputeQueryUsingStats(org.apache.hadoop.hive.ql.metadata.Tabl
         }
       }
     }
-    return false;
+    return true;
```
zhangbutao, line 619: In case of delete files, if Iceberg statsSource can not compute query( eg. count(*)) using stats, i think HMS stats can't either. They both come from the same place -- `SnapshotSummary`.

We should not give a wrong impression that HMS can give the accurate stats.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -600,7 +600,10 @@ private ColumnStatistics readColStats(Table table, Path statsPath) {
 
   @Override
   public boolean canComputeQueryUsingStats(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
-    if (getStatsSource().equals(HiveMetaHook.ICEBERG) && hmsTable.getMetaTable() == null) {
+    if (hmsTable.getMetaTable() != null) {
```
zhangbutao, line 603: Query against Iceberg Branch/Tag can also benefit from the stats. We can optimize this later.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/StatsOptimizer.java
zhangbutao: > I am not very certain how accurate is the TOTAL_RECORDS_PROP from the snapshot summary especially when there are deletes.

Iceberg table with equal delets should not be optimized by the `count `query optimization, so we can skip to get the stats or return null in case of existing deletes.

> yes, i think iceberg.hive.keep.stats should be enabled when stats source is not iceberg

`iceberg.hive.keep.stats` should be always enabled(true) when stats source is iceberg.  `iceberg.hive.keep.stats` true means that the stats is accurate. 
HMS stats for iceberg can be not accurate if some other engines(Spark、Trino) write the table but not update the HMS stats.
But if the statsSource is iceberg, the stats is retrieved from iceberg `SnapshotSummary` which is real time and accurate.

> could you please check the comments in #5215 and maybe incorporate changes from this PR into yours

#5215 I want to optimize the `count `query with no care the values of  `iceberg.hive.keep.stats`. But i am not do the limit like your change that `not use HMS stats when statsSource is Iceberg`.  I am ok to incorporate this thange as well as some other supplements to #5215, we can continue to disscuss there.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: even if we do alter table compute stats?

zhangbutao: Good catch! 
In case of delete files, `analyze table compute stats`  job can get the accurate stats as it will launch tez task to compute the stats.

And after the job `analyze table compute stats`, the HMS stats will be updated & accurate and  `iceberg.hive.keep.stats` will be true, so we can use the HMS stats to optimize the `count `query. 

But if the statsSource is Iceberg & in case of delete files, even we have done the job `analyze table compute stats`, we won't update the Iceberg `SnapshotSummary`, so we can not optimize the `count `query.

This will look a little weird. Users do a job `analyze table compute stats` to update the stats, but they can not optimize the `count `query if the statsSource is Iceberg & in case of delete files.

PR #5401, opened 2024-08-21
HIVE-27080: Support project pushdown in JDBC storage handler even whe…

soumyakanti3578:
…n filters are not pushed

Support RelFieldTrimmer for JDBC nodes, so that trimmed projects can be pushed below HiveJdbcConverter

Currently we don't push projects when filters are not pushed. More details: https://issues.apache.org/jira/browse/HIVE-27080

No, except changes in plans.

No

```
mvn test -pl itests/qtest -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile="jdbc_project_pushdown.q"
```


ql/src/test/results/clientpositive/llap/jdbc_table_with_schema_mariadb.q.out
```diff
@@ -33,7 +33,8 @@ POSTHOOK: Input: default@country_0
 CBO PLAN:
 HiveJdbcConverter(convention=[JDBC.MYSQL])
   JdbcAggregate(group=[{}], agg#0=[count()])
-    JdbcHiveTableScan(table=[[default, country_0]], table:alias=[country_0])
+    JdbcProject(DUMMY=[0])
```
kasakrisz, line 36: What is the purpose of this `JdbcProject`? Is it necessary?

soumyakanti3578: Since RelFieldTrimmer can now bypass the HiveJdbcConverter barrier, it will always add a `JdbcProject` on top of `JdbcHiveTableScan`.
This happens because of [this](https://github.com/apache/hive/blob/bb0509f916d2edfad104e8af95ea3e304f920e6e/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/RelFieldTrimmer.java#L1066).

Although not necessary, I don't think an extra Project on top of TableScan here is a bad thing, especially if the table is wide!

kasakrisz: it seems that it doesn't affect the physical plan. At least in the case of this query
```
POSTHOOK: query: EXPLAIN SELECT COUNT(*) FROM country_0
POSTHOOK: type: QUERY
POSTHOOK: Input: default@country_0
STAGE DEPENDENCIES:
  Stage-0 is a root stage

STAGE PLANS:
  Stage: Stage-0
    Fetch Operator
      limit: -1
      Processor Tree:
        TableScan
          alias: country_0
          properties:
            hive.sql.query SELECT COUNT(*) AS `$f0`
FROM `country`
            hive.sql.query.fieldNames $f0
            hive.sql.query.fieldTypes bigint
            hive.sql.query.split false
          Select Operator
            expressions: $f0 (type: bigint)
            outputColumnNames: _col0
            ListSink
```

I guess this is the query which is pushed via JDBC
```
SELECT COUNT(*) AS `$f0`
FROM `country`
```

soumyakanti3578: I noticed that there is a `HiveAggregateProjectMergeRule` which merges this kind of agrregates and projects. I tested by excluding the rule and observed similar behaviour (dummy projects) below aggregates for non jdbc plans.

I have added a new rule [JdbcAggregateProjectMergeRule](https://github.com/apache/hive/pull/5401/files#diff-fec68de4add4936ac409e9e56362fe0444af9406d6d21fa4ae530d18d2d633e9) that resolves the issue here.

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -2296,6 +2297,7 @@ private RelNode applyPostJoinOrderingTransform(RelNode basePlan, RelMetadataProv
         rules.add(JDBCFilterJoinRule.INSTANCE);
         rules.add(JDBCFilterPushDownRule.INSTANCE);
         rules.add(JDBCProjectPushDownRule.INSTANCE);
+        rules.add(JdbcAggregateProjectMergeRule.INSTANCE);
```
kasakrisz, line 2300: Should it be called `JBDCAggregateProjectMergeRule` to follow the convention?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/JdbcAggregateProjectMergeRule.java
```diff
@@ -0,0 +1,59 @@
...
+    final Aggregate aggregate = call.rel(0);
+    // Rule cannot be applied if there are GroupingId because it will change the
+    // value as the position will be changed.
+    for (AggregateCall aggCall : aggregate.getAggCallList()) {
+      if (aggCall.getAggregation().equals(HiveGroupingID.INSTANCE)) {
+        return false;
+      }
+    }
+    return super.matches(call);
+  }
```
kasakrisz, line 58: This method is the same as the one in `HiveAggregateProjectMergeRule`.
IIUC the only difference between `HiveAggregateProjectMergeRule` and `JdbcAggregateProjectMergeRule` is the plan pattern definition.
How about adding a constructor to `HiveAggregateProjectMergeRule` to get the plan pattern as parameters
```
protected HiveAggregateProjectMergeRule(Class<? extends Aggregate> aggregateClass, Class<? extends Project> projectClass, RelBuilderFactory relBuilderFactory) {
  super(...);
}

JdbcAggregateProjectMergeRule extends HiveAggregateProjectMergeRule {
...
}

```

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
soumyakanti3578: done!

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/JdbcAggregateProjectMergeRule.java
soumyakanti3578: done!

kasakrisz (approved):

PR #5403, opened 2024-08-22
HIVE-28264: OOM/slow compilation when query contains SELECT clauses with nested expressions

kasakrisz:
Apply changes in [CALCITE-6513](https://issues.apache.org/jira/browse/CALCITE-6513) into `HiveFilterProjectTransposeRule`.

Applying the rule to nested project operators that contain nested expressions may cause OOM. Please see jira for details
[CALCITE-6513](https://issues.apache.org/jira/browse/CALCITE-6513)
[HIVE-28264](https://issues.apache.org/jira/browse/HIVE-28264)

No.

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestMiniLlapLocalCliDriver -Dqfile=subquery_nested_expressions.q -pl itests/qtest -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveFilterProjectTransposeRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveFilterProjectTransposeRule.java
@@ -36,6 +36,7 @@ import org.apache.calcite.rel.core.Filter;
 import org.apache.calcite.rel.core.Join;
 import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.rules.FilterProjectTransposeRule;
+import org.apache.calcite.rel.rules.ProjectMergeRule;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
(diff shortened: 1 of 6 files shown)


okumin (approved):
LGTM

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveFilterProjectTransposeRule.java
```diff
@@ -66,43 +67,47 @@ public class HiveFilterProjectTransposeRule extends FilterProjectTransposeRule {
...
   public static final HiveFilterProjectTransposeRule INSTANCE =
           new HiveFilterProjectTransposeRule(Filter.class, HiveProject.class,
-                  HiveRelFactories.HIVE_BUILDER, false, false);
+                  HiveRelFactories.HIVE_BUILDER, false, false, ProjectMergeRule.DEFAULT_BLOAT);
 
   private final boolean onlyDeterministic;
 
   private final boolean pushThroughWindowing;
 
+  private final int bloat;
```
zabetak, line 93: Do we need to modify `HiveFilterProjectTSTransposeRule` as well so that it takes bloat into account? Do we need to add more tests for covering this path?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveFilterProjectTransposeRule.java
```diff
@@ -66,43 +67,47 @@ public class HiveFilterProjectTransposeRule extends FilterProjectTransposeRule {
   public static final HiveFilterProjectTransposeRule DETERMINISTIC_WINDOWING_ON_NON_FILTERING_JOIN =
       new HiveFilterProjectTransposeRule(
           operand(Filter.class, operand(Project.class, operand(Join.class, any()))),
-          HiveRelFactories.HIVE_BUILDER, true, true);
+          HiveRelFactories.HIVE_BUILDER, true, true, ProjectMergeRule.DEFAULT_BLOAT);
```
zabetak, line 70: We could make the bloat parameter configurable via HiveConf but this is orthogonal to this change so we can do it in a follow-up if necessary.

zabetak (commented):
LGTM. Let's clarify what needs to be done for `HiveFilterProjectTSTransposeRule` and other than that its good to go.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveFilterProjectTransposeRule.java
kasakrisz: I replaced all occurrences of  `RelOptUtil.pushPastProject` to `HiveRelOptUtils.pushPastProjectUnlessBloat`
Based on the patches added the `RelOptUtil.pushPastProject` lots of test are affected. Let's see the next PTest results

zabetak (approved):

PR #5406, opened 2024-08-24
HIVE-28480: Disable SMB on partition hash generator mismatch across join branches…

himanshu-mishra:
… in previous RS

As SMB replaces last RS op from the joining branches and the JOIN op with MERGEJOIN, we need to ensure the RS before these RS, in both branches, are partitioning using same hash generator.

Hash code generator differs based on ReducerTraits.UNIFORM i.e. ReduceSinkOperator#computeMurmurHash()  or ReduceSinkOperator#computeHashCode(), leading to different hash code for same value.

Skip SMB join in such cases.

It fixes bug that results in incorrect join result. Example query and behavior explanation is added in https://issues.apache.org/jira/browse/HIVE-28480

No

No

Tested via `auto_sortmerge_join_18.q` added in this PR


ngsg:
The patch seems reasonable. LGTM

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
```diff
@@ -870,6 +871,35 @@ private boolean checkConvertJoinSMBJoin(JoinOperator joinOp, OptimizeTezProcCont
...
 
+    /* As SMB replaces last RS op from the joining branches and the JOIN op with MERGEJOIN, we need to ensure
+     * the RS before these RS, in both branches, are partitioning using same hash generator. It
+     * differs depending on ReducerTraits.UNIFORM i.e. ReduceSinkOperator#computeMurmurHash or
+     * ReduceSinkOperator#computeHashCode, leading to different code for same value. Skip SMB join in such cases.
+     */
+    Boolean prevRsHasUniformTrait = null;
+    for (Operator<? extends OperatorDesc> parentOp : joinOp.getParentOperators()) {
+      // Assertion of mandatory single parent is already being done in bucket version check earlier
+      Operator<?> op = parentOp.getParentOperators().get(0);
```
okumin, line 882: Can I assume we want to find ReduceSinkOperator-1 of `Operators -> ReduceSinkOperator-1 -> Operators -> ReduceSinkOperator-2(parentOp) ->JoinOperator`?

okumin (commented):
This looks fine to me at this point.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
okumin: That makes sense. Thanks

himanshu-mishra:
@okumin if this looks good, please merge the PR

okumin:
@himanshu-mishra Sorry, I don't have the right to merge a PR as I am not a committer. You need approval of any committer first.

okumin (approved):

himanshu-mishra:
@kasakrisz can you please review or add reviewer

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
```diff
@@ -870,6 +871,35 @@ private boolean checkConvertJoinSMBJoin(JoinOperator joinOp, OptimizeTezProcCont
...
+      while (op != null && !(op instanceof TableScanOperator || op instanceof ReduceSinkOperator
+              || op instanceof CommonJoinOperator)) {
+        // If op has parents it is guaranteed to be 1.
+        List<Operator<?>> parents = op.getParentOperators();
+        Preconditions.checkState(parents.size() == 0 || parents.size() == 1);
+        op = parents.size() == 1 ? parents.get(0) : null;
+      }
```
kasakrisz, line 889: There are several utility methods to find operators in branches of a plan in `ql.exec.OperatorUtils`
```
OperatorUtils.find*Operator*()
```
Is it possible to use one of them to find parent RS operators?

kasakrisz (commented):
This is a good candidate to fix this issue, but have you checked why the reducer traits are different?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
himanshu-mishra: There isn't a utility method there that returns first ancestor operator encountered from given set of classes. Here, we are only interested in ancestor RS if that is the source of data for current RS, if TS, JOIN are encountered, they are the source, and we do not need verifying them.

himanshu-mishra:
> This is a good candidate to fix this issue, but have you checked why the reducer traits are different?

Consider the sample query
```sql
SELECT * FROM (
    SELECT k, COUNT(DISTINCT v), SUM(v)
    FROM t_asj_18 GROUP BY k
) a LEFT JOIN (
    SELECT k, COUNT(v)
    FROM t_asj_18 GROUP BY k
) b ON a.k = b.k; 
```

In this query the aggregation `k, count(distinct v).. group by k` leads to group by and hence following RS operator's keys to be `k, v` while keeping partition column `k`, as we are aggregating on `k`. In such cases where key columns and partition columns differ, the `UNIFORM` trait is removed [refer commit](https://github.com/apache/hive/commit/d0acc5d309bba3bb4e6110f92d10e1ea247edd7e).
Note that the group by keys and partition columns differ only with `distinct` aggregate, removing that would lead to both being `a` thereby having `UNIFORM` trait.

himanshu-mishra:
@kasakrisz let me know if there are pending concerns.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
kasakrisz: a few lines earlier the same while loop exists https://github.com/apache/hive/blob/b13f529d944dc4c72d02d18e63757622985ada28/ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java#L853-L859

Could you please extract these to a utility method and move it to `OperatorUtils`

kasakrisz:
> @kasakrisz let me know if there are pending concerns.

Sorry, I was occupied in the last few days. I left only one comment.

kasakrisz (approved):
+1

PR #5413, opened 2024-08-27
HIVE-21584 to branch-3.1:Fix issue Java 11 preparation: system class loader is not URLClassLoader

beat4ocean:
HIVE-21584 to branch-3.1:Fix issue Java 11/17 preparation: system class loader is not URLClassLoader

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/common/StringInternUtils.java
+++ b/common/src/java/org/apache/hadoop/hive/common/StringInternUtils.java
@@ -29,74 +29,15 @@ import java.util.Map;
  */
 public class StringInternUtils {
 
-  // When a URI instance is initialized, it creates a bunch of private String
-  // fields, never bothering about their possible duplication. It would be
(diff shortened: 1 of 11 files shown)


zhangbutao:
I don't think we should continue to backport changes to branch-3.1.
You may try to use the latest Hive4 version which has many fixes.

beat4ocean:
> I don't think we should continue to backport changes to branch-3.1. You may try to use the latest Hive4 version which has many fixes.

But the usage ratio of Hive 3 series is very high, and currently many ecosystem components depend on the Hive 3 series.

zhangbutao:
> But the usage ratio of Hive 3 series is very high, and currently many ecosystem components depend on the Hive 3 series.

I know that. But IMHO, we won't put too much effort into branch-3, and I don't think we would release a new version based on branch-3. Most of developments focus on the branch-4 & master branch now.
I would suggest Hive users to try to upgrade hive version to Hive 4.
Thanks.

PR #4545, opened 2023-08-03
HIVE-27586: Parse dates from strings ignoring trailing (potentialy) invalid chars

zabetak:
## What changes were proposed in this pull request?

1. Change `DateParser` to allow partial parsing of the input string and late resolution.
2. Remove special handling of 'T' and space ' ' in date parser

## Why are the changes needed?

1. Return valid dates (not nulls) from timestamp strings with delimiters other than space and 'T' (e.g., `2023-08-03_16:02:00`)
2. Improve backwards compatibility by allowing partial parsing supported before HIVE-20007
3. Handle trailing invalid chars in a uniform way. Now `2023-08-03TGARBAGE` and `2023-08-03_GARBAGE` both return `2023-08-03`

## Does this PR introduce any user-facing change?

Various SQL functions and operations on dates will now return a valid date instead of `null`. A list of affected functions and concrete examples are present in HIVE-27586.

## How was this patch tested?

New + existing tests
```
mvn test -Dtest=TestDateParser
```

PR #5421, opened 2024-08-29
HIVE-28483: Fix for casting invalid date give wrong result instead of…

zratkai:
Fix for casting invalid date give wrong result instead of null

Change-Id: Iffee3e070fadbbfca83adb98561e95f216f7722e

Casting invalid date may gave back strange date instead of null.

Yes, it gives back null if casting an invalid date, instead of a strange result.

No.

Unit test.

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/common/type/Date.java
+++ b/common/src/java/org/apache/hadoop/hive/common/type/Date.java
@@ -84,7 +84,7 @@ public class Date implements Comparable<Date> {
   private static final LocalDate EPOCH = LocalDate.of(1970, 1, 1);
 
   private static final DateTimeFormatter PARSE_FORMATTER =
-      new DateTimeFormatterBuilder().appendValue(YEAR, 1, 10, SignStyle.NORMAL).appendLiteral('-')
+      new DateTimeFormatterBuilder().appendValue(YEAR, 1, 4, SignStyle.NORMAL).appendLiteral('-')
           .appendValue(MONTH_OF_YEAR, 1, 2, SignStyle.NORMAL).appendLiteral('-')
           .appendValue(DAY_OF_MONTH, 1, 2, SignStyle.NORMAL).toFormatter().withResolverStyle(ResolverStyle.STRICT);
 
@@ -177,13 +177,22 @@ public class Date implements Comparable<Date> {
    * @throws NullPointerException if {@code text} is null
    */
   public static Date valueOf(final String text) {
-    String s = Objects.requireNonNull(text).trim();
+    String trimedText = Objects.requireNonNull(text).trim();
     ParsePosition pos = new ParsePosition(0);
     try {
-      TemporalAccessor t = PARSE_FORMATTER.parseUnresolved(s, pos);
+      TemporalAccessor t = PARSE_FORMATTER.parse(trimedText, pos);
       if (pos.getErrorIndex() >= 0) {
-        throw new DateTimeParseException("Text could not be parsed to date", s, pos.getErrorIndex());
+        throw new DateTimeParseException("Text could not be parsed to date", trimedText, pos.getErrorIndex());
       }
+      //Check if there is still text left after parsing
+      if(pos.getIndex() < trimedText.length()) {
+        char lastChar = trimedText.charAt(pos.getIndex());
+        //Check if the first character of the remaining is a digit, e.g. "2023-08-0800" if yes, then it is a parse error and must throw an exception
+        if (lastChar >= '0' && lastChar <= '9'){
+          throw new DateTimeParseException("Text '" + trimedText + "' could not be parsed, unparsed text found at index " + pos.getIndex(), trimedText,
+              pos.getIndex());
+        }
+      }  
       return new Date(LocalDate.of(t.get(YEAR), t.get(MONTH_OF_YEAR), t.get(DAY_OF_MONTH)));
     } catch (DateTimeException e) {
       throw new IllegalArgumentException("Cannot create date, parsing error");
--- a/common/src/test/org/apache/hive/common/util/TestDateParser.java
+++ b/common/src/test/org/apache/hive/common/util/TestDateParser.java
@@ -45,35 +45,27 @@ public class TestDateParser {
 
   @Test
   public void testValidCases() throws Exception {
-    checkValidCase("1945-12-31", Date.valueOf("1945-12-31"));
-    checkValidCase("1946-01-01", Date.valueOf("1946-01-01"));
-    checkValidCase("2001-11-12", Date.valueOf("2001-11-12"));
-    checkValidCase("0004-05-06", Date.valueOf("0004-05-06"));
-    checkValidCase("1678-09-10", Date.valueOf("1678-09-10"));
-    checkValidCase("9999-10-11", Date.valueOf("9999-10-11"));
+    checkValidCase("1945-12-31", Date.of(1945,12,31));
+    checkValidCase("1946-01-01", Date.of(1946,1,1));
+    checkValidCase("2001-11-12", Date.of(2001,11,12));
+    checkValidCase("0004-05-06", Date.of(4,5,6));
+    checkValidCase("1678-09-10", Date.of(1678,9,10));
+    checkValidCase("9999-10-11", Date.of(9999,10,11));
 
     // Timestamp strings should parse ok
-    checkValidCase("2001-11-12 01:02:03", Date.valueOf("2001-11-12"));
+    checkValidCase("2001-11-12 01:02:03", Date.of(2001,11,12));
 
     // Leading spaces
-    checkValidCase(" 1946-01-01", Date.valueOf("1946-01-01"));
-    checkValidCase(" 2001-11-12 01:02:03", Date.valueOf("2001-11-12"));
+    checkValidCase(" 1946-01-01", Date.of(1946,01,01));
+    checkValidCase(" 2001-11-12 01:02:03", Date.of(2001,11,12));
   }
 
   @Test
   public void testParseDateFromTimestampWithCommonTimeDelimiter() {
-    for (String d : new String[] { "T", " ", "-", ".", "_", "" }) {
+    for (String d : new String[] { "T", " ", "-", ".", "_" }) {
       String ts = "2023-08-03" + d + "01:02:03";
       assertEquals("Parsing " + ts, Date.of(2023, 8, 3), DateParser.parseDate(ts));
     }
-  }
-
-  @Test
-  public void testParseDateFromValidDateLiteralWithTrailingDigits() {
-    assertEquals(Date.of(2023, 8, 8), DateParser.parseDate("2023-08-0800"));
-    // The result may seem unexpected but for many "08-08-20" is a valid date so there is no reason to reject
-    // "08-08-2023" and return null unless in the future Hive becomes stricter in terms of parsing dates.
-    assertEquals(Date.of(8, 8, 20), DateParser.parseDate("08-08-2023"));
   }
 
   @Test
@@ -86,5 +78,8 @@ public class TestDateParser {
     checkInvalidCase("0000-00-00");
     checkInvalidCase("2001-13-12");
     checkInvalidCase("2001-11-31");
+    checkInvalidCase("19999-10-11");
+    checkInvalidCase("08-08-2023");
+    checkInvalidCase("2023-08-0800");
   }
 }


common/src/test/org/apache/hive/common/util/TestDateParser.java
```diff
@@ -45,19 +45,19 @@ void checkInvalidCase(String strValue) {
...
-    checkValidCase("2001-11-12", Date.valueOf("2001-11-12"));
-    checkValidCase("0004-05-06", Date.valueOf("0004-05-06"));
-    checkValidCase("1678-09-10", Date.valueOf("1678-09-10"));
-    checkValidCase("9999-10-11", Date.valueOf("9999-10-11"));
+    checkValidCase("1945-12-31", Date.of(1945,12,31));
+    checkValidCase("1946-01-01", Date.of(1946,1,1));
+    checkValidCase("2001-11-12", Date.of(2001,11,12));
+    checkValidCase("0004-05-06", Date.of(4,5,6));
+    checkValidCase("1678-09-10", Date.of(1678,9,10));
+    checkValidCase("9999-10-11", Date.of(9999,10,11));
```
okumin, line 53: I wonder if this is just a clean-up or if the test fails without this change

common/src/test/org/apache/hive/common/util/TestDateParser.java
```diff
@@ -71,9 +71,6 @@ public void testParseDateFromTimestampWithCommonTimeDelimiter() {
   @Test
   public void testParseDateFromValidDateLiteralWithTrailingDigits() {
     assertEquals(Date.of(2023, 8, 8), DateParser.parseDate("2023-08-0800"));
-    // The result may seem unexpected but for many "08-08-20" is a valid date so there is no reason to reject
-    // "08-08-2023" and return null unless in the future Hive becomes stricter in terms of parsing dates.
-    assertEquals(Date.of(8, 8, 20), DateParser.parseDate("08-08-2023"));
```
okumin, line 76: I prefer the new behavior, but it looks intentional. I have no objections to change it

common/src/test/org/apache/hive/common/util/TestDateParser.java
zratkai: Actually it checked itself...
Here is the checkValidCase implementation:
  void checkValidCase(String strValue, Date expected) {
    Date dateValue = DateParser.parseDate(strValue);
    assertEquals(expected, dateValue);

    assertTrue(DateParser.parseDate(strValue, date));
    assertEquals(expected, date);
  }
  
  in the first row:
      Date dateValue = DateParser.parseDate(strValue);
 
then the implementation of DateParser.parseDate :

public static Date parseDate(final String text) {
    Objects.requireNonNull(text);

    // Date is a mutable class; do not return cached value
    Date result = new Date();
    return (parseDate(text, result)) ? result : null;
  }
    
and parseDate uses:

   Date date = DATE_CACHE.get(text);
   
 and:
 
   private static final LoadingCache<String, Date> DATE_CACHE =
      CacheBuilder.newBuilder().maximumSize(10 * 365).build(new CacheLoader<String, Date>() {
        @Override
        public Date load(final String text) throws Exception {
          return Date.valueOf(text);
        }
      });
      
So on both sides Date.valueOf was actually used, so it tested itself, which makes no sense for a test to test itself.

zratkai (commented):
@okumin many thanks for the review! Please check my comments!
Fixing the failing tests!

common/src/test/org/apache/hive/common/util/TestDateParser.java
okumin: That makes sense. Thanks for the explanation

common/src/test/org/apache/hive/common/util/TestDateParser.java
```diff
@@ -86,5 +78,8 @@ public void testInvalidCases() throws Exception {
     checkInvalidCase("0000-00-00");
     checkInvalidCase("2001-13-12");
     checkInvalidCase("2001-11-31");
+    checkInvalidCase("19999-10-11");
+    checkInvalidCase("08-08-2023");
+    checkInvalidCase("2023-08-0800");
```
okumin, line 83: I double-checked 4.0.0 returns `9999-10-11`, `0008-08-20`, and `2023-08-08`. The new behavior looks more reasonable to me.

common/src/test/org/apache/hive/common/util/TestDateParser.java
okumin: I guess we should ping @zabetak as it is him who implemented the test case
https://github.com/apache/hive/pull/4545

common/src/java/org/apache/hadoop/hive/common/type/Date.java
```diff
@@ -177,13 +177,22 @@ public void setTimeInMillis(long epochMilli) {
...
       }
+      //Check if there is still text left after parsing
+      if(pos.getIndex() < trimedText.length()) {
+        char lastChar = trimedText.charAt(pos.getIndex());
+        //Check if the first character of the remaining is a digit, e.g. "2023-08-0800" if yes, then it is a parse error and must throw an exception
+        if (lastChar >= '0' && lastChar <= '9'){
+          throw new DateTimeParseException("Text '" + trimedText + "' could not be parsed, unparsed text found at index " + pos.getIndex(), trimedText,
+              pos.getIndex());
+        }
+      }
```
okumin, line 195: This logic could be a little naive but I don't come up with a better solution

okumin (approved):
Looks good to me!

zabetak (commented):
I left some comments under HIVE-28483. The current behavior is expected and documented so I am not sure if we should change it. 

The changes here make the implementation more complex and more difficult to document. There are also better ways to `CAST` strings to dates than relying on defaults that are subject to change across versions.

zratkai:
@zabetak @okumin I replied in the ticket, here I copy it as well:

I can undestand this is documented,but I have doubts that what a data analyst would expect when giving an input like:  "03-08-2024" and get a result "0003-08-20", and say just drop the "24" from the string, and say it must be some garbage.  This issue comes from a real user who is complaining it is a bug, and I totally agree with it.

As a human I can say the "03-08-2024" is a reverse of "2024-08-03" and it meant 3rd of August 2024, and not 20th of August the year of 3, and 24 must be some kind of garbage. Saying, we just skip the 24 and act like there is nothing is absolutely misleading, I think there is no user who would expect that.

If a string do not conform with the default pattern the best is give back null, to make it clear that it does not conform with the pattern, instead giving back some strange date is absolutely misleading, and error-prone.


Is there any other system which acts like that? I have never seen any.

common/src/java/org/apache/hadoop/hive/common/type/Date.java
zratkai: I tried to find better solutions, but this was the most effective and clear option.

common/src/test/org/apache/hive/common/util/TestDateParser.java
zratkai: Agree.

zratkai (commented):
Please check my replies.

zabetak (commented):
The changes lead to behavior changes in various SQL functions. If agree to merge the patch then the changes need to be clearly documented in the JIRA and possibly other places as well.

common/src/test/org/apache/hive/common/util/TestDateParser.java
```diff
@@ -45,37 +45,29 @@ void checkInvalidCase(String strValue) {
...
-    checkValidCase(" 1946-01-01", Date.valueOf("1946-01-01"));
-    checkValidCase(" 2001-11-12 01:02:03", Date.valueOf("2001-11-12"));
+    checkValidCase(" 1946-01-01", Date.of(1946,01,01));
+    checkValidCase(" 2001-11-12 01:02:03", Date.of(2001,11,12));
   }
 
   @Test
   public void testParseDateFromTimestampWithCommonTimeDelimiter() {
-    for (String d : new String[] { "T", " ", "-", ".", "_", "" }) {
+    for (String d : new String[] { "T", " ", "-", ".", "_" }) {
```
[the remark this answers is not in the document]
zratkai: Yes, since it is absolutely misleading. Someone can say it is year 2023 August 3, other can say, it is year 301 August 23 and just drop the 20, as it is done now with the trailing characters, mentioned earlier. If we can consider dropping the trailing characters, why can not we consider doing the same with leading ones?
Please consider this example: "2023-08-21:02:03" .  Is it 2023 August 2 and 1:02:03, or 2023 August 21 and 02:03? Both interpretation can be argued, but a parsing should not rely on guessing, it must be on clear formatting.

common/src/java/org/apache/hadoop/hive/common/type/Date.java
```diff
@@ -84,7 +84,7 @@ public class Date implements Comparable<Date> {
   private static final LocalDate EPOCH = LocalDate.of(1970, 1, 1);
 
   private static final DateTimeFormatter PARSE_FORMATTER =
-      new DateTimeFormatterBuilder().appendValue(YEAR, 1, 10, SignStyle.NORMAL).appendLiteral('-')
+      new DateTimeFormatterBuilder().appendValue(YEAR, 1, 4, SignStyle.NORMAL).appendLiteral('-')
```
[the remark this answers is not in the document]
zratkai: Years must be not 4 characters long, since date 1-1-1, 45-10-20, 457-05-16, 896-08-01 are valid dates, and as we allow the month and days to be one or two characters long it would be strange to say, years must be at least 4 character long, but it can be 10 character long. Most of the programming languages support dates up to years 9999 or even less, so we are good with that limit.

common/src/test/org/apache/hive/common/util/TestDateParser.java
```diff
@@ -86,5 +78,8 @@ public void testInvalidCases() throws Exception {
     checkInvalidCase("0000-00-00");
     checkInvalidCase("2001-13-12");
     checkInvalidCase("2001-11-31");
+    checkInvalidCase("19999-10-11");
```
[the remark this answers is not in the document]
zratkai: I do not see the reason supporting something is not documented.

common/src/java/org/apache/hadoop/hive/common/type/Date.java
```diff
@@ -177,13 +177,22 @@ public void setTimeInMillis(long epochMilli) {
    * @throws NullPointerException if {@code text} is null
    */
   public static Date valueOf(final String text) {
```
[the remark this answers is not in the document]
zratkai: It won't be valid, since it is misleading. Is it year 6 July 8 or year 8 July 6?

zratkai (commented):
Please check my answers!

zratkai:
The ticket which is referred : "HIVE-27586: Parse dates from strings ignoring trailing (potentially) invalid chars" is about ignoring  trailing (potentially) characters, and not about truncating valid, but reversely formatted dates, which is a bug here reported by a customer. In the PR description it is even not mentioning this behavior.

Could you please be more specific what you mean by this?
"The changes lead to behavior changes in various SQL functions."

All the test are green, so I do not see what SQL functions would break.


If a user likes to convert from the not standard format, it can be done with CAST(<string-value> AS DATE FORMAT <date-pattern>) command, but the default one should be not something which is not now across the tech industry. Could you please mention any language which supports this format as default?

zratkai:
@zabetak @okumin Please check the updated ticket description. I made a research on other DB systems (Postgre, Oracle, Mysql). None of them give back error-prone result, instead exception or null.

common/src/java/org/apache/hadoop/hive/common/type/Date.java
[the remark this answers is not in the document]
zabetak: In most cases in Hive, DATE values describe a particular year/month/day, in the form YYYY-­MM-­DD. With this in mind, '06-07-08' is not that ambiguous; it means July 8, of year 6.

From a quick test, both before and after this PR the result is the same: July 8, of year 6

common/src/test/org/apache/hive/common/util/TestDateParser.java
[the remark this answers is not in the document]
zabetak: Various things are not documented but Hive users rely on them. I would like to enforce stricter bounds but I am sure if the change will be well-received by everyone.

common/src/test/org/apache/hive/common/util/TestDateParser.java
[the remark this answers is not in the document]
zabetak: Since Hive Dates are expected to be in the default format (YYYY-MM-DD), I would also consider `2023-08-21:02:03` to mean August 21, of year 2023. This is not very ambiguous once we know the default format.

> If we can consider dropping the trailing characters, why can not we consider doing the same with leading ones?

HIVE-27586 was mainly introduced to retain backward compatibility with previous versions that were dropping trailing chars. Leading invalid chars were never accepted I think.

zabetak:
> In the PR description it is even not mentioning this behavior.

In HIVE-27586, there is an explicit explanation about dates of the form `03-08-2023` and documents how these are treated after the changes there. There were also explicit unit tests added to guard against accidental changes.

> Could you please be more specific what you mean by this?
> "The changes lead to behavior changes in various SQL functions."

The changes in the parser incurred by this PR will affect at least the same SQL functions that are mentioned under HIVE-27586. Some test cases were modified by this PR and new limits were introduced so there are behavior changes.

> I made a research on other DB systems (Postgre, Oracle, Mysql). None of them give back error-prone result, instead exception or null.

Thanks for doing that! I fully agree with you that the default date format should be consistent across SQL engines. Ideally, I would like the parser to be very strict and only accept dates in the form 2023-08-03 (YYYY-MM-DD) but this behavior was not appreciated by all users and this is why HIVE-27586 came to the picture.

zratkai:
@zabetak Is there anything else apart from the debate here and in the JIRA ticket holding off this PR? I want to get a feeling what is pending to finalize this work to see how to advance.

zabetak:
@zratkai It seems that we are close in reaching consensus. I suggest to wait 48 hours to allow people to review all new elements and then merge it.

PR #5423, opened 2024-08-30
HIVE-28488: Merge multiple adjacent union distinct into single adjacent union distinct

ngsg:

This PR proposed a new optimization to reduce the shuffle when computing union distinct of multiple tables. The new optimization merges GroupBy operators computing distinct after Union, thus reducing the number of edges involved by UNION DISTINCT.
A new configuration key, `hive.optimize.merge.adjacent.union.distinct`, is introduced to configure this optimization.

Please check out the attached slides in the JIRA page (HIVE-28488) for further explanations.

To improve execution time of query containing many `UNION DISTINCT`.

No

No

The proposed optimization is tested by running TPC-DS query 49 and 75. This PR contains a qfile test to confirm that the patch optimizes query plan.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.hive.ql.optimizer;
+
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.ql.exec.GroupByOperator;
+import org.apache.hadoop.hive.ql.exec.Operator;
+import org.apache.hadoop.hive.ql.exec.ReduceSinkOperator;
+import org.apache.hadoop.hive.ql.exec.UnionOperator;
+import org.apache.hadoop.hive.ql.lib.DefaultGraphWalker;
+import org.apache.hadoop.hive.ql.lib.DefaultRuleDispatcher;
+import org.apache.hadoop.hive.ql.lib.Node;
+import org.apache.hadoop.hive.ql.lib.NodeProcessorCtx;
+import org.apache.hadoop.hive.ql.lib.RuleRegExp;
+import org.apache.hadoop.hive.ql.lib.SemanticDispatcher;
+import org.apache.hadoop.hive.ql.lib.SemanticGraphWalker;
+import org.apache.hadoop.hive.ql.lib.SemanticNodeProcessor;
+import org.apache.hadoop.hive.ql.lib.SemanticRule;
+import org.apache.hadoop.hive.ql.parse.ParseContext;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Stack;
+
+public class UnionDistinctMerger extends Transform {
+  private static final Logger LOG = LoggerFactory.getLogger(UnionDistinctMerger.class);
+
+  private static class UnionMergeProcessor implements SemanticNodeProcessor {
+    @Override
+    public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
+        Object... nodeOutputs) throws SemanticException {
+      UnionOperator upperUnionOperator = (UnionOperator) stack.get(stack.size() - 8);
+      GroupByOperator upperMiddleGroupByOperator = (GroupByOperator) stack.get(stack.size() - 7);
+      ReduceSinkOperator upperReduceSinkOperator = (ReduceSinkOperator) stack.get(stack.size() - 6);
+      GroupByOperator upperFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 5);
+
+      UnionOperator lowerUnionOperator = (UnionOperator) stack.get(stack.size() - 4);
+      GroupByOperator lowerFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 1);
+
+      if (upperFinalGroupByOperator.getConf().getAggregators().isEmpty() &&
+          lowerFinalGroupByOperator.getConf().getAggregators().isEmpty() &&
+          upperUnionOperator.getChildOperators().size() == 1 &&
+          upperMiddleGroupByOperator.getChildOperators().size() == 1 &&
+          upperReduceSinkOperator.getChildOperators().size() == 1 &&
+          upperFinalGroupByOperator.getChildOperators().size() == 1) {
+        LOG.info("Detect duplicate UNION-DISTINCT GBY patterns. Remove the first one.");
+
+        lowerUnionOperator.removeParent(upperFinalGroupByOperator);
+        for (Operator<?> upperUnionParent: upperUnionOperator.getParentOperators()) {
+          upperUnionParent.replaceChild(upperUnionOperator, lowerUnionOperator);
+          lowerUnionOperator.getParentOperators().add(upperUnionParent);
+        }
+        upperUnionOperator.setParentOperators(new ArrayList<>());
+
+        lowerUnionOperator.getConf().setNumInputs(lowerUnionOperator.getNumParent());
+      }
+
+      return null;
+    }
+  }
+
+  public ParseContext transform(ParseContext pCtx) throws SemanticException {
+    Map<SemanticRule, SemanticNodeProcessor> testRules = new LinkedHashMap<>();
+    StringBuilder pattern = new StringBuilder();
+    pattern.append(UnionOperator.getOperatorName()).append("%")
+        .append(GroupByOperator.getOperatorName()).append("%")
+        .append(ReduceSinkOperator.getOperatorName()).append("%")
+        .append(GroupByOperator.getOperatorName()).append("%")
+        .append(UnionOperator.getOperatorName()).append("%")
+        .append(GroupByOperator.getOperatorName()).append("%")
+        .append(ReduceSinkOperator.getOperatorName()).append("%")
+        .append(GroupByOperator.getOperatorName()).append("%");
+
+    testRules.put(new RuleRegExp("AdjacentDistinctUnion", pattern.toString()), new UnionMergeProcessor());
+
+    SemanticDispatcher disp = new DefaultRuleDispatcher(null, testRules, null);
+    SemanticGraphWalker ogw = new DefaultGraphWalker(disp);
+
+    List<Node> topNodes = new ArrayList<>();
+    topNodes.addAll(pCtx.getTopOps().values());
+    ogw.startWalking(topNodes, null);
+
+    return pCtx;
+  }
+}
+
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
@@ -81,7 +81,7 @@ public class UnionDistinctMerger extends Transform {
           upperMiddleGroupByOperator.getChildOperators().size() == 1 &&
           upperReduceSinkOperator.getChildOperators().size() == 1 &&
           upperFinalGroupByOperator.getChildOperators().size() == 1) {
-        LOG.info("Detect duplicate UNION-DISTINCT GBY patterns. Remove the later one.");
+        LOG.info("Detect duplicate UNION-DISTINCT GBY patterns. Remove the latter one.");
 
         lowerUnionOperator.removeParent(upperFinalGroupByOperator);
         for (Operator<?> lowerUnionParent: lowerUnionOperator.getParentOperators()) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
@@ -60,41 +60,77 @@ public class UnionDistinctMerger extends Transform {
     public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
         Object... nodeOutputs) throws SemanticException {
       UnionMergeContext context = (UnionMergeContext) procCtx;
+
+      // The stack contains at least 8 operators, UNION-GBY-RS-GBY-UNION-GBY-RS-GBY.
+      // The leftmost UNION is on stack.size() - 8 and the rightmost GBY is on stack.size() - 1.
       Collection<Operator> allOps = context.pCtx.getAllOps();
       for (int i = 1; i <= 8; i ++) {
-        if (!allOps.contains(stack.get(stack.size() - i))) {
+        Operator<?> op = (Operator<?>) stack.get(stack.size() - i);
+        
+        // We do not apply the optimization if some operators do not belong to query plan.
+        // This can be happened when we already merged some UNIONs before.
+        // For example, Suppose that a query plan looks like the below graph:
+        //   (1)UNION-GBY-RS-GBY-(3)UNION-GBY-RS-GBY
+        //   (2)UNION-GBY-RS-GBY-
+        // Then we merge (1)-(3) to (1) and them move on to merging (2)-(3). Without checking the presence of
+        // operators in (2) and (3), the merge process will fail as (3) is already removed.
+        if (!allOps.contains(op)) {
+          return null;
+        }
+        
+        // We do not apply the optimization if intermediate outputs are used by other operators.
+        if (i != 1 && op.getChildOperators().size() > 1) {
           return null;
         }
       }
 
       UnionOperator upperUnionOperator = (UnionOperator) stack.get(stack.size() - 8);
-      GroupByOperator upperMiddleGroupByOperator = (GroupByOperator) stack.get(stack.size() - 7);
-      ReduceSinkOperator upperReduceSinkOperator = (ReduceSinkOperator) stack.get(stack.size() - 6);
       GroupByOperator upperFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 5);
 
       UnionOperator lowerUnionOperator = (UnionOperator) stack.get(stack.size() - 4);
       GroupByOperator lowerFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 1);
 
       if (upperFinalGroupByOperator.getConf().getAggregators().isEmpty() &&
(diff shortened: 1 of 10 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  private class UnionMergeProcessor implements SemanticNodeProcessor {
+    @Override
+    public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
+        Object... nodeOutputs) throws SemanticException {
+      UnionMergeContext context = (UnionMergeContext) procCtx;
+      Collection<Operator> allOps = context.pCtx.getAllOps();
+      for (int i = 1; i <= 8; i ++) {
```
deniskuzZ, line 64: could we please add some comments around this `magic` indices

ngsg: Added some comments to explain the meaning of indices and logic around them.

ql/src/test/results/clientpositive/llap/explainuser_2.q.out
```diff
@@ -650,266 +650,267 @@ POSTHOOK: Output: hdfs://### HDFS PATH ###
...
-Reducer 20 <- Union 19 (SIMPLE_EDGE)
-Reducer 25 <- Union 24 (SIMPLE_EDGE), Union 26 (CONTAINS)
-Reducer 27 <- Union 26 (SIMPLE_EDGE), Union 28 (CONTAINS)
-Reducer 29 <- Union 28 (SIMPLE_EDGE)
-Reducer 3 <- Reducer 13 (SIMPLE_EDGE), Reducer 2 (SIMPLE_EDGE), Union 4 (CONTAINS)
-Reducer 5 <- Union 4 (SIMPLE_EDGE), Union 6 (CONTAINS)
-Reducer 7 <- Union 6 (SIMPLE_EDGE)
-Reducer 8 <- Reducer 2 (SIMPLE_EDGE), Reducer 29 (SIMPLE_EDGE), Union 6 (CONTAINS)
-Reducer 9 <- Map 1 (SIMPLE_EDGE), Map 15 (SIMPLE_EDGE)
+Map 10 <- Union 11 (CONTAINS)
```
deniskuzZ, line 653: should we keep `hive.optimize.merge.adjacent.union.distinct` = false for the existing tests? You've added a new test anyways

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -0,0 +1,184 @@
...
+        // GraphWalker uses retMap to determine if an operator has been visited.
+        // Clearing it after each walk() ensures that we visit GroupBy operator in every possible path.
+        retMap.clear();
+      }
+    }
+  }
+
+  public ParseContext transform(ParseContext pCtx) throws SemanticException {
+    Map<SemanticRule, SemanticNodeProcessor> testRules = new LinkedHashMap<>();
+    StringBuilder pattern = new StringBuilder();
```
deniskuzZ, line 162: do we need to construct this pattern every time? can we extract it to constants?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -0,0 +1,184 @@
...
+        Operator<?> op = (Operator<?>) stack.get(stack.size() - i);
+        
+        // We do not apply the optimization if some operators do not belong to query plan.
+        // This can be happened when we already merged some UNIONs before.
+        // For example, Suppose that a query plan looks like the below graph:
+        //   (1)UNION-GBY-RS-GBY-(3)UNION-GBY-RS-GBY
+        //   (2)UNION-GBY-RS-GBY-
+        // Then we merge (1)-(3) to (1) and them move on to merging (2)-(3). Without checking the presence of
+        // operators in (2) and (3), the merge process will fail as (3) is already removed.
+        if (!allOps.contains(op)) {
```
deniskuzZ, line 77: should we put `allOps` into the HashSet?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -0,0 +1,184 @@
...
+      UnionOperator lowerUnionOperator = (UnionOperator) stack.get(stack.size() - 4);
+      GroupByOperator lowerFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 1);
+
+      if (upperFinalGroupByOperator.getConf().getAggregators().isEmpty() &&
+          lowerFinalGroupByOperator.getConf().getAggregators().isEmpty()) {
+        LOG.info("Detect duplicate UNION-DISTINCT GBY patterns. Remove the latter one.");
+
+        // Step 0. UNION1->GBY1->RS1->GBY2->UNION2->GBY3->RS2->GBY4
+
+        // Step 1. Cut GBY2->UNINO1
```
deniskuzZ, line 99: GBY2->UNION2

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -0,0 +1,184 @@
...
+        //    OP2-
+        //    OP3-
+        //    OP4-
+        for (Operator<?> lowerUnionParent: lowerUnionOperator.getParentOperators()) {
+          lowerUnionParent.replaceChild(lowerUnionOperator, upperUnionOperator);
+          upperUnionOperator.getParentOperators().add(lowerUnionParent);
+        }
+        lowerUnionOperator.setParentOperators(new ArrayList<>());
+
+        // Step 3. Redirect the output of lowerFinalGroupByOperator to children of
```
deniskuzZ, line 119: children of upperFinalGroupByOperator
operator tree looks confusing since it's the same as on step 2

ql/src/test/results/clientpositive/llap/explainuser_2.q.out
ngsg: Okay. I add `set hive.optimize.merge.adjacent.union.distinct=false` in explainuser_2.q and unionDistinct_3.q to preserve their output files.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: Moved the pattern to outside of constructor.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: Changed `allOps` from Collection(List) to HashSet.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: fixed

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: Yes, the comment should be `children of upperFinalGroupByOperator`. I updated the comment with a correct diagram explaining step 3. Sorry for the confusion due to the incomplete comment.

ql/src/test/results/clientpositive/llap/explainuser_2.q.out
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -63,10 +75,10 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
 
       // The stack contains at least 8 operators, UNION-GBY-RS-GBY-UNION-GBY-RS-GBY.
       // The leftmost UNION is on stack.size() - 8 and the rightmost GBY is on stack.size() - 1.
-      Collection<Operator> allOps = context.pCtx.getAllOps();
+      HashSet<Operator> allOps = new HashSet<>(context.pCtx.getAllOps());
```
deniskuzZ, line 78: please code to interface
````
Set<Operator> allOps = new HashSet<>(context.pCtx.getAllOps())
````

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java [hunk 8]
```diff
@@ -0,0 +1,181 @@
...
+        //   (1)UNION-GBY-RS-GBY-(3)UNION-GBY-RS-GBY
+        //   (2)UNION-GBY-RS-GBY-
+        // Then we merge (1)-(3) to (1) and them move on to merging (2)-(3). Without checking the presence of
+        // operators in (2) and (3), the merge process will fail as (3) is already removed.
+        if (!allOps.contains(op)) {
+          return null;
+        }
+
+        // We do not apply the optimization if intermediate outputs are used by other operators.
+        if (i != 1 && op.getChildOperators().size() > 1) {
```
deniskuzZ, line 94: sorry, I don't follow here. why in that case we are doing 
````
Redirect the output of lowerFinalGroupByOperator to children of upperFinalGroupByOperator
````
upperFinalGroupByOperator index is `stack.size() - 5`. Should we ignore it as well?

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java [hunk 9]
```diff
@@ -0,0 +1,181 @@
...
+        }
+      }
+
+      UnionOperator upperUnionOperator = (UnionOperator) stack.get(stack.size() - 8);
+      GroupByOperator upperFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 5);
+
+      UnionOperator lowerUnionOperator = (UnionOperator) stack.get(stack.size() - 4);
+      GroupByOperator lowerFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 1);
+
+      if (upperFinalGroupByOperator.getConf().getAggregators().isEmpty() &&
```
deniskuzZ, line 105: could you please add a comment on what it means if `aggregators().isEmpty()`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: fixed

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: added a comment about `aggregators().isEmpty()`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: I made a mistake in the comment again; I have rewritten it with clearer words.

The goal of this optimization is merging two distinct-union-computations (UNION-GBY-RS-GBY) by disconnecting lower one from operator graph. So we don't connect(redirect) lowerFinalGroupByOperator and the children of upperFinalGroupByOperator. Instead, we connect upperFinalGroupByOperator and the children of lowerFinalGroupByOperator. Since we skip the check when `i == 1`, there can be multiple children under lowerFinalGroupByOpeartor.

deniskuzZ: 👍

deniskuzZ (approved):
LGTM +1, pending tests
@ngsg, could you please check sonar: https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=5423&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

ngsg:
@deniskuzZ , I checked sonar and resolved 3 out of 6 reported issues. The remaining issues are not resolved because of the following reasons:

1.
https://github.com/apache/hive/blob/612afeb8d7268ebc46cacd44651795bb7e5f999e/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java#L73
This method uses return statement just for quick exit. Since its returned values is not used, returning the same value(null) is not harmful.

2.
https://github.com/apache/hive/blob/612afeb8d7268ebc46cacd44651795bb7e5f999e/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java#L79
The type of `context.pCtx.getAllOps()` is raw `Collection<Operator>`, so the `Set` uses it, too.

3.
https://github.com/apache/hive/blob/612afeb8d7268ebc46cacd44651795bb7e5f999e/ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java#L136
Sonar determines this comment as the block of commented-out code, which is not true.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -0,0 +1,185 @@
...
+      UnionOperator upperUnionOperator = (UnionOperator) stack.get(stack.size() - 8);
+      GroupByOperator upperFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 5);
+
+      UnionOperator lowerUnionOperator = (UnionOperator) stack.get(stack.size() - 4);
+      GroupByOperator lowerFinalGroupByOperator = (GroupByOperator) stack.get(stack.size() - 1);
+
+      // We can apply the optimization if there is no aggregators in final GroupBy operators. The absence of
+      // aggregators ensures that we are merging two distinct computation.
+      if (upperFinalGroupByOperator.getConf().getAggregators().isEmpty() &&
+          lowerFinalGroupByOperator.getConf().getAggregators().isEmpty()) {
```
okumin, line 109: Can we finally optimize where UDAFs are used? I guess we can as long as aggregators have compatibility. I could be wrong

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -0,0 +1,185 @@
...
+        lowerUnionOperator.removeParent(upperFinalGroupByOperator);
+
+        // Step 2.
+        //   Connect the parent of lowerUnionOperator and upperUnionOperator.
+        //   Disconnect lowerUnionOperator from operator graph.
+        // Before step 2:
+        //    {OP1, 2}-UNION1->GBY1->RS1->GBY2-{}
+        //    {OP3, 4}-UNION2->GBY3->RS2->GBY4-{OP5, 6, ...}
+        // After step 2:
+        //    {OP1, 2}-UNION1->GBY1->RS1->GBY2-{}
```
okumin, line 124: `{OP1, 2, 3, 4}-UNION1->GBY1->RS1->GBY2-{}` or `{OP1, 2, ...}-UNION1->GBY1->RS1->GBY2-{}` If I understand correctly?

ql/src/test/queries/clientpositive/explainuser_2.q
```diff
@@ -6,6 +6,7 @@ set hive.strict.checks.bucketing=false;
 set hive.explain.user=true;
 set hive.metastore.aggregate.stats.cache.enabled=false;
 set hive.cbo.fallback.strategy=NEVER;
+set hive.optimize.merge.adjacent.union.distinct=false;
```
okumin, line 9: Why do we disable it here? I am worried that the result of non-EXPLAIN could change

ql/src/test/queries/clientpositive/unionDistinct_3.q
```diff
@@ -3,6 +3,7 @@
 --! qt:dataset:src
 set hive.mapred.mode=nonstrict;
 set hive.explain.user=false;
+set hive.optimize.merge.adjacent.union.distinct=false;
```
okumin, line 6: Why do we disable it here? I am worried that the result of non-EXPLAIN could change

okumin (approved):
Looks good as long as we can explain why we have to disable it in `explainuser_2.q` and `unionDistinct_3.q`.

ql/src/test/queries/clientpositive/explainuser_2.q
deniskuzZ: hi @okumin, I've asked to keep the existing qtests unchanged as a regression when this feature is disabled + there is a new qtest added specifically to validate new feature. 
Note, you could check the outputs for those 2 qtests with enabled optimization in https://github.com/apache/hive/pull/5423/commits/74178b15f72bb8d53ea7b5fb934103c296f6c754#diff-f40b3c8ff967ea71b2e610dc10d9ed9a749f1054ad0d7144a3326f1b5820eb3f

ql/src/test/queries/clientpositive/unionDistinct_3.q
okumin: Now, I understand this is intended

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
deniskuzZ: missed that. @ngsg could you please check

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: The goal of this optimization is merging UNION DISTINCT pattern, so I didn't considered any UDAF here. In my opinion, if there exists any aggregator, we cannot merge 2 UNION-GBY-RS-GBY pattern.

A GroupBy with aggregator forwards key columns and aggregator output columns. Therefore, if an upperFinalGroupBy contains aggregators, than lowerUnion takes not only key columns but also aggregator output columns, which cannot be obtained in merged plan.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
ngsg: It should be {OP1, 2, 3, 4}. Thanks for the thorough review.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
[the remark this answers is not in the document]
ngsg: Yes, UNION2 is correct. It was fixed in previous commits.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/UnionDistinctMerger.java
```diff
@@ -0,0 +1,185 @@
...
+    public final ParseContext pCtx;
+
+    public UnionMergeContext(ParseContext pCtx) {
+      this.pCtx = pCtx;
+    }
+  }
+
+  private class UnionMergeProcessor implements SemanticNodeProcessor {
+    @Override
+    public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
```
deniskuzZ, line 73: @ngsg, should we return `Void`?

ngsg: Yes. Changed to Void.

PR #5424, opened 2024-08-30
HIVE-28489: Partition the input data of GroupBy with GroupingSet

ngsg:

This patch introduces a new optimization for GroupBy with GroupingSet. GroupingSet often emits too much rows and becomes a bottleneck of query execution. The proposed optimization tries mitigating this situation by reducing the cardinality of GroupBy key. For the detailed explanation, please checkout the slides attached in the JIRA page (HIVE-28489).

This patch introduces a new configuration key: `hive.optimize.grouping.set.threshold`. The default value of 1,000,000,000 is chosen because it showed best performance in 10TB TPC-DS experiment. Setting this value to negative number disables the optimization.

To improve query execution time of GroupBy with GroupingSet.

No

No

We tested this patch using 10TB TPC-DS experiment. This patch contains a qfile test to verify optimized query plan.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
@@ -1,0 +1,344 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.hive.ql.optimizer;
+
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.metastore.api.FieldSchema;
+import org.apache.hadoop.hive.ql.exec.AbstractMapJoinOperator;
+import org.apache.hadoop.hive.ql.exec.ColumnInfo;
+import org.apache.hadoop.hive.ql.exec.GroupByOperator;
+import org.apache.hadoop.hive.ql.exec.Operator;
+import org.apache.hadoop.hive.ql.exec.OperatorFactory;
+import org.apache.hadoop.hive.ql.exec.ReduceSinkOperator;
+import org.apache.hadoop.hive.ql.exec.RowSchema;
+import org.apache.hadoop.hive.ql.exec.SelectOperator;
+import org.apache.hadoop.hive.ql.exec.Utilities;
+import org.apache.hadoop.hive.ql.io.AcidUtils;
+import org.apache.hadoop.hive.ql.lib.DefaultGraphWalker;
+import org.apache.hadoop.hive.ql.lib.DefaultRuleDispatcher;
+import org.apache.hadoop.hive.ql.lib.Node;
+import org.apache.hadoop.hive.ql.lib.NodeProcessorCtx;
+import org.apache.hadoop.hive.ql.lib.RuleRegExp;
+import org.apache.hadoop.hive.ql.lib.SemanticDispatcher;
+import org.apache.hadoop.hive.ql.lib.SemanticGraphWalker;
+import org.apache.hadoop.hive.ql.lib.SemanticNodeProcessor;
+import org.apache.hadoop.hive.ql.lib.SemanticRule;
+import org.apache.hadoop.hive.ql.parse.ParseContext;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+import org.apache.hadoop.hive.ql.plan.ColStatistics;
+import org.apache.hadoop.hive.ql.plan.ExprNodeColumnDesc;
+import org.apache.hadoop.hive.ql.plan.ExprNodeDesc;
+import org.apache.hadoop.hive.ql.plan.ExprNodeFieldDesc;
+import org.apache.hadoop.hive.ql.plan.MapJoinDesc;
+import org.apache.hadoop.hive.ql.plan.PlanUtils;
+import org.apache.hadoop.hive.ql.plan.ReduceSinkDesc;
+import org.apache.hadoop.hive.ql.plan.SelectDesc;
+import org.apache.hadoop.hive.ql.plan.TableDesc;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Stack;
+
+public class GroupingSetOptimizer extends Transform {
+  private static final Logger LOG = LoggerFactory.getLogger(GroupingSetOptimizer.class);
+
+  private static class GroupingSetProcessorContext implements NodeProcessorCtx {
+    public final long bytesPerReducer;
+    public final int maxReducers;
+    public final long groupingSetThreshold;
+
+    public GroupingSetProcessorContext(HiveConf hiveConf) {
+      bytesPerReducer = hiveConf.getLongVar(HiveConf.ConfVars.BYTES_PER_REDUCER);
+      maxReducers = hiveConf.getIntVar(HiveConf.ConfVars.MAX_REDUCERS);
+      groupingSetThreshold = hiveConf.getLongVar(HiveConf.ConfVars.HIVE_OPTIMIZE_GROUPING_SET_THRESHOLD);
+    }
+  }
+
+  private static class GroupingSetProcessor implements SemanticNodeProcessor {
+    @Override
+    public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
+        Object... nodeOutputs) throws SemanticException {
+      GroupingSetProcessorContext context = (GroupingSetProcessorContext) procCtx;
+      GroupByOperator gby = (GroupByOperator) nd;
+      if (!isGroupByFeasible(gby, context)) {
+        return null;
+      }
+
+      Operator<?> parentOp = gby.getParentOperators().get(0);
+      if (!isParentOpFeasible(parentOp)) {
+        return null;
+      }
+
+      String partitionCol = selectPartitionColumn(gby, parentOp);
+      if (partitionCol == null) {
+        return null;
+      }
+
+      LOG.info("Applying GroupingSetOptimization: partitioning the input data of {} by {}",
+          gby, partitionCol);
+
+      ReduceSinkOperator rs = createReduceSink(parentOp, partitionCol, context);
+
+      parentOp.removeChild(gby);
+      // gby.setParentOperators(Arrays.asList(rs));
+      // NOTE: The above expression does not work because GBY refers to _colN
+      //  while input columns are VALUE._colN. Therefore, we should either modify GBY expressions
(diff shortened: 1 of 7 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,363 @@
...
+import org.apache.hadoop.hive.ql.lib.SemanticDispatcher;
+import org.apache.hadoop.hive.ql.lib.SemanticGraphWalker;
+import org.apache.hadoop.hive.ql.lib.SemanticNodeProcessor;
+import org.apache.hadoop.hive.ql.lib.SemanticRule;
+import org.apache.hadoop.hive.ql.parse.ParseContext;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+import org.apache.hadoop.hive.ql.plan.ColStatistics;
+import org.apache.hadoop.hive.ql.plan.ExprNodeColumnDesc;
+import org.apache.hadoop.hive.ql.plan.ExprNodeDesc;
+import org.apache.hadoop.hive.ql.plan.ExprNodeFieldDesc;
```
BsoBird, line 47: Remove unused import entries.

BsoBird (commented):
It seems like there are no major issues. However, it would be best to have a few more people review it together.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
ngsg: removed an unused import.

okumin:
The intention is admirable, but I am not 100% confident that this is the best approach because I need to know how GROUPING SETS works. It still needs time for me

BsoBird:
> The intention is admirable, but I am not 100% confident that this is the best approach because I need to know how GROUPING SETS works. It still needs time for me

Hello. @okumin  Trino seems to have adopted a similar scheme to optimize grouping sets. It appears that the modifications in this submission are intended to port the optimization rules from Trino to HIVE. Since Trino has been using this rule for a long time, and we have received positive feedback from users after introducing this patch, I personally believe that the approach used in this submission is a viable solution. After all, if we find a better solution, it would be quite easy for us to replace the changes introduced by this PR.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,362 @@
...
+        ExprNodeDesc key = gby.getConf().getKeys().get(groupingSetKeyPosition);
+
+        if (key instanceof ExprNodeColumnDesc) {
+          candidates.add(((ExprNodeColumnDesc) key).getColumn());
+        }
+      }
+      candidates.retainAll(colNamesInSignature);
+
+      List<ColStatistics> columnStatistics = new ArrayList<>(parentOp.getStatistics().getColumnStats());;
+      columnStatistics.sort(new Comparator<ColStatistics>() {
```
okumin, line 218: We may simplify it by using `Comparator.comparingLong(...).reversed()`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,362 @@
...
+      List<Integer> groupingSetKeys = listGroupingSetKeyPositions(gby.getConf().getListGroupingSets());
+      Set<String> candidates = new HashSet<>();
+      for (Integer groupingSetKeyPosition: groupingSetKeys) {
+        ExprNodeDesc key = gby.getConf().getKeys().get(groupingSetKeyPosition);
+
+        if (key instanceof ExprNodeColumnDesc) {
+          candidates.add(((ExprNodeColumnDesc) key).getColumn());
+        }
+      }
+      candidates.retainAll(colNamesInSignature);
```
okumin, line 215: I am not understanding why we can put all of the parent operator.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,362 @@
...
+        }
+      }
+
+      List<FieldSchema> valueFields =
+          PlanUtils.getFieldSchemasFromColumnList(valueColumns, valueColumnNames, 0, "");
+      TableDesc valueTable = PlanUtils.getReduceValueTableDesc(valueFields);
+
+      TableDesc keyTable = PlanUtils.getReduceKeyTableDesc(new ArrayList<>(), "", "");
+
+      List<ExprNodeDesc> keyColumns = new ArrayList<>();
```
okumin, line 279: I could be wrong. Should we also use the partition key as a sort key to optimize the hash table utilization? I see we already achieved a good result with this implementation, so I might be wrong.
If we do that, I wonder if we can generalize the utility to insert REDUCE_SINK + SEL somewhere because it is a useful conversion...

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,362 @@
...
+      List<String> keyColumnNames = new ArrayList<>();
+      List<List<Integer>> distinctColumnIndices = new ArrayList<>();
+
+      ReduceSinkDesc rsConf = new ReduceSinkDesc(keyColumns, 0, valueColumns, keyColumnNames,
+          distinctColumnIndices, valueColumnNames, -1, partCols, -1, keyTable, valueTable,
```
okumin, line 284: I remember the argument between `partCols` and `keyTable` is # of reducers. We can move L294.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,362 @@
...
+      List<ColumnInfo> selSignature = new ArrayList<>();
+      Map<String, ExprNodeDesc> colExprMap = new HashMap<>();
+
+      for (ColumnInfo pColInfo: signature) {
+        String origColName = pColInfo.getInternalName();
+        String rsColName = Utilities.ReduceField.VALUE + "." + origColName;
+
+        ColumnInfo selColInfo = new ColumnInfo(pColInfo);
+
+        ExprNodeDesc selExpr = new ExprNodeColumnDesc(pColInfo.getType(), rsColName, rsColName, false);
```
okumin, line 312: Should the 3rd arg be NULL?

ql/src/test/queries/clientpositive/groupingset_optimize_hive_28489.q
```diff
@@ -0,0 +1,26 @@
+create table grp_set_test (key string, value string, col0 int, col1 int, col2 int, col3 int);
```
okumin, line 1: We may test this feature with both EXPLAIN and non-EXPLAIN. That's because all of relevant test cases, i.e. `groupingset_optimize_hive_28489.q` and TPC-DS, run only EXPLAIN.

ngsg: Added two identical select commands that run with and without optimization.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    private String selectPartitionColumn(GroupByOperator gby, Operator<?> parentOp) {
+      if (!(parentOp.getSchema() != null && parentOp.getSchema().getSignature() != null)) {
+        LOG.debug("Skip grouping-set optimization as the parent operator {} does not provide signature",
+            parentOp);
+        return null;
+      }
+
+      if (!(parentOp.getStatistics() != null && parentOp.getStatistics().getColumnStats() != null)) {
```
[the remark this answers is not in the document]
ngsg: fixed.
I checked the suggested methods, but they are not appropriate to check `Statistics#getNumRows() > 0` and `ColStatistics#getCountDistint() > 0`. So I added manual checks for them instead of calling the methods.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
ngsg: I agree that inserting RS + SEL might be useful optimization if we can find the place where heavy sort happens. However, I'm not sure whether HashTable can fully take advantage of pre-sorting.
I'll test your idea in our cluster and share the result later on.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,362 @@
...
+        // There is no partitioning followed by this parentOp. Continue optimization.
+        return true;
+      }
+
+      if (rs.getConf().getPartitionCols() != null && rs.getConf().getPartitionCols().size() > 0) {
+        // This rs might be irrelevant to the target GroupBy operator. For example, the following query:
+        //   SELECT a, b, sum(c) FROM (SELECT a, b, c FROM tbl DISTRIBUTE BY c) z GROUP BY rollup(a, b)
+        // won't be optimized although 'DISTRIBUTE BY c' is irrelevant to the key columns of GroupBy.
+        LOG.debug("Skip grouping-set optimization in order not to introduce possibly redundant shuffle.");
+        return false;
```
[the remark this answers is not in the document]
ngsg: I think yes, but then we have to implement complex column tracker to determine whether the optimization introduces redundant partitioning or not.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
ngsg: In the above code, we gather the parent columns that are used for groupingSet. The for-loop gathers all of the such columns, and I call retainAll() just to ensure that the columns gathered from groupingSet actually exist in parent operator.

okumin: Ah, ok. Please ignore. I thought I saw not `retainAll` but `addAll`. My bad

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
ngsg: @okumin

I tested your idea using 10TB TPC-DS dataset. I used query18, query22, query67 and set `hive.optimize.grouping.set.threshold` to 1,000,000,000. I ran each query 3 times.

||query18|query22|query67|
|---|---|---|---|
|# GBY input rows estimation|382,977,011|258,698,384|5,279,977,323|
|# GBY output rows estimation|1,914,885,055|1,293,491,920|47,519,795,907|
|Query execution time(current impl.) |27.575, 24.089, 23.265|14.827, 13.925, 12.959|426.62, 419.793, 414.403|
|Query execution time(with additional sort)|27.636, 24.836, 24.054|16.531, 15.729, 15.624|318.425, 323.733, 336.545|

There is almost no change in the query execution time of relatively small GroupBys (query18 and query22). However, it appears that a large GroupBy benefits from the additional sort. Therefore, I conclude that adding a sort before large GroupBys is worthwhile. The latest commit(efe982c9) includes an implementation of additional sort that I used for the test.

okumin (commented):
Mostly, looks good to me based on my experience. Only one comment.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/GroupingSetOptimizer.java
```diff
@@ -0,0 +1,379 @@
...
+      List<FieldSchema> keyFields =
+          PlanUtils.getFieldSchemasFromColumnList(keyColumns, keyColumnNames, 0, "");
+      TableDesc keyTable = PlanUtils.getReduceKeyTableDesc(keyFields, "+", "z");
+      List<List<Integer>> distinctColumnIndices = new ArrayList<>();
+
+      // If we run SetReducerParallelism after this optimization, then we don't have to compute numReducers.
+      int numReducers = Utilities.estimateReducers(
+          parentOp.getStatistics().getDataSize(), context.bytesPerReducer, context.maxReducers, false);
+
+      ReduceSinkDesc rsConf = new ReduceSinkDesc(keyColumns, 0, valueColumns, keyColumnNames,
```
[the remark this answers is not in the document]
ngsg: You're right. Fixed it to use `keyColumns.size()`.

okumin (approved):
+1

PR #5425, opened 2024-08-30
HIVE-28490: Stop removing retainable DPP sources

ngsg:

This patch changes the implementation of SharedWorkOptimizer#findAscendantWorkOperators. For a given operator, original findAscendantWorkOperators() returns a set of operators that belongs to ancestor vertices of the vertex that given operator belongs to. Therefore, some operators that are irrelevant to the given operator also belong to the resultant set, which leads to wrong DPP source removal. The new implementation returns a set of ancestor operators of the given operator.
Please check out the slides attached in the JIRA page (HIVE-28490) for further explanations.

In order to prevent the removal of retainable DPP sources that are still useful in runtime.

No

No

We tested this patch by running all TPC-DS queries in cluster. This patch contains a qfile test that reproduces the wrong DPP source removal.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
@@ -28,6 +28,7 @@ import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -1195,7 +1196,7 @@ public class SharedWorkOptimizer extends Transform {
       Operator<?> op = dppsOp1.get(i);
       if (op instanceof ReduceSinkOperator) {
         Set<Operator<?>> ascendants =
-            findAscendantWorkOperators(optimizerCache, op);
+            findAscendantOperators(optimizerCache, op);
         if (ascendants.contains(tsOp2)) {
           // This should not happen, we cannot merge
           return false;
@@ -1206,7 +1207,7 @@ public class SharedWorkOptimizer extends Transform {
       Operator<?> op = dppsOp2.get(i);
       if (op instanceof ReduceSinkOperator) {
         Set<Operator<?>> ascendants =
-            findAscendantWorkOperators(optimizerCache, op);
+            findAscendantOperators(optimizerCache, op);
         if (ascendants.contains(tsOp1)) {
           // This should not happen, we cannot merge
           return false;
@@ -1633,7 +1634,7 @@ public class SharedWorkOptimizer extends Transform {
         Collection<Operator<?>> c = optimizerCache.tableScanToDPPSource
             .get((TableScanOperator) op);
         for (Operator<?> dppSource : c) {
-          Set<Operator<?>> ascendants = findAscendantWorkOperators(optimizerCache, dppSource);
+          Set<Operator<?>> ascendants = findAscendantOperators(optimizerCache, dppSource);
           if (!Collections.disjoint(ascendants, discardedOps)) {
             // Remove branch
             removeBranch(dppSource, dppBranches, ops);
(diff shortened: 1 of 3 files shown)


BsoBird:
tips: HIVE-24890 => HIVE-28490

ngsg:
@BsoBird , thanks for pointing out the typo. I fixed it.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1963,33 +1962,24 @@ private static Set<Operator<?>> findParentWorkOperators(ParseContext pctx,
...
-          // Check for DPP and semijoin DPP
-          for (Operator<?> parent : optimizerCache.tableScanToDPPSource.get((TableScanOperator) op)) {
-            set.addAll(findWorkOperators(optimizerCache, parent));
-          }
+  private static Set<Operator<?>> findAscendantWorkOperators(SharedWorkOptimizerCache optimizerCache,
+      Operator<?> start) {
+    Set<Operator<?>> visited = new HashSet<>();
+    visited.add(start);
+
+    List<Operator<?>> remaining = new ArrayList<>(start.getParentOperators());
```
deniskuzZ, line 1970: maybe use Queue<Operator<?>> remaining = new LinkedList<>()

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1963,33 +1962,24 @@ private static Set<Operator<?>> findParentWorkOperators(ParseContext pctx,
...
+    while (!remaining.isEmpty()) {
+      Operator<?> op = remaining.remove(0);
+      if (!visited.contains(op)) {
+        visited.add(op);
+        remaining.addAll(op.getParentOperators());
+        if (op instanceof TableScanOperator) {
```
deniskuzZ, line 1976: please restore removed comments

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1963,33 +1962,24 @@ private static Set<Operator<?>> findParentWorkOperators(ParseContext pctx,
...
       }
-      workOps = set;
-      result.addAll(set);
     }
-    return result;
+
+    return visited;
   }
 
   private static Set<Operator<?>> findChildWorkOperators(ParseContext pctx,
```
deniskuzZ, line 1985: can we drop it since it's not used?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1963,33 +1962,24 @@ private static Set<Operator<?>> findParentWorkOperators(ParseContext pctx,
...
+      Operator<?> start) {
+    Set<Operator<?>> visited = new HashSet<>();
+    visited.add(start);
+
+    List<Operator<?>> remaining = new ArrayList<>(start.getParentOperators());
+    while (!remaining.isEmpty()) {
+      Operator<?> op = remaining.remove(0);
+      if (!visited.contains(op)) {
+        visited.add(op);
+        remaining.addAll(op.getParentOperators());
```
deniskuzZ, line 1975: should we ignore `ReduceSinkOperator` parents?
````
if (parent instanceof ReduceSinkOperator) {
  continue;
}
````

Comment (deniskuzZ) [inline ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java:1975]: can we get NULL from `op.getParentOperators()`?

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java [hunk 5]
```diff
@@ -1963,33 +1962,24 @@ private static Set<Operator<?>> findParentWorkOperators(ParseContext pctx,
...
-            if (parent instanceof ReduceSinkOperator) {
-              set.addAll(findWorkOperators(optimizerCache, parent));
-            }
-          }
-        } else if (op instanceof TableScanOperator) {
-          // Check for DPP and semijoin DPP
-          for (Operator<?> parent : optimizerCache.tableScanToDPPSource.get((TableScanOperator) op)) {
-            set.addAll(findWorkOperators(optimizerCache, parent));
-          }
+  private static Set<Operator<?>> findAscendantWorkOperators(SharedWorkOptimizerCache optimizerCache,
```
deniskuzZ, line 1965: do I understand it right that original `findWorkOperators` collects recursively both operator's parents and children starting from `start`, but we only interested in parents?

Note findWorkOperators leverages the cache, do we need the same here?

deniskuzZ:
thanks for the fix @ngsg! sorry for the late reply. I've added a few comments

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
ngsg: changed to use Queue and LinkedList

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
ngsg: restored the comments

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
ngsg: Yes, we can. Removed `findChildWorkOperators` and `findParentWorkOperators` method.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
ngsg: Yes. Changed to do null-check before add `op.getParentOperators()`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
ngsg: 1 .
Yes, `findWorkOperators` returns the set of operators that will be packed in to the same work (`Vertex` in Tez and `BaseWork` in Hive).
Using `findWorkOperators`, we can compute the set of operators that belongs to the same work or its ancestor work. But this set is not our interest because the dependency between works does not always imply the dependency between operators.

For example, we can find 5 works from the below operator graph. (The same as the one in the slide)
```
Work1: MAPJOIN_217, SEL_27, GBY_29, SEL_140, GBY_141, RS_142, SEL_185, GBY_186, RS_187
Work2: GBY_143, RS_144
Work3: TS_3, FIL_116
Work4: GBY_188, RS_189
Work5: TS_41, FIL_121
```
In the example, RS_189 belongs to Work4, which depends on Work1. But RS_189 does not depend on RS_142, which is a member of Work1.

2 .
Since **the set of ascendant operators** is used less frequently than **the set of operators that belong to the same work**, I think we don't need another cache that is dedicated to `findAscendantWorkOperators`.

3 .
We don't use `findParentWorkOperators` anymore, and I don't know its original usage. Since findParentWorkOperators is just a one-step version of findAscendantWorkOperators, there might be the same issue if one used this method to find any dependency conflicts between operators.

4 .
I have changed the name of the method to `findAscendantOperators` as the method does not use the concept of work here.

h

deniskuzZ: 👍 , thanks for the explanation @ngsg!

deniskuzZ (approved):
LGTM +1

deniskuzZ:
@ramesh0201, @kasakrisz, could you please take a look at this PR as a second pair of eyes? thanks!

deniskuzZ:
ok, since there were no other reviews, I am merging this PR

PR #5428, opened 2024-09-03
HIVE-28494: Iceberg: mvn build enables iceberg module by default

zhangbutao:

[HIVE-25027](https://issues.apache.org/jira/browse/HIVE-25027) hidden the iceberg module by default. IMO, we have put lots of effort into iceberg module and it is more stable than before. We should enable the iceberg module by default in case of mvn build.

No

No

After this change, you can build the package including iceberg moudle with this cmd:
`mvn clean install -DskipTests -Pdist`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,6 +60,7 @@
     <module>packaging</module>
     <module>standalone-metastore</module>
     <module>kafka-handler</module>
+    <module>iceberg</module>
   </modules>
   <properties>
     <standalone-metastore.version>4.1.0-SNAPSHOT</standalone-metastore.version>
@@ -2097,12 +2098,6 @@
       <id>itests</id>
       <modules>
         <module>itests</module>
-      </modules>
-    </profile>
-    <profile>
-      <id>iceberg</id>
-      <modules>
-        <module>iceberg</module>
       </modules>
     </profile>
     <profile>
--- a/.github/workflows/docker-GA-images.yml
+++ b/.github/workflows/docker-GA-images.yml
@@ -119,7 +119,7 @@ jobs:
         run: echo "TEZ_VERSION=$(mvn -f "pom.xml" -q help:evaluate -Dexpression=tez.version -DforceStdout)" >> $GITHUB_ENV
 
       - name: Build project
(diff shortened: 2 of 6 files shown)


deniskuzZ (approved):
LGTM +1

pom.xml
```diff
@@ -2099,12 +2100,6 @@
         <module>itests</module>
       </modules>
     </profile>
-    <profile>
```
Indhumathi27, line 2102: We have Iceberg profile in itests module https://github.com/apache/hive/blob/3f6f940af3f60cc28834268e5d7f5612e3b13c30/itests/pom.xml#L501 as well which runs tests in qtest-iceberg module. This also should be enabled by default

zhangbutao: Good catch! I checked the original ticket HIVE-25088.
Fixed in https://github.com/apache/hive/pull/5428/commits/8c1b26b13cd894085c2fab0a5d1b62c74432b757

Please help to check if this change is ok.
Thanks @Indhumathi27

zabetak (commented):
In the past (HIVE-25998), this was breaking the compilation of the project in IntelliJ. Are these problems resolved?

zhangbutao:
> In the past ([HIVE-25998](https://issues.apache.org/jira/browse/HIVE-25998)), this was breaking the compilation of the project in IntelliJ. Are these problems resolved?

@zabetak IMO, before HIVE-25998, we need to shade&patch lots of Apache Iceberg API&Core code by two patched moudle `patched-iceberg-api `& `patched-iceberg-core`, to make our hive-iceberg handler development more faster, and these shaded code can't be read easily by developers unless they manually add the missing dependency in the Intellij idea dependency moule. 

But now, the two patched moudle `patched-iceberg-api `& `patched-iceberg-core` are rarely used, and hive-iceberg handler module is more stable than before. So we can safely enable the iceberg module by default.

zhangbutao:
> It seems that you also need to update the following file: `.github/workflows/docker-GA-images.yml`

Fixed in https://github.com/apache/hive/pull/5428/commits/a2735295654d16d6f9a95b66e63fb2381a4f1426
@zabetak Thanks a lot!

zabetak (approved):

PR #5430, opened 2024-09-03
HIVE-28496: Address CVE-2020-28487 due to 4.20.0 version of vis.js

KiranVelumuri:

https://issues.apache.org/jira/browse/HIVE-28496
Upgrade vis.js from 4.20.0 to address CVE-2020-28487

To address CVE-2020-28487

No. Please see the next comment for screenshots.

Yes. Not attaching the dependency tree output since this does not show up

Using the HTML generated from the template file [QueryProfileTmpl.jamon](https://github.com/apache/hive/blob/master/service/src/jamon/org/apache/hive/tmpl/QueryProfileTmpl.jamon) in the checkWebuiShowGraph test from [TestQueryDisplay.java](https://github.com/apache/hive/blob/master/itests/hive-unit/src/test/java/org/apache/hive/service/cli/session/TestQueryDisplay.java#L164C15-L164C34)


KiranVelumuri:
Link to resource vis-network.min.js from [vis.js](https://visjs.org/) library:
[vis-network.min.js](https://unpkg.com/browse/vis-network@9.1.9/standalone/umd/vis-network.min.js)

## Before (with vis.min.js):
image

image

image


## After (with vis-network.min.js):
image

image

image

KiranVelumuri:
@okumin @zhangbutao 
Could you please review this?

zhangbutao:
> Does this PR introduce any user-facing change?

> Yes. Please see the next comment for  screenshots.

You said this pr introduced user-facing change, but i don't see any difference from the screenshots you proveided.

KiranVelumuri:
> You said this pr introduced user-facing change, but i don't see any difference from the screenshots you provided.
Yes, they are identical even after the library upgrade. I meant in the sense that this change is at the user-facing end. Updated the comment now.

zhangbutao (approved):
+1 LGTM
Thanks for the fix!

KiranVelumuri:
Thank you @zhangbutao for the review!

okumin:
I am checking the content on CDN and that on this PR.

### CDN

```
% curl -o unpkg.js https://unpkg.com/vis-network@9.1.9/standalone/umd/vis-network.min.js
% shasum unpkg.js
021306dceabc3358057f91b84c7dbbe6f48ac12e  unpkg.js
```

### This PR

The checksum didn't match.

```
% shasum service/src/resources/hive-webapps/static/js/vis-network.min.js
173617cf2974080c9857f949ddeb424b50410c87  service/src/resources/hive-webapps/static/js/vis-network.min.js
```

The diff is too big to put here.

```
% diff service/src/resources/hive-webapps/static/js/vis-network.min.js /tmp/vis/unpkg.js
```

I have some questions.

- Is it legal to add the Apache 2.0 header on our own? Looks like, the file includes [a part with dual licenses of Apache 2.0 and MIT](https://github.com/apache/hive/blob/49f66c20e1f29afcd3570a401d8df67b0b186e5c/service/src/resources/hive-webapps/static/js/vis-network.min.js#L18-L42), and [a part with MIT](https://github.com/apache/hive/blob/49f66c20e1f29afcd3570a401d8df67b0b186e5c/service/src/resources/hive-webapps/static/js/vis-network.min.js#L44-L49). I am not confident that our manner is acceptable
- Could you please create `vis-network.min.js` so that it is exactly(or + Apache 2.0 header) the same as the file on the CDN? We have to ensure the new file is definitely safe, but it is hard to review a minified JS codes

KiranVelumuri:
> Thanks! The motivation looks good to me.
> 
> I am checking the content on CDN and that on this PR.
> 
> ### CDN
> ```
> % curl -o unpkg.js https://unpkg.com/vis-network@9.1.9/standalone/umd/vis-network.min.js
> % shasum unpkg.js
> 021306dceabc3358057f91b84c7dbbe6f48ac12e  unpkg.js
> ```
> 
> ### This PR
> The checksum didn't match.
> 
> ```
> % shasum service/src/resources/hive-webapps/static/js/vis-network.min.js
> 173617cf2974080c9857f949ddeb424b50410c87  service/src/resources/hive-webapps/static/js/vis-network.min.js
> ```
> 
> The diff is too big to put here.
> 
> ```
> % diff service/src/resources/hive-webapps/static/js/vis-network.min.js /tmp/vis/unpkg.js
> ```
> 
> I have some questions.
> 
> * Is it legal to add the Apache 2.0 header on our own? Looks like, the file includes [a part with dual licenses of Apache 2.0 and MIT](https://github.com/apache/hive/blob/49f66c20e1f29afcd3570a401d8df67b0b186e5c/service/src/resources/hive-webapps/static/js/vis-network.min.js#L18-L42), and [a part with MIT](https://github.com/apache/hive/blob/49f66c20e1f29afcd3570a401d8df67b0b186e5c/service/src/resources/hive-webapps/static/js/vis-network.min.js#L44-L49). I am not confident that our manner is acceptable
> * Could you please create `vis-network.min.js` so that it is exactly(or + Apache 2.0 header) the same as the file on the CDN? We have to ensure the new file is definitely safe, but it is hard to review a minified JS codes

Thank you for the review.

- I have updated the file `vis-network.min.js` to be the same as from the CDN.
shasum - 021306dceabc3358057f91b84c7dbbe6f48ac12e

- About the licensing part, I have added the license following from earlier file `vis.min.js`, and yes I think it's better to keep the license and the file as is from the CDN. Also, not having the license header should be okay as it is mentioned `vis.js may be distributed under either license.`

okumin:
Thanks. I verified the content of 55ca3d83690154dcb598450dbd02319491f91bcd is precisely the same as that in the internet.

```
$ shasum service/src/resources/hive-webapps/static/js/vis-network.min.js
021306dceabc3358057f91b84c7dbbe6f48ac12e  service/src/resources/hive-webapps/static/js/vis-network.min.js
$ shasum /tmp/vis/unpkg.js                                              
021306dceabc3358057f91b84c7dbbe6f48ac12e  /tmp/vis/unpkg.js
$ diff service/src/resources/hive-webapps/static/js/vis-network.min.js /tmp/vis/unpkg.js
$
```

I also confirmed the revision is likely to work in the same way as 4.0.0.

image

Configuration I used.

```
  <property>
    <name>hive.server2.webui.explain.output</name>
    <value>true</value>
  </property>
  <property>
    <name>hive.server2.webui.show.graph</name>
    <value>true</value>
  </property>
```

CI failed but I believe it is unrelated.

https://ci.hive.apache.org/blue/rest/organizations/jenkins/pipelines/hive-precommit/branches/PR-5430/runs/3/nodes/455/log/?start=0

As for the license, the MIT part about `Hammer.JS` shall be mentioned here? To be honest, I am not confident with the rule.

https://github.com/apache/hive/blob/master/LICENSE

okumin (approved):
I approved. I would like someone to check the license point as well.

KiranVelumuri:
> I approved. I would like someone to check the license point as well.

@ayushtkn Could you please comment on the license point? Thank you.

KiranVelumuri:
Both [Apache Hive](https://github.com/apache/hive/blob/master/LICENSE) and [vis.js](https://github.com/visjs/vis-network/blob/master/LICENSE-APACHE-2.0) are licensed under Apache License 2.0. In [Applying the Apache license, version 2.0](https://www.apache.org/legal/apply-license.html#:~:text=Each%20original%20source%20document%20(code%20and%20documentation%2C%20but%20not%20the%20LICENSE%20and%20NOTICE%20files)%20should%20include%20a%20short%20license%20header%20at%20the%20top.), it is mentioned that each source file must include a header.

@okumin @zhangbutao
So, we should add the Apache LICENSE 2.0 header at the top of vis-network.min.js to comply with the standards. Please reply/react so that I will apply the header to the current change and update the commit.

okumin:
I assume you are not an author of vis.js, and the rule is unlikely applied if I undeistand correctly.
https://www.apache.org/legal/src-headers.html#3party

zabetak:
When we include code not owned by the contributor it is best to leave the file as it is (no modifications).
Since the file is not developed by us it may be nice to add a mention in LICENSE although not strictly necessary since it is licensed under AL2.

More general, it is best to avoid minified javascript files in Apache source releases but this is a bit of a grey area so we can continue as is for the moment.

KiranVelumuri:
@zabetak Thank you, we will proceed as is.
@okumin @zhangbutao I think no further changes are needed, please help in merging. Thank you.

okumin:
@zabetak Just a question from my curiosity. [L44- of the file is not licensed as AL2](https://github.com/apache/hive/blob/49f66c20e1f29afcd3570a401d8df67b0b186e5c/service/src/resources/hive-webapps/static/js/vis-network.min.js#L44-L51), as it is MIT. Don't we need special care in this case? Additionally, if it was GPL or something, would we have to take care of it?

I think Butao can merge this PR as a committer.

zabetak:
@okumin Good catch! I haven't noticed that portions of the file (L44-49) are licensed differently. Yes, we need to handle such cases carefully especially when we are dealing with licenses that are not in [category-a](https://www.apache.org/legal/resolved.html#category-a). Since it seems that portion of the file is only under the MIT license then we should include both a copy of the [license](https://github.com/naver/hammer.js/blob/master/LICENSE.md) in our source code distribution and an appropriate entry in our main LICENSE file.

In general, whenever there is a doubt about how to handle a licensing case, the correct reaction is to contact the Apache legal team (e.g., by opening a LEGAL jira ticket) and request guidance.

service/src/jamon/org/apache/hive/tmpl/QueryProfileTmpl.jamon
```diff
@@ -57,7 +57,7 @@ private boolean showStats = false;
       </%if>
     </%if>
     <%if showGraph %>
-      <script type="text/javascript" src="/static/js/vis.min.js"></script>
+      <script type="text/javascript" src="/static/js/vis-network.min.js"></script>
```
zabetak, line 60: Do we need to include the `vis-network.min.js` file in our source and binary distribution? Can't we simply link to it without copying its source code in this repo?
```
<script type="text/javascript" src="https://unpkg.com/vis-network@9.1.9/standalone/umd/vis-network.min.js"></script>
```

zabetak (changes requested):

service/src/jamon/org/apache/hive/tmpl/QueryProfileTmpl.jamon
KiranVelumuri: I included the file to follow the same manner as earlier.
Yes, we can have it this way as well. I have validated with this way of `src` and it works fine. Please let me know if we can proceed this way.

zabetak: I don't know the history behind the original decision to bundle it inside but for me this approach is better in every aspect.

zhangbutao: I'm not very clear about license. But the latest change is good to me. :)

zhangbutao (approved):

okumin:
I confirmed it would also work on my machine
image

zabetak (approved):

PR #5432, opened 2024-09-04
[HIVE-28500] Update alterSchemaVersion in ObjectStore.java

kimjye-dev:
Only the alterSchemaVersion method of ObjectStore does not perform rollbackTranscation() if committed is false in finally.

so if commited = false, commitTransaction -> rollbackTranscation

https://issues.apache.org/jira/browse/HIVE-28500

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
@@ -13369,7 +13369,7 @@ public class ObjectStore implements RawStore, Configurable {
       committed = commitTransaction();
     } finally {
       if (!committed) {
-        commitTransaction();
+        rollbackTransaction();
       }
     }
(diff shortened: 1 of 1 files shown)


InvisibleProgrammer:
Hi, 

For the first sight it looks good. But it opens some questions: 
What is the current behaviour if the call cannot be committed? Is the method throws an exception? 
After that change, in case it have to be rolled back, from the caller's standpoint, it looks like a successful call: I call the method, no errors given, no return value. So I assume the call was successful. 
Shouldn't we throw at least a meta exception (if we don't want a breaking change) or modify the interface to return with a true/false value that tells if the call was successful (like in `alterDatabase`)? 

Also, can you add a test for a change? `TestObjectStoreSchemaMethods` can be a proper place to do that.

mr3project:
Calling commitTransaction() instead of rollbackTransaction() seems just like a minor error, and I think we don't need a new test case for this change.

deniskuzZ (approved):
LGTM +1

PR #5433, opened 2024-09-06
HIVE-28503: Wrong results(NULL) when string concat operation with || operator for ORC file format when vectorization enabled

maheshrajus:
Set the output vector flag correctly about whether it has NULL or NOT NULL in case of all inputs values are NOT NULL.
StringGroupConcatColCol->evaluate() method:

if (inV1.noNulls && !inV2.noNulls) {    //if any one input has NULL, then output should be NULL. 
     outV.noNulls = false;                      // setting this flag false as all values in this are NULLs
 --- code --- 
}  else if (!inV1.noNulls && inV2.noNulls) {    //if any one input has NULL, then output should be NULL.
 // propagate nulls
   outV.noNulls = false;                  // setting this flag false as all values in this are NULLs
}  else if (!inV1.noNulls && !inV2.noNulls) {   // if two inputs are NULL, then output should be NULL.
   // propagate nulls
   outV.noNulls = false;           //setting this flag false as all values in this are NULLs
    --- code --- 
}  else  {                       // there are no nulls in either input vector
    outV.noNulls = true;    // this has to be set true, as there are no NULL values, this check is missed currently.
    // perform data operation
   --- code --- 
}

While doing concat() operation, In StringGroupConcatColCol class, if input batch vector has mixed of NULL and NOT NULL values of inputs then we are not setting output vector batch flags related to NULL and NOT NULLS correctly . Each value in the vector has the flag whether it is NULL or NOT NULL. But  here we are not setting correctly the whole output vector flag (outV.noNulls). 

No

No

Existing Q files. Cant add new tests as this was able to reproduce only in cluster with more input records. Tested in cluster and verified.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/StringGroupConcatColCol.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/StringGroupConcatColCol.java
@@ -320,7 +320,8 @@ public class StringGroupConcatColCol extends VectorExpression {
       /*
        * Do careful maintenance of the outputColVector.noNulls flag.
        */
-
+      // propagate not nulls flag.
+      outV.noNulls = true;
       // perform data operation
       if (inV1.isRepeating && inV2.isRepeating) {
 


ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/StringGroupConcatColCol.java
```diff
@@ -320,7 +320,8 @@ public void evaluate(VectorizedRowBatch batch) throws HiveException {
       /*
        * Do careful maintenance of the outputColVector.noNulls flag.
        */
-
+      // propagate not nulls flag.
```
zratkai, line 323: Above lines says:
     // there are no nulls in either input vector
      /*
       * Do careful maintenance of the outputColVector.noNulls flag.
       */
       
  So the else branch says there are no nulls, it seems strange to then says noNulls = true

zratkai (changes requested):
Please add a meaningful description, explanation and not a java code in the PR text, since it is hard to understand without the context.

Also try to create a test. From the description I do not understand why it only happens in a cluster, and can not be reproduced in a test.

InvisibleProgrammer:
`ColumnVector` initialises `noNulls` as true. I assume you tested it with a sample data received from a special case of a customer. 
I wonder, is there a chance that the original problem has other vector expressions and maybe one of the previous steps sets noNulls to false. 

About testing that path: There is a test in `ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestVectorStringExpressions.java` that looks like testing that code path: `testColConcatCol`

```java
// no nulls, not repeating
    batch = makeStringBatch2In1Out();
    batch.cols[0].noNulls = true;
    batch.cols[1].noNulls = true;
    expr.evaluate(batch);
    outCol = (BytesColumnVector) batch.cols[2];

   ... 

    Assert.assertTrue(outCol.noNulls);
    Assert.assertFalse(outCol.isRepeating);
```

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/StringGroupConcatColCol.java
maheshrajus: @zratkai there are two vectors. one is input vector which contains input strings in our case. And the second vector is output which contains the result and flags related to nulls/not nulls. So here the case is if input vector contains mix of NULLs and NOT NULLs then previous steps sets noNulls to false(NULL case) and same flag carried to "else" case[NOT NULL] and we are not setting the **outV.noNulls** flag to true.
I just modified the code to set the **outV.noNulls** flag true in else case and i verified in cluster. Fix is working fine. 
I am trying to add the qtest but i am not able to reproduce the issue locally with q file.

zratkai (commented):
Please check my answer!

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/StringGroupConcatColCol.java
```diff
@@ -320,7 +320,8 @@ public void evaluate(VectorizedRowBatch batch) throws HiveException {
...
+      outV.noNulls = true;
```
mdayakar, line 324: @maheshrajus this assignment was there earlier and got removed as a part of [HIVE-18622](https://issues.apache.org/jira/browse/HIVE-18622). Now you are adding it again so please check carefully why it was removed as a part of [HIVE-18622](https://issues.apache.org/jira/browse/HIVE-18622).

maheshrajus: @mdayakar thanks for review. I will check reason behind it.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/StringGroupConcatColCol.java
maheshrajus: @zratkai  yes, but with simple data insert issue is not able to reproduce.

PR #5434, opened 2024-09-06
HIVE-28502: Refactor method names that start with capital letters in …

difin:
…PasswdAuthenticationProvider class

Rename PasswdAuthenticationProvider's 'Authenticate' method to 'authenticate' according to Java naming convention.

To fix a method name that is wrong according to Java naming conventions.

No

No

Existing unit tests and query tests.


zhangbutao:
Looks good! Need to rebase to fix the conflicts.

BTW, I think if we can do checkstyle in the ci pipeline, then we will avoid lots of similar code style issue.
But hive never use the checkstyle except the hive-iceberg handler module. :(   see HIVE-28499

deniskuzZ (approved):
LGTM +1

PR #5438, opened 2024-09-10
HIVE-28515: Iceberg: Concurrent queries fail during commit with ValidationException.

ayushtkn:

Don't fail when the retry strategy is write_conflict in case of a commit failure, but rather retry

To avoid query failures where multiple queries are running concurrently on the same table & writes on similar set of files

Yes, In case of concurrent queries with conflicting files, the query doesn't fail but rather retries. 

No

UT

### DIFF
--- a/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestOptimisticRetry.java
+++ b/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestOptimisticRetry.java
@@ -74,6 +74,27 @@ public class TestOptimisticRetry extends HiveIcebergStorageHandlerWithEngineBase
 
   }
 
+  @Test
+  public void testConcurrentOverwriteAndUpdate() {
+    testTables.createTable(shell, "customers", HiveIcebergStorageHandlerTestUtils.CUSTOMER_SCHEMA,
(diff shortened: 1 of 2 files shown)


iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestOptimisticRetry.java
```diff
@@ -74,6 +74,27 @@ public void testConcurrentOverlappingUpdates() {
 
   }
 
+  @Test
```
simhadri-g, line 77: Would recommend moving this to TestConflictingDataFiles.java, because phaser code in @Before and @After are not present here.

simhadri-g: Rest of the code LGTM

ayushtkn: I changed as we discussed

simhadri-g: Thanks!

simhadri-g (approved):
LGTM pending green test run.

PR #1126, opened 2020-06-16
HIVE-23598: Add option to rewrite NTILE and RANK to sketch functions

kgyrtkirk:

PR #5444, opened 2024-09-12
HIVE-28520: Upgrade to datasketches 2.0.0

zhangbutao:

mvn dependency:tree

No

Yes. 
[dependency_upgrade.txt](https://github.com/user-attachments/files/16972308/dependency_upgrade.txt)

Existing Uts.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -219,7 +219,7 @@
     <rs-api.version>2.0.1</rs-api.version>
     <json-path.version>2.9.0</json-path.version>
     <janino.version>3.1.12</janino.version>
-    <datasketches.version>1.2.0</datasketches.version>
+    <datasketches.version>2.0.0</datasketches.version>
     <spotbugs.version>4.0.3</spotbugs.version>
     <validation-api.version>1.1.0.Final</validation-api.version>
     <aws-secretsmanager-caching.version>1.0.1</aws-secretsmanager-caching.version>
--- a/standalone-metastore/pom.xml
+++ b/standalone-metastore/pom.xml
@@ -64,7 +64,7 @@
     <commons-lang3.version>3.12.0</commons-lang3.version>
     <commons-logging.version>1.1.3</commons-logging.version>
     <commons-dbcp2.version>2.12.0</commons-dbcp2.version>
-    <datasketches.version>1.2.0</datasketches.version>
+    <datasketches.version>2.0.0</datasketches.version>
     <datanucleus-api-jdo.version>5.2.8</datanucleus-api-jdo.version>
     <datanucleus-core.version>5.2.10</datanucleus-core.version>
     <datanucleus-jdo.version>3.2.0-release</datanucleus-jdo.version>


zhangbutao:
hi @okumin I think the 2.0.0 datasketchs can fix the issue you have mentioned: https://github.com/apache/hive/pull/5144#issuecomment-2014130113

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/common/histogram/KllHistogramEstimator.java
```diff
@@ -30,11 +30,11 @@ public class KllHistogramEstimator {
   private final KllFloatsSketch kll;
 
   KllHistogramEstimator() {
-    this.kll = new KllFloatsSketch();
+    this.kll = KllFloatsSketch.newHeapInstance();
```
zhangbutao, line 33: In this commit: https://github.com/apache/datasketches-java/commit/f3d80294e1e34278d1cd7a46071929ae267d633c#diff-255436ca709c4ad22450438c6b07cb84771cbb9694ebaec9b15ab201f8f03673

KllFloatsSketch was changed to `abstract`. Use `KllFloatsSketch.newHeapInstance()` to get the instance.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/common/histogram/kll/KllUtils.java
```diff
@@ -100,7 +101,7 @@ public static KllFloatsSketch deserializeKll(final byte[] buf) {
    * @return the length of the given KLL sketch according to the given java data model
    */
   public static int lengthFor(JavaDataModel model, KllFloatsSketch kll) {
-    return model == null ? KllFloatsSketch.getMaxSerializedSizeBytes(kll.getK(), kll.getN())
+    return model == null ? KllFloatsSketch.getMaxSerializedSizeBytes(kll.getK(), kll.getN(), KllSketch.SketchType.FLOATS_SKETCH, false)
```
zhangbutao, line 104: In this commit: https://github.com/apache/datasketches-java/commit/60cbfaabd43ab33eb542148eb4dbd8fbef98a264#diff-dacbef77551b8e94e3095811486b859725948654627b5141625c453f0ecad774

`KllFloatsSketch::getMaxSerializedSizeBytes` added two extra parameters:
method_change

I set the updatable to false to keep it the same as before. If we want optimize this to set true or false, we can make a configuration here in the future.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/stats/FilterSelectivityEstimator.java
```diff
@@ -511,7 +511,7 @@ private static double rangedSelectivity(KllFloatsSketch kll, float val1, float v
    * @return the selectivity of a predicate "column &gt; value" in the range [0, 1]
    */
   public static double greaterThanSelectivity(KllFloatsSketch kll, float value) {
-    float max = kll.getMaxValue();
+    float max = kll.getMaxItem();
```
zhangbutao, line 514: https://github.com/apache/datasketches-java/commit/ce9bed7f753ddaa27fc1162f88bae350af08231b#diff-3a7fd0aa04e700d6abbaa5025d434d4098a80d60699b192dc2995a7201ff3aea This commit changed the method name.

okumin:
For myself, [datasketches-hive 1.2.0](https://github.com/apache/datasketches-hive/blob/1.2.0/pom.xml#L86) uses [datasketches-java 3.1.0](https://github.com/apache/datasketches-java/tree/3.1.0). [datasketches-hive 2.0.0](https://github.com/apache/datasketches-hive/blob/2.0.0/pom.xml#L86) uses [datasketches-java 6.0.0](https://github.com/apache/datasketches-java/tree/6.0.0).

zhangbutao:
> For myself, [datasketches-hive 1.2.0](https://github.com/apache/datasketches-hive/blob/1.2.0/pom.xml#L86) uses [datasketches-java 3.1.0](https://github.com/apache/datasketches-java/tree/3.1.0). [datasketches-hive 2.0.0](https://github.com/apache/datasketches-hive/blob/2.0.0/pom.xml#L86) uses [datasketches-java 6.0.0](https://github.com/apache/datasketches-java/tree/6.0.0).

@okumin Yes, actually, this change is mainly at datasketches-java: upgarde datasketches-java from 3.1.0 to 6.0.0

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/common/histogram/KllHistogramEstimator.java
okumin: dataskatches-hive 2.0 can read the data written by 1.2?

okumin (approved):
I confirmed the warning of `exec.FunctionRegistry: UDF Class org.apache.hive.org.apache.datasketches.hive` would disappear. I think this PR is ok if the new datasketches-java can read existing sketches persisted in metastore or somewhere.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/stats/FilterSelectivityEstimator.java
zhangbutao: I didn't do much research, but I think it is ok. 
To be honest, it's hard to explore the reasons for some commits of apache datasketches. Apache datasketches often doesn't give more details about its commit. :(

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/common/histogram/KllHistogramEstimator.java
zhangbutao: > dataskatches-hive 2.0 can read the data written by 1.2?

I think yes.

zhangbutao:
@asolimando Could you also please take a look? The code changes is related to your last commit of histogram-based stats HIVE-26221
IMO, the kll stats  is just a rough estimate, so after the datasketch upgrading,  it is reasonable for the changes related Uts?

asolimando:
> @asolimando Could you also please take a look? The code changes is related to your last commit of histogram-based stats [HIVE-26221](https://issues.apache.org/jira/browse/HIVE-26221) IMO, the kll stats is just a rough estimate, so after the datasketch upgrading, it is reasonable for the changes related Uts?

@zhangbutao, I have skimmed the changes and although I agree that KLL stats are estimates, the specific unit-tests I have added for HIVE-26221 were hand-crafted and double checked manually on a small sample, so I wouldn't expect changes there (especially the estimation of rows surviving a particular filter is sometimes "big").

My suggestion would be to double check if the initialization changes haven't impacted the way the KLL data sketch is computed, and to double check manually with small samples (even outside Hive, with just the library in some Java code).

zhangbutao:
> > @asolimando Could you also please take a look? The code changes is related to your last commit of histogram-based stats [HIVE-26221](https://issues.apache.org/jira/browse/HIVE-26221) IMO, the kll stats is just a rough estimate, so after the datasketch upgrading, it is reasonable for the changes related Uts?
> 
> @zhangbutao, I have skimmed the changes and although I agree that KLL stats are estimates, the specific unit-tests I have added for [HIVE-26221](https://issues.apache.org/jira/browse/HIVE-26221) were hand-crafted and double checked manually on a small sample, so I wouldn't expect changes there (especially the estimation of rows surviving a particular filter is sometimes "big").
> 
> My suggestion would be to double check if the initialization changes haven't impacted the way the KLL data sketch is computed, and to double check manually with small samples (even outside Hive, with just the library in some Java code).

@asolimando Thanks for you suggestion. I will double check these tests changes.

okumin:
I found the diff of stats_histogram.q would disappear with this change. There could be some more changes on datasketches-java side.

```
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/ddl/ShowUtils.java b/ql/src/java/org/apache/hadoop/hive/ql/ddl/ShowUtils.java
index 0d52c85289..ec197a3264 100644
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/ShowUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/ShowUtils.java
@@ -23,6 +23,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.datasketches.kll.KllFloatsSketch;
 import org.apache.datasketches.memory.Memory;
+import org.apache.datasketches.quantilescommon.QuantileSearchCriteria;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -257,7 +258,7 @@ private static String convertHistogram(byte[] buffer, ColumnStatisticsData._Fiel
     final KllFloatsSketch kll = KllFloatsSketch.heapify(Memory.wrap(buffer));
     // to keep the visualization compact, we print only the quartiles (Q1, Q2 and Q3),
     // as min and max are displayed as separate statistics already
-    final float[] quantiles = kll.getQuantiles(new double[]{ 0.25, 0.5, 0.75 });
+    final float[] quantiles = kll.getQuantiles(new double[]{ 0.25, 0.5, 0.75 }, QuantileSearchCriteria.EXCLUSIVE);
```

ql/src/test/results/clientpositive/llap/compute_kll_sketch.q.out
```diff
@@ -7,22 +7,25 @@ POSTHOOK: type: QUERY
...
-   Empty                : false
-   Estimation Mode      : true
-   Levels               : 6
-   Sorted               : false
-   Buffer Capacity Items: 547
-   Retained Items       : 535
-   Storage Bytes        : 2192
-   Min Value            : -64.0
-   Max Value            : 62.0
+### KllKllFloatsSketch Summary:
```
asolimando, line 10: The name seems redundant, I guess `KllFloatSketch Summary` is better if we need to change the existing name (even if I don't see much interest, what type of KLL sketch we use it's an internal detail I'd say).

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/stats/TestFilterSelectivityEstimator.java
```diff
@@ -159,7 +159,7 @@ public void testIsHistogramAvailableWhenEmptyArray() {
 
   @Test
   public void testLessThanSelectivity() {
-    Assert.assertEquals(0.6153846153846154, lessThanSelectivity(KLL, 3), DELTA);
```
asolimando, line 162: I am still not convinced that those values should change, from what I understood @okumin found a way to have the same results? Have you applied his suggestions on the way we compute and query the data sketch?

zhangbutao: Fixed in https://github.com/apache/hive/pull/5444/commits/ab110b6d7ce29b2a0bc916cf3dd62634fdf6df23

I think all histogram stats related tests are changed back now. :)

zhangbutao: But i still can not figure why skecth functions Qtests changed. Such as `sketches_materialized_view_cume_dist.q.out`.

I noticed that https://github.com/apache/hive/pull/1126#discussion_r442360252 also did similar change. 
Maybe @kgyrtkirk you can give some thread? Thanks.

ql/src/test/results/clientpositive/llap/compute_kll_sketch.q.out
zhangbutao: I think this needs to change related codes in apache/datasketches. :)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/common/histogram/KllHistogramEstimator.java
asolimando: It's crucial to make sure backward compatibility, especially because stats in metastore are meant to be interoperable between engines, so if we are breaking compatibility with other engines using an older datasketch version, we must highlight it clearly.

I suggest to just do a quick test locally, compute with 1.2, print the quantiles, write to a file, load it with the current version and display the same. It can be just a blank java project with a main class, you don't even need Hive to test that, as we rely on the serde of the library.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/stats/TestFilterSelectivityEstimator.java
zhangbutao: I have found the root cause: some skecth functions have multiple `evaluate` method, but in Hive we always get the first `evaluate` method, see:
https://github.com/apache/hive/blob/548990dfad78b3d89a334c875d8a6708ef475e88/ql/src/java/org/apache/hadoop/hive/ql/exec/DataSketchesFunctions.java#L288-L297

Here i gave a example: In this commit https://github.com/apache/datasketches-hive/commit/b6c4d01ff9539d2aff520569277c14d479695bb0#diff-722be040808b3a7ef18a0e86ced7e686fcf553002f82d8332d02d29f28b265a9 , it added another `evaluate `method with the new added param `QuantileSearchCriteria.INCLUSIVE`, and then, some Qtest which uses `cdf` will change its result.

And some other skecth functions may have more `evaluate ` method, we also choose the first one.

IMO, we can not find a good way to choose the best `evaluate` method, what i can do is regenerate the skecth functions Qtests outout. :(

Do you have any other good ideas? @asolimando

okumin: How about modifying datasketches-java so that it is consistent with the previous version?

asolimando: @okumin, if they are open to that change, we could definitely try and that would solve our problems. But since the behavior changed in a major release, technically it's something that could be intended and expected, although it makes our life more complicated.

@zhangbutao sorry I missed your previous message! What about moving the evaluate method into a map which is filled when invoking `register` for each skect function we support? It's less elegant maybe than the existing, but we would regain full control on what we invoke at eval time and we are anyway already specifying many info in the register function, so it does seem to be coherent with the current design.

What do you guys think?

okumin: > How about modifying datasketches-java so that it is consistent with the previous version?

Sorry, I intended modifying **datasketches-hive**, not datasketches-java. The Hive community may have some control over the library.

zhangbutao: > What about moving the evaluate method into a map which is filled when invoking register for each skect function we support?

@asolimando But we don't know what `evaluate `methods of sketch funtion we should support. In fact, we can run all `evaluate` methods of each function, but we just pick the first one & we can not determine which `evaluate ` method is better or is what we need.

> The Hive community may have some control over the library datasketches-hive.

@okumin I think this is too hard :(. I don't think we can do anything about the datasketches. Unless we backport the whole base code of datasketches-hive into Hive repo, but this may cause some codes inconsistency&conflict with datasketches in the futrue.

okumin: Just let me clarify the problem. In the case of `GetCdfUDF`, originally, we provided `(ByteWritable, Float...) -> List<Double>`. Datasketches Hive 2.0 provides `(ByteWritable, Float...) -> List<Double>` and `(ByteWritable, Boolean, Float...) -> List<Double>`, and the first one is now an alias of `(ByteWritable, true, Float...) -> List<Double>`. While using Datasketches Hive 1.2, the method with 2-arity was working as `(ByteWritable, false, Float...) -> List<Double>`. So, the original query unit test failed as the 2-arity method started using the INCLUSIVE mode.

One more, I guess we don't have to keep 100% compatibility of sketch UDFs because it returns an approximate result. However, we will encounter incompatibilities when the sketches, not the final result, are directly stored as binary. Information stored in HMS is the primary example. Do we have another case to directly store sketches on the system side? e.g. Materialized View

asolimando: > I have found the root cause: some skecth functions have multiple `evaluate` method, but in Hive we always get the first `evaluate` method, see:
> 
> https://github.com/apache/hive/blob/548990dfad78b3d89a334c875d8a6708ef475e88/ql/src/java/org/apache/hadoop/hive/ql/exec/DataSketchesFunctions.java#L288-L297
> 
> Here i gave a example: In this commit [apache/datasketches-hive@b6c4d01#diff-722be040808b3a7ef18a0e86ced7e686fcf553002f82d8332d02d29f28b265a9](https://github.com/apache/datasketches-hive/commit/b6c4d01ff9539d2aff520569277c14d479695bb0#diff-722be040808b3a7ef18a0e86ced7e686fcf553002f82d8332d02d29f28b265a9) , it added another `evaluate `method with the new added param `QuantileSearchCriteria.INCLUSIVE`, and then, some Qtest which uses `cdf` will change its result.
> 
> And some other skecth functions may have more `evaluate ` method, we also choose the first one.
> 
> IMO, we can not find a good way to choose the best `evaluate` method, what i can do is regenerate the skecth functions Qtests outout. :(
> 
> Do you have any other good ideas? @asolimando

Actually the method you cite only infers the return type, it's not exactly what we want, although the root cause it's something similar I guess, someplace else.

I have traced some functions finding by reflection the correct `evaluate` method for UDF and UADF, respectively https://github.com/apache/hive/blob/master/udf/src/java/org/apache/hadoop/hive/ql/exec/UDFMethodResolver.java#L49
and https://github.com/apache/hive/blob/master/udf/src/java/org/apache/hadoop/hive/ql/exec/DefaultUDAFEvaluatorResolver.java#L57 

The problem is not in those methods, as they are receiving the actual function inputs and just matching on them, what we would need to do, is to intercept where this UDF/UDAF are written in the AST/plan and replace `$udf($binary, $float)` with `$udf($binary, false, $float)` so that the right evaluate function is then invoked.

I tried to track down places where the sketches are introduced (e.g., rules like https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveRewriteToDataSketchesRules.java) but without much luck so far.

zhangbutao: > what we would need to do, is to intercept where this UDF/UDAF are written in the AST/plan and replace $udf($binary, $float) with $udf($binary, false, $float) so that the right evaluate function is then invoked.

This seems like a good idea. :)   I will try to explore this way, too.

Because incompatible datasketches changes lead to a lot of work, it seems difficult to control datasketches-hive.  I hope this PR is the last one to do this work.  :(

okumin: Can I expect this is still in progress?

zhangbutao: I've been busy with other work stuff recently.  Will have a better look at this after a while. :)

deniskuzZ:
what's pending here?

zhangbutao:
> what's pending here?

Due to the changes in the underlying datasketches interface, We are unable to select a fixed sketch strategy on the Hive side, which leads to many QTests needing to be changed.

Trying to find a way on the Hive side to avoid modifying qtests & keeping backward compatibility each time, but so far haven't had any good luck.

vrozov:
@deniskuzZ @zhangbutao This PR is critical for Apache Spark to be able to upgrade to Hive 4.x.

deniskuzZ:
@vrozov, noted. 
@zhangbutao, anything pending here?  Can we merge?

deniskuzZ (commented):
LGTM, however, we should test the backward compatibility

zhangbutao:
> @vrozov, noted. @zhangbutao, anything pending here? Can we merge?
> LGTM, however, we should test the backward compatibility

To be honest, I really haven't found a good solution to address the changes in the qtests output results caused by the Sketch library code.  Perhaps I'm a little too lazy to continue researching this issue anymore. :)

About this PR, IMO, I think there shouldn't be any compatibility issues. The affected are some estimation functions(cume_dist & ds_kll_sketch & ntile ). These functions are only used by users in SQL execution, and the results of these estimation functions will not be stored in HMS. So, at least from the HMS perspective, there won't be any compatibility issues.

However, the user might need to modify the SQL execution result if they . But I think such incompatibility changes can be accepted by users because they need to understand that this function is an estimated calculation and the error variation is reasonable.

BTW, If we want to inform users that the results of using this estimation function may vary**, we might need to add a documentation note in the next release version to remind users that the results of using these functions may change in the new version.**

At last, In the long run, we might need to port these function codes from the sketches library to the hive library so that hive can fully handle any changes to these functions.


Well, saying all that might just be an excuse I made up to avoid continuing to study this issue. :)

CC  @okumin @asolimando

zhangbutao:
Don't have any further effective ways to optimize this PR.
I will merge this change next week.

zhangbutao:
Merged!
Thanks @okumin @asolimando @deniskuzZ for the review!

PR #5447, opened 2024-09-13
Feature/Improve performance when drop partition or drop table

Leonidas963:
Reduce getDnsPath times to improve performance

When the number of partitions is large, such as 200,000, serial traversing and obtaining wh.getDnsPath will take a huge amount of time, and performance optimization can be done for this place

No

No

By ut

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -3221,11 +3221,11 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
         for (String partName : partNames) {
           String pathString = partitionLocations.get(partName);
           if (pathString != null) {
-            Path partPath = wh.getDnsPath(new Path(pathString));
             // Double check here. Maybe Warehouse.getDnsPath revealed relationship between the
             // path objects
             if (tableDnsPath == null ||
-                !FileUtils.isSubdirectory(tableDnsPath, partPath.toString())) {
+                !FileUtils.isSubdirectory(tableDnsPath, pathString)) {
+              Path partPath = wh.getDnsPath(new Path(pathString));
               if (!wh.isWritable(partPath.getParent())) {
                 throw new MetaException("Table metadata not deleted since the partition "
                     + partName + " has parent location " + partPath.getParent()
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
@@ -38,6 +38,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.lang.reflect.*;
+
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 
 import com.google.common.collect.Sets;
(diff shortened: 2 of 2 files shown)


huiboliu2020:
When is the tableDnsPath empty ?   And why can this Improve performance?

zhangbutao:
Could you please provide more detaill about this optimization?
For example, if you guess `wh.getDnsPath` will take much time when dropping partition, you can do a `jstack ` on HMS java pid, and you should see the `wh.getDnsPath`  thread is in a `WAITING `or a `BLOCK `state for a long time.

So ,could you do a jstack on HMS to help we know more about this?
Thanks.

Leonidas963:
> Could you please provide more detaill about this optimization? For example, if you guess `wh.getDnsPath` will take much time when dropping partition, you can do a `jstack ` on HMS java pid, and you should see the `wh.getDnsPath` thread is in a `WAITING `or a `BLOCK `state for a long time.
> 
> So ,could you do a jstack on HMS to help we know more about this? Thanks.

It is executed in serial, which should not show that the state is waiting or block. I have a time-consuming printing here, which can show that there is a performance problem in this piece.
I think that's a good indication of the problem

get_dns_path_cost

Leonidas963:
> When is the tableDnsPath empty ? And why can this Improve performance?

When table type is view the tableDnsPath will be null, By reduce times to get filesytem client.

zhangbutao:
@Leonidas963 I checked https://issues.apache.org/jira/browse/HIVE-24838 which wanted to reduce the object stores fs call. I think it maybe can also be used to hdfs filesystem.
HIVE-24838 was merged to Hive 4. Can you try to test HIVE-24838  in your hadoop cluster? and set the metastore property as following:  `hive.blobstore.supported.schemes=hdfs,s3,s3a,s3n`

BTW, i will check your fix later.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -3221,11 +3221,11 @@ private List<Path> dropPartitionsAndGetLocations(RawStore ms, String catName, St
         for (String partName : partNames) {
           String pathString = partitionLocations.get(partName);
           if (pathString != null) {
-            Path partPath = wh.getDnsPath(new Path(pathString));
             // Double check here. Maybe Warehouse.getDnsPath revealed relationship between the
             // path objects
             if (tableDnsPath == null ||
-                !FileUtils.isSubdirectory(tableDnsPath, partPath.toString())) {
+                !FileUtils.isSubdirectory(tableDnsPath, pathString)) {
```
zhangbutao, line 3227: After checking, i think your change has no meaning for the current branch(Hive4 & master branch).

Because HIVE-19783 in Hive4 changed some codes logic about fetching partition location. See https://github.com/apache/hive/commit/e36f6e4fbda354f33ba9cef6cf25e5573c78d618#diff-cf3f64ef2e2a3a8ee2867032c2d3a1ac0b567a60f82b00299cc0dc99e00888a8 , HIVE-19783 did the optimization to only retrieve partitions locations instead of all objects of partitions by `ObjectStore::getPartitionLocations`.  And this method only **return location string only outside the location**, see comment https://issues.apache.org/jira/browse/HIVE-19783?focusedCommentId=16512315&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16512315.

As a result, after HIVE-19783, if partition location is subdirectory of table path, the variable `pathString` will always null, and if partition location is different from table location, this code snippet will take effect.

I guess your parition location is subdirectory of table path. e.g. table location = hdfs://ns1/db/testtbl, and partition location = hdfs://ns1/db/testtbl/dt=2024. If so, after the optimization of HIVE-19783, ` if (pathString != null) ` will false, then you won't see any perfomance issue.


**But I guess you are using Hive3? You change may be useful for Hive3.**
BTW, it would be if you try to use this optimization HIVE-24838. It can also optimize some fs call.

PR #5450, opened 2024-09-16
HIVE-28282: Addendum: Merge-Insert with reserved keywords as column names

deniskuzZ:

Merge-Insert with reserved keywords as column names

Fix

No

No

Qtest

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/sql/MultiInsertSqlGenerator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/sql/MultiInsertSqlGenerator.java
@@ -27,6 +27,7 @@ import org.apache.hadoop.hive.ql.metadata.Table;
 import java.util.ArrayDeque;
 import java.util.List;
 import java.util.Deque;
+import java.util.function.Function;
 
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/sql/MultiInsertSqlGenerator.java
```diff
@@ -155,22 +156,22 @@ public void appendPartColsOfTargetTableWithComma(String alias) {
...
-
-  public void appendCols(List<FieldSchema> columns) {
-    appendCols(columns, null, null);
+  
+  public <T> void appendCols(List<T> columns, Function<T, String> func) {
+    appendCols(columns, null, null, func);
   }
 
-  public void appendCols(List<FieldSchema> columns, String alias, String prefix) {
+  public <T> void appendCols(List<T> columns, String alias, String prefix, Function<T, String> func) {
```
kasakrisz, line 174: nit. rename `func` to `stringConverter`

kasakrisz (approved):
+1

kasakrisz:
Please add the original commit message of [HIVE-28282](https://issues.apache.org/jira/browse/HIVE-28282) too.

ql/src/java/org/apache/hadoop/hive/ql/parse/rewrite/sql/MultiInsertSqlGenerator.java
deniskuzZ: renamed

deniskuzZ:
> Please add the original commit message of [HIVE-28282](https://issues.apache.org/jira/browse/HIVE-28282) too.

I am not sure where I should add it?

PR #5451, opened 2024-09-17
HIVE-28505: OTEL: Implement OTEL Exporter to expose query details fro…

tanishq-chugh:
…m HiveServer2

Implement OTEL Exporter to expose basic query details from HiveServer2

Better Usability & introduce various OTEL Clients to fetch query data

OTEL Clients can be configured to fetch data from HS2

No

Manual Testing by integrating it with OTEL collector & various receivers such as Jaeger, Zipkin.

### DIFF
--- a/itests/hive-unit/pom.xml
+++ b/itests/hive-unit/pom.xml
@@ -213,6 +213,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-exporter-otlp</artifactId>
+      <version>${otel.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-hdfs</artifactId>
       <classifier>tests</classifier>
--- a/itests/hive-unit/pom.xml
+++ b/itests/hive-unit/pom.xml
@@ -213,11 +213,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>io.opentelemetry</groupId>
-      <artifactId>opentelemetry-exporter-otlp</artifactId>
-      <version>${otel.version}</version>
-    </dependency>
-    <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-hdfs</artifactId>
       <classifier>tests</classifier>
--- a/service/src/java/org/apache/hive/service/servlet/OTELExporter.java
+++ b/service/src/java/org/apache/hive/service/servlet/OTELExporter.java
@@ -18,7 +18,6 @@
 
 package org.apache.hive.service.servlet;
 
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
--- a/service/src/java/org/apache/hive/service/servlet/OTELExporter.java
+++ b/service/src/java/org/apache/hive/service/servlet/OTELExporter.java
@@ -26,7 +26,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
-import com.google.common.base.Joiner;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.trace.Span;
@@ -45,11 +44,11 @@ public class OTELExporter extends Thread {
   private static final String INSTRUMENTATION_NAME = OTELExporter.class.getName();
   private static final Logger LOG = LoggerFactory.getLogger(OTELExporter.class);
   private final OperationManager operationManager;
-  private Set<String> historicalQueryId;
+  private final Set<String> historicalQueryId;
   private final long frequency;
   private final Tracer tracer;
-  private Map<String, Span> queryIdToSpanMap;
-  private Map<String, List<String>> queryIdToTasksMap;
+  private final Map<String, Span> queryIdToSpanMap;
+  private final Map<String, Set<String>> queryIdToTasksMap;
 
   public OTELExporter(OpenTelemetry openTelemetry, SessionManager sessionManager, long frequency) {
     this.tracer = openTelemetry.getTracer(INSTRUMENTATION_NAME);
@@ -88,10 +87,10 @@ public class OTELExporter extends Thread {
       //In case of live query previously encountered in past loops
       if (rootspan != null) {
         for (QueryDisplay.TaskDisplay task : lQuery.getQueryDisplay().getTaskDisplays()) {
-          if (task.getReturnValue() != null && task.getEndTime() != null && !queryIdToTasksMap.get(queryID).contains(task.getTaskId())) {
-            queryIdToTasksMap.get(queryID).add(task.getTaskId());
+          if (task.getReturnValue() != null && task.getEndTime() != null
+                  && queryIdToTasksMap.get(queryID).add(task.getTaskId())) {
             Context parentContext = Context.current().with(rootspan);
-            tracer.spanBuilder(queryID+ " - " + task.getTaskId() + " - live")
+            tracer.spanBuilder(queryID + " - " + task.getTaskId() + " - live")
                     .setParent(parentContext).setAllAttributes(addTaskAttributes(task))
                     .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan()
                     .end(task.getEndTime(), TimeUnit.MILLISECONDS);
@@ -99,17 +98,16 @@ public class OTELExporter extends Thread {
         }
       } else {
         // In case of live queries being seen for first time and has initialized its queryDisplay
-        rootspan = tracer.spanBuilder(queryID + " - live")
-                .startSpan();
-        List<String> completedTasks = new ArrayList<>();
+        rootspan = tracer.spanBuilder(queryID + " - live").startSpan();
+        Set<String> completedTasks = new HashSet<>();
         Context parentContext = Context.current().with(rootspan);
 
         Span initSpan = tracer.spanBuilder(queryID + " - live").setParent(parentContext).startSpan()
-                .setAttribute("queryID", lQuery.getQueryDisplay().getQueryId())
-                .setAttribute("queryString", lQuery.getQueryDisplay().getQueryString())
-                .setAttribute("Begin Time", lQuery.getBeginTime());
+                .setAttribute("QueryId", queryID)
+                .setAttribute("QueryString", lQuery.getQueryDisplay().getQueryString())
+                .setAttribute("BeginTime", lQuery.getBeginTime());
         if (lQuery.getQueryDisplay().getErrorMessage() != null) {
-          initSpan.setAttribute("Error Message", lQuery.getQueryDisplay().getErrorMessage());
+          initSpan.setAttribute("ErrorMessage", lQuery.getQueryDisplay().getErrorMessage());
         }
         initSpan.end();
 
@@ -132,15 +130,15 @@ public class OTELExporter extends Thread {
     for (QueryInfo hQuery : historicalQueries) {
       String hQueryId = hQuery.getQueryDisplay().getQueryId();
       historicalQueryIDs.add(hQueryId);
-      Span rootspan = queryIdToSpanMap.get(hQueryId);
+      Span rootspan = queryIdToSpanMap.remove(hQueryId);
+      Set<String> completedTasks = queryIdToTasksMap.remove(hQueryId);
 
       //For queries that were live till last loop but have ended before start of this loop
       if (rootspan != null) {
         for (QueryDisplay.TaskDisplay task : hQuery.getQueryDisplay().getTaskDisplays()) {
-          if (!queryIdToTasksMap.get(hQueryId).contains(task.getTaskId())) {
-            queryIdToTasksMap.get(hQueryId).add(task.getTaskId());
+          if (!completedTasks.contains(task.getTaskId())) {
             Context parentContext = Context.current().with(rootspan);
-            tracer.spanBuilder(hQueryId+ " - " + task.getTaskId() + " - completed")
+            tracer.spanBuilder(hQueryId + " - " + task.getTaskId() + " - completed")
                     .setParent(parentContext).setAllAttributes(addTaskAttributes(task))
                     .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan()
                     .end(task.getEndTime(), TimeUnit.MILLISECONDS);
@@ -149,30 +147,27 @@ public class OTELExporter extends Thread {
 
         //Update the rootSpan name & attributes before ending it
         rootspan.updateName(hQueryId + " - completed").setAllAttributes(addQueryAttributes(hQuery)).end();
-        queryIdToSpanMap.remove(hQueryId);
-        queryIdToTasksMap.remove(hQueryId);
-
         historicalQueryId.add(hQueryId);
       }
 
       //For queries that already ended either before OTEL service started or in between OTEL loops
-      if (historicalQueryId.add(hQuery.getQueryDisplay().getQueryId())) {
-        rootspan = tracer.spanBuilder(hQuery.getQueryDisplay().getQueryId() + " - completed")
-                .startSpan();
+      if (historicalQueryId.add(hQueryId)) {
+        rootspan = tracer.spanBuilder(hQueryId + " - completed").startSpan();
         Context parentContext = Context.current().with(rootspan);
 
-        Span initSpan = tracer.spanBuilder(hQuery.getQueryDisplay().getQueryId() + " - completed").setParent(parentContext).startSpan()
-                .setAttribute("queryID", hQuery.getQueryDisplay().getQueryId())
-                .setAttribute("queryString", hQuery.getQueryDisplay().getQueryString())
-                .setAttribute("Begin Time", hQuery.getBeginTime());
+        Span initSpan = tracer.spanBuilder(hQueryId + " - completed")
+                .setParent(parentContext).startSpan()
+                .setAttribute("QueryId", hQueryId)
+                .setAttribute("QueryString", hQuery.getQueryDisplay().getQueryString())
+                .setAttribute("BeginTime", hQuery.getBeginTime());
         if (hQuery.getQueryDisplay().getErrorMessage() != null) {
-          initSpan.setAttribute("Error Message", hQuery.getQueryDisplay().getErrorMessage());
+          initSpan.setAttribute("ErrorMessage", hQuery.getQueryDisplay().getErrorMessage());
         }
         initSpan.end();
 
         for (QueryDisplay.TaskDisplay task : hQuery.getQueryDisplay().getTaskDisplays()) {
           parentContext = Context.current().with(rootspan);
-          tracer.spanBuilder(hQuery.getQueryDisplay().getQueryId()+ " - " + task.getTaskId() + " - completed")
+          tracer.spanBuilder(hQueryId + " - " + task.getTaskId() + " - completed")
                   .setParent(parentContext).setAllAttributes(addTaskAttributes(task))
                   .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan()
                   .end(task.getEndTime(), TimeUnit.MILLISECONDS);
@@ -185,21 +180,20 @@ public class OTELExporter extends Thread {
 
   private AttributesMap addQueryAttributes(QueryInfo query){
     AttributesMap attributes = AttributesMap.create(Long.MAX_VALUE, Integer.MAX_VALUE);
-    attributes.put(AttributeKey.stringKey("queryId"), query.getQueryDisplay().getQueryId());
+    attributes.put(AttributeKey.stringKey("QueryId"), query.getQueryDisplay().getQueryId());
     attributes.put(AttributeKey.stringKey("QueryString"), query.getQueryDisplay().getQueryString());
     attributes.put(AttributeKey.longKey("QueryStartTime"), query.getQueryDisplay().getQueryStartTime());
-    attributes.put(AttributeKey.longKey("End Time"), query.getEndTime());
-    attributes.put(AttributeKey.stringKey("Operation Id"), query.getOperationId());
-    attributes.put(AttributeKey.stringKey("Operation Log Location"), query.getOperationLogLocation());
-    attributes.put(AttributeKey.stringKey("Error Message"), query.getQueryDisplay().getErrorMessage());
-    attributes.put(AttributeKey.stringKey("explainPlan"), query.getQueryDisplay().getExplainPlan());
-    attributes.put(AttributeKey.stringKey("fullLogLocation"), query.getQueryDisplay().getFullLogLocation());
-    attributes.put(AttributeKey.stringKey("taskDisplays"), Joiner.on("\t").join(query.getQueryDisplay().getTaskDisplays()));
+    attributes.put(AttributeKey.longKey("EndTime"), query.getEndTime());
+    attributes.put(AttributeKey.stringKey("OperationId"), query.getOperationId());
+    attributes.put(AttributeKey.stringKey("OperationLogLocation"), query.getOperationLogLocation());
+    attributes.put(AttributeKey.stringKey("ErrorMessage"), query.getQueryDisplay().getErrorMessage());
+    attributes.put(AttributeKey.stringKey("ExplainPlan"), query.getQueryDisplay().getExplainPlan());
+    attributes.put(AttributeKey.stringKey("FullLogLocation"), query.getQueryDisplay().getFullLogLocation());
     attributes.put(AttributeKey.stringKey("Running"), String.valueOf(query.isRunning()));
     attributes.put(AttributeKey.longKey("Runtime"), query.getRuntime());
-    attributes.put(AttributeKey.stringKey("User Name"), query.getUserName());
+    attributes.put(AttributeKey.stringKey("UserName"), query.getUserName());
     attributes.put(AttributeKey.stringKey("State"), query.getState());
-    attributes.put(AttributeKey.stringKey("Session Id"), query.getSessionId());
+    attributes.put(AttributeKey.stringKey("SessionId"), query.getSessionId());
     return attributes;
   }
 
@@ -207,14 +201,14 @@ public class OTELExporter extends Thread {
     AttributesMap attributes = AttributesMap.create(Long.MAX_VALUE, Integer.MAX_VALUE);
     attributes.put(AttributeKey.stringKey("TaskId"), taskDisplay.getTaskId());
     attributes.put(AttributeKey.stringKey("Name"), taskDisplay.getName());
-    if (taskDisplay.getTaskType() != null) {
+    if(taskDisplay.getTaskType() != null){
       attributes.put(AttributeKey.stringKey("TaskType"), taskDisplay.getTaskType().toString());
     }
     attributes.put(AttributeKey.stringKey("Status"), taskDisplay.getStatus());
     attributes.put(AttributeKey.stringKey("StatusMessage"), taskDisplay.getStatusMessage());
     attributes.put(AttributeKey.stringKey("ExternalHandle"), taskDisplay.getExternalHandle());
     attributes.put(AttributeKey.stringKey("ErrorMsg"), taskDisplay.getErrorMsg());
-    if (taskDisplay.getReturnValue() != null) {
+    if(taskDisplay.getReturnValue() != null ){
       attributes.put(AttributeKey.longKey("ReturnValue"), taskDisplay.getReturnValue().longValue());
     }
     attributes.put(AttributeKey.longKey("BeginTime"), taskDisplay.getBeginTime());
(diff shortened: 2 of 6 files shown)


itests/hive-unit/pom.xml
```diff
@@ -212,6 +212,11 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-exporter-otlp</artifactId>
+      <version>${otel.version}</version>
+    </dependency>
```
ayushtkn, line 219: Why are we adding in itest? We are only having changes in the prod code

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+    List<QueryInfo> liveQueries = operationManager.getLiveQueryInfos();
+    List<QueryInfo> historicalQueries = operationManager.getHistoricalQueryInfos();
+
+    LOG.debug("Found {} liveQueries and {} historicalQueries", liveQueries.size(), historicalQueries.size());
+
+    for (QueryInfo lQuery: liveQueries){
+
+      while (lQuery.getQueryDisplay() == null) {
+        try {
+          Thread.sleep(500);
```
ayushtkn, line 85: Why is there are sleep here?

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+        } catch (InterruptedException e) {
+          throw new RuntimeException(e);
+        }
+      }
+
+      String queryID = lQuery.getQueryDisplay().getQueryId();
+
+      //In case of live query previously encountered in past loops
+      if(queryIdToSpanMap.containsKey(queryID)){
+        Span rootspan = queryIdToSpanMap.get(queryID);
```
ayushtkn, line 95:
```
      Span rootspan = queryIdToSpanMap.get(queryID);
      if (rootspan != null) {
```

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+            queryIdToTasksMap.get(queryID).add(task.getTaskId());
+            Context parentContext = Context.current().with(rootspan);
+            Span currSpan = tracer.spanBuilder(queryID+ " - " + task.getTaskId() + " - live").setParent(parentContext).setAllAttributes(addTaskAttributes(task))
+                    .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan();
+            while (task.getEndTime() == null){
+              try {
+                Thread.sleep(500);
+              } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+              }
```
ayushtkn, line 108: this is same as below, refactor into a method

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+        for (QueryDisplay.TaskDisplay task : lQuery.getQueryDisplay().getTaskDisplays()) {
+          if (task.getReturnValue() != null) {
+            completedTasks.add(task.getTaskId());
+            Context parentContext = Context.current().with(rootspan);
+            Span currSpan = tracer.spanBuilder(queryID + " - " + task.getTaskId() + " - live").setParent(parentContext).setAllAttributes(addTaskAttributes(task))
+                    .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan();
+
+            while (task.getEndTime() == null) {
+              try {
+                Thread.sleep(500);
```
ayushtkn, line 127: We can't have sleep in the prod code like this, that too when the code is sequential, if it is null, just drop this task in this iteration and we will catch this in next iteration

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+              } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+              }
+            }
+            currSpan.end(task.getEndTime(), TimeUnit.MILLISECONDS);
+          }
+        }
+
+        if(rootspan != null){
```
ayushtkn, line 136: if which case `rootspan` will be `null` here?

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+      }
+    }
+
+
+    for (QueryInfo hQuery : historicalQueries) {
+
+      //For queries that were live till last loop but have ended before start of this loop
+      if(queryIdToSpanMap.containsKey(hQuery.getQueryDisplay().getQueryId())){
+        String hQueryId = hQuery.getQueryDisplay().getQueryId();
+        Span rootspan = queryIdToSpanMap.get(hQueryId);
```
ayushtkn, line 150: Change to 
```
      String hQueryId = hQuery.getQueryDisplay().getQueryId();
      Span rootspan = queryIdToSpanMap.get(hQueryId);
      if (rootspan != null) {
```

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+        for (QueryDisplay.TaskDisplay task : hQuery.getQueryDisplay().getTaskDisplays()) {
+          if(queryIdToTasksMap.get(hQueryId) == null || !queryIdToTasksMap.get(hQueryId).contains(task.getTaskId())){
```
ayushtkn, line 152: can we use ``rootspan`` instead of ``queryIdToTasksMap.get(hQueryId)``

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+            queryIdToTasksMap.get(hQueryId).add(task.getTaskId());
+            Context parentContext = Context.current().with(rootspan);
+            Span currSpan = tracer.spanBuilder(hQueryId+ " - " + task.getTaskId() + " - completed").setParent(parentContext).setAllAttributes(addTaskAttributes(task))
+                    .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan();
+            currSpan.end(task.getEndTime(), TimeUnit.MILLISECONDS);
```
ayushtkn, line 157: I don't think we need to store `currSpan` we can chain the calls
```
            tracer.spanBuilder(hQueryId + " - " + task.getTaskId() + " - completed")
                .setParent(parentContext).setAllAttributes(addTaskAttributes(task))
                .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan()
                .end(task.getEndTime(), TimeUnit.MILLISECONDS);
```

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+          }
+        }
+
+        //Update the rootSpan name & attributes before ending it
+        queryIdToSpanMap.get(hQueryId).updateName(hQueryId + " - completed");
+        queryIdToSpanMap.get(hQueryId).setAllAttributes(addQueryAttributes(hQuery));
+        queryIdToSpanMap.get(hQueryId).end();
```
ayushtkn, line 164: Does this work instead
```
        //Update the rootSpan name & attributes before ending it
        rootspan.updateName(hQueryId + " - completed").setAllAttributes(addQueryAttributes(hQuery)).end();
```

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+
+        queryIdToSpanMap.remove(hQueryId);
+        queryIdToTasksMap.remove(hQueryId);
+
+        historicalQueryId.add(hQueryId);
+      }
+
+      //For queries that were already either before OTEL service started or in between OTEL loops
+      if (!historicalQueryId.contains(hQuery.getQueryDisplay().getQueryId())) {
+        historicalQueryId.add(hQuery.getQueryDisplay().getQueryId());
```
ayushtkn, line 174: It is a Set, you don't need to do contains then add, you can directly add and check the return value, if it is true, means the element wasn't there, if false, means the element was already there

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,229 @@
...
+                .startSpan();
+        for (QueryDisplay.TaskDisplay task : hQuery.getQueryDisplay().getTaskDisplays()) {
+          Context parentContext = Context.current().with(rootSpan);
+          Span currSpan = tracer.spanBuilder(hQuery.getQueryDisplay().getQueryId()+ " - " + task.getTaskId() + " - completed").setParent(parentContext).setAllAttributes(addTaskAttributes(task))
+                  .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan();
+          currSpan.end(task.getEndTime(), TimeUnit.MILLISECONDS);
+        }
+        rootSpan.setAllAttributes(addQueryAttributes(hQuery)).end();
+      }
+
```
ayushtkn, line 185: Your `historicalQueryId` will grow infinitely, you need to reset it the only ones returned by ``    List<QueryInfo> historicalQueries = operationManager.getHistoricalQueryInfos();``

This would return say only 25 (by default) but you would be storing it infinitely which will ultimately lead to OOM on the HS2 side

ayushtkn (changes requested):
Thanx @tanishq-chugh, have dropped some comments, please check & I think your code ain't properly formatted

itests/hive-unit/pom.xml
tanishq-chugh: This is required in case of users running OTEL Service for HS2 which is running locally built from source code.

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: In some cases, OTEL service was capturing the liveQueries in a state where queryDisplay was not initialised and leading to Null Pointer Exception. This is changed to drop such queries for the current OTEL run and capture it in next run.

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: My bad, this slipped in from my previous version of this code. Removed it.

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: Yes, i have refactored the code such that at the end of each OTEL loop, all the queries which have been removed from `historicalQueries` will also be removed from our `historicalQueryId`.

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: Have refactored the code to remove the while loop of thread.sleep and usage of currSpan variable as well leading to single line of code for this block.

itests/hive-unit/pom.xml
ayushtkn: We don't package itests as part of the distribution, this is integration test module. HS2 should run locally without this itests, it ain't there in the class path either

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+import org.apache.hadoop.hive.ql.QueryDisplay;
+import org.apache.hadoop.hive.ql.QueryInfo;
+import org.apache.hive.service.cli.operation.OperationManager;
+import org.apache.hive.service.cli.session.SessionManager;
+
+public class OTELExporter extends Thread {
+  private static final String INSTRUMENTATION_NAME = OTELExporter.class.getName();
+  private static final Logger LOG = LoggerFactory.getLogger(OTELExporter.class);
+  private final OperationManager operationManager;
+  private Set<String> historicalQueryId;
```
ayushtkn, line 48: Should be `final`

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+  private final long frequency;
+  private final Tracer tracer;
+  private Map<String, Span> queryIdToSpanMap;
+  private Map<String, List<String>> queryIdToTasksMap;
```
ayushtkn, line 52: should be `final`

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+      if(lQuery.getQueryDisplay() == null){
+        continue;
+      }
+      String queryID = lQuery.getQueryDisplay().getQueryId();
+      Span rootspan = queryIdToSpanMap.get(queryID);
+
+      //In case of live query previously encountered in past loops
+      if (rootspan != null) {
+        for (QueryDisplay.TaskDisplay task : lQuery.getQueryDisplay().getTaskDisplays()) {
+          if (task.getReturnValue() != null && task.getEndTime() != null && !queryIdToTasksMap.get(queryID).contains(task.getTaskId())) {
```
ayushtkn, line 91: the line length is more than the allowed limit of 120

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+            tracer.spanBuilder(queryID+ " - " + task.getTaskId() + " - live")
+                    .setParent(parentContext).setAllAttributes(addTaskAttributes(task))
+                    .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan()
+                    .end(task.getEndTime(), TimeUnit.MILLISECONDS);
+          }
+        }
+      } else {
+        // In case of live queries being seen for first time and has initialized its queryDisplay
+        rootspan = tracer.spanBuilder(queryID + " - live")
+                .startSpan();
```
ayushtkn, line 103: nit
```
        rootspan = tracer.spanBuilder(queryID + " - live").startSpan();
```

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+        List<String> completedTasks = new ArrayList<>();
+        Context parentContext = Context.current().with(rootspan);
+
+        Span initSpan = tracer.spanBuilder(queryID + " - live").setParent(parentContext).startSpan()
+                .setAttribute("queryID", lQuery.getQueryDisplay().getQueryId())
+                .setAttribute("queryString", lQuery.getQueryDisplay().getQueryString())
+                .setAttribute("Begin Time", lQuery.getBeginTime());
+        if (lQuery.getQueryDisplay().getErrorMessage() != null) {
+          initSpan.setAttribute("Error Message", lQuery.getQueryDisplay().getErrorMessage());
```
ayushtkn, line 112: Can you remove the Space and make it `BeginTime` & `ErrorMessage`

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+                    .setParent(parentContext).setAllAttributes(addTaskAttributes(task))
+                    .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan()
+                    .end(task.getEndTime(), TimeUnit.MILLISECONDS);
+          }
+        }
+      } else {
+        // In case of live queries being seen for first time and has initialized its queryDisplay
+        rootspan = tracer.spanBuilder(queryID + " - live")
+                .startSpan();
+        List<String> completedTasks = new ArrayList<>();
```
ayushtkn, line 104: Shouldn't this be a `Set`

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+        queryIdToSpanMap.put(queryID,rootspan);
+        queryIdToTasksMap.put(queryID,completedTasks);
+      }
+    }
+
+    List<String> historicalQueryIDs = new ArrayList<>();
+    for (QueryInfo hQuery : historicalQueries) {
+      String hQueryId = hQuery.getQueryDisplay().getQueryId();
+      historicalQueryIDs.add(hQueryId);
+      Span rootspan = queryIdToSpanMap.get(hQueryId);
```
ayushtkn, line 135: this should be `remove` instead of `get`

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+
+      //For queries that were live till last loop but have ended before start of this loop
+      if (rootspan != null) {
+        for (QueryDisplay.TaskDisplay task : hQuery.getQueryDisplay().getTaskDisplays()) {
+          if (!queryIdToTasksMap.get(hQueryId).contains(task.getTaskId())) {
+            queryIdToTasksMap.get(hQueryId).add(task.getTaskId());
```
ayushtkn, line 141: can we do ``queryIdToTasksMap.remove(hQueryId)`` once & store it in a variable and reuse & why do we need to add here, we won't ever come back for this query again, right?

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+            Context parentContext = Context.current().with(rootspan);
+            tracer.spanBuilder(hQueryId+ " - " + task.getTaskId() + " - completed")
```
ayushtkn, line 143: nit
space before +

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+        //Update the rootSpan name & attributes before ending it
+        rootspan.updateName(hQueryId + " - completed").setAllAttributes(addQueryAttributes(hQuery)).end();
+        queryIdToSpanMap.remove(hQueryId);
+        queryIdToTasksMap.remove(hQueryId);
+
+        historicalQueryId.add(hQueryId);
+      }
+
+      //For queries that already ended either before OTEL service started or in between OTEL loops
+      if (historicalQueryId.add(hQuery.getQueryDisplay().getQueryId())) {
```
ayushtkn, line 159: should be
```
      if (historicalQueryId.add(hQueryId))
```

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+        rootspan = tracer.spanBuilder(hQuery.getQueryDisplay().getQueryId() + " - completed")
+                .startSpan();
+        Context parentContext = Context.current().with(rootspan);
+
+        Span initSpan = tracer.spanBuilder(hQuery.getQueryDisplay().getQueryId() + " - completed").setParent(parentContext).startSpan()
```
ayushtkn, line 164: nit
line length is more than 120

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+    attributes.put(AttributeKey.stringKey("queryId"), query.getQueryDisplay().getQueryId());
+    attributes.put(AttributeKey.stringKey("QueryString"), query.getQueryDisplay().getQueryString());
+    attributes.put(AttributeKey.longKey("QueryStartTime"), query.getQueryDisplay().getQueryStartTime());
+    attributes.put(AttributeKey.longKey("End Time"), query.getEndTime());
+    attributes.put(AttributeKey.stringKey("Operation Id"), query.getOperationId());
+    attributes.put(AttributeKey.stringKey("Operation Log Location"), query.getOperationLogLocation());
+    attributes.put(AttributeKey.stringKey("Error Message"), query.getQueryDisplay().getErrorMessage());
+    attributes.put(AttributeKey.stringKey("explainPlan"), query.getQueryDisplay().getExplainPlan());
+    attributes.put(AttributeKey.stringKey("fullLogLocation"), query.getQueryDisplay().getFullLogLocation());
+    attributes.put(AttributeKey.stringKey("taskDisplays"), Joiner.on("\t").join(query.getQueryDisplay().getTaskDisplays()));
```
ayushtkn, line 197: this ain't required IMO, we are having spans for each task now

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+        rootspan.setAllAttributes(addQueryAttributes(hQuery)).end();
+      }
+    }
+    historicalQueryId.retainAll(historicalQueryIDs);
+  }
+
+  private AttributesMap addQueryAttributes(QueryInfo query){
+    AttributesMap attributes = AttributesMap.create(Long.MAX_VALUE, Integer.MAX_VALUE);
+    attributes.put(AttributeKey.stringKey("queryId"), query.getQueryDisplay().getQueryId());
```
ayushtkn, line 188: I think queryId is already there right? if not q -> Q

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+    attributes.put(AttributeKey.stringKey("QueryString"), query.getQueryDisplay().getQueryString());
+    attributes.put(AttributeKey.longKey("QueryStartTime"), query.getQueryDisplay().getQueryStartTime());
+    attributes.put(AttributeKey.longKey("End Time"), query.getEndTime());
+    attributes.put(AttributeKey.stringKey("Operation Id"), query.getOperationId());
+    attributes.put(AttributeKey.stringKey("Operation Log Location"), query.getOperationLogLocation());
+    attributes.put(AttributeKey.stringKey("Error Message"), query.getQueryDisplay().getErrorMessage());
```
ayushtkn, line 194: remove space

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+      //For queries that already ended either before OTEL service started or in between OTEL loops
+      if (historicalQueryId.add(hQuery.getQueryDisplay().getQueryId())) {
+        rootspan = tracer.spanBuilder(hQuery.getQueryDisplay().getQueryId() + " - completed")
+                .startSpan();
+        Context parentContext = Context.current().with(rootspan);
+
+        Span initSpan = tracer.spanBuilder(hQuery.getQueryDisplay().getQueryId() + " - completed").setParent(parentContext).startSpan()
+                .setAttribute("queryID", hQuery.getQueryDisplay().getQueryId())
+                .setAttribute("queryString", hQuery.getQueryDisplay().getQueryString())
+                .setAttribute("Begin Time", hQuery.getBeginTime());
```
ayushtkn, line 167: shouldn't we instead call ``addQueryAttributes`` here?

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: Yes, changed the Map to store the tasks in a set and have initially stored the set in a variable by removing it from `queryIdToTasksMap`. Also, have removed the add of new taskID as its not required.

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: have made the change at other similar points as well.

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: Yes, removed taskDisplays attribute.

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: It would make initSpan and rootSpan storing the same content in case of historical queries. Would be better if we keep only these minimal attributes in initSpan so that there's consistency between live and historical traces.

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,219 @@
...
+            completedTasks.add(task.getTaskId());
+            parentContext = Context.current().with(rootspan);
+            tracer.spanBuilder(queryID + " - " + task.getTaskId() + " - live")
+                    .setParent(parentContext).setAllAttributes(addTaskAttributes(task))
+                    .setStartTimestamp(task.getBeginTime(), TimeUnit.MILLISECONDS).startSpan()
+                    .end(task.getEndTime(), TimeUnit.MILLISECONDS);
+          }
+        }
+        queryIdToSpanMap.put(queryID,rootspan);
+        queryIdToTasksMap.put(queryID,completedTasks);
```
ayushtkn, line 125: this isn't formatted, can you format the code properly here & the other similar places

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,219 @@
...
+      }
+    }
+
+    List<String> historicalQueryIDs = new ArrayList<>();
```
ayushtkn, line 129: Should be a set

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,219 @@
...
+      } else {
+        // In case of live queries being seen for first time and has initialized its queryDisplay
+        rootspan = tracer.spanBuilder(queryID + " - live").startSpan();
+        Set<String> completedTasks = new HashSet<>();
+        Context parentContext = Context.current().with(rootspan);
+
+        Span initSpan = tracer.spanBuilder(queryID + " - live").setParent(parentContext).startSpan()
+                .setAttribute("QueryId", queryID)
+                .setAttribute("QueryString", lQuery.getQueryDisplay().getQueryString())
+                .setAttribute("BeginTime", lQuery.getBeginTime());
```
ayushtkn, line 108: I think we should set ``setStartTimestamp`` for the intiSpan with this value rather than the attribute

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,219 @@
...
+          }
+        }
+      } else {
+        // In case of live queries being seen for first time and has initialized its queryDisplay
+        rootspan = tracer.spanBuilder(queryID + " - live").startSpan();
+        Set<String> completedTasks = new HashSet<>();
+        Context parentContext = Context.current().with(rootspan);
+
+        Span initSpan = tracer.spanBuilder(queryID + " - live").setParent(parentContext).startSpan()
+                .setAttribute("QueryId", queryID)
```
ayushtkn, line 106: Add user as well

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,219 @@
...
+                .setAttribute("QueryString", lQuery.getQueryDisplay().getQueryString())
```
ayushtkn, line 107: can let queryString present only in rootSpan, some queries are huge to be put twice and are more eye catching as compared to just query id

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,219 @@
...
+                .setAttribute("BeginTime", lQuery.getBeginTime());
+        if (lQuery.getQueryDisplay().getErrorMessage() != null) {
+          initSpan.setAttribute("ErrorMessage", lQuery.getQueryDisplay().getErrorMessage());
+        }
+        initSpan.end();
```
ayushtkn, line 112: this I believe should have the begin time as the end time here

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: Added queryString into initSpan and removed it from rootSpan

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
```diff
@@ -0,0 +1,225 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hive.service.servlet;
+
+import java.util.ArrayList;
```
ayushtkn, line 21: unused import

ayushtkn (approved):
LGTM

service/src/java/org/apache/hive/service/servlet/OTELExporter.java
tanishq-chugh: my bad, removing it.

PR #5452, opened 2024-09-17
HIVE-24167: Compilation fails due to equivalence mapping violation when CTE materialization is enabled

okumin:

With this change, Hive skips using or persisting runtime stats when PlanMapper throws [equivalence mapping violation](https://github.com/apache/hive/blob/rel/release-4.0.0/ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java#L220).

In some more contexts, we've tried to resolve the violation in #5037 and #5077. However, we couldn't find an obvious way to ideally avoid the issue, which means a modification will likely cause another problem. That's why we are currently biased toward the current approach.

When this issue occurs, the query always fails. We think the disadvantage of failure is not well balanced with the advantage of fast failure. A query might run with a bit poorer performance without runtime stats, but it is not critical in most cases. That's why we would try disabling runtime stats in this case.

In the new behavior, affected queries will not fail but run without runtime statistics.

No.

qtests.


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -809,6 +809,10 @@ public static enum ConfVars {
     HIVE_IN_TEST_REPL("hive.in.repl.test", false, "internal usage only, true in replication test mode", true),
     HIVE_IN_TEST_IDE("hive.in.ide.test", false, "internal usage only, true if test running in ide",
         true),
+    HIVE_IN_TEST_PLANMAPPER_STRICT_VALIDATION("hive.in.test.planmapper.strict.validation", false,
+        "internal use only, whether to raise an error when unexpected links are found. We ignore equivalence mapping "
+        + "violation because it introduces only minor problems. But we want to strictly check it in qtest so that we "
+        + "can prevent further degradations"),
```
zabetak, line 815: I am unsure about this newly introduced property. It may prevent future "degradation" but it puts an extra burden on contributors and reviewers.

As a reviewer, when I see people setting on/off properties in tests I always ask why? And if the answer is "because there is a problem" then I will ask for more information. The contributor will try to understand what's going on and eventually it will find this ticket and argue that this feature is broken so its best to disable the property for their use-case (as out of scope). The reviewer may agree or push back and say that this is blocking the PR from merging.

An exception is by definition serious, a warning on the other hand is not so much and I feel that for this use-case it should have been a warning in the first-place.

ql/src/java/org/apache/hadoop/hive/ql/parse/TezCompiler.java
```diff
@@ -991,35 +987,6 @@ private void removeSemiJoinIfNoStats(OptimizeTezProcContext procCtx)
     ogw.startWalking(topNodes, null);
   }
 
-  private static class CollectAll implements SemanticNodeProcessor {
```
zabetak, line 994: Why are we touching this?

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -217,7 +230,11 @@ private void link(Object o1, Object o2, boolean mayMerge) {
     }
     if (mGroups.size() > 1) {
       if (!mayMerge) {
-        throw new RuntimeException("equivalence mapping violation");
+        LOG.warn("Illegally linking {} and {}", o1, o2);
+        if (failsWithIllegalLink) {
+          throw new RuntimeException("equivalence mapping violation");
+        }
+        isBroken.set(true);
```
zabetak, line 237: It seems to me that if we raise a warning and return we don't really break the `PlanMapper` since the "invalid" expression is not added. This would allow the mapper to be used based on the existing content. Is it necessary to invalidate everything that's already in?

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -256,20 +277,15 @@ public <T> List<T> getAll(Class<T> clazz) {
...
-  public <T> List<T> lookupAll(Class<T> clazz, Object key) {
+  private <T> List<T> lookupAll(Class<T> clazz, Object key) {
     EquivGroup group = objectMap.get(key);
     if (group == null) {
       throw new NoSuchElementException(Objects.toString(key));
     }
     return group.getAll(clazz);
   }
 
+  @VisibleForTesting
```
zabetak, line 288: Why adding the annotation?

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -279,7 +295,6 @@ public <T> T lookup(Class<T> clazz, Object key) {
     return all.get(0);
   }
 
-  @VisibleForTesting
```
zabetak, line 282: Why removing the annotation?

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
```diff
@@ -32,21 +32,34 @@
...
  */
 public class PlanMapper {
+  private static final Logger LOG = LoggerFactory.getLogger(PlanMapper.class);
 
-  Set<EquivGroup> groups = new HashSet<>();
-  private Map<Object, EquivGroup> objectMap = new CompositeMap<>(OpTreeSignature.class, AuxOpTreeSignature.class);
+  private final Set<EquivGroup> groups = new HashSet<>();
+  private final Map<Object, EquivGroup> objectMap = new CompositeMap<>(OpTreeSignature.class, AuxOpTreeSignature.class);
+  private final boolean failsWithIllegalLink;
+  private final AtomicBoolean isBroken = new AtomicBoolean(false);
```
zabetak, line 58: Why AtomicBoolean? The whole class does not look thread-safe.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
okumin: Thanks. Let's say our qtest covers 99.9% of use cases of Hive(I trust the coverage 😃 ), which means 0.1% are uncovered. I assume we agreed that it is too costly to achieve 100% promptly.
I personally thought it'd be more reasonable to use CI if we wanted to keep the stability high enough, e.g. 90%. I'm a bit pessimistic here and I'm going to remove this property.

ql/src/java/org/apache/hadoop/hive/ql/parse/TezCompiler.java
okumin: This is not used anywhere. I assume it is costly to ensure unused classes works correctly compared with advantages to keep it.

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
okumin: I lost my memory but I also think we have no reasons to make only this variable thread-safe. I will make it a primitive boolean.

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
okumin: This is used only by unit tests.

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
okumin: Because this is used by StatsSources.

okumin (commented):
Thanks for reviewing the PR. I am updating the branch based on the discussion.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
okumin: Done by https://github.com/apache/hive/pull/5452/commits/4f60470184e766798f6bf32319aca90c97446312

ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java
okumin: This change would let PlanMapper return empty links.
https://github.com/apache/hive/pull/5452/commits/319a26410674d12290dc7b98e5e7032483f534b1

My assumptions.
- Maybe, it is understandable and safe to invalidate the entire PlanMapper once we find something wrong
- Empty links vs IllegalStateException. I am wondering which is better honestly

zabetak: I don't fully understand the reasoning behind the new proposal. 

From my perspective what I had in mind was something like the following:
```java
if (!mayMerge) {
  LOG.warn("...");
  return;
}
```
I am still trying to understand if it is necessary to invalidate the `PlanMapper` or if we can just skip the "problematic" entry.

okumin: Now, I understand what you said, maybe. I was convinced that it was all or nothing, which means I thought a partially valid PlanMapper could never exist. Your comment should be worth considering.

I'm checking the source code with my mind updated.

```
if (mGroups.size() > 1) {
  if (!mayMerge) {
    LOG.warn("Illegally linking {} and {}", o1, o2);
    return;
  }
```

I guess we should be a bit more defensive. That's because we have two places using [AuxOpTreeSignature](https://github.com/apache/hive/blob/454415b4798dbb7accf5e5de23db3dc4801b7d68/ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/AuxOpTreeSignature.java#L76) which [links all Operators with mayMerge=true](https://github.com/apache/hive/blob/454415b4798dbb7accf5e5de23db3dc4801b7d68/ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/PlanMapper.java#L198-L219). It is not evident why AuxOpTreeSignature can safely relink the Operators which have been rejected once before. It could be illegal.

```
if (mGroups.size() > 1) {
  if (!mayMerge) {
    LOG.warn("Illegally linking {} and {}", o1, o2);
    mGroups.forEach(g -> g.add(new OperatorStats.IncorrectRuntimeStatsMarker()));
    return;
  }
```

I guess this can keep the valid part alive, marking the unexpected part invalidated and removing `PlanMapper#isBroken`. I am thinking of that approach.

okumin: I checked the use cases querying PlanMapper.
(1) [RuntimeStatsPersistenceCheckerHook](https://github.com/apache/hive/blob/b431e11eb19def7df978547bd161ba102d59083c/ql/src/java/org/apache/hadoop/hive/ql/hooks/RuntimeStatsPersistenceCheckerHook.java) iterates all signatures to test RuntimeStatsPersister can correctly encode/decode OpTreeSignature. Only test codes use this class, and this class probably works even if a signature has illegal links.
(2) [StatsSources](https://github.com/apache/hive/blob/b431e11eb19def7df978547bd161ba102d59083c/ql/src/java/org/apache/hadoop/hive/ql/plan/mapper/StatsSources.java) iterates all elements to get valid ones for the optimizer. This is likely to work even with illegal links as long as we attach `IncorrectRuntimeStatsMarker` to them.
(3) [ReOptimizePlugin](https://github.com/apache/hive/blob/b431e11eb19def7df978547bd161ba102d59083c/ql/src/java/org/apache/hadoop/hive/ql/reexec/ReOptimizePlugin.java) iterates all Operators to identify whether the execution plan is changed or not. This logic probably works even if there are illegal links.

I think both (1) and (3) don't rely on the existence of mapping violations. We have to take care of (2) because we may fill runtime stats of an unrelated Operator if a mapping violation is a real problem. Anyway, I'll try a PoC

okumin: @zabetak Does this align with your thoughts? If yes, I will push it to this PR
- [new commit](https://github.com/okumin/hive/commit/63cb34c83d7c34c3d1ff1a9847f53f913657e4d3)
- [entire diff](https://github.com/okumin/hive/compare/49a6ff8b3cea365ceb62ea470c43fee2d1d252eb...63cb34c83d7c34c3d1ff1a9847f53f913657e4d3)

zabetak: @okumin I am good with the new approach. It's more defensive than what I had in mind but for sure you are more familiar with this code than myself so I trust your judgement.

okumin:
@zabetak Updated. Thanks for reviewing this PR and giving many suggestions. I believe the change is much simpler

okumin:
I rebased it again because the previous base revision was affected by HIVE-28611

zabetak (approved):
LGTM! 

@okumin Now that you are a committer you can merge this yourself.

The PR/JIRA summary are a bit narrow given the nature of the problem and the fix that we are about to merge. Consider updating the commit message and JIRA with something more general/descriptive.

**Examples**
HIVE-24167: Compilation fails due to equivalence mapping violation when CTE materialization is enabled
HIVE-24167: Equivalence mapping violation in PlanMapper when CTE materialization is enabled

okumin:
Thanks. Let me pick up the first option.

> Ensure that the Pull Request has a +1 vote, and that 24 hours have elapsed since the first +1 vote was cast on the Pull Request. Note that this rule appears in the Hive Bylaws. Do not ignore it.

Based on this rule, I will wait for 1 day and return here.

PR #5454, opened 2024-09-18
HIVE-28526 may produce null pointer when struct type value is null

terrytlu:
null pointer check

 it is a bug.

No

No

sql test, add ql/src/test/queries/clientpositive/struct_null_select.q
create table test_struct
(
    f1 string,
    demo_struct struct\<f1:string,f2:string,f3:string\>,
    datestr string
);
insert into test_struct(f1, datestr) select 'test_f1','datestr_1';

### DIFF
--- a/ql/src/test/queries/clientpositive/struct_null_select.q
+++ b/ql/src/test/queries/clientpositive/struct_null_select.q
@@ -1,0 +1,10 @@
+create table test_struct
+(
+    f1 string,
+    demo_struct struct<f1:string,f2:string,f3:string>,
+    datestr string
+);
+
+insert into test_struct(f1, datestr) select 'test_f1','datestr_1';
+
+drop table test_struct;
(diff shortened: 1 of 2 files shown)


ayushtkn (commented):
You have to generate q.out file for your .q file as well, the test fails otherwise
https://ci.hive.apache.org/job/hive-precommit/job/PR-5454/1/testReport/junit/org.apache.hadoop.hive.cli.split22/TestMiniLlapLocalCliDriver/Testing___split_14___PostProcess___testCliDriver_struct_null_select_/

ql/src/test/queries/clientpositive/struct_null_select.q
```diff
@@ -0,0 +1,10 @@
+create table test_struct
```
difin, line 1: The name of the q-test is a little bit confusing to me because the test doesn't seem to select any nulls.
Also, is the .q.out file is missing for this new test?

ngsg:
I failed reproduce the problem with your test on TestMiniLlapLocalCliDriver, and it seems that this issue is fixed by HIVE-27690. Could you check and share your test environment?

PR #5455, opened 2024-09-18
HIVE-28343 Iceberg: Major QB Compaction support filter in compaction …

zratkai:
HIVE-28343 Iceberg: Major QB Compaction support filter in compaction request in OPTIMIZE TABLE command

Change-Id: I6e3c09f67423cf559f315d6f448877c967da7479

With this PR user can execute optimize table command with condition.

To execute optimize table with condition.

Yes,  user. can use optimize table with condition.

No.

Unit test, qtest.

### DIFF
--- a/parser/src/test/org/apache/hadoop/hive/ql/parse/TestParseOptimizeTable.java
+++ b/parser/src/test/org/apache/hadoop/hive/ql/parse/TestParseOptimizeTable.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hive.ql.parse;
+
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+public class TestParseOptimizeTable {
+  ParseDriver parseDriver = new ParseDriver();
+  private static final String EXPECTED_WHERE_FILTER = "\n" + 
+      "nil\n" + 
+      "   TOK_ALTERTABLE\n" + 
+      "      TOK_TABNAME\n" + 
+      "         tbl0\n" + 
+      "      TOK_ALTERTABLE_COMPACT\n" + 
+      "         'MAJOR'\n" + 
+      "         TOK_BLOCKING\n" + 
+      "         TOK_WHERE\n" + 
+      "            and\n" + 
+      "               TOK_FUNCTION\n" + 
+      "                  in\n" + 
+      "                  TOK_TABLE_OR_COL\n" + 
+      "                     col01\n" + 
+      "                  'A'\n" + 
+      "                  'B'\n" + 
+      "               <\n" + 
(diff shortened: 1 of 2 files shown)


difin:
Looks good to me, pending tests and minor ask related to the new unit test.

parser/src/test/org/apache/hadoop/hive/ql/parse/TestParseOptimizeTable.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.hamcrest.core.Is.is;
+
+public class TestParseOptimizeTable {
+  ParseDriver parseDriver = new ParseDriver();
+  @Test
+  public void testOptimizeTableWithWhere() throws Exception {
+    ASTNode tree = parseDriver.parse(
+        " optimize table tbl0 rewrite data where (col01 in ('A', 'B') and col02 < '2024-09-17 00:00:00')", null).getTree();
+    String expected = "\n" + "nil\n" + "   TOK_ALTERTABLE\n" + "      TOK_TABNAME\n" + "         tbl0\n" + "      TOK_ALTERTABLE_COMPACT\n" + "         'MAJOR'\n" + "         TOK_BLOCKING\n" + "         TOK_WHERE\n" + "            and\n" + "               TOK_FUNCTION\n" + "                  in\n" + "                  TOK_TABLE_OR_COL\n" + "                     col01\n" + "                  'A'\n" + "                  'B'\n" + "               <\n" + "                  TOK_TABLE_OR_COL\n" + "                     col02\n" + "                  '2024-09-17 00:00:00'\n" + "   <EOF>\n";
+    assertThat(tree.dump(), is(expected));
```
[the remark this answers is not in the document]
zratkai: It contains the TOK_WHERE which comes from WHERE clause, so it is tested.

zratkai (commented):
Thanks for the review! Please check my reply.

aturoczy (approved):

iceberg/iceberg-handler/src/test/queries/positive/iceberg_optimize_table_partition_evolution_w_dyn_spec_w_filter.q
```diff
@@ -0,0 +1,85 @@
...
+-- Mask compaction id as they will be allocated in parallel threads
+--! qt:replace:/^[0-9]/#Masked#/
+-- Mask removed file size
+--! qt:replace:/(\S\"removed-files-size\\\":\\\")(\d+)(\\\")/$1#Masked#$3/
+
+set hive.llap.io.enabled=true;
+set hive.vectorized.execution.enabled=true;
+set hive.optimize.shared.work.merge.ts.schema=true;
+
+create table ice_orc (
```
deniskuzZ, line 24: why do we need such a complex test that is just a copy-paste of the existing test? Can't we just use explain?

zratkai: It tests the same logic, but with "optimize table" instead of "alter table" and for that it needs the same preconditions. I wanted to ensure everything is same, so it is easy to compare if it does what it supposed to.

deniskuzZ: it doesn't make sense to execute a duplicate heavy test just for a syntax "sugar"

zratkai: In the test it check e.g. the numFiles changes. Without a table, data, it won't be testable. 
Checking the generated plan would not be the same. To test compaction the best way is:
- create a table
- insert data with multiple insert statement
- check the content of the table
- check table description (e.g. number of files)
- run compaction
- check table content (there must be no change)
- check table description (e.g. number of files decreased)

And this is what this test does. Making it simpler would mean it won't test the whole process.

deniskuzZ (changes requested):
You haven't implemented any new compaction functionality, but added just a parser sugar. There is no need to test the whole Hive for such a minor change. 
The new test is a complete copy-paste of the existing test without sugar, so it's just a waste of Jenkins resources. qtests are expensive and are overkill for such minor changes. 
You need to test that optimize command is being correctly rewritten to alter and that is it.

zratkai:
@deniskuzZ I deleted the qtest.

deniskuzZ (approved):
LGTM +1

PR #4077, opened 2023-02-21
HIVE-26735: Ability to sort the data during rebalancing compaction

veghlaci05:
### What changes were proposed in this pull request?
The REBALANCE compaction now accepts an ORDER BY clause.

### Why are the changes needed?
 To be able to reorder the data during the compaction.

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
Manually and through unit tests.

PR #5456, opened 2024-09-18
HIVE-28524: Iceberg: Major QB Compaction add sort order support

difin:

Added sort order support for Hive Iceberg Compaction that allows to optionally order the data as part of compaction operation.

Sort ordering the data as part of compaction can give additional performance improvement.

Yes, Hive Iceberg compaction command now supports ORDER BY clause.

No

New q-tests.

### DIFF
--- a/iceberg/iceberg-handler/src/test/results/positive/llap/iceberg_major_compaction_unpartitioned_ordered.q.out
+++ b/iceberg/iceberg-handler/src/test/results/positive/llap/iceberg_major_compaction_unpartitioned_ordered.q.out
@@ -16,140 +16,36 @@ tblproperties ('format-version'='2')
 POSTHOOK: type: CREATETABLE
 POSTHOOK: Output: database:default
 POSTHOOK: Output: default@ice_orc
-PREHOOK: query: insert into ice_orc VALUES ('fn1','ln1')
+PREHOOK: query: insert into ice_orc VALUES 
+('fn1','ln1'),
+('fn2','ln2'),
+('fn3','ln3'),
+('fn4','ln4'),
+('fn5','ln5'),
+('fn6','ln6'),
+('fn7','ln7')
 PREHOOK: type: QUERY
 PREHOOK: Input: _dummy_database@_dummy_table
 PREHOOK: Output: default@ice_orc
-POSTHOOK: query: insert into ice_orc VALUES ('fn1','ln1')
+POSTHOOK: query: insert into ice_orc VALUES 
+('fn1','ln1'),
+('fn2','ln2'),
+('fn3','ln3'),
+('fn4','ln4'),
+('fn5','ln5'),
+('fn6','ln6'),
+('fn7','ln7')
 POSTHOOK: type: QUERY
 POSTHOOK: Input: _dummy_database@_dummy_table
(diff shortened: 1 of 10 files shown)


iceberg/iceberg-handler/src/test/results/positive/llap/iceberg_major_compaction_unpartitioned_ordered.q.out
```diff
@@ -184,21 +80,21 @@ Table Parameters:
...
 #### A masked pattern was here ####
-	numFiles            	11                  
-	numRows             	11                  
+	numFiles            	1                   
+	numRows             	4                   
 	parquet.compression 	zstd                
 #### A masked pattern was here ####
 	rawDataSize         	0                   
 	serialization.format	1                   
-	snapshot-count      	15
```
simhadri-g, line 198: nit: snapshot-count went from 15 to 2, is this expected?

difin: The files `iceberg_major_compaction_unpartitioned_ordered.q` and `iceberg_major_compaction_unpartitioned_ordered.q.out` are files of a new q-test that I am adding as part of this PR, but for unclear reason git thinks that these files are renames of `iceberg_optimize_table_unpartitioned.q` which was deleted in one of the previous PRs...

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
```diff
@@ -337,6 +337,7 @@ private static CompactionResponse requestCompaction(CompactionInfo ci, String ru
     }
     compactionRequest.setInitiatorVersion(ci.initiatorVersion);
     compactionRequest.setPoolName(ci.poolName);
+    compactionRequest.setOrderByClause(ci.orderByClause);
```
simhadri-g, line 340: Will enabling this compactionRequest.setOrderByClause affect hive acid table compaction as well?

difin: Yes, it affect Hive ACID compaction as well because compaction implementation is common for both ACID and Iceberg tables. But I believe this change is not dangerous because it just copies an order by clause from compaction info object to the compaction request which was missing; if the order by clause is not defined, it will set it as null. Hive ACID compaction also supports sorting. https://github.com/apache/hive/pull/4077

simhadri-g: Thats great. Thanks!

I was worried about the side effect this would have on hive acid compactions. This clears it up.

simhadri-g (approved):
LGTM +1

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorUtil.java
simhadri-g: Rest of the code looks good to me!

difin: Thanks!

PR #5458, opened 2024-09-19
HIVE-28512: retain properties in HiveConf.ConfVars.DDL_CTL_PARAMETERS…

jjiang037:
Retain properties in HiveConf.ConfVars.DDL_CTL_PARAMETERS_WHITELIST for CTLT query

Useful for avro base tables as the schema can evolve over time 

No

No

Unit test would be added.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
@@ -162,7 +162,7 @@ public class CreateTableLikeOperation extends DDLOperation<CreateTableLikeDesc> 
     if (paramsStr != null) {
       Set<String> retainer = new HashSet<String>(Arrays.asList(paramsStr.split(",")));
       for (String key : retainer) {
-        if (originalProperties.containsKey(key)) {
+        if (originalProperties.containsKey(key) && !tbl.getParameters().containsKey(key)) {
           tbl.getParameters().put(key, originalProperties.get(key));
         }
       }
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
@@ -155,17 +155,14 @@ public class CreateTableLikeOperation extends DDLOperation<CreateTableLikeDesc> 
     // with other query engines like mysql, redshift.
     originalProperties.putAll(tbl.getParameters());
     tbl.getParameters().clear();
-    if (desc.getTblProps() != null) {
-      tbl.setParameters(desc.getTblProps());
-    }
     String paramsStr = HiveConf.getVar(context.getConf(), HiveConf.ConfVars.DDL_CTL_PARAMETERS_WHITELIST);
     if (paramsStr != null) {
       Set<String> retainer = new HashSet<String>(Arrays.asList(paramsStr.split(",")));
-      for (String key : retainer) {
-        if (originalProperties.containsKey(key) && !tbl.getParameters().containsKey(key)) {
-          tbl.getParameters().put(key, originalProperties.get(key));
-        }
-      }
+      originalProperties.entrySet().stream().filter(entry -> retainer.contains(entry.getKey()))
+          .forEach(entry -> tbl.getParameters().put(entry.getKey(), entry.getValue()));
+    }
+    if (desc.getTblProps() != null) {
+      tbl.setParameters(desc.getTblProps());
     }
     HiveStorageHandler storageHandler = tbl.getStorageHandler();
     if (storageHandler != null) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
@@ -18,8 +18,7 @@
 
 package org.apache.hadoop.hive.ql.ddl.table.create.like;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
 
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.common.StatsSetupConst;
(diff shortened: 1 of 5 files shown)


ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
```diff
@@ -18,8 +18,7 @@
 
 package org.apache.hadoop.hive.ql.ddl.table.create.like;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
```
saihemanth-cloudera, line 21: nit: incorrect import

saihemanth-cloudera:
Can you add a test regarding why we need to retain `DDL_CTL_PARAMETERS_WHITELIST` property?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
```diff
@@ -155,6 +154,15 @@ private void setTableParameters(Table tbl, Map<String, String> originalPropertie
     if (desc.getTblProps() != null) {
       tbl.setParameters(desc.getTblProps());
     }
+    String paramsStr = HiveConf.getVar(context.getConf(), HiveConf.ConfVars.DDL_CTL_PARAMETERS_WHITELIST);
+    if (paramsStr != null) {
+        Set<String> retainer = new HashSet<String>(Arrays.asList(paramsStr.split(",")));
```
nareshpr, line 159: nit: 2 space

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
```diff
@@ -155,6 +158,15 @@ private void setTableParameters(Table tbl, Map<String, String> originalPropertie
     if (desc.getTblProps() != null) {
       tbl.setParameters(desc.getTblProps());
     }
+    String paramsStr = HiveConf.getVar(context.getConf(), HiveConf.ConfVars.DDL_CTL_PARAMETERS_WHITELIST);
```
dengzhhu653, line 161: nit: can we move this changes above 
```
  if (desc.getTblProps() != null) {
      tbl.setParameters(desc.getTblProps());
    }
```
and change it to:
```
 String paramsStr = HiveConf.getVar(context.getConf(), HiveConf.ConfVars.DDL_CTL_PARAMETERS_WHITELIST);
    if (paramsStr != null) {
      Set<String> retainer = new HashSet<String>(Arrays.asList(paramsStr.split(",")));
      originalProperties.entrySet().stream().filter(entry -> retainer.contains(entry.getKey()))
          .forEach(entry -> tbl.getParameters().put(entry.getKey(), entry.getValue()));
    }

File: ql/src/test/queries/clientpositive/create_like3.q [hunk 4]
```diff
@@ -0,0 +1,80 @@
...
+    "namespace":"dummy_namespace",
+    "fields":[
+      {"name":"field1","type":["null","string"],"default":null},
+      {"name":"field2","type":["null",{"type":"bytes","logicalType":"decimal","precision":10,"scale":3}],"default":null},
+      {"name":"field3","type":["null","string"],"default":null}
+    ]}'
+);
+
+SHOW CREATE TABLE dummy_table_1;
+CREATE EXTERNAL TABLE dummy_like_table_2 LIKE dummy_table_1;
```
dengzhhu653, line 35: what's this purpose? we have already checked this by `CREATE EXTERNAL TABLE dummy_like_table_1 LIKE dummy_table_1;`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
jjiang037: Will the issue arise when re-adding the whitelisted original properties from `originalProperties` back to the table, even if they've been explicitly overridden?

dengzhhu653: you can make this change after:
```
if (desc.getTblProps() != null) {
      tbl.setParameters(desc.getTblProps());
    }
    ```
so it will be overwritten if some are specified explicitly

File: ql/src/test/results/clientpositive/llap/create_like2.q.out [hunk 5]
```diff
@@ -40,7 +40,9 @@ Retention:          	0
 Table Type:         	MANAGED_TABLE       	 
 Table Parameters:	 	 
 	COLUMN_STATS_ACCURATE	{\"BASIC_STATS\":\"true\",\"COLUMN_STATS\":{\"a\":\"true\",\"b\":\"true\"}}
+	a                   	1
```
dengzhhu653, line 43: the test is failing, could you check it please?

ql/src/test/results/clientpositive/llap/create_like3.q.out
```diff
@@ -0,0 +1,375 @@
...
+  'org.apache.hadoop.hive.ql.io.avro.AvroContainerInputFormat' 
+OUTPUTFORMAT 
+  'org.apache.hadoop.hive.ql.io.avro.AvroContainerOutputFormat'
+LOCATION
+#### A masked pattern was here ####
+TBLPROPERTIES (
+  'bucketing_version'='2', 
+#### A masked pattern was here ####
+PREHOOK: query: ALTER TABLE dummy_table_1 SET TBLPROPERTIES (
+  'avro.schema.literal'='{
```
dengzhhu653, line 80: why the `avro.schema.literal` is present given we don't set the whitelist?

ql/src/test/results/clientpositive/llap/partition_discovery.q.out
```diff
@@ -510,6 +510,7 @@ Table Parameters:
 	COLUMN_STATS_ACCURATE	{\"BASIC_STATS\":\"true\"}
 	EXTERNAL            	TRUE                
 	bucketing_version   	2                   
+	discover.partitions 	false
```
dengzhhu653, line 513: why `discover.partitions` is changing here?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
```diff
@@ -148,13 +151,20 @@ private void setUserSpecifiedLocation(Table table) {
 
   private void setTableParameters(Table tbl, Map<String, String> originalProperties) throws HiveException {
     // With Hive-25813, we'll not copy over table properties from the source.
-    // CTLT should should copy column schema but not table properties. It is also consistent
+    // CTLT should copy column schema but not table properties. It is also consistent
     // with other query engines like mysql, redshift.
     originalProperties.putAll(tbl.getParameters());
     tbl.getParameters().clear();
     if (desc.getTblProps() != null) {
```
dengzhhu653, line 158: The `desc.getTblProps()` could be overwritten by the retain properties, is there any reason to do this?

dengzhhu653 (approved):

PR #5461, opened 2024-09-23
HIVE-28530: Set files in thread safe manner in HiveSequenceFileInputFormat

SourabhBadhya:
Set files in thread safe manner in HiveSequenceFileInputFormat

Set files in thread safe manner in HiveSequenceFileInputFormat

Setting files is not thread safe right now.

No

No

Concurrent tests run on a cluster with extremely high number of SELECT query execution.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveSequenceFileInputFormat.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/HiveSequenceFileInputFormat.java
@@ -21,6 +21,7 @@ package org.apache.hadoop.hive.ql.io;
 import java.io.IOException;
 import java.util.Set;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.hive.serde2.columnar.BytesRefArrayWritable;
 import org.apache.hadoop.io.LongWritable;
@@ -40,24 +41,27 @@ import org.apache.hadoop.mapred.SequenceFileInputFormat;
 public class HiveSequenceFileInputFormat<K extends LongWritable, V extends BytesRefArrayWritable>
     extends SequenceFileInputFormat<K, V> {
 
+  private static final ThreadLocal<Set<FileStatus>> fileStatuses = new ThreadLocal<>();
+
   public HiveSequenceFileInputFormat() {
     setMinSplitSize(SequenceFile.SYNC_INTERVAL);
   }
 
-  private Set<FileStatus> fileStatuses = null;
-
-  public void setFiles(Set<FileStatus> fileStatuses) {
-    this.fileStatuses= fileStatuses;
+  public void setFiles(Set<FileStatus> files) {
+    fileStatuses.set(files);
   }
 
   @Override
   protected FileStatus[] listStatus(JobConf job) throws IOException {
-    if(fileStatuses== null || fileStatuses.isEmpty()) {
-      // In cases where list of files to fetch is not provided we will use SequenceFileInputFormat
-      // e.g. SELECT without a job
-      return super.listStatus(job);
+    try {
+      if (CollectionUtils.isEmpty(fileStatuses.get())) {
+        // In cases where list of files to fetch is not provided we will use SequenceFileInputFormat
+        // e.g. SELECT without a job
+        return super.listStatus(job);
+      }
+      return fileStatuses.get().toArray(new FileStatus[0]);
+    } finally {
+      fileStatuses.remove();
     }
-    FileStatus[] fsStatusArray = new FileStatus[fileStatuses.size()];
-    return fileStatuses.toArray(fsStatusArray);
   }
 }


ql/src/java/org/apache/hadoop/hive/ql/io/HiveSequenceFileInputFormat.java
```diff
@@ -44,20 +44,20 @@ public HiveSequenceFileInputFormat() {
...
-    if(fileStatuses== null || fileStatuses.isEmpty()) {
+    if(fileStatuses.get() == null || fileStatuses.get().isEmpty()) {
       // In cases where list of files to fetch is not provided we will use SequenceFileInputFormat
       // e.g. SELECT without a job
       return super.listStatus(job);
     }
-    FileStatus[] fsStatusArray = new FileStatus[fileStatuses.size()];
-    return fileStatuses.toArray(fsStatusArray);
+    FileStatus[] fsStatusArray = new FileStatus[fileStatuses.get().size()];
+    return fileStatuses.get().toArray(fsStatusArray);
```
deniskuzZ, line 61: could be simplified to `fileStatuses.get().toArray(new FileStatus[0])`

ql/src/java/org/apache/hadoop/hive/ql/io/HiveSequenceFileInputFormat.java
```diff
@@ -44,20 +44,20 @@ public HiveSequenceFileInputFormat() {
     setMinSplitSize(SequenceFile.SYNC_INTERVAL);
   }
 
-  private Set<FileStatus> fileStatuses = null;
+  private ThreadLocal<Set<FileStatus>> fileStatuses = new ThreadLocal<>();
```
deniskuzZ, line 47: javadoc :
````
"Each thread holds an implicit reference to its copy of a thread-local variable as long as the thread is alive and the ThreadLocal instance is accessible; after a thread goes away, all of its copies of thread-local instances are subject to garbage collection (unless other references to these copies exist).
````
If your application uses a thread pool, to prevent memory  leaks call the `ThreadLocal.remove()`.

File: ql/src/java/org/apache/hadoop/hive/ql/io/HiveSequenceFileInputFormat.java (hunk 1 above)
Comment (SourabhBadhya) [reply ql/src/java/org/apache/hadoop/hive/ql/io/HiveSequenceFileInputFormat.java]: Done.

File: ql/src/java/org/apache/hadoop/hive/ql/io/HiveSequenceFileInputFormat.java [hunk 3]
```diff
@@ -44,20 +44,20 @@ public HiveSequenceFileInputFormat() {
...
 
   public void setFiles(Set<FileStatus> fileStatuses) {
-    this.fileStatuses= fileStatuses;
+    this.fileStatuses.set(fileStatuses);
   }
 
   @Override
   protected FileStatus[] listStatus(JobConf job) throws IOException {
-    if(fileStatuses== null || fileStatuses.isEmpty()) {
+    if(fileStatuses.get() == null || fileStatuses.get().isEmpty()) {
```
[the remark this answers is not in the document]
SourabhBadhya: Done

ql/src/java/org/apache/hadoop/hive/ql/io/HiveSequenceFileInputFormat.java
SourabhBadhya: Calling `remove()` after the usage of fileStatuses object in the `listStatus()` function. Done.

deniskuzZ:
````
public class HiveSequenceFileInputFormat<K extends LongWritable, V extends BytesRefArrayWritable>
  extends SequenceFileInputFormat<K, V> {

  private static final ThreadLocal<Set<FileStatus>> fileStatuses = new ThreadLocal<>();

  public HiveSequenceFileInputFormat() {
    setMinSplitSize(SequenceFile.SYNC_INTERVAL);
  }

  public void setFiles(Set<FileStatus> files) {
    fileStatuses.set(files);
  }

  @Override
  protected FileStatus[] listStatus(JobConf job) throws IOException {
    try {
      if (CollectionUtils.isEmpty(fileStatuses.get())) {
        // In cases where list of files to fetch is not provided we will use SequenceFileInputFormat
        // e.g. SELECT without a job
        return super.listStatus(job);
      }
      return fileStatuses.get().toArray(new FileStatus[0]);
    } finally {
      fileStatuses.remove();
    }
  }
}
````

SourabhBadhya: @deniskuzZ I think its called only once. Checked it here - https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapred/FileInputFormat.java#L334

deniskuzZ: 👍 
1 note, remove this from static var

deniskuzZ (approved):
LGTM +1, pending tests

PR #5462, opened 2024-09-23
HIVE-28532 Map Join Reuse cache allows to share hashtables for different join types

ramesh0201:
Map join container cache key should be unique for each map join hash table if they are of different join types. This will prevent hash table from getting shared between different join types. 

Sharing same hash table between different join types will not work because they expect different hash table type based on the join type. Otherwise we might run into ClassCastException and also potential wrong results in certain cases.

No

No

cd ql
mvn test -Dtest=TestSharedWorkOptimizer#testMapJoinCacheReuse

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
@@ -36,6 +36,8 @@ import java.util.Objects;
 import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.Set;
+
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hive.common.TableName;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
@@ -1006,6 +1008,16 @@ public class SharedWorkOptimizer extends Transform {
     if (mapJoinOp1.getConf().getPosBigTable() != mapJoinOp2.getConf().getPosBigTable()) {
       return false;
     }
+
+    // Map Joins when vectorized can have different formats for the hash tables built on the small table.
+    // Reusing hash tables between different join type can lead to ClassCastException or even wrong results.
+    if (ArrayUtils.isNotEmpty(mapJoinOp1.getConf().getConds())
+        && ArrayUtils.isNotEmpty(mapJoinOp2.getConf().getConds())
+        && mapJoinOp1.getConf().getConds()[0].getType() != mapJoinOp2.getConf().getConds()[0].getType()
+        && (mapJoinOp1.getConf().isNoOuterJoin() || mapJoinOp2.getConf().isNoOuterJoin())) {
+        return false;
+    }
+
 
     for (int i = 0; i < mapJoinOp1.getNumParent(); i ++) {
       if (i == mapJoinOp1.getConf().getPosBigTable()) {
--- a/ql/src/test/org/apache/hadoop/hive/ql/optimizer/TestSharedWorkOptimizer.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/optimizer/TestSharedWorkOptimizer.java
@@ -27,12 +27,14 @@ import org.apache.hadoop.hive.ql.exec.TableScanOperator;
 import org.apache.hadoop.hive.ql.metadata.Table;
 import org.apache.hadoop.hive.ql.optimizer.SharedWorkOptimizer.SharedWorkOptimizerCache;
 import org.apache.hadoop.hive.ql.optimizer.SharedWorkOptimizer.TSComparator;
+import org.apache.hadoop.hive.ql.parse.JoinType;
 import org.apache.hadoop.hive.ql.parse.SemanticException;
 import org.apache.hadoop.hive.ql.plan.ExprNodeColumnDesc;
 import org.apache.hadoop.hive.ql.plan.ExprNodeConstantDesc;
 import org.apache.hadoop.hive.ql.plan.ExprNodeDesc;
 import org.apache.hadoop.hive.ql.plan.ExprNodeGenericFuncDesc;
 import org.apache.hadoop.hive.ql.plan.FilterDesc;
+import org.apache.hadoop.hive.ql.plan.JoinCondDesc;
 import org.apache.hadoop.hive.ql.plan.MapJoinDesc;
 import org.apache.hadoop.hive.ql.plan.OperatorDesc;
 import org.apache.hadoop.hive.ql.plan.ReduceSinkDesc;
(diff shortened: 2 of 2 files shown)


ql/src/test/org/apache/hadoop/hive/ql/optimizer/TestSharedWorkOptimizer.java
```diff
@@ -323,39 +314,72 @@ public void testMapJoinCacheReuse() {
...
+            JoinType.RIGHTOUTER.ordinal(), JoinType.LEFTOUTER.ordinal(), true);
+
+    // case 10. same as case4 with full outer join and inner join
+    setupJoinOperatorsAndTest(ts1, ts2, rsA1, rsA2, rsB1, rsB2, 0, 0,
+            JoinType.FULLOUTER.ordinal(), JoinType.INNER.ordinal(), false);
+  }
 
-    runMapJoinCacheReuseOptimization(mapJoin5, mapJoin6);
-    assertNotEquals(mapJoin5.getConf().getCacheKey(), mapJoin6.getConf().getCacheKey());
+  private void setupJoinOperatorsAndTest(Operator<?> join1Op1, Operator<?> join2Op1, Operator<?> join1Op2,
```
kasakrisz, line 357: This method can be simplified further if the `MapJoinOperator` creation is extracted to another method like
```
MapJoinOperator setupMapJoin(List<Operator<?>> joinSource, int joinType, int join1PosBigTable) {
  ...
  return mapJoinOperator;
}
```
Then it can be called twice to construct the two map join operators

ql/src/test/org/apache/hadoop/hive/ql/optimizer/TestSharedWorkOptimizer.java
```diff
@@ -323,39 +314,72 @@ public void testMapJoinCacheReuse() {
...
+    setupJoinOperatorsAndTest(ts1, ts2, rsA1, rsA2, rsB1, rsB2, 0, 0,
+        JoinType.INNER.ordinal(), JoinType.INNER.ordinal(), true);
 
-    MapJoinOperator mapJoin1 = (MapJoinOperator) OperatorFactory.getAndMakeChild(
-        cCtx, getMapJoinDesc(0), joinSource1);
-    MapJoinOperator mapJoin2 = (MapJoinOperator) OperatorFactory.getAndMakeChild(
-        cCtx, getMapJoinDesc(0), joinSource2);
+    // case 2. MapJoin1: (big, A, B), MapJoin2: (big, A, C)
+    setupJoinOperatorsAndTest(ts1, ts2, rsA1, rsA2, rsB1, rsC2, 0, 0,
+        JoinType.INNER.ordinal(), JoinType.INNER.ordinal(), false);
```
kasakrisz, line 322: It would be better to move each case to its test method because in the current form the context (ts1, smallTableA, rsA1, etc) of the tests is reused. The code or any test may makes changes on these objects and causes interference between test cases.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1006,6 +1006,16 @@ private boolean canShareBroadcastInputs(
       return false;
     }
 
+    // Map Joins when vectorized can have different formats for the hash tables built on the small table.
+    // Reusing hash tables between different join type can lead to ClassCastException or even wrong results.
+    if (mapJoinOp1.getConf().getConds() != null && mapJoinOp1.getConf().getConds().length > 0
+        && mapJoinOp2.getConf().getConds() != null && mapJoinOp2.getConf().getConds().length > 0
+        && mapJoinOp1.getConf().getConds()[0].getType() != mapJoinOp2.getConf().getConds()[0].getType()
+        && (mapJoinOp1.getConf().isNoOuterJoin() || mapJoinOp2.getConf().isNoOuterJoin())) {
```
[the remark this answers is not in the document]
ramesh0201: Yes, if they are both outer joins, we still generate the same hash table type.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/TestSharedWorkOptimizer.java
```diff
@@ -323,39 +314,72 @@ public void testMapJoinCacheReuse() {
...
+
+    // case 10. same as case4 with full outer join and inner join
+    setupJoinOperatorsAndTest(ts1, ts2, rsA1, rsA2, rsB1, rsB2, 0, 0,
+            JoinType.FULLOUTER.ordinal(), JoinType.INNER.ordinal(), false);
+  }
 
-    runMapJoinCacheReuseOptimization(mapJoin5, mapJoin6);
-    assertNotEquals(mapJoin5.getConf().getCacheKey(), mapJoin6.getConf().getCacheKey());
+  private void setupJoinOperatorsAndTest(Operator<?> join1Op1, Operator<?> join2Op1, Operator<?> join1Op2,
+                                         Operator<?> join2Op2, Operator<?> join1Op3, Operator<?> join2Op3,
```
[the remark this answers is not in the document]
ramesh0201: Modified this.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/TestSharedWorkOptimizer.java
ramesh0201: Addressed this

ql/src/test/org/apache/hadoop/hive/ql/optimizer/TestSharedWorkOptimizer.java
ramesh0201: Addressed this

ql/src/test/org/apache/hadoop/hive/ql/optimizer/TestSharedWorkOptimizer.java
```diff
@@ -27,18 +27,9 @@
...
-import org.apache.hadoop.hive.ql.plan.ExprNodeDesc;
-import org.apache.hadoop.hive.ql.plan.ExprNodeGenericFuncDesc;
-import org.apache.hadoop.hive.ql.plan.FilterDesc;
-import org.apache.hadoop.hive.ql.plan.MapJoinDesc;
-import org.apache.hadoop.hive.ql.plan.OperatorDesc;
-import org.apache.hadoop.hive.ql.plan.ReduceSinkDesc;
-import org.apache.hadoop.hive.ql.plan.Statistics;
-import org.apache.hadoop.hive.ql.plan.TableDesc;
-import org.apache.hadoop.hive.ql.plan.TableScanDesc;
+import org.apache.hadoop.hive.ql.plan.*;
```
deniskuzZ, line 32: star import

deniskuzZ:
@kasakrisz, do you know if there is anything left here? could this be merged?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1007,6 +1007,16 @@ private boolean canShareBroadcastInputs(
       return false;
     }
 
+    // Map Joins when vectorized can have different formats for the hash tables built on the small table.
+    // Reusing hash tables between different join type can lead to ClassCastException or even wrong results.
+    if (mapJoinOp1.getConf().getConds() != null && mapJoinOp1.getConf().getConds().length > 0
```
deniskuzZ, line 1012: minor: could be simplified with util method:
````
ArrayUtils.isNotEmpty(mapJoinOp1.getConf().getConds())
````

deniskuzZ (approved):
+1

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
ramesh0201: Thank you! Modified this

PR #5465, opened 2024-09-24
HIVE-28534: Improve HMS Client Exception Handling for Hive-3

sercanCyberVision:
**MOTIVATION**
When the HMS client fails to connect to the server due to a `TTransportException`, there is no issue with error reporting.

However, when the failure is caused by an `IOException`, the exception object, which is used for reporting purposes, remains null. As a result, it does not properly capture the root cause, and end-users encounter an unrelated NPE, masking the actual issue.
```
Exception in thread "main" java.lang.AssertionError: Unable to connect to HMS!
	at TestHMS.main(TestHMS.java:20)
Caused by: java.lang.NullPointerException
	at org.apache.hadoop.util.StringUtils.stringifyException(StringUtils.java:90)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.open(HiveMetaStoreClient.java:613)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.<init>(HiveMetaStoreClient.java:233)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.<init>(HiveMetaStoreClient.java:145)
	at TestHMS.main(TestHMS.java:13)
```

**SOLUTION**
I have added an `IOException` object to capture the corresponding exception during connection attempts. If the `IOException` is not null, it will now be output, providing clearer information about the failure.

**CHECKS**
After applying the patch, instead of an NPE, I now receive a proper error message:
```
Exception in thread "main" java.lang.AssertionError: Unable to connect to HMS!
	at TestHMS.main(TestHMS.java:20)
Caused by: MetaException(message:Could not connect to meta store using any of the URIs provided. Most recent failure: 
javax.security.auth.login.LoginException: Unable to obtain MapR credentials)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.open(HiveMetaStoreClient.java:618)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.<init>(HiveMetaStoreClient.java:233)
	at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.<init>(HiveMetaStoreClient.java:145)
	at TestHMS.main(TestHMS.java:13)
```

### DIFF
--- a/standalone-metastore/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
+++ b/standalone-metastore/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
@@ -429,6 +429,7 @@ public class HiveMetaStoreClient implements IMetaStoreClient, AutoCloseable {
   private void open() throws MetaException {
     isConnected = false;
     TTransportException tte = null;
+    IOException ioException = null;
     boolean useSSL = MetastoreConf.getBoolVar(conf, ConfVars.USE_SSL);
     boolean useSasl = MetastoreConf.getBoolVar(conf, ConfVars.USE_THRIFT_SASL);
     boolean useFramedTransport = MetastoreConf.getBoolVar(conf, ConfVars.USE_THRIFT_FRAMED_TRANSPORT);
@@ -456,6 +457,7 @@ public class HiveMetaStoreClient implements IMetaStoreClient, AutoCloseable {
                   trustStorePath, trustStorePassword );
               LOG.info("Opened an SSL connection to metastore, current connections: " + connCount.incrementAndGet());
             } catch(IOException e) {
+              ioException = e;
               throw new IllegalArgumentException(e);
             } catch(TTransportException e) {
               tte = e;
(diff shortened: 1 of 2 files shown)


standalone-metastore/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -456,6 +457,7 @@ private void open() throws MetaException {
                   trustStorePath, trustStorePassword );
               LOG.info("Opened an SSL connection to metastore, current connections: " + connCount.incrementAndGet());
             } catch(IOException e) {
+              ioException = e;
```
difin, line 460: This assignment is excessive cause it's not doing anything after the IllegalArgumentException is thrown on the next line.

sercanCyberVision: Thank you @difin for the review.
You're right! 

if the `IllegalArgumentException` isn't explicitly intended to terminate reattempts, I can change it to `MetaException` to log the error message and allow the connection process to make another attempt.

If not, I can remove the assignment.

sercanCyberVision: @deniskuzZ, could you please review the options mentioned above and let us know which one you think we should proceed with?

deniskuzZ: Hive-3.x should be declared. EOL soon

PR #5466, opened 2024-09-24
HIVE-28533: Fix compaction with custom pools

difin:

Fix for getting matching configs based on a regex pattern in HiveConf. The bug is that the matcher returns the whole pattern instead of the first matching group as was expected. This functionality is used in compaction when assigning compaction workers to custom pools. On startup, HS2 retrieves custom pool names from the configs that are having naming format such as 'hive.compactor.worker.{poolName}.threads'.

To fix the compaction functionality of assigning compaction workers to custom pools.

No

New unit test was added;
Manually tested on an internal cloud environment.

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -7247,7 +7247,7 @@ public class HiveConf extends Configuration {
     for (Map.Entry<String, String> entry : this) {
       Matcher matcher = regex.matcher(entry.getKey());
       if (matcher.matches()) {
-        matchingEntries.add(new AbstractMap.SimpleEntry<>(entry.getKey(), matcher.group(0)));
+        matchingEntries.add(new AbstractMap.SimpleEntry<>(entry.getKey(), matcher.group(1)));
       }
     }
     return matchingEntries;
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -62,7 +62,6 @@ import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.time.ZoneId;
-import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -7241,15 +7241,4 @@ public class HiveConf extends Configuration {
       set(e.getKey(), e.getValue());
     }
   }
-
-  public List<Map.Entry<String, String>> getMatchingEntries(Pattern regex) {
-    List<Map.Entry<String, String>> matchingEntries = new ArrayList<>();
-    for (Map.Entry<String, String> entry : this) {
-      Matcher matcher = regex.matcher(entry.getKey());
-      if (matcher.matches()) {
-        matchingEntries.add(new AbstractMap.SimpleEntry<>(entry.getKey(), matcher.group(1)));
-      }
-    }
-    return matchingEntries;
-  }
 }
--- a/common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java
+++ b/common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java
@@ -31,8 +31,6 @@ import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 
--- a/common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java
+++ b/common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java
@@ -254,16 +254,4 @@ public class TestHiveConf {
     f2.delete();
     fileBakHiveSite.delete();
   }
-
-  @Test
-  public void testGetMatchingEntries() {
-    HiveConf conf = new HiveConf();
-    conf.setInt("hive.compactor.worker.iceberg.threads", 4);
-
-    List<Map.Entry<String, String>> entries = conf.getMatchingEntries(Constants.COMPACTION_POOLS_PATTERN);
-    
-    Assert.assertEquals(entries.size(), 1);
-    Assert.assertEquals(entries.get(0).getKey(), "hive.compactor.worker.iceberg.threads");
-    Assert.assertEquals(entries.get(0).getValue(), "iceberg");
-  }
 }
--- a/common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java
+++ b/common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java
@@ -31,6 +31,8 @@ import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 
@@ -251,5 +253,17 @@ public class TestHiveConf {
     FileUtils.copyFile(fileBakHiveSite, fileHiveSite);
     f2.delete();
     fileBakHiveSite.delete();
+  }
+
+  @Test
+  public void testGetMatchingEntries() {
+    HiveConf conf = new HiveConf();
+    conf.setInt("hive.compactor.worker.iceberg.threads", 4);
+
+    List<Map.Entry<String, String>> entries = conf.getMatchingEntries(Constants.COMPACTION_POOLS_PATTERN);
+    
+    Assert.assertEquals(entries.size(), 1);
+    Assert.assertEquals(entries.get(0).getKey(), "hive.compactor.worker.iceberg.threads");
+    Assert.assertEquals(entries.get(0).getValue(), "iceberg");
   }
 }
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
@@ -190,14 +190,8 @@ public class CompactionPoolOnTezTest extends CompactorOnTezTest {
     Map<String, Integer> customPools = CompactorUtil.getPoolConf(conf);
     Assert.assertEquals(customPools.size(), 1);
     Assert.assertEquals(customPools.get("pool1"), new Integer(1));
-    
-    Worker worker = new Worker();
-    worker.setConf(conf);
-    worker.setPoolName("pool1");
-    worker.init(new AtomicBoolean(true));
 
-    worker.run();
-
+    TxnCommandsBaseForTests.runWorker(conf, "pool1");
     checkCompactionRequest("ready for cleaning", "pool1");
   }
 }
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
@@ -17,11 +17,8 @@
  */
 package org.apache.hadoop.hive.ql.txn.compactor;
 
-import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hive.conf.Constants;
 import org.apache.hadoop.hive.conf.HiveConf;
-import org.apache.hadoop.hive.metastore.api.CompactionRequest;
-import org.apache.hadoop.hive.metastore.api.CompactionType;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.ShowCompactRequest;
 import org.apache.hadoop.hive.metastore.api.ShowCompactResponse;
@@ -29,7 +26,6 @@ import org.apache.hadoop.hive.metastore.api.ShowCompactResponseElement;
 import org.apache.hadoop.hive.metastore.txn.TxnStore;
 import org.apache.hadoop.hive.metastore.txn.TxnUtils;
 import org.apache.hadoop.hive.ql.TxnCommandsBaseForTests;
-import org.apache.hadoop.hive.ql.processors.CommandProcessorResponse;
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -37,9 +33,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
-import java.io.*;
 
 import static org.apache.hadoop.hive.ql.txn.compactor.TestCompactorBase.executeStatementOnDriver;
 
@@ -176,22 +170,23 @@ public class CompactionPoolOnTezTest extends CompactorOnTezTest {
 
   @Test
   public void testCompactionWithCustomPool() throws Exception {
-    conf.setInt("hive.compactor.worker.pool1.threads", 1);
+    String poolName = "pool1";
+    conf.setInt(String.format("hive.compactor.worker.%s.threads", poolName), 1);
 
     Map<String, String> properties = new HashMap<>();
-    properties.put(Constants.HIVE_COMPACTOR_WORKER_POOL, "pool1");
+    properties.put(Constants.HIVE_COMPACTOR_WORKER_POOL, poolName);
     provider.createFullAcidTable(null, DEFAULT_TABLE_NAME, false, false, properties);
     provider.insertTestData(DEFAULT_TABLE_NAME, false);
 
     TxnCommandsBaseForTests.runInitiator(conf);
 
-    checkCompactionRequest("initiated", "pool1");
+    checkCompactionRequest("initiated", poolName);
 
     Map<String, Integer> customPools = CompactorUtil.getPoolConf(conf);
     Assert.assertEquals(customPools.size(), 1);
-    Assert.assertEquals(customPools.get("pool1"), new Integer(1));
+    Assert.assertEquals(customPools.get(poolName), new Integer(1));
 
-    TxnCommandsBaseForTests.runWorker(conf, "pool1");
-    checkCompactionRequest("ready for cleaning", "pool1");
+    TxnCommandsBaseForTests.runWorker(conf, poolName);
+    checkCompactionRequest("ready for cleaning", poolName);
   }
 }
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
@@ -20,6 +20,8 @@ package org.apache.hadoop.hive.ql.txn.compactor;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hive.conf.Constants;
 import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.metastore.api.CompactionRequest;
+import org.apache.hadoop.hive.metastore.api.CompactionType;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.ShowCompactRequest;
 import org.apache.hadoop.hive.metastore.api.ShowCompactResponse;
@@ -35,6 +37,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
 import java.io.*;
 
@@ -171,4 +174,30 @@ public class CompactionPoolOnTezTest extends CompactorOnTezTest {
     Assert.assertTrue(p.matcher(results.get(1).toString()).matches());
   }
 
+  @Test
+  public void testCompactionWithCustomPool() throws Exception {
+    conf.setInt("hive.compactor.worker.pool1.threads", 1);
+
+    Map<String, String> properties = new HashMap<>();
+    properties.put(Constants.HIVE_COMPACTOR_WORKER_POOL, "pool1");
+    provider.createFullAcidTable(null, DEFAULT_TABLE_NAME, false, false, properties);
+    provider.insertTestData(DEFAULT_TABLE_NAME, false);
+
+    TxnCommandsBaseForTests.runInitiator(conf);
+
+    checkCompactionRequest("initiated", "pool1");
+
+    Map<String, Integer> customPools = CompactorUtil.getPoolConf(conf);
+    Assert.assertEquals(customPools.size(), 1);
+    Assert.assertEquals(customPools.get("pool1"), new Integer(1));
+    
+    Worker worker = new Worker();
+    worker.setConf(conf);
+    worker.setPoolName("pool1");
+    worker.init(new AtomicBoolean(true));
+
+    worker.run();
+
+    checkCompactionRequest("ready for cleaning", "pool1");
+  }
 }
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
@@ -183,8 +183,8 @@ public class CompactionPoolOnTezTest extends CompactorOnTezTest {
     checkCompactionRequest("initiated", poolName);
 
     Map<String, Integer> customPools = CompactorUtil.getPoolConf(conf);
-    Assert.assertEquals(customPools.size(), 1);
-    Assert.assertEquals(customPools.get(poolName), new Integer(1));
+    Assert.assertEquals(1, customPools.size());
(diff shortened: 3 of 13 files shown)


common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java
```diff
@@ -252,4 +254,16 @@ public void testAdditionalConfigFiles() throws Exception{
     f2.delete();
     fileBakHiveSite.delete();
   }
+
+  @Test
+  public void testGetMatchingEntries() {
+    HiveConf conf = new HiveConf();
+    conf.setInt("hive.compactor.worker.iceberg.threads", 4);
+
+    List<Map.Entry<String, String>> entries = conf.getMatchingEntries(Constants.COMPACTION_POOLS_PATTERN);
```
deniskuzZ, line 263: why is that a `List<Map.Entry<String, String>>` and not just Map<String, String>?

deniskuzZ, line 263: also, COMPACTION_POOLS_PATTERN is incorrect, please change to 
````
^hive\.compactor\.worker\.(.+)\.threads$
````

Comment (deniskuzZ) [reply common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java:263]: that is some crazy code, could you please refactor it with
````
  public Map<String, Integer> getPoolConfigEntries(HiveConf hiveConf) {
    Map<String, Integer> poolConfigs = new HashMap<>();
    for (Map.Entry<String, String> entry : hiveConf) {
      Matcher matcher = COMPACTION_POOLS_PATTERN.matcher(entry.getKey());
      if (matcher.matches()) {
        matchingEntries.put(matcher.group(0), NumberUtils.toInt(entry.getValue(), 0));
      }
    }
    return poolConfigs;
  }
````

File: common/src/java/org/apache/hadoop/hive/conf/HiveConf.java [hunk 2]
```diff
@@ -7247,7 +7247,7 @@ public List<Map.Entry<String, String>> getMatchingEntries(Pattern regex) {
     for (Map.Entry<String, String> entry : this) {
```
deniskuzZ, line 7247: that is some crazy code, could you please refactor it with below snippet and extract to HS2 (where it's used)
````
  public Map<String, Integer> getPoolConfigEntries(HiveConf hiveConf) {
    Map<String, Integer> poolConfigs = new HashMap<>();
    for (Map.Entry<String, String> entry : hiveConf) {
      Matcher matcher = COMPACTION_POOLS_PATTERN.matcher(entry.getKey());
      if (matcher.matches()) {
        matchingEntries.put(matcher.group(1), NumberUtils.toInt(entry.getValue(), 0));
      }
    }
    return poolConfigs;
  }
````
wonder how it was even tested?

Comment (difin) [reply common/src/java/org/apache/hadoop/hive/conf/HiveConf.java:7247]: Done

File: common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java (hunk 1 above)
Comment (difin) [reply common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java:263]: Fixed

File: common/src/test/org/apache/hadoop/hive/conf/TestHiveConf.java [hunk 3]
```diff
@@ -31,6 +31,8 @@
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.ArrayList;
+import java.util.List;
```
deniskuzZ, line 34: is that leftovers? i don't see any new code using them

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
```diff
@@ -171,4 +174,30 @@ public void testShowCompactionsRespectPoolName() throws Exception {
...
+
+    TxnCommandsBaseForTests.runInitiator(conf);
+
+    checkCompactionRequest("initiated", "pool1");
+
+    Map<String, Integer> customPools = CompactorUtil.getPoolConf(conf);
+    Assert.assertEquals(customPools.size(), 1);
+    Assert.assertEquals(customPools.get("pool1"), new Integer(1));
+    
+    Worker worker = new Worker();
```
deniskuzZ, line 194: could you use helper method: `TxnCommandsBaseForTests.runWorker(conf)`?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -1204,17 +1205,21 @@ private static void startHiveServer2() throws Throwable {
...
       sb.append("Global pool size: ").append(numWorkers).append("\n");
 
       LOG.info("Initializing the compaction pools with using the global worker limit: {} ", numWorkers);
-      while (numWorkers > 0 && entries.size() > 0) {
-        Map.Entry<String, String> entry = entries.remove(0);
-        String poolName = entry.getValue();
-        int poolWorkers = hiveConf.getInt(entry.getKey(), 0);
+      for (Map.Entry<String, Integer> customPool: customPools.entrySet()) {
+        
+        if (numWorkers <= 0) {
```
deniskuzZ, line 1217: why do we need this check inside the loop?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -1204,17 +1205,21 @@ private static void startHiveServer2() throws Throwable {
...
       StringBuilder sb = new StringBuilder(2048);
       sb.append("This HS2 instance will act as Compactor with the following worker pool configuration:\n");
       sb.append("Global pool size: ").append(numWorkers).append("\n");
 
       LOG.info("Initializing the compaction pools with using the global worker limit: {} ", numWorkers);
-      while (numWorkers > 0 && entries.size() > 0) {
-        Map.Entry<String, String> entry = entries.remove(0);
-        String poolName = entry.getValue();
-        int poolWorkers = hiveConf.getInt(entry.getKey(), 0);
+      for (Map.Entry<String, Integer> customPool: customPools.entrySet()) {
```
deniskuzZ, line 1215: can we use for-each here?
````
customPools.forEach((poolName, poolWorkers) -> {
      ....
})
````

File: service/src/java/org/apache/hive/service/server/HiveServer2.java [hunk 7]
```diff
@@ -1240,9 +1245,9 @@ private void maybeStartCompactorThreads(HiveConf hiveConf) throws Exception {
 
       if (numWorkers == 0) {
         LOG.warn("No default compaction pool configured, all non-labeled compaction requests will remain unprocessed!");
-        if (entries.size() > 0) {
-          for (Map.Entry<String, String> entry : entries) {
-            String poolName = entry.getValue();
+        if (customPools.size() > 0) {
+          for (Map.Entry<String, Integer> entry : customPools.entrySet()) {
```
deniskuzZ, line 1249:
````
customPools.keySet().forEach(poolName -> {
      
    });
````

File: itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java (hunk 4 above)
Comment (difin) [reply itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java:194]: Done

File: service/src/java/org/apache/hive/service/server/HiveServer2.java (hunk 5 above)
Comment (difin) [reply service/src/java/org/apache/hive/service/server/HiveServer2.java:1217]: Because in each iteration it subtracts `poolWorkers` from `numWorkers` and the loop execution needs to stop when there are no more workers left. `poolWorkers` is the number of custom pool workers and `numWorkers` is the global limit of workers.

```
String poolName = customPool.getKey();
int poolWorkers = customPool.getValue();
...
numWorkers -= poolWorkers;
```

File: service/src/java/org/apache/hive/service/server/HiveServer2.java (hunk 7 above)
Comment (difin) [reply service/src/java/org/apache/hive/service/server/HiveServer2.java:1249]: I refactored this part

File: itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java [hunk 8]
```diff
@@ -171,4 +174,24 @@ public void testShowCompactionsRespectPoolName() throws Exception {
...
+    conf.setInt("hive.compactor.worker.pool1.threads", 1);
+
+    Map<String, String> properties = new HashMap<>();
+    properties.put(Constants.HIVE_COMPACTOR_WORKER_POOL, "pool1");
+    provider.createFullAcidTable(null, DEFAULT_TABLE_NAME, false, false, properties);
+    provider.insertTestData(DEFAULT_TABLE_NAME, false);
+
+    TxnCommandsBaseForTests.runInitiator(conf);
+
+    checkCompactionRequest("initiated", "pool1");
```
deniskuzZ, line 188: would be nice if we define a localvar, however, if not other issues we can ignore this
````
String poolName = "pool1"
````

File: ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorThread.java [hunk 9]
```diff
@@ -111,12 +111,15 @@ protected String tableName(Table t) {
...
-    thread.init(new AtomicBoolean());
-    thread.start();
+  public static void initializeAndStartThread(CompactorThread thread, Configuration conf) {
+    try {
+      LOG.info("Starting compactor thread of type " + thread.getClass().getName());
+      thread.setConf(conf);
+      thread.init(new AtomicBoolean());
+      thread.start(); 
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
deniskuzZ, line 121: why is this change? is there some compaction related runtime exception we could use?

ql/src/test/org/apache/hadoop/hive/ql/TxnCommandsBaseForTests.java
```diff
@@ -246,6 +246,13 @@ public static void runInitiator(HiveConf hiveConf) throws Exception {
   public static void runWorker(HiveConf hiveConf) throws Exception {
     runCompactorThread(hiveConf, CompactorThreadType.WORKER);
   }
+  public static void runWorker(HiveConf hiveConf, String poolName) throws Exception {
```
deniskuzZ, line 249: why do you need this method. can we add  
````
((Worker)t).setPoolName(poolName);
````
in `runCompactorThread` and keep the original signature with `default` value?

File: service/src/java/org/apache/hive/service/server/HiveServer2.java [hunk 11]
```diff
@@ -1201,63 +1204,63 @@ private static void startHiveServer2() throws Throwable {
     }
   }
 
-  private void maybeStartCompactorThreads(HiveConf hiveConf) throws Exception {
+  public Map<String, Integer> maybeStartCompactorThreads(HiveConf hiveConf) {
+    Map<String, Integer> startedWorkers = new HashMap<>();
     if (MetastoreConf.getVar(hiveConf, MetastoreConf.ConfVars.HIVE_METASTORE_RUNWORKER_IN).equals("hs2")) {
-      int numWorkers = MetastoreConf.getIntVar(hiveConf, MetastoreConf.ConfVars.COMPACTOR_WORKER_THREADS);
-      List<Map.Entry<String, String>> entries = hiveConf.getMatchingEntries(Constants.COMPACTION_POOLS_PATTERN);
+      AtomicInteger numWorkers = new AtomicInteger(MetastoreConf.getIntVar(hiveConf, MetastoreConf.ConfVars.COMPACTOR_WORKER_THREADS));
```
deniskuzZ, line 1210: why do we need AtomicInteger, it shouldn't be executed only in HMS main thread

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -1201,63 +1204,63 @@ private static void startHiveServer2() throws Throwable {
...
         if (poolWorkers == 0) {
           LOG.warn("Compaction pool ({}) configured with zero workers. Skipping pool initialization", poolName);
           sb.append("Pool not initialized, 0 size: ").append(poolName).append("\n");
-          continue;
         }
-        if (poolWorkers > numWorkers) {
-          LOG.warn("Global worker pool exhausted, compaction pool ({}) will be configured with less workers than the " +
-              "required number. ({} -> {})", poolName, poolWorkers, numWorkers);
-          poolWorkers = numWorkers;
+        else if (numWorkers.get() == 0) {
```
deniskuzZ, line 1223: that should be an IllegalArgumentException

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -1201,63 +1204,63 @@ private static void startHiveServer2() throws Throwable {
...
-          CompactorThread.initializeAndStartThread(w, hiveConf);
-          sb.append("Worker - Name: ").append(w.getName()).append(", Pool: ").append(poolName)
-              .append(", Priority: ").append(w.getPriority()).append("\n");
+          LOG.info("Initializing compaction pool ({}) with {} workers.", poolName, poolWorkers);
+          IntStream.range(0, poolWorkers).forEach(i -> {
+            Worker w = new Worker();
+            w.setPoolName(poolName);
+            CompactorThread.initializeAndStartThread(w, hiveConf);
+            startedWorkers.put(poolName, startedWorkers.getOrDefault(poolName, 0)+1);
+            sb.append("Worker - Name: ").append(w.getName()).append(", Pool: ").append(poolName)
```
deniskuzZ, line 1240: not sure, but wouldn't this be more readable
````
sb.append(
String.format("Worker - Name: %s, Pool: %s, Priority: %i, w.getName(), poolName,w.getPriority() ));
````

File: service/src/java/org/apache/hive/service/server/HiveServer2.java [hunk 14]
```diff
@@ -1201,63 +1204,63 @@ private static void startHiveServer2() throws Throwable {
...
+                .append(", Priority: ").append(w.getPriority()).append("\n");
+          });
+          numWorkers.set(numWorkers.get() - poolWorkers);
```
deniskuzZ, line 1243: with Ref<Integer> you can do 
````
numWorkers.value -= poolWorkers
````

File: service/src/java/org/apache/hive/service/server/HiveServer2.java [hunk 15]
```diff
@@ -1201,63 +1204,63 @@ private static void startHiveServer2() throws Throwable {
...
           sb.append("Pool not initialized, no remaining free workers: default\n");
         }
       } else {
-        LOG.info("Initializing default compaction pool with {} workers.", numWorkers);
-        for (int i = 0; i < numWorkers; i++) {
+        LOG.info("Initializing default compaction pool with {} workers.", numWorkers.get());
+        IntStream.range(0, numWorkers.get()).forEach(i -> {
           Worker w = new Worker();
           CompactorThread.initializeAndStartThread(w, hiveConf);
+          startedWorkers.put(Constants.COMPACTION_DEFAULT_POOL, startedWorkers.getOrDefault(Constants.COMPACTION_DEFAULT_POOL, 0)+1);
```
deniskuzZ, line 1257: use compute instead
````
startedWorkers.compute(Constants.COMPACTION_DEFAULT_POOL, (k, v) -> (v == null) ? 0 : v + 1)
````

File: service/src/java/org/apache/hive/service/server/HiveServer2.java [hunk 16]
```diff
@@ -1201,63 +1204,63 @@ private static void startHiveServer2() throws Throwable {
...
-          w.setPoolName(poolName);
-          CompactorThread.initializeAndStartThread(w, hiveConf);
-          sb.append("Worker - Name: ").append(w.getName()).append(", Pool: ").append(poolName)
-              .append(", Priority: ").append(w.getPriority()).append("\n");
+          LOG.info("Initializing compaction pool ({}) with {} workers.", poolName, poolWorkers);
+          IntStream.range(0, poolWorkers).forEach(i -> {
+            Worker w = new Worker();
+            w.setPoolName(poolName);
+            CompactorThread.initializeAndStartThread(w, hiveConf);
+            startedWorkers.put(poolName, startedWorkers.getOrDefault(poolName, 0)+1);
```
deniskuzZ, line 1239: use `compute` instead

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/CompactionPoolOnTezTest.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorThread.java
difin: This is called only from `HiveServer2#maybeStartCompactorThreads`, the method that I changed in this PR. 
It is called from within `customPools.forEach(...)` and checked exceptions in functional code are problematic in Java. In reality, a checked exception can never be thrown in this `CompactorThread#initializeAndStartThread` method. It's declared in `CompactorThread#init`, but it is not thrown there. It is done only because classes that extend `CompactorThread` throw exceptions from the `init` method.

ql/src/test/org/apache/hadoop/hive/ql/TxnCommandsBaseForTests.java
difin: Done

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: This was in the existing code, but fixed as you suggested.

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: Done

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: Done

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: I think this is not related to preventing compaction for selected db/tables.
If we do IllegalArgumentException here, it will be different from the intended design as described here:
https://cwiki.apache.org/confluence/display/Hive/Compaction+pooling

"_Worker allocation
The already existing hive.compactor.worker.threads configuration value holds the maximum number of worker threads. The worker allocation happens as follows:

Labeled pools are initialized in a sequential manner with random ordering.
Each pool decreases the number of available workers by its own worker count.
If the number of assignable workers is less than the configured one, the pool size will be adjusted (In other words: if the requested pool size is 5 but there are only 3 workers remaining, than the pool size will be decreased to 3).
**If the number of assignable workers is 0, the pool won’t be initialized.**
All remaining workers not used up by the labeled pools, are assigned to the default pool._"

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: Done

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: Done

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorThread.java
deniskuzZ: could we remove then throw from `CompactorThread#initializeAndStartThread/init`?

ql/src/test/org/apache/hadoop/hive/ql/TxnCommandsBaseForTests.java
```diff
@@ -241,17 +242,20 @@ public static String makeValuesClause(int[][] rows) {
   }
 
   public static void runInitiator(HiveConf hiveConf) throws Exception {
-    runCompactorThread(hiveConf, CompactorThreadType.INITIATOR);
+    runCompactorThread(hiveConf, CompactorThreadType.INITIATOR, Constants.COMPACTION_DEFAULT_POOL);
```
deniskuzZ, line 245: no need for changes here, just keep the original signature and use defauls
````
private static void runCompactorThread(HiveConf hiveConf, CompactorThreadType type) {
    runCompactorThread(hiveConf, type, Constants.COMPACTION_DEFAULT_POOL)
}
````

File: ql/src/test/org/apache/hadoop/hive/ql/TxnCommandsBaseForTests.java [hunk 18]
```diff
@@ -241,17 +242,20 @@ public static String makeValuesClause(int[][] rows) {
...
   }
   public static void runWorker(HiveConf hiveConf) throws Exception {
-    runCompactorThread(hiveConf, CompactorThreadType.WORKER);
+    runCompactorThread(hiveConf, CompactorThreadType.WORKER, Constants.COMPACTION_DEFAULT_POOL);
```
deniskuzZ, line 248:
````
runWorker(hiveConf, Constants.COMPACTION_DEFAULT_POOL)
````

File: ql/src/test/org/apache/hadoop/hive/ql/TxnCommandsBaseForTests.java [hunk 19]
```diff
@@ -261,6 +265,9 @@ private static void runCompactorThread(HiveConf hiveConf, CompactorThreadType ty
         break;
       case WORKER:
         t = new Worker();
+        if (!poolName.equals(Constants.COMPACTION_DEFAULT_POOL)) {
```
deniskuzZ, line 268: not safe, could be null
````
!Objects.equals(poolName, Constants.COMPACTION_DEFAULT_POOL)
````

File: service/src/java/org/apache/hive/service/server/HiveServer2.java [hunk 20]
```diff
@@ -1201,63 +1204,64 @@ private static void startHiveServer2() throws Throwable {
...
           sb.append("Pool not initialized, 0 size: ").append(poolName).append("\n");
-          continue;
         }
-        if (poolWorkers > numWorkers) {
-          LOG.warn("Global worker pool exhausted, compaction pool ({}) will be configured with less workers than the " +
-              "required number. ({} -> {})", poolName, poolWorkers, numWorkers);
-          poolWorkers = numWorkers;
+        else if (numWorkers.value == 0) {
+          LOG.warn("There are no available workers for the following compaction pool: {} ", poolName);
+          sb.append("Pool not initialized, no remaining free workers: ").append(poolName).append("\n");
```
deniskuzZ, line 1225: i would just log this line instead of the prev one. Als,o why do we need to add this into `sb`?

deniskuzZ (commented):
LGTM, some minor things

ql/src/test/org/apache/hadoop/hive/ql/TxnCommandsBaseForTests.java
difin: Done

ql/src/test/org/apache/hadoop/hive/ql/TxnCommandsBaseForTests.java
difin: Done

ql/src/test/org/apache/hadoop/hive/ql/TxnCommandsBaseForTests.java
difin: Right, fixed.

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: This was added into sb in the existing code.
I removed adding to sb and logged warning instead.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorThread.java
difin: I was wrong saying that the exception can never be thrown from here because this method is called with the Worker object type which can throw exceptions in the init method. So I added a new specialized exception class that extends RuntimeException and used it here.

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: Yes, I just pushed new commit

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/CompactorThread.java
deniskuzZ: ok, but could we wrap with try-catch only init, not the whole method body

difin: Yes, done.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/RuntimeCompactionException.java
```diff
@@ -0,0 +1,92 @@
...
+
+package org.apache.hadoop.hive.ql.txn.compactor;
+
+import org.apache.hadoop.hive.common.classification.InterfaceAudience;
+import org.apache.hadoop.hive.common.classification.InterfaceStability;
+import org.apache.hadoop.hive.ql.ErrorMsg;
+
+@InterfaceAudience.Public
+@InterfaceStability.Stable
+public class RuntimeCompactionException extends RuntimeException {
```
deniskuzZ, line 27: please rename to simply `CompactionException`

deniskuzZ (approved):
+1, pending tests

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/RuntimeCompactionException.java
difin: Done

difin:
> +1, pending tests

Tests passed. Thanks, @deniskuzZ!

deniskuzZ:
@difin i was about to merge, but noticed number of sonar violations, could you please address only ones created by new code: https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=5466&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

difin:
@deniskuzZ, I fixed all Sonar violations that could be fixed.

PR #5467, opened 2024-09-26
HIVE-21481: MERGE correctness issues with null safe equality

kasakrisz:
Merge statements in Hive are rewritten to multi-insert statements (Merge on Read) or insert select statements where the select clause has union all branches. In the case of both approaches the target and the source tables must be joined to find records that need to be inserted/updated/deleted. This is performed by outer join in most cases. Then the joined records are pushed to the insert/select branches (MOR) or union branches (COW). Each operation (I/U/D) has its branch with a where clause to filter the records only relevant to the branch. In the current implementation, this filter predicate is inherited from the join condition: a record should be inserted or updated depending on whether the target tables' join keys have null values or not in the joined record. This logic can not handle the case when both the source and target table join key columns are null and we want to update these records.
Instead of this logic, I propose to use flags to indicate whether a source and a target table record exist in the joined record:
```
FROM SELECT (SELECT *, true AS targetRecordExists FROM target) t RIGHT OUTER JOIN source src ON t.a <=> src.a
-- insert new records
INSERT INTO target
   SELECT ... WHERE targetRecordExists IS NULL
-- update records matching join criteria
INSERT INTO target
   SELECT ... WHERE targetRecordExists
```

See jira https://issues.apache.org/jira/browse/HIVE-21481

Yes, 
* execution plan of merge statements.
* result set of queries executed after merge statements merging into tables having null values on merge join columns

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=sqlmerge_nullsafe.q -pl itests/qtest -Pitests
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestIcebergCliDriver -Dqfile=merge_iceberg_copy_on_write_partitioned.q,merge_iceberg_copy_on_write_partitioned_nullsafe.q,merge_iceberg_copy_on_write_unpartitioned.q,merge_iceberg_copy_on_write_unpartitioned_nullsafe.q,merge_iceberg_orc.q,merge_iceberg_orc_nullsafe.q,merge_iceberg_partitioned_orc.q,merge_iceberg_partitioned_orc_nullsafe.q -pl itests/qtest-iceberg -Pitests,iceberg
```

### DIFF
--- a/iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_copy_on_write_partitioned.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_copy_on_write_partitioned.q
@@ -7,8 +7,8 @@ drop table if exists source;
 create external table target_ice(a int, b string, c int) partitioned by spec (bucket(16, a), truncate(3, b)) stored by iceberg tblproperties ('format-version'='2', 'write.merge.mode'='copy-on-write');
 create table source(a int, b string, c int);
 
-insert into target_ice values (1, 'one', 50), (2, 'two', 51), (111, 'one', 55), (333, 'two', 56);
-insert into source values (1, 'one', 50), (2, 'two', 51), (3, 'three', 52), (4, 'four', 53), (5, 'five', 54), (111, 'one', 55);
+insert into target_ice values (1, 'match', 50), (2, 'not match', 51), (3, 'delete', 55), (4, 'not delete null', null), (null, 'match null', 56);
+insert into source values (1, 'match', 50), (22, 'not match', 51), (3, 'delete', 55), (4, 'not delete null', null), (null, 'match null', 56);
 
 -- merge
 explain
--- a/iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_copy_on_write_partitioned.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_copy_on_write_partitioned.q
@@ -13,29 +13,29 @@ insert into source values (1, 'match', 50), (22, 'not match', 51), (3, 'delete',
 -- merge
 explain
 merge into target_ice as t using source src ON t.a = src.a
-when matched and t.a > 100 THEN DELETE
-when matched then update set b = 'Merged', c = t.c + 10;
-
-merge into target_ice as t using source src ON t.a = src.a
-when matched and t.a > 100 THEN DELETE
-when matched then update set b = 'Merged', c = t.c + 10;
-
-select * from target_ice;
-
-explain
-merge into target_ice as t using source src ON t.a = src.a
-when not matched then insert values (src.a, src.b, src.c);
-
-merge into target_ice as t using source src ON t.a = src.a
-when not matched and src.a <= 5 then insert values (src.a, src.b, src.c);
-
-select * from target_ice;
+when matched and t.c > 50 THEN DELETE
+when matched then update set b = concat(t.b, ' Merged'), c = t.c + 10
+when not matched then insert values (src.a, concat(src.b, ' New'), src.c);
 
 -- insert clause with a column list
+explain
 merge into target_ice as t using source src ON t.a = src.a
-when not matched then insert (a, c) values (src.a, src.c);
+when matched and t.c > 50 THEN DELETE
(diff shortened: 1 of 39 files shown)


iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_copy_on_write_partitioned.q
```diff
@@ -7,8 +7,8 @@ drop table if exists source;
 create external table target_ice(a int, b string, c int) partitioned by spec (bucket(16, a), truncate(3, b)) stored by iceberg tblproperties ('format-version'='2', 'write.merge.mode'='copy-on-write');
 create table source(a int, b string, c int);
 
-insert into target_ice values (1, 'one', 50), (2, 'two', 51), (111, 'one', 55), (333, 'two', 56);
-insert into source values (1, 'one', 50), (2, 'two', 51), (3, 'three', 52), (4, 'four', 53), (5, 'five', 54), (111, 'one', 55);
+insert into target_ice values (1, 'match', 50), (2, 'not match', 51), (3, 'delete', 55), (4, 'not delete null', null), (null, 'match null', 56);
```
deniskuzZ, line 10: `when matched and t.a > 100 THEN DELETE` is never triggered after the change

iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_copy_on_write_unpartitioned.q
```diff
@@ -7,8 +7,8 @@ drop table if exists source;
 create external table target_ice(a int, b string, c int) stored by iceberg tblproperties ('format-version'='2', 'write.merge.mode'='copy-on-write');
 create table source(a int, b string, c int);
 
-insert into target_ice values (1, 'one', 50), (2, 'two', 51), (111, 'one', 55), (333, 'two', 56);
-insert into source values (1, 'one', 50), (2, 'two', 51), (3, 'three', 52), (4, 'four', 53), (5, 'five', 54), (111, 'one', 55);
+insert into target_ice values (1, 'match', 50), (2, 'not match', 51), (3, 'delete', 55), (4, 'not delete null', null), (null, 'match null', 56);
```
deniskuzZ, line 10: same thing here

iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_orc.q
```diff
@@ -4,8 +4,8 @@ set hive.explain.user=false;
 create external table target_ice(a int, b string, c int) stored by iceberg stored as orc tblproperties ('format-version'='2');
 create table source(a int, b string, c int);
 
-insert into target_ice values (1, 'one', 50), (2, 'two', 51), (111, 'one', 55), (333, 'two', 56);
-insert into source values (1, 'one', 50), (2, 'two', 51), (3, 'three', 52), (4, 'four', 53), (5, 'five', 54), (111, 'one', 55);
+insert into target_ice values (1, 'match', 50), (2, 'not match', 51), (3, 'delete', 55), (4, 'not delete null', null), (null, 'match null', 56);
```
deniskuzZ, line 7: here

iceberg/iceberg-handler/src/test/queries/positive/merge_iceberg_partitioned_orc.q
```diff
@@ -4,8 +4,8 @@ set hive.explain.user=false;
 create external table target_ice(a int, b string, c int) partitioned by spec (bucket(16, a), truncate(3, b)) stored by iceberg stored as orc tblproperties ('format-version'='2');
 create table source(a int, b string, c int);
 
-insert into target_ice values (1, 'one', 50), (2, 'two', 51), (111, 'one', 55), (333, 'two', 56);
-insert into source values (1, 'one', 50), (2, 'two', 51), (3, 'three', 52), (4, 'four', 53), (5, 'five', 54), (111, 'one', 55);
+insert into target_ice values (1, 'match', 50), (2, 'not match', 51), (3, 'delete', 55), (4, 'not delete null', null), (null, 'match null', 56);
```
deniskuzZ, line 7: and here

PR #5470, opened 2024-09-27
HIVE-27262: send notifications on add / drop / grant / revoke role and grant / revoke privileges

bagipriyank:

[HIVE-27262](https://issues.apache.org/jira/browse/HIVE-27262)

Sending an event to transactional listeners on adding / removing / granting / revoking a role or granting / revoking privileges. No notification events were being generated for these scenarios before this change.

No

No

On local setup. Also added tests to itests.


hcatalog/server-extensions/pom.xml
```diff
@@ -40,6 +40,11 @@
   <dependencies>
     <!-- dependencies are always listed in sorted order by groupId, artifectId -->
     <!-- intra-project -->
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-standalone-metastore</artifactId>
```
bharos, line 45: Do we need the standalone metastore dependendy in hcatalog extensions

bagipriyank:
@nrg4878 can you please review?

nrg4878:
@bagipriyank Thank you for the patch. While I agree that this is a gap in the product, I think the need for this is somewhat negated with the use of an external authorization manager like Ranger. I do not know what % of users use Hive's default authorizer.
Having said that can you provide a bit more context on what some of the other the usecases might be ? One of my concerns would be increased generation of notification_ids which has been somewhat of a bottleneck under heavy loads. What sort of testing have you done with high loads. Thank you

PR #5471, opened 2024-09-27
HIVE-28541: Incorrectly treating materialized CTE as Table when privi…

shuaiqig:

https://issues.apache.org/jira/browse/HIVE-28541

 - check `isMaterializedTable()` before adding elements into `Collection<ReadEntity>`
 - add materialized CTE's real `ReadEntity` into parent's `inputs`

it is a bug, which caused an exception when I used `hive.optimize.cte.materialize.threshold`

No

No

It is hard to test with simple units. I tested and deployed it in my production environment.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SimpleFetchOptimizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SimpleFetchOptimizer.java
@@ -425,7 +425,9 @@ public class SimpleFetchOptimizer extends Transform {
       Utilities.addSchemaEvolutionToTableScanOperator(table, scanOp);
       TableDesc tableDesc = Utilities.getTableDesc(table);
       if (!table.isPartitioned()) {
-        inputs.add(new ReadEntity(table, parent, !table.isView() && parent == null));
+        if (!table.isMaterializedTable()) {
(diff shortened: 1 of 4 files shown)


okumin:
The problem statement is obvious. I'm reviewing and finding an excellent way to test it...

okumin:
@shuaiqig I guess we can test it using the integration test like this. Could you add some test cases? I guess we want 4 test cases as it is a security related feature.
- Positive test with CBO. If the user has access to the source table, they can read the source table and materialized CTEs
- Positive test without CBO
- Negative test with CBO. If the user doesn't have access to the source table, they fail to run the query
- Negative test without CBO

```
--! qt:dataset:src
set hive.security.authorization.manager=org.apache.hadoop.hive.ql.security.authorization.DefaultHiveAuthorizationProvider;

create table src_auth_test_cte_materialization as select * from src;

set hive.security.authorization.enabled=true;

grant select on table src_auth_test_cte_materialization to user hive_test_user;

set hive.optimize.cte.materialize.threshold=1;

explain
with x as (select count(*) from src_auth_test_cte_materialization) select * from x;

with x as (select count(*) from src_auth_test_cte_materialization) select * from x;
```

You can see some guidance here. Feedback is always welcome
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=27362107#HowToContribute-QueryUnitTest

PR #5473, opened 2024-09-27
HIVE-27262: send notifications on add / drop / grant / revoke role and grant / revoke privileges

bagipriyank:

[HIVE-27262](https://issues.apache.org/jira/browse/HIVE-27262)

Sending an event to transactional listeners on adding / removing / granting / revoking a role or granting / revoking privileges. No notification events were being generated for these scenarios before this change.

No

No

On local setup by running itests. Added tests to itests.


bagipriyank:
@nrg4878 can you please review?

PR #5474, opened 2024-09-28
HIVE-28460: Determine the database type once the PersistenceManagerFa…

dengzhhu653:
…ctory created

Determine the Database product only once

Every time ObjectStore initializes, it needs to determine the database type by asking the connection the product name. It's a waste of time and resource to do this, as long as we don't change the database connection information, the db type shouldn't be changed, sometimes retrieving the connection for the secondary connection pool(which size is 2) could be time cost, which can lead to the client taking much time to connect to the Metastore.
This PR also fixes the potential connection leaks on retrying to create the PersistenceManagerFactory.

No

No

Existing tests

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
@@ -18,6 +18,7 @@
 
 package org.apache.hadoop.hive.metastore;
 
+import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.SQLIntegrityConstraintViolationException;
 import java.sql.SQLTransactionRollbackException;
@@ -43,6 +44,8 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+
+import javax.sql.DataSource;
 
 /** Database product inferred via JDBC. Encapsulates all SQL logic associated with
  * the database product.
@@ -70,6 +73,8 @@ public class DatabaseProduct implements Configurable {
   private static DatabaseProduct theDatabaseProduct;
 
   Configuration myConf;
+
+  private String productName;
   /**
    * Protected constructor for singleton class
    */
@@ -82,6 +87,16 @@ public class DatabaseProduct implements Configurable {
   public static final String POSTGRESQL_NAME = "postgresql";
   public static final String ORACLE_NAME = "oracle";
   public static final String UNDEFINED_NAME = "other";
+
+  public static DatabaseProduct determineDatabaseProduct(DataSource connPool,
+      Configuration conf) {
+    try (Connection conn = connPool.getConnection()) {
+      String s = conn.getMetaData().getDatabaseProductName();
+      return determineDatabaseProduct(s, conf);
+    } catch (SQLException e) {
+      throw new RuntimeException("Unable to get database product name", e);
+    }
+  }
 
   /**
    * Determine the database product type
@@ -144,6 +159,7 @@ public class DatabaseProduct implements Configurable {
         }
 
         theDatabaseProduct.dbType = dbt;
+        theDatabaseProduct.productName = productName;
       }
     }
     return theDatabaseProduct;
@@ -817,4 +833,7 @@ public class DatabaseProduct implements Configurable {
     return dbType == DbType.DERBY || dbType == DbType.ORACLE;
   }
 
+  public String getProductName() {
+    return productName;
+  }
 }
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
@@ -94,7 +94,7 @@ public class DatabaseProduct implements Configurable {
       String s = conn.getMetaData().getDatabaseProductName();
       return determineDatabaseProduct(s, conf);
     } catch (SQLException e) {
-      throw new RuntimeException("Unable to get database product name", e);
+      throw new IllegalStateException("Unable to get database product name", e);
     }
   }
 
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
@@ -306,8 +306,9 @@ public class PersistenceManagerProvider {
     if (dsp == null) {
       pmf = JDOHelper.getPersistenceManagerFactory(dsProp);
     } else {
+      DataSource ds = null;
+      DataSource ds2 = null;
       String sourceName = forCompactor ? "objectstore-compactor" : "objectstore";
-      DataSource ds = null, ds2 = null;
       try (DataSourceProvider.DataSourceNameConfigurator configurator =
                new DataSourceProvider.DataSourceNameConfigurator(conf, sourceName)) {
         ds = (maxPoolSize > 0) ? dsp.create(conf, maxPoolSize) : dsp.create(conf);
@@ -336,7 +337,7 @@ public class PersistenceManagerProvider {
         }
       }
     }
-    DataStoreCache dsc = pmf.getDataStoreCache();
+    DataStoreCache dsc = pmf != null ? pmf.getDataStoreCache() : null;
     if (dsc != null) {
       String objTypes = MetastoreConf.getVar(conf, ConfVars.CACHE_PINOBJTYPES);
       LOG.info(
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
@@ -311,6 +311,7 @@ public class PersistenceManagerProvider {
       try (DataSourceProvider.DataSourceNameConfigurator configurator =
                new DataSourceProvider.DataSourceNameConfigurator(conf, sourceName)) {
         ds = (maxPoolSize > 0) ? dsp.create(conf, maxPoolSize) : dsp.create(conf);
+        databaseProduct = DatabaseProduct.determineDatabaseProduct(ds, conf);
         // The secondary connection factory is used for schema generation, and for value generation operations.
         // We should use a different pool for the secondary connection factory to avoid resource starvation.
         // Since DataNucleus uses locks for schema generation and value generation, 2 connections should be sufficient.
@@ -318,7 +319,6 @@ public class PersistenceManagerProvider {
         ds2 = dsp.create(conf, /* maxPoolSize */ 2);
         dsProp.put(PropertyNames.PROPERTY_CONNECTION_FACTORY, ds);
         dsProp.put(PropertyNames.PROPERTY_CONNECTION_FACTORY2, ds2);
-        databaseProduct = DatabaseProduct.determineDatabaseProduct(ds, conf);
         dsProp.put(ConfVars.MANAGER_FACTORY_CLASS.getVarname(),
             "org.datanucleus.api.jdo.JDOPersistenceManagerFactory");
         pmf = JDOHelper.getPersistenceManagerFactory(dsProp);
@@ -516,8 +516,8 @@ public class PersistenceManagerProvider {
 
   public static DatabaseProduct getDatabaseProduct() {
     if (databaseProduct == null) {
-      throw new RuntimeException(
-          "Cannot determine the database product. PersistenceManagerFactory has not initialized yet");
+      throw new IllegalStateException(
+          "Cannot determine the database product. PersistenceManagerFactory has not been initialized yet");
     }
     return databaseProduct;
   }
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
@@ -83,6 +83,7 @@ import static org.apache.hadoop.hive.metastore.conf.MetastoreConf.ConfVars.COMPA
 public class PersistenceManagerProvider {
   private static PersistenceManagerFactory pmf;
   private static PersistenceManagerFactory compactorPmf;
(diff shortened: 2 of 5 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
```diff
@@ -83,6 +88,16 @@ protected DatabaseProduct() {}
   public static final String ORACLE_NAME = "oracle";
   public static final String UNDEFINED_NAME = "other";
 
+  public static DatabaseProduct determineDatabaseProduct(DataSource connPool,
+      Configuration conf) {
+    try (Connection conn = connPool.getConnection()) {
+      String s = conn.getMetaData().getDatabaseProductName();
+      return determineDatabaseProduct(s, conf);
+    } catch (SQLException e) {
+      throw new RuntimeException("Unable to get database product name", e);
```
difin, line 97: Is it possible to use some related Metastore runtime exception, instead of generic RuntimeException?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
```diff
@@ -499,6 +514,14 @@ public static PersistenceManager getPersistenceManager(boolean forCompactor) {
     }
   }
 
+  public static DatabaseProduct getDatabaseProduct() {
+    if (databaseProduct == null) {
+      throw new RuntimeException(
```
difin, line 519: Same here, is it possible to use some specialized Metastore runtime exception?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
dengzhhu653: if throwing other exceptions, we need to change our existing methods, such as the `ObjectStore.setConf` doesn't specify the allowed exceptions in `Configurable.setConf`.
Somewhere we catch and throw the RuntimeException upon the initialization. For example:
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L412
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java#L299

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
```diff
@@ -292,11 +292,10 @@ public void setConf(Configuration conf) {
             }
           }
           if (dbProduct == null) {
-            try (Connection dbConn = getDbConn(Connection.TRANSACTION_READ_COMMITTED, connPool)) {
-              determineDatabaseProduct(dbConn);
-            } catch (SQLException e) {
-              LOG.error("Unable to determine database product", e);
-              throw new RuntimeException(e);
+            dbProduct = DatabaseProduct.determineDatabaseProduct(connPool, conf);
+            if (dbProduct.isUNDEFINED()) {
```
deniskuzZ, line 296: that would never happen, `DatabaseProduct.determineDatabaseProduct` throws a runtime exception

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
deniskuzZ: maybe change to `IllegalStateException`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
```diff
@@ -302,23 +307,33 @@ private static PersistenceManagerFactory initPMF(Configuration conf, boolean for
...
+        ds = (maxPoolSize > 0) ? dsp.create(conf, maxPoolSize) : dsp.create(conf);
         // The secondary connection factory is used for schema generation, and for value generation operations.
         // We should use a different pool for the secondary connection factory to avoid resource starvation.
         // Since DataNucleus uses locks for schema generation and value generation, 2 connections should be sufficient.
         configurator.resetName("objectstore-secondary");
-        DataSource ds2 = forCompactor ? ds : dsp.create(conf, /* maxPoolSize */ 2);
+        ds2 = forCompactor ? ds : dsp.create(conf, /* maxPoolSize */ 2);
         dsProp.put(PropertyNames.PROPERTY_CONNECTION_FACTORY, ds);
         dsProp.put(PropertyNames.PROPERTY_CONNECTION_FACTORY2, ds2);
+        databaseProduct = DatabaseProduct.determineDatabaseProduct(ds, conf);
```
deniskuzZ, line 321: can we do this after the `ds` initialization?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
```diff
@@ -302,23 +307,33 @@ private static PersistenceManagerFactory initPMF(Configuration conf, boolean for
...
         dsProp.put(ConfVars.MANAGER_FACTORY_CLASS.getVarname(),
             "org.datanucleus.api.jdo.JDOPersistenceManagerFactory");
         pmf = JDOHelper.getPersistenceManagerFactory(dsProp);
       } catch (SQLException e) {
         LOG.warn("Could not create PersistenceManagerFactory using "
             + "connection pool properties, will fall back", e);
         pmf = JDOHelper.getPersistenceManagerFactory(prop);
+      } finally {
+        if (pmf == null && ds instanceof AutoCloseable) {
+          try (AutoCloseable close1 = (AutoCloseable) ds;
```
deniskuzZ, line 331: why is this magic, can't we call close directly?

dengzhhu653: The `DataSource` doesn't have the `close` directly to free the resource. For `HikariDataSource` and `PoolingDataSource`, they are implemented both the `AutoCloseable` and `DataSource`, so here need to cast the `ds` to the `AutoCloseable` explicitly.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
dengzhhu653: sure

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
dengzhhu653: It could, as this is default in `getDbType(String productName)`: https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java#L167

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
dengzhhu653: sure

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
deniskuzZ: can't we refactor to use try-with-resources?
````
try (DataSource ds = create.., ds2 = create..) {
....
}
````

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java:331]: can we add a `close` method to `DataSourceProvider`
````
  static void close(DataSource ds) {
    if (ds instanceof AutoCloseable) {
      try {
        ((AutoCloseable)ds).close();
      } catch (Exception e) {
        LOG.warn("Failed to close the DataSource", e);
      }
      }
    }
  }
````

Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java:331]: I would like to keep it, we don't introduce anything other than `close` in this method, and we have no chance to reuse it elsewhere.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java:331]: c3p0 provides the same method for handling pooled DS. why use try-with-resources if the block body is empty??

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java (hunk 1 above)
Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java:97]: can we throw `IllegalStateException`?

**Comment by deniskuzZ** [comment]
please check the sonar issues as well

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java (hunk 5 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java:332]: The `close` method in `AutoCloseable` declared throwing the generic exception, on the caller, we should catch it anyhow. The issues mentioned in the link is because the `pmf` might be null, but this should never happen, as the exception should be thrown during create the `pmf`, the code cannot reach here in this case, but I also fixed it in the new commit.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java:332]: Nested blocks of code should not be left empty [java:S108](https://sonarcloud.io/organizations/apache/rules?open=java%3AS108&rule_key=java%3AS108)

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java [hunk 6]
```diff
@@ -319,9 +327,17 @@ private static PersistenceManagerFactory initPMF(Configuration conf, boolean for
...
+               AutoCloseable close2 = (AutoCloseable) ds2 ) {
+          } catch (Exception e) {
+            LOG.warn("Failed to close the DataSource", e);
+          }
+        }
       }
     }
-    DataStoreCache dsc = pmf.getDataStoreCache();
+    DataStoreCache dsc = pmf != null ? pmf.getDataStoreCache() : null;
```
deniskuzZ, line 340: warn would be misleading with that change when pmf is null
````
LOG.warn("PersistenceManagerFactory returned null DataStoreCache object
````

Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java:340]: yes, when reaching here the `pmf` shouldn't be null, the `finally` block will be executed whether there is a JDOFatalUserException or not, if in finally the `pmf` is null, there should be the exception thrown, adding the `return null` makes the impression this method can returning the nullable `pmf`, it makes little sense from my perspective.

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java (hunk 5 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java:332]: Added a debug log in this block

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java (hunk 6 above)
Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java]: agree, but I don't see that you throw an exception in finally when `pmf` is null

Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java]: I write a simple code, 
```java
 public static void main(String[] args) throws Exception {
    String a = null;
    try {
     a();
    } catch (SQLException e) {
      throw new JDOFatalUserException();
    } finally {
      if (a == null) {
        System.out.println("print a");
      }
    }
    System.out.println("print b");
  }
  private static void a() throws Exception {
    throw new JDOFatalUserException();
  }
```
the output:
```
print a
Exception in thread "main" javax.jdo.JDOFatalUserException
	at org.example.FinallyException.a(FinallyException.java:23)
	at org.example.FinallyException.main(FinallyException.java:11)
```
as see, the finally is executed before the exception is thrown.

deniskuzZ: @dengzhhu653  that's because you catch SQLException, but throw JDOException
````
  public  static void main(String ... args) {
    String a = null;
    try {
      a();
    } catch (Exception e) {
      System.out.println("catch block");
      throw new JDOFatalUserException("outer");
    } finally {
      System.out.println("finally");
      if (a == null) {
        System.out.println("print a");
      }
    }
    System.out.println("print b");
  }
  private static void a() throws Exception {
    throw new JDOFatalUserException("inner");
  } 
````
catch block
finally
print a
Exception in thread "main" javax.jdo.JDOFatalUserException: outer
	at org.apache.Main.main(Main.java:17)

deniskuzZ: i think i get you point. `pmf` would be only null if  JDOFatalUserException is thrown, but we should still close the datasources

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
deniskuzZ: @dengzhhu653, can we make this change? it's still a RuntimeException.

deniskuzZ (commented):
LGTM, some minor stuff

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DatabaseProduct.java
dengzhhu653: sorry, somehow I missed this comment, fixed.

deniskuzZ (approved):
+1

PR #5475, opened 2024-09-28
HIVE-28543: Previous snapshotId is stored in backend database for iceberg tables

Aggarwal-Raghav:
[HIVE-28543](https://issues.apache.org/jira/browse/HIVE-28543)

In iceberg table, post inserting the data into iceberg table, the _current-snapshot-id_ in _desc formatted_ output is different from the value stored in backend db (for examlpe: mysql)

NO

NO

On local setup

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -89,7 +89,6 @@ import org.apache.iceberg.PartitionSpecParser;
 import org.apache.iceberg.PartitionsTable;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SchemaParser;
-import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.TableMetadataParser;
@@ -1062,10 +1061,6 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
     if (hmsTable != null) {
       try {
         Table tbl = IcebergTableUtil.getTable(conf, hmsTable);
-        Snapshot snapshot = tbl.currentSnapshot();
-        if (snapshot != null) {
-          hmsTable.getParameters().put("current-snapshot-id", String.valueOf(snapshot.snapshotId()));
-        }
         String formatVersion = String.valueOf(((BaseTable) tbl).operations().current().formatVersion());
         hmsTable.getParameters().put(TableProperties.FORMAT_VERSION, formatVersion);
         // Set the serde info


Aggarwal-Raghav:
**Explanation:** 
When we run insert query on iceberg table, in _StatTask_ _getTable_ call is made https://github.com/apache/hive/blob/d85b87cfd750623d365d39c73df6d58e1220128a/ql/src/java/org/apache/hadoop/hive/ql/exec/StatsTask.java#L103 this _getTable_ contains the stale snapshotID, reason being in _HiveIcebergMetaHook.java_  the getTable call  to IcebergUtil uses the skipcache=false and  it returns the previous snapshotID, causing the problem
 https://github.com/apache/hive/blob/d85b87cfd750623d365d39c73df6d58e1220128a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java#L1067

_For iceberg tables, DML and DDL operations are not disjoint sets anymore, i.e. insert queries makes changes to metadata hence we cannot use cache_ 

Attaching the screenshot for before and after the fix state:
**Before:**
Before_fix_beeline
Before_fix_mysql

**Even though the latest snapshotID is 4050123687297981987 but in backend db 8504103028255089587 is stored.**

**After:**
After_fix_beeline
After_fix_mysql

Aggarwal-Raghav:
Issue will be observed post the second insert query.

Aggarwal-Raghav:
@deniskuzZ @zhangbutao , can you please review this?

zhangbutao (approved):
LGTM +1
Good catch!

deniskuzZ:
sorry, maybe I am missing something, but how doesn't it cause any data loss, since writers might be updating the prev snapshot and not the latest?

Aggarwal-Raghav:
@deniskuzZ , thanks for the review. Your both questions are genuine and valid.

The reason why we are not seeing the data is loss despite stale snapshotId present in backend is because it is getting the value of current-snapshot-id from the latest metadata.json file which contains the correct snapshotId. Meaning, the first  getTable call during insert will give previous snapshotId but it will get overwritten during HiveTableOperation.java
https://github.com/apache/hive/blob/4472bba342efd62ecb90e21ad9838c62dbb1776a/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L165

Screenshot 2024-09-30 at 7 16 30 PM


**To answer your second question**, why only snapshotId is stale and not the metadata location, in HiveIcebergMetaHook#postGetTable, the hmsTable (present in method argument **contains updated value**) and tbl **contains stale value.**
https://github.com/apache/hive/blob/4472bba342efd62ecb90e21ad9838c62dbb1776a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java#L1064

_Attaching 1 instance of debugging, here check the metadata location._ 

Screenshot 2024-09-30 at 7 33 11 PM

**As we are only updating current-snapshot-id in the method, hence snapshotID is stale and rest all info is correct.**
https://github.com/apache/hive/blob/4472bba342efd62ecb90e21ad9838c62dbb1776a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java#L1067

deniskuzZ:
> @deniskuzZ , thanks for the review. Your both questions are genuine and valid.
> 
> The reason why we are not seeing the data is loss despite stale snapshotId present in backend is because it is getting the value of current-snapshot-id from the latest metadata.json file which contains the correct snapshotId. Meaning, the first getTable call during insert will give previous snapshotId but it will get overwritten during HiveTableOperation.java
> 
> https://github.com/apache/hive/blob/4472bba342efd62ecb90e21ad9838c62dbb1776a/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L165
> 
> Screenshot 2024-09-30 at 7 16 30 PM
> **To answer your second question**, why only snapshotId is stale and not the metadata location, in HiveIcebergMetaHook#postGetTable, the hmsTable (present in method argument **contains updated value**) and tbl **contains stale value.**
> 
> https://github.com/apache/hive/blob/4472bba342efd62ecb90e21ad9838c62dbb1776a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java#L1064
> 
> _Attaching 1 instance of debugging, here check the metadata location._
> 
> Screenshot 2024-09-30 at 7 33 11 PM
> **As we are only updating current-snapshot-id in the method, hence snapshotID is stale and rest all info is correct.**
> 
> https://github.com/apache/hive/blob/4472bba342efd62ecb90e21ad9838c62dbb1776a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java#L1067

@Aggarwal-Raghav, thank you for the detailed explanation! Instead of refreshing the cache, could we try to use `current-snapshot-id` from `hmsTable` if present?

deniskuzZ:
based on a code location is taken from hmsTable as well
````
    if (properties.get(Catalogs.LOCATION) == null &&
        hmsTable.getSd() != null && hmsTable.getSd().getLocation() != null) {
      properties.put(Catalogs.LOCATION, hmsTable.getSd().getLocation());
    }
````
I think we don't even need iceberg Table instance in `postGetTable`..

**Comment by ayushtkn** [comment]
Rajesh had some use case I believe, I don’t recall exactly but we shouldn’t skip cache every time, I will try to find the conversation in sometime

we can maybe dodge this by putting the current snapshot id only for show create table or describe table commands, for rest of them maybe we need not to put it in

**Comment by Aggarwal-Raghav** [comment]
@ayushtkn, I tried reverting the java changes done as part of [HIVE-27049](https://issues.apache.org/jira/browse/HIVE-27049) and checked in both hive4 and spark 3.4.1 + iceberg_spark_runtime.jar (1.4.3). _current-snapshot-id_ still is present in output of _show create table / desc formatted table_ . **If possible can you please provide the use case of** [HIVE-27049](https://issues.apache.org/jira/browse/HIVE-27049)?

**Just INFO:**
`Table tbl = IcebergTableUtil.getTable(conf, hmsTable);`
is doing a hdfs read on metadata.json file and from that we are obtaining the _current-snaptshot-id_ and _format-version_ in _postGetTable_. Based on the comments in this iceberg issue, https://github.com/apache/iceberg/issues/6194#issuecomment-1330841257, Spark and Impala are also obtaining _format-version_ from metadata.json file and hive is doing the same. That mean 1 HDFS read call for every iceberg getTable call.

**What's next:**
Based on the sentiment of the review comments, looks like skipCache=true is not the way forward. So, there are 2 options:
- [option 1] Go ahead with @deniskuzZ  suggestion
```
 hmsTable.getParameters().computeIfAbsent(hive_metastoreConstants.META_TABLE_SNAPSHOT_ID, k ->
    snapshot != null? String.valueOf(snapshot.snapshotId()) : null);
```
- [option 2] Revert the java changes done in [HIVE-27049](https://issues.apache.org/jira/browse/HIVE-27049).

ayushtkn:
The first option can solve problem in few cases when HMS has it, IIRC correctly in few cases it ain’t there, I am more inclined on setting this only during show table or describe table calls rather than all & that should be the ideal fix IMO

deniskuzZ:
> The first option can solve problem in few cases when HMS has it, IIRC correctly in few cases it ain’t there, I am more inclined on setting this only during show table or describe table calls rather than all & that should be the ideal fix IMO

if we can't repro the issue with `show table` or `describe table` ops after the revert of [HIVE-27049](https://issues.apache.org/jira/browse/HIVE-27049) , my vote is for option 2

Aggarwal-Raghav:
@ayushtkn, without the java changes [HIVE-27049](https://issues.apache.org/jira/browse/HIVE-27049) and rebuilding the project, the q files modified in that JIRA are still passing:
Screenshot 2024-10-08 at 8 42 58 PM

Ran the following command from _itests/qtest-iceberg_ dir:
`mvn test -Dtest=TestIcebergCliDriver -Dqfile=show_create_iceberg_table.q,mv_iceberg_partitioned_orc.q,alter_table_to_iceberg.q`

Screenshot 2024-10-08 at 8 51 09 PM

That is why I suggested the revert.

ayushtkn:
Not sure how it wasn’t working before, else the q.out should have that before the change.

so in that case remove the prod change

PR #5477, opened 2024-09-30
HIVE-28545: Remove global lock on HMSHandler.getMSForConf which is pr…

dengzhhu653:
…one to deadlock

Remove global lock on HMSHandler.getMSForConf

This global lock is prone to deadlock and performance penalty.

No

No

Run TestCompactionMetrics.testInitiatorFailuresCountedCorrectly and TestMetaStoreDeadlock

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
@@ -308,8 +308,8 @@ public class PersistenceManagerProvider {
         // The secondary connection factory is used for schema generation, and for value generation operations.
         // We should use a different pool for the secondary connection factory to avoid resource starvation.
         // Since DataNucleus uses locks for schema generation and value generation, 2 connections should be sufficient.
-        configurator.resetName("objectstore-secondary");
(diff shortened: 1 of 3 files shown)


deniskuzZ (approved):
LGTM +1

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
```diff
@@ -308,8 +308,8 @@ private static PersistenceManagerFactory initPMF(Configuration conf, boolean for
         // The secondary connection factory is used for schema generation, and for value generation operations.
         // We should use a different pool for the secondary connection factory to avoid resource starvation.
         // Since DataNucleus uses locks for schema generation and value generation, 2 connections should be sufficient.
-        configurator.resetName("objectstore-secondary");
-        DataSource ds2 = forCompactor ? ds : dsp.create(conf, /* maxPoolSize */ 2);
+        configurator.resetName(forCompactor ? "objectstore-compactor-secondary" : "objectstore-secondary");
```
deniskuzZ, line 311: how about `sourceName+"-secondary"`

dengzhhu653: done, thank you for the suggestion!

PR #5478, opened 2024-10-01
HIVE-28548: Subdivide memory size allocated to parallel operators

okumin:

Let each operator know how much it can use at maximum.

We observed OOM happens when SharedWorkOptimizer merges heavy operators such as GroupByOperators with a map-side hash aggregation. I guess it is hard for GroupByOperator to control its memory correctly in that case.

I confirmed Tez has a similar feature to reallocate memory when a task is connected to multiple edges.
https://github.com/apache/tez/blob/rel/release-0.10.4/tez-runtime-library/src/main/java/org/apache/tez/runtime/library/resources/WeightedScalingMemoryDistributor.java

https://issues.apache.org/jira/browse/HIVE-28548

I also found it is still problematic when # of merged operators is huge, e.g. 100, because the assignment per operator gets tiny. I will handle such case in HIVE-28549.

No.

No.

Checked local logs.

```
--! qt:dataset:src

set hive.auto.convert.join=true;

SELECT *
FROM (SELECT key, count(*) AS num FROM src WHERE key LIKE '%1%' GROUP BY key) t1
LEFT OUTER JOIN (SELECT key, count(*) AS num FROM src WHERE key LIKE '%2%' GROUP BY key) t2 ON t1.key = t2.key
LEFT OUTER JOIN (SELECT key, count(*) AS num FROM src WHERE key LIKE '%3%' GROUP BY key) t3 ON t1.key = t3.key
LEFT OUTER JOIN (SELECT key, count(*) AS num FROM src WHERE key LIKE '%4%' GROUP BY key) t4 ON t1.key = t4.key
LEFT OUTER JOIN (SELECT key, count(*) AS num FROM src WHERE key LIKE '%5%' GROUP BY key) t5 ON t1.key = t5.key;

set hive.vectorized.execution.enabled=false;

SELECT *
FROM (SELECT key, count(*) AS num FROM src WHERE key LIKE '%1%' GROUP BY key) t1
LEFT OUTER JOIN (SELECT key, count(*) AS num FROM src WHERE key LIKE '%2%' GROUP BY key) t2 ON t1.key = t2.key
LEFT OUTER JOIN (SELECT key, count(*) AS num FROM src WHERE key LIKE '%3%' GROUP BY key) t3 ON t1.key = t3.key
LEFT OUTER JOIN (SELECT key, count(*) AS num FROM src WHERE key LIKE '%4%' GROUP BY key) t4 ON t1.key = t4.key
LEFT OUTER JOIN (SELECT key, count(*) AS num FROM src WHERE key LIKE '%5%' GROUP BY key) t5 ON t1.key = t5.key;
```

The total memory reduced from 358MB to 71MB.

```
% cat org.apache.hadoop.hive.cli.TestMiniLlapLocalCliDriver-output.txt | grep OperatorUtils | grep Assigning
2024-09-30T23:54:38,486  INFO [TezTR-672468_1_3_0_0_0] exec.OperatorUtils: Assigning 75161926 bytes to 5 operators
2024-09-30T23:54:38,831  INFO [TezTR-672468_1_3_0_0_1] exec.OperatorUtils: Assigning 75161926 bytes to 5 operators
2024-09-30T23:54:42,015  INFO [TezTR-672468_1_4_0_0_0] exec.OperatorUtils: Assigning 75161926 bytes to 5 operators
2024-09-30T23:54:42,310  INFO [TezTR-672468_1_4_0_0_1] exec.OperatorUtils: Assigning 75161926 bytes to 5 operators
```

```
% cat org.apache.hadoop.hive.cli.TestMiniLlapLocalCliDriver-output.txt | grep GroupByOperator | grep 'Max hash table'
2024-09-30T23:54:36,280  INFO [TezTR-672468_1_2_0_0_0] exec.GroupByOperator: Max hash table memory: 179.20MB (358.40MB * 0.5)
2024-09-30T23:54:42,015  INFO [TezTR-672468_1_4_0_0_0] exec.GroupByOperator: Max hash table memory: 35.84MB (71.68MB * 0.5)
2024-09-30T23:54:42,017  INFO [TezTR-672468_1_4_0_0_0] exec.GroupByOperator: Max hash table memory: 35.84MB (71.68MB * 0.5)
2024-09-30T23:54:42,018  INFO [TezTR-672468_1_4_0_0_0] exec.GroupByOperator: Max hash table memory: 35.84MB (71.68MB * 0.5)
2024-09-30T23:54:42,018  INFO [TezTR-672468_1_4_0_0_0] exec.GroupByOperator: Max hash table memory: 35.84MB (71.68MB * 0.5)
...
```

```
% cat org.apache.hadoop.hive.cli.TestMiniLlapLocalCliDriver-output.txt | grep VectorGroupByOperator | grep 'GBY memory limits'
2024-09-30T23:54:38,491  INFO [TezTR-672468_1_3_0_0_0] vector.VectorGroupByOperator: GBY memory limits - isTez: true isLlap: true maxHashTblMemory: 64.51MB (71.68MB * 0.9) fixSize:660 (key:472 agg:144)
2024-09-30T23:54:38,499  INFO [TezTR-672468_1_3_0_0_0] vector.VectorGroupByOperator: GBY memory limits - isTez: true isLlap: true maxHashTblMemory: 64.51MB (71.68MB * 0.9) fixSize:660 (key:472 agg:144)
2024-09-30T23:54:38,500  INFO [TezTR-672468_1_3_0_0_0] vector.VectorGroupByOperator: GBY memory limits - isTez: true isLlap: true maxHashTblMemory: 64.51MB (71.68MB * 0.9) fixSize:660 (key:472 agg:144)
2024-09-30T23:54:38,500  INFO [TezTR-672468_1_3_0_0_0] vector.VectorGroupByOperator: GBY memory limits - isTez: true isLlap: true maxHashTblMemory: 64.51MB (71.68MB * 0.9) fixSize:660 (key:472 agg:144)
...
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java
@@ -35,6 +35,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.common.type.TimestampTZ;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.llap.LlapDaemonInfo;
+import org.apache.hadoop.hive.llap.LlapUtil;
 import org.apache.hadoop.hive.ql.CompilationOpContext;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.parse.OpParseContext;
@@ -389,14 +390,14 @@ public class GroupByOperator extends Operator<GroupByDesc> implements IConfigure
     isLlap = LlapDaemonInfo.INSTANCE.isLlap();
     numExecutors = isLlap ? LlapDaemonInfo.INSTANCE.getNumExecutors() : 1;
     firstRow = true;
+    memoryMXBean = ManagementFactory.getMemoryMXBean();
+    maxMemory = isTez ? getConf().getMaxMemoryAvailable() : memoryMXBean.getHeapMemoryUsage().getMax();
     // estimate the number of hash table entries based on the size of each
     // entry. Since the size of a entry
     // is not known, estimate that based on the number of entries
     if (hashAggr) {
       computeMaxEntriesHashAggr();
     }
-    memoryMXBean = ManagementFactory.getMemoryMXBean();
-    maxMemory = isTez ? getConf().getMaxMemoryAvailable() : memoryMXBean.getHeapMemoryUsage().getMax();
     memoryThreshold = this.getConf().getMemoryThreshold();
     LOG.info("isTez: {} isLlap: {} numExecutors: {} maxMemory: {}", isTez, isLlap, numExecutors, maxMemory);
   }
@@ -411,13 +412,12 @@ public class GroupByOperator extends Operator<GroupByDesc> implements IConfigure
    *         aggregation only
    **/
   private void computeMaxEntriesHashAggr() throws HiveException {
-    float memoryPercentage = this.getConf().getGroupByMemoryUsage();
-    if (isTez) {
-      maxHashTblMemory = (long) (memoryPercentage * getConf().getMaxMemoryAvailable());
-    } else {
-      maxHashTblMemory = (long) (memoryPercentage * Runtime.getRuntime().maxMemory());
+    final float memoryPercentage = this.getConf().getGroupByMemoryUsage();
+    maxHashTblMemory = (long) (memoryPercentage * maxMemory);
+    if (LOG.isInfoEnabled()) {
+      LOG.info("Max hash table memory: {} ({} * {})", LlapUtil.humanReadableByteCount(maxHashTblMemory),
+          LlapUtil.humanReadableByteCount(maxMemory), memoryPercentage);
     }
-    LOG.info("Max hash table memory: {} bytes", maxHashTblMemory);
     estimateRowSize();
   }
 
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/OperatorUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/OperatorUtils.java
@@ -385,17 +385,22 @@ public class OperatorUtils {
   }
 
   public static void setMemoryAvailable(final List<Operator<? extends OperatorDesc>> operators,
-    final long memoryAvailableToTask) {
-    if (operators == null) {
+    final long memoryAvailable) {
+    if (operators == null || operators.isEmpty()) {
       return;
     }
 
+    final long memoryAvailablePerOperator = memoryAvailable / operators.size();
+    Preconditions.checkArgument(memoryAvailablePerOperator > 0);
+    if (operators.size() > 1) {
+      LOG.info("Assigning {} bytes to {} operators", memoryAvailablePerOperator, operators.size());
+    }
     for (Operator<? extends OperatorDesc> op : operators) {
       if (op.getConf() != null) {
-        op.getConf().setMaxMemoryAvailable(memoryAvailableToTask);
+        op.getConf().setMaxMemoryAvailable(memoryAvailablePerOperator);
       }
       if (op.getChildOperators() != null && !op.getChildOperators().isEmpty()) {
-        setMemoryAvailable(op.getChildOperators(), memoryAvailableToTask);
+        setMemoryAvailable(op.getChildOperators(), memoryAvailablePerOperator);
       }
     }
   }
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorGroupByOperator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorGroupByOperator.java
@@ -450,7 +450,7 @@ public class VectorGroupByOperator extends Operator<GroupByDesc>
                 + "minReductionHashAggr:{} ", maxHtEntries, groupingSets.length,
             numRowsCompareHashAggr, minReductionHashAggr);
       }
-      computeMemoryLimits();
+      computeMemoryLimits(HiveConf.getVar(hconf, HiveConf.ConfVars.HIVE_EXECUTION_ENGINE).equals("tez"));
       LOG.debug("using hash aggregation processing mode");
 
       if (keyWrappersBatch.getVectorHashKeyWrappers()[0] instanceof VectorHashKeyWrapperGeneral) {
@@ -616,7 +616,7 @@ public class VectorGroupByOperator extends Operator<GroupByDesc>
     /**
      * Computes the memory limits for hash table flush (spill).
      */
-    private void computeMemoryLimits() {
+    private void computeMemoryLimits(boolean isTez) {
       JavaDataModel model = JavaDataModel.get();
 
       fixedHashEntrySize =
@@ -625,7 +625,7 @@ public class VectorGroupByOperator extends Operator<GroupByDesc>
           aggregationBatchInfo.getAggregatorsFixedSize();
 
       MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
-      maxMemory = isLlap ? getConf().getMaxMemoryAvailable() : memoryMXBean.getHeapMemoryUsage().getMax();
+      maxMemory = isTez ? getConf().getMaxMemoryAvailable() : memoryMXBean.getHeapMemoryUsage().getMax();
       memoryThreshold = conf.getMemoryThreshold();
       // Tests may leave this unitialized, so better set it to 1
       if (memoryThreshold == 0.0f) {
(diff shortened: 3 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorGroupByOperator.java
```diff
@@ -625,7 +625,7 @@ private void computeMemoryLimits() {
           aggregationBatchInfo.getAggregatorsFixedSize();
 
       MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
-      maxMemory = isLlap ? getConf().getMaxMemoryAvailable() : memoryMXBean.getHeapMemoryUsage().getMax();
+      maxMemory = isTez ? getConf().getMaxMemoryAvailable() : memoryMXBean.getHeapMemoryUsage().getMax();
```
okumin, line 628: I guess Tez without LLAP also should use this conf. This is the only case where checking if it is LLAP or not.

ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java
```diff
@@ -411,13 +412,12 @@ protected void initializeOp(Configuration hconf) throws HiveException {
    *         aggregation only
    **/
   private void computeMaxEntriesHashAggr() throws HiveException {
```
deniskuzZ, line 414: can we pass `maxMemory` as a param here to avoid reordering issues after refactor in future?

ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java
```diff
@@ -389,14 +390,14 @@ protected void initializeOp(Configuration hconf) throws HiveException {
     isLlap = LlapDaemonInfo.INSTANCE.isLlap();
     numExecutors = isLlap ? LlapDaemonInfo.INSTANCE.getNumExecutors() : 1;
     firstRow = true;
+    memoryMXBean = ManagementFactory.getMemoryMXBean();
+    maxMemory = isTez ? getConf().getMaxMemoryAvailable() : memoryMXBean.getHeapMemoryUsage().getMax();
```
deniskuzZ, line 394: is that safe to use `memoryMXBean.getHeapMemoryUsage().getMax()`?
````
Returns the maximum amount of memory in bytes that can be used for memory management. This method returns -1 if the maximum memory size is undefined.
````
when `Runtime.getRuntime().maxMemory()`
````
Returns the maximum amount of memory that the Java virtual machine will attempt to use. If there is no inherent limit then the value Long.MAX_VALUE will be returned.
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/OperatorUtils.java [hunk 4]
```diff
@@ -385,17 +385,22 @@ public static int countOperatorsUpstream(Operator<?> start, Set<Class<? extends
...
 
   public static void setMemoryAvailable(final List<Operator<? extends OperatorDesc>> operators,
-    final long memoryAvailableToTask) {
-    if (operators == null) {
+    final long memoryAvailable) {
+    if (operators == null || operators.isEmpty()) {
       return;
     }
 
+    final long memoryAvailablePerOperator = memoryAvailable / operators.size();
```
deniskuzZ, line 393: what if we have uneven distribution? won't it cause even more frequent OOMs?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorGroupByOperator.java
```diff
@@ -450,7 +450,7 @@ public void initialize(Configuration hconf) throws HiveException {
                 + "minReductionHashAggr:{} ", maxHtEntries, groupingSets.length,
             numRowsCompareHashAggr, minReductionHashAggr);
       }
-      computeMemoryLimits();
+      computeMemoryLimits(HiveConf.getVar(hconf, HiveConf.ConfVars.HIVE_EXECUTION_ENGINE).equals("tez"));
```
deniskuzZ, line 453: can we set `isTez` in initializeOp, similar to `isLlap`?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorGroupByOperator.java
deniskuzZ: cc @abstractdog

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorGroupByOperator.java
okumin: I agree but I am wondering if we should take # of executors into account if it is LLAP. I'd like you to find an LLAP expert if possible...
https://github.com/apache/hive/blob/18f34e75da0141d37d9a8f1cef4f7f64ba09fadb/ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java#L875-L877

deniskuzZ: cc @abstractdog

ql/src/java/org/apache/hadoop/hive/ql/exec/GroupByOperator.java
ngsg, line 394: I think we could use a number less than `operators.size()` since not all operators require significant memory that should be managed by `getMaxMemoryAvailable()`. For instance, we can likely ignore select and filter operators, as they behave more like fire-and-forget operations and do not consume much memory. Given that `OperatorDesc.getMaxMemoryAvailable()` is referred by MapJoin, ReduceSink (with TopN), and GroupBy (with Hash Aggregate Mode), I think #MapJoin + #(RS with TopN) + #(GBY with HashMode) + 1(for the rest) could be a candidate.

ql/src/java/org/apache/hadoop/hive/ql/exec/OperatorUtils.java
ngsg, line 393: I think we could use a number less than `operators.size()` since not all operators require significant memory that should be managed by `getMaxMemoryAvailable()`. For instance, we can likely ignore select and filter operators, as they behave more like fire-and-forget operations and do not consume much memory. Given that `OperatorDesc.getMaxMemoryAvailable()` is referenced by MapJoin, ReduceSink (with TopN), and GroupBy (with Hash Aggregate Mode), I think #MapJoin + #(RS with TopN) + #(GBY with HashMode) + 1(for the rest) could be a candidate.

ngsg:
Just for your information, we tested this patch using TPC-DS queries 1-25 on a 10TB dataset. The total execution times were similar, though query 17 became slower (54s → 94s), while query 5 improved (142s → 101s). I haven't investigated the differences in detail, so they could be due to factors like stragglers or other irrelevant issues. Overall, I think the total execution times show that the patch doesn’t introduce any significant performance problems in typical cases, and the patch seems like a reasonable change to me.

ql/src/java/org/apache/hadoop/hive/ql/exec/OperatorUtils.java
okumin: In my impression, my naive approach is likely to work in most cases because SWO is likely to merge similar processing. But it would be nice if we could find good heuristics to calculate the cost here.

okumin:
@ngsg @deniskuzZ @abstractdog Do you think this approach is reasonable? If the answer is yes, I will rebase this branch

ngsg:
I think this is a reasonable approach.

Regarding the `isTez`/`isLlap` issue in `VectorGroupByOperator`, I'm inclined to use `isTez` and drop `isLlap`, as `VectorGroupByOperator` checks overall heap usage via `gcCanary`, not `MemoryMXBean#getHeapMemoryUsage` and `memoryThreshold`. However, I'm not certain whether Tez without LLAP also respects `getConf().getMaxMemoryAvailable()`, so it would be nice if others could confirm this as well.

okumin:
I rebased the branch and CI is green. I'm open to any suggestions

ql/src/java/org/apache/hadoop/hive/ql/exec/OperatorUtils.java
```diff
@@ -385,17 +385,22 @@ public static int countOperatorsUpstream(Operator<?> start, Set<Class<? extends
...
+      LOG.info("Assigning {} bytes to {} operators", memoryAvailablePerOperator, operators.size());
+    }
     for (Operator<? extends OperatorDesc> op : operators) {
       if (op.getConf() != null) {
-        op.getConf().setMaxMemoryAvailable(memoryAvailableToTask);
+        op.getConf().setMaxMemoryAvailable(memoryAvailablePerOperator);
       }
       if (op.getChildOperators() != null && !op.getChildOperators().isEmpty()) {
-        setMemoryAvailable(op.getChildOperators(), memoryAvailableToTask);
+        setMemoryAvailable(op.getChildOperators(), memoryAvailablePerOperator);
```
ngsg, line 403: I have a minor question here. If I understand correctly, this code may subdivide the memory size as shown in the diagram. Is this the intended behavior, given that we only consider operators running in parallel?
```
Grandparent(400M) - Parent1(200M) - Child11(100M)
                                  - Child12(100M)
                  - Parent2(200M) - Child21(100M)
                                  - Child22(100M)
```

okumin: That's likely true. Only operators that buffer records are expected to use `OperatorDesc#getMaxMemoryAvailable`. I suspect parallel operators could concurrently buffer records.

PR #5479, opened 2024-10-01
HIVE-28544: Ensure using UTF-8 encoding in some String/Char/Varchar related operations

abstractdog:
Ensure UTF-8 encoding in some places in the code, that were revealed by customer cases.

Encoding mismatch, leading to stuff like this:

```
create external table test1(a int, b varchar(50), c string) TBLPROPERTIES ("external.table.purge"="true");
insert into test1 values(1,'上海','徐汇'),(2,'北京','海淀');
SELECT * FROM test1 where b='北京';


SELECT * FROM test1;
-- correct, only because it's using a serde where the computation doesn't fall into BytesColumnVector (not the OrcSerde)

+--------------+--------------+--------------+
| test1.a  | test1.b  | test1.c  |
+--------------+--------------+--------------+
| 1            | 上海           | 徐汇           |
| 2            | 北京           | 海淀           |
+--------------+--------------+--------------+


create table test2 as SELECT * FROM test1 where b='北京';

select * from test2;
+--------------+--------------+--------------+
| test2.a  | test2.b  | test2.c  |
+--------------+--------------+--------------+
| 2            | ������       | 海淀           |
+--------------+--------------+--------------+


create table test3 as SELECT * FROM test1;

    select * from test3;

+--------------+--------------+--------------+
| test3.a  | test3.b  | test3.c  |
+--------------+--------------+--------------+
| 1            | ������       | 徐汇           |
| 2            | ������       | 海淀           |
+--------------+--------------+--------------+
```

Yes, for all those customers using special characters on clusters with -Dfile.encoding other than UTF-8, this will fix some problems.

No.

Tested on cluster.

### DIFF
--- a/itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestUtil.java
+++ b/itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestUtil.java
@@ -28,6 +28,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


simhadri-g (approved):
LGTM +1

abstractdog:
thanks a lot @simhadri-g for the review! I just added description and will merge today

PR #5482, opened 2024-10-03
HIVE-28551 Check the transactional table is recreated by its Id

czxm:

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
@@ -44,6 +44,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.apache.hadoop.conf.Configuration;
@@ -88,11 +89,13 @@ public final class QueryResultsCache {
   public static class LookupInfo {
     private String queryText;
     private Supplier<ValidTxnWriteIdList> txnWriteIdListProvider;
+    private Set<Long> txnTables;
 
-    public LookupInfo(String queryText, Supplier<ValidTxnWriteIdList> txnWriteIdListProvider) {
+    public LookupInfo(String queryText, Supplier<ValidTxnWriteIdList> txnWriteIdListProvider, Set<Long> txnTables) {
       super();
       this.queryText = queryText;
       this.txnWriteIdListProvider = txnWriteIdListProvider;
+      this.txnTables = txnTables;
     }
 
     public String getQueryText() {
@@ -671,6 +674,15 @@ public final class QueryResultsCache {
    */
   private boolean entryMatches(LookupInfo lookupInfo, CacheEntry entry, Set<CacheEntry> entriesToRemove) {
     QueryInfo queryInfo = entry.getQueryInfo();
+
+    if(!queryInfo.getInputs()
+            .stream()
+            .map(ReadEntity::getTable)
+            .map(Table::getTTable)
+            .map(org.apache.hadoop.hive.metastore.api.Table::getId)
+            .collect(Collectors.toSet()).containsAll(lookupInfo.txnTables))
+        return false;
+
     for (ReadEntity readEntity : queryInfo.getInputs()) {
       // Check that the tables used do not resolve to temp tables.
       if (readEntity.getType() == Type.TABLE) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
@@ -15789,12 +15789,22 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
     return null;
   }
 
+  private Set<Long> getTransactionedTables() throws SemanticException {
+    return tablesFromReadEntities(inputs)
+            .stream()
+            .filter(AcidUtils::isTransactionalTable)
+            .map(Table::getTTable)
+            .map(org.apache.hadoop.hive.metastore.api.Table::getId)
+            .collect(Collectors.toSet());
+  }
+
   private QueryResultsCache.LookupInfo createLookupInfoForQuery(ASTNode astNode) throws SemanticException {
     QueryResultsCache.LookupInfo lookupInfo = null;
     String queryString = getQueryStringForCache(astNode);
     if (queryString != null) {
       ValidTxnWriteIdList writeIdList = getQueryValidTxnWriteIdList();
-      lookupInfo = new QueryResultsCache.LookupInfo(queryString, () -> writeIdList);
+      Set<Long> txnTables = getTransactionedTables();
+      lookupInfo = new QueryResultsCache.LookupInfo(queryString, () -> writeIdList, txnTables);
     }
     return lookupInfo;
   }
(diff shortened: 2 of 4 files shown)


difin:
Please fill in the PR template to give some information about these changes.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -15789,12 +15789,22 @@ private ValidTxnWriteIdList getQueryValidTxnWriteIdList() throws SemanticExcepti
     return null;
   }
 
+  private Set<Long> getTransactionedTables() throws SemanticException {
```
difin, line 15792: Typo: should be `Transactional` instead of `Transactioned`, and also the method name is confusing because it returns table IDs, not tables. I think `getTransactionalTableIDs` would be a more clear name.

ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
```diff
@@ -671,6 +674,15 @@ public void notifyTableChanged(String dbName, String tableName, long updateTime)
    */
   private boolean entryMatches(LookupInfo lookupInfo, CacheEntry entry, Set<CacheEntry> entriesToRemove) {
     QueryInfo queryInfo = entry.getQueryInfo();
+
+    if(!queryInfo.getInputs()
+            .stream()
+            .map(ReadEntity::getTable)
+            .map(Table::getTTable)
```
deniskuzZ, line 681: i don't think that would work with the MV
````
          Table tbl = entity.getTable();
          if (tbl.isMaterializedView() && tbl.getMVMetadata() != null) {
            return tbl.getMVMetadata().getSourceTables().stream().map(SourceTable::getTable);
          }
````

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java:681]: avoid code duplication, the same thing in done in SA

File: ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java [hunk 3]
```diff
@@ -671,6 +674,15 @@ public void notifyTableChanged(String dbName, String tableName, long updateTime)
...
+            .map(org.apache.hadoop.hive.metastore.api.Table::getId)
+            .collect(Collectors.toSet()).containsAll(lookupInfo.txnTables))
+        return false;
```
zabetak, line 684: Adding a holistic check here incurs some overhead since we duplicate some work that is already done as part of the for loop just below. Since we are already iterating over the read entities it may be better to simplify and move the check inside the loop.

```java
lookupInfo.tableIds.contains(tableUsed.getTTable().getId())
```

Moreover, if we detect that a certain cache entry contains a table ID that is not part of the lookupInfo we should determine if we should/can invalidate and remove that entry from the cache. In the current, approach we simply bail-out and leave the entry inside the cache.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -15789,12 +15789,22 @@ private ValidTxnWriteIdList getQueryValidTxnWriteIdList() throws SemanticExcepti
...
+            .collect(Collectors.toSet());
+  }
+
   private QueryResultsCache.LookupInfo createLookupInfoForQuery(ASTNode astNode) throws SemanticException {
     QueryResultsCache.LookupInfo lookupInfo = null;
     String queryString = getQueryStringForCache(astNode);
     if (queryString != null) {
       ValidTxnWriteIdList writeIdList = getQueryValidTxnWriteIdList();
-      lookupInfo = new QueryResultsCache.LookupInfo(queryString, () -> writeIdList);
+      Set<Long> txnTables = getTransactionedTables();
```
zabetak, line 15806: Getting the tables ids is almost a one-liner so you could possibly just inline the code.
```java
Set<Long> tableIds =
          tablesFromReadEntities(inputs).stream().map(Table::getTTable).map(t -> t.getId()).collect(Collectors.toSet());
```

ql/src/test/queries/clientpositive/results_cache_invalidation3.q
```diff
@@ -0,0 +1,16 @@
+set hive.support.concurrency=true;
+set hive.txn.manager=org.apache.hadoop.hive.ql.lockmgr.DbTxnManager;
+
+set hive.query.results.cache.enabled=true;
+set hive.query.results.cache.nontransactional.tables.enabled=false;
```
zabetak, line 5: It would be nice to add also another test case (e.g., `results_cache_invalidation4.q`) when `hive.query.results.cache.nontransactional.tables.enabled=true` using non-transactional tables. It seems that if the table is dropped with the new logic we should be able to detect if the cache entry is valid or not.

ql/src/test/queries/clientpositive/results_cache_invalidation3.q
```diff
@@ -0,0 +1,16 @@
...
+
+CREATE TABLE author (fname STRING) STORED AS ORC TBLPROPERTIES('transactional'='true');
+INSERT INTO author VALUES ('Victor');
+SELECT fname FROM author;
+
+DROP TABLE author;
+
+CREATE TABLE author (fname STRING) STORED AS ORC TBLPROPERTIES('transactional'='true');
+INSERT INTO author VALUES ('Alexander');
+SELECT fname FROM author;
```
zabetak, line 16: nit: Consider adding a new line 

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

zabetak (commented):
The proposed solution looks good. I added a few comments that may improve performance and also cover a few more problematic cases.

zabetak:
Hey @czxm , do you have time to drive this PR forward and address the review comments?

zabetak:
Hey bot please keep this PR open!

deniskuzZ:
@zabetak, we'll need to find a volunteer to drive this PR

ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
thomasrebele: @deniskuzZ, could you please explain what you mean by these two messages?

deniskuzZ: Please see `BaseSemanticAnalyzer#isRequiresOpenTransaction`. Also it looks like the code has changed since the PR was created

thomasrebele: @deniskuzZ, could you please elaborate your concerns a bit more? I'm not familiar with Hive's source code yet, so please provide some context.

I tried to create a similar test case that uses materialized views, however, with non-transactional tables I get an error `org.apache.hadoop.hive.ql.parse.SemanticException: Automatic rewriting for materialized view cannot be enabled if the materialized view uses non-transactional tables`, and with transactional tables I get an error `org.apache.hadoop.hive.ql.metadata.HiveException: MetaException(message:Cannot drop table as it is used in the following materialized views [default.author_mv])` when trying to drop the table that the MV depends on.

ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
thomasrebele: I think the condition is that all `lookupInfo.tableIds` need to be contained in the cache entry. The condition you've mentioned might have some lookupInfo.tableIds that are not covered.

We can collect the available table ids in the loop and move the check to the end of the method, though.

thomasrebele:
I've opened a new PR, continuing the work of @czxm: https://github.com/apache/hive/pull/6064

I've taken the review comments into account and implemented almost all of them, except for the [discussion related to MVs](https://github.com/apache/hive/pull/5482#discussion_r1789961954), where I don't understand the purpose of the comment yet.

PR #5484, opened 2024-10-03
HIVE-28416: Another process is rebuilding the materialized view is thrown for subsequent MV rebuilds when the current MV rebuild is aborted for the same MV.

mdayakar:
Due to any issue if Materialized View Rebuild fails then the transaction will be aborted in that flow the MV rebuild lock entry is not getting removed from MATERIALIZATION_REBUILD_LOCKS table so subsequent rebuilds for that MV are failing. So as a part of this PR, in abort flow also deleting the MV rebuild lock from MATERIALIZATION_REBUILD_LOCKS table so that subsequent MV rebuild will not have any issue.

Actually there is one MaterializationsRebuildLockCleanerTask which cleans the outdated MV rebuild locks based on the lock heartbeat time but MaterializationRebuildLockHeartbeater heartbeating periodically without checking state of the transaction due to that MaterializationsRebuildLockCleanerTask is not cleaning the outdated locks. Now before heartbearing, checking transaction is aborted one or not, if not then only heartbearing the MV rebuild lock.

Without the changes, subsequent MV rebuild will always fail if previous MV rebuild had aborted due to any reason.

No

No

Test cases added.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
@@ -67,6 +67,7 @@ import javax.annotation.concurrent.NotThreadSafe;
 import java.io.IOException;
 import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -1139,6 +1140,9 @@ public final class DbTxnManager extends HiveTxnManagerImpl {
    */
   private static class MaterializationRebuildLockHeartbeater implements Runnable {
 
+    private static final List<TxnType> EXCLUDE_TXN_TYPES =
+        Arrays.asList(TxnType.DEFAULT, TxnType.REPL_CREATED, TxnType.READ_ONLY, TxnType.COMPACTION, TxnType.SOFT_DELETE,
+            TxnType.REBALANCE_COMPACTION);
     private final DbTxnManager txnMgr;
     private final String dbName;
     private final String tableName;
@@ -1165,7 +1169,7 @@ public final class DbTxnManager extends HiveTxnManagerImpl {
           AcidUtils.getFullTableName(dbName, tableName), queryId);
       boolean refreshed = false;
       try {
-        if (!txnMgr.getValidTxns().isTxnAborted(txnId)) {
+        if (!txnMgr.getValidTxns(EXCLUDE_TXN_TYPES).isTxnAborted(txnId)) {
           refreshed = txnMgr.heartbeatMaterializationRebuildLock(dbName, tableName, txnId);
         } else {
           LOG.info(
--- a/ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
@@ -1163,9 +1163,15 @@ public final class DbTxnManager extends HiveTxnManagerImpl {
     public void run() {
       LOG.trace("Heartbeating materialization rebuild lock for {} for query: {}",
           AcidUtils.getFullTableName(dbName, tableName), queryId);
-      boolean refreshed;
+      boolean refreshed = false;
       try {
-        refreshed = txnMgr.heartbeatMaterializationRebuildLock(dbName, tableName, txnId);
+        if (!txnMgr.getValidTxns().isTxnAborted(txnId)) {
+          refreshed = txnMgr.heartbeatMaterializationRebuildLock(dbName, tableName, txnId);
+        } else {
+          LOG.info(
+              "Transaction {} is aborted so stopping the heartbeat for materialization rebuild lock for {} for query: {}",
+              txnId, AcidUtils.getFullTableName(dbName, tableName), queryId);
+        }
       } catch (LockException e) {
         LOG.error("Failed trying to acquire lock", e);
         throw new RuntimeException(e);
--- a/ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
@@ -67,7 +67,6 @@ import javax.annotation.concurrent.NotThreadSafe;
 import java.io.IOException;
 import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -1140,9 +1139,6 @@ public final class DbTxnManager extends HiveTxnManagerImpl {
    */
   private static class MaterializationRebuildLockHeartbeater implements Runnable {
 
-    private static final List<TxnType> EXCLUDE_TXN_TYPES =
-        Arrays.asList(TxnType.DEFAULT, TxnType.REPL_CREATED, TxnType.READ_ONLY, TxnType.COMPACTION, TxnType.SOFT_DELETE,
-            TxnType.REBALANCE_COMPACTION);
     private final DbTxnManager txnMgr;
     private final String dbName;
     private final String tableName;
@@ -1167,15 +1163,9 @@ public final class DbTxnManager extends HiveTxnManagerImpl {
     public void run() {
       LOG.trace("Heartbeating materialization rebuild lock for {} for query: {}",
           AcidUtils.getFullTableName(dbName, tableName), queryId);
-      boolean refreshed = false;
+      boolean refreshed;
       try {
-        if (!txnMgr.getValidTxns(EXCLUDE_TXN_TYPES).isTxnAborted(txnId)) {
-          refreshed = txnMgr.heartbeatMaterializationRebuildLock(dbName, tableName, txnId);
-        } else {
-          LOG.info(
-              "Transaction {} is aborted so stopping the heartbeat for materialization rebuild lock for {} for query: {}",
-              txnId, AcidUtils.getFullTableName(dbName, tableName), queryId);
-        }
+        refreshed = txnMgr.heartbeatMaterializationRebuildLock(dbName, tableName, txnId);
       } catch (LockException e) {
         LOG.error("Failed trying to acquire lock", e);
         throw new RuntimeException(e);
--- a/ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
@@ -4702,12 +4702,12 @@ public class TestDbTxnManager2 extends DbTxnManagerEndToEndTestBase {
     driver.run("drop materialized view if exists mv_tab_acid");
     dropTable(new String[] { "tab_acid" });
 
-    driver.run(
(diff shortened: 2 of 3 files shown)


kasakrisz:
Please fix the issues found by sonarcloud

ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
```diff
@@ -4696,4 +4697,37 @@ public void testMaterializedViewEnableRewriteNonBlocking() throws Exception {
...
+
+    try {
+      driver.compileAndRespond("alter materialized view mv_tab_acid rebuild");
+    }
+    catch (Exception se) {
+      Assert.assertEquals(
+          "FAILED: SemanticException org.apache.hadoop.hive.ql.parse.SemanticException: Another process is rebuilding the materialized view default.mv_tab_acid",
+          se.getMessage());
+      Assert.fail("Should not throw exception here");
+    }
```
kasakrisz, line 4729: nit.: `Assert.assertThrows` can be used instead of this.

mdayakar:
> Please fix the issues found by sonarcloud

First comment is related to **Invoke method(s) only conditionally.** at [line](https://github.com/apache/hive/pull/5484/files#diff-64f581d32594c91d8cdf4a64745604a68b9f915653a65ba1695929f01bf0ad68R1173) which looks false positive as it is part of else block and also the method will be invoked when INFO is enabled so it is already invoked conditionally.

Second comment (related to unused import) fixed.

ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
mdayakar: Here I am not expecting exception, thats why Assert.fail() is added in catch block to fail the test case. As per my understanding Assert.assertThrows() is getting used when we expect some exception.

ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
```diff
@@ -4696,4 +4696,37 @@ public void testMaterializedViewEnableRewriteNonBlocking() throws Exception {
...
```
difin, line 4728: If an exception is not expected here, than why is this Assert needed?

```
Assert.assertEquals(
          "FAILED: SemanticException org.apache.hadoop.hive.ql.parse.SemanticException: Another process is rebuilding the materialized view default.mv_tab_acid",
          se.getMessage());
```

mdayakar: Before fix, it was throwing exception and after fix it shouldn't throw exception. 
If later someone changes code in such a way that again it throws exception then Assert.fail() won't give much information so added assertEquals also which will fail if it throws any different exception. I can include exception message in Assert.fail() also but anyway I feel above code is also fine.

kasakrisz: >  it won't give much information

AFAIK the exception message and stacktrace are present in failed test report. What is the missing information you refer to?

mdayakar: I mean here I just included "Should not throw exception here" message in fail() so failed report will show only this message. Which exception is thrown and other details are will not be shown. Anyway to remove the confusion I will change Assert.fail() message and remove assertEquals().

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
```diff
@@ -1163,9 +1163,15 @@ private static class MaterializationRebuildLockHeartbeater implements Runnable {
     public void run() {
       LOG.trace("Heartbeating materialization rebuild lock for {} for query: {}",
           AcidUtils.getFullTableName(dbName, tableName), queryId);
-      boolean refreshed;
+      boolean refreshed = false;
       try {
-        refreshed = txnMgr.heartbeatMaterializationRebuildLock(dbName, tableName, txnId);
+        if (!txnMgr.getValidTxns().isTxnAborted(txnId)) {
```
InvisibleProgrammer, line 1168: Selecting all the transactions in a heartbeat, just to check if the txn is aborted looks like an unnecessary resource consumption that we could avoid.

kasakrisz, line 1168: IIUC only MV rebuild txns are counts here.
Can we use `getValidTxns(List<TxnType> excludeTxnTypes)` and exclude everything but `MATER_VIEW_REBUILD`?

kasakrisz (changes requested):

zratkai:
Instead of the issue "Hive materialized views rebuild issue." give a description what is solves. Think like when you create a description about a new version a software you do not describe the issues, but the fixes. Something like : "Ensuring that on aborting a Materialized View rebuild, the rebuild lock is removed to prevent subsequent failures, and improves lock cleaning by checking the transaction state before heartbeating."

InvisibleProgrammer:
If I understood it correctly, currently, an aborted MV build blocks future rebuilds. So that, we want to narrow down the time window between the abort and the clean up by utilizing the heartbeat process. Am I correct? 

If yes, I would suggest to think about fixing the question why an aborted txn blocks future transaction instead of spreading the cleaning logic between multiple components. I believe a cleaning should be done by a cleaner process. Not a heartbeat one.

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
mdayakar: Changed as suggested.

mdayakar: Changed now as per @kasakrisz suggestion.

mdayakar:
> Instead of the issue "Hive materialized views rebuild issue." give a description what is solves. Think like when you create a description about a new version a software you do not describe the issues, but the fixes. Something like : "Ensuring that on aborting a Materialized View rebuild, the rebuild lock is removed to prevent subsequent failures, and improves lock cleaning by checking the transaction state before heartbeating."

Here "Hive materialized views rebuild issue." text is description for the issue. Issue will be raised before knowing the root cause for the main problem thats why the text shows what is the issue. After analysis and rootcause identification the details are explained in the PR description.

zratkai:
"Here "Hive materialized views rebuild issue." text is description for the issue" I understand that. In JIRA ticket it is great. But this is a commit message.
Please have a look on the commits what kind of message is used in general:
image

We use terms like "Fix for...", "Address CVE", "Impove" etc.

ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
difin: Looks fine now

difin: Yes, makes sense.

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
```diff
@@ -1163,9 +1167,15 @@ private static class MaterializationRebuildLockHeartbeater implements Runnable {
     public void run() {
       LOG.trace("Heartbeating materialization rebuild lock for {} for query: {}",
           AcidUtils.getFullTableName(dbName, tableName), queryId);
-      boolean refreshed;
+      boolean refreshed = false;
       try {
-        refreshed = txnMgr.heartbeatMaterializationRebuildLock(dbName, tableName, txnId);
+        if (!txnMgr.getValidTxns(EXCLUDE_TXN_TYPES).isTxnAborted(txnId)) {
```
InvisibleProgrammer, line 1172: That change improves nothing speaking of the original idea that I wrote before. And also, it introduce a new problem. 

I wrote this originally: 
> Selecting all the transactions in a heartbeat, just to check if the txn is aborted looks like an unnecessary resource consumption that we could avoid.

If you check the implementation in TxnHandler, you will see that excluding transaction types is just a cosmetic change: you save some bytes in API calls but still, you select all the transactions from the metastore database. So now, you can only save some network bandwith between metastore and HS2. 
You can see the select statements in `GetOpenTxnsListHandler` and the filtering logic in `OpenTxnList.toOpenTxnsResponse`. 

Secondly, that implementation defines everything as an exclusion. And keeps one type. That makes hard to introduce new transaction types. Imagine: you have to introduce a new transaction type, 2 years from now. Will you remember that you have to add that type in an exclusion list at a materialized view lock heartbeater? 
Imagine an other approach: what about using inclusions, not exclusions and saying we only need `TxnType.MATER_VIEW_REBUILD`? 

At this point, I don't know what is the best solution: keeping this logic is a no from my side as it increases the complexity without having any performance boost. And also introduces a possible problem in case we want to extend `TxnType`. 
But I also now that using an inclusion, not an exclusion, leads way more code changes: changes in the sql string , modifying the TxnHandler interface, re-generating Thrift stuff... It is a tedious task to do. But without it, I see more problem with that change than value.

ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
```diff
@@ -4696,4 +4696,37 @@ public void testMaterializedViewEnableRewriteNonBlocking() throws Exception {
     txnMgr.rollbackTxn();
     driver.run("drop materialized view mv_tab_acid");
   }
+
+  @Test
+  public void testMaterializedViewRebuildLockForSameMV() throws Exception {
```
InvisibleProgrammer, line 4701: It seems that test case covers the problem. 
Some minor thoughts about it: 
- I would remove the unnecessary string concatenations, , like in line 4706 and 4710. I assume they are leftovers and they were multiple line statements at a previous iteration. 
- As the test case checks the proper locking, I would check the locks as well, like in `testMaterializedViewRebuildNoLocks`

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DbTxnManager.java
mdayakar: Totally agree with you. I will do one thing, I will create a separate issue to handle MV rebuild cleaner task/heart beater issue and as a part of this issue I will fix only current issue(Removing locak entry from MATERIALIZATION_REBUILD_LOCKS table in case transaction is aborted.)

mdayakar: Created [HIVE-28563](https://issues.apache.org/jira/browse/HIVE-28563) to fix the cleaner/heartbeater related issue.

ql/src/test/org/apache/hadoop/hive/ql/lockmgr/TestDbTxnManager2.java
mdayakar: - Regarding the string concatenations, its just copied from above testcase but during code format they came in a single line. I believe they used string concatenation for readability purpose as the string is big.
- Here we are already checking the entries from MATERIALIZATION_REBUILD_LOCKS table which is related to actual problem.

kasakrisz (approved):

PR #5487, opened 2024-10-04
HIVE-28555: Adding custom plugins to ReExecution

ryukobayashi:

ReExecution currently only works with defined plugins. By extending this, it becomes possible to extend it so that only specific actions are taken when a specific failure occurs. For example, we can rewrite properties or prevent re-execution only when a specific exception occurs due to a specific cause.

We can write a plugin like this:
OOM has occurred, but we want to identify the message and turn off Vectorized only.

Yes, it might be a good idea to add this to the following documentation:
https://cwiki.apache.org/confluence/display/Hive/Query+ReExecution

No.

I added a UnitTest to test it.

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -5684,6 +5684,11 @@ public class HiveConf extends Configuration {
             + "  recompile_without_cbo: recompiles query after a CBO failure\n"
             + "  reexecute_lost_am: reexecutes query if it failed due to tez am node gets decommissioned\n "
             + "  write_conflict: retries the query once if the query failed due to write_conflict"),
+    HIVE_QUERY_CUStOM_REEXECUTION_STRATEGIES("hive.query.custome.reexecution.strategies",
+        "",
+        "Define a custom reexecution strategies for hive.query.reexecution.strategies.:\n"
+            + "  e.g.\n"
+            + "    org.apache.hadoop.hive.ql.reexec.custom.CustomPlugin1,org.apache.hadoop.hive.ql.reexec.custom.CustomPlugin2"),
     HIVE_QUERY_REEXECUTION_STATS_PERSISTENCE("hive.query.reexecution.stats.persist.scope", "metastore",
         new StringSet("query", "hiveserver", "metastore"),
         "Sets the persistence scope of runtime statistics\n"
--- a/ql/src/java/org/apache/hadoop/hive/ql/DriverFactory.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/DriverFactory.java
@@ -58,6 +58,14 @@ public final class DriverFactory {
       plugins.add(buildReExecPlugin(string));
     }
 
+    String customeStrategies = queryState.getConf().getVar(ConfVars.HIVE_QUERY_CUStOM_REEXECUTION_STRATEGIES);
+    for (String string : customeStrategies.split(",")) {
+      if (string.trim().isEmpty()) {
+        continue;
+      }
+      plugins.add(buildCustomReExecPlugin(string));
+    }
+
     return new ReExecDriver(queryState, queryInfo, plugins);
   }
 
@@ -66,6 +74,22 @@ public final class DriverFactory {
     try {
       return pluginType.newInstance();
     } catch (InstantiationException | IllegalAccessException e) {
+      throw new RuntimeException(
+          "Unknown re-execution plugin: " + name + " (" + ConfVars.HIVE_QUERY_CUStOM_REEXECUTION_STRATEGIES.varname + ")");
+    }
+  }
+
+  private static IReExecutionPlugin buildCustomReExecPlugin(String name) throws RuntimeException {
+    try {
+      Class<?> cls = Class.forName(name);
+      Object o = cls.newInstance();
+      if (!(o instanceof IReExecutionPlugin)) {
+        throw new RuntimeException(
+            "Not re-execution plugin: " + name);
+      }
+
+      return (IReExecutionPlugin) o;
+    } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
       throw new RuntimeException(
           "Unknown re-execution plugin: " + name + " (" + ConfVars.HIVE_QUERY_REEXECUTION_STRATEGIES.varname + ")");
     }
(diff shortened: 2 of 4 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -5684,6 +5684,11 @@ public static enum ConfVars {
             + "  recompile_without_cbo: recompiles query after a CBO failure\n"
             + "  reexecute_lost_am: reexecutes query if it failed due to tez am node gets decommissioned\n "
             + "  write_conflict: retries the query once if the query failed due to write_conflict"),
+    HIVE_QUERY_CUSTOM_REEXECUTION_STRATEGIES("hive.query.custom.reexecution.strategies",
```
deniskuzZ, line 5687: not sure if we need both configs. I think we can create a generic config `hive.query.reexecution.strategies` and list existing strategies there similar to `metastore.task.threads.remote`
````
org.apache.hadoop.hive.ql.reexec.ReExecutionOverlayPlugin,
org.apache.hadoop.hive.ql.reexec.ReExecuteLostAMQueryPlugin,
org.apache.hadoop.hive.ql.reexec.ReCompileWithoutCBOPlugin,
org.apache.hadoop.hive.ql.reexec.ReExecutionDagSubmitPlugin,
org.apache.hadoop.hive.ql.reexec.ReExecuteOnWriteConflictPlugin;
````

Comment (deniskuzZ) [reply common/src/java/org/apache/hadoop/hive/conf/HiveConf.java:5687]: or maybe have mix of enum defined strategies and custom classes
````
overlay,reoptimize,reexecute_lost_am,dagsubmit,recompile_without_cbo,write_conflict,org.apache.hadoop.hive.ql.reexec.custom.CustomPlugin1
````

Comment (ryukobayashi) [reply common/src/java/org/apache/hadoop/hive/conf/HiveConf.java:5687]: I see. I will do that.

Comment (ryukobayashi) [reply common/src/java/org/apache/hadoop/hive/conf/HiveConf.java]: @deniskuzZ  I fixed that.

File: ql/src/java/org/apache/hadoop/hive/ql/DriverFactory.java [hunk 2]
```diff
@@ -49,20 +47,29 @@ public static IDriver newDriver(QueryState queryState, QueryInfo queryInfo) {
...
+      } else {
+        plugins.add(buildCustomReExecPlugin(string));
+      }
     }
 
     return new ReExecDriver(queryState, queryInfo, plugins);
   }
 
   private static IReExecutionPlugin buildReExecPlugin(String name) throws RuntimeException {
     Class<? extends IReExecutionPlugin> pluginType = ReExecutionStrategyType.getPluginClassByName(name);
```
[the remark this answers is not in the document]
ryukobayashi: Sorry, I just wanted to delete `toLowerCase`, but I ended up deleting one line. So, I changed it back and remove `toLowerCase`.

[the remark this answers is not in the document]
ryukobayashi: Sorry, I just wanted to remove `toLowerCase`, but I ended up removing one line. So, I changed it back and remove `toLowerCase`.

ql/src/java/org/apache/hadoop/hive/ql/DriverFactory.java
```diff
@@ -49,20 +47,29 @@ public static IDriver newDriver(QueryState queryState, QueryInfo queryInfo) {
...
+    if (pluginType == null) {
```
deniskuzZ, line 72: why not catch  exception and do a fallback here

````
    Class<?> pluginType;
    try {
      pluginType = ReExecutionStrategyType.getPluginClassByName(name);
    } catch (IllegalArgumentException e) {
      pluginType = Class.forName(name);
      if (!pluginType.isAssignableFrom(IReExecutionPlugin.class)) {
        throw new RuntimeException("Not re-execution plugin: " + name);
      }
    }
    try {
      return (IReExecutionPlugin) pluginType.newInstance();
    } catch (InstantiationException | IllegalAccessException e) {
      throw new RuntimeException(
          "Unknown re-execution plugin: " + name + " (" + ConfVars.HIVE_QUERY_REEXECUTION_STRATEGIES.varname + ")");
    }
````

File: ql/src/java/org/apache/hadoop/hive/ql/reexec/ReExecutionStrategyType.java [hunk 4]
```diff
@@ -39,6 +37,10 @@ public Class<? extends IReExecutionPlugin> getPluginClass() {
   }
 
   public static Class<? extends IReExecutionPlugin> getPluginClassByName(String strategy) {
-    return ReExecutionStrategyType.valueOf(strategy.toUpperCase()).getPluginClass();
+    try {
+      return ReExecutionStrategyType.valueOf(strategy.toUpperCase()).getPluginClass();
+    } catch (IllegalArgumentException e) {
+      return null;
```
deniskuzZ, line 43: no need for that change

ql/src/java/org/apache/hadoop/hive/ql/reexec/ReExecDriver.java
```diff
@@ -83,6 +83,11 @@ public int compile(String command, boolean resetTaskIds) {
     return coreDriver.compile(command, resetTaskIds);
   }
 
+  @VisibleForTesting
+  public List<IReExecutionPlugin> getPlugins() {
```
deniskuzZ, line 87: could it be package private if we move the testt under reexec package?

ql/src/java/org/apache/hadoop/hive/ql/DriverFactory.java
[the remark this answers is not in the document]
ryukobayashi: OK, I will remove trim.

ql/src/java/org/apache/hadoop/hive/ql/DriverFactory.java
ryukobayashi: OK, move the process.

ql/src/java/org/apache/hadoop/hive/ql/reexec/ReExecutionStrategyType.java
ryukobayashi: Yes, with the above changes it is no longer necessary.

ql/src/java/org/apache/hadoop/hive/ql/reexec/ReExecDriver.java
ryukobayashi: I think we cannot make a package private by moving it.

deniskuzZ (approved):
LGTM +1

ryukobayashi:
@deniskuzZ Thanks. I fixed unused imports, etc.

ryukobayashi:
The test failed so I rebased it, but some conflicts occurred so I will fix them.

PR #5488, opened 2024-10-04
HIVE-28556: Do not clone Statistics before scaleToRowCount()

ngsg:

Remove an unnecessary Statsistics.clone() invocation in StatsRulesProcFactory.applyRuntimeStats().

Current StatsRulesProcFactory.applyRuntimeStats() clones original stats and then calls scaleToRowCount() on it.
But Statistics.scaleToRowCount() always returns the clone of itself, so we don't have to clone the original stats in StatsRulesProcFactory.applyRuntimeStats().

No

No

Tested in local environment

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
@@ -3294,7 +3294,6 @@ public class StatsRulesProcFactory {
       return stats;
     }
     LOG.debug("using runtime stats for {}; {}", op, os.get());
-    // Statistics.scaleToRowCount() creates and returns the clone of itself.
     Statistics outStats = stats.scaleToRowCount(os.get().getOutputRecords(), false);
     outStats.setRuntimeStats(true);
     return outStats;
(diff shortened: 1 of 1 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
```diff
@@ -3294,8 +3294,8 @@ private static Statistics applyRuntimeStats(Context context, Statistics stats, O
       return stats;
     }
     LOG.debug("using runtime stats for {}; {}", op, os.get());
-    Statistics outStats = stats.clone();
-    outStats = outStats.scaleToRowCount(os.get().getOutputRecords(), false);
+    // Statistics.scaleToRowCount() creates and returns the clone of itself.
```
deniskuzZ, line 3297: not sure if we need this comment

deniskuzZ (approved):
+1, pending tests

okumin (approved):

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
ngsg: removed the comment

PR #5489, opened 2024-10-04
HIVE-28557: Eliminate a lot of metrics null check from LlapTaskSchedulerService

abstractdog:
1. Introduce a DummyLlapTaskSchedulerMetrics to eliminate 29 null checks.
2. Moving all metrics system initialization to the LlapTaskSchedulerMetrics class where it belongs.

Eliminate boilerplate.
Repeating null check is a code smell that implies there is no clear contract about whether the field is null. This is now handled by initializing a dummy instance and letting the rest of the code ignore what's going on with metrics actually.

No.

No.

Precommit.

### DIFF
--- a/llap-tez/src/java/org/apache/hadoop/hive/llap/tezplugins/metrics/DummyLlapTaskSchedulerMetrics.java
+++ b/llap-tez/src/java/org/apache/hadoop/hive/llap/tezplugins/metrics/DummyLlapTaskSchedulerMetrics.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


abstractdog:
test failures are related, need a bit of rework before the next patch

llap-tez/src/java/org/apache/hadoop/hive/llap/tezplugins/metrics/DummyLlapTaskSchedulerMetrics.java
```diff
@@ -0,0 +1,23 @@
...
+
+package org.apache.hadoop.hive.llap.tezplugins.metrics;
+
+import org.apache.hadoop.conf.Configuration;
+
+public class DummyLlapTaskSchedulerMetrics extends LlapTaskSchedulerMetrics {
+  public DummyLlapTaskSchedulerMetrics(Configuration conf) {
+    super(conf);
+  }
+}
```
difin, line 23: This class seems to have the same behavior as its parent `LlapTaskSchedulerMetrics` and therefore it is not a dummy metrics handler? I though the idea of `DummyLlapTaskSchedulerMetrics` was to implement metrics handling with dummy functionality.

abstractdog: yeah, that's what cause so many testing failures, I'm working on it!
thanks for taking a look

PR #5491, opened 2024-10-07
HIVE-28560: AssertionError for jdbc queries because of incorrec…

soumyakanti3578:
…t expectedRel in SqlImplementor

https://issues.apache.org/jira/browse/HIVE-28560

Call `setFetch` before the expectedRel is changed to a `Project`

We get an `AssertionError` otherwise

No

No

```
mvn test -pl itests/qtest  -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile=jdbc_sort_limit.q
```


ql/src/test/queries/clientpositive/jdbc_sort_limit.q
```diff
@@ -0,0 +1,132 @@
+--!qt:database:postgres:q_test_tpcds_tables.sql
+
+CREATE EXTERNAL TABLE IF NOT EXISTS `date_dim`(
+  `d_date_sk` int, 
+  `d_date_id` string,
```
difin, line 5: I don't understand the purpose of including this file `q_test_tpcds_tables.sql` here.
The same tables that are created in that file are also created in this q-test.

soumyakanti3578: `q_test_tpcds_tables.sql` creates them in a postgres database in a docker container.
In this file, we create external tables for the jdbc tables, and eventually use these external tables for running qtests.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/HiveJdbcImplementor.java
```diff
@@ -68,6 +68,18 @@ public HiveJdbcImplementor(SqlDialect dialect, JavaTypeFactory typeFactory) {
...
+      builder = x.builder(e);
+      builder.setFetch(builder.context.toSql(null, e.fetch));
+      x = builder.result();
+    }
+    if (e.offset != null) {
+      builder = x.builder(e);
+      builder.setOffset(builder.context.toSql(null, e.offset));
+      x = builder.result();
+    }
+
```
zabetak, line 82: The first question that comes to mind is if this problem affects Calcite or if it is only Hive specific. I get the impression that it is mostly due to the Hive override that seems somewhat broken. From a quick look, I think we can get rid of this override completely and rely entirely on Calcite `visit(Sort)` method.

ql/src/test/queries/clientpositive/jdbc_sort_limit.q
```diff
@@ -0,0 +1,132 @@
...
+   and store_sales.ss_item_sk = item.i_item_sk
+   and item.i_manufact_id = 436
+   and dt.d_moy=12
+ group by dt.d_year
+      ,item.i_brand
+      ,item.i_brand_id
+ order by dt.d_year
+         ,sum_agg desc
+         ,brand_id
+limit 100;
```
zabetak, line 132: The repro test is unnecessarily verbose making hard to understand where exactly is the problem and why it is fixed by the proposed changes. According to the description the problem should be visible in any query containing a limit.

How about just adding the following "jdbc_table_limit_postgres.q":
```sql
--! qt:database:postgres:q_test_country_table.sql

CREATE EXTERNAL TABLE country (id int, name varchar(20))
STORED BY 'org.apache.hive.storage.jdbc.JdbcStorageHandler'
TBLPROPERTIES (
    "hive.sql.database.type" = "POSTGRES",
    "hive.sql.jdbc.driver" = "org.postgresql.Driver",
    "hive.sql.jdbc.url" = "jdbc:postgresql://localhost:5432/qtestDB",
    "hive.sql.dbcp.username" = "qtestuser",
    "hive.sql.dbcp.password" = "qtestpassword",
    "hive.sql.table" = "country");

EXPLAIN CBO SELECT * FROM country ORDER BY id LIMIT 2;
EXPLAIN SELECT * FROM country ORDER BY id LIMIT 2;
SELECT * FROM country ORDER BY id LIMIT 2;
```
The above test case is minimal, and allows us to test all aspects of the plan (compilation+execution).

It may also be nice to reword the JIRA case using the minimal example.

ql/src/test/queries/clientpositive/jdbc_sort_limit.q
```diff
@@ -0,0 +1,132 @@
...
+TBLPROPERTIES (                                    
+    "hive.sql.database.type" = "POSTGRES",
+    "hive.sql.jdbc.driver" = "org.postgresql.Driver",
+    "hive.sql.jdbc.url" = "jdbc:postgresql://localhost:5432/qtestDB",
+    "hive.sql.dbcp.username" = "qtestuser",
+    "hive.sql.dbcp.password" = "qtestpassword",
+    "hive.sql.table" = "item"
+);
+
+set hive.mapred.mode=nonstrict;
```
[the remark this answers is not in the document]
soumyakanti3578: Removed this and added a new test as recommended below

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/HiveJdbcImplementor.java
soumyakanti3578: Removed the overridden method from the class.

ql/src/test/queries/clientpositive/jdbc_sort_limit.q
soumyakanti3578: Thanks! I have added the test as suggested here.

zabetak (approved):
LGTM, will merge soon!

PR #5492, opened 2024-10-08
HIVE-28549: Limit the maximum number of operators merged by SharedWorkOptimizer

okumin:

This PR would limit the maximum number of table scan operators which SWO tries to merge.

https://issues.apache.org/jira/browse/HIVE-28549

We observed SWO makes a negative impact when it merges too many, e.g. 50, operators. If operators are memory intensive, they might throw OOM or might slow down.

I believe we can resolve OOM with the following patch, but we still want an upper limit so that we can tune concurrency or RAM per operator reasonably.
https://github.com/apache/hive/pull/5478

No.

No.

I added a qtest

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
@@ -37,7 +37,7 @@ import java.util.Objects;
 import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.Set;
-
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hive.common.TableName;
@@ -174,7 +174,6 @@ public class SharedWorkOptimizer extends Transform {
     gatherDPPTableScanOps(pctx, optimizerCache);
 
     final int batchSize = HiveConf.getIntVar(pctx.getConf(), ConfVars.HIVE_SHARED_WORK_MAX_SIBLINGS);
-    Preconditions.checkArgument(batchSize == -1 || batchSize > 0);
     for (List<TableScanOperator> scans : groupTableScanOperators(sortedTables, tableNameToOps, batchSize)) {
       // Execute shared work optimization
       runSharedWorkOptimization(pctx, optimizerCache, scans, Mode.SubtreeMerge);
@@ -254,14 +253,16 @@ public class SharedWorkOptimizer extends Transform {
 
   private static List<List<TableScanOperator>> groupTableScanOperators(List<Entry<String, Long>> sortedTables,
       ArrayListMultimap<String, TableScanOperator> tableNameToOps, int batchSize) {
+    if (batchSize == -1) {
+      return Collections.singletonList(sortedTables.stream().map(Entry::getKey)
+          .flatMap(tableName -> tableNameToOps.get(tableName).stream()).collect(Collectors.toList()));
+    }
+
+    Preconditions.checkArgument(batchSize > 0);
     final List<List<TableScanOperator>> batches = new ArrayList<>();
     for (Entry<String, Long> tablePair : sortedTables) {
       final String tableName = tablePair.getKey();
       final List<TableScanOperator> scans = tableNameToOps.get(tableName);
-      if (batchSize == -1) {
-        batches.add(scans);
-        continue;
-      }
 
       final int limit = scans.size();
       int from = 0;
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -248,6 +249,32 @@ public ParseContext transform(ParseContext pctx) throws SemanticException {
...
 
+  private static List<List<TableScanOperator>> groupTableScanOperators(List<Entry<String, Long>> sortedTables,
+      ArrayListMultimap<String, TableScanOperator> tableNameToOps, int batchSize) {
+    final List<List<TableScanOperator>> batches = new ArrayList<>();
+    for (Entry<String, Long> tablePair : sortedTables) {
+      final String tableName = tablePair.getKey();
+      final List<TableScanOperator> scans = tableNameToOps.get(tableName);
+      if (batchSize == -1) {
+        batches.add(scans);
+        continue;
```
deniskuzZ, line 260: using `continue` keyword is a bad practice

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -248,6 +249,32 @@ public ParseContext transform(ParseContext pctx) throws SemanticException {
...
+      }
+
+      final int limit = scans.size();
+      int from = 0;
+      while (from != limit) {
+        final int to = Math.min(limit, from + batchSize);
+        // We have to copy the list because it is mutated later
+        final List<TableScanOperator> subList = new ArrayList<>(scans.subList(from, to));
+        Preconditions.checkState(!subList.isEmpty());
```
ngsg, line 269: Could we change this Preconditions to an assertion or some other expressions that guarantee `!scans.subList(from, to).isEmpty()`? (e.g. `Preconditions.checkArgument(batchSize > 0)` at the top of the method.) Using Preconditions to check the invariant of code, rather than caller's or client's requirements, seems a bit odd to me.

okumin: The level of intention of this assertion might be lower than what you are thinking. I just added it to ensure I correctly implemented L265-L272 because I am worried too much. I think we can remove the line basically. Will do

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -248,6 +251,32 @@ public ParseContext transform(ParseContext pctx) throws SemanticException {
     return pctx;
   }
 
+  private static List<List<TableScanOperator>> groupTableScanOperators(List<Entry<String, Long>> sortedTables,
+      ArrayListMultimap<String, TableScanOperator> tableNameToOps, int batchSize) {
+    if (batchSize == -1) {
+      return Collections.singletonList(sortedTables.stream().map(Entry::getKey)
+          .flatMap(tableName -> tableNameToOps.get(tableName).stream()).collect(Collectors.toList()));
```
ngsg, line 258: This puts all TS ops into a single List, regardless of their source. Maybe you intend the following code?
```
    if (batchSize == -1) {
      return sortedTables.stream()
          .map(entry -> tableNameToOps.get(entry.getKey()))
    }
```

ngsg (approved):
LGTM, +1
(The failed test seems just a flaky one. In my local environment, the test finished successfully.)

okumin:
Retriggered CI just in case

okumin:
I am checking why `partition_explain_ddl` always fails. In my impression, this PR is unrelated.

okumin:
CI passed after merging https://github.com/apache/hive/pull/5555

deniskuzZ:
hi @okumin, should we rebase?

okumin:
@deniskuzZ 
Hi, I've rebased the branch on the master branch

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -251,6 +254,31 @@ public ParseContext transform(ParseContext pctx) throws SemanticException {
...
+    if (batchSize == -1) {
+      return sortedTables.stream().map(entry -> tableNameToOps.get(entry.getKey())).collect(Collectors.toList());
+    }
+
+    final List<List<TableScanOperator>> batches = new ArrayList<>();
+    for (Entry<String, Long> tablePair : sortedTables) {
+      final String tableName = tablePair.getKey();
+      final List<TableScanOperator> scans = tableNameToOps.get(tableName);
+
+      final int limit = scans.size();
```
deniskuzZ, line 268: why not use `scans.size()` directly?

deniskuzZ (approved):
LGTM, minor comment

deniskuzZ:
hi @okumin, not sure if you noticed the above comment, do you think it's legit?

okumin:
@deniskuzZ Yes, I do. I'm using my machine power to review and test another PR. I will likely update this one tomorrow.

okumin:
Thanks

PR #5494, opened 2024-10-08
HIVE-28564: AssertionError for jdbc queries when JDBCExpandExpression…

soumyakanti3578:
…sRule produces non flat Filter conditions

https://issues.apache.org/jira/browse/HIVE-28564
Simplify filter/join condition in JDBCExpandExpressionsRule so that they are flat

Without these changes, sometimes the new conditions are not flat and we get an AssertionError as shown in the ticket.

No

No

```
mvn test -pl itests/qtest  -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile=jdbc_non_flat_filter_condition.q
```


soumyakanti3578:
Hi @zabetak, could you please review this?

ql/src/test/queries/clientpositive/jdbc_filter_expand_row_operator.q
```diff
@@ -0,0 +1,21 @@
...
+
+explain cbo
+select * from book 
+where id = 0 or (id = 1 and author = 11) or (id = 2 and author = 22);
+
+-- non jdbc plan containing ROW, for reference
+CREATE EXTERNAL TABLE book1 (id int, title varchar(100), author int);
+explain cbo
+select * from book1
+where id = 0 or (id = 1 and author = 11) or (id = 2 and author = 22);
```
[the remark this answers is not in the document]
soumyakanti3578: Yes, it is not really required. I added it to highlight differences in the plan generated for JDBC and non JDBC queries. I have removed it in the latest commit.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/JDBCExpandExpressionsRule.java
```diff
@@ -159,9 +160,16 @@ public void onMatch(RelOptRuleCall call) {
...
   }
 
-  RexNode analyzeRexNode(RexBuilder rexBuilder, RexNode condition) {
-    RexTransformIntoOrAndClause transformIntoInClause = new RexTransformIntoOrAndClause(rexBuilder);
+  RexNode analyzeRexNode(RelOptCluster cluster, RexNode condition) {
+    RexTransformIntoOrAndClause transformIntoInClause = new RexTransformIntoOrAndClause(cluster.getRexBuilder());
     RexNode newCondition = transformIntoInClause.apply(condition);
+
+    if (!RexUtil.isFlat(newCondition)) {
+      newCondition = new RexSimplify(
```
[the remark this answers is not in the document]
soumyakanti3578: I tried `flatten` before trying `RexSimplify` and it didn't work. To me it looks like there is a bug in [isFlat(List<? extends RexNode> exprs, final SqlOperator op)](https://github.com/apache/calcite/blob/calcite-1.25.0/core/src/main/java/org/apache/calcite/rex/RexUtil.java#L1429):
```
  private static boolean isFlat(
      List<? extends RexNode> exprs, final SqlOperator op) {
    return !isAssociative(op)
        || !exists(exprs, (Predicate1<RexNode>) expr -> isCallTo(expr, op));
```
and [isCallTo](https://github.com/apache/calcite/blob/calcite-1.25.0/core/src/main/java/org/apache/calcite/rex/RexUtil.java#L942):
```
  public static boolean isCallTo(RexNode expr, SqlOperator op) {
    return (expr instanceof RexCall)
        && (((RexCall) expr).getOperator() == op);
  }
```
where we just check if the List of RexNodes are flat with respect to the SqlOperator. In this case, for the condition
```
AND(OR(=($0, 0), OR(AND(=($0, 1), =($2, 11)), AND(=($0, 2), =($2, 22)))), IN($0, 0, 1, 2))
```
it just checks whether the outer `AND` is flat with `IN`, and with the outer `OR`, but it doesn't go further and check if the list of operands themselves are flat. In this case, as we can see, `OR(=($0, 0), OR(AND(=($0, 1), =($2, 11)), AND(=($0, 2), =($2, 22))))` is not flat.

Comparing this to the overloaded [isFlat(RexNode expr)](https://github.com/apache/calcite/blob/calcite-1.25.0/core/src/main/java/org/apache/calcite/rex/RexUtil.java#L1436):
```
  public static boolean isFlat(RexNode expr) {
    if (!(expr instanceof RexCall)) {
      return true;
    }
    final RexCall call = (RexCall) expr;
    return isFlat(call.getOperands(), call.getOperator())
        && all(call.getOperands(), RexUtil::isFlat);
  }
```
we see a call to check if `all` operands are flat or not. This returns `false` for the `condition` shown above, and we call this method when checking the condition in Hive.

I was able to use `flatten` elsewhere in the new commit, and for now it seems to resolve this because it tries to flatten bottom up, right after an expression is created.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/JDBCExpandExpressionsRule.java
```diff
@@ -159,9 +160,16 @@ public void onMatch(RelOptRuleCall call) {
 
   }
 
-  RexNode analyzeRexNode(RexBuilder rexBuilder, RexNode condition) {
-    RexTransformIntoOrAndClause transformIntoInClause = new RexTransformIntoOrAndClause(rexBuilder);
+  RexNode analyzeRexNode(RelOptCluster cluster, RexNode condition) {
```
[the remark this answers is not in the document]
soumyakanti3578: Reverted this change.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/JDBCExpandExpressionsRule.java
```diff
@@ -178,6 +177,9 @@ protected static class RexTransformIntoOrAndClause extends RexShuttle {
     @Override
     public RexNode visitCall(RexCall inputCall) {
       RexNode node = super.visitCall(inputCall);
+      if (!RexUtil.isFlat(node)) {
+        node = RexUtil.flatten(rexBuilder, node);
+      }
```
zabetak, line 182: If we flatten the call here aren't we risking to create a non-flat expression during the IN to OR transformation below?

ql/src/test/results/clientpositive/llap/jdbc_filter_expand_row_operator.q.out
```diff
@@ -0,0 +1,41 @@
...
+#### A masked pattern was here ####
+POSTHOOK: query: explain cbo
+select * from book 
+where id = 0 or (id = 1 and author = 11) or (id = 2 and author = 22)
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@book
+#### A masked pattern was here ####
+CBO PLAN:
+HiveJdbcConverter(convention=[JDBC.POSTGRES])
+  JdbcFilter(condition=[AND(OR(=($0, 0), AND(=($0, 1), =($2, 11)), AND(=($0, 2), =($2, 22))), IN($0, 0, 1, 2))])
```
zabetak, line 39: The `IN($0, 0, 1, 2)` seems kinda redundant any idea why it's there?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/JDBCExpandExpressionsRule.java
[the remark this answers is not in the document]
zabetak: Not sure if there is a bug I haven't read the javadoc/code of the respective methods. If you think there is a bug feel free to raise a CALCITE ticket.

Apart from that, I am ok with the idea to flatten the expression bottom up.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/JDBCExpandExpressionsRule.java
zabetak: In fact, I was thinking that it may be better to flatten after the transformation but if you are saying that it's not necessary that's fine with me.
```java
public RexNode visitCall(RexCall inputCall) {
      RexNode node = super.visitCall(inputCall);
      if (node instanceof RexCall) {
        RexCall call = (RexCall) node;
        switch (call.getKind()) {
          case IN:
            node = transformIntoOrAndClause(rexBuilder, call);
          default:
            break;
        }
      }
      return RexUtil.isFlat(node) ? node : RexUtil.flatten(rexBuilder, node);
    }
```

ql/src/test/results/clientpositive/llap/jdbc_filter_expand_row_operator.q.out
zabetak: True it's probably out of the scope of this PR so let's create a separate ticket to track it further.

I am suspecting that `IN(ROW($6, $8), ROW(1999, 12), ROW(2001, 1))` cannot be easily pushed in some JDBC databases cause they may not support the ROW operator. Nevertheless, I don't see why we should have both
`OR(=($0, 0), AND(=($0, 1), =($2, 11)), AND(=($0, 2), =($2, 22)))` and  `IN($0, 0, 1, 2)` especially for the JDBC storage handler where the underlying DB is gonna apply its own optimizations.

zabetak (approved):

ql/src/test/results/clientpositive/llap/jdbc_filter_expand_row_operator.q.out
soumyakanti3578: No but for jdbc databases, we are expanding all the ROW operators. The issue is present for non-jdbc plans too. I will create a ticket for this 👍🏼

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/jdbc/JDBCExpandExpressionsRule.java
soumyakanti3578: Although it is doing the same thing, this indeed looks safer, so I have included it in the latest commit.

ql/src/test/results/clientpositive/llap/jdbc_filter_expand_row_operator.q.out
soumyakanti3578: I will create a new ticket for this after this PR is merged, so that I can reference the q.out file

PR #5495, opened 2024-10-09
HIVE-28565: Set lock.numretries to small value for tests

zabetak:
Set `hive.lock.sleep.between.retries` and `metastore.lock.sleep.between.retries` to 2 similar to what was done HIVE-6464.

Fail fast when there is problem to obtain a lock saving resources and providing quick feedback to the developpers instead of waiting ~1.6hrs.

No
No

Existing tests

### DIFF
--- a/conf/hive-site.xml
+++ b/conf/hive-site.xml
@@ -19,4 +19,8 @@
 
 <configuration>
 
+  <property>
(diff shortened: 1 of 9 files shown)


zabetak:
Can someone please review this PR?

simhadri-g (approved):
LGTM

PR #5496, opened 2024-10-09
HIVE-28566: Simplify version identifier in patched-iceberg modules

zabetak:
1. Drop redundant groupId specification
2. Use project.version in patched-iceberg modules

Align the patched-iceberg versioning with the rest of Hive modules and get rid of maven warnings.
```
[WARNING] 'version' contains an expression but should be a constant.
```
More info under HIVE-28566.

It may affect only projects using patched-iceberg modules via Maven.

In the next release the patched-iceberg-api and patched-iceberg-core modules will use the same version identifier as every other Hive module.

For instance, the next version for `org.apache.hive:patched-iceberg-api` will be `4.1.0` instead of `patched-1.6.1-4.1.0`

No

```
mvn clean install -DskipTests
Inspect content of:
ls ~/.m2/repository/org/apache/hive/patched-iceberg-*/4.1.0-SNAPSHOT/
```

### DIFF
--- a/iceberg/patched-iceberg-api/pom.xml
+++ b/iceberg/patched-iceberg-api/pom.xml
@@ -20,9 +20,7 @@
     <relativePath>../pom.xml</relativePath>
   </parent>
   <modelVersion>4.0.0</modelVersion>
-  <groupId>org.apache.hive</groupId>
   <artifactId>patched-iceberg-api</artifactId>
-  <version>patched-${iceberg.version}-${project.parent.version}</version>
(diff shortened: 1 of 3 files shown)


Aggarwal-Raghav:
LGTM +1 
- verified that the maven warning is gone.
- in local m2, dir is created as 4.1.0-SNAPSHOT instead of patched-1.6.1-4.1.0-SNAPSHOT.

zhangbutao (approved):
+1 LGTM

PR #5497, opened 2024-10-09
HIVE-28571: Basic UNIONTYPE support in CBO

zabetak:
1. Model union as struct at the CBO layer
2. Remove CalciteSemanticException from TypeConverter APIs (no longer thrown)
3. Update q.out files since queries with UNIONTYPE can now use CBO

Support UNIONTYPE in the CBO path to take advantage of the powerful optimizations that are performed in the CBO layer and avoid relying on the fallback mechanism. The changes do not aim to cover new use-cases but just to ensure that existing queries that involve UNIONTYPE can exploit the CBO.

Better plans

No

```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=unionDistinct_3.q,union21.q,join_thrift.q,udf_isnull_isnotnull.q,uniontype_cbo.q
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
@@ -56,7 +56,6 @@ import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.ql.Context;
 import org.apache.hadoop.hive.ql.exec.ColumnInfo;
 import org.apache.hadoop.hive.ql.log.PerfLogger;
-import org.apache.hadoop.hive.ql.optimizer.calcite.CalciteSemanticException;
 import org.apache.hadoop.hive.ql.optimizer.calcite.HiveTypeSystemImpl;
 import org.apache.hadoop.hive.ql.optimizer.calcite.RelOptHiveTable;
 import org.apache.hadoop.hive.ql.optimizer.calcite.reloperators.HiveRelNode;
@@ -421,13 +420,7 @@ public final class HiveMaterializedViewsRegistry {
     }
 
     // 1.3 Build row type from field <type, name>
-    RelDataType rowType;
-    try {
-      rowType = TypeConverter.getType(cluster, rr, null);
-    } catch (CalciteSemanticException e) {
-      // Bail out
-      return null;
-    }
+    RelDataType rowType = TypeConverter.getType(cluster, rr, null);
 
     // 2. Build RelOptAbstractTable
     List<String> fullyQualifiedTabName = new ArrayList<>();
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
@@ -34,7 +34,7 @@ public class CalciteSemanticException extends SemanticException {
     Having_clause_without_any_groupby, Invalid_column_reference, Invalid_decimal,
     Less_than_equal_greater_than, Others, Same_name_in_multiple_expressions,
     Schema_less_table, Select_alias_in_having_clause, Select_transform, Subquery,
-    Table_sample_clauses, UDTF, Union_type, Unique_join,
+    Table_sample_clauses, UDTF, Unique_join,
     HighPrecisionTimestamp // CALCITE-1690
   };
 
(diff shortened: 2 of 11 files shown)


zabetak:
Hey @soumyakanti3578 can you please review the PR? The test failure is a known issue (https://issues.apache.org/jira/browse/HIVE-26168) will deal with it later.

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
```diff
@@ -421,13 +420,7 @@ private static RelNode createMaterializedViewScan(HiveConf conf, Table viewTable
...
 
     // 1.3 Build row type from field <type, name>
-    RelDataType rowType;
-    try {
-      rowType = TypeConverter.getType(cluster, rr, null);
-    } catch (CalciteSemanticException e) {
-      // Bail out
-      return null;
-    }
+    RelDataType rowType = TypeConverter.getType(cluster, rr, null);
```
aturoczy, line 423: Calcite exception does not need to handle anymore? I guess the internal exception will throw in a higher level, and does not need to handle the null. But I want to highlight, this could be behavior change if the caller handle null.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
```diff
@@ -34,7 +34,7 @@ public enum UnsupportedFeature {
     Having_clause_without_any_groupby, Invalid_column_reference, Invalid_decimal,
     Less_than_equal_greater_than, Others, Same_name_in_multiple_expressions,
     Schema_less_table, Select_alias_in_having_clause, Select_transform, Subquery,
-    Table_sample_clauses, UDTF, Union_type, Unique_join,
+    Table_sample_clauses, UDTF, Unique_join,
```
aturoczy, line 37: Clarification: The Union_type is not need anymore?

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
aturoczy: btw it is more cleaner. :)

asolimando (approved):
LGTM

ql/src/test/results/clientpositive/llap/uniontype_cbo.q.out
```diff
@@ -0,0 +1,171 @@
+PREHOOK: query: CREATE TABLE utable (cu UNIONTYPE<INTEGER, STRING>)
```
okumin, line 1: I double-checked the diff between the result of the almost latest master and this file would be expected.

```
16c16
< PREHOOK: query: EXPLAIN CBO
---
> PREHOOK: query: EXPLAIN CBO 
23c23
< POSTHOOK: query: EXPLAIN CBO
---
> POSTHOOK: query: EXPLAIN CBO 
29a30,33
> CBO PLAN:
> HiveTableFunctionScan(invocation=[inline(ARRAY(ROW(create_union(0, 10, _UTF-16LE'ten':VARCHAR(2147483647) CHARACTER SET "UTF-16LE")), ROW(create_union(1, 10, _UTF-16LE'ten':VARCHAR(2147483647) CHARACTER SET "UTF-16LE"))))], rowType=[RecordType(RecordType(INTEGER $tag_0, VARCHAR(2147483647) $tag_1) col1)])
>   HiveTableScan(table=[[_dummy_database, _dummy_table]], table:alias=[_dummy_table])
> 
52a57,60
> CBO PLAN:
> HiveProject(cu=[$0])
>   HiveTableScan(table=[[default, utable]], table:alias=[utable])
> 
72a81,84
> CBO PLAN:
> HiveProject(_o__c0=[extract_union($0)])
>   HiveTableScan(table=[[default, utable]], table:alias=[utable])
> 
113a126,129
> CBO PLAN:
> HiveProject(u_age_name=[create_union(MOD($0, 2), $2, $1)])
>   HiveTableScan(table=[[default, author]], table:alias=[author])
> 
141a158,161
> CBO PLAN:
> HiveProject(u_age_name=[$0])
>   HiveTableScan(table=[[default, uauthor]], table:alias=[uauthor])
> 
```

okumin (approved):
LGTM!

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
zabetak: `CalciteSemanticException` is a checked exception so we were forced to catch and handle it here. However, after the changes in this PR it can't be thrown anymore from `TypeConverter.getType` methods because we are able to handle all type conversions.

The caller will never get back null as a result so it doesn't need to handle it.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CalciteSemanticException.java
zabetak: This is an internal enumeration that keeps track of `UnsupportedFeature`s in CBO. Since now UNIONTYPE is supported in CBO the enum entry is not necessary.

PR #5498, opened 2024-10-09
HIVE-28581: Support Partition Prunning stats optimizations for Iceberg tables

deniskuzZ:

Adds support for Iceberg partition prune stats optimizations. 

1. Adds functionality to compute, store and fetch basic partition stats per snapshot;
2. Enhances query plan with basic stats, aggregated for pruned partitions;
3. Enables PartitionConditionRemover (PCR) optimization;
4. Improves StatsOptimizer, allows to answers queries like count(*)  with partition predicate directly from stats;

Performance

No

No

mvn test -Dtest=TestIcebergCliDriver -Dqfile=iceberg_stats_with_ppr.q -Drat.skip=true


ql/src/java/org/apache/hadoop/hive/ql/metadata/DummyPartition.java
```diff
@@ -51,11 +48,15 @@ public DummyPartition(Table tbl, String name) {
     this.name = name;
   }  
 
-  public DummyPartition(Table tbl, String name,
-      Map<String, String> partSpec) {
-    setTable(tbl);
+  public DummyPartition(Table tbl, String name, Map<String, String> partSpec) throws HiveException {
```
deniskuzZ, line 51: Can we reuse this object or better create another abstraction, like [Virtual/Hidden]Partition?
cc @kasakrisz

kasakrisz:
@deniskuzZ 
Could you please add the jira Id.
I haven't found the file `iceberg_partition_stats_with_ppr.q` in the patch

deniskuzZ:
> Add support for Iceberg partition prune stats optimization



> @deniskuzZ Could you please add the jira Id. I haven't found the file `iceberg_partition_stats_with_ppr.q` in the patch

updated

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -445,9 +444,9 @@ public static List<DeleteFile> getDeleteFiles(Table table, String partitionPath)
         t -> ((PositionDeletesScanTask) t).file()));
   }
 
-  public static Expression generateExpressionFromPartitionSpec(Table table, Map<String, String> partitionSpec)
-      throws SemanticException {
-    Map<String, PartitionField> partitionFieldMap = getPartitionFields(table).stream()
+  public static Expression generateExpressionFromPartitionSpec(Table table, Map<String, String> partitionSpec,
+      boolean latestSpecOnly) throws SemanticException {
+    Map<String, PartitionField> partitionFieldMap = getPartitionFields(table, latestSpecOnly).stream()
```
kasakrisz, line 449: The method `generateExpressionFromPartitionSpec` is called from two places and the actual parameter value `latestSpecOnly` is always `true`. How about calling
```
icebergTable.spec().fields()
```
instead of
```
getPartitionFields(table, latestSpecOnly)
```

ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnInfo.java
```diff
@@ -166,9 +170,13 @@ public String getTabAlias() {
   }
 
   public boolean getIsVirtualCol() {
-    return isVirtualCol;
+    return isVirtualCol || isPartitionCol;
```
kasakrisz, line 173: Please don't mix virtual columns and partition columns. These are very different things. `getIsVirtualCol()` should depend on `isVirtualCol` only.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -129,19 +129,19 @@ public static Table getTable(Configuration configuration, org.apache.hadoop.hive
...
+
     if (metaTable != null) {
+      props.computeIfAbsent(InputFormatConfig.CATALOG_NAME, properties::get);
       // HiveCatalog, HadoopCatalog uses NAME to identify the metadata table
-      properties.setProperty(Catalogs.NAME, tableName + "." + metaTable);
+      props.computeIfAbsent(Catalogs.NAME, k -> properties.get(k) + "." + metaTable);
       // HadoopTable uses LOCATION to identify the metadata table
-      properties.setProperty(Catalogs.LOCATION, location + "#" + metaTable);
+      props.computeIfAbsent(Catalogs.LOCATION, k -> properties.get(k) + "#" + metaTable);
     }
```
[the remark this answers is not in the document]
deniskuzZ: WDYM? new code has the same 1 `if (metaTable != null)`. 
Note, the original method has side effects: it modifies `tableDesc.properties[name]` by appending `metaTable` to it, on a 2nd invocation name would be corrupted (i.e. `default.ice01.partitions.partitions`)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -471,6 +483,22 @@ public Map<String, String> getBasicStatistics(Partish partish) {
...
+
+            if (summary.containsKey(key)) {
+              StringTokenizer tokenizer = new StringTokenizer(summary.get(key), ",");
+              while (tokenizer.hasMoreTokens()) {
+                String token = tokenizer.nextToken();
+                String[] keyValue = token.split("=");
+                map.put(keyValue[0], keyValue[1]);
+              }
+            }
+            summary = map;
```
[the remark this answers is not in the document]
deniskuzZ: noted, I'll refactor

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
deniskuzZ: it's called from `IcebergTableUtil.getPartitionNames(Table icebergTable, Map<String, String> partitionSpec, boolean latestSpecOnly)`, that is called from `.HiveIcebergStorageHandlergetPartitions(org.apache.hadoop.hive.ql.metadata.Table table,
      Map<String, String> partitionSpec, boolean latestSpecOnly)`

deniskuzZ: or are you saying that we could refactor it to
````
  public static List<PartitionField> getPartitionFields(Table table, boolean latestSpecOnly) {
    return latestSpecOnly ? table.spec().fields() :
      table.specs().values().stream()
        .flatMap(spec -> spec.fields().stream()).distinct()
  }
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnInfo.java (hunk 3 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnInfo.java:173]: ATM all partitionCol were virtual ones, I've added isPartitionCol to include then into Scan

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/DummyPartition.java (hunk 1 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/DummyPartition.java]: we use overriden `getValues` and `getSpec`. And yes, they represent real partitions (not in HMS)

File: ql/src/test/results/clientpositive/llap/vector_bucket.q.out [hunk 6]
```diff
@@ -61,10 +61,11 @@ STAGE PLANS:
             Execution mode: llap
             LLAP IO: no inputs
             Map Vectorization:
-                enabled: false
+                enabled: true
                 enabledConditionsMet: hive.vectorized.use.vectorized.input.format IS true
-                enabledConditionsNotMet: Could not enable vectorization due to partition column names size 1 is greater than the number of table column names size 0 IS false
                 inputFileFormats: org.apache.hadoop.hive.ql.io.NullRowsInputFormat
+                notVectorizedReason: UDTF Operator (UDTF) not supported
+                vectorized: false
```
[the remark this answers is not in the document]
deniskuzZ: there was a bug in a Vectorizer, I noticed it while troubleshooting. I can move it into a different PR

ql/src/java/org/apache/hadoop/hive/ql/metadata/DummyPartition.java
deniskuzZ: `DummyPartition` is used to represent non-HMS partitions. Before, it had just two properties: `table` and `name`. 
But I agree the name is not self explanatory

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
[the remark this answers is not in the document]
kasakrisz: These are the `if`s in new code:
* `(metaTable != null) ? new Properties() : properties`
* `if (metaTable != null) { ... }`
* 3 `computeIfAbsent` calls

When `metaTable` present a new empty `Properties` instance is created and `computeIfAbsent` is always absent.
Ternary operator is one line but it still an if-else with the same condition like the follow-up if statement.

I agree that the side affect you mentioned should be fixed. It can be achieved by only one if statement:
```
    Properties props;
    if (metaTable != null) {
      props = new Properties();
      props.put(InputFormatConfig.CATALOG_NAME, properties.get(InputFormatConfig.CATALOG_NAME));
      props.put(Catalogs.NAME, properties.get(Catalogs.NAME) + "." + metaTable);
      props.put(Catalogs.LOCATION, properties.get(Catalogs.LOCATION) + "#" + metaTable);
    } else {
      props = properties;
    }
```
The else block can also be removed:
```
   Properties props = new Properties(properties); // use input properties as default
   if (metaTable != null) {
...
   }
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
kasakrisz: > or are you saying that we could refactor it to
> 
> ```
>   public static List<PartitionField> getPartitionFields(Table table, boolean latestSpecOnly) {
>     return latestSpecOnly ? table.spec().fields() :
>       table.specs().values().stream()
>         .flatMap(spec -> spec.fields().stream()).distinct()
>         .collect(Collectors.toList());
>   }
> ```

yes.

Actually at the top of the call stack, from we call this method, we already know whether we need only the current spec or all. :)

ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnInfo.java
kasakrisz: I still has the impression that this change is not safe. Example:
here we create a new instance of `ColumnInfo` based on an existing one. If the source `ColumnInfo` is a partition column the field `isPartitionCol` is still `false` in the result which is not valid.

https://github.com/apache/hive/blob/b2a5933c0f396ade25273ac75cbf35e205969b21/ql/src/java/org/apache/hadoop/hive/ql/optimizer/ColumnPrunerProcFactory.java#L196-L197

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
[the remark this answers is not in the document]
deniskuzZ: hashtable's `get` has a O(1) complexity, so that could be ignored, but I don't mind to refactor it as you proposed

ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnInfo.java
deniskuzZ: true, but how else should I call this method? `isRequiredCol` ??

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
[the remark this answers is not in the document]
deniskuzZ: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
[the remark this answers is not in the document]
deniskuzZ: done

ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnInfo.java
deniskuzZ: renamed to `isHiddenPartitionCol `

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
deniskuzZ: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -425,6 +426,16 @@ public boolean canProvideBasicStatistics() {
     return true;
   }
 
+  @Override
+  public boolean canProvidePartitionStatistics(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
+    Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
+    if (table.currentSnapshot() != null) {
```
zhangbutao, line 432: We can also check branch/tag here.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/StatsOptimizer.java
```diff
@@ -932,7 +932,7 @@ private Collection<List<ColumnStatisticsObj>> verifyAndGetPartColumnStats(
     private Long getRowCnt(
         ParseContext pCtx, TableScanOperator tsOp, Table tbl) throws HiveException {
       Long rowCnt = 0L;
-      if (tbl.isPartitioned()) {
+      if (tbl.isPartitioned() && StatsUtils.checkCanProvidePartitionStats(tbl)) {
         for (Partition part : pctx.getPrunedPartitions(
             tsOp.getConf().getAlias(), tsOp).getPartitions()) {
           if (!StatsUtils.areBasicStatsUptoDateForQueryAnswering(part.getTable(), part.getParameters())) {
```
zhangbutao, line 938: `StatsUtils::areBasicStatsUptoDateForQueryAnswering` is not applicable to Iceberg table, and it will check table param `COLUMN_STATS_ACCURATE` and then determine to get stats or not. But we always get partition stats from iceberg metadata file, so `COLUMN_STATS_ACCURATE` should be always `true`.

The reason the iceberg qtest for table&partition's stats looks good is because we already set `COLUMN_STATS_ACCURATE` to true in hive-site.xml. But in fact, i think no users will care this param. So i think if we want to use iceberg partition stats, we should consider to remove this param.
https://github.com/apache/hive/blob/48a67a4f2cc7a65bf9aac4a1ed518958c5b00027/data/conf/hive-site.xml#L334-L339

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: @zhangbutao thanks for checking. I am again stepping on https://github.com/apache/hive/pull/5215
Could you please address the review comments so we can merge it and reuse here?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/StatsOptimizer.java
deniskuzZ: fixed

ql/src/test/results/clientpositive/llap/vector_bucket.q.out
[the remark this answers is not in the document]
deniskuzZ: created https://github.com/apache/hive/pull/5519

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution_w_id_spec_w_filter.q
```diff
@@ -65,21 +65,6 @@ delete from ice_orc where last_name in ('ln1', 'ln9');
 delete from ice_orc where last_name in ('ln3', 'ln11');
 delete from ice_orc where last_name in ('ln5', 'ln13');
 
-alter table ice_orc set partition spec(team_id);
```
zratkai, line 68: Why did you delete this ?

iceberg/patched-iceberg-core/src/main/java/org/apache/iceberg/SnapshotProducer.java
```diff
@@ -0,0 +1,697 @@
...
+import static org.apache.iceberg.TableProperties.COMMIT_NUM_RETRIES_DEFAULT;
+import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS;
+import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT;
+import static org.apache.iceberg.TableProperties.MANIFEST_TARGET_SIZE_BYTES;
+import static org.apache.iceberg.TableProperties.MANIFEST_TARGET_SIZE_BYTES_DEFAULT;
+import static org.apache.iceberg.TableProperties.SNAPSHOT_ID_INHERITANCE_ENABLED;
+import static org.apache.iceberg.TableProperties.SNAPSHOT_ID_INHERITANCE_ENABLED_DEFAULT;
+
+@SuppressWarnings("UnnecessaryAnonymousClass")
+abstract class SnapshotProducer<ThisT> implements SnapshotUpdate<ThisT> {
```
zhangbutao, line 78: It would be good to add some comments to tell others what new things were added in this class.

Or try to push the change to upstream Apache Iceberg.

Then we can remove or change the patched class safely in the next iceberg upgrade.

iceberg/iceberg-handler/src/test/results/positive/iceberg_insert_overwrite_partition.q.out
```diff
@@ -487,32 +488,32 @@ STAGE PLANS:
...
                       sort order: ++
                       Map-reduce partition columns: _col2 (type: string), _col3 (type: string)
-                      Statistics: Num rows: 4 Data size: 1084 Basic stats: COMPLETE Column stats: COMPLETE
+                      Statistics: Num rows: 4 Data size: 2224 Basic stats: COMPLETE Column stats: NONE
                       value expressions: _col0 (type: string), _col1 (type: int), _col2 (type: string), _col3 (type: string)
                     Select Operator
                       expressions: _col0 (type: string), _col1 (type: int), _col2 (type: string), _col3 (type: string)
                       outputColumnNames: name, age, country, state
-                      Statistics: Num rows: 4 Data size: 1084 Basic stats: COMPLETE Column stats: COMPLETE
+                      Statistics: Num rows: 4 Data size: 2224 Basic stats: COMPLETE Column stats: NONE
```
abstractdog, line 506: many occurrences of COMPLETE -> NONE, is it a statistics regression?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4146,12 +4146,12 @@ public List<Partition> getPartitions(Table tbl) throws HiveException {
     PerfLogger perfLogger = SessionState.getPerfLogger();
     perfLogger.perfLogBegin(CLASS_NAME, PerfLogger.HIVE_GET_PARTITIONS);
     try {
-      if (tbl.getStorageHandler() != null && tbl.getStorageHandler().alwaysUnpartitioned()) {
-        return tbl.getStorageHandler().getPartitions(tbl, Collections.EMPTY_MAP);
+      if (tbl.alwaysUnpartitioned()) {
+        return tbl.getStorageHandler().getPartitions(tbl);
       } else {
-        int batchSize= MetastoreConf.getIntVar(Hive.get().getConf(), MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX);
+        int batchSize= MetastoreConf.getIntVar(conf, MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX);
```
abstractdog, line 4152: nit: space (as you're already touching this line)
```
int batchSize =  MetastoreConf...
```

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4146,12 +4146,12 @@ public List<Partition> getPartitions(Table tbl) throws HiveException {
...
         return new ArrayList<>(getAllPartitionsInBatches(tbl, batchSize, DEFAULT_BATCH_DECAYING_FACTOR, MetastoreConf
-                .getIntVar(Hive.get().getConf(), MetastoreConf.ConfVars.GETPARTITIONS_BATCH_MAX_RETRIES),
+                .getIntVar(conf, MetastoreConf.ConfVars.GETPARTITIONS_BATCH_MAX_RETRIES),
```
abstractdog, line 4154: there are 8 occurrences of this Hive.get().getConf() non-sense, using conf makes sense, you fixed 6 of them as far as I can see, can you please take care of the rest 2 in getPartitionsWithSpecs

iceberg/iceberg-handler/src/test/results/positive/iceberg_insert_overwrite_partition.q.out
deniskuzZ: no, stats was completely wrong

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/desc/DescTableOperation.java
```diff
@@ -190,16 +192,16 @@ private void getColumnsNoColumnPath(Table table, Partition partition, List<Field
...
-      throws SemanticException, HiveException, MetaException {
+      throws HiveException, MetaException {
     // when column name is specified in describe table DDL, colPath will be db_name.table_name.column_name
     String colName = desc.getColumnPath().split("\\.")[2];
     List<String> colNames = Lists.newArrayList(colName.toLowerCase());
 
     TableName tableName = HiveTableName.of(desc.getDbTableName());
     if (null == part) {
-      if (table.isPartitioned()) {
+      if (table.isPartitioned() && !table.alwaysUnpartitioned()) {
```
abstractdog, line 202: basically, I can see 2 different condition combinations:
```
table.isPartitioned() && !table.alwaysUnpartitioned()
```
and
```
!table.isPartitioned() || table.alwaysUnpartitioned()
```

first looks like something partitioned, second looks like something non-partitioned, could this be merged into a single call? does it make sense to handle them separately?

this is kind of inline with my javadoc request for alwaysUnpartitioned

iceberg/iceberg-handler/src/test/results/positive/iceberg_insert_overwrite_partition.q.out
abstractdog: okay, I'm assuming those can still be collected with ANALYZE TABLE ... FOR COLUMNS if we want to have them (it's just not the scope of this ticket)

ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java
```diff
@@ -641,7 +642,7 @@ public boolean isPartitionKey(String colName) {
   // TODO merge this with getBucketCols function
   public String getBucketingDimensionId() {
     List<String> bcols = tTable.getSd().getBucketCols();
-    if (bcols == null || bcols.size() == 0) {
+    if (bcols == null || bcols.isEmpty()) {
```
abstractdog, line 645:
```
CollectionUtils.isEmpty(bcols)
```

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/desc/DescTableOperation.java
deniskuzZ: table.alwaysUnpartitioned() means that table supports native partitioning, we don't store anything in HMS
````
  /**
   * Check if partition columns should be removed and added to the list of regular columns in HMS.
   * This can be useful for non-native tables where the table format/layout differs from the standard Hive table layout,
   * e.g. Iceberg tables. For these table formats, the partition column values are stored in the data files along with
   * regular column values, therefore the object inspectors should include the partition columns as well.
   * Any partitioning scheme provided via the standard HiveQL syntax will be honored but stored in someplace
   * other than HMS, depending on the storage handler implementation.
   *
   * @return whether table should always be unpartitioned from the perspective of HMS
   */
  default boolean alwaysUnpartitioned() {
    return false;
  }
````
not a great name, maybe better `isMetadataEnhanced` ?
````
table.isPartitioned() && !table.isMetadataEnhanced()
````
would mean that we need to fetch information from somewhere like HMS

Comment (abstractdog) [reply ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/desc/DescTableOperation.java:202]: oh, I see, this is not even a new method...
at first sight, I like isMetadataEnhanced but it's too generic and hides the original meaning

really tricky, thinking about this
**always unpartitioned** is just one of the consequences of the root cause that the table format - which is handled by the storage handler - can handle partitions natively, so hive should treat it as unpartitioned (and doesn't have to add its mindfuck logic for partition columns), so for me it seems more like:
```
table.hasBuiltInPartitionSupport()
```

which is more in line with most of the usages, like:
```
    if (tbl.hasBuiltInPartitionSupport()) {
      return tbl.getStorageHandler().getPartition(...
      ...
```
that can be read as: if the table has built-in partition support, then we must acquire partitions from it instead of HMS

just idea, this is your choice

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java (hunk 15 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java]: fixed

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java (hunk 13 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java:4154]: done

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java:4154]: done + renamed the `alwaysUnpartitioned` to `isMetadataEnhanced`

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java (hunk 12 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java]: fixed

**Comment by abstractdog** [comment]
in general, I cannot see any test coverage of this new feature, how does the added iceberg_stats_with_ppr.q.out show that it kicks in?

**Comment by deniskuzZ** [comment]
> in general, I cannot see any test coverage of this new feature, how does the added iceberg_stats_with_ppr.q.out show that it kicks in?
1. table has 5 rows, predicate returns 3 (before would be 5), see stats:
https://github.com/apache/hive/pull/5498/files#diff-d6f6d58b1308192b6b3c9fcd75e2f5047e3817ad24900e2ecaca684569956d6eR56-R57
2. count(*) explain shows no tasks, simple fetch from stats (before Map task was launched) 
https://github.com/apache/hive/pull/5498/files#diff-d6f6d58b1308192b6b3c9fcd75e2f5047e3817ad24900e2ecaca684569956d6eR78-R97

File: iceberg/iceberg-handler/src/test/results/positive/iceberg_insert_overwrite_partition.q.out (hunk 11 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/test/results/positive/iceberg_insert_overwrite_partition.q.out:506]: yes, there is a q test for that - `puffin_col_stats.q`

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/data/PartitionStatsHandler.java [hunk 16]
```diff
@@ -0,0 +1,336 @@
...
+        GenericFileWriterFactory.builderFor(table)
+        .dataSchema(dataSchema)
+        .dataFileFormat(fileFormat(outputFile.location()))
+        .build();
+    DataWriter<Record> writer =
+        factory.newDataWriter(
+        EncryptedFiles.encryptedOutput(outputFile, EncryptionKeyMetadata.EMPTY),
+        PartitionSpec.unpartitioned(),
+        null);
+    try (Closeable toClose = writer) {
```
zratkai, line 176: I think this approach can lead to errors. Think about writer is closed after this try, but later it can be still used in this method. I see now it is not used, but later it can be changed, refactored etc. and can lead to potential errors.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/Catalogs.java
```diff
@@ -183,7 +184,11 @@ public static boolean dropTable(Configuration conf, Properties props) {
    * @return true if the Catalog is HiveCatalog
    */
   public static boolean hiveCatalog(Configuration conf, Properties props) {
-    String catalogName = props.getProperty(InputFormatConfig.CATALOG_NAME);
+    return hiveCatalog(conf, props::getProperty);
+  }
+
+  public static boolean hiveCatalog(Configuration conf, UnaryOperator<String> props) {
```
zratkai, line 190: Why is this necessary?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -165,6 +165,41 @@ static Table getTable(Configuration configuration, Properties properties) {
     return getTable(configuration, properties, false);
   }
 
+  static Snapshot getTableSnapshot(Table table, org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
```
zratkai, line 168: There is already a method in this class with the same name, same input types, but reverse order:
public static Snapshot getTableSnapshot(org.apache.hadoop.hive.ql.metadata.Table hmsTable, Table table)

It is confusing, and hard to understand what is the difference. Please use the same order of parameters, as the old one, and add other name which expresses the difference.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -179,6 +214,12 @@ static Optional<Path> getColStatsPath(Table table, long snapshotId) {
       .findAny();
   }
 
+  static PartitionStatisticsFile getPartitionStatsFile(Table table, long snapshotId) {
```
zratkai, line 217: getPartitionStatsFileBySnapshotId?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/data/PartitionStatsHandler.java
deniskuzZ: same appoach is used everywhere in iceberg repo.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/Catalogs.java
deniskuzZ: because i need that method signature

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
deniskuzZ: `table` is the main argument and `hmsTable` is supplementary config provider

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/data/PartitionStatsHandler.java
```diff
@@ -0,0 +1,336 @@
...
+    Schema schema = schema(partitionType);
+
+    Collection<PartitionStats> stats = PartitionStatsUtil.computeStats(table, currentSnapshot);
+    List<PartitionStats> sortedStats = PartitionStatsUtil.sortStats(stats, partitionType);
+    Iterator<PartitionStatsRecord> convertedRecords = statsToRecords(sortedStats, schema);
+    return writePartitionStatsFile(table, currentSnapshot.snapshotId(), schema, convertedRecords);
+  }
+
+  @VisibleForTesting
+  static PartitionStatisticsFile writePartitionStatsFile(
```
difin, line 163: This method is not called from unit tests, @VisibleForTesting seems to be not needed.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/data/PartitionStatsHandler.java
```diff
@@ -0,0 +1,336 @@
...
+    stats.set(Column.LAST_UPDATED_AT.id(), record.get(Column.LAST_UPDATED_AT.id(), Long.class));
+    stats.set(
+        Column.LAST_UPDATED_SNAPSHOT_ID.id(),
+        record.get(Column.LAST_UPDATED_SNAPSHOT_ID.id(), Long.class));
+
+    return PartitionStatsRecord.create(record.struct(), stats);
+  }
+
+  @VisibleForTesting
+  static Iterator<PartitionStatsRecord> statsToRecords(
```
difin, line 277: This method is not called from unit tests, @VisibleForTesting seems to be not needed.

iceberg/iceberg-handler/src/test/results/positive/iceberg_insert_overwrite_partition.q.out
deniskuzZ: @abstractdog, sorry my bad. 
Now, since we no longer consider table as unpartitioned, we need to calculate, store and fetch col stats per partition (same as for basic stats implemented here). Table col stats can't be used anymore.
I'll create a follow-up ticket for that since it's a big change.

iceberg/patched-iceberg-core/src/main/java/org/apache/iceberg/data/PartitionStatsRecord.java
```diff
@@ -0,0 +1,174 @@
...
+    return partitionStats.get(pos, Object.class);
+  }
+
+  @Override
+  public <T> T get(int pos, Class<T> javaClass) {
+    Object value = get(pos);
+    if (value == null || javaClass.isInstance(value)) {
+      return javaClass.cast(value);
+    } else {
+      throw new IllegalStateException("Not an instance of " + javaClass.getName() + ": " + value);
```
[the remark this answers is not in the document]
deniskuzZ: please don't review patched stuff or classes with `TODO: remove class`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/data/PartitionStatsHandler.java
deniskuzZ: would be removed once we upgrade iceberg dep

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/data/PartitionStatsHandler.java
deniskuzZ: removed

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/data/PartitionStatsHandler.java
deniskuzZ: removed

iceberg/iceberg-handler/src/test/results/positive/iceberg_insert_overwrite_partition.q.out
deniskuzZ: added fallback to table column stats in https://github.com/apache/hive/pull/5498/commits/8f24713a3d2e303dac5fc8df1c7929e26e7dc561

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/Catalogs.java
deniskuzZ: we can drop it if the tests pass without this check: https://github.com/apache/hive/pull/5498/commits/f71c124d85efe91431f93f9d2e79e7f46f08338f

deniskuzZ: few failed, I've fixed them and removed this method

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution_w_id_spec_w_filter.q
deniskuzZ: only current spec partition columns should be allowed in the compaction command predicate. 
alteration of `PartitionPruner#onlyContainsPartnCols` broke the partition pruning logic, disabling all partition optimizations

difin:
There are several Sonar issues in the modified classes.
Other than that LGTM +1.

difin (approved):

iceberg/iceberg-handler/src/test/results/positive/llap/iceberg_bucket_map_join_4.q.out
```diff
@@ -242,23 +242,23 @@ Stage-0
...
+                  Select Operator [SEL_32] (rows=3918 width=89)
                     Output:["_col0"]
-                    Filter Operator [FIL_31] (rows=1476 width=89)
+                    Filter Operator [FIL_31] (rows=3918 width=89)
                       predicate:(part is not null and key is not null)
-                      TableScan [TS_3] (rows=1476 width=89)
-                        default@srcbucket_mapjoin_part_2_n6,b,Tbl:COMPLETE,Col:COMPLETE,Output:["key","part"]
-              <-Select Operator [SEL_35] (rows=2952 width=4)
+                      TableScan [TS_3] (rows=3918 width=89)
+                        default@srcbucket_mapjoin_part_2_n6,b,Tbl:PARTIAL,Col:COMPLETE,Output:["key","part"]
```
deniskuzZ, line 255: hi @okumin, after adding support in iceberg for basic partition stats, I've noticed that a few TableScans have PARTIAL stats. 
The reason is that the partition name seems to include the bucketing number
`srcbucket_mapjoin_part_1_n2(part=2/key_bucket_4=0)`, however, stats record doesn't have it `part=2/key_bucket=0`
Was it necessary to have the bucketing number in the name or it can be removed?

ayushtkn (approved):
LGTM

PR #5499, opened 2024-10-10
HIVE-28569: ReExecuteLostAMQueryPlugin doesn't kick in if the message contains line break and diagnostics

abstractdog:
Change the pattern matcher to consume a pattern with line breaks.

Because the plugin didn't kick in if the message contains diagnostics which is separated from the original trigger message with line break

No.

No.

Unit test added.


deniskuzZ (approved):
+1

PR #4271, opened 2023-04-26
HIVE-27299: Upgrade guava version to 31.1-jre to fix CVE

Aggarwal-Raghav:
### What changes were proposed in this pull request?
Upgrade guava to fix the CVE's

### Why are the changes needed?
The guava version in Hive master branch is 22.0 which has 2 Direct CVE:
[CVE-2020-8908](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-8908)
[CVE-2018-10237](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-10237)
Component like Tez 0.10.2 (used in hive) has also moved to 31.1-jre of guava version.

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
By building hive on local machine

PR #5500, opened 2024-10-10
HIVE-28191: Upgrade Hadoop Version to 3.4.1

zhangbutao:

Upgrade Hadoop version from 3.3.6 to 3.4.1

No

Yes. See the file: 
[hadoop_upgrade_dep.txt](https://github.com/user-attachments/files/18392110/hadoop_upgrade_dep.txt)

Existing Tests.

### DIFF
--- a/itests/hcatalog-unit/pom.xml
+++ b/itests/hcatalog-unit/pom.xml
@@ -357,23 +357,4 @@
       </plugin>
     </plugins>
   </build>
-  <profiles>
-    <profile>
-      <!-- See HIVE-24473 -->
-      <id>gahbase</id>
-      <activation>
-        <property>
-          <name>hbase.version</name>
-        </property>
-      </activation>
-      <dependencies>
-        <dependency>
-          <groupId>org.apache.hbase</groupId>
-          <artifactId>hbase-zookeeper</artifactId>
-          <classifier>tests</classifier>
-          <scope>test</scope>
-        </dependency>
-      </dependencies>
-    </profile>
-  </profiles>
 </project>
--- a/itests/util/src/main/java/org/apache/hadoop/hive/hbase/HBaseTestSetup.java
+++ b/itests/util/src/main/java/org/apache/hadoop/hive/hbase/HBaseTestSetup.java
@@ -106,7 +106,7 @@ public class HBaseTestSetup extends QTestSetup {
         "org.apache.hadoop.hbase.shaded.");
 
     Configuration hbaseConf = HBaseConfiguration.create(conf);
-    // A workaround for HBASE-28908
+    // A workaround for HBASE-28908. Should be removed once HBase 2.7.0 is released & HIVE-28740 is fixed.
     hbaseConf.set("hbase.wal.provider", "filesystem");
     util = new HBaseTestingUtility(hbaseConf);
 
--- a/itests/util/src/main/java/org/apache/hadoop/hive/hbase/HBaseTestSetup.java
+++ b/itests/util/src/main/java/org/apache/hadoop/hive/hbase/HBaseTestSetup.java
@@ -106,6 +106,8 @@ public class HBaseTestSetup extends QTestSetup {
         "org.apache.hadoop.hbase.shaded.");
 
     Configuration hbaseConf = HBaseConfiguration.create(conf);
+    // A workaround for HBASE-28908
+    hbaseConf.set("hbase.wal.provider", "filesystem");
     util = new HBaseTestingUtility(hbaseConf);
 
     util.startMiniDFSCluster(1);
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
     <hadoop.version>3.4.0</hadoop.version>
     <hadoop.bin.path>${basedir}/${hive.path.to.root}/testutils/hadoop</hadoop.bin.path>
     <hamcrest.version>1.3</hamcrest.version>
-    <hbase.version>2.5.6-hadoop3</hbase.version>
+    <hbase.version>2.5.10-hadoop3</hbase.version>
     <hppc.version>0.7.2</hppc.version>
     <!-- required for logging test to avoid including hbase which pulls disruptor transitively -->
     <disruptor.version>3.3.7</disruptor.version>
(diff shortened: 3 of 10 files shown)


slfan1989:
@zhangbutao LGTM +1.

Aggarwal-Raghav:
@zhangbutao, thanks for driving this forward.

_For tez project:_
1. **We need to exclude logback jar from hadoop transitive dependency in tez project or move to hadoop 3.4.1**, Otherwise it can cause classloading issues. IIRC, I had faced issue because logback jar was getting picked first and hive-log4j2.properties not getting honoured. If possible please go through the following:
a. https://github.com/apache/hadoop/pull/6582#issue-2151551682
b. [HADOOP-19153](https://issues.apache.org/jira/browse/HADOOP-19153):  hadoop-common exports logback as a dependency (**This fix is not in hadoop 3.4.0**)
2. zookeeper version, I would prefer to keep it in sync:
a. hive => 3.8.4
b. hadoop3.4.0 => 3.8.3
c. hadoop3.4.1 => 3.8.4 

_For hive:_

- Please fill out the PR template and attach the dependecy tree.
- I think we can upgrade guava version also and keep it in sync with hadoop, there was some discussion to keep it in sync with the version used in hadoop. Please check #4271

zhangbutao:
@Aggarwal-Raghav Thanks for your insightful thought!  Will check this later.

ayushtkn:
should consider 3.4.1 now :-)

devaspatikrishnatri:
@zhangbutao Is this PR going to go in , does this affect the upgrade of guava in hive?

zhangbutao:
> @zhangbutao Is this PR going to go in , does this affect the upgrade of guava in hive?

@devaspatikrishnatri I will check this next week. ：）

Aggarwal-Raghav:
@zhangbutao, please check by updating the guava version to 27.0-jre (used in Hadoop 3.4.1) instead of 32.0.1-jre.

zhangbutao:
> 27.0-jre

@Aggarwal-Raghav Sure, Let me try.

zhangbutao:
Interestingly, faild tests show that guava-27.0-jre would change some qtests(sql explain).

zhangbutao:
I haven't figured out why the some qtests changed after upgrading guava. But these changes https://github.com/apache/hive/pull/5500/commits/78357d29a269049531d218395ab7af3b3700ab2c are just the names of the columns in the explain, so I think the guava upgrade is acceptable.

ayushtkn:
I am not sure we should chase Guava upgrade as part of Hadoop upgrade. We can track that separately I believe.

btw. Hadoop doesn't use guava version specified in its POM, that is kept only for its transitive dependency. It uses the Guava coming from hadoop-thirdparty: HADOOP-17288 and that is 30+ as of today, should be 30+ for 3.4.1 as well If I am not mistaken

https://github.com/apache/hadoop-thirdparty/blob/trunk/pom.xml#L101

Aggarwal-Raghav:
> "Hadoop doesn't use guava version specified in its POM, that is kept only for its transitive dependency."

Oh, I was not aware of this. Then maybe we can track it in separate ticket.

Just info: In our codebase, we have guava version 32.0.1-jre in tez (0.10.3), hadoop(3.3.6) and hive(4.0.0) and I didn't observed any UT failures there. Something to investigate on my end.

zhangbutao:
guava 32.0.1-jre would cause lots of qtests failure, including some class not found exeception. 

guava 27.0-jre would casue some minor explain qtests changes.   https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5500/9/pipeline/

So, Maybe we can upgrade to this 27.0 version first and then consider upgrading to 32.0 version.

In short, it makes more sense to study the guava version carefully in subsequent ticket before upgrading.

I will revert the guava upgrade in this PR. @Aggarwal-Raghav @ayushtkn

itests/util/src/main/java/org/apache/hadoop/hive/hbase/HBaseTestSetup.java
```diff
@@ -106,6 +106,8 @@ private void setUpFixtures(HiveConf conf) throws Exception {
         "org.apache.hadoop.hbase.shaded.");
 
     Configuration hbaseConf = HBaseConfiguration.create(conf);
+    // A workaround for HBASE-28908
+    hbaseConf.set("hbase.wal.provider", "filesystem");
```
ayushtkn, line 110: can you create a Hive ticket to upgrade HBase to 2.7.0. & link the mentioned ticket + this line to be removed + HBASE-28939

pom.xml
```diff
@@ -140,10 +140,10 @@
     <guava.version>22.0</guava.version>
     <groovy.version>2.4.21</groovy.version>
     <h2database.version>2.2.220</h2database.version>
-    <hadoop.version>3.3.6</hadoop.version>
+    <hadoop.version>3.4.1</hadoop.version>
     <hadoop.bin.path>${basedir}/${hive.path.to.root}/testutils/hadoop</hadoop.bin.path>
     <hamcrest.version>1.3</hamcrest.version>
-    <hbase.version>2.5.6-hadoop3</hbase.version>
+    <hbase.version>2.5.10-hadoop3</hbase.version>
```
ayushtkn, line 146: The HBase upgrade can be split into another ticket unless it can't go ahead without Hadoop upgade. If it is the case maybe then we can change the title to show that

itests/hcatalog-unit/pom.xml
```diff
@@ -357,23 +357,4 @@
       </plugin>
     </plugins>
   </build>
-  <profiles>
-    <profile>
-      <!-- See HIVE-24473 -->
```
ayushtkn, line 362: Why are we removing these?

ayushtkn (commented):
Minor comments rest looks good

itests/util/src/main/java/org/apache/hadoop/hive/hbase/HBaseTestSetup.java
zhangbutao: Created the ticket HIVE-28740.

itests/hcatalog-unit/pom.xml
zhangbutao: See https://issues.apache.org/jira/browse/HIVE-24473?focusedCommentId=17254875&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17254875

HIVE-24473 added a profile(-Dhbase.version) to make hive&hadoop3 can build with the ga hbase. But we can build with hbase without this profile now. No need it any more.

ayushtkn (approved):
LGTM

PR #5501, opened 2024-10-10
HIVE-28570: Iceberg: Aggregate queries on translated iceberg table fails post rename.

ayushtkn:

Avoid MetdataTransformer to modify the table location in case the translated table is an iceberg table

To make the aggregate query work post rename

Yes, count(*) on renamed iceberg tables work

No

UT


simhadri-g (approved):
LGTM +1

PR #2085, opened 2021-03-17
HIVE-24838. Reduce FS creation in Warehouse::getDnsPath for object stores

zeroflag:
Based on its javadoc comment getDnsPath is supposed to change the authority part (ip) of a path to a domain name. To do so it creates a new FileSystem object and gets the default path from it, and uses the authority from the the default path to replace the authority of the input.

There are multiple problems with this. Instantiating a new FileSystem is expensive. Replacing the authority on a blobstore path (s3a) is not needed since s3 uses bucket name instead of a hostname in the path.

Even in HDFS case the original function doesn't do what the original intention might was. The new file system is initialized with the input path so the default FS path is always same as the input path (unless a relative path is used as an input). The only useful thing this function does is transforming a relative path to an absolute path.

For example: 

```
input = /warehouse/tablespace/managed/hive
output = hdfs://namenode-address.site:8020/warehouse/tablespace/managed/hive
```

But this can be achieved by doing a simple config lookup (FileSystem.getDefaultUri(conf)) there is no need to create a new FS object every time.

PR #5504, opened 2024-10-12
HIVE-28575: Reduce hdfs filesystem rpc call

zhangbutao:

HIVE-24838 https://github.com/apache/hive/pull/2085 added the optimization to reduce BlobStorages fs call. I think it is also can be used to reduce HDFS filesystem call.

This can also optimize some issues like HIVE-28523(if partition location is outside from table location)

No

The existing tests

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
@@ -133,14 +133,8 @@ public class Warehouse {
 
 
   /**
-   * Hadoop File System reverse lookups paths with raw ip addresses The File
-   * System URI always contains the canonical DNS name of the Namenode.
-   * Subsequently, operations on paths with raw ip addresses cause an exception
-   * since they don't match the file system URI.
-   *
-   * This routine solves this problem by replacing the scheme and authority of a
-   * path with the scheme and authority of the FileSystem that it maps to.
-   *
+   * Get a path with canonical scheme and authority.
+   * <p>
    * Since creating a new file system object is expensive, this method
    * mimics getFileSystem() without creating an actual FileSystem object.
    * When the input path lacks a scheme or an authority this is added
@@ -151,40 +145,24 @@ public class Warehouse {
    * @return Path with canonical scheme and authority
    */
   public static Path getDnsPath(Path path, Configuration conf) throws MetaException {
-    if (isBlobStorageScheme(conf, path.toUri().getScheme())) {
-      String scheme = path.toUri().getScheme();
-      String authority = path.toUri().getAuthority();
-      URI defaultUri = FileSystem.getDefaultUri(conf);
-      if ((authority == null && scheme == null)
-              || StringUtils.equalsIgnoreCase(scheme, defaultUri.getScheme())) {
-        if (authority == null) {
-          authority = defaultUri.getAuthority();
-        }
-        if (scheme == null) {
-          scheme = defaultUri.getScheme();
-        }
-        String uriPath = path.toUri().getPath();
-        if (StringUtils.isEmpty(uriPath)) {
-          uriPath = "/";
-        }
-        return new Path(scheme, authority, uriPath);
+    String scheme = path.toUri().getScheme();
+    String authority = path.toUri().getAuthority();
+    URI defaultUri = FileSystem.getDefaultUri(conf);
+    if ((authority == null && scheme == null)
+        || StringUtils.equalsIgnoreCase(scheme, defaultUri.getScheme())) {
+      if (authority == null) {
+        authority = defaultUri.getAuthority();
       }
-      return path;
-    } else { // fallback: for other FS type make the FS instance
-      FileSystem fs = getFs(path, conf);
+      if (scheme == null) {
+        scheme = defaultUri.getScheme();
+      }
       String uriPath = path.toUri().getPath();
       if (StringUtils.isEmpty(uriPath)) {
         uriPath = "/";
       }
-      return (new Path(fs.getUri().getScheme(), fs.getUri().getAuthority(), uriPath));
+      return new Path(scheme, authority, uriPath);
     }
-  }
-
-  private static boolean isBlobStorageScheme(Configuration conf, String scheme) {
-    final String uriScheme = scheme == null ? FileSystem.getDefaultUri(conf).getScheme() : scheme;
-    return MetastoreConf.getStringCollection(conf, MetastoreConf.ConfVars.HIVE_BLOBSTORE_SUPPORTED_SCHEMES)
-            .stream()
-            .anyMatch(each -> each.equalsIgnoreCase(uriScheme));
+    return path;
   }
 
   public Path getDnsPath(Path path) throws MetaException {
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -1796,8 +1796,6 @@ public class MetastoreConf {
         "hive.metastore.custom.database.product.classname", "none",
           "Hook for external RDBMS. This class will be instantiated only when " +
           "metastore.use.custom.database.product is set to true."),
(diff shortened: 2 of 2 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
```diff
@@ -133,14 +133,8 @@ public FileSystem getFs(Path f) throws MetaException {
 
 
   /**
-   * Hadoop File System reverse lookups paths with raw ip addresses The File
-   * System URI always contains the canonical DNS name of the Namenode.
-   * Subsequently, operations on paths with raw ip addresses cause an exception
-   * since they don't match the file system URI.
```
zhangbutao, line 139: I realy don't understand what's meaning of this comment.
https://github.com/apache/hive/blob/master/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestWarehouseDnsPath.java The tests show that `getDnsPath` has nothing to do with `Hadoop File System reverse lookups`.

IMO, the hdfs filesystem should be same as object stores, s3a/s3n, we can also optimizie hdfs fs like HIVE-24838.  https://github.com/apache/hive/pull/2085#issue-833982804 also said `getDnsPath` : `The only useful thing this function does is transforming a relative path to an absolute path.`

I hope someone can help me to understand this comment about the `Hadoop File System reverse lookups paths`.
@ayushtkn Do you have any thought?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1796,8 +1796,6 @@ public enum ConfVars {
         "hive.metastore.custom.database.product.classname", "none",
           "Hook for external RDBMS. This class will be instantiated only when " +
           "metastore.use.custom.database.product is set to true."),
-    HIVE_BLOBSTORE_SUPPORTED_SCHEMES("hive.blobstore.supported.schemes", "hive.blobstore.supported.schemes", "s3,s3a,s3n",
```
zhangbutao, line 1799: I think we can delete this propertity, we can do optimizie the fs call for all schemas including hdfs.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
ayushtkn: I think this comment is around if someone gives the ip:port of the namenode rather than the canonical name or the nameservice like he supplies hdfs://192.168.1.100:8020 rather hdfs://namenode.example.com:8020 or the namespace & this can lead to mismatch when we are doing some URI matching or so. I haven't gone through this code. but somewhere it is talking about the case where some one provides the IP:Port of the namenode directly

cc. @cnauroth in case you have some pointers around this

cnauroth: That is some very old code, going all the way back to [HIVE-467](https://issues.apache.org/jira/browse/HIVE-467). `TestWarehouseDnsPath` wasn't added until [HIVE-24838](https://issues.apache.org/jira/browse/HIVE-24838), and I see the discussion on the pull request about how the method doesn't really do what the comments state.

I can't think of any reason why the logic wouldn't work for HDFS. Maybe at the time of HIVE-467, the HDFS client had different behavior than it does today, but I can't think of anything specific.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
cnauroth: This property is also defined in `HiveConf` and triggers optimizations specific to blob stores. I just want to confirm, you are only proposing removing it from metastore?

cnauroth (commented):
@zhangbutao , thank you for the patch. It looks good overall. AFAIK, this logic can work for all file systems, including HDFS. I entered one small question.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
zhangbutao: Thanks @cnauroth ! Good catch! Let me check it later.

zhangbutao: @cnauroth The property defined in HiveConf has other purposes, namely to distinguish between hdfs and s3.

But the property in MetastoreConf is only used to optimize the s3 rpc call in **Warehouse::getDnsPath**, and i think the optimization can be used both s3 and hdfs. So we can remove the property in MetastoreConf & no need to distinguish between s3 and hdfs.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
zhangbutao: Yes, this is a very old code&comment. :(

Bsed on the test https://github.com/apache/hive/blob/master/itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/TestWarehouseDnsPath.java , I didn't find any clues about `Hadoop File System reverse lookups paths`. But maybe this test can not cover the case which has multiple namenodes(hdfs federation).  I also don't have a hdfs federation cluster to have test. :(

The purpose of this code is to assume that s3 and hdfs have the same absolute path acquisition method. That is to say `hive.blobstore.supported.schemes` can also include hdfs, then this can also reduce some hdfs rpc call.  If this, we can simplify this code to remove the property `hive.blobstore.supported.schemes` and let the optimization apply to all schemas(s3a, hdfs) by default.

Now If we want to be careful about this optimization, we can just add `hdfs ` into `hive.blobstore.supported.schemes` in MetastoreConf, and people can remove `hdfs `from this property if finding some issue about getting hdfs path.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
```diff
@@ -151,40 +145,24 @@ public FileSystem getFs(Path f) throws MetaException {
...
+        || StringUtils.equalsIgnoreCase(scheme, defaultUri.getScheme())) {
+      if (authority == null) {
+        authority = defaultUri.getAuthority();
+      }
+      if (scheme == null) {
+        scheme = defaultUri.getScheme();
       }
-      return path;
-    } else { // fallback: for other FS type make the FS instance
-      FileSystem fs = getFs(path, conf);
```
zhangbutao, line 174: `FileSystem fs = getFs(path, conf);`  This PR is just want to remove this code for all schemas(s3, hdfs), so this can reduce some filesystem rpc call.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
cnauroth: I am onboard (+1 non-binding) with applying this logic to all file systems, not just blobstores. I can't think of any harm from this.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
cnauroth: I agree with you. It seems we can safely remove from metastore, and leave the configuration available to the rest of Hive.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
cnauroth: Agreed, the optimization seems good for all file systems.

I think it's more impactful for blobstores like S3, where the bucket is part of the authority, and therefore different buckets mean different `FileSystem` instances. The optimization is still valid for HDFS though.

deniskuzZ:
@zhangbutao should we merge?

deniskuzZ (approved):
+1, pending tests

PR #5510, opened 2024-10-14
HIVE-28576: Add jdbc tests for tpcds queries

soumyakanti3578:

Add a new test driver for jdbc tests `TestMiniLlapLocalJdbcCliDriver` to create tpcds tables in dockerized postgres, and create corresponding external tables. Generate explain plans for tpcds queries.

To improve JDBC test coverage

No

No

```
mvn test -pl itests/qtest  -Pitests -Dtest=TestMiniLlapLocalJdbcCliDriver -Dtest.output.overwrite=true
```


deniskuzZ:
@soumyakanti3578, why should it be MiniLlapDriver?

soumyakanti3578:
@deniskuzZ No, TestTezTPCDS30TBPerfCliDriver generates logical plans and uses a stats dump. This is different because here we actually create tpcds tables in Postgres, create external tables in Hive corresponding to the tables in Postgres, and then generate the jdbc plans.

deniskuzZ:
> @deniskuzZ No, TestTezTPCDS30TBPerfCliDriver generates logical plans and uses a stats dump. This is different because here we actually create tpcds tables in Postgres, create external tables in Hive corresponding to the tables in Postgres, and then generate the jdbc plans.

why not simply parameterize the db to run tests against? default is derby, but can't be overridden. why create a separate driver for that?

soumyakanti3578:
@deniskuzZ 

> why not simply parameterize the db to run tests against?

Only option I saw was to add something like `--!qt:database:postgres:<script-to-run-in-postgres>`. Did you have this in mind as well? I didn't go with this approach because I need to run all tpcds queries on external tables. There are 2 ways to do this:
1. Have a single qfile - `test_all_tpcds_queries_postgres.q` with `--!qt:database:postgres:q_test_tpcds_tables.postgres.sql` ([q_test_tpcds_tables.postgres.sql](https://github.com/apache/hive/pull/5510/files#diff-a05f632368b15e78cbf5c2f71aeb19278b867d94e6a370168263295602c61966)) on the top of the file to create all tpcds queries in postgres. Then we would need the contents of [q_test_external_tpcds_tables_postgres.q](https://github.com/apache/hive/pull/5510/files#diff-a9cdca90dd8b5872a171faa1c815504f2078040df7328370482b754e6c32bcd4) in `test_all_tpcds_queries_postgres.q` to create all external tables. And then, we will need all tpcds queries too. This would be a huge file, and the resulting qout file could be very difficult to maintain too.
2. Other approach that I can think of is to have individual q files for each tpcds query, but then we will need to create `<script-to-run-in-postgres>` for each one of them if we want to only create the required tables in postgres. We can also just create all tpcds tables for each q file but that would be inefficient I guess. We would still need to create external tables in each q file. For example, for [cbo_query3.q](https://github.com/apache/hive/blob/57f720da75ba5c7416f303730c7a03fae3a08655/ql/src/test/queries/clientpositive/perf/cbo_query3.q), we would need to create 3 jdbc tables (date_dim, store_sales, item) in `<script-to-run-in-postgres-for-cbo_query3>`, and 3 external tables in `cbo_query3_postgres.q`. This could become very difficult to maintain too.

With the new driver, we can configure [CliConfigs.java](https://github.com/apache/hive/pull/5510/files#diff-b27a4591a3b7f43fc1fb488aa1d4b2d1b55b67fb1dea63c2871827fc427a4021) to run 1 file to create all tpcds tables once with 
```
setJdbcInitScript("q_test_tpcds_tables.postgres.sql");
```
and run 1 file to create all external tables once with
```
setExternalTablesForJdbcInitScript("q_test_external_tpcds_tables_postgres.q");
```
and in [CoreJdbcCliDriver.java](https://github.com/apache/hive/pull/5510/files#diff-6275c227d594ccf98ef6bf8adaad5eb8915bc0bb406a139c7ee7c036991fcecf), we can launch the docker container before starting the tests with `beforeClass()`, run all the tests, then stop the docker container in `shutdown()` method.

I think this approach is easier to maintain.

But please let me know what your thoughts are regarding this, and if you have something else in mind too!

Thanks!

deniskuzZ:
hi @soumyakanti3578, sorry for not replying, please give me some time, I am overwhelmed with diff requests atm

soumyakanti3578:
@deniskuzZ No problem at all. Thanks for looking into this :)

deniskuzZ:
q_test_jdbc_schema_init-postgres.sql, 
q_test_hive_schema_init-postgres.sql

itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CoreJdbcCliDriver.java
```diff
@@ -0,0 +1,104 @@
...
+    if (cliConfig.getJdbcInitScript() != null) {
+      LOG.info("Launching docker container, running jdbc init script...");
+      java.nio.file.Path scriptFile = Paths.get(
+          QTestUtil.getScriptsDir(getQt().getConf()) + File.separator + cliConfig.getJdbcInitScript()
+      );
+      if (Files.notExists(scriptFile)) {
+        LOG.info("No jdbc init script detected. Skipping");
+        return;
+      }
+      externalDB = QTestDatabaseHandler.DatabaseType.valueOf("POSTGRES").create();
```
deniskuzZ, line 61: can we reuse QTestDatabaseHandler here?

soumyakanti3578:
Thanks @deniskuzZ for reopening the PR. I will look into your suggestions, and also the failure, and revive this PR soon. :)

itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CoreJdbcCliDriver.java
soumyakanti3578: Unfortunately I don't think we can reuse `QTestDatabaseHandler` here.
We want to create a DB in `@BeforeClass` method so that the tables are available for all 200 tests. Otherwise we will have to create tpcds tables and corresponding hive external tables again and again for each test file.

`QTestDatabaseHandler` is set up to create DB and launch docker for each individual test, as it is done in `beforeTest` method.

Please let me know if I am missing something, and if there is a better way to handle this.

deniskuzZ:
````
QTestDatabaseHandler dbHandler = qt.getDatabaseHandler();
dbHandler.initDb("postgres", scriptFile)
````

File: itests/util/src/main/java/org/apache/hadoop/hive/cli/control/AbstractCliConfig.java [hunk 2]
```diff
@@ -345,6 +347,22 @@ protected void setInitScript(String initScript) {
       this.initScript = initScript;
     }
   }
+
+  public String getJdbcInitScript() {
```
deniskuzZ, line 351: can't we reuse the existing `initScript`?

PR #5512, opened 2024-10-17
HIVE-28580: Create table using import command gives ddl pointing to i…

dengzhhu653:
…ncorrect location

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java
@@ -772,21 +772,7 @@ public class ImportSemanticAnalyzer extends BaseSemanticAnalyzer {
         tgtPath = new Path(table.getDataLocation().toString(),
             Warehouse.makePartPath(partSpec.getPartSpec()));
       } else {
-        try {
-          Table translatedTable =
-              x.getHive().getTranslateTableDryrun(tblDesc.toTable(x.getConf()).getTTable());
-          Path tablePath = translatedTable.getDataLocation();
-          if (tablePath == null) {
-            Database parentDb = x.getHive().getDatabase(tblDesc.getDatabaseName());
-            tablePath = wh.getDefaultTablePath(parentDb, tblDesc.getTableName(), tblDesc.isExternal());
-          }
-          tgtPath = new Path(tablePath, Warehouse.makePartPath(partSpec.getPartSpec()));
-        } catch (MetaException | HiveException | IOException e) {
-          throw e;
-        } catch (Exception e) {
-          throw new HiveException("Unable to determine the target path of the partition: " +
-              partSpec.getPartSpec(), e);
-        }
+        tgtPath = new Path(getTableDataLocation(wh, tblDesc, x), Warehouse.makePartPath(partSpec.getPartSpec()));
       }
     } else {
       tgtPath = new Path(tblDesc.getLocation(),
@@ -1119,7 +1105,7 @@ public class ImportSemanticAnalyzer extends BaseSemanticAnalyzer {
           if (tblDesc.getLocation() != null) {
             tablePath = new Path(tblDesc.getLocation());
           } else {
-            tablePath = wh.getDefaultTablePath(parentDb, tblDesc.getTableName(), tblDesc.isExternal());
+            tablePath = getTableDataLocation(wh, tblDesc, x);
           }
           FileSystem tgtFs = FileSystem.get(tablePath.toUri(), x.getConf());
           checkTargetLocationEmpty(tgtFs, tablePath, replicationSpec,x.getLOG());
@@ -1414,4 +1400,15 @@ public class ImportSemanticAnalyzer extends BaseSemanticAnalyzer {
     }
   }
 
+  private static Path getTableDataLocation(Warehouse wh, ImportTableDesc tblDesc,
+      EximUtil.SemanticAnalyzerWrapperContext x) throws HiveException, MetaException {
+    Table translatedTable =
+        x.getHive().getTranslateTableDryrun(tblDesc.toTable(x.getConf()).getTTable());
+    Path tablePath = translatedTable.getDataLocation();
+    if (tablePath == null) {
+      Database parentDb = x.getHive().getDatabase(tblDesc.getDatabaseName());
+      tablePath = wh.getDefaultTablePath(parentDb, tblDesc.getTableName(), tblDesc.isExternal());
+    }
+    return tablePath;
+  }
 }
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java
```diff
@@ -772,10 +772,21 @@ we use isExternal and not tableType() method since that always gives type as man
         tgtPath = new Path(table.getDataLocation().toString(),
             Warehouse.makePartPath(partSpec.getPartSpec()));
       } else {
-        Database parentDb = x.getHive().getDatabase(tblDesc.getDatabaseName());
-        tgtPath = new Path(
-            wh.getDefaultTablePath( parentDb, tblDesc.getTableName(), tblDesc.isExternal()),
-            Warehouse.makePartPath(partSpec.getPartSpec()));
+        try {
+          Table translatedTable =
```
deniskuzZ, line 776: can we extract the get location logic and the construct the. final tgt path?
````
String path = getLocation(tblDesc, partSpec, table);
tgtPath = new Path(path, Warehouse.makePartPath(partSpec.getPartSpec()));
````

File: ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java [hunk 2]
```diff
@@ -772,10 +772,21 @@ we use isExternal and not tableType() method since that always gives type as man
...
+              x.getHive().getTranslateTableDryrun(tblDesc.toTable(x.getConf()).getTTable());
+          Path tablePath = translatedTable.getDataLocation();
+          if (tablePath == null) {
+            Database parentDb = x.getHive().getDatabase(tblDesc.getDatabaseName());
+            tablePath = wh.getDefaultTablePath(parentDb, tblDesc.getTableName(), tblDesc.isExternal());
+          }
+          tgtPath = new Path(tablePath, Warehouse.makePartPath(partSpec.getPartSpec()));
+        } catch (MetaException | HiveException | IOException e) {
```
deniskuzZ, line 784: why do we need to catch this, method has throws definition

ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java
dengzhhu653: sure will do that

ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java
```diff
@@ -772,10 +772,7 @@ we use isExternal and not tableType() method since that always gives type as man
         tgtPath = new Path(table.getDataLocation().toString(),
             Warehouse.makePartPath(partSpec.getPartSpec()));
       } else {
-        Database parentDb = x.getHive().getDatabase(tblDesc.getDatabaseName());
-        tgtPath = new Path(
-            wh.getDefaultTablePath( parentDb, tblDesc.getTableName(), tblDesc.isExternal()),
-            Warehouse.makePartPath(partSpec.getPartSpec()));
+        tgtPath = new Path(getTableDataLocation(wh, tblDesc, x), Warehouse.makePartPath(partSpec.getPartSpec()));
```
deniskuzZ, line 775: why don't you store data location into `tablePath` and extract from all the brances method for building the part path
````
tgtPath = new Path(tablePath, Warehouse.makePartPath(partSpec.getPartSpec()))
````

File: ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java [hunk 4]
```diff
@@ -1108,7 +1105,7 @@ private static void createRegularImportTasks(
           if (tblDesc.getLocation() != null) {
             tablePath = new Path(tblDesc.getLocation());
           } else {
-            tablePath = wh.getDefaultTablePath(parentDb, tblDesc.getTableName(), tblDesc.isExternal());
+            tablePath = getTableDataLocation(wh, tblDesc, x);
```
deniskuzZ, line 1108: why don't you move above `if (tblDesc.getLocation() != null)` to the helper method?

ql/src/java/org/apache/hadoop/hive/ql/plan/ImportTableDesc.java
```diff
@@ -175,7 +176,7 @@ public TableType tableType() {
     return TableType.MANAGED_TABLE;
   }
 
-  public Table toTable(HiveConf conf) throws Exception {
+  public Table toTable(HiveConf conf) throws HiveException {
```
saihemanth-cloudera, line 179: Will the CreateTableLike operation also use the same code path? It would be good to a test case for the same.

ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java
```diff
@@ -1403,4 +1400,15 @@ public static Table tableIfExists(ImportTableDesc tblDesc, Hive db) throws HiveE
     }
   }
 
+  private static Path getTableDataLocation(Warehouse wh, ImportTableDesc tblDesc,
+      EximUtil.SemanticAnalyzerWrapperContext x) throws HiveException, MetaException {
+    Table translatedTable =
+        x.getHive().getTranslateTableDryrun(tblDesc.toTable(x.getConf()).getTTable());
```
saihemanth-cloudera, line 1406: Any chance that SemanticAnalyzerWrapperContext can be null?

ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java
dengzhhu653: yes, we can. I will fix it. Thank you!

ql/src/java/org/apache/hadoop/hive/ql/plan/ImportTableDesc.java
dengzhhu653: I think we don't need to, this method will only throw HiveException.

ql/src/java/org/apache/hadoop/hive/ql/parse/ImportSemanticAnalyzer.java
dengzhhu653: no, we always create it at ImportSemanticAnalyzer#L170

deniskuzZ (approved):
LGTM +1, thanks for the patch @dengzhhu653!

PR #5516, opened 2024-10-24
HIVE-17306: Support MySQL InnoDB Cluster

KiranVelumuri:

Introduce primary key columns to support HA for MySQL cluster. This works for MySQL 8.0.23 and above.

[HIVE-17306](https://issues.apache.org/jira/browse/HIVE-17306)

No

No

Perform schema upgrade and verify transactions are working fine, and the meta tables are getting updated as earlier.


KiranVelumuri:
mysql in test environment is at 5.x.x, which does not have support for INVISIBLE columns

PR #5518, opened 2024-10-25
HIVE-28589: Not null constraint does not enforced at invalid cast

kasakrisz:
Move Filter operator with enforce constraint call generation after conversion Select operator generation and make the Filter operator the child operator of the conversion Select operator.

Currently, implicit type conversion occurs after constraints are enforced; however, the result of an invalid cast can be `null`. In these cases, the `null` value is inserted despite the fact that a not-null constraint is specified on the target column.

See the execution plan if this simple insert:
```
set hive.stats.autogather=false;

create table t1(
  date_col date not null
);

explain
insert into t1 values ('2024-99-99');
```
```
                        Filter Operator
                          predicate: enforce_constraint(_col0 is not null) (type: boolean)
                          Statistics: Num rows: 1 Data size: 8 Basic stats: COMPLETE Column stats: COMPLETE
                          Select Operator
                            expressions: CAST( _col0 AS DATE) (type: date)
                            outputColumnNames: _col0
                            Statistics: Num rows: 1 Data size: 56 Basic stats: COMPLETE Column stats: COMPLETE
```

Yes. Such inserts will fail with an error message like:
```
org.apache.hadoop.hive.ql.exec.errors.DataConstraintViolationError: Either CHECK or NOT NULL constraint violated!
```

No

```
mvn test -Dtest.output.overwrite -Dtest=TestNegativeLlapLocalCliDriver -Dqfile=constraint_invalid_cast.q,constraint_invalid_cast_partition.q -pl itests/qtest -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
@@ -7527,9 +7527,6 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
         dpCtx.setRootPath(queryTmpdir);
       }
 
-      // Add NOT NULL constraint check
-      input = genConstraintsPlan(dest, qb, input);
-
       if (!qb.getIsQuery()) {
         isAlreadyContainsPartCols = Optional.ofNullable(destinationTable)
             .map(Table::getStorageHandler)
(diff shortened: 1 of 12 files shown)


deniskuzZ (approved):
LGTM +1

kasakrisz:
@deniskuzZ 
Thanks for the review. I rebased the patch to master to avoid any conflicts.

PR #5520, opened 2024-10-28
HIVE-28579: Add query id to HiveSQLException

difin:

Add query id to HiveSQLException's message. Using query id more info can easily be retrieved from Hiveserver2 log.

- When `'hive.server2.async.exec.async.compile' = 'true' `and an exception happens, an operational log is not displayed in Beeline and Hue (because stmtHandle is empty) and only exception message is displayed. Query id is not displayed.
- When `'hive.server2.async.exec.async.compile' = 'false'` and operational log level != 'NONE', operational log is displayed with query id and exception stack trace and also exception message is displayed.

When a query users should know Query ID, because using one they can find stack traces and other info in HiveServer2 log.

Yes

No

Added new JDBC test to test that query id is added to exception message.

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestExceptionMessageWithJdbc.java
+++ b/itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestExceptionMessageWithJdbc.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
 import org.apache.hadoop.hive.conf.HiveConfForTest;
 import org.apache.hadoop.hive.ql.ServiceContext;
+import org.apache.hadoop.hive.ql.hooks.HiveProtoLoggingHook.ExecutionMode;
 import org.apache.hive.jdbc.miniHS2.MiniHS2;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
@@ -139,7 +140,7 @@ public class TestExceptionMessageWithJdbc {
 
   private static HiveConf getNewHiveConf() {
     HiveConf conf = new HiveConfForTest(TestExceptionMessageWithJdbc.class);
-    conf.setVar(ConfVars.HIVE_EXECUTION_ENGINE, "mr");
+    conf.setVar(ConfVars.HIVE_EXECUTION_ENGINE, ExecutionMode.TEZ.name());
     return conf;
   }
 
--- a/itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestExceptionMessageWithJdbc.java
+++ b/itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestExceptionMessageWithJdbc.java
@@ -1,0 +1,157 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hive.jdbc;
+
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
+import org.apache.hadoop.hive.conf.HiveConfForTest;
+import org.apache.hadoop.hive.ql.ServiceContext;
+import org.apache.hive.jdbc.miniHS2.MiniHS2;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class TestExceptionMessageWithJdbc {
+  private static MiniHS2 miniHS2;
+  private static Connection conDefault;
+  private static Connection conTestDb;
+  private static final String TEST_DB_NAME = "test_logging_level";
+  private static final String USERNAME = System.getProperty("user.name");
+  private static final String PASSWORD = "bar";
+
+  @BeforeClass
+  public static void setupBeforeClass() throws Exception {
+    MiniHS2.cleanupLocalDir();
+    HiveConf conf = getNewHiveConf();
+    try {
+      startMiniHS2(conf);
+    } catch (Exception e) {
+      System.out.println("Unable to start MiniHS2: " + e);
+      throw e;
+    }
+
+    try {
+      openDefaultConnections();
+    } catch (Exception e) {
+      System.out.println("Unable to open default connections to MiniHS2: " + e);
+      throw e;
+    }
+    Statement stmt = conDefault.createStatement();
+    stmt.execute("drop database if exists " + TEST_DB_NAME + " cascade");
+    stmt.execute("create database " + TEST_DB_NAME);
+    stmt.close();
+
+    try {
+      openTestConnections();
+    } catch (Exception e) {
+      System.out.println("Unable to open default connections to MiniHS2: " + e);
+      throw e;
+    }
+  }
+
+  private static Connection getConnection() throws Exception {
+    return getConnection(miniHS2.getJdbcURL(), USERNAME, PASSWORD);
+  }
+
+  private static Connection getConnection(String dbName) throws Exception {
+    return getConnection(miniHS2.getJdbcURL(dbName), USERNAME, PASSWORD);
+  }
+
+  private static Connection getConnection(String jdbcURL, String user, String pwd)
+      throws SQLException {
+    Connection conn = DriverManager.getConnection(jdbcURL, user, pwd);
+    assertNotNull(conn);
+    return conn;
+  }
+
+  @AfterClass
+  public static void tearDownAfterClass() throws Exception {
+    Statement stmt = conDefault.createStatement();
+    stmt.execute("set hive.support.concurrency = false");
+    stmt.execute("drop database if exists " + TEST_DB_NAME + " cascade");
+    stmt.close();
+    if (conTestDb != null) {
+      conTestDb.close();
+    }
+    if (conDefault != null) {
+      conDefault.close();
+    }
+    stopMiniHS2();
+    cleanupMiniHS2();
+  }
+
+  private static void startMiniHS2(HiveConf conf) throws Exception {
+    MiniHS2.Builder builder = new MiniHS2.Builder().withConf(conf).cleanupLocalDirOnStartup(false);
+    miniHS2 = builder.build();
+    
+    Map<String, String> confOverlay = new HashMap<>();
+    miniHS2.start(confOverlay);
+  }
+
+  private static void stopMiniHS2() {
+    if ((miniHS2 != null) && (miniHS2.isStarted())) {
+      miniHS2.stop();
+    }
+  }
+
+  private static void cleanupMiniHS2() throws IOException {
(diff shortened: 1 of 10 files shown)


beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -182,6 +182,7 @@ public class BeeLine implements Closeable {
   private static final String HIVE_CONF_PREFIX = "--hiveconf";
   private static final String PROP_FILE_PREFIX = "--property-file";
   public static final String PASSWD_MASK = "[passwd stripped]";
+  public static final String BEELINE_DEFAULT_OPERATIONAL_LOG_LEVEL = "EXECUTION";
```
deniskuzZ, line 185: keep it simple: DEFAULT_OPERATIONAL_LOG_LEVEL, it's inside of BeeLine class

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -4117,7 +4117,7 @@ public static enum ConfVars {
         "${system:java.io.tmpdir}" + File.separator + "${system:user.name}" + File.separator +
             "operation_logs",
         "Top level directory where operation logs are stored if logging functionality is enabled"),
-    HIVE_SERVER2_LOGGING_OPERATION_LEVEL("hive.server2.logging.operation.level", "EXECUTION",
+    HIVE_SERVER2_LOGGING_OPERATION_LEVEL("hive.server2.logging.operation.level", "NONE",
```
deniskuzZ, line 4120: wait, why change here, when you override on client side?

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
```diff
@@ -58,6 +60,10 @@ public HiveSQLException(String reason) {
     super(reason);
   }
 
+  public HiveSQLException(String reason, QueryState queryState) {
+    super(ServiceUtils.addQueryId(reason, queryState), reason);
```
deniskuzZ, line 64: i think ServiceUtils should provide `getQueryId` method and final message could be constructed in HiveSQLException

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -880,6 +881,12 @@ private boolean connectUsingArgs(BeelineParser beelineParser, CommandLine cl) {
       setHiveConfVar(key, hiveConfs.getProperty(key));
     }
 
+    String opLogLevelVarName = HiveConf.ConfVars.HIVE_SERVER2_LOGGING_OPERATION_LEVEL.varname;
```
deniskuzZ, line 884: i think that should be downstream only change where could disable `HIVE_SERVER2_LOGGING_OPERATION_LEVEL`

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
difin: We override only on Beeline because as we said Beeline is mostly used by developers.
It's for other Hive client tools, like Hue.

beeline/src/java/org/apache/hive/beeline/BeeLine.java
difin: This is related to the previous comment.

beeline/src/java/org/apache/hive/beeline/BeeLine.java
difin: Done

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
difin: Done

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
deniskuzZ: are we disabling it for the JDBC connections?

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
```diff
@@ -150,4 +189,19 @@ public static TStatus toTStatus(Exception e) {
     return tStatus;
   }
 
+  private static String addQueryId(String message, QueryState queryState) {
```
deniskuzZ, line 192: maybe `getErrorMsg(QueryState queryState, String reason)`

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
```diff
@@ -150,4 +189,19 @@ public static TStatus toTStatus(Exception e) {
...
+    boolean opLogEnabled = HiveConf.getBoolVar(queryState.getConf(),
+        HiveConf.ConfVars.HIVE_SERVER2_LOGGING_OPERATION_ENABLED, false);
+    String opLogLevel = HiveConf.getVar(queryState.getConf(),
+        HiveConf.ConfVars.HIVE_SERVER2_LOGGING_OPERATION_LEVEL, StringUtils.EMPTY);
+    boolean asyncCompile = HiveConf.getBoolVar(queryState.getConf(),
+        HiveConf.ConfVars.HIVE_SERVER2_ASYNC_EXEC_ASYNC_COMPILE, false);
+
+    if ((!opLogEnabled || opLogLevel.equals(OperationLog.LoggingLevel.NONE.name()) || !asyncCompile) &&
```
deniskuzZ, line 200: why wrap in brackets OR predicates?

service/src/java/org/apache/hive/service/cli/operation/SQLOperation.java
```diff
@@ -211,13 +210,13 @@ private void prepare(QueryState queryState) throws HiveSQLException {
       setHasResultSet(driver.hasResultSet());
     } catch (CommandProcessorException e) {
       setState(OperationState.ERROR);
-      throw toSQLException("Error while compiling statement", e);
+      throw toSQLException("Error while compiling statement", e, queryState);
```
deniskuzZ, line 213: don't we add the queryId based on operationLog mode?

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -880,6 +881,11 @@ private boolean connectUsingArgs(BeelineParser beelineParser, CommandLine cl) {
...
+    if (getOpts().getHiveConfVariables().get(opLogLevelVarName) == null) {
```
deniskuzZ, line 885: can we use `contains`?

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
deniskuzZ, line 200: 1. what is with `asyncCompile`? we return the operation log regardless if it's enabled?
2. why are we checking if the message doesn't contain QUERY_ID pattern? What if some exceptions do and we still want to get the operation log?

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
difin: Done

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
```diff
@@ -150,4 +185,11 @@ public static TStatus toTStatus(Exception e) {
     return tStatus;
   }
 
+  private static String getErrorMsg(String message, String queryId) {
```
deniskuzZ, line 188: i think it could be done simpler, just override the Throwable#getMessage
````
@Override
String getMessage(){
String errorMsg = super.getMessage();
 if (!errorMsg.contains(ServiceContext.QUERY_ID)) {
      return String.format("%1$s; %2$s: %3$s", errorMsg, ServiceContext.QUERY_ID, queryId);
    } else {
      return errorMsg;
    }
}
````

File: beeline/src/java/org/apache/hive/beeline/BeeLine.java [hunk 10]
```diff
@@ -182,6 +182,7 @@ public class BeeLine implements Closeable {
   private static final String HIVE_CONF_PREFIX = "--hiveconf";
   private static final String PROP_FILE_PREFIX = "--property-file";
   public static final String PASSWD_MASK = "[passwd stripped]";
+  public static final String DEFAULT_OPERATIONAL_LOG_LEVEL = "EXECUTION";
```
deniskuzZ, line 185: let's drop it

itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestExceptionMessageWithJdbc.java
```diff
@@ -0,0 +1,157 @@
...
+    conDefault = getConnection();
+  }
+
+  private static void openTestConnections() throws Exception {
+    conTestDb = getConnection(TEST_DB_NAME);
+  }
+
+  private static HiveConf getNewHiveConf() {
+    HiveConf conf = new HiveConfForTest(TestExceptionMessageWithJdbc.class);
+    conf.setVar(ConfVars.HIVE_EXECUTION_ENGINE, "mr");
```
ayushtkn, line 142: Why are we running the tests with MR, MR is deprecated we should avoid adding new tests with it

ql/src/java/org/apache/hadoop/hive/ql/ServiceContext.java
```diff
@@ -30,6 +30,7 @@ public class ServiceContext {
   private final String clusterId;
   private final Supplier<String> host;
   private final Supplier<Integer> port;
+  public static final String QUERY_ID = "Query ID";
```
ayushtkn, line 33: I am not sure if this is the correct place to keep this variable, The Javadoc of ``ServiceContext`` itself says ``Information context about the service, most probably HiveServer2.``

service/src/java/org/apache/hive/service/ServiceUtils.java
```diff
@@ -72,5 +72,4 @@ public static boolean canProvideProgressLog(HiveConf hiveConf) {
     return "tez".equals(hiveConf.getVar(HiveConf.ConfVars.HIVE_EXECUTION_ENGINE)) && hiveConf
         .getBoolVar(HiveConf.ConfVars.HIVE_SERVER2_INPLACE_PROGRESS);
   }
-
```
ayushtkn, line 75: nit
avoid

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
```diff
@@ -150,4 +191,13 @@ public static TStatus toTStatus(Exception e) {
     return tStatus;
   }
 
+  @Override
+  public String getMessage(){
+    String errorMsg = super.getMessage() == null ? "Error" : super.getMessage();
+    if (!errorMsg.contains(ServiceContext.QUERY_ID)) {
+      return String.format("%s; Query ID: %s", errorMsg, queryId);
```
ayushtkn, line 198: Shouldn't we check ``queryId`` not null as well? else someone having an old jar which creates this HiveSqlException without QueryId will start getting this with null?

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
```diff
@@ -150,4 +191,13 @@ public static TStatus toTStatus(Exception e) {
     return tStatus;
   }
 
+  @Override
+  public String getMessage(){
+    String errorMsg = super.getMessage() == null ? "Error" : super.getMessage();
+    if (!errorMsg.contains(ServiceContext.QUERY_ID)) {
```
ayushtkn, line 197: curious: is the contains to avoid appending queryId multiple times, there could be cases where query id is already appended & we want to avoid appending it again?
Something like that?

service/src/java/org/apache/hive/service/cli/operation/Operation.java
```diff
@@ -367,13 +367,13 @@ protected void validateFetchOrientation(FetchOrientation orientation,
       EnumSet<FetchOrientation> supportedOrientations) throws HiveSQLException {
     if (!supportedOrientations.contains(orientation)) {
       throw new HiveSQLException("The fetch type " + orientation.toString() +
-          " is not supported for this resultset", "HY106");
+          " is not supported for this resultset", "HY106", queryState.getQueryId());
     }
   }
 
-  protected HiveSQLException toSQLException(String prefix, CommandProcessorException e) {
+  protected HiveSQLException toSQLException(String prefix, CommandProcessorException e, QueryState queryState) {
```
ayushtkn, line 374: Why are you having ``queryState`` as a parameter & then extracting ``queryId`` from it, we can directly have ``queryId`` as a parameter? as in case of constructor of ``HiveSQLException``

service/src/java/org/apache/hive/service/cli/HiveSQLException.java
```diff
@@ -70,6 +72,11 @@ public HiveSQLException(Throwable cause) {
     super(cause);
   }
 
+  public HiveSQLException(Throwable cause, String queryId) {
```
ayushtkn, line 75: We should even deprecate the old constructors without queryId & remove in the subsequent release

ayushtkn (approved):
LGTM

PR #5522, opened 2024-10-28
HIVE-28593: INSERT OVERWRITE DIRECTORY is not cleaning up the HDFS data

saihemanth-cloudera:
INSERT OVERWRITE DIRECTORY will clean up the hdfs directory when zero rows are available to insert.

HDFS directory is not cleaned up and it can lead to data correctness issues.

No.

No.

Manual testing

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/plan/LoadFileDesc.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/plan/LoadFileDesc.java
@@ -41,6 +41,7 @@ public class LoadFileDesc extends LoadDesc implements Serializable {
   private transient CreateMaterializedViewDesc createViewDesc;
   private boolean isMmCtas;
   private String moveTaskId;
+  private boolean isOverwrite;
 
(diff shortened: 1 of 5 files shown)


ql/src/java/org/apache/hadoop/hive/ql/plan/LoadFileDesc.java
```diff
@@ -152,4 +153,12 @@ public String getMoveTaskId() {
...
     this.moveTaskId = moveTaskId;
   }
+
+  public boolean getIsOverwrite() {
+    return isOverwrite;
+  }
+
+  public void setIsOverwrite(boolean isOverwrite) {
+    this.isOverwrite = isOverwrite;
+  }
```
difin, line 163: It is better to name the getters/setters for boolean members like `getOverwrite()` and `setOverwrite`.

https://stackoverflow.com/questions/5322648/for-a-boolean-field-what-is-the-naming-convention-for-its-getter-setter

PR #5524, opened 2024-10-29
HIVE-26582: Cartesian join fails if the query has an empty table when…

soumyakanti3578:
… cartesian product edge is used

https://issues.apache.org/jira/browse/HIVE-26582

Add a new rule to prune empty tables from plans. Currently the config `HiveZeroMaxRowsRuleConfig` is a copy of Calcite's `ZeroMaxRowsRuleConfig` - [CALCITE-5314](https://issues.apache.org/jira/browse/CALCITE-5314), but it can be removed after upgrading to at least Calcite 1.33

It can lead to issues as described in HIVE-26582

No

No

```
mvn test -pl itests/qtest  -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile=prune_empty_table.q
```


ql/src/test/queries/clientpositive/prune_empty_table.q
```diff
@@ -0,0 +1,24 @@
+set hive.tez.cartesian-product.enabled=true;
```
zabetak, line 1: Is the property necessary?

ql/src/test/queries/clientpositive/prune_empty_table.q
```diff
@@ -0,0 +1,24 @@
...
+create table tmp2 (a int) stored as orc;
+
+insert into table c values (3);
+insert into table tmp1 values (3);
+
+explain cbo
+with
+first as (
+select a1 from c where a1 = 3
+),
```
zabetak, line 14: CTEs make the query more complex and add another dimension to the problem. If possible let's just skip it.

Can't we repro the problem with just two tables that one of them is empty? Do we need union all + 3 tables?

Since the query was failing at runtime consider adding also a full execution of the query with the results.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/stats/HiveRelMdMaxRowCount.java
```diff
@@ -0,0 +1,54 @@
...
+    if (!StatsUtils.areBasicStatsUptoDateForQueryAnswering(
+        table.getHiveTableMD(), table.getHiveTableMD().getParameters())) {
+      return null;
+    }
+    // if basic stats are up-to-date and the table is not dummy, return 0.0D if table is empty
+    // super returns infinity. 
+    return !table.getName().equals(SemanticAnalyzer.DUMMY_DATABASE + "." + SemanticAnalyzer.DUMMY_TABLE) &&
+        StatsUtils.isTableEmpty(table) ?
+        Double.valueOf(0.0) :
+        super.getMaxRowCount(hiveTableScan, mq);
```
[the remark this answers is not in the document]
soumyakanti3578: Yes, this can be moved to RelOptHiveTable.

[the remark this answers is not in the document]
soumyakanti3578: I will return `Double.POSITIVE_INFINITY` when table is not empty.

ql/src/test/queries/clientpositive/prune_empty_table.q
soumyakanti3578: It can be removed 👍🏼

ql/src/test/queries/clientpositive/prune_empty_table.q
soumyakanti3578: I will simplify the test and also add another query without explain.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/stats/HiveRelMdMaxRowCount.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.hadoop.hive.ql.optimizer.calcite.RelOptHiveTable;
+import org.apache.hadoop.hive.ql.optimizer.calcite.reloperators.HiveTableScan;
+import org.apache.hadoop.hive.ql.parse.SemanticAnalyzer;
+import org.apache.hadoop.hive.ql.stats.StatsUtils;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Extends {@link RelMdMaxRowCount} to get max row count for {@link HiveTableScan}
+ */
+public class HiveRelMdMaxRowCount extends RelMdMaxRowCount {
```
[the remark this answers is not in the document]
soumyakanti3578: Actually, I understood that it can't be removed completely right now. But I have moved the logic to RelOptHiveTable

deniskuzZ:
hi @soumyakanti3578 , @zabetak, is that still relevant? should we reopen?

PR #5526, opened 2024-10-30
HIVE-28599 lock stmt parser

mlorek:
Moving lock related statements to a separate Parser file to avoid `code too large` error.


mlorek:
@zabetak could you please add your review?

zabetak (approved):
There isn't an immediate need to do this refactoring but since the PR is ready I guess we can get this in.

PR #5404, opened 2024-08-23
HIVE-26473: Hive JDK-17 support for compile and runtime

akshat0395:
### What changes were proposed in this pull request?

[HIVE-26473](https://issues.apache.org/jira/browse/HIVE-26473): Hive JDK-17 support for compile and runtime

- Various version upgrades for jdk-17 compatibility
- Various test fixes for runtime issues
- CI config changes for jdk-17
- Java 17 coding convention upgrades
- Datanucleus upgrade with patached modules for rdbms

### Why are the changes needed?

To support JDK-17 migration for Hive 

### Does this PR introduce _any_ user-facing change?

Yes
### Is the change a dependency upgrade?

Yes

### How was this patch tested?

CI Tests and Sanity

PR #5527, opened 2024-11-02
HIVE-28603 Fixing Multiple Flaky tests in module serde which are due to non determinism

nikunjagarwal321:
Sorting the fields in function `getDeclaredNonStaticFields()` in ObjectInspectorUtils.java class in order to make the function deterministic. The below flaky tests will be fixed by this.

Multiple flaky tests were detected in _serde_ module while trying to run the tests using the nondex tool. [NonDex](https://github.com/TestingResearchIllinois/NonDex) is a tool for detecting and debugging wrong assumptions on under-determined Java APIs. 

#### Steps to reproduce flakiness using nondex -
```
mvn install -pl serde -DskipTests

mvn -pl serde test -Dtest=org.apache.hadoop.hive.serde2.objectinspector.TestProtocolBuffersObjectInspectors#testProtocolBuffersObjectInspectors

mvn -pl serde edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=org.apache.hadoop.hive.serde2.objectinspector.TestProtocolBuffersObjectInspectors#testProtocolBuffersObjectInspectors
```

#### ERROR logs: 
 

```
[ERROR] Failures: 
[ERROR]   TestProtocolBuffersObjectInspectors.testProtocolBuffersObjectInspectors:71 expected:<test> but was:<[one, two]>
```

Screenshot 2024-11-02 at 2 48 35 PM

### Reason for Failure 

The method **getDeclaredNonStaticFields(Class<?> c)** in class **ObjectInspectorUtils.java** uses class.getDeclaredFields() method which is nondeterministic as per [java docs](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredFields--). 

` Field[] f = c.getDeclaredFields();`

Screenshot 2024-11-02 at 4 26 58 PM

As per the test cases, it is assumed that the return order of the fields will always be the same, which may not be the case.

All the below test cases are impacted because of the same reason : 

- org.apache.hadoop.hive.serde2.objectinspector.TestProtocolBuffersObjectInspectors.testProtocolBuffersObjectInspectors
- org.apache.hadoop.hive.serde2.objectinspector.TestThriftObjectInspectors.testThriftObjectInspectors
- org.apache.hadoop.hive.serde2.objectinspector.TestReflectionObjectInspectors.testReflectionObjectInspectors
- org.apache.hadoop.hive.serde2.columnar.TestLazyBinaryColumnarSerDe.testLazyBinaryColumnarSerDeWithEmptyBinary
- org.apache.hadoop.hive.serde2.columnar.TestLazyBinaryColumnarSerDe.testSerDeInnerNulls
- org.apache.hadoop.hive.serde2.lazybinary.TestLazyBinarySerDe.testLazyBinarySerDe
- org.apache.hadoop.hive.serde2.TestStatsSerde.testLazyBinarySerDe
- org.apache.hadoop.hive.serde2.lazy.TestLazySimpleSerDe#testSerDeParameters
- org.apache.hadoop.hive.serde2.binarysortable.TestBinarySortableSerDe#testBinarySortableSerDe

No

No

The changes were tested by running the unit tests locally. The code is already covered in the above test cases and is actually fixing the flaky tests.

Please let me know if you have any questions or would like to discuss this further for any clarificaitions.

### DIFF
--- a/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
+++ b/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
@@ -1666,7 +1666,7 @@ public final class ObjectInspectorUtils {
       slotField.setAccessible(true);
       return slotField.getInt(field);
     } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException e) {
-      LOG.info("Field not found: {}", e);
+      LOG.info("Field not found:", e);
       return 0; // Default value if there's an error
     }
   }
--- a/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
+++ b/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
@@ -1666,8 +1666,8 @@ public final class ObjectInspectorUtils {
       slotField.setAccessible(true);
       return slotField.getInt(field);
     } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException e) {
-      LOG.info("Field not found:", e);
-      return 0; // Default value if there's an error
+      LOG.error("Error getting a slot value:", e);
+      throw new RuntimeException("Error getting a slot value");
     }
   }
 }
--- a/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
+++ b/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
@@ -1661,13 +1661,20 @@ public final class ObjectInspectorUtils {
    * @return
    */
   private static int getSlotValue(Field field) {
+    Field slotField = null;
+    boolean originalAccessible = false;
     try {
-      Field slotField = Field.class.getDeclaredField("slot");
+      slotField = Field.class.getDeclaredField("slot");
+      originalAccessible = slotField.isAccessible();
       slotField.setAccessible(true);
       return slotField.getInt(field);
     } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException e) {
       LOG.error("Error getting a slot value:", e);
       throw new RuntimeException("Error getting a slot value");
+    } finally {
+      if (slotField != null) {
+        slotField.setAccessible(originalAccessible);
+      }
     }
   }
 }
--- a/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
+++ b/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
@@ -24,6 +24,7 @@ import java.lang.reflect.Type;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -558,6 +559,7 @@ public final class ObjectInspectorUtils {
   public static Field[] getDeclaredNonStaticFields(Class<?> c) {
     Field[] f = c.getDeclaredFields();
     ArrayList<Field> af = new ArrayList<Field>();
+    Arrays.sort(f, Comparator.comparingInt(ObjectInspectorUtils::getSlotValue));
     for (int i = 0; i < f.length; ++i) {
       if (!Modifier.isStatic(f[i].getModifiers())) {
         af.add(f[i]);
@@ -1651,5 +1653,21 @@ public final class ObjectInspectorUtils {
 
   private ObjectInspectorUtils() {
     // prevent instantiation
+  }
+
+  /**
+   * Returns slot value used for ordering the fields to make it deterministic
+   * @param field : field of a given class
+   * @return
+   */
+  private static int getSlotValue(Field field) {
+    try {
+      Field slotField = Field.class.getDeclaredField("slot");
+      slotField.setAccessible(true);
+      return slotField.getInt(field);
+    } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException e) {
+      LOG.info("Field not found: {}", e);
+      return 0; // Default value if there's an error
+    }
   }
 }


serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
```diff
@@ -1652,4 +1654,20 @@ public static boolean hasAllFieldsSettable(ObjectInspector oi,
...
+   * @param field : field of a given class
+   * @return
+   */
+  private static int getSlotValue(Field field) {
+    try {
+      Field slotField = Field.class.getDeclaredField("slot");
+      slotField.setAccessible(true);
+      return slotField.getInt(field);
+    } catch (NoSuchFieldException | IllegalAccessException | IllegalArgumentException e) {
+      LOG.info("Field not found: {}", e);
```
difin, line 1669: I think this will not log properly.
These braces '{}' are used when you want to format the log message substituting them with provided values, but the second element here is not a string, but an Exception, so it will log an exception with the message "Field not found: {}" as is.

nikunjagarwal321: Hi, 
Thanks for the review. Understood. I have made the changes to log the error. Please let me know if this looks good.

difin: Yes, just I think it would be better to make the exception message something like "Error getting a slot number
If an exception happens here, why do we log it on info level and return 0?
In Comparator zero means equality. 
Maybe it's better to log it on error level and throw a Hive-specific runtime exception?

nikunjagarwal321: Agreed, basically the change is to make the function deterministic so that the mentioned tests are not flaky anymore. As you said, returning 0 can still return random order as it means equality. 
I'll have changed this to throw a new RuntimeException(similar to the exceptions used in other parts of the class) with the same error message.

serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
```diff
@@ -1652,4 +1654,20 @@ public static boolean hasAllFieldsSettable(ObjectInspector oi,
...
+
+  /**
+   * Returns slot value used for ordering the fields to make it deterministic
+   * @param field : field of a given class
+   * @return
+   */
+  private static int getSlotValue(Field field) {
+    try {
+      Field slotField = Field.class.getDeclaredField("slot");
+      slotField.setAccessible(true);
```
difin, line 1666: I don't know if there are any implications of leaving the slot field accessible.
Maybe it is better to revert this flag to the original value in a `Finaly` block.

difin:
LGTM, pending one comment.

serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
nikunjagarwal321: Hey, the slotField is only used within the getSlotValue method, so the accessibility change is limited to this local instance and has no lasting or global impact. Since each call to getSlotValue uses a separate instance, ideally there should be no repercussions. 

Still to ensure security, I'll make the change to handle this and revert it back to its original value.

nikunjagarwal321: Hey, the slotField is only used within the getSlotValue method, so the accessibility change is limited to this local instance and should ideally have no lasting or global impact. Since each call to getSlotValue uses a separate instance, there should be no repercussions.

Still, to ensure security, I have added the code to revert the slot field accessible value back to its original value. Please let me know if this looks good.

nikunjagarwal321: Hi @difin ,
After making the change, I can see in the pipeline that testcase org.apache.hadoop.hive.ql.TestWarehouseExternalDir is failing, but when I try to run the same test in local, it seems to pass. 
Screenshot 2024-11-09 at 1 55 41 PM



The new commit only sets the variable value back to the initial one and the test cases seem to pass in the earlier commits. 
Just had this query regarding if this case legit or is there a way to rerun the pipeline? Just wanted to reproduce the same failure.

difin: You can retriger the entire pipeline.
You can amend your commit on your local `git commit --amend` to make it look modified and force push into your remote branch, this will retriger the pipeline without introducing a new commit.

nikunjagarwal321: I have made changes as per the comment. @difin please let me know if the changes look good.

difin: Yes, looks good to me (LGTM). 
I will be able to merge your changes in a few days, and in the meanwhile let's see if your change gets more reviews.

difin:
@nikunjagarwal321 please retrigger the build, it needs to pass

serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
nikunjagarwal321: Sure @difin , thanks!!

nikunjagarwal321:
Thanks @difin !

kokila-19:
@nikunjagarwal321 Hi, It appears that this commit does not work with JDK 17 and has resulted in multiple test failures. Could you please check this issue?

You can check out the PR here: https://github.com/apache/hive/pull/5404

nikunjagarwal321:
> @nikunjagarwal321 Hi, It appears that this commit does not work with JDK 17 and has resulted in multiple test failures. Could you please check this issue?
> 
> You can check out the PR here: #5404

Hey, 
Thanks for bringing this to my attention. From my understanding, the current master branch is still on Java 8, and the changes I made were based on that version. The modifications I implemented focused on resolving flaky tests in the existing main code, and all related unit tests passed successfully at the time.

I see that the transition to Java 17 is a new initiative you're working on. If the previous changes are not fully compatible with Java 17, we may need to explore alternative approaches. I’d be happy to provide more context on my changes or collaborate on finding a resolution.

Let’s continue the discussion in the new PR where the Java 17 upgrade is being worked on, as this change seems primarily related to the version upgrade and keeping all discussions in one place would be helpful.

PR #5528, opened 2024-11-04
HIVE-28604: Allow HMS to configure the DataNucleus level 1 cache

dengzhhu653:
Allow HMS to configure the DataNucleus level 1 cache

Under heavy memory load, a null value could be obtained from this cache,  leading to the opaque NPE.

Introduce a new property

No

Test against a heavy HMS


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -724,6 +724,9 @@ public enum ConfVars {
     DATANUCLEUS_AUTOSTART("datanucleus.autoStartMechanismMode",
         "datanucleus.autoStartMechanismMode", "ignored", new StringSetValidator("ignored"),
         "Autostart mechanism for datanucleus.  Currently ignored is the only option supported."),
+    DATANUCLEUS_CACHE_LEVEL1_TYPE("datanucleus.cache.level1.type",
+        "datanucleus.cache.level1.type", "soft", new StringSetValidator("none", "soft", "weak", "strong"),
+        "Stores instances within a PersistenceManager"),
```
soumyakanti3578, line 729: Do we need a similar change for `DATANUCLEUS_CACHE_LEVEL2_TYPE` ?

soumyakanti3578 (commented):
@dengzhhu653 could you please add a brief description about the tests that were performed to validate this, either here or preferably in the Jira?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
dengzhhu653: it's already been here: https://github.com/apache/hive/blob/master/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java#L729

dengzhhu653: I would like to leave it as it is, there are multiple implementations for this level cache, some are DataNucleus-provided, others are external cache, https://www.datanucleus.org/products/accessplatform_4_2/extensions/level2_cache.html

saihemanth-cloudera (approved):
LGTM +1

soumyakanti3578 (approved):
LGTM

PR #5389, opened 2024-08-07
HIVE-28047: Iceberg: Major QB Compaction on unpartitioned tables with…

difin:
… a single commit

### What changes were proposed in this pull request?

Currently, when an unpartitioned Iceberg table is compacted, it creates 2 commits: one that truncates the table and the other one that inserts compacted data. This PR proposes to replace this double commit process with one commit using Iceberg replaceFiles API.

### Why are the changes needed?

To compact an unpartitioned table in one commit. This avoids problems with time travel in case a user queries the table at the state of the commit that truncated the table and before the commit when compacted files are inserted.

### Does this PR introduce _any_ user-facing change?

No

### Is the change a dependency upgrade?

No

### How was this patch tested?

Existing query tests

PR #5529, opened 2024-11-05
HIVE-28600: Iceberg: Check that table/partition requires compaction b…

difin:
…efore compacting

Added checking if Iceberg table/partition requires compaction before starting a manual compaction request.

Performance improvement to skip tables/partitions which do not require compaction.

No

No

Updated existing q-tests, added new unit test.


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -458,6 +458,52 @@ public static List<DeleteFile> getDeleteFiles(Table table, String partitionPath)
         t -> ((PositionDeletesScanTask) t).file()));
   }
 
+  public static float getFileSizeRatio(Table table, String partitionPath, long fileSizeInBytesThreshold) {
```
zratkai, line 461: The name of the method does not show what ratio it, as it does nothing with the file actual size, rather it check the number of files. 
Like e.g. getUncompactedRatio would be better.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -458,6 +458,52 @@ public static List<DeleteFile> getDeleteFiles(Table table, String partitionPath)
...
+
+  private static long getDataFileCount(Table table, String partitionPath, long fileSizeInBytesThreshold,
+      boolean isLess) {
+    CloseableIterable<FileScanTask> fileScanTasks =
+        table.newScan().useSnapshot(table.currentSnapshot().snapshotId()).ignoreResiduals().planFiles();
+    CloseableIterable<FileScanTask> filteredFileScanTasks =
+        CloseableIterable.filter(fileScanTasks, t -> {
+          DataFile file = t.asFileScanTask().file();
+          return (!table.spec().isPartitioned() ||
+              partitionPath == null && file.specId() != table.spec().specId() ||
```
zratkai, line 482: Please use better separation between operands to make it readable/understandable and reliable, since now the precedence can cause issues.

Like: ((a==b && c!=2) && (a!=b && c=2) && (...) )

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -458,6 +458,52 @@ public static List<DeleteFile> getDeleteFiles(Table table, String partitionPath)
...
+  }
+
+  public static long countDeleteRecords(Table table, String partitionPath) {
+    Table deletesTable =
+        MetadataTableUtils.createMetadataTableInstance(table, MetadataTableType.POSITION_DELETES);
+    CloseableIterable<ScanTask> deletesScanTasks = deletesTable.newBatchScan().planFiles();
+    CloseableIterable<ScanTask> filteredDeletesScanTasks =
+        CloseableIterable.filter(deletesScanTasks, t -> {
+          DeleteFile file = ((PositionDeletesScanTask) t).file();
+          return !table.spec().isPartitioned() ||
```
zratkai, line 498: Please use better separation between operands to make it readable/understandable and reliable, since now the precedence can cause issues.

Like: ((a==b && c!=2) && (a!=b && c=2) && (...) )

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -458,6 +458,52 @@ public static List<DeleteFile> getDeleteFiles(Table table, String partitionPath)
         t -> ((PositionDeletesScanTask) t).file()));
   }
 
+  public static float getFileSizeRatio(Table table, String partitionPath, long fileSizeInBytesThreshold) {
+    long uncompactedFilesCount = getDataFileCount(table, partitionPath, fileSizeInBytesThreshold, true);
+    long compactedFilesCount = getDataFileCount(table, partitionPath, fileSizeInBytesThreshold, false);
```
zratkai, line 463: Why the fileSizeInBytesThreshold is the comparator if a file is uncompacted or compacted? 
Isn't there a better approach to make this check?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution_ordered.q
```diff
@@ -34,27 +34,30 @@ tblproperties ('format-version'='2');
...
 ('fn9', 'ln9',  3, 20, 101),
-('fn10','ln10', 3, 20, 101),
+('fn10','ln10', 3, 20, 101);
+
+insert into ice_orc VALUES 
 ('fn11','ln11', 3, 20, 101),
 ('fn12','ln12', 3, 20, 101);
 
-select * from ice_orc where company_id = 100;
-select * from ice_orc where dept_id = 2;
```
zratkai, line 56: Why did you delete those selects?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
```diff
@@ -118,6 +118,13 @@ else if (desc.getPartitionSpec() != null) {
         Optional<String> partitionName = partitionMap.keySet().stream().findFirst();
         partitionName.ifPresent(compactionRequest::setPartitionname);
       }
+      if (table.getStorageHandler() != null && 
+          !table.getStorageHandler().canCompact(context.getConf(), table, compactionRequest.getPartitionname(), 
+              compactionRequest.getType())) {
+        LOG.info("Table=%s%s doesn't meet requirements for compaction", table.getTableName(),
```
zratkai, line 124: Please add more description what requirement didn't met. 
e.g.: What was expected, what is the actual?

zratkai (changes requested):
Please check my comments!

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Similar to previous comment, previously I used braces in very similar case, and @deniskuzZ asked to remove them: https://github.com/apache/hive/pull/5389#discussion_r1711415417

The precedence won't cause issues because in Java && operands has higher precedence over ||, so it will be evaluated deterministically.

difin: hm, but there seems to be no contradiction, your test confirms that && has higher precedence than ||.

First case:
```
x && y || z
false && true || true
false || true
true
```

Second case:
```
x && (y || z)
false && (true || true)
false && (true)
false
```

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution_ordered.q
difin: Because with the other changes done to the q file in the PR the order of the records in the selects is not fixed and changes randomly in different executions of the test.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: > it does nothing with the file actual size, rather it check the number of files.
The method uses the actual file sizes, it checks for each file if its actual size is above or below the threshold and returns `(number of files whose size is below threshold) / (total number of files)`.
I think `getUncompactedRatio` is fine, I will rename like that.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/storage/compact/AlterTableCompactOperation.java
difin: I removed logging from this place and instead added it to `HiveIcebergStorageHandler#canCompact`.
This way it will log the exact reason why a table/partition doesn't meet compaction requirements.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution_ordered.q
kasakrisz: how about adding `-- SORT_QUERY_RESULTS` ?

difin: This q-test tests if records get sorted after executing compaction with sort clause.
I removed `-- SORT_QUERY_RESULTS` intentionally, otherwise it won't be possible to test if records are sorted after compaction.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
zratkai: It definitely confirms how important parenthesis is when there are multiple conditions in if. With or without parenthesis it could have a different result, as shown.Which means if it is done the wrong way it results in error.
It also not readable for others who didn't developed it, just try to understand or modify later. Imagine someone has to add or remove a condition in the future and it leads to bugs.

Please check Oracle's recommendation to this topic:
https://www.oracle.com/java/technologies/javase/codeconventions-programmingpractices.html

"10.5 Miscellaneous Practices
10.5.1 Parentheses
It is generally a good idea to use parentheses liberally in expressions involving mixed operators
to avoid operator precedence problems. Even if the operator precedence seems clear to you, it
might not be to others—you shouldn’t assume that other programmers know precedence as
well as you do.
if (a == b && c == d) // AVOID!
if ((a == b) && (c == d)) // RIGHT"

Also some valuable experience from others for this topic. I can not express myself better then the comments in this :
https://softwareengineering.stackexchange.com/questions/201175/should-i-fully-parenthesize-expressions-or-rely-on-precedence-rules
"As for correctness, conditions without parentheses are a recipe for silly mistakes. When they happen, they can be bugs that are hard to find--because often an incorrect condition will behave correctly most of the time, and only occasionally fail."

"Good developers strive to write code that is clear and correct. Parentheses in conditionals, even if they are not strictly required, help with both."
"Furthermore, extra parentheses, just like indentations, whitespace, and other style standards, help visually organize the code in a logical way."
"And even if you get it right, the next person to work on your code may not, either adding errors to the expression or misunderstanding your logic and thus adding errors elsewhere (as LarsH rightly points out)."

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
zratkai: See other response to this topic.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution_ordered.q
zratkai: Why is there an impact of this PR to change the order of the result of this query?

difin: Because of changes that I did on the lines 35-60 in iceberg_major_compaction_partition_evolution_ordered.q

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Hi @deniskuzZ, can you please provide your view on this topic, should we add parenthesis here or not?
Currently it follows the recommendation from this review comment:
https://github.com/apache/hive/pull/5389#discussion_r1711415417

deniskuzZ: Up to you; it's a matter of style. I haven't seen any recommendations on that topic in either Clean Code or Effective Java.
Only remark it should be easy to read and understand.

````
Even if the operator precedence seems clear to you, it might not be to others—you shouldn’t assume that other programmers know precedence as well as you do.
if (a == b && c == d) // AVOID!
if ((a == b) && (c == d)) // RIGHT"
````
to be honest, that seems like a joke 

````
x && y || z
x && (y || z)
````
are 2 completely different expressions, if you know  boolean algebra

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 7]
```diff
@@ -2232,4 +2233,59 @@ private static List<FieldSchema> schema(List<VirtualColumn> exprs) {
   private static List<FieldSchema> orderBy(VirtualColumn... exprs) {
     return schema(Arrays.asList(exprs));
   }
+
+  @Override
+  public boolean canCompact(HiveConf hiveConf, org.apache.hadoop.hive.ql.metadata.Table table, String partitionPath,
```
deniskuzZ, line 2238: that should be part of StorageHandler API, it's Compactor/Worker responsibility

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergCompactionUtil.java
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Returns true if:
+   *  1. When the table is unpartitioned.
+   *  2. When partitionPath is not provided and the file belongs to the non-latest partition spec.
+   *  3. When partitionPath is provided and the file belongs to the same partition.
+   * @param table the iceberg table
+   * @param partitionPath partition path
+   * @param file Data or Delete file
+   */
+  public static boolean doesFileMatchPartition(Table table, String partitionPath, ContentFile<?> file) {
```
deniskuzZ, line 49: name doesn't reflect the inner logic, any other alternatives?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.iceberg.mr.hive.compaction.evaluator.amoro.TableConfiguration;
+import org.apache.iceberg.mr.hive.compaction.evaluator.amoro.TableFileScanHelper;
+import org.apache.iceberg.mr.hive.compaction.evaluator.amoro.TableFormat;
+import org.apache.iceberg.mr.hive.compaction.evaluator.amoro.TableRuntime;
+import org.apache.iceberg.mr.hive.compaction.evaluator.amoro.TableRuntimeMeta;
+import org.apache.iceberg.util.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CompactionEvaluator {
```
deniskuzZ, line 44: maybe add Iceberg: `IcebergCompactionEvaluator`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
```diff
@@ -0,0 +1,118 @@
...
+        table.currentSnapshot().snapshotId());
+    CommonPartitionEvaluator evaluator = null;
+    try (CloseableIterable<TableFileScanHelper.FileScanResult> results =
+             tableFileScanHelper.scan()) {
+      for (TableFileScanHelper.FileScanResult fileScanResult : results) {
+        DataFile file = fileScanResult.file();
+        if (IcebergCompactionUtil.doesFileMatchPartition(table, partitionPath, file)) {
+          PartitionSpec partitionSpec = table.specs().get(file.specId());
+          Pair<Integer, StructLike> partition = Pair.of(partitionSpec.specId(), fileScanResult.file().partition());
+          evaluator = evaluator == null ? new CommonPartitionEvaluator(tableRuntime, partition, 2) : evaluator;
```
deniskuzZ, line 109: what. is the magic number `2`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
```diff
@@ -0,0 +1,118 @@
...
+      HiveConf conf) {
+    TableRuntime tableRuntime = constructTableRuntime(table, conf);
+
+    TableFileScanHelper tableFileScanHelper = new IcebergTableFileScanHelper(table,
+        table.currentSnapshot().snapshotId());
+    CommonPartitionEvaluator evaluator = null;
+    try (CloseableIterable<TableFileScanHelper.FileScanResult> results =
+             tableFileScanHelper.scan()) {
+      for (TableFileScanHelper.FileScanResult fileScanResult : results) {
+        DataFile file = fileScanResult.file();
```
deniskuzZ, line 105: would it be extended with `deletes` as well?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
```diff
@@ -0,0 +1,118 @@
...
+    tableRuntimeMeta.setTableName(icebergTable.name());
+    tableRuntimeMeta.setFormat(TableFormat.ICEBERG);
+    tableRuntimeMeta.setLastFullOptimizingTime(1);
+    tableRuntimeMeta.setLastMinorOptimizingTime(1);
+    tableRuntimeMeta.setTableConfig(tableConfig);
+
+    return new HiveTableRuntime(tableRuntimeMeta);
+  }
+
+  private static CommonPartitionEvaluator constructCommonPartitionEvaluator(Table table, String partitionPath,
```
deniskuzZ, line 95: maybe use `create` instead of `construct`

deniskuzZ (commented):
LGTM, some minor things

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergCompactionUtil.java
difin: Updated Javadoc of this method to 
```
  /**
   * This method implements a common filter that is used in several places in Iceberg compaction code.
   * Its aim is to determine if the provided file needs to be handled when compacting a partition whose path is equal to 
   * the provided partitionPath. Returns true when one of the following conditions is true, otherwise returns false:
   *  1. table is unpartitioned
   *  2. partitionPath is null and the file belongs to the non-latest partition spec
   *  3. partitionPath is not null and the file belongs to the partition whose path is the partitionPath
   * @param table the iceberg table
   * @param partitionPath partition path
   * @param file Data or Delete file
   */
```

and updated method name to `isHandleFileInCompaction` to better reflect the intent of the method.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
difin: no, it's not needed.
I use `file` only to get `partitionSpec` and to check if this file and its deletes need to be handled when processing the given partition.
If the file does need to be handled with the given partitionPath, I'm adding the data file and the delete files to the compaction evaluator:
`evaluator.addFile(fileScanResult.file(), fileScanResult.deleteFiles());`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
difin: The number `2` is a hardcoded plan time at which we run compaction evaluator.
Amoro compaction evaluator has logic that uses last full optimize time, last minor optimize time and the current plan time. I defined last full optimize time and last minor optimize as `1`. With these hardcoded values the time-based conditions in Amoro compaction evaluator are always true and the results of the evaluation depends on other metrics.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
difin: Done

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -2229,12 +2229,18 @@ public static enum ConfVars {
...
     HIVE_ICEBERG_ALLOW_DATAFILES_IN_TABLE_LOCATION_ONLY("hive.iceberg.allow.datafiles.in.table.location.only", false,
         "If this is set to true, then all the data files being read should be withing the table location"),
-
+    HIVE_ICEBERG_MAJOR_COMPACTION_FILE_SIZE_THRESHOLD("hive.iceberg.major.compaction.file.size.threshold", "96mb",
+        new SizeValidator(), "Iceberg data file size in megabytes below which a file needs to be compacted."),
+    HIVE_ICEBERG_MINOR_COMPACTION_FILE_SIZE_THRESHOLD("hive.iceberg.minor.compaction.file.size.threshold", "16mb",
+        new SizeValidator(), "Iceberg data file size in megabytes below which a file needs to be compacted."),
+    ICEBERG_COMPACTION_FILE_SIZE_RATIO("hive.iceberg.compaction.file.size.ratio", 0.1f,
+        "Ratio of # data files below threshold / # data files above threshold above which compaction is needed"),
+    ICEBERG_COMPACTION_DELETE_RECORDS_THRESHOLD("hive.iceberg.delete.records.threshold", 100,
```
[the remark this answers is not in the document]
difin: I removed this from Conf, not sure it's needed with Amoro compaction evaluator.
For now it's not used, I think we can add it later when needed.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergCompactionUtil.java
deniskuzZ: still not great, what if we name it `isEligibleForCompaction` or `shouldIncludeForCompaction`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
deniskuzZ: would it make sence to use `System.currentTimeMillis()` everywhere istead of this constants?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergCompactionUtil.java
difin: `shouldIncludeForCompaction` seems better to me

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/CompactionEvaluator.java
difin: I think `System.currentTimeMillis()` makes sense for the current plan time, but for last optimize time we can use something hardcoded, because last optimize time was not done at `System.currentTimeMillis()` time.

if we use `System.currentTimeMillis()` as the last optimize time then it will be same as current plan time which is incorrect.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/IcebergCompactionEvaluator.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.iceberg.mr.hive.compaction.evaluator.amoro.TableRuntime;
+import org.apache.iceberg.mr.hive.compaction.evaluator.amoro.TableRuntimeMeta;
+import org.apache.iceberg.util.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IcebergCompactionEvaluator {
+
+  private static final long lastOptimizeTime = 0;
+  private static final int triggerInterval = 0;
```
deniskuzZ, line 47: constants should be in uppercase, sonar reports it as a major issue

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/IcebergCompactionEvaluator.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  private static TableRuntime createTableRuntime(Table icebergTable, HiveConf conf) {
+    long fileSizeInBytesThreshold = HiveConf.getSizeVar(conf,
+        HiveConf.ConfVars.HIVE_ICEBERG_COMPACTION_FILE_SIZE_THRESHOLD);
+
+    OptimizingConfig optimizingConfig = OptimizingConfig.parse(Collections.emptyMap());
+    optimizingConfig.setTargetSize(fileSizeInBytesThreshold);
+    optimizingConfig.setFullTriggerInterval(triggerInterval);
```
deniskuzZ, line 86: is that required since we trigger it manually or are there some internal checks?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/IcebergCompactionEvaluator.java
difin: Done.
Initially I had them in uppercase and Iceberg stylechecking was failing because of upper case. Works now.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/evaluator/IcebergCompactionEvaluator.java
difin: Amoro compaction evaluator has logic that depends on trigger intervals. It can refuse compaction requests based on them.
Setting trigger interval because defaults ones have the following values:
```
public static final int SELF_OPTIMIZING_MINOR_TRIGGER_INTERVAL_DEFAULT = 3600000; // 1 h
public static final int SELF_OPTIMIZING_FULL_TRIGGER_INTERVAL_DEFAULT = -1; // not trigger
```

Regarding overriding target size config, I think we need the ability to override the default value in order to use FILE_SIZE_THRESHOLD clause in a compaction command.

deniskuzZ (approved):
+1

PR #5531, opened 2024-11-05
HIVE-28609: HiveSequenceFileInputFormat should be cloned or not be cached

abstractdog:
Introduce an InputFormatCache in hive and handle the caching behavior of the input formats.

The code became cleaner and simpler by removing a ThreadLocal from a specific input format (HiveSequenceFileInputFormat) and separating the caching logic.

No.

No.

Precommit testing.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/InputFormatCache.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/InputFormatCache.java
@@ -44,9 +44,9 @@ public class InputFormatCache {
   /**
    * A cache of InputFormat instances.
    */
-  private static final Map<String, InputFormat> inputFormats = new HashMap<String, InputFormat>();
+  private static final Map<String, InputFormat<?, ?>> inputFormats = new HashMap<>();
 
-  public static InputFormat getInputFormatFromCache(
+  public static InputFormat getInputFormat(
       Class<? extends InputFormat> inputFormatClass, Configuration conf) throws IOException {
     if (Configurable.class.isAssignableFrom(inputFormatClass) ||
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/exec/InputFormatCache.java
```diff
@@ -0,0 +1,75 @@
...
+          inputFormats.put(inputFormatClass.getName(), format);
+        }
+      } catch (Exception e) {
+        throw new IOException("Cannot create an instance of InputFormat class: " + inputFormatClass.getName(), e);
+      }
+    }
+    if (CacheableInputFormat.class.isAssignableFrom(inputFormatClass)) {
+      format = ((CacheableInputFormat)format).newInstance(format);
+    }
+    return format;
```
SourabhBadhya, line 73: Ok so this logic itself seems to always create a new instance of HiveSequenceFileInputFormat whenever the inputFormatClass is HiveSequenceFileInputFormat?

SourabhBadhya (approved):
LGTM +1. @abstractdog I have one question, however I understand your logic.

ql/src/java/org/apache/hadoop/hive/ql/exec/InputFormatCache.java
abstractdog: yes, exactly, with this approach, I tried to achieve all that we had achieved previously, so to prevent fileStatuses contention as fixed in [HIVE-28530](https://issues.apache.org/jira/browse/HIVE-28530), but at the same time, moved the control of the caching behavior to the input format classes themselves (because they are the only ones being aware if caching them is safe or not)

PR #5534, opened 2024-11-07
HIVE-28616: OrcReader is not closed in OrcEncodedDataReader

abstractdog:
OrcEncodedDataReader fully owns the OrcReader instance it creates, so it should be closed there too

It causes connection leaks in case of FileSystems like S3 (where an HTTP connection pool is utilized)

No.

No.

Tested on cluster.

### DIFF
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
@@ -956,7 +956,6 @@ public class OrcEncodedDataReader extends CallableWithNdc<Void>
   private class DataWrapperForOrc implements LlapDataReader, DataCache, BufferObjectFactory {
     /** A reference to parent DataReader not owned by this object. */
     private final LlapDataReader orcDataReaderRef;
-    private boolean isOpen;
 
     public DataWrapperForOrc() throws IOException {
       ensureRawDataReader(false);
@@ -1024,7 +1023,7 @@ public class OrcEncodedDataReader extends CallableWithNdc<Void>
 
     @Override
     public boolean isOpen() {
-      return isOpen;
+      return orcDataReaderRef.isOpen();
     }
 
     @Override
@@ -1061,7 +1060,6 @@ public class OrcEncodedDataReader extends CallableWithNdc<Void>
       long startTime = counters.startTimeCounter();
       orcDataReaderRef.open();
       counters.recordHdfsTime(startTime);
-      isOpen = true;
     }
 
     @Override
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
@@ -275,15 +275,20 @@ public class OrcEncodedDataReader extends CallableWithNdc<Void>
 
   @Override
   protected Void callInternal() throws IOException, InterruptedException {
-    return ugi.doAs(new PrivilegedExceptionAction<Void>() {
-      @Override
-      public Void run() throws Exception {
-        try {
-          return performDataRead();
-        } finally {
-          cleanupReaders();
-        }
+    return ugi.doAs((PrivilegedExceptionAction<Void>) () -> {
+      long startTime = counters.startTimeCounter();
+      try {
+        performDataRead();
+        consumer.setDone();
+      } catch (Throwable t) {
+        consumer.setError(t);
+        trace.dumpLog(LOG);
+      } finally {
+        cleanupReaders();
+        recordReaderTime(startTime);
+        tracePool.offer(trace);
       }
+      return null;
     });
   }
 
@@ -298,31 +303,22 @@ public class OrcEncodedDataReader extends CallableWithNdc<Void>
     };
   }
 
-  protected Void performDataRead() throws IOException, InterruptedException {
-    long startTime = counters.startTimeCounter();
+  protected void performDataRead() throws IOException, InterruptedException {
     LlapIoImpl.LOG.info("Processing data for file {}: {}", fileKey, split.getPath());
     if (processStop()) {
-      recordReaderTime(startTime);
-      return null;
+      return;
     }
     counters.setDesc(QueryFragmentCounters.Desc.TABLE, cacheTag.getTableName());
     counters.setDesc(QueryFragmentCounters.Desc.FILE, split.getPath()
         + (fileKey == null ? "" : " (" + fileKey + ")"));
-    try {
-      validateFileMetadata();
 
-      // 2. Determine which stripes to read based on the split.
-      determineStripesToRead();
-    } catch (Throwable t) {
-      handleReaderError(startTime, t);
-      return null;
-    }
+    validateFileMetadata();
+
+    // 2. Determine which stripes to read based on the split.
+    determineStripesToRead();
 
     if (stripeRgs.length == 0) {
-      consumer.setDone();
-      recordReaderTime(startTime);
-      tracePool.offer(trace);
-      return null; // No data to read.
+      return;
     }
     counters.setDesc(QueryFragmentCounters.Desc.STRIPES,
         stripeIxFrom + "," + stripeRgs.length);
@@ -330,51 +326,37 @@ public class OrcEncodedDataReader extends CallableWithNdc<Void>
     // 3. Apply SARG if needed, and otherwise determine what RGs to read.
     int stride = fileMetadata.getRowIndexStride();
     ArrayList<OrcStripeMetadata> stripeMetadatas = null;
-    try {
-      if (sarg != null && stride != 0) {
-        // TODO: move this to a common method
-        // Note: this gets IDs by name, so we assume indices don't need to be adjusted for ACID.
-        int[] filterColumns = RecordReaderImpl.mapSargColumnsToOrcInternalColIdx(
-          sarg.getLeaves(), evolution);
-        // included will not be null, row options will fill the array with trues if null
-        sargColumns = new boolean[evolution.getFileSchema().getMaximumId() + 1];
-        for (int i : filterColumns) {
-          // filter columns may have -1 as index which could be partition column in SARG.
-          // TODO: should this then be >=?
-          if (i > 0) {
-            sargColumns[i] = true;
-          }
+    if (sarg != null && stride != 0) {
+      // TODO: move this to a common method
+      // Note: this gets IDs by name, so we assume indices don't need to be adjusted for ACID.
+      int[] filterColumns = RecordReaderImpl.mapSargColumnsToOrcInternalColIdx(
+        sarg.getLeaves(), evolution);
+      // included will not be null, row options will fill the array with trues if null
+      sargColumns = new boolean[evolution.getFileSchema().getMaximumId() + 1];
+      for (int i : filterColumns) {
+        // filter columns may have -1 as index which could be partition column in SARG.
+        // TODO: should this then be >=?
+        if (i > 0) {
+          sargColumns[i] = true;
         }
-
-        // If SARG is present, get relevant stripe metadata from cache or readers.
-        stripeMetadatas = readStripesMetadata(fileIncludes, sargColumns);
       }
 
-      // Now, apply SARG if any; w/o sarg, this will just initialize stripeRgs.
-      boolean hasData = determineRgsToRead(stride, stripeMetadatas);
-      if (!hasData) {
-        consumer.setDone();
-        recordReaderTime(startTime);
-        tracePool.offer(trace);
-        return null; // No data to read.
-      }
-    } catch (Throwable t) {
-      handleReaderError(startTime, t);
-      return null;
+      // If SARG is present, get relevant stripe metadata from cache or readers.
+      stripeMetadatas = readStripesMetadata(fileIncludes, sargColumns);
+    }
+
+    // Now, apply SARG if any; w/o sarg, this will just initialize stripeRgs.
+    boolean hasData = determineRgsToRead(stride, stripeMetadatas);
+    if (!hasData) {
+      return; // No data to read.
     }
 
     if (processStop()) {
-      recordReaderTime(startTime);
-      return null;
+      return;
     }
 
     // 4. Create encoded data reader.
-    try {
-      ensureDataReader();
-    } catch (Throwable t) {
-      handleReaderError(startTime, t);
-      return null;
-    }
+    ensureDataReader();
 
     // 6. Read data.
     // TODO: I/O threadpool could be here - one thread per stripe; for now, linear.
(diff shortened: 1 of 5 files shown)


llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -1072,7 +1073,9 @@ public Allocator getAllocator() {
 
     @Override
     public void close() throws IOException {
-      // Noop: orcDataReaderRef is owned by the parent object
+      if (orcReader != null){
```
zratkai, line 1076: Shouldn't we call cleanupReaders() here?

abstractdog: makes sense, let me check

deniskuzZ: it seems that was an intentional change, see HIVE-18933 : disable ORC codec pool for now; remove clone

abstractdog: yikes, in this I need to figure out whose parent is who :) ... and close from the parent, because while cluster testing, a socket leak was removed by closing the orc reader, so there is a problem with the current implementation

abstractdog: okay, uploaded new commit 
https://github.com/apache/hive/pull/5534/commits/df13c21aced3839aa3b532bdf5159203f4bd7610

tested on cluster, problem solved
looks like the cleanupReaders method is called as expected, so closing the orcReader there works

need to note that there were 2 different stacks where the leakage happened, @harishjp  is attaching them to jira as we speak

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -531,6 +531,15 @@ private void cleanupReaders() {
         // Ignore.
       }
     }
+
+    if (orcReader != null) {
```
deniskuzZ, line 535: I might be wrong, but I think that is not right. Based on the code we reuse `orcReader` instance in multiple splits and create `stripeReader` per split which is closed in `cleanupReaders`.

see HIVE-21005: LLAP: Reading more stripes per-split leaks ZlibCodecs

deniskuzZ: code is very strange. why do we create new `stripeReader` per stripe, can't we reuse?

abstractdog: it looks like we close everything else in cleanupReaders, which is called from performDataRead
by closing this, are you suspecting a failure, or a perf degradation (meaning that this reader will be reopened any time if it's found to be null)
(let me check the lifecycle of the OrcEncodedDataReader in the meantime more thoroughly)

deniskuzZ: maybe we should call `close` in a `call` method before we assign `orcReader = null`?

deniskuzZ: also would be great if we manage to create Hive test with repro

zratkai:
In private void cleanupReaders() {} the  rawDataReader is closed, but isRawDataReaderOpen is not set to false.

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
deniskuzZ: ok, looks like it's fine. 
we should close the reader in `getOrcTailForPath` as well. maybe use try-with-resources

deniskuzZ: also please refactor to use `IOUtils.closeQuietly(orcReader)`

deniskuzZ:
> In private void cleanupReaders() {} the rawDataReader is closed, but isRawDataReaderOpen is not set to false.

i don't think it matters cause we basically destroying `OrcEncodedDataReader` there

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -838,6 +847,7 @@ private void ensureRawDataReader(boolean isOpen) throws IOException {
       if (!isRawDataReaderOpen && isOpen) {
         long startTime = counters.startTimeCounter();
         rawDataReader.open();
+        isRawDataReaderOpen = true;
```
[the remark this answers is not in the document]
deniskuzZ: what i meant
````
if (!rawDataReader.isOpen() && isOpen) {
  rawDataReader.open();
}
````
but that's more like refactor, so can't be skipped here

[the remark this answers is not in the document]
Comment (abstractdog) [reply llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java]: Should we close the reader regardless of isRawDataReaderOpen
yeah, any problem will be caught as an IOException here and can continue cleanup

File: llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java (hunk 2 above)
Comment (abstractdog) [reply llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java]: yeah, I was thinking about making all the readers implement Closeable and use IOUtils.closeQuietly, so the question is where to stop refactoring the messy thing now :)

File: llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java [hunk 4]
```diff
@@ -1109,6 +1122,7 @@ public void open() throws IOException {
       long startTime = counters.startTimeCounter();
       orcDataReaderRef.open();
       counters.recordHdfsTime(startTime);
+      isOpen = true;
```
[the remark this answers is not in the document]
abstractdog: for clarity's sake  at least, yeah

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -523,14 +528,17 @@ private void cleanupReaders() {
...
+    if (rawDataReader != null) {
       try {
         rawDataReader.close();
         rawDataReader = null;
       } catch (IOException ex) {
         // Ignore.
       }
     }
+
+    IOUtils.closeQuietly(orcReader);
```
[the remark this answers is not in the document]
abstractdog: kind of agree, but INFO level still looks high, what's the desired log level here in your opinion?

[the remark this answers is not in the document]
harishjp: This would log only on close failures, which I'm assuming should not happen normally. I would set the level to be as high as WARN for this.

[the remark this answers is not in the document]
deniskuzZ: there is overloaded `closeQuietly` with the consumer

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -523,14 +528,17 @@ private void cleanupReaders() {
         // Ignore.
       }
     }
-    if (rawDataReader != null && isRawDataReaderOpen) {
+    if (rawDataReader != null) {
```
deniskuzZ, line 531: no need for null check
````
IOUtils.closeQuietly(rawDataReader);
rawDataReader = null;
````

File: llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java (hunk 4 above)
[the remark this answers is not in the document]
Comment (deniskuzZ) [reply llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java:1125]: we should use 
````
    public boolean isOpen() {
      return orcDataReaderRef.isOpen();
    }
````

File: llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java (hunk 2 above)
Comment (deniskuzZ) [reply llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java]: :) I've added same comment

Comment (harishjp) [reply llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java]: stop gets invoked in LlapRecordReader.close, but since its another thread, we cannot close there so it has to be done in callInternal (same thread), and since callInternal is only invoked once per OrcEncodedDataReader(confirmed by Denys), changing code to do this in a try finally inside callInternal.

File: llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java (hunk 4 above)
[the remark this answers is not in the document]
Comment (harishjp) [reply llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java]: Thanks, done.

File: llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java (hunk 6 above)
Comment (harishjp) [reply llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java]: Nice thanks! Done, changed all other classes/interfaces to be Closeable too.

File: llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java [hunk 7]
```diff
@@ -276,11 +275,20 @@ public void unpause() {
...
   protected Void callInternal() throws IOException, InterruptedException {
-    return ugi.doAs(new PrivilegedExceptionAction<Void>() {
-      @Override
-      public Void run() throws Exception {
-        return performDataRead();
+    return ugi.doAs((PrivilegedExceptionAction<Void>) () -> {
+      long startTime = counters.startTimeCounter();
+      try {
+        performDataRead();
+        consumer.setDone();
```
deniskuzZ, line 282: keep it in `performDataRead`

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -276,11 +275,20 @@ public void unpause() {
...
+      } catch (Throwable t) {
+        consumer.setError(t);
```
deniskuzZ, line 284: reuse `handleReaderError`

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -494,8 +451,6 @@ private void validateFileMetadata() throws IOException {
   private boolean processStop() {
```
deniskuzZ, line 451: why do we need this method, we can just use `!isStopped.get()`

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
harishjp: There are way too many return and anyone trying changing code should remember to do this correctly. This code refactors all the different call site and unifies handling error, timing and trace to just one place. Now if you see performDataRead is much much simpler.

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
harishjp: I deleted handleReaderError, this is the only place all handling happens.

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -276,11 +275,20 @@ public void unpause() {
...
+      long startTime = counters.startTimeCounter();
+      try {
+        performDataRead();
+        consumer.setDone();
+      } catch (Throwable t) {
+        consumer.setError(t);
+        trace.dumpLog(LOG);
+      } finally {
+        cleanupReaders();
+        recordReaderTime(startTime);
```
deniskuzZ, line 288: should this be the 1st step?

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
deniskuzZ: not sure i get this, `orcReader` is created in LlapRecordReader constructor

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
deniskuzZ: ok

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
deniskuzZ: ok

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
harishjp: I do not know, that is confusing for me too. If you look at previous code, in some case recordReaderTime is called before cleanupReaders (Ex: end of function and handleError) and in some cases its called after (Ex: processStop). I do not think it was intentional. I can do anything which seems correct, cleanupReader is closing IO objects, whether that part of IO or not is subjective. I can do it the way you feel strongly.

deniskuzZ: if that is supposed to profile the read time, it should probably be before the cleanup

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -516,21 +471,13 @@ private static Object determineFileId(Supplier<FileSystem> fsSupplier, FileSplit
...
-        rawDataReader = null;
-      } catch (IOException ex) {
-        // Ignore.
-      }
+    try {
+      IOUtils.close(stripeReader, rawDataReader, orcReader);
+    } catch (IOException e) {
+      LOG.warn("Error while closing readers: ", e);
     }
+    rawDataReader = null;
```
deniskuzZ, line 479: what about `stripeReader` should we set it to null as well

deniskuzZ (dismissed):
LGTM, some minor stuff, pending tests

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
deniskuzZ: @harishjp, we should close in `stop`! What if we don't reach the call method, orcReader would be still open

harishjp: Yes you are right, that was one of the things I handled earlier with the AtomicInteger with state. I'll add that again then :-(.

deniskuzZ: why do you need new state? there won't be 2 concurrent threads, just close the reader

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
harishjp: Done!

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
harishjp: Done

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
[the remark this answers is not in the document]
harishjp: Fixed this case too, PTAL.

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
harishjp: The implementation seems to be written such that stop() can be called when callInternal is executing, please look at how isStopped is being sent as AtomicBoolean and isStopped checks are in multiple places during callInternal, isn't that right? If yes then we need the new state, how else do you know that stop() is not being called when callInternal is executing?

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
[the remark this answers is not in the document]
abstractdog: Post-Traumatic Activity Level?

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -187,7 +188,17 @@ public Pool<OrcEncodedColumnBatch> createEncodedColumnBatchPool() {
    * Contains only stripes that are read, and only columns included. null => read all RGs.
    */
   private boolean[][] stripeRgs;
-  private AtomicBoolean isStopped = new AtomicBoolean(false);
+
+  private static final int STATE_IDLING = 0;
```
deniskuzZ, line 192: `STATE_IDLE`

deniskuzZ (approved):
LGTM +1

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
deniskuzZ: yes, looks like `isStopped` is used for thread interruption (i.e. `processStop`), but we are more interested when we can do the resource cleanup.

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
```diff
@@ -295,83 +318,60 @@ static Supplier<FileSystem> getFsSupplier(final Path path,
     };
   }
 
-  protected Void performDataRead() throws IOException, InterruptedException {
-    long startTime = counters.startTimeCounter();
+  protected void performDataRead() throws IOException, InterruptedException {
```
deniskuzZ, line 321: remove `InterruptedException `

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
abstractdog: ack, fixing this

llap-server/src/java/org/apache/hadoop/hive/llap/io/encoded/OrcEncodedDataReader.java
abstractdog: ack

abstractdog:
latest comment are addressed in: https://github.com/apache/hive/pull/5534/commits/6da66f618cad2a08e1d3f2e97592f435dd324600

waiting for a green test

PR #5537, opened 2024-11-12
HIVE-28620: Query result is cached in case of IOWD if the subquery is not trivial

abstractdog:


ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -15840,25 +15849,34 @@ private QueryResultsCache.QueryInfo createCacheQueryInfoForQuery(QueryResultsCac
...
       return false;
     }
     // HIVE-19096 - disable for explain and explain analyze
-    return ctx.getExplainAnalyze() == null;
+    if (ctx.getExplainAnalyze() != null) {
+      LOG.debug("Query type cannot use cache (explain analyze command)");
+      return false;
+    }
+
+    LOG.debug("Query type can use cache");
```
deniskuzZ, line 15878: i think this log message is redundant

abstractdog: agreed, doesn't add more value

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -12267,32 +12267,41 @@ private void setupStats(TableScanDesc tsDesc, QBParseInfo qbp, Table tab, String
     }
   }
 
-  private Operator genPlan(QB parent, QBExpr qbexpr) throws SemanticException {
+  private Operator genPlan(QB rootQB, QB parent, QBExpr qbexpr) throws SemanticException {
```
deniskuzZ, line 12270: that is confusing: rootQB & parent QB

abstractdog:
> It seems to me that this could be done with fewer code changes, however, I could be completely wrong, but to be safe I think we should ask review from @kasakrisz can't we pass the needed flags from the rootQB to qb?

I was thinking about the same
there are flags that should not change across the different levels, like the one that this patch is about: isInsertOverwriteDir
if it's set once, it applies to the whole query: there is no such thing as, some query blocks suggest that the query is IOWD, while others don't
how to implement this properly in this SemanticAnalyzer monster...is a completely different question :D open to ideas

abstractdog:
> > It seems to me that this could be done with fewer code changes, however, I could be completely wrong, but to be safe I think we should ask review from @kasakrisz can't we pass the needed flags from the rootQB to qb?
> 
> I was thinking about the same there are flags that should not change across the different levels, like the one that this patch is about: isInsertOverwriteDir if it's set once, it applies to the whole query theoretically: there is no such thing as "some query blocks suggest that the query is IOWD, while others don't" how to implement this properly in this SemanticAnalyzer monster...is a completely different question :D open to ideas
> 
> I believe the worst thing that's currently happening is this:
> 
> https://github.com/apache/hive/blob/e48cb8e986ec254eb44dff684659dcdb8bf118e8/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java#L12435
> 
> 
> a field variable is changed while running a recursive method, with all due respect, but without knowing the context, this looks awful :(
> however, I wasn't brave enough to touch it for the sake of this fix

wait, maybe I have to revise what I said
```
      Operator<?> operator = genPlan(qb, qbexpr);
```
looks like the original qb is passed to the method, so I can reuse it, let me try :D

abstractdog:
pushed https://github.com/apache/hive/pull/5537/commits/392c9b9b5aa71a730c618604c330e94e9fadb5a4
thanks for the good catch @deniskuzZ , looks like almost all the signature changes were unnecessary, except where the actual issue happened: getDestinationFilePath, where this.qb was silently used
I also changed qb param to qbParam in genPlan, it's dangerous that a param can hide the field in java, especially in this context

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -12292,10 +12292,10 @@ Operator genPlan(QB qb) throws SemanticException {
   }
 
   @SuppressWarnings("nls")
-  private Operator genPlan(QB qb, boolean skipAmbiguityCheck)
+  private Operator genPlan(QB qpParam, boolean skipAmbiguityCheck)
```
kasakrisz, line 12295: `qpParam` or `qbParam`?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -12405,14 +12405,14 @@ private Operator genPlan(QB qb, boolean skipAmbiguityCheck)
...
-          mergeJoinTree(qb);
+          mergeJoinTree(qpParam);
         }
       }
 
       // if any filters are present in the join tree, push them on top of the
       // table
-      pushJoinFilters(qb, qb.getQbJoinTree(), aliasToOpInfo);
-      srcOpInfo = genJoinPlan(qb, aliasToOpInfo);
+      pushJoinFilters(qpParam, qpParam.getQbJoinTree(), aliasToOpInfo);
```
kasakrisz, line 12414: Passing `qpParam.getQbJoinTree()` is not necessary if `qpParam` is also passed.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -12292,10 +12292,10 @@ Operator genPlan(QB qb) throws SemanticException {
   }
 
   @SuppressWarnings("nls")
-  private Operator genPlan(QB qb, boolean skipAmbiguityCheck)
+  private Operator genPlan(QB qbParam, boolean skipAmbiguityCheck)
```
deniskuzZ, line 12295: minor: i don't think renaming has some advantages here, however, could create conflicts during backport.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -15840,25 +15840,33 @@ private QueryResultsCache.QueryInfo createCacheQueryInfoForQuery(QueryResultsCac
   /**
    * Some initial checks for a query to see if we can look this query up in the results cache.
    */
-  private boolean queryTypeCanUseCache() {
-    if (this.qb == null || this.qb.getParseInfo() == null) {
+  private boolean queryTypeCanUseCache(QB qbParam) {
```
deniskuzZ, line 15843: minor: I would have used simply `qb`. I know it hides the instance param, but maybe we could refactor similar stuff in diff PR

deniskuzZ:
sonar reported number of new findings

abstractdog:
> sonar reported number of new findings

checked, most of them are not introduced now, I was only able to fix one exception in the unit test

abstractdog:
is this ready for a +1 guys? :)

kasakrisz (approved):

deniskuzZ (approved):
LGTM +1

PR #5539, opened 2024-11-12
HIVE-28601: Leverage configurable getPartitions API in HMS to decrease memory footprint in HS2

armitage420:

Leverage configurable getPartitionsWithSpecs API in HMS to decrease memory footprint in HS2

The primary motivation behind these changes is to optimize the performance of certain API calls within HS2. These calls often require specific elements, such as say location, without needing the complete storage descriptor in partition object. Transmitting the full object in these cases results in unnecessary overhead and resource consumption. By leveraging the getPartitionsWithSpecs API, we can fetch only the necessary Partition's sub-objects, thereby reducing the memory footprint and hence reducing unnecessary network usage. (A document is added with the jira for more details on how the APIs work, and performance impact it can make)

No

No

Provided tests and added JUnit test class

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java
@@ -22,6 +22,9 @@ import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.metastore.api.GetPartitionsRequest;
+import org.apache.hadoop.hive.metastore.api.GetProjectionsSpec;
+import org.apache.hadoop.hive.metastore.client.builder.GetPartitionProjectionsSpecBuilder;
 import org.apache.hadoop.hive.ql.metadata.Hive;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.metadata.Partition;
@@ -33,6 +36,7 @@ import org.slf4j.LoggerFactory;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
@@ -89,7 +93,18 @@ public abstract class ShowTableStatusFormatter {
     List<Path> locations = new ArrayList<Path>();
     if (table.isPartitioned()) {
       if (partition == null) {
-        for (Partition currPartition : db.getPartitions(table)) {
+      List<Partition> partitions;
+      GetProjectionsSpec getProjectionsSpec = new GetPartitionProjectionsSpecBuilder()
+              .addProjectFieldList(Arrays.asList("sd.location")).build();
+        GetPartitionsRequest request = new GetPartitionsRequest(table.getDbName(), table.getTableName(), getProjectionsSpec, null);
+        request.setCatName(table.getCatName());
+        try {
+          partitions = db.getPartitionsWithSpecs(table, request);
+        } catch (Exception e) {
+          throw new HiveException(e);
+        }
+
+        for (Partition currPartition : partitions) {
           if (currPartition.getLocation() != null) {
             locations.add(new Path(currPartition.getLocation()));
           }
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/exchange/AlterTableExchangePartitionAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/exchange/AlterTableExchangePartitionAnalyzer.java
@@ -18,11 +18,18 @@
 
 package org.apache.hadoop.hive.ql.ddl.table.partition.exchange;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.hadoop.hive.common.TableName;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
+import org.apache.hadoop.hive.metastore.api.GetPartitionsFilterSpec;
+import org.apache.hadoop.hive.metastore.api.GetPartitionsRequest;
+import org.apache.hadoop.hive.metastore.api.GetProjectionsSpec;
+import org.apache.hadoop.hive.metastore.api.PartitionFilterMode;
+import org.apache.hadoop.hive.metastore.client.builder.GetPartitionProjectionsSpecBuilder;
 import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
 import org.apache.hadoop.hive.ql.ErrorMsg;
 import org.apache.hadoop.hive.ql.QueryState;
@@ -77,9 +84,20 @@ public  class AlterTableExchangePartitionAnalyzer extends AbstractAlterTableAnal
     if (AcidUtils.isTransactionalTable(sourceTable) || AcidUtils.isTransactionalTable(destTable)) {
       throw new SemanticException(ErrorMsg.EXCHANGE_PARTITION_NOT_ALLOWED_WITH_TRANSACTIONAL_TABLES.getMsg());
     }
+    List<String> sourceProjectFilters = MetaStoreUtils.getPvals(sourceTable.getPartCols(), partitionSpecs);
 
     // check if source partition exists
-    PartitionUtils.getPartitions(db, sourceTable, partitionSpecs, true);
+    GetPartitionsFilterSpec sourcePartitionsFilterSpec = new GetPartitionsFilterSpec();
+    sourcePartitionsFilterSpec.setFilters(sourceProjectFilters);
+    sourcePartitionsFilterSpec.setFilterMode(PartitionFilterMode.BY_VALUES);
+
+    GetProjectionsSpec getProjectionsSpec = new GetPartitionProjectionsSpecBuilder()
+        .addProjectFieldList(Arrays.asList("values")).build();
+
+    GetPartitionsRequest request = new GetPartitionsRequest(sourceTable.getDbName(), sourceTable.getTableName(),
+        getProjectionsSpec, sourcePartitionsFilterSpec);
+    request.setCatName(sourceTable.getCatName());
+    PartitionUtils.getPartitionsWithSpecs(db, sourceTable, request, true);
 
     // Verify that the partitions specified are continuous
     // If a subpartition value is specified without specifying a partition's value then we throw an exception
@@ -88,13 +106,23 @@ public  class AlterTableExchangePartitionAnalyzer extends AbstractAlterTableAnal
       throw new SemanticException(ErrorMsg.PARTITION_VALUE_NOT_CONTINUOUS.getMsg(partitionSpecs.toString()));
     }
 
+    List<String> destProjectFilters = MetaStoreUtils.getPvals(destTable.getPartCols(), partitionSpecs);
+
+    // check if dest partition exists
+    GetPartitionsFilterSpec getDestPartitionsFilterSpec = new GetPartitionsFilterSpec();
+    getDestPartitionsFilterSpec.setFilters(destProjectFilters);
+    getDestPartitionsFilterSpec.setFilterMode(PartitionFilterMode.BY_VALUES);
+
     List<Partition> destPartitions = null;
+    GetPartitionsRequest destRequest = new GetPartitionsRequest(destTable.getDbName(), destTable.getTableName(),
+        getProjectionsSpec, getDestPartitionsFilterSpec);
+    destRequest.setCatName(destTable.getCatName());
     try {
-      destPartitions = PartitionUtils.getPartitions(db, destTable, partitionSpecs, true);
+      destPartitions = PartitionUtils.getPartitionsWithSpecs(db, destTable, destRequest, true);
     } catch (SemanticException ex) {
       // We should expect a semantic exception being throw as this partition should not be present.
     }
-    if (destPartitions != null) {
+    if (CollectionUtils.isNotEmpty(destPartitions)) {
       // If any destination partition is present then throw a Semantic Exception.
       throw new SemanticException(ErrorMsg.PARTITION_EXISTS.getMsg(destPartitions.toString()));
     }
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/repl/ReplExternalTables.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/repl/ReplExternalTables.java
@@ -23,7 +23,10 @@ import org.apache.hadoop.hdfs.DistributedFileSystem;
 import org.apache.hadoop.hive.common.FileUtils;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.TableType;
+import org.apache.hadoop.hive.metastore.api.GetPartitionsRequest;
+import org.apache.hadoop.hive.metastore.api.GetProjectionsSpec;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
+import org.apache.hadoop.hive.metastore.client.builder.GetPartitionProjectionsSpecBuilder;
 import org.apache.hadoop.hive.metastore.utils.StringUtils;
 import org.apache.hadoop.hive.ql.ErrorMsg;
 import org.apache.hadoop.hive.ql.exec.repl.util.FileList;
@@ -36,6 +39,7 @@ import org.apache.hadoop.hive.ql.metadata.Table;
 import org.apache.hadoop.hive.ql.parse.SemanticException;
 import org.apache.hadoop.hive.ql.parse.repl.PathBuilder;
 import org.apache.hadoop.hive.ql.parse.repl.dump.Utils;
+import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -43,6 +47,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -97,15 +102,20 @@ public class ReplExternalTables {
     }
     if (table.isPartitioned()) {
       List<Partition> partitions;
+      GetProjectionsSpec projectionSpec = new GetPartitionProjectionsSpecBuilder()
+              .addProjectFieldList(Arrays.asList("sd.location")).build();
+      GetPartitionsRequest request = new GetPartitionsRequest(table.getDbName(), table.getTableName(),
+              projectionSpec, null);
+      request.setCatName(table.getCatName());
       try {
-        partitions = Hive.get(hiveConf).getPartitions(table);
-      } catch (HiveException e) {
+        partitions = Hive.get(hiveConf).getPartitionsWithSpecs(table, request);
+      } catch (HiveException | TException e) {
         if (e.getCause() instanceof NoSuchObjectException) {
           // If table is dropped when dump in progress, just skip partitions data location dump
           LOG.debug(e.getMessage());
           return;
         }
-        throw e;
+        throw new HiveException(e);
       }
 
       for (Partition partition : partitions) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -118,10 +118,13 @@ import org.apache.hadoop.hive.metastore.api.CompactionRequest;
 import org.apache.hadoop.hive.metastore.api.CreateTableRequest;
 import org.apache.hadoop.hive.metastore.api.GetFunctionsRequest;
 import org.apache.hadoop.hive.metastore.api.GetPartitionsByNamesRequest;
+import org.apache.hadoop.hive.metastore.api.GetPartitionsRequest;
+import org.apache.hadoop.hive.metastore.api.GetPartitionsResponse;
 import org.apache.hadoop.hive.metastore.api.GetTableRequest;
 import org.apache.hadoop.hive.metastore.api.SourceTable;
 import org.apache.hadoop.hive.metastore.api.UpdateTransactionalStatsRequest;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
+import org.apache.hadoop.hive.metastore.Batchable;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.metastore.utils.RetryUtilities;
 import org.apache.hadoop.hive.ql.Context;
@@ -162,6 +165,7 @@ import org.apache.hadoop.hive.metastore.api.GetOpenTxnsInfoResponse;
 import org.apache.hadoop.hive.metastore.api.GetPartitionNamesPsRequest;
 import org.apache.hadoop.hive.metastore.api.GetPartitionNamesPsResponse;
 import org.apache.hadoop.hive.metastore.api.GetPartitionRequest;
+import org.apache.hadoop.hive.metastore.api.GetPartitionsFilterSpec;
 import org.apache.hadoop.hive.metastore.api.GetPartitionResponse;
 import org.apache.hadoop.hive.metastore.api.GetPartitionsPsWithAuthRequest;
 import org.apache.hadoop.hive.metastore.api.GetPartitionsPsWithAuthResponse;
@@ -177,6 +181,7 @@ import org.apache.hadoop.hive.metastore.api.Materialization;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.MetadataPpdResult;
 import org.apache.hadoop.hive.metastore.api.NotNullConstraintsRequest;
+import org.apache.hadoop.hive.metastore.api.PartitionFilterMode;
 import org.apache.hadoop.hive.metastore.api.PartitionSpec;
 import org.apache.hadoop.hive.metastore.api.PartitionWithoutSD;
 import org.apache.hadoop.hive.metastore.api.PartitionsByExprRequest;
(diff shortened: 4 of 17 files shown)


armitage420:
I noticed @miklosgergely has not been active for a while probably. 
@zhangbutao @dengzhhu653 @ayushtkn @deniskuzZ Could you please help with the review if/when you have the time?
Thank You ✨

deniskuzZ:
cc @wecharyu, maybe you would be interested in this as well

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/client/builder/GetPartitionProjectionsSpecBuilder.java
```diff
@@ -41,6 +42,12 @@ public GetPartitionProjectionsSpecBuilder addProjectField(String field) {
         return this;
     }
 
+    public GetPartitionProjectionsSpecBuilder addProjectFieldList(List<String> fields) {
+        fieldList.addAll(Arrays.asList("catName","dbName","tableName"));
```
wecharyu, line 46: why need these fields here?

armitage420: @wecharyu In case of sending an empty list, our API will fetch the whole object. Hence, we are using basic catName, dbName, tblName for cases like, we just need to check that partition exists.

armitage420:
@wecharyu seems like I am unable to re-run tests. And there is no button to re-run tests 
@ayushtkn any help with this? We used to have a rerun tests button in CI.

ayushtkn:
> seems like I am unable to re-run tests. And there is no button to re-run tests
@ayushtkn any help with this? We used to have a rerun tests button in CI.

@armitage420 there is a log in button on the top right, once you click that it should show you the old screen where you can trigger the tests, it uses Github creds If I remember right

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/exchange/AlterTableExchangePartitionAnalyzer.java
```diff
@@ -77,9 +85,24 @@ protected void analyzeCommand(TableName tableName, Map<String, String> partition
...
+    List<String> sourceProjectFilters = MetaStoreUtils.getPvals(sourceTable.getPartCols(), partitionSpecs);
 
     // check if source partition exists
-    PartitionUtils.getPartitions(db, sourceTable, partitionSpecs, true);
+    GetPartitionsFilterSpec getSourcePartitionsFilterSpec = new GetPartitionsFilterSpec();
+    getSourcePartitionsFilterSpec.setFilters(sourceProjectFilters);
+    getSourcePartitionsFilterSpec.setFilterMode(PartitionFilterMode.BY_VALUES);
+
+    GetProjectionsSpec getProjectionsSpec = new GetPartitionProjectionsSpecBuilder()
+        .addProjectFieldList(Arrays.asList("values","lastAccessTime")).build();
```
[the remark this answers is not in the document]
armitage420: @wecharyu This piece of code does not require "lastAccessTime", although it does require "values" in the process when _convertFromPartSpec_ method is called upon by HS2 in the very process to extract our partitions list.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4086,6 +4091,37 @@ public List<String> getPartitionNames(String dbName, String tblName,
...
+        req.setValidWriteIdList(validWriteIdList != null ? validWriteIdList.toString() : null);
+        req.setId(t.getTTable().getId());
+      }
+      GetPartitionNamesPsResponse res = getMSC().listPartitionNamesRequest(req);
+      names = res.getNames();
+    } catch (NoSuchObjectException nsoe) {
+      // this means no partition exists for the given partition spec
+      // key value pairs - thrift cannot handle null return values, hence
+      // listPartitionNames() throws NoSuchObjectException to indicate null partitions
+      return Lists.newArrayList();
```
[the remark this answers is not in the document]
wecharyu: I think we should wrap and throw an exception for `NoSuchObjectException`, as returning an empty result can be ambiguous.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/client/builder/GetPartitionProjectionsSpecBuilder.java
armitage420: Correct. In case where a user wants to keep all projectFieldList, they just do not have to call this method or addProjectField method at all. 
We would require calling this particular method only once, to add to our request, hence _"catName","dbName","tableName"_ would only be added once along with the list of project fields if any.

armitage420:
> > seems like I am unable to re-run tests. And there is no button to re-run tests
> > @ayushtkn any help with this? We used to have a rerun tests button in CI.
> 
> @armitage420 there is a log in button on the top right, once you click that it should show you the old screen where you can trigger the tests, it uses Github creds If I remember right

@ayushtkn I am able to see only a `go to classic` button instead that redirects us to jenkins. And we do not have permissions to run precommit pipelines in jenkins
image

ayushtkn:
Hmm, post this:
https://lists.apache.org/thread/c697bb63xxsd3g7zws4n31z59gtbgt4t

I think now that access has been removed, you need to trigger the build by pushing empty commit only

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java
```diff
@@ -89,7 +93,18 @@ List<Path> getLocations(Hive db, Partition partition, Table table) throws HiveEx
     List<Path> locations = new ArrayList<Path>();
     if (table.isPartitioned()) {
       if (partition == null) {
-        for (Partition currPartition : db.getPartitions(table)) {
+      List<Partition> partitions;
+      GetProjectionsSpec getProjectionsSpec = new GetPartitionProjectionsSpecBuilder()
+              .addProjectFieldList(Arrays.asList("sd.location")).build();
```
deniskuzZ, line 98: where is the `sd` alias come from?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java
```diff
@@ -89,7 +93,18 @@ List<Path> getLocations(Hive db, Partition partition, Table table) throws HiveEx
...
+        GetPartitionsRequest request = new GetPartitionsRequest(table.getDbName(), table.getTableName(), getProjectionsSpec, null);
```
deniskuzZ, line 99: can we use `GetPartitionsRequestBuilder` here, maybe add simpler constructor

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java
```diff
@@ -89,7 +93,18 @@ List<Path> getLocations(Hive db, Partition partition, Table table) throws HiveEx
...
+        request.setCatName(table.getCatName());
+        try{
+          partitions = db.getPartitionsWithSpecs(table, request);
```
deniskuzZ, line 102: why not use new helper method from `PartitionUtils`?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/exchange/AlterTableExchangePartitionAnalyzer.java
```diff
@@ -77,9 +84,20 @@ protected void analyzeCommand(TableName tableName, Map<String, String> partition
     if (AcidUtils.isTransactionalTable(sourceTable) || AcidUtils.isTransactionalTable(destTable)) {
       throw new SemanticException(ErrorMsg.EXCHANGE_PARTITION_NOT_ALLOWED_WITH_TRANSACTIONAL_TABLES.getMsg());
     }
+    List<String> sourceProjectFilters = MetaStoreUtils.getPvals(sourceTable.getPartCols(), partitionSpecs);
 
     // check if source partition exists
-    PartitionUtils.getPartitions(db, sourceTable, partitionSpecs, true);
+    GetPartitionsFilterSpec getSourcePartitionsFilterSpec = new GetPartitionsFilterSpec();
```
deniskuzZ, line 90: please remove `get` from var names

ql/src/java/org/apache/hadoop/hive/ql/exec/repl/ReplExternalTables.java
```diff
@@ -97,15 +100,21 @@ void dataLocationDump(Table table, FileList fileList, HashMap<String, Boolean> s
     }
     if (table.isPartitioned()) {
       List<Partition> partitions;
+      GetPartitionsRequest request = new GetPartitionsRequest(table.getDbName(), table.getTableName(),
+          null, null);
+      request.setCatName(table.getCatName());
+      request.setProjectionSpec(
```
deniskuzZ, line 106: why not set it through constructor?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4086,6 +4091,37 @@ public List<String> getPartitionNames(String dbName, String tblName,
     return names;
   }
 
+  // get partition names from provided partition values
+  public List<String> getPartitionNames(String dbName, String tblName,
+      short max, List<String> pVals) throws HiveException {
```
deniskuzZ, line 4096: why max is `short` type + I think it should be the last param

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4086,6 +4091,37 @@ public List<String> getPartitionNames(String dbName, String tblName,
     return names;
   }
 
+  // get partition names from provided partition values
+  public List<String> getPartitionNames(String dbName, String tblName,
```
deniskuzZ, line 4095: it seems to be a partial code duplication. why not extract and reuse same logic from `List<String> getPartitionNames(String dbName, String tblName, Map<String, String> partSpec, short max)`

deniskuzZ: `max` should be the last param

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4466,6 +4526,71 @@ public List<Partition> getPartitionsByFilter(Table tbl, String filter)
...
   }
 
+  public List<Partition> getPartitionsWithSpecs(Table tbl, GetPartitionsRequest request)
+      throws HiveException, TException {
+
+    if (!tbl.isPartitioned()) {
+      throw new HiveException(ErrorMsg.TABLE_NOT_PARTITIONED, tbl.getTableName());
+    }
+    int batchSize= MetastoreConf.getIntVar(Hive.get().getConf(), MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX);
+    if(batchSize > 0){
```
deniskuzZ, line 4536: nit: space

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4466,6 +4526,71 @@ public List<Partition> getPartitionsByFilter(Table tbl, String filter)
...
+      return new ArrayList<>(getAllPartitionsWithSpecsInBatches(tbl, batchSize, DEFAULT_BATCH_DECAYING_FACTOR, MetastoreConf.getIntVar(
+          Hive.get().getConf(), MetastoreConf.ConfVars.GETPARTITIONS_BATCH_MAX_RETRIES), request));
+    }else{
```
deniskuzZ, line 4539: please format

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4466,6 +4526,71 @@ public List<Partition> getPartitionsByFilter(Table tbl, String filter)
...
+      return getPartitionsWithSpecsInternal(tbl, request);
+    }
+  }
+
+  public List<Partition> getPartitionsWithSpecsInternal(Table tbl, GetPartitionsRequest request)
```
deniskuzZ, line 4544: should it be public?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4230,6 +4266,30 @@ public Void execute(int size) throws HiveException {
     return result;
   }
 
+  public Set<Partition> getAllPartitionsWithSpecsInBatches(Table tbl, int batchSize, int decayingFactor,
```
deniskuzZ, line 4269: should it be public?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
deniskuzZ: should it be public?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -4466,6 +4526,71 @@ public List<Partition> getPartitionsByFilter(Table tbl, String filter)
...
+    }
+    GetPartitionsResponse response = getMSC().getPartitionsWithSpecs(request);
+    List<org.apache.hadoop.hive.metastore.api.PartitionSpec> partitionSpecs = response.getPartitionSpec();
+    List<Partition> partitions = new ArrayList<>();
+    partitions.addAll(convertFromPartSpec(partitionSpecs.iterator(), tbl));
+
+    return partitions;
+  }
+
+  List<Partition> getPartitionsWithSpecsByNames(Table tbl, List<String> partNames, GetPartitionsRequest request)
```
deniskuzZ, line 4558: what is the benefit of splitting the filter into batches?

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
```diff
@@ -99,12 +104,16 @@ private void getNextBatch() {
           batchCounter++;
         }
         try {
-          if (isAuthRequired) {
-            batchIter = db.getPartitionsAuthByNames(table, nameBatch, userName, groupNames).iterator();
-          } else {
-            batchIter = db.getPartitionsByNames(table, nameBatch, getColStats).iterator();
+          if(getPartitionsRequest == null){
```
deniskuzZ, line 107: please format the code

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
```diff
@@ -99,12 +104,16 @@ private void getNextBatch() {
...
+            if (isAuthRequired) {
+              batchIter = db.getPartitionsAuthByNames(table, nameBatch, userName, groupNames).iterator();
+            } else {
+              batchIter = db.getPartitionsByNames(table, nameBatch, getColStats).iterator();
+            }
+          }else {
+            batchIter = db.getPartitionsWithSpecsByNames(table, nameBatch, getPartitionsRequest).iterator();
```
deniskuzZ, line 114: does it handle `isAuthRequired`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
```diff
@@ -272,9 +272,8 @@ void checkTable(String catName, String dbName, String tableName, byte[] filterEx
       } else {
         GetPartitionsRequest request = new GetPartitionsRequest(table.getDbName(), table.getTableName(),
             null, null);
-        request.setProjectionSpec(new GetPartitionProjectionsSpecBuilder().addProjectField("sd.location")
-            .addProjectField("createTime").addProjectField("tableName")
-            .addProjectField("values").build());
+        request.setProjectionSpec(new GetPartitionProjectionsSpecBuilder()
+                .addProjectFieldList(Arrays.asList("sd.location","createTime","values")).build());
```
deniskuzZ, line 276: why not extract `projectionSpec` into local var and pass to constructor

ql/src/test/results/clientnegative/exchange_partition_neg_partition_missing.q.out
```diff
@@ -20,4 +20,4 @@ PREHOOK: Input: default@exchange_part_test1
 POSTHOOK: query: SHOW PARTITIONS exchange_part_test1
 POSTHOOK: type: SHOWPARTITIONS
 POSTHOOK: Input: default@exchange_part_test1
-FAILED: SemanticException [Error 10006]: Partition not found {ds=2013-04-05}
+FAILED: SemanticException [Error 10006]: Partition not found GetPartitionsFilterSpec(filterMode:BY_VALUES, filters:[2013-04-05])
```
deniskuzZ, line 23: partition name is not complete!

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
```diff
@@ -271,4 +277,69 @@ List<Partition> getPartitionsByFilter(final String filter) throws MetaException
...
+            LOG.error("JavaScript script engine is not found, therefore partition filtering "
+                    + "for temporary tables is disabled.");
+            break;
+          }
+
+          for (String filter : filterSpec.getFilters()) {
+            try {
+              se.put("partition", partition);
+              matches = (Boolean) se.eval(filter);
+              if (!matches) break;
```
deniskuzZ, line 325: seems unnnesesary

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
```diff
@@ -173,6 +183,42 @@ public PartitionIterable(Hive db, Table table, Map<String, String> partialPartit
     this(db, table, partialPartitionSpec, batchSize, getColStats, false, null, null);
   }
 
+  public PartitionIterable(Hive db, Table table, int batchSize, GetPartitionsRequest getPartitionsRequest)
```
deniskuzZ, line 186: `batchSize` should be the last param

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
```diff
@@ -173,6 +183,42 @@ public PartitionIterable(Hive db, Table table, Map<String, String> partialPartit
...
+    if (batchSize < 1) {
+      throw new HiveException("Invalid batch size for partition iterable. Please use a batch size greater than 0");
+    }
+    this.currType = Type.LAZY_FETCH_PARTITIONS;
+    this.db = db;
+    this.table = table;
+    this.batchSize = batchSize;
+    this.getPartitionsRequest = getPartitionsRequest;
+    List<String> pVals = null;
+    if(getPartitionsRequest.isSetFilterSpec()){
```
deniskuzZ, line 197: please format the code

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
```diff
@@ -173,6 +183,42 @@ public PartitionIterable(Hive db, Table table, Map<String, String> partialPartit
...
+      pVals = this.getPartitionsRequest.getFilterSpec().getFilters();
+    }
+    if (pVals == null){
+      partitionNames = db.getPartitionNames(
+          table.getDbName(),table.getTableName(), (short) -1);
+    } else {
+      PartitionFilterMode filterMode = getPartitionsRequest.getFilterSpec().getFilterMode();
```
deniskuzZ, line 204: same code is used in `[PartitionTree.java](https://github.com/apache/hive/pull/5539/files#diff-f83827d7c6ec0a73bbc558101e64145afb555daff397c999c7491373253168c7)` try to reuse code and avoid code duplication

deniskuzZ:
thanks for the patch @armitage420! 
in general, LGTM, however, please take a look at the review comments and issues reported by sonar. 
thanks!

deniskuzZ:
@armitage420 any plans to continue with this PR? looks abandoned

armitage420:
@deniskuzZ Thank you for the review! I was a bit busy with some work. Will be updating this PR by today!

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java
armitage420: I am a bit skeptical about adding the new constructor for this use-case, as the code complexity approximately would be the same here. GetPartitionsRequestBuilder was built so that we would not have to write multiple set_methods for our projections,

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java
armitage420: PartitionUtils is a part of org.apache.hadoop.hive.ql.ddl.table.partition
package implying it is made for DDL operations for partitioned tables. Here we are not performing any DDL operation.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
armitage420: 1. fixed in line 4069

2. setting `max` to the last param results in ambiguous method calls when public List<String> `getPartitionNames(String dbName, String tblName, Map<String, String> partSpec, short max)` is called with `partSpec` set to null.

3. Third party users using HS2 can benefit from this in order to get partition names with just given filter values.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
armitage420: Third party users that use HS2 internally can benefit from this method. Moreover, the presence of InterfaceAudience.Public in GetPartitionsRequest class strongly indicates that this class(one of our method parameters) is intended for public use.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
armitage420: This particular case scenario is made for huge partitioned tables where our thrift network would hit a 2GB data threshold. 

We might need to change the value METASTORE_BATCH_RETRIEVE_MAX in order to benefit from this very method. We will have to choose a very approximate value for the same though. As, I am not able to come up with a real time calculation of max batchsize required. And the data size now is going to be pretty dynamic with dynamic projections for partitions.

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
armitage420: The switch cases does have the same conditions, while what the code is doing for each cases are different. 
`PartitionTree` was made for handling temp tables, along with their partitions in HS2 itself(the very temp table and map of partition objects are initialized, populated in the same class).
`PartitionIterable` on the other hand has to call upon (or just add if provided based on cases) partition names from HMS instead to do batch operations.

ql/src/java/org/apache/hadoop/hive/ql/exec/repl/ReplExternalTables.java
armitage420: my bad

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
armitage420: fixed

ql/src/test/results/clientnegative/exchange_partition_neg_partition_missing.q.out
armitage420: fixed!

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
deniskuzZ: 👍

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java
deniskuzZ: could we pass `Hive db, Table table` through the GetPartitionsRequest
````
public PartitionIterable(GetPartitionsRequest getPartitionsRequest, int batchSize) {
...
}
````

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java (hunk 23 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java:203]: ok

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java (hunk 21 above)
Comment (armitage420) [reply ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionIterable.java]: We would need a `Hive db` to be passed on. We can definitely do without a `Table table`. 
```
public PartitionIterable(Hive db, GetPartitionsRequest getPartitionsRequest, int batchSize) {
...
}
```

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java [hunk 24]
```diff
@@ -89,7 +93,18 @@ List<Path> getLocations(Hive db, Partition partition, Table table) throws HiveEx
     List<Path> locations = new ArrayList<Path>();
     if (table.isPartitioned()) {
       if (partition == null) {
-        for (Partition currPartition : db.getPartitions(table)) {
+      List<Partition> partitions;
+      GetProjectionsSpec getProjectionsSpec = new GetPartitionProjectionsSpecBuilder()
+              .addProjectFieldList(Arrays.asList("sd.location")).build();
+        GetPartitionsRequest request = new GetPartitionsRequest(table.getDbName(), table.getTableName(), getProjectionsSpec, null);
+        request.setCatName(table.getCatName());
+        try{
```
deniskuzZ, line 101: nit: space

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/status/formatter/ShowTableStatusFormatter.java
deniskuzZ: `PartitionUtils#getPartitionsWithSpecs` is a public method, so I don't get what is an issue. 
Anyways that's fine

deniskuzZ (approved):
+1, pending tests

armitage420:
@deniskuzZ Thank you so much for your review again! 
The pending test is fixed. Waiting for the pipeline run!

armitage420:
The build is green! phew! :')

PR #5541, opened 2024-11-14
HIVE-28586 Support write order for Iceberg tables at CREATE TABLE

zratkai:
Change-Id: Ia9a0a92d19d33693887137c797e0662088a314db

Support for write order in iceberg tables. Like: 
CREATE TABLE prod.db.sample WRITE ORDERED BY category ASC NULLS LAST, id DESC NULLS FIRST

To support write ordering in iceberg tables.

Yes, user can now use this syntax:
CREATE TABLE prod.db.sample WRITE ORDERED BY category ASC NULLS LAST, id DESC NULLS FIRST;

No.

With qtest.


parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g
```diff
@@ -1840,6 +1841,14 @@ tableImplBuckets
     -> ^(TOK_ALTERTABLE_BUCKETS $num)
     ;
 
+tableWriteOrdered
+@init { pushMsg("table sorted specification", state); }
+@after { popMsg(state); }
+    :
+      KW_WRITE KW_ORDERED KW_BY sortCols=columnNameOrderList
```
zhangbutao, line 1848: Does this syntax equal to Spark-Iceberg ` alter-table-write-ordered-by`?
https://iceberg.apache.org/docs/1.7.0/spark-ddl/#alter-table-write-ordered-by

Do you consider also to add the ` alter-table-write-ordered-by` syntax? So that user can alter the existing iceberg table to support write order.

zhangbutao: BTW, Trino added the `write-sort ` through table property:
https://trino.io/docs/current/connector/iceberg.html#sorted-tables
https://www.starburst.io/blog/improving-performance-with-iceberg-sorted-tables/


```
CREATE TABLE
catalog_sales_sorted (
   cs_sold_date_sk bigint,
   more columns...
)
WITH
(
   format = 'PARQUET',
   sorted_by = ARRAY['cs_sold_date_sk'],
   type = 'ICEBERG'
)
```

This can avoid introducing new create syntax. Anyway, we should find a simple syntax which user can use.

I also want to hear other folks opinions about the syntax. 
cc @ayushtkn @deniskuzZ @okumin

okumin: Is the semantics of the sort order of Iceberg different from that of Hive's SORTED BY of bucketed sorted tables? If it is the same, we may use it. Maybe, it is different. I also feel the global ordering is hard to achieve with the current Hive.

deniskuzZ: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL
````
The CLUSTERED BY and SORTED BY creation commands do not affect how data is inserted into a table – only how it is read. This means that users must be careful to insert data correctly by specifying the number of reducers to be equal to the number of buckets, and using CLUSTER BY and SORT BY commands in their query.
````
@okumin, this PR intends to add support for Iceberg sorted tables, same as in a trino link you've added. I do not really like that we have to create new syntax for that. 
`SORTED BY` kinda makes sense, but it would be overloaded for Iceberg since it would be used to specify data insert order.

Spark only supports [alter-table-write-ordered-by](https://iceberg.apache.org/docs/1.7.0/spark-ddl/#alter-table-write-ordered-by) to define the write order. 

Maybe we could have a combination of ALTER + Trino table properties styles?

Comment (okumin) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g:1848]: I understand the SORT order of the Iceberg is a hint in writing.
- [Table Spec](https://iceberg.apache.org/spec/#sorting) says, "Writers should use this default sort order to sort the data on write, but are not required to if the default order is prohibitively expensive, as it would be for streaming writes."
- [Spark DDL](https://learn.microsoft.com/en-us/cli/azure/role/definition?view=azure-cli-latest#az-role-definition-list) says, "Table write order does not guarantee data order for queries. It only affects how data is written to the table."
- [Trino has a toggle, sorted_writing_enabled, to obey the hint](https://trino.io/docs/current/connector/iceberg.html#iceberg-sorted-files)

So, I agree that reusing SORTED BY is confusing. I'm also checking what syntaxes are used in other engines...

Comment (okumin) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g:1848]: ## How other Iceberg-friendly query engines support SORT orders

### ALTER TABLE
- [Spark supports WRITE ORDERED BY and WRITE LOCALLY ORDERED BY](https://iceberg.apache.org/docs/1.7.0/spark-ddl/#alter-table-write-ordered-by)
- [Trino supports it through table properties](https://trino.io/docs/current/connector/iceberg.html#alter-table-set-properties)
- [Dremio supports LOCALSORT BY](https://docs.dremio.com/cloud/reference/sql/commands/alter-table)

### CREATE TABLE
- [Spark doesn't support it?](https://github.com/apache/iceberg/blob/apache-iceberg-1.7.1/spark/v3.5/spark-extensions/src/main/antlr/org.apache.spark.sql.catalyst.parser.extensions/IcebergSqlExtensions.g4)
- [Trino supports it through table properties](https://trino.io/docs/current/connector/iceberg.html#sorted-tables)
- [Dremio supports LOCALSORT BY](https://docs.dremio.com/cloud/reference/sql/commands/create-table/)

## Related discussions

I found a related discussion although it would not include the conclusion we want.
https://github.com/apache/iceberg/issues/3547

## My current opinion

If we introduce `ALTER TABLE ... WRITE ORDERED BY`, we can reuse it in CREATE TABLE in my opinion. It is consistent and no new keywords are introduced for CREATE TABLE. If we don't, I prefer to use a property for both purposes.

Comment (zratkai) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g:1848]: @zhangbutao , @okumin , @deniskuzZ 
First of all, thanks for your comments! This syntax comes from the official Iceberg recommendation, which is for alter table, and to be consistend I modified the ALTER to CREATE in the syntax, so at CREATE table it is the same as ALTER TABLE:
[https://iceberg.apache.org/docs/1.7.0/spark-ddl/#alter-table-write-ordered-by]

My plan is to work as a next step on ALTER TABLE after this PR is done. I already created the ticket for that at the time this ticket was born: 
[https://issues.apache.org/jira/browse/HIVE-28587]

About table property idea and comparisms: 
The iceberg syntax defines multiple keywords in the above page, not just the colum names:

"ALTER TABLE prod.db.sample WRITE ORDERED BY category ASC NULLS LAST, id DESC NULLS FIRST"

The mentioned examples do not cover all of that. E.g. I could not find in the linked Trino documentation how the ordering is defined (ASC/DESC) or the null ordering (NULLS FIRST/LAST). With that the table property solution would look like ugly and hard to remember and write correctly for the user. 

Since it is passedas a table property deep in the code here, it works with table property as well with this commit :
e.g.:
create table ice_orc_sorted (id int, text string) stored by iceberg stored as orc TBLPROPERTIES('iceberg.write-order'='{"order-id":1,"fields":[{"transform":"identity","source-id":1,"direction":"asc","null-order":"nulls-last"}]}');

The syntax comes from Iceberg built in format for sorted order, but I do not find it user friendly and I do not recommend to go that way.


About SORTED BY/ SORT BY/LOCALSORT BY: as you already mentioned it would be confusing, and the "ALTER/CREATE TABLE prod.db.sample WRITE ORDERED BY category ASC NULLS LAST, id DESC NULLS FIRST" is recommended by Iceberg, and easy to understand and remember.

So every DB engine has it's own syntax for this purpose and **neither is following Iceberg recommendation except this one.**

As far as I found Spark does not even support yet sort order at alter table:
https://spark.apache.org/docs/3.5.3/sql-ref-syntax-ddl-alter-table.html#content
https://spark.apache.org/docs/3.5.3/sql-ref-syntax-ddl-create-table.html


At the linked iceberg discussion even Ryan Blue said this:
"No, I don't think so. These aren't table properties... "
And someone else said:
" I don't think using TBLPROPERTIES is a good idea..."

I hope I answered all the questions!

Comment (deniskuzZ) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g:1848]: I like the `LOCALSORT BY` it gives more context than `WRITE ORDERED BY`

Comment (zratkai) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g:1848]: @zhangbutao @okumin @deniskuzZ I would like to finish this PR. Could you please give feedback if it is acceptable as I described the design of this ticket?

Comment (okumin) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g:1848]: > The mentioned examples do not cover all of that. E.g. I could not find in the linked Trino documentation how the ordering is defined (ASC/DESC) or the null ordering (NULLS FIRST/LAST)

I think this is a critical point, and now I am biased toward to `WRITE ORDERED BY` + `WRITE LOCALLY ORDERED BY`. I am trying to talk to a Trino maintainers

Comment (okumin) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g:1848]: JFYI, I heard Trino would support some detailed properties and verified the behavior.

```
trino> CREATE TABLE iceberg.default.test1(a int) WITH (sorted_by = ARRAY['a']);
CREATE TABLE
trino> CREATE TABLE iceberg.default.test2(a int) WITH (sorted_by = ARRAY['a DESC NULLS FIRST']);
CREATE TABLE
```

```
$ hdfs dfs -cat /user/hive/warehouse/test1-d38df51e56ee445aa4b100fc6306cd78/metadata/00000-77ea73cf-95ec-44fc-bf47-2287e19a4195.metadata.json
...
  "default-sort-order-id" : 1,
  "sort-orders" : [ {
    "order-id" : 1,
    "fields" : [ {
      "transform" : "identity",
      "source-id" : 1,
      "direction" : "asc",
      "null-order" : "nulls-first"
    } ]
  } ],
```

```
$ hdfs dfs -cat /user/hive/warehouse/test2-1d2d11372fad4f3e945d298f63afe322/metadata/00000-62c0c21b-2032-4883-a092-012bd1572e83.metadata.json
...
  "default-sort-order-id" : 1,
  "sort-orders" : [ {
    "order-id" : 1,
    "fields" : [ {
      "transform" : "identity",
      "source-id" : 1,
      "direction" : "desc",
      "null-order" : "nulls-first"
    } ]
  } ],
```

File: parser/src/test/org/apache/hadoop/hive/ql/parse/TestReservedWords.java [hunk 2]
```diff
@@ -119,6 +119,7 @@ public static Collection<String> data() {
         "ONLY",
         "OR",
         "ORDER",
+        "ORDERED",
```
okumin, line 122: I would like to include it in non-reserved words because `ORDERED` is not listed in SQL:2023.
https://github.com/apache/hive/blob/cbb14ddb1738466b08c01082470c921d51bb8046/parser/src/java/org/apache/hadoop/hive/ql/parse/IdentifiersParser.g#L931

parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g
okumin: First of all, I think the consistency between ALTER TABLE and CREATE TABLE is the most important.
On top of that, I think Spark is the most common engine for Iceberg. I'm reading O'Reilly's "Apache Iceberg: The Definitive Guide", and the first example of any feature is almost always Spark SQL. My preference here is to use the syntax introduced for Spark.

zhangbutao: @zratkai you have created the ticket [HIVE-28587](https://issues.apache.org/jira/browse/HIVE-28587), which means that you will implement the `alter table `syntax like spark syntax. https://iceberg.apache.org/docs/1.7.0/spark-ddl/#alter-table-write-ordered-by  Right?

> This syntax comes from the official Iceberg recommendation, which is for alter table, and to be consistend I modified the ALTER to CREATE in the syntax, so at CREATE table it is the same as ALTER TABLE:

I already know why you give this create syntax. But since there is no uniform `Create `syntax for the `write order`, i think we no need to make `Create `syntax to be consistend with the `Alter `syntax. If the table property is more simple, i prefer to `create + table property` like trino.

If you think `create + table property` is also not friendly to users. I am not against the currrent implementation. You just need to update the wiki doc to tell users the create dialect for iceberg write order.

ayushtkn: If there is no common agreement, We can reach out the dev mailing list & can proceed with what majority of the folks like, This Syntax discussion is kind of pure personal preference stuff I believe, what each of us find more convenient & there is no way we can conclude like this which is better.

Was reading this doc:
https://www.tabular.io/apache-iceberg-cookbook/data-engineering-table-write-order/
It had something like
```
ALTER TABLE logs WRITE LOCALLY ORDERED BY level, event_ts
```
& 

here:
https://iceberg.apache.org/docs/latest/spark-ddl/#alter-table-write-ordered-by

```
ALTER TABLE prod.db.sample WRITE LOCALLY ORDERED BY category, id
```

If LOCALLY isn't specified it is Global Sort order as the doc says 'To order within each task, not across tasks, use LOCALLY ORDERED BY:' and I believe this PR is just supporting LOCAL sort order not the Global one.

So, this kind of looks fine to me, We can just suffix in the Create command as well something like "WRITE LOCALLY ORDERED BY level, event_ts" which will be in sync with ALTER command & all such above docs.

Again just my preference, if all of us aren't convinced on one choice, we can fallback to dev@ and go what the majority feels more soothing to eyes.

zratkai: @ayushtkn thanks Ayush for the comment!
I agree with you and @deniskuzZ,  LOCALLY ORDERED BY is the right choice here.

@okumin @zhangbutao if you can accept that I will modify it accordingly.

zhangbutao: It's fine to me. 😃 Just do it. We can optimize this later if someone has new good idea in the future.

okumin: I give +1 to WRITE LOCALLY ORDERED BY.

I double-checked `WRITE ORDERED BY` vs `WRITE LOCALLY ORDERED BY` with Spark 3.5.1 and Iceberg 1.6.1.

```
spark-sql (default)> CREATE TABLE hadoop_prod.default.test2 (a int) USING iceberg;
Time taken: 0.089 seconds
spark-sql (default)> ALTER TABLE hadoop_prod.default.test2 WRITE ORDERED BY a;
Time taken: 0.182 seconds
spark-sql (default)> CREATE TABLE hadoop_prod.default.test3 (a int) USING iceberg;
Time taken: 0.086 seconds
spark-sql (default)> ALTER TABLE hadoop_prod.default.test3 WRITE LOCALLY ORDERED BY a;
```

This is the diff.

```
zookage@client-node-0:~$ hdfs dfs -cat /user/hive/warehouse/catalog/default/test2/metadata/v2.metadata.json > /tmp/test2.json
zookage@client-node-0:~$ hdfs dfs -cat /user/hive/warehouse/catalog/default/test3/metadata/v2.metadata.json > /tmp/test3.json
zookage@client-node-0:~$ diff /tmp/test2.json /tmp/test3.json 
3,4c3,4
<   "table-uuid" : "821c3cc2-1320-45dc-bb2a-c805778caa91",
<   "location" : "hdfs://hdfs-namenode-0.hdfs-namenode:8020/user/hive/warehouse/catalog/default/test2",
---
>   "table-uuid" : "094951f1-6229-43cd-bffe-cb2f086b8dda",
>   "location" : "hdfs://hdfs-namenode-0.hdfs-namenode:8020/user/hive/warehouse/catalog/default/test3",
6c6
<   "last-updated-ms" : 1733994463428,
---
>   "last-updated-ms" : 1733994491871,
40c40
<     "write.distribution-mode" : "range",
---
>     "write.distribution-mode" : "none",
50,51c50,51
<     "timestamp-ms" : 1733994460005,
<     "metadata-file" : "hdfs://hdfs-namenode-0.hdfs-namenode:8020/user/hive/warehouse/catalog/default/test2/metadata/v1.metadata.json"
---
>     "timestamp-ms" : 1733994472725,
>     "metadata-file" : "hdfs://hdfs-namenode-0.hdfs-namenode:8020/user/hive/warehouse/catalog/default/test3/metadata/v1.metadata.json"
```

Looks like, the meaningful difference is only `write.distribution-mode=range` or `write.distribution-mode=none`. I guess adding LOCALLY makes more sense unless we give `range`.

I am not a specialist of Apache Spark. Please feel free to correct me if I am wrong.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/InputFormatConfig.java
```diff
@@ -88,6 +88,7 @@ private InputFormatConfig() {
   public static final String CATALOG_CLASS_TEMPLATE = "iceberg.catalog.%s.catalog-impl";
   public static final String CATALOG_DEFAULT_CONFIG_PREFIX = "iceberg.catalog-default.";
   public static final String QUERY_FILTERS = "iceberg.query.filters";
+  public static final String INSERT_WRITE_ORDER = "iceberg.write-order";
```
deniskuzZ, line 91: it's a table property, right? why not just `write.order`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -271,6 +275,23 @@ public void preCreateTable(CreateTableRequest request) {
     setOrcOnlyFilesParam(hmsTable);
     // Remove hive primary key columns from table request, as iceberg doesn't support hive primary key.
     request.setPrimaryKeys(null);
+    addSortOrder(hmsTable, schema, catalogProperties);
```
deniskuzZ, line 278: isn't it set in 
````
HiveTableOperations.setHmsTableParameters(..) ->  setSortOrder(metadata, parameters)
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/Catalogs.java [hunk 5]
```diff
@@ -140,15 +143,23 @@ public static Table createTable(Configuration conf, Properties props) {
...
+          .withLocation(location).withProperties(map).withSortOrder(sortOrder).create();
     }
 
     Preconditions.checkNotNull(location, "Table location not set");
-    return new HadoopTables(conf).create(schema, spec, map, location);
+    return new HadoopTables(conf).create(schema, spec, sortOrder, map, location);
+  }
+
+  private static SortOrder getSortOrder(Properties props, Schema schema) {
+    String sortOrderJsonString = props.getProperty(InputFormatConfig.INSERT_WRITE_ORDER);
```
deniskuzZ, line 159:
````
return SortOrderParser.fromJson(schema(), sortOrderAsJson)
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java [hunk 6]
```diff
@@ -271,6 +275,23 @@ public void preCreateTable(CreateTableRequest request) {
...
+  }
+
+  private void addSortOrder(org.apache.hadoop.hive.metastore.api.Table hmsTable, Schema schema,
+      Properties properties) {
+    SortOrder.Builder sortOderBuilder = SortOrder.builderFor(schema);
+    hmsTable.getSd().getSortCols().forEach(item -> {
```
deniskuzZ, line 284: item -> field

deniskuzZ: iceberg support sort for expressions like `sort by day(time)`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -271,6 +275,23 @@ public void preCreateTable(CreateTableRequest request) {
...
+      NullOrder nullOrder = item.getNullOrdering() == NullOrderingType.NULLS_FIRST ?
+          NullOrder.NULLS_FIRST : NullOrder.NULLS_LAST;
+      if (item.getOrder() == 0) {
```
deniskuzZ, line 287: sortOderBuilder.sortBy(item.getCol(), (field.getCol() == 0) ? ASC:DESC, nullOrder)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -271,6 +275,23 @@ public void preCreateTable(CreateTableRequest request) {
     setOrcOnlyFilesParam(hmsTable);
     // Remove hive primary key columns from table request, as iceberg doesn't support hive primary key.
     request.setPrimaryKeys(null);
+    addSortOrder(hmsTable, schema, catalogProperties);
+  }
+
+  private void addSortOrder(org.apache.hadoop.hive.metastore.api.Table hmsTable, Schema schema,
```
deniskuzZ, line 281: `setSortOrder`

iceberg/iceberg-handler/src/test/queries/positive/iceberg_create_sorted_table.q
```diff
@@ -0,0 +1,33 @@
...
+--! qt:replace:/(\s+current-snapshot-timestamp-ms\s+)\S+(\s*)/$1#Masked#$2/
+--! qt:replace:/(MAJOR\s+succeeded\s+)[a-zA-Z0-9\-\.\s+]+(\s+manual)/$1#Masked#$2/
+-- Mask iceberg version
+--! qt:replace:/(\S\"iceberg-version\\\":\\\")(\w+\s\w+\s\d+\.\d+\.\d+\s\(\w+\s\w+\))(\\\")/$1#Masked#$3/
+set hive.llap.io.enabled=true;
+set hive.vectorized.execution.enabled=true;
+set hive.optimize.shared.work.merge.ts.schema=true;
+
+
+create table ice_orc_sorted (id int, text string) write ordered by id desc nulls first, text asc nulls last stored by iceberg stored as orc;
```
deniskuzZ, line 25: Should we specify the sort column/expressions in parentheses?

iceberg/iceberg-handler/src/test/results/positive/create_iceberg_table_stored_by_iceberg_with_serdeproperties.q.out
```diff
@@ -31,6 +31,7 @@ Table Parameters:
 	current-schema      	{\"type\":\"struct\",\"schema-id\":0,\"fields\":[{\"id\":1,\"name\":\"i\",\"required\":false,\"type\":\"int\"},{\"id\":2,\"name\":\"s\",\"required\":false,\"type\":\"string\"},{\"id\":3,\"name\":\"ts\",\"required\":false,\"type\":\"timestamp\"},{\"id\":4,\"name\":\"d\",\"required\":false,\"type\":\"date\"}]}
 	format-version      	2                   
 	iceberg.orc.files.only	true                
+	iceberg.write-order 	{\"order-id\":0,\"fields\":[]}
```
deniskuzZ, line 34: should we add this if no order specified?

parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g
```diff
@@ -1840,6 +1841,14 @@ tableImplBuckets
...
+    -> ^(TOK_ALTERTABLE_BUCKETS $sortCols?)
```
deniskuzZ, line 1849: we should support iceberg transform expressions as well, not just identity

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
```diff
@@ -966,15 +967,10 @@ protected List<Order> getColumnNamesOrder(ASTNode ast) throws SemanticException
...
-      if (child.getToken().getType() != HiveParser.TOK_NULLS_FIRST && directionCode == DirectionUtils.ASCENDING_CODE) {
-        throw new SemanticException(
-                "create/alter bucketed table: not supported NULLS LAST for SORTED BY in ASC order");
-      }
-      if (child.getToken().getType() != HiveParser.TOK_NULLS_LAST && directionCode == DirectionUtils.DESCENDING_CODE) {
-        throw new SemanticException(
-                "create/alter bucketed table: not supported NULLS FIRST for SORTED BY in DESC order");
-      }
-      colList.add(new Order(unescapeIdentifier(child.getChild(0).getText()).toLowerCase(), directionCode));
+      Order order = new Order(unescapeIdentifier(child.getChild(0).getText()).toLowerCase(), directionCode);
```
deniskuzZ, line 970: why did we remove nullOrder and direction checks?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -14119,12 +14120,19 @@ ASTNode analyzeCreateTable(
...
       case HiveParser.TOK_ALTERTABLE_BUCKETS:
-        bucketCols = getColumnNames((ASTNode) child.getChild(0));
-        if (child.getChildCount() == 2) {
-          numBuckets = Integer.parseInt(child.getChild(1).getText());
-        } else {
-          sortCols = getColumnNamesOrder((ASTNode) child.getChild(1));
-          numBuckets = Integer.parseInt(child.getChild(2).getText());
+        switch(child.getChildCount()){
+          case 1:
+            sortCols = getColumnNamesOrder((ASTNode) child.getChild(0));
```
deniskuzZ, line 14125: why it's under TOK_ALTERTABLE_BUCKETS?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -14444,6 +14453,22 @@ ASTNode analyzeCreateTable(
     return null;
   }
 
+  private void checkSortCols(List<Order> sortCols, StorageFormat storageFormat) throws SemanticException{
+    if("org.apache.iceberg.mr.hive.HiveIcebergStorageHandler".equalsIgnoreCase(storageFormat.getStorageHandler())){
```
deniskuzZ, line 14457: nit: space

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
```diff
@@ -281,6 +281,11 @@ enum DatabaseType {
   REMOTE = 2
 }
 
+enum NullOrderingType {
```
deniskuzZ, line 284: do we store it in HMS? why changes in thrift?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/Catalogs.java
zratkai: I think I would use  "asJson"  in a method name to express a transformation is happening like toString(), asJson(), asXML ect not in a variable name.

deniskuzZ: i don't insist, just saw this naming is used in iceberg

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
zratkai: Done.

parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g
zratkai: I will create a follow up ticket.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_create_locally_ordered_table.q
```diff
@@ -0,0 +1,44 @@
...
+create table ice_orc (id int, text string) stored by iceberg stored as orc;
+
+insert into ice_orc values (3, "3"),(2, "2"),(4, "4"),(5, "5"),(1, "1"),(2, "3"),(3,null),(2,null),(null,"a");
+
+describe formatted ice_orc;
+describe extended ice_orc;
+set hive.fetch.task.conversion=more;
+select * from ice_orc;
+
+create table ice_orc_sorted (id int, text string) write locally ordered by id desc nulls first, text asc nulls last stored by iceberg stored as orc;
```
okumin, line 33: Can we add one more sort key that uses the default direction and null ordering? I expect ASC + NULLS FIRST is used by default.

```
spark-sql (default)> CREATE TABLE hadoop_prod.default.test_spark (a int) USING iceberg;
Time taken: 1.956 seconds
spark-sql (default)> ALTER TABLE hadoop_prod.default.test_spark WRITE ORDERED BY a;
Time taken: 0.181 seconds
```

```
$ hdfs dfs -cat /user/hive/warehouse/catalog/default/test_spark/metadata/v2.metadata.json
...
  "default-sort-order-id" : 1,
  "sort-orders" : [ {
    "order-id" : 0,
    "fields" : [ ]
  }, {
    "order-id" : 1,
    "fields" : [ {
      "transform" : "identity",
      "source-id" : 1,
      "direction" : "asc",
      "null-order" : "nulls-first"
    } ]
  } ],
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -271,6 +281,31 @@ public void preCreateTable(CreateTableRequest request) {
...
+
+  private void setSortOrder(org.apache.hadoop.hive.metastore.api.Table hmsTable, Schema schema,
+      Properties properties) {
+    String sortOderJSONString = hmsTable.getParameters().get(TableProperties.DEFAULT_SORT_ORDER);
+    SortFields sortFields = null;
+    if (!Strings.isNullOrEmpty(sortOderJSONString)) {
+      try {
+        sortFields = jsonObjectMapper.readValue(sortOderJSONString, SortFields.class);
+      } catch (JsonProcessingException e) {
+        return;
```
okumin, line 295: Can we add warning?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -16089,7 +16097,26 @@ protected void addPartitionColsToInsert(List<FieldSchema> partCols,
...
+      sortFieldDescList.add(new SortFieldDesc(name, sortDirection, nullOrder));
+    }
+    try {
+      return jsonObjectMapper.writeValueAsString(sortFields);
+    } catch (JsonProcessingException e) {
+      return null;
+    }
+    //return new JSONObject().put("fields", new JSONArray(sortFields.stream().map(SortFieldDesc::toString).collect(Collectors.toList()))).toString();
+  }
   @Override
```
okumin, line 16120: We may want a new line here.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -16089,7 +16097,26 @@ protected void addPartitionColsToInsert(List<FieldSchema> partCols,
...
+      String name = unescapeIdentifier(child.getChild(0).getText()).toLowerCase();
+      SortFieldDesc.NullOrder nullOrder = child.getToken().getType() == HiveParser.TOK_NULLS_FIRST ? SortFieldDesc.NullOrder.NULLS_FIRST : SortFieldDesc.NullOrder.NULLS_LAST;
+      sortFieldDescList.add(new SortFieldDesc(name, sortDirection, nullOrder));
+    }
+    try {
+      return jsonObjectMapper.writeValueAsString(sortFields);
+    } catch (JsonProcessingException e) {
+      return null;
+    }
+    //return new JSONObject().put("fields", new JSONArray(sortFields.stream().map(SortFieldDesc::toString).collect(Collectors.toList()))).toString();
```
okumin, line 16118: Can we delete this one?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -16089,7 +16097,26 @@ protected void addPartitionColsToInsert(List<FieldSchema> partCols,
...
+      SortFieldDesc.SortDirection sortDirection = child.getToken()
+          .getType() == HiveParser.TOK_TABSORTCOLNAMEDESC ? SortFieldDesc.SortDirection.DESC : SortFieldDesc.SortDirection.ASC;
+      child = (ASTNode) child.getChild(0);
+      String name = unescapeIdentifier(child.getChild(0).getText()).toLowerCase();
+      SortFieldDesc.NullOrder nullOrder = child.getToken().getType() == HiveParser.TOK_NULLS_FIRST ? SortFieldDesc.NullOrder.NULLS_FIRST : SortFieldDesc.NullOrder.NULLS_LAST;
+      sortFieldDescList.add(new SortFieldDesc(name, sortDirection, nullOrder));
+    }
+    try {
+      return jsonObjectMapper.writeValueAsString(sortFields);
+    } catch (JsonProcessingException e) {
```
okumin, line 16115: Can we add a warning message here?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -14415,7 +14423,7 @@ ASTNode analyzeCreateTable(
         throw new SemanticException(
             "Partition columns can only declared using their names in CTAS statements");
       }
-
+
```
okumin, line 14426: Please clean up this

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
```diff
@@ -115,6 +117,7 @@
 import org.apache.hadoop.security.alias.AbstractJavaKeyStoreProvider;
 import org.apache.hadoop.security.alias.CredentialProvider;
 import org.apache.hadoop.security.alias.CredentialProviderFactory;
+import org.json.JSONArray;
```
okumin, line 120: Could you please revert changes on this file?

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/sortoder/SortFields.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.hadoop.hive.ql.ddl.misc.sortoder;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class SortFields {
+  
+  private List<SortFieldDesc> sortFields;
+
+  public SortFields(){
```
okumin, line 28: I'd add a space after the right paren

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/sortoder/SortFields.java
```diff
@@ -0,0 +1,46 @@
...
+    this.sortFields = new LinkedList<>();
+  }
```
okumin, line 30: I'd insert a line break

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/sortoder/SortFieldDesc.java
```diff
@@ -0,0 +1,73 @@
...
+
+package org.apache.hadoop.hive.ql.ddl.misc.sortoder;
+
+public class SortFieldDesc {
+
+  private String columnName;
+  private NullOrder nullOrder;
+  private SortDirection direction;
+
+  public SortFieldDesc(){
```
okumin, line 27: I'd add a space after the right paren

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
```diff
@@ -966,15 +967,10 @@ protected List<Order> getColumnNamesOrder(ASTNode ast) throws SemanticException
...
-        throw new SemanticException(
-                "create/alter bucketed table: not supported NULLS LAST for SORTED BY in ASC order");
-      }
-      if (child.getToken().getType() != HiveParser.TOK_NULLS_LAST && directionCode == DirectionUtils.DESCENDING_CODE) {
-        throw new SemanticException(
-                "create/alter bucketed table: not supported NULLS FIRST for SORTED BY in DESC order");
-      }
-      colList.add(new Order(unescapeIdentifier(child.getChild(0).getText()).toLowerCase(), directionCode));
+      Order order = new Order(unescapeIdentifier(child.getChild(0).getText()).toLowerCase(), directionCode);
+      order.setNullOrdering(child.getToken().getType()== HiveParser.TOK_NULLS_FIRST?
```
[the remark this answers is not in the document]
zratkai: Not relevant anymore.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_create_locally_ordered_table.q
zratkai: It works currently this way as you mentioned.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
zratkai: Sure.

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
zratkai: Sure.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
zratkai: Sure.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_create_sorted_table.q
zratkai: Iceberg doc uses this semantic, so I would use this.

iceberg/iceberg-handler/src/test/results/positive/create_iceberg_table_stored_by_iceberg_with_serdeproperties.q.out
zratkai: Not relevant anymore.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
zratkai: Not relevant anymore.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
zratkai: Not relevant anymore.

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
zratkai: Not relevant anymore.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -781,7 +802,7 @@ private void setCommonHmsTablePropertiesForIceberg(org.apache.hadoop.hive.metast
    * @param hmsTable Table for which we are calculating the properties
    * @return The properties we can provide for Iceberg functions, like {@link Catalogs}
    */
-  private static Properties getCatalogProperties(org.apache.hadoop.hive.metastore.api.Table hmsTable) {
+  private Properties getCatalogProperties(org.apache.hadoop.hive.metastore.api.Table hmsTable) {
```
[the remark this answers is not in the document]
zratkai: Fixed.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
zratkai: Sure.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
zratkai: Done.

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/sortoder/SortFieldDesc.java
zratkai: Done.

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/sortoder/SortFields.java
zratkai: Done.

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/sortoder/SortFields.java
zratkai: Done.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
zratkai: Done.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_create_locally_ordered_table.q
deniskuzZ, line 33: does it support parenthesis?
````
ordered by (id desc nulls first, text asc nulls last)
````

File: parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g [hunk 28]
```diff
@@ -2201,6 +2212,8 @@ columnNameOrder
             ^(TOK_TABSORTCOLNAMEDESC ^(TOK_NULLS_LAST identifier))
     -> {$orderSpec.tree.getType()==HiveParser.KW_ASC}?
             ^(TOK_TABSORTCOLNAMEASC ^($nullSpec identifier))
+    -> {$orderSpec.tree.getType()==HiveParser.KW_DESC}?
```
deniskuzZ, line 2215: KW_DESC wasn't supported before?

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/sortoder/SortFieldDesc.java
```diff
@@ -0,0 +1,73 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.hive.ql.ddl.misc.sortoder;
+
+public class SortFieldDesc {
```
deniskuzZ, line 21: how so we didn't have it before? :)

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -16089,7 +16097,26 @@ protected void addPartitionColsToInsert(List<FieldSchema> partCols,
...
+      ASTNode child = (ASTNode) ast.getChild(i);
+      SortFieldDesc.SortDirection sortDirection = child.getToken()
+          .getType() == HiveParser.TOK_TABSORTCOLNAMEDESC ? SortFieldDesc.SortDirection.DESC : SortFieldDesc.SortDirection.ASC;
+      child = (ASTNode) child.getChild(0);
+      String name = unescapeIdentifier(child.getChild(0).getText()).toLowerCase();
+      SortFieldDesc.NullOrder nullOrder = child.getToken().getType() == HiveParser.TOK_NULLS_FIRST ? SortFieldDesc.NullOrder.NULLS_FIRST : SortFieldDesc.NullOrder.NULLS_LAST;
+      sortFieldDescList.add(new SortFieldDesc(name, sortDirection, nullOrder));
+    }
+    try {
+      return jsonObjectMapper.writeValueAsString(sortFields);
```
deniskuzZ, line 16114: is there a reason why we use jsonObjectMapper and not byte serialization like
````
conf.set(FILTER_EXPRESSION, SerializationUtil.serializeToBase64(expression))
````

File: ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java [hunk 31]
```diff
@@ -14255,7 +14261,9 @@ ASTNode analyzeCreateTable(
       isExt = isExternalTableChanged(tblProps, isTransactional, isExt, isDefaultTableTypeChanged);
       addDbAndTabToOutputs(new String[] {qualifiedTabName.getDb(), qualifiedTabName.getTable()},
           TableType.MANAGED_TABLE, isTemporary, tblProps, storageFormat);
-
+      if(!Strings.isNullOrEmpty(sortOrder)) {
```
deniskuzZ, line 14264: nit: space

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -137,6 +146,7 @@
 
 public class HiveIcebergMetaHook implements HiveMetaHook {
   private static final Logger LOG = LoggerFactory.getLogger(HiveIcebergMetaHook.class);
+  private final ObjectMapper jsonObjectMapper = new ObjectMapper();
```
deniskuzZ, line 149: can it be static?

deniskuzZ, line 149: see https://stackoverflow.com/questions/3907929/should-i-declare-jacksons-objectmapper-as-a-static-field
````
keep static ObjectMapper somewhere, but only get ObjectReader / ObjectWriter instances (via helper methods), retain references to those in other places (or dynamically call). These reader/writer objects are not only fully thread-safe wrt reconfiguration, but also very light-weight (wrt mapper instances). So keeping thousands of references does not add much memory usage
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java [hunk 33]
```diff
@@ -271,6 +281,32 @@ public void preCreateTable(CreateTableRequest request) {
...
+        sortFields = jsonObjectMapper.readValue(sortOderJSONString, SortFields.class);
+      } catch (JsonProcessingException e) {
+        LOG.warn("Can not read write order json: {}", sortOderJSONString, e);
+        return;
+      }
+      if (sortFields != null && sortFields.getSortFields().size() != 0) {
+        SortOrder.Builder sortOderBuilder = SortOrder.builderFor(schema);
+        sortFields.getSortFields().forEach(fieldDesc -> {
+          NullOrder nullOrder = fieldDesc.getNullOrder() == SortFieldDesc.NullOrder.NULLS_FIRST ?
+              NullOrder.NULLS_FIRST : NullOrder.NULLS_LAST;
```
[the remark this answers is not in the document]
zratkai: Yes.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
zratkai: Fixed.

zratkai (commented):
Please have a look on my replies.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
deniskuzZ: ok, so the idea was to have it readable in tbl properties

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -137,6 +145,7 @@
 
 public class HiveIcebergMetaHook implements HiveMetaHook {
   private static final Logger LOG = LoggerFactory.getLogger(HiveIcebergMetaHook.class);
+  private static final ObjectMapper jsonObjectMapper = new ObjectMapper();
```
deniskuzZ, line 148: static vars / constants should be in upper registry

deniskuzZ (approved):
+1, pending tests

deniskuzZ:
@zratkai please take a look at sonar issues: https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=5541&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
zratkai: Fixed.

okumin (approved):

okumin:
I gave +1. The remaining "Brain Method" of SemanticAnalyzer seems unavoidable.

I retriggered CI because it failed, maybe because of an accident. I'm waiting for it to be completed.

okumin:
@zratkai I reran CI multiple times, but `query_iceberg_metadata_of_unpartitioned_table` always fails. Could you please have a look at if this PR causes it?
https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/activity/?branch=PR-5541

My eyes could not locate the difference.
https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5541/17/tests

PR #5542, opened 2024-11-15
HIVE-27879: Improve the test coverage on package.jdo

dengzhhu653:

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreDirectSqlUtils.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreDirectSqlUtils.java
@@ -20,6 +20,7 @@
 package org.apache.hadoop.hive.metastore;
 
 import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.Function;
 import org.apache.hadoop.hive.metastore.api.MetaException;
@@ -78,7 +79,7 @@ class MetastoreDirectSqlUtils {
       }
       LOG.warn(errorBuilder.toString() + "]", ex);
       // We just logged an exception with (in case of JDO) a humongous callstack. Make a new one.
-      throw new MetaException("See previous errors; " + ex.getMessage() + errorBuilder.toString() + "]");
+      throw new MetaException("See previous errors; " + ExceptionUtils.getRootCauseMessage(ex) + " " +  errorBuilder.toString() + "]");
     }
   }
 
@@ -144,6 +145,7 @@ class MetastoreDirectSqlUtils {
         if (fields == null && !iter.hasNext())
           break;
         long id = entry.getKey();
+        T value = entry.getValue();
         while (fields != null || iter.hasNext()) {
           if (fields == null) {
             fields = iter.next();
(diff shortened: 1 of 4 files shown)


standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestPackageJdo.java
```diff
@@ -0,0 +1,155 @@
...
+    PersistenceManager pm = PersistenceManagerProvider.getPersistenceManager();
+    for (Class clazz : classes) {
+      try {
+        pm.makePersistent(JavaUtils.newInstance(clazz, new Class[0], new Object[0]));
+      } catch (Exception ignore) {
+      }
+    }
+  }
+
+  static class ColumnInfo {
```
deniskuzZ, line 79: should it be private?

dengzhhu653: updated, thank you!

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreDirectSqlUtils.java
```diff
@@ -152,9 +154,18 @@ static <T> int loopJoinOrderedResult(PersistenceManager pm, TreeMap<Long, T> tre
           if (nestedId < id) {
             throw new MetaException("Found entries for unknown ID " + nestedId);
           }
-          if (nestedId > id)
+          if (nestedId > id) {
+            if (!tree.containsKey(nestedId)) {
+              Throwable throwable = (new Throwable()).fillInStackTrace();
```
deniskuzZ, line 159: why do we need to create Throwable?

dengzhhu653: it helps to know what the `many` is (one-to-many mapping), for example if the `value` is Partition, then the `many` can be parameters or partition values.

deniskuzZ: that's fine to log the `value.getClass().getSimpleName()`, however, i don't see benefit of `throwable` here

dengzhhu653: not enough, for example, if the value is a `Partition`, then what are missing for this `nestedId` Partition, it might be the values stored in `PARTITION_KEY_VALS` or the parameters in `PARTITION_PARAMS`, if we shows the message who makes the call, it will direct us to the table quickly to check if this table is corrupted or not.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1556,7 +1556,8 @@ public static String getPartitionName(Table table, Partition partition) {
     try {
       return Warehouse.makePartName(getPartCols(table), partition.getValues());
     } catch (MetaException e) {
-      throw new RuntimeException(e);
+      throw new RuntimeException("Invalid partition found, location is: " +
```
deniskuzZ: minor: simply "location: " ?

deniskuzZ (approved):
+1, pending tests

PR #5543, opened 2024-11-15
HIVE-28598:  NPE in vectorized murmur_hash function on columns with repeating values

ngsg:

`MurmurHashExpressions` now use the first item for every row if `isRepeating` is true.

In current implementation, `MurmurHashExpression` does not check `isRepeating` and always reads whole rows. So if one of `ColumnVector` has `isRepeating == true`, then `MurmurHashExpression` reads the second row and then throws `NullPointerException`.

No

No

Tested using the input data provided on HIVE-28598 JIRA page, and also by TestMiniLlapLocalCliDriver.
However, I failed to reproduce the issue with a file smaller than 11MB. I'll try to reproduce this issue with a smaller input file and update the qfile.


BsoBird (commented):
Hello, I have reviewed your patch. I also used a similar patch to test and fix the issue some time ago. I only fixed MurmurHashStringColStringCol.java, and the repair plan is similar to yours. The NPE indeed disappeared, but my calculation result is missing about 3739 data compared to the correct result. I am currently investigating whether the issue is caused by our hash-result=0 or if there is a data loss issue with Iceberg-vector-read itself...

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/MurmurHashStringColStringCol.java
```diff
@@ -47,12 +47,15 @@ protected void hash(ColumnVector inputColVector1, ColumnVector inputColVector2,
     BytesColumnVector inV1 = (BytesColumnVector) inputColVector1;
     BytesColumnVector inV2 = (BytesColumnVector) inputColVector2;
 
+    int idx1 = inputColVector1.isRepeating ? 0 : i;
+    int idx2 = inputColVector2.isRepeating ? 0 : i;
+
     // hash of value from 1. column
-    int hash = inV1.isNull[i] ? 0
-      : Murmur3.hash32(inV1.vector[i], inV1.start[i], inV1.length[i], Murmur3.DEFAULT_SEED);
+    int hash = inV1.isNull[idx1] ? 0
```
BsoBird, line 54: Hello, I have reviewed your patch. I also used a similar patch to test and fix the issue some time ago. I only fixed MurmurHashStringColStringCol.java, and the repair plan is similar to yours. The NPE indeed disappeared, but my calculation result is missing about 3739 data compared to the correct result. I am currently investigating whether the issue is caused by our hash-result=0 or if there is a data loss issue with Iceberg-vector-read itself...

but maybe i'm wrong

BsoBird (commented):
Leave some minor comments, but let me try again; the patch I tested before has a slight difference from the current one.

BsoBird:
@okumin @ngsg @deniskuzZ @zhangbutao @zabetak 
Hello.
Surprisingly, this patch works well. All NPEs are gone, and the correct results are returned. Thank you for your work!@ngsg
By the way, I'd like to understand why using hash(varchar, varchar) does not trigger an NPE when I'm not using JOIN?
Is there actually something wrong with vectorReader?

BsoBird (approved):
Now I understand, for the case where isRepeating=true, any function or other logic that uses inputVector[] needs to check and be compatible with this scenario. Otherwise, it will trigger an NPE similar to the current one.

LGTM.

BsoBird:
@okumin @ngsg @deniskuzZ @zhangbutao @zabetak
Hello.
Surprisingly, this patch works well. All NPEs are gone, and the correct results are returned. Thank you for your work!@ngsg
By the way, I'd like to understand why using hash(varchar, varchar) does not trigger an NPE when I'm not using JOIN?

zabetak (commented):
The code changes seem reasonable and I understand the suggested fix. However, I don't fully understand the root cause and who is breaking the ColumnVector.

Can we create a .q test with explicit calls to `murmur_hash(c1,c2)` and bunch of repeating values to trigger the NPE? If we don't add coverage this part may break again sometime soon.

If it is not possible to write a .q test, we can get this in as it is but I am just wondering if we examined that possibility.

deniskuzZ:
LGTM, +1 on adding a test for repeating values, see TestMurmurHashExpression

ngsg:
@zabetak @deniskuzZ
I have added a qfile and unit tests for repeating values. Could you please review the added tests? Thank you.

ngsg:
@zabetak
The null values in `BytesColumnVector` come from ORC code.

In the ORC code, ORC reader reset only `vector[]`. [ORC code1](https://github.com/apache/orc/blob/9699c4cbc50bbcc13a5c61623c01763c8b9507c5/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L2326-L2328)
And if the new `ColumnVector` has `isRepeating=true`, the reader updates the first item only. [ORC code2](https://github.com/apache/orc/blob/9699c4cbc50bbcc13a5c61623c01763c8b9507c5/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L2390-L2396)
So the previously used `start[]` and `length[]` are not updated nor reset.

In my opinion, these null values are not a problem because it seems that `isRepeating=true` indicates that only `vector[0]` is valid and the other entries should be ignored.
https://github.com/apache/hive/blob/afe05b968026dfdda631de1e2b090665f0820ef3/storage-api/src/java/org/apache/hadoop/hive/ql/exec/vector/ColumnVector.java#L71-L75

ngsg:
@BsoBird
I don't know the exact answer to your question, but in my experience, the NPE issue did not occur in the following situation:
1. used `hash()` instead of `murmur_hash()`. Both of them appear as `hash` in query plan, but only `murmur_hash` causes the NPE issue.
2. `hive.llap.io.enabled=true` & small inputs. I don't know the exact reason, but I have to disable LLAP IO to reproduce this issue using qtests with a small file.
3. ORC file using direct encoding. If a string column uses direct encoding, then ORC reader always set `isRepeating` to false.
4. length[i] == 0 for all 1<=i<=1023. Then `Murmur3.hash32` does not access `vector[][]` and NPE does not occur.

ql/src/test/queries/clientpositive/vectorized_murmurhash_repeating.q
```diff
@@ -0,0 +1,29 @@
+set hive.llap.io.enabled=false;
+set hive.vectorized.execution.enabled=true;
+
+CREATE TABLE source(id string, id2 string) stored as orc;
+LOAD DATA LOCAL INPATH '../../data/files/2048.orc' overwrite into table source;
```
zabetak, line 5: The .orc are binary files that cannot be easily inspected/reviewed. Moreover, ASF imposes various limitations on the content of binary files so whenever we can avoid them it is good.
Is it possible to use a .csv file as an input and use a staging TEXT table for loading the data to the ORC table? If it's not possible then that's fine we can merge the PR as is.

ql/src/test/queries/clientpositive/vectorized_murmurhash_repeating.q
```diff
@@ -0,0 +1,29 @@
...
+
+-- 2048.orc has 2 columns (id, id2) and 2048 rows.
+-- For the first 1024 rows, both id and id2 are not repeating.
+-- For the last 1024 rows, id is repeating while id2 is not.
+
+-- Test MurmurHashStringColStringCol
+explain
+select sum(murmur_hash(id, id2)) from source;
```
zabetak, line 13: Is the `sum` function necessary for reproducing the problem?

ql/src/test/queries/clientpositive/vectorized_murmurhash_repeating.q
```diff
@@ -0,0 +1,29 @@
...
+
+CREATE TABLE source(id string, id2 string) stored as orc;
+LOAD DATA LOCAL INPATH '../../data/files/2048.orc' overwrite into table source;
+
+-- 2048.orc has 2 columns (id, id2) and 2048 rows.
+-- For the first 1024 rows, both id and id2 are not repeating.
+-- For the last 1024 rows, id is repeating while id2 is not.
+
+-- Test MurmurHashStringColStringCol
+explain
```
zabetak, line 12: The simple explain is not very useful in this case since it doesn't tell us which and if the expected murmur implementation is used. `EXPLAIN VECTORIZATION DETAIL` should show all the necessary info.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestMurmurHashExpression.java
```diff
@@ -241,4 +241,143 @@ public void testMurmurHashWithNullsInt() throws HiveException {
           ((LongColumnVector) vrb.cols[2]).vector[i]);
     }
   }
+
+  @Test
```
zabetak, line 245: There seems to be a bit of duplication in the tests. Do we have anything to factor out?

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestMurmurHashExpression.java
```diff
@@ -241,4 +241,143 @@ public void testMurmurHashWithNullsInt() throws HiveException {
...
+  public void testMurmurHashStringColStringColSingleRepeating() throws HiveException {
+    BytesColumnVector cvString1 = (BytesColumnVector) ColumnVectorGenUtil.generateColumnVector(
+        TypeInfoFactory.getPrimitiveTypeInfo("string"), false, false, SIZE, rand);
+
+    byte[] repeatedValue = new byte[10];
+    rand.nextBytes(repeatedValue);
```
zabetak, line 251: Random makes the tests harder to debug and less predictable so if we can avoid it that would be better.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestMurmurHashExpression.java
```diff
@@ -241,4 +241,143 @@ public void testMurmurHashWithNullsInt() throws HiveException {
...
+
+    Assert.assertEquals(false, vrb.cols[2].isRepeating);
+
+    Text t2 = new Text(repeatedValue);
+    for (int i = 0; i < SIZE; i++) {
+      Text t1 = new Text();
+      t1.set(cvString1.vector[i], cvString1.start[i], cvString1.length[i]);
+
+      Assert.assertEquals(
+          ObjectInspectorUtils.getBucketHashCode(
```
zabetak, line 278: What exactly are we trying to achieve with these assertions? They seem a bit useless since the implementation of the function internally calls this method. If we want to check the correctness of the method then we should hardcode somewhere the expected values and use these in the assert.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestMurmurHashExpression.java
```diff
@@ -241,4 +241,143 @@ public void testMurmurHashWithNullsInt() throws HiveException {
...
+    }
+
+    VectorizedRowBatch vrb = new VectorizedRowBatch(3, SIZE);
+    vrb.cols[0] = cvString1;
+    vrb.cols[1] = cvString2;
+    vrb.cols[2] = new LongColumnVector(SIZE);
+
+    new MurmurHashStringColStringCol(0, 1, 2).evaluate(vrb);
+
+    Assert.assertEquals(false, vrb.cols[2].isRepeating);
```
zabetak, line 270: Why do we have this assertion? Is it something that we want to enforce?

zabetak (commented):
Many thanks for the extra tests. I left some small comments/suggestions.

ql/src/test/queries/clientpositive/vectorized_murmurhash_repeating.q
ngsg: Sure we can. I changed qfile to create ORC table from text file and removed ORC file.

ql/src/test/queries/clientpositive/vectorized_murmurhash_repeating.q
ngsg: It's not necessary; I just used `sum` to check the result quickly. Removed from the qfile.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestMurmurHashExpression.java
ngsg: Yes, it ensures that `MurmurHash` produces `isRepeating=false` ColumnVector for `isRepeating=true` and `isRepeating=false` ColumnVectors. Currrent tests check `true, true`  and `false, false` cases only.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestMurmurHashExpression.java
ngsg: removed Random from the tests.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestMurmurHashExpression.java
ngsg: I agree that the values we are comparing are computed in the same way. Removed calling `getBucketHashCode` and modified the tests to compare against some hardcoded values.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/expressions/TestMurmurHashExpression.java
ngsg: I refactored `ColumnVector` creation part to reduce code redundancy and stop using random vector.

ql/src/test/results/clientpositive/llap/vectorized_murmurhash_repeating.q.out
```diff
@@ -0,0 +1,8374 @@
...
+
+STAGE PLANS:
+  Stage: Stage-0
+    Fetch Operator
+      limit: -1
+      Processor Tree:
+        TableScan
+          alias: source
+          Select Operator
+            expressions: hash(id,id2) (type: int)
```
zabetak, line 60: In this and all subsequent explains we can see that the vectorized murmur implementation is not used. In other words, the tests here do not cover the problematic NPE case that was reported. This can be easily seen since this test pass cleanly on apache master without the changes in this PR.

ql/src/test/results/clientpositive/llap/vectorized_murmurhash_repeating.q.out
```diff
@@ -0,0 +1,8374 @@
...
+
+PREHOOK: query: select murmur_hash(id, id2) from source
+PREHOOK: type: QUERY
+PREHOOK: Input: default@source
+#### A masked pattern was here ####
+POSTHOOK: query: select murmur_hash(id, id2) from source
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@source
+#### A masked pattern was here ####
+-2117694071
```
zabetak, line 72: Give that the result set is very large lets add some limit to the respective queries or some other kind of method to reduce the result set. I suppose you used the sum function for this purpose so now I see why it makes sense.

ql/src/test/results/clientpositive/llap/vectorized_murmurhash_repeating.q.out
ngsg: Sorry for the wrong test; I thought `sum()` is irrelevant to qtests so I missed checking the qtest output. It seems that we need `sum()` to enforce vectorization and using `murmur_hash()`. I reverted the query to use sum().

ql/src/test/results/clientpositive/llap/vectorized_murmurhash_repeating.q.out
ngsg: reverted to sum as I explained above

zabetak (approved):
LGTM! I will merge this once tests come back green.

PR #5544, opened 2024-11-15
HIVE-28626: Display MoveTask/StatsTask duration on the query summary

abstractdog:
Display MoveTask's and StatsTask's duration on the console.

Better observability.

Yes, query output on the console slightly changes.

No.

On cluster. Unfortunately, due to HIVE-28627, the console message is broken, works only intermittently, but in the logs, the very same messages appear.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/MoveTask.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/MoveTask.java
@@ -78,6 +78,7 @@ import org.apache.hadoop.hive.ql.plan.api.StageType;
 import org.apache.hadoop.hive.ql.session.SessionState;
 import org.apache.hadoop.hive.ql.util.DirectionUtils;
 import org.apache.hadoop.util.StringUtils;
+import org.apache.hadoop.util.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -688,7 +689,7 @@ public class MoveTask extends Task<MoveWork> implements Serializable {
     Map<Path, Utilities.PartitionDetails> dps = Utilities.getFullDPSpecs(conf, dpCtx, dynamicPartitionSpecs);
 
     console.printInfo(System.getProperty("line.separator"));
-    long startTime = System.currentTimeMillis();
+    long startTime = Time.monotonicNow();
     // load the list of DP partitions and return the list of partition specs
     // TODO: In a follow-up to HIVE-1361, we should refactor loadDynamicPartitions
     // to use Utilities.getFullDPSpecs() to get the list of full partSpecs.
@@ -714,7 +715,7 @@ public class MoveTask extends Task<MoveWork> implements Serializable {
     }
 
     String loadTime = String.format("Time taken to load dynamic partitions:\t %.3f seconds",
-        (System.currentTimeMillis() - startTime) / 1000.0);
+        (Time.monotonicNow() - startTime) / 1000.0);
     console.printInfo(loadTime);
     LOG.info(loadTime);
 
@@ -723,7 +724,7 @@ public class MoveTask extends Task<MoveWork> implements Serializable {
           " To turn off this error, set hive.error.on.empty.partition=false.");
     }
 
-    startTime = System.currentTimeMillis();
+    startTime = Time.monotonicNow();
     // for each partition spec, get the partition
     // and put it to WriteEntity for post-exec hook
     for(Map.Entry<Map<String, String>, Partition> entry : dp.entrySet()) {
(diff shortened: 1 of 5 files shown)


ql/src/java/org/apache/hadoop/hive/ql/exec/MoveTask.java
```diff
@@ -569,7 +569,8 @@ public int execute() {
         }
         releaseLocks(tbd);
       }
-
+      long moveFilesDuration = perfLogger.getDuration(PerfLogger.FILE_MOVES);
```
InvisibleProgrammer, line 572: Two questions about this: 

1.: This method already contains a perfLogEnd call: 
```java 
      // if _blob_files_kept is present, use it to move the files. Else fall back to normal case.
      if (moveFilesUsingManifestFile(fs, sourcePath, targetPath)) {
        perfLogger.perfLogEnd("MoveTask", PerfLogger.FILE_MOVES);
        return;
      }
```

Does it stops the perfLogger? If yes, how that change will behave in that case? 

2: perfLogEnd is already called at line 196. Can it affect the output of this getDuration call?

abstractdog: 1. it doesn't stop, just registers and endTime for the duration of FILE_MOVES
2. yeah, depending on the outcome, the ending time should still be registered anyways

InvisibleProgrammer (approved):
LGTM

kokila-19 (approved):

ql/src/java/org/apache/hadoop/hive/ql/exec/MoveTask.java
```diff
@@ -758,8 +759,8 @@ private DataContainer handleDynParts(Hive db, Table table, LoadTableDesc tbd,
       }
       LOG.info("Loading partition " + entry.getKey());
     }
-    console.printInfo("\t Time taken for adding to write entity : " +
-        (System.currentTimeMillis() - startTime)/1000.0 + " seconds");
+    console.printInfo(String.format("Time taken for adding to write entity:\t %.3f seconds",
+        (System.currentTimeMillis() - startTime) / 1000.0));
```
ayushtkn, line 763: this entire class does this not something new, but still I will take the opportunity here to say. ``System.currentTimeMillis()`` isn't the best thing to do when calculating elaspsed time, it is subject to wall clock changes, Should use ``nanoTime()`` always, or Time.monotonicNow()

https://github.com/apache/hadoop/blob/31b8b4a8bd5602e04e780776acf878dacda01656/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/Time.java#L49-L69

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
```diff
@@ -102,7 +102,6 @@ public BasicStatsTask(HiveConf conf, BasicStatsWork work) {
 
   @Override
   public int process(Hive db, Table tbl) throws Exception {
-
```
ayushtkn, line 105: avoid

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -5047,7 +5047,7 @@ private static void deleteAndRename(FileSystem destFs, Path destFile, FileStatus
     try {
       // rename cannot overwrite non empty destination directory, so deleting the destination before renaming.
       destFs.delete(destFile);
-      LOG.info("Deleted destination file" + destFile.toUri());
+      LOG.info("Deleted destination file: " + destFile.toUri());
```
ayushtkn, line 5050: Since you are touching it, can you change it to
LOG.info("Deleted destination file: {}", destFile.toUri());

ayushtkn (approved):
Thanx Laszlo for the PR, minor nits, rest LGTM

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
abstractdog: ack

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
abstractdog: intentionally removed the line break from this method, let me know if you have any concerns

ql/src/java/org/apache/hadoop/hive/ql/exec/MoveTask.java
abstractdog: right, we're doing this currentTimeMillis thing all the time :D
I agree, let me fix this as I'm touching this part now

deniskuzZ:
@ayushtkn, should we merge it?

abstractdog:
> @ayushtkn, should we merge it?

abstractdog:
rebased on master

ayushtkn:
> @ayushtkn, should we merge it?

yep, if the build is still green

PR #5548, opened 2024-11-17
HIVE-28631: Close opened iterables in HiveDeleteFilter to stop resource leak.

harishjp:

The Iterables created in HiveDeleteFilter are CloseableIterators and it actually opens multiple files during iteration, if left behind especially without finishing iteration it leaves opened resources behind.

No.

No.

Tested by deploying this change in a cluster and running a large merge query accessing files in HiveDeleteFilter across multiple mappers.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/vector/HiveDeleteFilter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/vector/HiveDeleteFilter.java
@@ -22,6 +22,7 @@ package org.apache.iceberg.mr.hive.vector;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.NoSuchElementException;
+import org.apache.commons.io.IOUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.iceberg.FileScanTask;
@@ -142,7 +143,7 @@ public class HiveDeleteFilter extends DeleteFilter<HiveRow> {
 
           @Override
           public void close() throws IOException {
-            srcIterator.close();
+            IOUtils.close(srcIterator, deleteInputIterable, deleteOutputIterable);
           }
         };
       }


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/vector/HiveDeleteFilter.java
```diff
@@ -142,7 +143,7 @@ public HiveBatchContext next() {
 
           @Override
           public void close() throws IOException {
-            srcIterator.close();
+            IOUtils.close(srcIterator, deleteInputIterable, deleteOutputIterable);
```
abstractdog, line 146: looks good to me, however I would suggest an additional minor refactor:
deleteInputIterable, deleteOutputIterable are defined outside of new CloseableIterable<HiveBatchContext>. in the filterBatch, can you move them as final fields into the CloseableIterable, emphasizing that it owns these iterables, so it's assumed to close them also
otherwise, their closure makes sense

deniskuzZ (changes requested):
this method is not even called upstream!

harishjp:
> this method is not even called upstream!

I'm sorry, I do not understand which method is not called upstream? close or filterBatch?

deniskuzZ:
> > This method is not even called upstream!
> 
> I'm sorry, I do not understand which method is not called upstream? close or filterBatch?

CloseableIterator#close(). Instead CloseableIterable#close is invoked. 
I've refactored the IcebergRecordReader#nextKeyValue not to use iceber's `CloseableIterable.concat` and looking into positional-delete readers leak - seems to be some race condition in deletes loading. 

I can repro the issue locally, and I do not see how proposed change even supposed to help. `deleteInputIterable` is just a closeable wrapper over `srcIterator`, and `deleteOutputIterable` is the result of filter transform on the same `srcIterator`. Closing `srcIterator` should be enough to release the data-files connection.

harishjp:
> > > This method is not even called upstream!
> > 
> > 
> > I'm sorry, I do not understand which method is not called upstream? close or filterBatch?
> 
> CloseableIterator#close(). Instead CloseableIterable#close is invoked. I've refactored the IcebergRecordReader#nextKeyValue not to use iceber's `CloseableIterable.concat` and looking into positional-delete readers leak - seems to be some race condition in deletes loading.
> 
> I can repro the issue locally, and I do not see how proposed change even supposed to help. `deleteInputIterable` is just a closeable wrapper over `srcIterator`, and `deleteOutputIterable` is the result of filter transform on the same `srcIterator`. Closing `srcIterator` should be enough to release the data-files connection. Screenshot 2024-11-17 at 18 45 37

Thank you for the clarification, yes you are right. This patch is not required for master. I do not know if there is a leak in the master, I'll close this ticket and patch.

PR #5550, opened 2024-11-18
HIVE-28632: Fix issues in JSON SerDe implementations related to Boolean, Binary

armitage420:

1. Ensure case-sensitive handling of boolean strings in JSON deserialization and resolve inconsistencies between INSERT statements and direct JSON loading.
2. Implement automatic detection and appropriate deserialization of JSON nodes, distinguishing between base64-encoded and raw strings.

The changes are needed to follow SQL standards by treating any string other than "false" (case-insensitive) as true and to resolve inconsistencies between outputs when using an INSERT statement and loading a JSON value directly. Additionally, Hive needs to automatically detect if a JSON node is base64-encoded and handle it accordingly, ensuring proper deserialization of both base64-encoded and raw strings.
[Steps to reproduce the issue and more details mentioned in jira description: [HIVE-28632](https://issues.apache.org/jira/browse/HIVE-28632)]

Yes, this PR introduces user-facing changes:

Boolean Handling: The JSON deserializer will now treat any string other than "false" (case-insensitive) as true. This change ensures compliance with SQL standards and resolves inconsistencies between INSERT statements and direct JSON loading.

Binary Handling: Hive will automatically detect if a JSON node is base64-encoded and handle it accordingly. This ensures proper deserialization of both base64-encoded and raw strings, improving user experience and consistency.

No

My local machine

### DIFF
--- a/serde/src/java/org/apache/hadoop/hive/serde2/JsonSerDe.java
+++ b/serde/src/java/org/apache/hadoop/hive/serde2/JsonSerDe.java
@@ -139,7 +139,6 @@ public class JsonSerDe extends AbstractSerDe {
     this.jsonReader = new HiveJsonReader(this.soi, tsParser);
     this.jsonWriter = new HiveJsonWriter(this.binaryEncoding, getColumnNames());
 
-    this.jsonReader.setBinaryEncoding(binaryEncoding);
     this.jsonReader.enable(HiveJsonReader.Feature.COL_INDEX_PARSING);
 
     if (writeablePrimitivesDeserialize) {
--- a/serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
+++ b/serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
@@ -112,8 +112,10 @@ public class HiveJsonReader {
   private final ObjectMapper objectMapper;
 
   private final TimestampParser tsParser;
-  private BinaryEncoding binaryEncoding;
   private final ObjectInspector oi;
+  private static final Pattern BASE64_PATTERN = Pattern.compile(
+          "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$"
+  );
 
   /**
    * Enumeration that defines all on/off features for this reader.
@@ -168,7 +170,6 @@ public class HiveJsonReader {
    * @param tsParser Custom timestamp parser
    */
   public HiveJsonReader(ObjectInspector oi, TimestampParser tsParser) {
-    this.binaryEncoding = BinaryEncoding.BASE64;
     this.tsParser = tsParser;
     this.oi = oi;
     this.objectMapper = new ObjectMapper();
@@ -397,7 +398,11 @@ public class HiveJsonReader {
     case LONG:
       return Long.valueOf(leafNode.asLong());
     case BOOLEAN:
-      return Boolean.valueOf(leafNode.asBoolean());
+      if ("false".equalsIgnoreCase(leafNode.asText())) {
+        return Boolean.FALSE;
+      } else {
+        return Boolean.valueOf(leafNode.asBoolean(true));
+      }
     case FLOAT:
       return Float.valueOf((float) leafNode.asDouble());
     case DOUBLE:
@@ -450,20 +455,34 @@ public class HiveJsonReader {
    */
   private byte[] getByteValue(final JsonNode binaryNode) throws SerDeException {
     try {
-      switch (this.binaryEncoding) {
+      BinaryEncoding binaryEncoding = getBinaryEncodingForNode(binaryNode);
+      switch (binaryEncoding) {
       case RAWSTRING:
-        final String byteText = binaryNode.textValue();
+        final String byteText = binaryNode.asText();
+        if (byteText == null) {
+          return null;
+        }
         return byteText.getBytes(StandardCharsets.UTF_8);
       case BASE64:
         return binaryNode.binaryValue();
       default:
         throw new SerDeException(
-            "No such binary encoding: " + this.binaryEncoding);
+            "No such binary encoding: " + binaryEncoding);
       }
     } catch (IOException e) {
       throw new SerDeException("Error generating JSON binary type from record.",
           e);
     }
+  }
+
+  private BinaryEncoding getBinaryEncodingForNode(JsonNode binaryNode) {
+    String jsonValue = binaryNode.textValue();
+
+    if (jsonValue == null || jsonValue.length() % 4 != 0 || !BASE64_PATTERN.matcher(jsonValue).matches()) {
+      return BinaryEncoding.RAWSTRING;
+    }
+
+    return BinaryEncoding.BASE64;
   }
 
   /**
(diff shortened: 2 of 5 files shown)


armitage420:
@ayushtkn @abstractdog @deniskuzZ Could you please help with the review if/when you have the time? Thank you!

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
```diff
@@ -397,7 +400,11 @@ private Object visitLeafNode(final JsonNode leafNode,
     case LONG:
       return Long.valueOf(leafNode.asLong());
     case BOOLEAN:
-      return Boolean.valueOf(leafNode.asBoolean());
+      if ("false".equalsIgnoreCase(leafNode.asText())) {
```
deniskuzZ, line 403: why do we need a string match here, won't the `Boolean.valueOf(leafNode.asBoolean(true))` handle both cases

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
```diff
@@ -466,6 +476,16 @@ private byte[] getByteValue(final JsonNode binaryNode) throws SerDeException {
...
   }
 
+  private BinaryEncoding getBinaryEncodingForNode(JsonNode binaryNode) {
+    String jsonValue = binaryNode.textValue();
+
+    if(jsonValue == null || jsonValue.length() % 4 != 0 || !BASE64_PATTERN.matcher(jsonValue).matches()) {
+      return BinaryEncoding.RAWSTRING;
+    }
+
+    return BinaryEncoding.BASE64;
```
deniskuzZ, line 486: return `binaryEncoding`?

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
```diff
@@ -450,9 +457,12 @@ private Object visitLeafNode(final JsonNode leafNode,
    */
   private byte[] getByteValue(final JsonNode binaryNode) throws SerDeException {
     try {
-      switch (this.binaryEncoding) {
+      switch (getBinaryEncodingForNode(binaryNode)) {
```
deniskuzZ, line 460: please extract to local var `binaryEncoding `

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
```diff
@@ -466,6 +476,16 @@ private byte[] getByteValue(final JsonNode binaryNode) throws SerDeException {
     }
   }
 
+  private BinaryEncoding getBinaryEncodingForNode(JsonNode binaryNode) {
+    String jsonValue = binaryNode.textValue();
+
+    if(jsonValue == null || jsonValue.length() % 4 != 0 || !BASE64_PATTERN.matcher(jsonValue).matches()) {
```
deniskuzZ, line 482: nit: space

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
armitage420: Thank you so for the review! 
Dropped this variable from HiveJsonReader as here we are detecting the encoding automatically. Although, HiveJsonWriter will use BASE64 encoding to serialized data for insert statements.

armitage420:
@deniskuzZ I have made the required changes, and the tests have passed in my local. Although, seems like the ci integration test is skipped and I do have the permission to re-trigger it.

armitage420:
@deniskuzZ The tests have passed, I hope you could check when you have the time! Thanks!

serde/src/java/org/apache/hadoop/hive/serde2/JsonSerDe.java
```diff
@@ -139,7 +139,6 @@ private void initialize(final Configuration conf, final Properties tbl,
     this.jsonReader = new HiveJsonReader(this.soi, tsParser);
     this.jsonWriter = new HiveJsonWriter(this.binaryEncoding, getColumnNames());
 
-    this.jsonReader.setBinaryEncoding(binaryEncoding);
```
deniskuzZ, line 142: can we drop the `binaryEncoding` var as well?

armitage420: @deniskuzZ Thanks for the review again!

Before [HIVE-21240](https://issues.apache.org/jira/browse/HIVE-21240), RAWSTRING was the default `binaryEncoding,` but it caused serialization issues. While most of these were addressed with precondition checks, switching to BASE64 binaryEncoding provides better handling for cases involving, say, special characters.

Moreover, if we decide to use RAWSTRING again in the future, the change can be made easily by updating a single variable, ensuring flexibility.

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
```diff
@@ -450,22 +455,36 @@ private Object visitLeafNode(final JsonNode leafNode,
...
   private byte[] getByteValue(final JsonNode binaryNode) throws SerDeException {
     try {
-      switch (this.binaryEncoding) {
+      BinaryEncoding binaryEncoding = getBinaryEncodingForNode(binaryNode);
+      switch (binaryEncoding) {
       case RAWSTRING:
-        final String byteText = binaryNode.textValue();
+        final String byteText = binaryNode.asText();
+        if (byteText == null) {
+          return null;
```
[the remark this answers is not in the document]
armitage420: Returning an empty array instead of null can compromise data consistency and risk misinterpreting the data in ways the user may not intend. I don't think we should do so.

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
armitage420: @deniskuzZ  `HiveJsonWriter` is used for serializing values to JSON during `INSERT` statements, so using a single encoding type should suffice. If future bugs arise (though no such scenarios are currently identified), we can consider supporting both encoding types as needed.

serde/src/java/org/apache/hadoop/hive/serde2/JsonSerDe.java
deniskuzZ: sorry, my bad, I haven't noticed that we pass encoding to the writer as well

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
deniskuzZ: wait, why do we need to determine the encoding if we could simply pass the encoding used in the writer as it was done originally?

armitage420: The encoding needs to be determined because loading JSON directly (e.g., using `LOAD DATA INPATH`) causes issues. Users must encode their JSON to BASE64 beforehand; otherwise, an error is thrown. This behavior is inconsistent since we previously supported deserializing RAWSTRING JSON values provided by users.

Steps to reproduce this issue and the inconsistency are detailed in the JIRA description: [HIVE-28632](https://issues.apache.org/jira/browse/HIVE-28632).

deniskuzZ (approved):
LGTM +1

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
deniskuzZ: yes, I got that. updated the comment

armitage420: My bad, I just saw the updated comment. You are right. 
Thank you so much for the review again!

PR #5559, opened 2024-11-25
HIVE-28029: Make unit tests based on TxnCommandsBaseForTests/DbTxnManagerEndToEndTestBase run on Tez

kasakrisz:
Run tests derived from `TxnCommandsBaseForTests` and `DbTxnManagerEndToEndTestBase` using `tez` execution engine instead of `mr`.
The test configurations are provided by `HiveConfForTest`.
Added
```
    hiveConf.set("tez.grouping.max-size", "10");
    hiveConf.set("tez.grouping.min-size", "1");
```
because lots of test expects more than one buckets.

`Tez` is the default execution engine in Hive and `mr` is deprecated

It affects mostly tests.
Tez diagnostics status can contain locking and heart-beating related error messages.

No.

```
mvn test -Dtest=TestDbTxnManager2,TestDbTxnManagerIsolationProperties -pl ql
mvn test -Dtest=TestTxnAddPartition,TestTxnCommands,TestTxnCommands2,TestTxnCommands3,TestTxnCommandsForMmTable,TestTxnConcatenate,TestTxnExIm,TestTxnLoadData,TestTxnNoBuckets,TestMSCKRepairOnAcid,TestHiveStrictManagedMigration,TestUpgradeTool -pl ql
```


ql/src/test/org/apache/hadoop/hive/ql/TezBaseForTests.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hive.ql;
+
+import org.apache.hadoop.hive.conf.HiveConf;
+
+import java.io.File;
+
+public abstract class TezBaseForTests {
```
deniskuzZ, line 24: can we reuse it in CompactorOnTezTest#setupTez?

abstractdog: I think that's what HiveConfForTest is for, see my comment above

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java
```diff
@@ -280,6 +280,7 @@ public int monitorExecution() {
             // best effort
           }
           console.printError("Execution has failed. stack trace: " + ExceptionUtils.getStackTrace(e));
+          diagnostics.append(e.getMessage());
```
deniskuzZ, line 283: is that an intentional change? just asking

ql/src/test/org/apache/hadoop/hive/ql/TestTxnAddPartition.java
```diff
@@ -63,15 +63,15 @@ protected String getTestDataDir() {
 
 
   @Test
-  public void addPartition() throws Exception {
-
-    addPartition(false);
+  public void testAddPartition() throws Exception {
+    hiveConf.setBoolVar(HiveConf.ConfVars.HIVE_VECTORIZATION_ENABLED, false);
```
deniskuzZ, line 67: why not do the config set in `addPartition` itself?

ql/src/test/org/apache/hadoop/hive/ql/TestTxnAddPartition.java
```diff
@@ -261,9 +259,9 @@ public void addPartitionRename() throws Exception {
     List<String> rs = runStatementOnDriver(
         "select ROW__ID, p, a, b, INPUT__FILE__NAME from T order by p, ROW__ID");
     String[][] expected = new String[][]{
+            {"{\"writeid\":1,\"bucketid\":536870912,\"rowid\":0}\t0\t2\t4",
```
deniskuzZ, line 262: where is the (2,4) ?

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
```diff
@@ -1335,6 +1319,7 @@ public void testInitiatorWithMinorCompactionForInsertOnlyTable() throws Exceptio
    * Make sure there's no FileSystem$Cache$Key leak due to UGI use
    * @throws Exception
    */
+  @Ignore("FileSystem.closeAllForUGI is never called from Compaction related threads")
```
deniskuzZ, line 1322: if that is the case, why do we need this test at all?

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
```diff
@@ -2450,6 +2435,9 @@ public void testCleanerForTxnToWriteId() throws Exception {
     // Keep an open txn which refers to the aborted txn.
     Context ctx = new Context(hiveConf);
     HiveTxnManager txnMgr = TxnManagerFactory.getTxnManagerFactory().getTxnManager(hiveConf);
+    // Txn is not considered committed or aborted until TXN_OPENTXN_TIMEOUT expires
+    // See MinOpenTxnIdWaterMarkFunction, OpenTxnTimeoutLowBoundaryTxnIdHandler
+    waitUntilAllTxnFinished();
```
deniskuzZ, line 2440: why do we need to wait for all txn to commit here?

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
```diff
@@ -2805,7 +2793,7 @@ public void testDynPartUpdateWithAborts() throws Exception {
     hiveConf.setBoolVar(HiveConf.ConfVars.HIVE_TEST_MODE_FAIL_LOAD_DYNAMIC_PARTITION, true);
     runStatementOnDriverWithAbort("update " + Table.ACIDTBLPART + " set b=a+2 where a<5");
     hiveConf.setBoolVar(HiveConf.ConfVars.HIVE_TEST_MODE_FAIL_LOAD_DYNAMIC_PARTITION, false);
-    verifyDeltaDirAndResult(2, Table.ACIDTBLPART.toString(), "p=p1", resultData1);
+    verifyDeltaDirAndResult(3, Table.ACIDTBLPART.toString(), "p=p1", resultData1);
```
deniskuzZ, line 2796: why do we have changes of expected results after the engine change?

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java
```diff
@@ -91,8 +91,8 @@ public void testRenameTable() throws Exception {
             "s/delta_0000001_0000001_0000/bucket_00000_0"},
         {"{\"writeid\":2,\"bucketid\":536870913,\"rowid\":0}\t4\t6",
             "s/delta_0000002_0000002_0001/bucket_00000_0"}};
-    checkResult(expected, testQuery, false, "check data", LOG);
-
+    List<String> rs = runStatementOnDriver(testQuery);
```
deniskuzZ, line 94: should the `checkResult` run the query and do the validation?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/MinOpenTxnIdWaterMarkFunction.java
```diff
@@ -61,7 +61,7 @@ public Long execute(MultiDataSourceJdbcResource jdbcResource) throws MetaExcepti
           return id;
         }));
     long lowWaterMark = jdbcResource.execute(new OpenTxnTimeoutLowBoundaryTxnIdHandler(openTxnTimeOutMillis));
-    LOG.debug("MinOpenTxnIdWaterMark calculated with minOpenTxn {}, lowWaterMark {}", minOpenTxn, lowWaterMark);
+    LOG.info("MinOpenTxnIdWaterMark calculated with minOpenTxn {}, lowWaterMark {}", minOpenTxn, lowWaterMark);
```
deniskuzZ, line 64: i don't think we should be changing to info here, it's a debug information

ql/src/test/org/apache/hadoop/hive/ql/TestTxnNoBuckets.java
```diff
@@ -429,33 +424,33 @@ now that T is Acid, data for each writerId is treated like a logical bucket (tho
...
     runStatementOnDriver("update T set b = 88 where b = 80");
-    assertVectorized(shouldVectorize(), "delete from T where b = 8");
+    assertVectorized("delete from T where b = 8");
     runStatementOnDriver("delete from T where b = 8");
     String expected3[][] = {
-        {"{\"writeid\":0,\"bucketid\":537001984,\"rowid\":0}\t1\t2",  "warehouse/t/000002_0"},
-        {"{\"writeid\":0,\"bucketid\":537001984,\"rowid\":1}\t2\t4",  "warehouse/t/000002_0"},
-        {"{\"writeid\":0,\"bucketid\":536870912,\"rowid\":0}\t5\t6",  "warehouse/t/000000_0"},
-        {"{\"writeid\":0,\"bucketid\":536870912,\"rowid\":1}\t9\t10", "warehouse/t/000000_0"},
+        {"{\"writeid\":0,\"bucketid\":536870912,\"rowid\":1}\t1\t2",  "warehouse/t/HIVE_UNION_SUBDIR_1/000000_0"},
```
deniskuzZ, line 448: now we have just 1 bucket?

ql/src/test/org/apache/hadoop/hive/ql/TestTxnAddPartition.java
kasakrisz: The default value of `HiveConf.ConfVars.HIVE_VECTORIZATION_ENABLED` is `true` and this test doesn't change this at setup.

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java
abstractdog: while I find this very useful, a bit confused why it's needed, is there a chance you can repro a failure to see how is the error messages are handled?
I mean:
https://github.com/apache/hive/blob/cf83d4751271622a9a700c6f2330dbfff38801d2/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java#L296
so in the finally block, the final dagStatus (given by the tez am) also contains diagnostics, which is then printed to the console and also added to diagnostics
so this change assumes that there is an error which is swallowed here and is not part of the final dagStatus
if that's the case, it's okay append to the diagnostics, we need to just understand what happened

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
kasakrisz: The next command is `txnMgr.openTxn(ctx, "u1");`. During txn opening a new record is inserted into `MIN_HISTORY_LEVEL`
https://github.com/apache/hive/blob/cf83d4751271622a9a700c6f2330dbfff38801d2/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/OpenTxnsFunction.java#L150
It has an impact cleaning records from `TXN_TO_WRITE_ID`

The column value `MHL_MIN_OPEN_TXNID` is coming from 
https://github.com/apache/hive/blob/cf83d4751271622a9a700c6f2330dbfff38801d2/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/OpenTxnsFunction.java#L101C26-L101C55
It is calculated by taking the smaller value from `minOpenTxn` and `lowWaterMark`. `minOpenTxn` is the last open txn id which his null so we go with `lowWaterMark`.
`lowWaterMark` is queried by 
https://github.com/apache/hive/blob/cf83d4751271622a9a700c6f2330dbfff38801d2/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/queries/OpenTxnTimeoutLowBoundaryTxnIdHandler.java#L42-L43

Maybe my test computer is running the queries with TEZ faster than MR and `openTxnTimeOutMillis` is not expired.
When I added `Thread.sleep(openTxnTimeOutMillis)` the test passed so I assume this is a timing issue.

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
kasakrisz: In this test we make the update statement fail explicitly by setting `HIVE_TEST_MODE_FAIL_LOAD_DYNAMIC_PARTITION` to `true`.
With Tez the jobs has a chance to finish their work and both the delta and delete delta directories and buckets are written hence we end up with 3 deltas and 1 delete delta.
With MR the insert job fails before creating the delta directory so we end up with 2 deltas only.
The txn is marked failed and rolled back with both execution engines.

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java
kasakrisz: `checkresults` also checks whether mappers are vectorized. It is not required here.
We can extract it to a new method but in that case I would rename the current `checkresults` to something like `checkResultsAndVectorization` first but it is widely used.
WDYT?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/MinOpenTxnIdWaterMarkFunction.java
kasakrisz: Rolled back

ql/src/test/org/apache/hadoop/hive/ql/TestTxnNoBuckets.java
kasakrisz: No. We still have several buckets. They are in separate sub directories though but `HIVE_UNION_SUBDIR_2` for has 2
t/HIVE_UNION_SUBDIR_2/000000_0
t/HIVE_UNION_SUBDIR_2/000001_0

IIUC Hive compiles the union operator in a different way with Tez hence the `HIVE_UNION_SUBDIR_`s.

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java
```diff
@@ -197,10 +193,10 @@ private void testDeleteEventPruning() throws Exception {
 
     String[][] expected2 = new String[][]{
         {"{\"writeid\":1,\"bucketid\":536870912,\"rowid\":1}\t4\t5",
-            "warehouse/t/base_0000001/bucket_00000"},
+            "warehouse/t/base_0000003_v0000012/bucket_00000"},
```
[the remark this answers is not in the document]
kasakrisz: It was never `base_0000001` even with MR but this was never asserted because this test always runs in vectorized mode so `INPUT__FILE__NAME` was not queried.
I altered the test query by adding `INPUT__FILE__NAME` since it is no longer a show stopper for vectorization and I had to change the expected base directory name accordingly.

ql/src/test/org/apache/hadoop/hive/ql/TezBaseForTests.java
```diff
@@ -0,0 +1,46 @@
...
+  protected static final String TEST_DATA_DIR = new File(
+          System.getProperty("java.io.tmpdir") + File.separator +
+                  TezBaseForTests.class.getCanonicalName() + "-" + System.currentTimeMillis())
+          .getPath().replaceAll("\\\\", "/");
+
+  protected void setupTez(HiveConf conf) {
```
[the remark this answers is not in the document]
kasakrisz: I switched to `HiveConfForTest` and it works.

ql/src/test/org/apache/hadoop/hive/ql/TezBaseForTests.java
kasakrisz: Switched to HiveConfForTest.

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java
kasakrisz: The code you pointed is about copying messages from the `DAGStatus` instance. However in this special test case  this instance is `null` because the `LockException` is thrown before pulling the status.
https://github.com/apache/hive/blob/cf83d4751271622a9a700c6f2330dbfff38801d2/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java#L171

https://github.com/apache/hive/blob/cf83d4751271622a9a700c6f2330dbfff38801d2/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java#L183

So it seems that heartbeat related checks are handled independently from the tez jobs status and the response coming from the first one is not fed into diagnostics.

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
kasakrisz: It makes sense to remove it. Done

ql/src/test/org/apache/hadoop/hive/ql/TestTxnAddPartition.java
deniskuzZ: i see you added the same for `testAddPartitionMM`. 👍

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
deniskuzZ: 👍

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
deniskuzZ: ok

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java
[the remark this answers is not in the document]
deniskuzZ: 👍

ql/src/test/org/apache/hadoop/hive/ql/TestTxnNoBuckets.java
deniskuzZ: i do not see `000001_0` in asserts only "bucketid\":536870912
after compaction also only `bucket_00000`

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java
deniskuzZ: maybe we can keep overloaded version without vectorization check, however, I am ok with rename as well

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
deniskuzZ: this part is a simulation of the following sequence of events:
````
commit-abort-open-abort-commit
````
I am not sure why we need to wait here since commit & abort are finite states

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java (hunk 2 above)
Comment (abstractdog) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/monitoring/TezJobMonitor.java:283]: I see, this is an edge where we don't even reach real dag monitoring then
not sure where the diagnostics object is used, but does it make sense to use the full trace, like:
```
         String reportedException = "Execution has failed, stack trace: " + ExceptionUtils.getStackTrace(e);
          console.printError(reportedException);
          diagnostics.append(reportedException);
```

File: ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java (hunk 6 above)
Comment (kasakrisz) [reply ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java:2396]: You can repro the issue on the master branch using MR. Just set
`metastore.txn.opentxn.timeout` high enough. The default is 1sec and the insert statements in the test run  within 1 sec with Tez but more than 1 sec with MR on my test PC.

File: ql/src/test/org/apache/hadoop/hive/ql/TestTxnNoBuckets.java (hunk 10 above)
Comment (kasakrisz) [reply ql/src/test/org/apache/hadoop/hive/ql/TestTxnNoBuckets.java:448]: Oh, ok. In case of bucket files which are not in acid format the row_id is generated at read. The bucket id is coming from the file name.
https://github.com/apache/hive/blob/c27d31722c2a7426f3236d3d892dbe1e206e840d/ql/src/java/org/apache/hadoop/hive/ql/io/AcidUtils.java#L547C5-L547C44
In case of acid writes it comes from the taskId
https://github.com/apache/hive/blob/c27d31722c2a7426f3236d3d892dbe1e206e840d/ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java#L969C13-L969C26

In case of Tez the taskIds are mapped differently to the files

I changed the update statement to update more than one record to achieve having more than one bucket in the new delta

File: ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java (hunk 8 above)
Comment (kasakrisz) [reply ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java]: Added `checkResultsAndVectorization` and `checkresults`.  The first one calls `checkresults`.

File: ql/src/test/org/apache/hadoop/hive/ql/TestTxnNoBuckets.java (hunk 10 above)
Comment (deniskuzZ) [reply ql/src/test/org/apache/hadoop/hive/ql/TestTxnNoBuckets.java:448]: 👍

File: ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java (hunk 8 above)
Comment (deniskuzZ) [reply ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java]: 👍

File: ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java (hunk 6 above)
Comment (deniskuzZ) [reply ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java:2396]: I'll check

**Comment by deniskuzZ** [comment]
thanks, @kasakrisz for the tests refactor! 
LGTM, but I'd like to validate the `testCleanerForTxnToWriteId` scenario, it concerns me

File: ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java [hunk 13]
```diff
@@ -256,17 +251,17 @@ public void testAcidMetaColumsDecode() throws Exception {
...
-    testSdpoBucketed(true, false, 1);
-    testSdpoBucketed(false, true, 1);
-    testSdpoBucketed(false, false,1);
-
-    testSdpoBucketed(true, true, 2);
-    testSdpoBucketed(true, false, 2);
-    testSdpoBucketed(false, true, 2);
-    testSdpoBucketed(false, false,2);
+    testSdpoBucketed(true, 1);
+    testSdpoBucketed(true, 1);
```
zratkai, line 255: There are multiple test cases with the same input parameters, so testing the same twice. Please eliminate the duplications !

zratkai (changes requested):
Please check my comment!

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands3.java
kasakrisz: Removed the duplications.

ql/src/test/org/apache/hadoop/hive/ql/TestTxnCommands2.java
deniskuzZ: I'm sorry, it might take a bit more time, but I don't want to hold it as a blocker. So, if everything else is resolved, let's merge it and I'll come back with the findings later if any.

deniskuzZ: @kasakrisz,  i've checked out your branch and removed `waitUntilAllTxnFinished`, but still can't repro, even tried to increase `metastore.txn.opentxn.timeout` to 2 sec

deniskuzZ: note, that `runCleaner` has the same wait logic before execution

deniskuzZ: @kasakrisz. please take a look, otherwise, I think we can merge

kasakrisz: I have a repro on a slower computer too but I had to set `metastore.txn.opentxn.timeout` to 5sec. With 2 sec the test passed on the slower computer and failed on the faster.
Anyway, I added a TODO to revisit this logic.

abstractdog:
this was huge, thanks @kasakrisz, @deniskuzZ !

PR #5560, opened 2024-11-26
HIVE-28642:column stats state unnecessarily created when hive.stats.c…

yigress:
do not set column stats state when load new partition if hive.stats.column.autogather=false

simplify query string when load new partition by removing the column stats state setup when HIVE_STATS_COL_AUTOGATHER=false

avoid PARAM_VALUE too long error from metastore db.

no

no

tested locally

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -2820,8 +2820,10 @@ public class Hive {
       if (oldPart == null) {
         newTPart.getTPartition().setParameters(new HashMap<String,String>());
         if (this.getConf().getBoolVar(HiveConf.ConfVars.HIVE_STATS_AUTOGATHER)) {
+          List<String> colNames = this.getConf().getBoolVar(ConfVars.HIVE_STATS_COL_AUTOGATHER)?
+              MetaStoreUtils.getColumnNames(tbl.getCols()) : null;
           StatsSetupConst.setStatsStateForCreateTable(newTPart.getParameters(),
-              MetaStoreUtils.getColumnNames(tbl.getCols()), StatsSetupConst.TRUE);
+              colNames, StatsSetupConst.TRUE);
         }
         // Note: we are creating a brand new the partition, so this is going to be valid for ACID.
         List<FileStatus> filesForStats = null;


ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -2820,8 +2820,10 @@ private Partition loadPartitionInternal(Path loadPath, Table tbl, Map<String, St
       if (oldPart == null) {
         newTPart.getTPartition().setParameters(new HashMap<String,String>());
         if (this.getConf().getBoolVar(HiveConf.ConfVars.HIVE_STATS_AUTOGATHER)) {
+          List<String> colNames = this.getConf().getBoolVar(ConfVars.HIVE_STATS_COL_AUTOGATHER)?
```
ngsg, line 2823: NIT, I think we miss a space before `?`.

yigress:
thanks @ngsg for the review. I changed CreateTableLikeOperation, AbstractAddPartitionAnalyzer  and the loadPartition to all use null so that they don't set column stats state at the time of the call. for loadPartition the actual column stats is updated later. for CreateTableLikeOperation & AbstractAddPartitionAnalyzer the call conditions are those new objects should not have column stats so setting state true is wrong. stats computation are done by statswork and not triggered with parsing the COLUMN_STATS_ACCURATE string.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/like/CreateTableLikeOperation.java
```diff
@@ -74,8 +74,7 @@ public int execute() throws HiveException {
 
     if (desc.getLocation() == null && !tbl.isPartitioned() &&
         context.getConf().getBoolVar(HiveConf.ConfVars.HIVE_STATS_AUTOGATHER)) {
-      StatsSetupConst.setStatsStateForCreateTable(tbl.getTTable().getParameters(),
-          MetaStoreUtils.getColumnNames(tbl.getCols()), StatsSetupConst.TRUE);
+      StatsSetupConst.setStatsStateForCreateTable(tbl.getTTable().getParameters(),null, StatsSetupConst.TRUE);
```
ngsg, line 77: NIT, we missed a space between `tbl.getTTable().getParameters()` and `null`.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/add/AbstractAddPartitionAnalyzer.java
```diff
@@ -120,8 +120,9 @@ private AlterTableAddPartitionDesc.PartitionDesc createPartitionDesc(Table table
     Map<String, String> params = null;
     if (conf.getBoolVar(HiveConf.ConfVars.HIVE_STATS_AUTOGATHER) && location == null) {
       params = new HashMap<String, String>();
-      StatsSetupConst.setStatsStateForCreateTable(params,
-          MetaStoreUtils.getColumnNames(table.getCols()), StatsSetupConst.TRUE);
+      List<String> colNames = conf.getBoolVar(HiveConf.ConfVars.HIVE_STATS_COL_AUTOGATHER) ?
+          MetaStoreUtils.getColumnNames(table.getCols()) : null;
+      StatsSetupConst.setStatsStateForCreateTable(params,colNames, StatsSetupConst.TRUE);
```
ngsg, line 125: NIT, the same space issue here.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -2820,8 +2820,7 @@ private Partition loadPartitionInternal(Path loadPath, Table tbl, Map<String, St
       if (oldPart == null) {
         newTPart.getTPartition().setParameters(new HashMap<String,String>());
         if (this.getConf().getBoolVar(HiveConf.ConfVars.HIVE_STATS_AUTOGATHER)) {
-          StatsSetupConst.setStatsStateForCreateTable(newTPart.getParameters(),
-              MetaStoreUtils.getColumnNames(tbl.getCols()), StatsSetupConst.TRUE);
+          StatsSetupConst.setStatsStateForCreateTable(newTPart.getParameters(),null, StatsSetupConst.TRUE);
```
ngsg, line 2823: NIT, the same space issue here, too.

ngsg:
Could you check the result of Hive CI? It seems that your last change disables `HIVE_STATS_COL_AUTOGATHER` feature.

ngsg:
LGTM with one nit and updating 3 qfile outputs

I have checked that 3 qfiles failed because Hive marks all columns as accurate in `COLUMN_STATS_ACCURATE` for newly create TableLikes and Partitions. I think these test qfile verify your patch on `CreateTableLikeOperation` and `AbstractAddPartitionAnalyzer`.

I wonder if there exists any incorrect column statistics computation issues which this patch can fix, but I couldn't find such problems so far. The following [link](https://gist.github.com/ngsg/04b8e4e39901891527557143774500c2) contains a query that I used to investigate. It would be appreciated if anyone know and could share an example of incorrect column statistics computation.

okumin:
I wonder why we don't have to apply a similar update on the other classes using `HIVE_STATS_AUTOGATHER`. `CreateTableDesc` or `AlterTableConcatenateAnalyzer` are examples

deniskuzZ:
hi @yigress, any plans to continue with the PR?

yigress:
@deniskuzZ sorry I don't have time to work on this these days.

deniskuzZ:
> @deniskuzZ sorry I don't have time to work on this these days.

ok, no worries

PR #5562, opened 2024-11-26
HIVE-15221: Improvement for MapJoin checkMemoryStatus, adding gc befor…

BsoBird:
…e throwing Exception

https://issues.apache.org/jira/browse/HIVE-15221

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/mapjoin/MapJoinMemoryExhaustionHandler.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/mapjoin/MapJoinMemoryExhaustionHandler.java
@@ -96,7 +96,12 @@ public class MapJoinMemoryExhaustionHandler {
         + tableContainerSize + "\tMemory usage:\t" + usedMemory + "\tpercentage:\t" + percentageNumberFormat.format(percentage);
     console.printInfo(msg);
     if(percentage > maxMemoryUsage) {
-      throw new MapJoinMemoryExhaustionError(msg);
+      System.gc();
+      usedMemory = memoryMXBean.getHeapMemoryUsage().getUsed();
+      percentage = (double) usedMemory / (double) maxHeapSize;
+      if(percentage > maxMemoryUsage) {
+        throw new MapJoinMemoryExhaustionError(msg);
+      }
     }
    }
 }


BsoBird:
@zhangbutao 
hello. can you check this? Tks.

ql/src/java/org/apache/hadoop/hive/ql/exec/mapjoin/MapJoinMemoryExhaustionHandler.java
```diff
@@ -96,7 +96,12 @@ public void checkMemoryStatus(long tableContainerSize, long numRows)
         + tableContainerSize + "\tMemory usage:\t" + usedMemory + "\tpercentage:\t" + percentageNumberFormat.format(percentage);
     console.printInfo(msg);
     if(percentage > maxMemoryUsage) {
-      throw new MapJoinMemoryExhaustionError(msg);
+      System.gc();
```
zhangbutao, line 99: I am not sure how to test this code. Do you have some ut or local test to show that this code work well?

btw, I think adding a configuration will be good here & Keep the conf disabled.

okumin: Is this safe or desirable with LLAP?

BsoBird: @zhangbutao @okumin 
As a whole, this patch doesn't change any execution logic, it just rechecks the memory space after trying to call system.gc() once in case of out-of-memory.

okumin: I wondered if we can use System.gc, a JVM global method, for a task-local problem. I basically don't know why we implemented MapJoinMemoryExhaustionHandler. I might be saying something strange

BsoBird: I've used this patch in our production environment, and the OOM and MapJoinMemoryExhaustionError related issues have really been reduced. However, I do agree with you to some extent that the current implementation is a bit problematic.

okumin: First, I might not have sufficient knowledge to give this +1. I've never faced this error.
Reading the source code, I found that only Hive on MapReduce uses `MemoryExhaustionChecker`. So, not only LLAP but also Tez are not involved with this error. My original concern, i.e., the entire LLAP daemon could get degraded, is gone.
But I am still wondering why Hive on MapReduce needs `MemoryExhaustionChecker`. We may remove the class and let MapReduce crush with OOM when creating a giant hash table.

BsoBird: @okumin Hello. Although I haven't carefully reviewed this part of the code yet, using the TEZ engine will still trigger the MemoryExhaustionChecker.

PR #5563, opened 2024-11-26
HIVE-28633: Insert to Bucketed Partition table fails with CBO=false and dynamic sort partition optimization enabled

Indhumathi27:
Disabled Dynamic sorting partition optimization with non-cbo flow.

Post [HIVE-20703](https://issues.apache.org/jira/browse/HIVE-20703), Dynamic Sorting Partition Optimization  was integrated into the Tez query flow. However, for non-CBO query plans, the RowSchema is not populated correctly, leading to NullPointerException (NPE) errors. Addressing this would require extensive changes to the query plan, which is not necessary since Dynamic Sorting Partition Optimization is based on cost-based decisions.

No

No

Test case added

### DIFF
--- a/ql/src/test/queries/clientpositive/dynpart_bucket.q
+++ b/ql/src/test/queries/clientpositive/dynpart_bucket.q
@@ -1,0 +1,9 @@
+set hive.cbo.enable=false;
+
+drop table if exists dynpart_bucket;
+CREATE TABLE dynpart_bucket (bn string) PARTITIONED BY (br string) CLUSTERED BY (bn) INTO 2 BUCKETS ROW FORMAT DELIMITED FIELDS TERMINATED BY ',' STORED AS TEXTFILE ;
+INSERT into TABLE dynpart_bucket VALUES ('tv_0', 'tv');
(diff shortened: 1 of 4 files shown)


ql/src/test/queries/clientpositive/dynpart_bucket.q
```diff
@@ -0,0 +1,8 @@
+set hive.cbo.enable=false;
+
+drop table if exists dynpart_bucket;
+CREATE TABLE dynpart_bucket (bn string) PARTITIONED BY (br string) CLUSTERED BY (bn) INTO 2 BUCKETS ROW FORMAT DELIMITED FIELDS TERMINATED BY ',' STORED AS TEXTFILE ;
+INSERT into TABLE dynpart_bucket VALUES ('tv_0', 'tv');
+set hive.cbo.enable=true;
+INSERT into TABLE dynpart_bucket VALUES ('tv_1', 'tv');
+select * from dynpart_bucket;
```
soumyakanti3578, line 8: Can you please add a newline at EOF?

abstractdog (approved):

PR #5566, opened 2024-11-28
HIVE-28649: Remove date in resourceMap from QueryState when retrying …

kokila-19:
…query execution.

When concurrent write fails in iceberg, the query is retried. On retry, the previous execution's commit info in resourceMap is
 available which leads to FileNotFoundException.

Reset the resourceMap in Querystate in releaseContext()

When concurrent write fails in iceberg, the query might fail with 'Found conflicting files'. This typically occurs when two or more operations conflict over writing to the same table, leading to a situation where the write cannot be completed successfully due to file-level conflicts.
When this error occurs, the query is retried based on the retry strategies defined. 
During retry, in prepareContext(), 
```
if (!driverContext.isRetrial()) {
        closeInProcess(false);
}
```
closeInProcess is called which in return invokes releaseContext() which removes the queryState from SessionState.
However, the driverContext still holds the same queryState, and this queryState is added back to SessionState.

```
if (SessionState.get() != null) {
      QueryState queryState = getQueryState();
      SessionState.get().addQueryState(queryState.getQueryId(), queryState);
    }
```

When the query is retried, depending on the tez session pool's availability , it can be assigned a different Tez session than the one used during the previous execution. If the application ID is different than the previous execution, then job ID will also change. If we don't clear the resourceMap , it will contain both old + new job ID. This will lead to FileNotFoundException when the files of old job id is accessed.

No

No

Tested Manually in a cluster.


ql/src/java/org/apache/hadoop/hive/ql/QueryState.java
```diff
@@ -217,6 +217,11 @@ public Object getResource(String resourceIdentifier) {
     return resourceMap.get(resourceIdentifier);
   }
 
+  // Resets the resourceMap by removing all stored resource information,
```
abstractdog, line 220: nit: comment ends with a comma, implying the comment is not finished :)

abstractdog:
the fix looks good, I think we do bad things to the resource map in general :)
can you please add a unit test? I don't insist on doing the whole scenario, but somethng like:
1. you can use TestDriver (it already has examples)
2. start SessionState
3. put a value in resourcemap, assert that value exists
4. call Driver.closeInProcess (you can open it for package-protected @VisibleForTesting if needed)
5. assert that the value in resourcemap is not present

kokila-19:
@deniskuzZ I can replicate only "Found conflicting files" issue that occurs on first execution. On retry, in cluster we get "FileNotFoundException" and this is not reproducible in unit test. In UT, it always uses same application ID while retrying the query, so we wont hit this issue. I can only reproduce this currently in a cluster with yarn and tez setup.

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerTestUtils.java
```diff
@@ -51,6 +51,18 @@ public class HiveIcebergStorageHandlerTestUtils {
           optional(3, "Last_name", Types.StringType.get())
   );
 
+  static final Schema SOURCE_SCHEMA = new Schema(
```
deniskuzZ, line 54: why do we need both `SOURCE_SCHEMA` & `TARGET_SCHEMA` they are the same? can't we create some generic object like `USER_CLICKS_SCHEMA`

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerTestUtils.java
```diff
@@ -75,6 +87,20 @@ public class HiveIcebergStorageHandlerTestUtils {
       .add(3L, "Trudy", "Henderson")
       .build();
 
+  static final List<Record> SOURCE_RECORDS = TestHelper.RecordsBuilder
```
deniskuzZ, line 90: USER_CLICKS_RECORDS_1 & USER_CLICKS_RECORDS_2?

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestOptimisticRetry.java
```diff
@@ -166,4 +166,44 @@ public void testConcurrent2MergeInserts() {
     Assert.assertEquals(6, res.size());
   }
 
+  @Test
+  public void testConcurrent2MergeUpdates() {
```
deniskuzZ, line 170: 👍

deniskuzZ (commented):
LGTM, some minor test comments

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerTestUtils.java
kokila-19: updated.

deniskuzZ (approved):
+1

PR #5567, opened 2024-11-28
HIVE-28578: Fix concurrency issue in ObjectStore#updateTableColumnStatistics

InvisibleProgrammer:
There can be a concurrency issue and persisting table parameters can have duplicate key error when replication wants to store table column statistics and the table have statistics for multiple engines. In this change, in a parallel process already saved those, we retry it so that, DataNucleus will try to do update instead of insert. 

To avoid duplicate key errors.

No

No

Repro and test process is described in the ticket: https://issues.apache.org/jira/browse/HIVE-28578?focusedCommentId=17901761&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17901761
Tested manually, on a cluster.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
@@ -267,6 +267,7 @@ import org.apache.thrift.TException;
 import org.datanucleus.ExecutionContext;
 import org.datanucleus.api.jdo.JDOPersistenceManager;
 import org.datanucleus.api.jdo.JDOTransaction;
+import org.datanucleus.exceptions.NucleusDataStoreException;
 import org.datanucleus.store.rdbms.exceptions.MissingTableException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -10262,7 +10263,7 @@ public class ObjectStore implements RawStore, Configurable {
     }
     return statsMap;
   }
-  
+
   @Override
   public Map<String, String> updateTableColumnStatistics(ColumnStatistics colStats, String validWriteIds, long writeId)
       throws NoSuchObjectException, MetaException, InvalidObjectException, InvalidInputException {
@@ -10270,7 +10271,7 @@ public class ObjectStore implements RawStore, Configurable {
 
     List<ColumnStatisticsObj> statsObjs = colStats.getStatsObj();
     ColumnStatisticsDesc statsDesc = colStats.getStatsDesc();
-    
+
     Lock tableLock = getTableLockFor(statsDesc.getDbName(), statsDesc.getTableName());
     tableLock.lock();
     try {
@@ -10296,18 +10297,23 @@ public class ObjectStore implements RawStore, Configurable {
       }
 
       // TODO: (HIVE-20109) ideally the col stats stats should be in colstats, not in the table!
-      // Set the table properties
-      // No need to check again if it exists.
-      String dbname = table.getDbName();
-      String name = table.getTableName();
-      MTable oldt = mTable;
       Map<String, String> newParams = new HashMap<>(table.getParameters());
-      StatsSetupConst.setColumnStatsState(newParams, colNames);
-      boolean isTxn = TxnUtils.isTransactionalTable(oldt.getParameters());
-      if (isTxn) {
-        if (!areTxnStatsSupported) {
-          StatsSetupConst.setBasicStatsState(newParams, StatsSetupConst.FALSE);
-        } else {
+
+      int retries = 3;
+      boolean success = false;
+      while (!success && retries > 0) {
+        // TODO## ideally the col stats stats should be in colstats, not in the table!
+        // Set the table properties
+        // No need to check again if it exists.
+        String dbname = table.getDbName();
+        String name = table.getTableName();
+        MTable oldt = mTable;
+        StatsSetupConst.setColumnStatsState(newParams, colNames);
+        boolean isTxn = TxnUtils.isTransactionalTable(oldt.getParameters());
+        if (isTxn) {
+          if (!areTxnStatsSupported) {
+            StatsSetupConst.setBasicStatsState(newParams, StatsSetupConst.FALSE);
+          } else {
           String errorMsg = verifyStatsChangeCtx(TableName.getDbTable(dbname, name),
             oldt.getParameters(), newParams, writeId, validWriteIds, true);
           if (errorMsg != null) {
@@ -10322,12 +10328,27 @@ public class ObjectStore implements RawStore, Configurable {
           oldt.setWriteId(writeId);
         }
       }
-      oldt.setParameters(newParams);
+        try {
+          oldt.setParameters(newParams);
+          success = true;
+        }
+        catch (NucleusDataStoreException e) {
+          retries--;
+
+          if (retries == 0) {
+            throw e;
+          }
+
+          newParams.clear();
+          LOG.info("Retry on updateTableColumnStatistics: {}", retries);
+        }
+      }
 
       committed = commitTransaction();
       // TODO: similar to update...Part, this used to do "return committed;"; makes little sense.
       return committed ? newParams : null;
-    } finally {
+    }
+    finally {
       try {
         if (!committed) {
           rollbackTransaction();
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
@@ -10306,30 +10306,29 @@ public class ObjectStore implements RawStore, Configurable {
         // Set the table properties
         // No need to check again if it exists.
         String dbname = table.getDbName();
-        String name = table.getTableName();
-        MTable oldt = mTable;
+        String tableName = table.getTableName();
         StatsSetupConst.setColumnStatsState(newParams, colNames);
-        boolean isTxn = TxnUtils.isTransactionalTable(oldt.getParameters());
+        boolean isTxn = TxnUtils.isTransactionalTable(mTable.getParameters());
         if (isTxn) {
           if (!areTxnStatsSupported) {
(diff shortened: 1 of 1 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10296,18 +10297,23 @@ public Map<String, String> updateTableColumnStatistics(ColumnStatistics colStats
...
-        } else {
+
+      int retries = 3;
+      boolean success = false;
+      while (!success && retries > 0) {
+        // TODO## ideally the col stats stats should be in colstats, not in the table!
+        // Set the table properties
+        // No need to check again if it exists.
+        String dbname = table.getDbName();
+        String name = table.getTableName();
```
zratkai, line 10309: Please use proper name like tableName.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10296,18 +10297,23 @@ public Map<String, String> updateTableColumnStatistics(ColumnStatistics colStats
...
-      StatsSetupConst.setColumnStatsState(newParams, colNames);
-      boolean isTxn = TxnUtils.isTransactionalTable(oldt.getParameters());
-      if (isTxn) {
-        if (!areTxnStatsSupported) {
-          StatsSetupConst.setBasicStatsState(newParams, StatsSetupConst.FALSE);
-        } else {
+
+      int retries = 3;
+      boolean success = false;
+      while (!success && retries > 0) {
```
zratkai, line 10304: Should not be waiting between retries?

InvisibleProgrammer: No. 

```Summary: 
updateTableColumnStatistics can throw SQLIntegrityConstraintViolationException during replication if HA is on and two different HMS instance gets the same call but with different engine. 
 
Workaround: 
Update table column statistics in single threaded. 
 
Details: 
updateTableColumnStatistics has a relative long running transaction. In that transaction, it validates the actual parameters, queries the metastore db against the TABLE_PARAMS that are already stored and makes a decision based on that. After this, it uses data nucleus to persist the new statistics. 
From the two HMS instances, one can save the column statistics. And the other cannot as the first instance already saved them. 
```

The point is that both process A and process B decides to store the new data. On db level, it is an insert. Process A commits the insert. Process B fails with constraint violation as it is already exists. If we retry process B, it queries the current state of the statistics again so now it will make a decision to do update, instead of insert. 
Unfortunately, DataNucleus doesn't know such a thing like upsert - it would be way easier in that way...

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
InvisibleProgrammer: Sure

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10296,18 +10297,23 @@ public Map<String, String> updateTableColumnStatistics(ColumnStatistics colStats
...
+        // TODO## ideally the col stats stats should be in colstats, not in the table!
+        // Set the table properties
+        // No need to check again if it exists.
+        String dbname = table.getDbName();
+        String name = table.getTableName();
+        MTable oldt = mTable;
```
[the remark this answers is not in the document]
InvisibleProgrammer: I double checked the code and it seems that variable is completely unnecessary. I removed it, ran the test on TestObjectStore. Let's see the result of the pipeline test run.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10296,38 +10297,57 @@ public Map<String, String> updateTableColumnStatistics(ColumnStatistics colStats
...
+        }
+      }
+        try {
+          mTable.setParameters(newParams);
+          success = true;
+        }
+        catch (NucleusDataStoreException e) {
+          retries--;
+
+          if (retries == 0) {
```
kokila-19, line 10337: How about logging the error here? Something like Maximum number of retries (Retry = 3) reached.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10296,38 +10297,57 @@ public Map<String, String> updateTableColumnStatistics(ColumnStatistics colStats
...
-        if (!areTxnStatsSupported) {
-          StatsSetupConst.setBasicStatsState(newParams, StatsSetupConst.FALSE);
-        } else {
-          String errorMsg = verifyStatsChangeCtx(TableName.getDbTable(dbname, name),
-            oldt.getParameters(), newParams, writeId, validWriteIds, true);
+
+      int retries = 3;
+      boolean success = false;
+      while (!success && retries > 0) {
+        // TODO## ideally the col stats stats should be in colstats, not in the table!
```
kokila-19, line 10305: " **col stats stats**" I think the stats was repeated by mistake, so this can be removed.

InvisibleProgrammer: I'm not sure if the author made it as a mistakes. Column statistics are a kind of statistics. So saying the statistics around column statistics shouldn't be in the table make sense to me.

deniskuzZ:
I am not sure I get this. There is already a table-level lock for atomic stats update and we are adding an extra re-try loop inside?

InvisibleProgrammer:
> I am not sure I get this. There is already a table-level lock for atomic stats update and we are adding an extra re-try loop inside?

The existing table-level lock is an in-memory lock. That means in case you have multiple HMS instances and two different processes are getting requests in the same time, it just doesn't have any effect - as on JVM level, there is only one single instance.

deniskuzZ:
> > I am not sure I get this. There is already a table-level lock for atomic stats update and we are adding an extra re-try loop inside?
> 
> The existing table-level lock is an in-memory lock. That means in case you have multiple HMS instances and two different processes are getting requests in the same time, it just doesn't have any effect - as on JVM level, there is only one single instance.

ok, in-memory lock serves diff purpose - thread safety, see HIVE-25904. 
Descriptions says "updateTableColumnStatistics can throw SQLIntegrityConstraintViolationException during replication if HA is on and two different HMS instance gets the same call but with different engine. "
isn't the engine part of PK? RawStore interface. methods are not retriable

InvisibleProgrammer:
> > > I am not sure I get this. There is already a table-level lock for atomic stats update and we are adding an extra re-try loop inside?
> > 
> > 
> > The existing table-level lock is an in-memory lock. That means in case you have multiple HMS instances and two different processes are getting requests in the same time, it just doesn't have any effect - as on JVM level, there is only one single instance.
> 
> ok, in-memory lock serves diff purpose - thread safety, see [HIVE-25904](https://issues.apache.org/jira/browse/HIVE-25904).
> 
> Descriptions says "updateTableColumnStatistics can throw SQLIntegrityConstraintViolationException during replication if HA is on and two different HMS instance gets the same call but with different engine. " isn't the `engine` part of PK? If not, maybe it should be..
> 
> Some of RawStore interface methods are not retriable by design, so I would expect others to be retriable without the need for extra hacks, simply use `RetryingMetaStoreClient`.
> 
> ```
> public interface RawStore extends Configurable {
>   /***
>    * Annotation to skip retries
>    */
>   @Target(value = ElementType.METHOD)
>   @Retention(value = RetentionPolicy.RUNTIME)
>   @interface CanNotRetry {
>   }
> ```

> Descriptions says "updateTableColumnStatistics can throw SQLIntegrityConstraintViolationException during replication if HA is on and two different HMS instance gets the same call but with different engine. " isn't the `engine` part of PK? If not, maybe it should be..

It is trickier than this. Most of the statistics are stored in `TAB_COL_STATS` and it does have an engine field. But we have at least one statistics that is stored in `TABLE_PARAMS`. I really don't know the reason, but even the code says it shouldn't be in that way. See that comment from 2018: 
```
      // TODO: (HIVE-20109) ideally the col stats stats should be in colstats, not in the table!
```
I assume that TODO never happened. 
And TABLE_PARAMS is just a key-value list. For example, at the customer that faced with the problem, the duplicate entry is `COLUMN_STATS_ACCURATE`. I really don't know what is the reason why this property is in TABLE_PARAMS, not TAB_COL_STATS. And I also have no information about how many other fields can be in the same situation. 

> Some of RawStore interface methods are not retriable by design, so I would expect others to be retriable without the need for extra hacks, simply use `RetryingMetaStoreClient`.
Let me check what that exactly does and if it is possible to apply at the customer. 
TBH, I'm not completely sure if retry is a proper expression about the current change: You can imagine, if, for example, we could cover the situation with a simple SQL script, it could be an upsert merge statement: insert if it doesn't exist. Update if it does. But DataNucleus is an ORM tool that cannot do such a thing. But if we force it to query the table state again, at the second attempt it will recognise that ups, we have records. So we should do an update instead of an insert. 
Do you know better way to do an insert or update statement, using DataNucleus?

InvisibleProgrammer:
Hi @deniskuzZ , 
Could you please explain the reason why you consider a wrong idea to try an insert if an update didn't succeed due to a concurrency issue? 

TBH, I'm not sure if using the Retrying handler for _all_ the metastore client calls is an approach that worth to use just because we have this one and only rare edge case.

And I also consider it the problem of using an ORM tool that cannot properly handle such a trivial case. It could be handled with a simple merge statement, using plain SQL.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
zratkai: Thanks for the clarification!

PR #5568, opened 2024-12-02
HIVE-28654: MSCK repair fails for default partition when partition column is of numeric type

Aggarwal-Raghav:
[HIVE-28654](https://issues.apache.org/jira/browse/HIVE-28654)

To ensure msck repair command works when default partition "__HIVE_DEFAULT_PARTITION__" is present

NO

NO

Using q files (present in PR)
```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=msck_repair_9.q -Drat.skip -Dtest.output.overwrite
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=msck_repair_8.q -Drat.skip -Dtest.output.overwrite
```

### DIFF
--- a/ql/src/test/queries/clientpositive/msck_repair_9.q
+++ b/ql/src/test/queries/clientpositive/msck_repair_9.q
@@ -1,0 +1,22 @@
+DROP TABLE IF EXISTS tbl_x;
+DROP TABLE IF EXISTS tbl_y;
+
+CREATE EXTERNAL TABLE tbl_x (id INT, name STRING) PARTITIONED BY (month INT, day INT) stored as ORC location '${system:test.tmp.dir}/apps/hive/warehouse/test.db/tbl_x/';
+
+INSERT INTO tbl_x values(1, 'aaa', 12, 2);
+INSERT INTO tbl_x values(2, 'bbb', 12, 3);
+INSERT INTO tbl_x (id, name, month) values(3, 'ccc', 12);
+
+SET hive.exec.default.partition.name=ANOTHER_PARTITION;
+INSERT INTO tbl_x (id, name, day) values(4, 'ddd', 3);
+
+SHOW PARTITIONS tbl_x;
+
+CREATE EXTERNAL TABLE tbl_y (id INT, name STRING) PARTITIONED BY (month INT, day INT) stored as ORC location '${system:test.tmp.dir}/apps/hive/warehouse/test.db/tbl_x/';
+
+MSCK REPAIR TABLE tbl_y;
+
+SHOW PARTITIONS tbl_y;
+
+DROP TABLE tbl_x;
+DROP TABLE tbl_y;
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
@@ -63,6 +63,7 @@ import org.apache.commons.beanutils.PropertyUtils;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.ListUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
@@ -1635,6 +1636,10 @@ public class MetaStoreServerUtils {
   }
 
   public static String getNormalisedPartitionValue(String partitionValue, String type) {
+
+    if (!NumberUtils.isParsable(partitionValue)) {
+      return partitionValue;
+    }
 
     LOG.debug("Converting '" + partitionValue + "' to type: '" + type + "'.");
 
--- a/ql/src/test/queries/clientpositive/msck_repair_8.q
+++ b/ql/src/test/queries/clientpositive/msck_repair_8.q
@@ -1,0 +1,11 @@
+DROP TABLE IF EXISTS tbl1;
+
+CREATE EXTERNAL TABLE tbl1 (id INT, name STRING) PARTITIONED BY (my_date BIGINT) stored as ORC location '${system:test.tmp.dir}/apps/hive/warehouse/test.db/tbl1/';
+
+dfs ${system:test.dfs.mkdir} -p ${system:test.tmp.dir}/apps/hive/warehouse/test.db/tbl1/my_date=__HIVE_DEFAULT_PARTITION__;
+
+MSCK REPAIR TABLE tbl1;
+
+show partitions tbl1;
+
+DROP TABLE tbl1;
--- a/ql/src/test/results/clientpositive/llap/msck_repair_8.q.out
+++ b/ql/src/test/results/clientpositive/llap/msck_repair_8.q.out
@@ -1,0 +1,41 @@
+PREHOOK: query: DROP TABLE IF EXISTS tbl1
+PREHOOK: type: DROPTABLE
+PREHOOK: Output: database:default
+POSTHOOK: query: DROP TABLE IF EXISTS tbl1
+POSTHOOK: type: DROPTABLE
+POSTHOOK: Output: database:default
+#### A masked pattern was here ####
+PREHOOK: type: CREATETABLE
+#### A masked pattern was here ####
+PREHOOK: Output: database:default
+PREHOOK: Output: default@tbl1
+#### A masked pattern was here ####
+POSTHOOK: type: CREATETABLE
+#### A masked pattern was here ####
+POSTHOOK: Output: database:default
+POSTHOOK: Output: default@tbl1
+PREHOOK: query: MSCK REPAIR TABLE tbl1
+PREHOOK: type: MSCK
+PREHOOK: Output: default@tbl1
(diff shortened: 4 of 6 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,6 +1636,9 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
 
   public static String getNormalisedPartitionValue(String partitionValue, String type) {
 
+    if (partitionValue.equals("__HIVE_DEFAULT_PARTITION__"))
```
ngsg, line 1639: I have a quick question here. Does the patch solve the issue regardless of the value of `hive.exec.default.partition.name`? If not, I think it would be better to retreive the default partition name from `HiveConf` rather than using `__HIVE_DEFAULT_PARTITION__`.

Aggarwal-Raghav: @ngsg, thanks for reviewing the PR. Yes, the issue is regarding `hive.exec.default.partition.name`. As hive-common is not a dependency in _standalone-metastore/metastore-server/pom.xml_ , i think it better to use MetaConf as done in `getPartitionListByFilterExp` https://github.com/apache/hive/blob/2ff3ed6d83a1633d17bbf5b3c852b34b51548ef8/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java#L270

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1634,7 +1634,10 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
     return result;
   }
 
-  public static String getNormalisedPartitionValue(String partitionValue, String type) {
+  public static String getNormalisedPartitionValue(String partitionValue, String type, String defaultPartName) {
```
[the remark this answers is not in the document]
Aggarwal-Raghav: We would need a conf object to fetch the value of DEFAULTPARTITIONNAME and conf is not attribute of MetaStoreServerUtils class. Therefore, String is passed in the function argument.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
```diff
@@ -440,12 +440,13 @@ void findUnknownPartitions(Table table, Set<Path> missingPartDirs,
       partColNames.add(fSchema.getName());
     }
 
+    String defaultPartName = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.DEFAULTPARTITIONNAME);
```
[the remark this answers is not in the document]
Aggarwal-Raghav: **Yes, that case the solution won't work.**
But IMO, its user's responsibility to create the default partition of same type as that of the partition column type and ensure no other datatype partition column is present. 

Consider 2 scenarios where partition column is of **INT** type:
**Scenario1:**
User1 sets default partition conf value to "ANOTHER_PARTITION" and user 2 in another beeline session tries to insert data in default partition, then hdfs dir structure will look something like this:
```
path to table/part=ANOTHER_PARITION
path to table/part=__HIVE_DEFAULT_PARTITION__
```
and running msck repair will never succeed for any user. One user will get this
```
Caused by: java.lang.NumberFormatException: For input string: "ANOTHER_PARITION"
```
and other will get this
```
Caused by: java.lang.NumberFormatException: For input string: "__HIVE_DEFAULT_PARTITION__"
```

**Scenario2:**
User set default partition conf value to INT. Lets say 13, so hdfs dir structure looks like this
```
path to table/part=13
path to table/part=__HIVE_DEFAULT_PARTITION__
```
and in same session user creates a new table pointing of location of old table and run msck repair then it will fail for \_\_HIVE_DEFAULT_PARTITION__.

Unless the user change the config back to \_\_HIVE_DEFAULT_PARTITION__


**Questions that I also have**, is going forward what should be the expected behaviour? That the user has to ensure that there shouldn't be 2 partitions of different datatype than that of partition column as shown in scenario 1?
	
CC @ayushtkn @deniskuzZ  -- please provide your insights on this?



Also, there are multiple places where \_\_HIVE_DEFAULT_PARTITION__ is explicitly used instead of the config value. For instance here:
https://github.com/apache/hive/blob/c27d31722c2a7426f3236d3d892dbe1e206e840d/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java#L307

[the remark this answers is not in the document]
ayushtkn: I just had a quick look seems like the problem is coming post HIVE-23347, which added this method ``getNormalisedPartitionValue`` which does this parsing for numeric values. Before that patch things should have worked & Scenario-1 would have passed as well, with both values being shown in `show partitions`

This method ``getNormalisedPartitionValue`` looks like is valid only for numeric values to sort out some 0s or so. Maybe we can just have a numeric check & return rather than this default partition value matching

```
    if(!NumberUtils.isParsable(partitionValue)) {
      return partitionValue;
    }
```

If we check the regular case without MSCK for scenario-1, the show partitions would show both the values, I tried locally
```
+------------------------------------------+
|                partition                 |
+------------------------------------------+
| month=12/day=2                           |
| month=12/day=3                           |
| month=12/day=__HIVE_DEFAULT_PARTITION__  |
| month=__ANOTHER_PARTITION__/day=9        |
| month=__HIVE_DEFAULT_PARTITION__/day=3   |
+------------------------------------------+
5 rows selected (0.052 seconds)
0: jdbc:hive2://localhost:10000/> 
```

So, post MSCK also we should show both those values. So, I believe we should just check whether the value is numeric or not, if not, just return the value as is..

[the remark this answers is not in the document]
Aggarwal-Raghav: @ayushtkn  Thanks for looking into this. Yes, I have update the PR based on your suggestion

ql/src/test/queries/clientpositive/msck_repair_9.q
```diff
@@ -0,0 +1,22 @@
...
+SHOW PARTITIONS tbl_x;
+
+CREATE EXTERNAL TABLE tbl_y (id INT, name STRING) PARTITIONED BY (month INT, day INT) stored as ORC location '${system:test.tmp.dir}/apps/hive/warehouse/test.db/tbl_x/';
+
+MSCK REPAIR TABLE tbl_y;
+
+SHOW PARTITIONS tbl_y;
+
+DROP TABLE tbl_x;
+DROP TABLE tbl_y;
```
okumin, line 22: Thanks for adding great test cases

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,6 +1637,10 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
 
   public static String getNormalisedPartitionValue(String partitionValue, String type) {
 
+    if (!NumberUtils.isParsable(partitionValue)) {
+      return partitionValue;
+    }
```
okumin, line 1642: I slightly suspect this can skip the case of `decimal`.
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#isParsable(java.lang.String)

Aggarwal-Raghav: @okumin , thanks for the input. I think there is already a UT for this and passing post the patch. Please check https://github.com/apache/hive/blob/0e75930d619e96a32ae35b59df9eeeb6466ea2c6/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/utils/TestMetaStoreServerUtils.java#L965

Let me know if I am missing something, but **NumberUtils.isParsable(partitionValue)** will return true for a decimal and later in same function it will be handled. https://github.com/apache/hive/blob/0e75930d619e96a32ae35b59df9eeeb6466ea2c6/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java#L1653

okumin: I originally thought it could not work when `Double.parseDouble` can not decode it but `new BigDecimal` can decode it. An example in my mind was a huge number. Looking at the implementation, now I feel the current implementation would work
https://github.com/apache/commons-lang/blob/daa63aa595ea39a69e1b44ac23bc50fdcf2afd62/src/main/java/org/apache/commons/lang3/math/NumberUtils.java#L1794-L1810

ayushtkn (approved):
LGTM

PR #5571, opened 2024-12-05
HIVE-28662: Enable the dynamic leader election for HMS

dengzhhu653:
Enable the dynamic leader election on HMS by default

In the Hive ACID world, we must enable the metastore.compactor.initiator.on and metastore.compactor.cleaner.on on HMS to trigger the table compaction automatically. In a real warehouse, we might have multiple HMS instances behind, or even deploy for different purposes, running the tasks like these on all instances could be a waste of resources, make the HMS scale out not so easily. 

Default value of the following properties changed:
metastore.housekeeping.leader.election host -> lock
metastore.housekeeping.threads.on false -> true
metastore.compactor.initiator.on false -> true
metastore.compactor.cleaner.on false -> true

No

The existing UTs

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -584,7 +584,7 @@ public class MetastoreConf {
         "it is desirable to configure metastore.thrift.bind.host on the intended leader HMS."),
     METASTORE_HOUSEKEEPING_LEADER_ELECTION("metastore.housekeeping.leader.election",
         "metastore.housekeeping.leader.election",
-        "host", new StringSetValidator("host", "lock"),
+        "lock", new StringSetValidator("host", "lock"),
         "Set to host, HMS will choose the leader by the configured metastore.housekeeping.leader.hostname.\n" +
         "Set to lock, HMS will use the Hive lock to elect the leader."),
     METASTORE_HOUSEKEEPING_LEADER_AUDITTABLE("metastore.housekeeping.leader.auditTable",
@@ -603,7 +603,7 @@ public class MetastoreConf {
         "metastore.housekeeping.leader.lock.namespace", "",
         "The database where the Hive lock sits when metastore.housekeeping.leader.election is set to lock."),
     METASTORE_HOUSEKEEPING_THREADS_ON("metastore.housekeeping.threads.on",
-        "hive.metastore.housekeeping.threads.on", false,
+        "hive.metastore.housekeeping.threads.on", true,
         "Whether to run the tasks under metastore.task.threads.remote on this metastore instance or not.\n" +
             "Set this to true on one instance of the Thrift metastore service as part of turning\n" +
             "on Hive transactions. For a complete list of parameters required for turning on\n" +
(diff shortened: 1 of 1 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -638,12 +638,12 @@ public enum ConfVars {
...
         "Percentage (fractional) size of the delta files relative to the base directory. Deltas smaller than this threshold " +
             "count as small deltas. Default 0.01 = 1%.)"),
-    COMPACTOR_INITIATOR_ON("metastore.compactor.initiator.on", "hive.compactor.initiator.on", false,
+    COMPACTOR_INITIATOR_ON("metastore.compactor.initiator.on", "hive.compactor.initiator.on", true,
         "Whether to run the initiator thread on this metastore instance or not.\n" +
             "Set this to true on one instance of the Thrift metastore service as part of turning\n" +
             "on Hive transactions. For a complete list of parameters required for turning on\n" +
             "transactions, see hive.txn.manager."),
-    COMPACTOR_CLEANER_ON("metastore.compactor.cleaner.on", "hive.compactor.cleaner.on", false,
+    COMPACTOR_CLEANER_ON("metastore.compactor.cleaner.on", "hive.compactor.cleaner.on", true,
```
ayushtkn, line 646: If you set it to `true` all the HMS instances will have it turned on, maybe that ain't the intention of this PR, IIRC we usually enable cleaner only on one HMS instance not all, which post this change will be enabled on all

ayushtkn (commented):
I am not sure about by default enabling the cleaner & all, it would enable for all HMS instances surprisingly once someone upgrades + enable for those who don't have ACID tables as well

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
dengzhhu653: We would have only one instance running the `initiator` and `cleaner`, as well as the housekeeping tasks, as the `metastore.housekeeping.leader.election` is changed to `lock`, all HMS instances will trying to acquire the exclusive Hive lock, only the winner will become the leader and run these maintenance tasks.

deniskuzZ: ACID is the default table type so that makes sense. However, once we switch to Iceberg, I don't think we should keep ACID housekeeping services enabled by default.

deniskuzZ: @dengzhhu653 keep in mind that once that's in, you'll need to explicitly disable these configs for some of the form factors downstream.

deniskuzZ (approved):
LGTM +1

ayushtkn (approved):
LGTM

simhadri-g:
This seems to be causing a deadlock and HMS is crashing. https://issues.apache.org/jira/browse/HIVE-28659?focusedCommentId=17905565&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17905565

PR #5572, opened 2024-12-05
HIVE-28657: Regenerate protobuf files

abstractdog:
Regenerated java sources from proto files.
```
mvn clean install -Pitests -T 1C -DskipTests -Pprotobuf
```

We need to make sure that our generated sources can always be compiled and used with the current hive version.

No. Proto messages are for internal use.

No.

Precommit.


abstractdog:
converted to draft, I compiled with a bad protoc version (2.5), hive is on 3.25.5

PR #5575, opened 2024-12-10
HIVE-28660: OTEL: Modify span names to remove inconsistency in case o…

tanishq-chugh:
…f completed queries

OTEL span names are modified such that live/completed identifiers causing inconsistencies in case of completed queries are addressed.

Current names not well understood/taken by the end user.

Yes, Changes in the OTEL span names.

No

Manual testing, by running the OTEL service with Jaeger and Zipkin recievers.


ayushtkn (approved):
Tried Locally, Works as discussed & expected 

LGTM

PR #5577, opened 2024-12-12
HIVE-28408: Support ARRAY field access in CBO

ramesh0201:
We need to support array field access in the CBO. This is currently already supported in the execution side. Only CBO fails with an error. So we will be by passing the CBO by doing no-op in this case to leave it to hive compiler and execution engine to do rest of the work

In order to enable CBO for all the queries

No

No

mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=nested_column_pruning.q

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
@@ -1103,6 +1103,9 @@ import com.google.common.annotations.VisibleForTesting;
   private int getStructFieldIndex(ExprNodeFieldDesc exprNodeFieldDesc) throws HiveException {
     ExprNodeDesc structNodeDesc = exprNodeFieldDesc.getDesc();
     String fieldName = exprNodeFieldDesc.getFieldName();
+    if (exprNodeFieldDesc.getIsList()) {
+      throw new HiveException("Could not vectorize expression with a LIST type without an index");
+    }
     StructTypeInfo structTypeInfo = (StructTypeInfo) structNodeDesc.getTypeInfo();
     int index = 0;
     boolean isFieldExist = false;
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
@@ -23,6 +23,7 @@ import java.nio.charset.Charset;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -39,12 +40,16 @@ import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSubQuery;
 import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlCollation;
+import org.apache.calcite.sql.SqlFunction;
+import org.apache.calcite.sql.SqlFunctionCategory;
 import org.apache.calcite.sql.SqlIntervalQualifier;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.fun.SqlQuantifyOperator;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.type.ArraySqlType;
 import org.apache.calcite.util.ConversionUtil;
 import org.apache.calcite.util.DateString;
 import org.apache.calcite.util.NlsString;
@@ -67,6 +72,7 @@ import org.apache.hadoop.hive.ql.optimizer.calcite.CalciteSemanticException.Unsu
 import org.apache.hadoop.hive.ql.optimizer.calcite.CalciteSubquerySemanticException;
 import org.apache.hadoop.hive.ql.optimizer.calcite.HiveCalciteUtil;
 import org.apache.hadoop.hive.ql.optimizer.calcite.reloperators.HiveRexExprList;
+import org.apache.hadoop.hive.ql.optimizer.calcite.reloperators.HiveComponentAccess;
 import org.apache.hadoop.hive.ql.optimizer.calcite.translator.TypeConverter;
 import org.apache.hadoop.hive.ql.parse.ASTNode;
 import org.apache.hadoop.hive.ql.parse.HiveParser;
@@ -622,11 +628,13 @@ public class RexNodeExprFactory extends ExprFactory<RexNode> {
     if (expr.getType().isStruct()) {
       // regular case of accessing nested field in a column
       return rexBuilder.makeFieldAccess(expr, fieldName, true);
+    } else if (expr.getType().getComponentType() != null) {
+      RexNode wrap = rexBuilder.makeCall(expr.getType().getComponentType(), HiveComponentAccess.COMPONENT_ACCESS,
+                  Collections.singletonList(expr));
+      return createNestedColumnRefExpr(typeInfo, wrap, fieldName,
+              expr.getType().getComponentType() instanceof ArraySqlType);
     } else {
-      // This may happen for schema-less tables, where columns are dynamically
-      // supplied by serdes.
-      throw new CalciteSemanticException("Unexpected rexnode : "
-          + expr.getClass().getCanonicalName(), UnsupportedFeature.Schema_less_table);
+      throw new CalciteSemanticException("Unexpected rexnode : " + expr.getClass().getCanonicalName());
     }
   }
 
--- a/ql/src/test/results/clientpositive/llap/vector_orc_nested_column_pruning.q.out
+++ b/ql/src/test/results/clientpositive/llap/vector_orc_nested_column_pruning.q.out
@@ -2730,15 +2730,15 @@ STAGE PLANS:
             Map Operator Tree:
                 TableScan
                   alias: nested_tbl_1
-                  Pruned Column Paths: s1.f6, s5.f16
+                  Pruned Column Paths: s5.f16, s1.f6
                   Statistics: Num rows: 1 Data size: 3196 Basic stats: COMPLETE Column stats: NONE
                   Select Operator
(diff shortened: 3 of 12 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -1867,7 +1867,7 @@ public static enum ConfVars {
             + "execution engine doesn't take advantage of statistics stored in the cache."),
     // CBO related
     HIVE_CBO_ENABLED("hive.cbo.enable", true, "Flag to control enabling Cost Based Optimizations using Calcite framework."),
-    HIVE_CBO_FALLBACK_STRATEGY("hive.cbo.fallback.strategy", "CONSERVATIVE",
+    HIVE_CBO_FALLBACK_STRATEGY("hive.cbo.fallback.strategy", "NEVER",
```
okumin, line 1870: Can we make NEVER default? I guess this is the last known issue and that's why this PR would change it. It is one of the greatest news in 2024.
https://issues.apache.org/jira/browse/HIVE-27831

I'd create a separate ticket to change the default, cleaning up some test resources as [Stamatis tried before](https://github.com/apache/hive/pull/4836/files).

ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
```diff
@@ -622,11 +634,13 @@ protected RexNode createNestedColumnRefExpr(
     if (expr.getType().isStruct()) {
       // regular case of accessing nested field in a column
       return rexBuilder.makeFieldAccess(expr, fieldName, true);
+    } else if (expr.getType().getComponentType() != null){
```
okumin, line 637: I'd add a space after the right paren

ql/src/test/results/clientpositive/llap/vector_orc_nested_column_pruning.q.out
```diff
@@ -3008,7 +3008,7 @@ STAGE PLANS:
...
...ysical.Vectorizer.vectorizeSelectOperator(Vectorizer.java:4806), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer.validateAndVectorizeOperator(Vectorizer.java:5443), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer.doProcessChild(Vectorizer.java:1011), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer.doProcessChildren(Vectorizer.java:898), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer.validateAndVectorizeOperatorTree(Vectorizer.java:868), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer.access$2500(Vectorizer.java:253), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer$VectorizationDispatcher.validateAndVectorizeMapOperators(Vectorizer.java:2118), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer$VectorizationDispatcher.validateAndVectorizeMapOperators(Vectorizer.java:2070), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer$VectorizationDispatcher.validateAndVectorizeMapWork(Vectorizer.java:2045), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer$VectorizationDispatcher.convertMapWork(Vectorizer.java:1199), org.apache.hadoop.hive.ql.optimizer.physical.Vectorizer$VectorizationDispatcher.dispatch(Vectorizer.java:1051), ...
```
okumin, line 3011: Although the result is unchanged, maybe we would like to have a better error here. ClassCastException sounds irregular. Also, this stack trace should not be robust

ramesh0201: Thank you @okumin. Please review my latest updated patch for this.

ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
ramesh0201: Resolved

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
ramesh0201: Yes I reverted this back

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
```diff
@@ -1103,6 +1103,9 @@ private VectorExpression getGenericUDFStructField(ExprNodeFieldDesc exprNodeFiel
   private int getStructFieldIndex(ExprNodeFieldDesc exprNodeFieldDesc) throws HiveException {
     ExprNodeDesc structNodeDesc = exprNodeFieldDesc.getDesc();
     String fieldName = exprNodeFieldDesc.getFieldName();
+    if (exprNodeFieldDesc.getIsList()) {
+      throw new HiveException("Could not vectorize expression: Vectorizing complex type LIST is not supported");
+    }
     StructTypeInfo structTypeInfo = (StructTypeInfo) structNodeDesc.getTypeInfo();
```
zabetak, line 1109: The fact that we are hitting a `ClassCastException` without the extra if check above is a bit worrisome. It means that the new logic has some impact on the vectorization codepath. Why are we hitting this now?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
```diff
@@ -1103,6 +1103,9 @@ private VectorExpression getGenericUDFStructField(ExprNodeFieldDesc exprNodeFiel
   private int getStructFieldIndex(ExprNodeFieldDesc exprNodeFieldDesc) throws HiveException {
     ExprNodeDesc structNodeDesc = exprNodeFieldDesc.getDesc();
     String fieldName = exprNodeFieldDesc.getFieldName();
+    if (exprNodeFieldDesc.getIsList()) {
+      throw new HiveException("Could not vectorize expression: Vectorizing complex type LIST is not supported");
```
zabetak, line 1107: The `Vectorizing complex type..` message is created by `getValidateDataTypeErrorMsg` during some validation type checks. Why aren't we hitting these checks before arriving into this method?

ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
```diff
@@ -613,6 +617,14 @@ protected RexNode createConstantExpr(TypeInfo typeInfo, Object constantValue)
         TypeConverter.convert(typeInfo, rexBuilder.getTypeFactory()), false);
   }
 
+  /**
+   * Special operator that is used as syntactic sugar to change the type of collection
+   * expressions in order to perform field access over them.
+   */
+  public static final SqlOperator COMPONENT_ACCESS =
```
zabetak, line 624: Most of the existing Hive specific operators/functions are under `org.apache.hadoop.hive.ql.optimizer.calcite.reloperators` package (e.g., `HiveDateAddSqlOperator`). Consider moving the new operator into a separate class under that package.

Having one class per function is not ideal but it may be a good idea to follow the existing paradigm at least till we decide to perform a holistic refactoring that puts them all together.

ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
```diff
@@ -622,11 +634,13 @@ protected RexNode createNestedColumnRefExpr(
...
+    } else if (expr.getType().getComponentType() != null) {
+      RexNode wrap =
+          rexBuilder.makeCall(expr.getType().getComponentType(), COMPONENT_ACCESS, Collections.singletonList(expr));
+      return createNestedColumnRefExpr(typeInfo, wrap, fieldName, isList);
     } else {
-      // This may happen for schema-less tables, where columns are dynamically
-      // supplied by serdes.
-      throw new CalciteSemanticException("Unexpected rexnode : "
-          + expr.getClass().getCanonicalName(), UnsupportedFeature.Schema_less_table);
+      // Safe exception. Shouldn't Ideally come here.
```
zabetak, line 642: The comment does not provide much info; consider dropping it.

ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
```diff
@@ -622,11 +634,13 @@ protected RexNode createNestedColumnRefExpr(
...
       return rexBuilder.makeFieldAccess(expr, fieldName, true);
+    } else if (expr.getType().getComponentType() != null) {
+      RexNode wrap =
+          rexBuilder.makeCall(expr.getType().getComponentType(), COMPONENT_ACCESS, Collections.singletonList(expr));
+      return createNestedColumnRefExpr(typeInfo, wrap, fieldName, isList);
     } else {
-      // This may happen for schema-less tables, where columns are dynamically
-      // supplied by serdes.
-      throw new CalciteSemanticException("Unexpected rexnode : "
-          + expr.getClass().getCanonicalName(), UnsupportedFeature.Schema_less_table);
```
zabetak, line 629: Can we remove the `UnsupportedFeature.Schema_less_table` from the enumeration.

zabetak:
It would be also nice to include a qfile with an EXPLAIN CBO statement that shows that we can treat ARRAY field access in CBO. Possibly the book example that I added under HIVE-28408 is a good fit.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
ramesh0201: We are hitting this message because we never do that check(call getValidateDataTypeErrorMsg) for the select clause expressions and do this check only for the group by expression, which is later in the operator tree for that particular vertex.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
ramesh0201: I was able to reproduce the ClassCastException even without my patch for a simpler query that I mentioned below:
EXPLAIN VECTORIZATION EXPRESSION
SELECT s5.f16.f18.f19 FROM nested_tbl_1;
I will be adding this to the list of queries in the same q file below. What I identified is that, we have certain nested access queries that were not handled well in the vectorization and we need to address that. But there is no new failures after our patch here. We can try to address this issue in a separate patch if you think it is reasonable idea.

ramesh0201:
@zabetak Can you please review the latest patch? I have addressed most of the comments, except the two  that have I left a comment above.

zabetak (approved):
LGTM, thanks for addressing the comments @ramesh0201 !

ramesh0201:
Thank you for the review @okumin and @zabetak

PR #5578, opened 2024-12-16
HIVE-28655: Implement HMS Related Drop Stats Changes

DanielZhu58:


standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
```diff
@@ -3019,9 +3036,15 @@ PartitionsResponse get_partitions_req(1:PartitionsRequest req)
   bool delete_partition_column_statistics(1:string db_name, 2:string tbl_name, 3:string part_name, 4:string col_name, 5:string engine) throws
               (1:NoSuchObjectException o1, 2:MetaException o2, 3:InvalidObjectException o3,
                4:InvalidInputException o4)
+  bool delete_partition_column_statistics_req(1:DeletePartitionColumnStatisticsRequest req) throws
+              (1:NoSuchObjectException o1, 2:MetaException o2, 3:InvalidObjectException o3,
+              4:InvalidInputException o4)
   bool delete_table_column_statistics(1:string db_name, 2:string tbl_name, 3:string col_name, 4:string engine) throws
               (1:NoSuchObjectException o1, 2:MetaException o2, 3:InvalidObjectException o3,
                4:InvalidInputException o4)
+  bool delete_table_column_statistics_req(1: DeleteTableColumnStatisticsRequest req) throws
```
dengzhhu653, line 3045: can we merge the two methods? i.e, merge them to `bool delete_column_statistics_req(DeleteColumnStatisticsRequest req)`

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -2716,6 +2719,33 @@ boolean deletePartitionColumnStatistics(String catName, String dbName, String ta
...
+                                                      List<String> colNames, String engine) throws TException;
+
+  /**
+   * Delete partition level column statistics given dbName, tableName, partName and colName, or
+   * all columns in a partition.
+   * @param req the DeletePartitionColumnStatisticsRequest which including
+   *            catalog name, database name, table name, partition name, column names, and engine name
+   * @throws TException thrift transport error
+   */
+  public boolean deletePartitionMultiColumnStatistics(DeletePartitionColumnStatisticsRequest req) throws TException;
```
dengzhhu653, line 2747: I would like to merge the `deletePartitionMultiColumnStatistics` and `deleteTableMultiColumnStatistics`, and keep only one method, i.e, a `deleteMultiColumnStatistics(DeleteColumnStatisticsRequest req)` for both dropping table statistics and partition statistics.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
```diff
@@ -1395,9 +1395,50 @@ List<ColumnStatistics> getPartitionColumnStatistics(
...
+   * @param partVals partition values.
+   * @param colNames a list of column names.
+   * @param engine engine for which we want to delete statistics
+   * @return Boolean indicating the outcome of the operation
+   * @throws NoSuchObjectException no such partition
+   * @throws MetaException error access the RDBMS
+   * @throws InvalidObjectException error dropping the stats
+   * @throws InvalidInputException bad input, such as null table or database name.
+   */
+  boolean deletePartitionColumnStatisticsInBatch(String catName, String dbName, String tableName,
```
dengzhhu653, line 1439: what's the difference between `deletePartitionMultiColumnStatistics` and `deletePartitionColumnStatisticsInBatch`, can them merge?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
```diff
@@ -1395,9 +1395,50 @@ List<ColumnStatistics> getPartitionColumnStatistics(
    * @throws InvalidInputException bad input, such as null table or database name.
    */
   boolean deletePartitionColumnStatistics(String catName, String dbName, String tableName,
```
dengzhhu653, line 1397: you can make this method as `default`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/cache/CachedStore.java
```diff
@@ -18,18 +18,7 @@
...
-import java.util.EmptyStackException;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.Stack;
+import java.util.*;
```
dengzhhu653, line 21: nit: restore the import

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7420,6 +7420,64 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+        startFunction("delete_column_statistics_by_partition",": table=" +
+                TableName.getQualified(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName) +
+                " partition=" + partName + " column=" + colName);
+      }
+    }
+    boolean ret = false, committed = false;
+
+    getMS().openTransaction();
+    try {
+      List<String> partVals = getPartValsFromName(getMS(), parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, partName);
```
dengzhhu653, line 7445: this `partName` in thrift defines as `optional`, should we take care of the nullable `partName`. To make the interface more generic, can we define the `partName` as `List<String>`, so we can delete multiple partitions' statistics?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7420,6 +7420,64 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+      Table table = getMS().getTable(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName);
+      // This API looks unused; if it were used we'd need to update stats state and write ID.
+      // We cannot just randomly nuke some txn stats.
+      if (TxnUtils.isTransactionalTable(table)) {
```
dengzhhu653, line 7449: why can't we drop the transactional statistics? can we make the partition stats as incorrect in partition params by `StatsSetupConst.COLUMN_STATS_ACCURATE`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7420,6 +7420,64 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+              partName, partVals, colNames, engine);
+      if (ret && colNames != null) {
+        for (String colName : colNames) {
+          if (transactionalListeners != null && !transactionalListeners.isEmpty()) {
+            MetaStoreListenerNotifier.notifyEvent(transactionalListeners,
+                    EventType.DELETE_PARTITION_COLUMN_STAT,
+                    new DeletePartitionColumnStatEvent(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName,
+                            partName, partVals, colName, engine, this));
+          }
+          if (!listeners.isEmpty()) {
```
dengzhhu653, line 7463: should this call be outside of the `transaction`? why do we notify the listener just on success and `colNames` is not null, will some events be missing?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,46 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+                + "  for which stats deletion is requested doesn't exist");
+      }
+      // Note: this does not verify ACID state; called internally when removing cols/etc.
+      //       Also called via an unused metastore API that checks for ACID tables.
+      MPartition mPartition = getMPartition(catName, dbName, tableName, partVals, mTable);
+      if (mPartition == null) {
+        throw new NoSuchObjectException("Partition " + partName
+                + " for which stats deletion is requested doesn't exist");
+      }
+      query = pm.newQuery(MPartitionColumnStatistics.class);
```
dengzhhu653, line 10358: where do we use this `query`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,46 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+      if (colNames != null) {
+        for (String colName : colNames){
+          deletePartitionColumnStatistics(catName, dbName, tableName, partName, partVals, colName, engine);
```
dengzhhu653, line 10361: can we drop all the columns' statistics in a round?

Aggarwal-Raghav:
@DanielZhu58, one question, to call the drop stats command it will be something like: `ALTER TABLE .. drop col/partition stats..`.  **So antlr file changes (HiveParser.g) are required right?**

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
DanielZhu58: Sometimes the user just want to drop the stats in a specific partition. 
In some cases, a certain partition stats becomes really huge so we want to drop stats for it only.
The 2 methods are designed for different use cases.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
DanielZhu58: Yes, I agree. I think we can merge these 2 methods.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/cache/CachedStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
DanielZhu58: Currently we don't need to delete the stats for multiple partitions, because currently there is no syntax for the user to drop stats for multiple partitions in HQL.
And also in the use cases, it's not very common that stats of multiple adjacent partitions become huge. The user can drop stats of multiple partitions one by one manually. 
I think this partName here is not optional. 
In hive_metastore.thrift, 
`struct DeletePartitionColumnStatisticsRequest {
  1: required string cat_name,
  2: required string db_name,
  3: required string tbl_name,
  4: required string part_name,
  5: optional list<string> col_names,
  6: required string engine
}`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
DanielZhu58: This comments are not written by me. We can mark this issue as to do, and discuss about it later.

DanielZhu58:
> @DanielZhu58, one question, to call the drop stats command it will be something like: `ALTER TABLE .. drop col/partition stats..`. **So antlr file changes (HiveParser.g) are required right?**

Hi @Aggarwal-Raghav , yes, the method changes in this patch are mainly focused on the HMS API changes. It's only the part of the big change.
There is another patch included the HiveParser.g changes coming soon, authored by @ramesh0201 
https://github.com/apache/hive/pull/5176

Aggarwal-Raghav:
> > @DanielZhu58, one question, to call the drop stats command it will be something like: `ALTER TABLE .. drop col/partition stats..`. **So antlr file changes (HiveParser.g) are required right?**
> 
> Hi @Aggarwal-Raghav , yes, the method changes in this patch are mainly focused on the HMS API changes. It's only the part of the big change. There is another patch included the HiveParser.g changes coming soon, authored by @ramesh0201 #5176

ok

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,108 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+    catName = normalizeIdentifier(catName);
+    if (tableName == null) {
+      throw new InvalidInputException("Table name is null.");
+    }
+    try {
+      openTransaction();
+      MTable mTable = getMTable(catName, dbName, tableName);
+      MPartitionColumnStatistics mStatsObj;
+      List<MPartitionColumnStatistics> mStatsObjColl;
+      if (mTable == null) {
```
dengzhhu653, line 10350: Is this check `mTable == null` redundant? as we validate the `mPartition` afterwards

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,108 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+      //       Also called via an unused metastore API that checks for ACID tables.
+      MPartition mPartition = getMPartition(catName, dbName, tableName, partVals, mTable);
+      if (mPartition == null) {
+        throw new NoSuchObjectException("Partition " + partName
+                + " for which stats deletion is requested doesn't exist");
+      }
+      query = pm.newQuery(MPartitionColumnStatistics.class);
+      String filter;
+      String parameters;
+      if (colNames != null) {
```
dengzhhu653, line 10364: nit: colNames != null && !colNames.isEmpty()

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,108 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+        parameters = "java.lang.String t1, java.lang.String t2, java.lang.String t3, "
+                + "java.util.Collection t4, java.lang.String t5" + (engine != null ? ", java.lang.String t6" : "");
+      } else {
+        filter = "partition.partitionName == t1 && partition.table.database.name == t2 && partition.table.tableName == t3 && partition.table.database.catalogName == t4" + (engine != null ? " && engine == t5" : "");
+        parameters = "java.lang.String t1, java.lang.String t2, java.lang.String t3, java.lang.String t4" + (engine != null ? ", java.lang.String t5" : "");
+      }
+      query.setFilter(filter);
+      query.declareParameters(parameters);
+      if (colNames != null) {
+        query.setUnique(true);
```
dengzhhu653, line 10376: as we might have multiple columns, there is no guarantee that the query will return only one record

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,108 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+        for (String colName : colNames){
+          // trim the extra spaces, and change to lowercase
+          normalizeIdentifier(colName);
```
dengzhhu653, line 10379: need to put the normalized column names into new list which will be passed to `executeWithArray`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,108 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+        }
+        if (engine != null) {
+          mStatsObj =
```
dengzhhu653, line 10382: this query result should be a collection

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,108 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+        }
+        pm.retrieve(mStatsObj);
+        if (mStatsObj != null) {
+          pm.deletePersistent(mStatsObj);
+        } else {
+          throw new NoSuchObjectException("Column stats doesn't exist for table="
+                  + TableName.getQualified(catName, dbName, tableName) +
+                  " partition=" + partName + " col=" + String.join(", ", colNames));
+        }
+      } else {
```
dengzhhu653, line 10405: nit: can we merge the `if else`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10330,6 +10330,108 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
...
+        }
+        pm.retrieveAll(mStatsObjColl);
+        if (mStatsObjColl != null) {
+          pm.deletePersistentAll(mStatsObjColl);
+        } else {
+          throw new NoSuchObjectException("Column stats don't exist for table="
+                  + TableName.getQualified(catName, dbName, tableName) + " partition" + partName);
+        }
+      }
+      ret = commitTransaction();
```
dengzhhu653, line 10428: after deleting the column stats, should we mark the `COLUMN_STATS_ACCURATE` to false in partition's parameters?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10420,6 +10522,100 @@ public boolean deleteTableColumnStatistics(String catName, String dbName, String
...
+        throw new NoSuchObjectException("Table " +
+                TableName.getQualified(catName, dbName, tableName)
+                + "  for which stats deletion is requested doesn't exist");
+      }
+      // Note: this does not verify ACID state; called internally when removing cols/etc.
+      //       Also called via an unused metastore API that checks for ACID tables.
+      query = pm.newQuery(MTableColumnStatistics.class);
+      String filter;
+      String parameters;
+      if (colNames != null) {
```
dengzhhu653, line 10551: nit: `colNames != null and !colNames.isEmpty()`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10420,6 +10522,100 @@ public boolean deleteTableColumnStatistics(String catName, String dbName, String
...
+        parameters = "java.lang.String t1, java.lang.String t2, java.lang.String t3, java.util.Collection t4" + (engine != null ? ", java.lang.String t5" : "");
+      } else {
+        filter = "table.tableName == t1 && table.database.name == t2 && table.database.catalogName == t3" + (engine != null ? " && engine == t4" : "");
+        parameters = "java.lang.String t1, java.lang.String t2, java.lang.String t3" + (engine != null ? ", java.lang.String t4" : "");
+      }
+
+      query.setFilter(filter);
+      query.declareParameters(parameters);
+      if (colNames != null) {
+        query.setUnique(true);
```
dengzhhu653, line 10562: remove the unique check

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10420,6 +10522,100 @@ public boolean deleteTableColumnStatistics(String catName, String dbName, String
...
+        for (String colName : colNames){
+          // trim the extra spaces, and change to lowercase
+          normalizeIdentifier(colName);
```
dengzhhu653, line 10565: put the normalize into a new list

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: I think we can keep this if condition. 
Because if mTable is null, the exception will be thrown from here, so we can know the "Table" does not exist. 
If the exception is thrown later with the mPartition, we can know the "Partition" does not exist.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Acknowledged. Yes, this should be removed.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Acknowledged. Created a new list `normalizedColNames`, to pass to `executeWithArray`.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: That's a good point. I don't have an answer for it now. 
We had the feature of dropping column stats for single column before, and the `COLUMN_STATS_ACCURATE ` was not mark to false.
So let's discuss about that later.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: One condition is (colNames != null) and the other one condition is (colNames == null). In this 2 conditions, the JDO queries are different. So it's clearer to not merge them for now.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10420,6 +10522,100 @@ public boolean deleteTableColumnStatistics(String catName, String dbName, String
...
+        }
+        pm.retrieve(mStatsObj);
+
+        if (mStatsObj != null) {
+          pm.deletePersistent(mStatsObj);
+        } else {
+          throw new NoSuchObjectException("Column stats doesn't exist for db=" + dbName + " table="
+                  + tableName + " col=" + String.join(", ", colNames));
+        }
+      } else {
```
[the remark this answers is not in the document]
DanielZhu58: Reasons same as mentioned above.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10234,93 +10234,105 @@ protected Integer getJdoResult(GetHelper<Integer> ctx) throws MetaException, NoS
...
+                          normalizeIdentifier(dbName),
+                          normalizeIdentifier(tableName),
+                          normalizedColNames,
+                          normalizeIdentifier(catName));
+        }
+        pm.retrieveAll(mStatsObjColl);
+        if (mStatsObjColl != null) {
+          pm.deletePersistentAll(mStatsObjColl);
         } else {
           throw new NoSuchObjectException("Column stats doesn't exist for table="
```
dengzhhu653, line 10311: nit: this exception should never throw?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10234,93 +10234,105 @@ protected Integer getJdoResult(GetHelper<Integer> ctx) throws MetaException, NoS
...
         }
       } else {
         if (engine != null) {
           mStatsObjColl =
-              (List<MPartitionColumnStatistics>) query.executeWithArray(partName.trim(),
-                  normalizeIdentifier(dbName),
-                  normalizeIdentifier(tableName),
-                  normalizeIdentifier(catName),
-                  engine);
+                  (List<MPartitionColumnStatistics>) query.executeWithArray(partName.trim(),
```
dengzhhu653, line 10318: nit: code format

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
```diff
@@ -1394,9 +1394,32 @@ List<ColumnStatistics> getPartitionColumnStatistics(
...
+   * @param colNames a list of column names.
+   * @param engine engine for which we want to delete statistics
+   * @return Boolean indicating the outcome of the operation
+   * @throws NoSuchObjectException no such partition
+   * @throws MetaException error access the RDBMS
+   * @throws InvalidObjectException error dropping the stats
+   * @throws InvalidInputException bad input, such as null table or database name.
+   */
+  boolean deletePartitionMultiColumnStatistics(String catName, String dbName, String tableName,
+    String partName, List<String> partVals, List<String> colNames, String engine)
```
dengzhhu653, line 1421: the `partVals` never be used?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10234,93 +10234,105 @@ protected Integer getJdoResult(GetHelper<Integer> ctx) throws MetaException, NoS
   public boolean deletePartitionColumnStatistics(String catName, String dbName, String tableName,
       String partName, List<String> partVals, String colName, String engine)
       throws NoSuchObjectException, MetaException, InvalidObjectException, InvalidInputException {
+    List<String> colNames = new ArrayList<>();
+    if (colName != null) {
+      colNames.add(colName);
+    }
+    return deletePartitionMultiColumnStatistics(catName, dbName, tableName, partName, partVals, colNames, engine);
```
dengzhhu653, line 10241: The direct sql path is missing, we'd better pushing the `colNames` into the direct sql

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
```diff
@@ -1394,9 +1394,32 @@ List<ColumnStatistics> getPartitionColumnStatistics(
    * @throws InvalidObjectException error dropping the stats
    * @throws InvalidInputException bad input, such as null table or database name.
    */
-  boolean deletePartitionColumnStatistics(String catName, String dbName, String tableName,
-      String partName, List<String> partVals, String colName, String engine)
-      throws NoSuchObjectException, MetaException, InvalidObjectException, InvalidInputException;
+  default boolean deletePartitionColumnStatistics(String catName, String dbName, String tableName,
+    String partName, List<String> partVals, String colName, String engine)
+    throws NoSuchObjectException, MetaException, InvalidObjectException, InvalidInputException{
+    return false;
```
dengzhhu653, line 1400: return deletePartitionMultiColumnStatistics(catName, dbName, tableName,
   partName, partVals, Arrays.asList(colName), engine);

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
```diff
@@ -1856,6 +1859,49 @@ public void testColumnStatistics() throws Throwable {
...
+      status = client.deleteTableMultiColumnStatistics(dbName, tblName, null, ENGINE);
+      assertTrue(status);
+      stats = client.getTableColumnStatistics(
+              dbName, tblName, Lists.newArrayList(colName[0]), ENGINE);
+      assertTrue("stats are not empty: " + stats, stats.isEmpty());
+      stats = client.getTableColumnStatistics(
+              dbName, tblName, Lists.newArrayList(colName[1]), ENGINE);
+      assertTrue("stats are not empty: " + stats, stats.isEmpty());
+
+      client.updateTableColumnStatistics(colStats);
```
dengzhhu653, line 1884: we have tested the single column case at https://github.com/apache/hive/pull/5578/files#diff-120e8efba348e9a8c233de4135cdcc4aaeb6bc0fe2540d9d952554af2c05fdecR1867

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
```diff
@@ -1856,6 +1859,49 @@ public void testColumnStatistics() throws Throwable {
...
+      assertTrue(status);
+      stats = client.getTableColumnStatistics(
+              dbName, tblName, Lists.newArrayList(colName[0]), ENGINE);
+      assertTrue("stats are not empty: " + stats, stats.isEmpty());
+
+      client.updateTableColumnStatistics(colStats);
+      //case 3: column names contains multiple column names, then delete the column stats in the column name list
+      status = client.deleteTableMultiColumnStatistics(dbName, tblName, Arrays.asList(colName), ENGINE);
+      assertTrue(status);
+      stats = client.getTableColumnStatistics(
```
dengzhhu653, line 1898: get the stats in a round, `client.getTableColumnStatistics(dbName, tblName, Lists.newArrayList(colName), ENGINE);`

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
```diff
@@ -1885,30 +1931,57 @@ public void testColumnStatistics() throws Throwable {
...
-
-     colStats2 = client.getPartitionColumnStatistics(dbName, tblName,
-         Lists.newArrayList(partName), Lists.newArrayList(colName[0]), ENGINE).get(partName).get(0);
-
-     // test get stats on a column for which stats doesn't exist
-     Map<String, List<ColumnStatisticsObj>> stats2 = client.getPartitionColumnStatistics(dbName, tblName,
-         Lists.newArrayList(partName), Lists.newArrayList(colName[1]), ENGINE);
-     assertTrue("stats are not empty: " + stats2, stats2.isEmpty());
+      // compare stats obj to ensure what we get is what we wrote
+      assertNotNull(colStats2);
```
dengzhhu653, line 1940: restore the code format

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Let's discuss this today. I am not sure if I understand this.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Let's discuss this today. I am not sure if I understand this.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
DanielZhu58: It's used around ObjectStore.java Line 10266

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
DanielZhu58: We don't want to change any of the performance of the current "delete column stats" related API, so that's why we added a new method in it. So if we need, we can merge these 2 methods in the next patch.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
DanielZhu58: Update: deletePartitionColumnStatisticsInBatch() is removed because we changed the implementation way of deletePartitionMultiColumnStatistics(). The performance should be good enough.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Update: The implementation way is changed.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: Yes, acknowledged.
The implementation way is changed.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
DanielZhu58: I am not sure if I understand this.
In this test file, we are testing 3 cases. (Columns names amount are null, 1, or multiple)

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetaStore.java
```diff
@@ -1885,30 +1931,57 @@ public void testColumnStatistics() throws Throwable {
...
+      Map<String, List<ColumnStatisticsObj>> stats2 = client.getPartitionColumnStatistics(dbName, tblName,
+              Lists.newArrayList(partName), Lists.newArrayList(colName[0]), ENGINE);
+      assertTrue("stats are not empty: " + stats2, stats2.isEmpty());
+
+      client.updatePartitionColumnStatistics(colStats);
+      // test delete multiple column stats
+      // case 1: column names are null, all column stats should be deleted
+      status = client.deletePartitionMultiColumnStatistics(dbName, tblName, partName, null, ENGINE);
+      assertTrue(status);
+      stats2 = client.getPartitionColumnStatistics(dbName, tblName,
```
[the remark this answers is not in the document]
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
dengzhhu653: The difference between the two conditions is the input parameters of the `Query`, we can use a list to collect all parameters and then push it to `query.executeWithArray(...`, example:
```
List parameters = new ArrayList();
parameters.add(normalizeIdentifier(tableName));
parameters.add(normalizeIdentifier(dbName));
parameters.add(normalizeIdentifier(catName));
if (colNames != null && !colNames.isEmpty()) {
  parameters.add(colNames);
}
if (engine != null) {
  parameters.add(engine);
}
 mStatsObjColl =
                  (List<MTableColumnStatistics>) query.executeWithArray(parameters.toArray());
...
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
dengzhhu653: we can add the direct sql implementation for dropping

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
DanielZhu58: I think we can keep the old way. The old way has a better readability.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -3658,6 +3658,20 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
         tableName, partName, colName, engine);
   }
 
+  @Override
+  public boolean deletePartitionMultiColumnStatistics(String dbName, String tableName, String partName,
```
saihemanth-cloudera, line 3662: I don't like introducing two APIs for the same purpose. Can we remove this API and use the other one?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -2716,6 +2719,33 @@ boolean deletePartitionColumnStatistics(String catName, String dbName, String ta
...
+   * @param colNames a list of column name, or null for all columns
+   * @param engine engine, or null for all engines
+   * @return boolean indicating outcome of the operation
+   * @throws NoSuchObjectException no such partition exists
+   * @throws InvalidObjectException error dropping the stats data
+   * @throws MetaException error accessing the RDBMS
+   * @throws TException thrift transport error
+   * @throws InvalidInputException input is invalid or null.
+   */
+  public boolean deletePartitionMultiColumnStatistics(String dbName, String tableName, String partName,
```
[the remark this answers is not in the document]
nrg4878: I agree that we should only have 1 method especially since the Request object already has everything that other method accepts.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
nrg4878: I agree with this. We should remove one of them, retain the one that takes the Request Object, rename it to something more generic like DeleteColumnStatisticsRequest, make the PartitionName optional and use this value to determine whether to delete the Table column stats or Partition columns stats.

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
```diff
@@ -2510,6 +2510,15 @@ struct GetAllWriteEventInfoRequest {
   3: optional string tableName
 }
 
+struct DeleteColumnStatisticsRequest {
+  1: required string cat_name,
```
dengzhhu653, line 2514: nit:  the `cat_name` should be optional to align with other request, the `engine` as well,
```
optional string engine = "hive"
```

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -3658,6 +3658,15 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
         tableName, partName, colName, engine);
   }
 
+//  @Override
```
dengzhhu653, line 3661: nit: remove them if the method is commented out

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -39,6 +39,9 @@
 import org.apache.hadoop.hive.metastore.partition.spec.PartitionSpecProxy;
 import org.apache.thrift.TException;
 
+import static org.apache.hadoop.hive.metastore.utils.MetaStoreUtils.getDefaultCatalog;
```
dengzhhu653, line 42: nit: unused imports

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7471,6 +7562,64 @@ public boolean delete_table_column_statistics(String dbName, String tableName, S
     return ret;
   }
 
+//  @Override
+//  public boolean delete_table_column_statistics_req(DeleteTableColumnStatisticsRequest req)
+//          throws TException {
```
dengzhhu653, line 7567: nit: remove the commented codes

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7420,6 +7420,97 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+    String engine = req.getEngine();
+    dbName = dbName.toLowerCase();
+    String[] parsedDbName = parseDbName(dbName, conf);
+    tableName = tableName.toLowerCase();
+
+    if (colNames != null) {
+      for (String colName : colNames) {
+        colName = colName.toLowerCase();
+        if (isDeletingTableLevel) {
+          startFunction("delete_column_statistics_by_table", ": table=" +
```
dengzhhu653, line 7441: No need to call multiple `startFunction`s, we can just put the colNames into the logs, such as:
```
 startFunction("delete_column_statistics_req", ": table=" +
                  TableName.getQualified(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName) +
                  " partition=" + partName + " column=" + colNames + " part_name=" + partName);
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7420,6 +7420,97 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+    }
+    boolean ret = false, committed = false;
+    getMS().openTransaction();
+    try {
+      Table table = getMS().getTable(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName);
+      boolean isPartitioned = table.isSetPartitionKeys() && table.getPartitionKeysSize() > 0;
+      if (TxnUtils.isTransactionalTable(table)) {
+        throw new MetaException("Cannot delete stats via this API for a transactional table");
+      }
+      if (isDeletingTableLevel && !isPartitioned){
```
dengzhhu653, line 7460: the `isDeletingTableLevel` is useless here, for `non-partitioned` table, the `isDeletingTableLevel` is always true

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7420,6 +7420,97 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+        ret = getMS().deleteTableMultiColumnStatistics(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, colNames, engine);
+        if (ret && colNames != null) {
+          for (String colName : colNames) {
```
dengzhhu653, line 7463: if the `colNames` is null, this `deleteTableMultiColumnStatistics` will drop all columns statistics, we might miss some DeleteTableColumnStatEvent in this case

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7420,6 +7420,97 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+          }
+        }
+      }
+      else {
+        if (isDeletingTableLevel){
+          ret = getMS().deletePartitionMultiColumnStatistics(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName,
+                  partName, null, colNames, engine);
+        }
+        List<String> partVals = getPartValsFromName(getMS(), parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, partName);
+        ret = getMS().deletePartitionMultiColumnStatistics(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName,
```
dengzhhu653, line 7485: whether the partName is null or not, both of them can call `ret = getMS().deletePartitionMultiColumnStatistics(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName,
Comment
whether the partName is null or not, both of them can invoke `getMS().deletePartitionMultiColumnStatistics(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, partName, colNames, engine);`, let RawStore to check the `partName` is null or not

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7420,6 +7420,97 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+                partName, partVals, colNames, engine);
+        if (ret && colNames != null) {
+          for (String colName : colNames) {
```
dengzhhu653, line 7488: we might miss some events when partName is null or colNames is null

dengzhhu653:
The latest change looks good to me, @DanielZhu58 could you skim through the sonar issues please to see if we can fix some of them? link: https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=5578&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

We still cannot drop the acid table's stats, can we open a new jira to track the problem? @DanielZhu58 
@saihemanth-cloudera @nrg4878 it would be great to take a another look on the PR before merging it. Thanks in advance!

saihemanth-cloudera (changes requested):
Please revert the MetastoreClient changes.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -2691,10 +2691,23 @@ Map<String, List<ColumnStatisticsObj>> getPartitionColumnStatistics(
    * @throws MetaException error accessing the RDBMS
    * @throws TException thrift transport error
    * @throws InvalidInputException input is invalid or null.
+   * @deprecated Use
+   *    {@link IMetaStoreClient#deleteColumnStatistics(org.apache.hadoop.hive.metastore.api.DeleteColumnStatisticsRequest)} instead
    */
-  boolean deletePartitionColumnStatistics(String dbName, String tableName,
```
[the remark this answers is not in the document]
dengzhhu653: it won't. the new change doesn't remove the api, it makes it `default` and add the implementation

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7379,47 +7381,97 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+        ret = rawStore.deleteTableColumnStatistics(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, colNames, engine);
+        if (ret) {
+          eventType = EventType.DELETE_TABLE_COLUMN_STAT;
+          for (String colName :
+              colNames == null ? table.getSd().getCols().stream().map(FieldSchema::getName).collect(Collectors.toList()) : colNames) {
+            if (transactionalListeners != null && !transactionalListeners.isEmpty()) {
+              MetaStoreListenerNotifier.notifyEvent(transactionalListeners, eventType,
+                  new DeleteTableColumnStatEvent(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, colName, engine, this));
+            }
+            events.add(new DeleteTableColumnStatEvent(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, colName, engine, this));
```
saihemanth-cloudera, line 7426: Can you create a follow-up patch to create/fire a single event for all the columns?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7379,47 +7381,97 @@ public boolean delete_partition_column_statistics(String dbName, String tableNam
...
+          for (String colName : colNames == null ? table.getSd().getCols().stream().map(FieldSchema::getName)
+              .collect(Collectors.toList()) : colNames) {
+            for (String partName : partNames) {
+              List<String> partVals = getPartValsFromName(getMS(), parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, partName);
+              if (transactionalListeners != null && !transactionalListeners.isEmpty()) {
+                MetaStoreListenerNotifier.notifyEvent(transactionalListeners, eventType,
+                    new DeletePartitionColumnStatEvent(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName,
+                        partName, partVals, colName, engine, this));
+              }
+              events.add(new DeletePartitionColumnStatEvent(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName,
```
saihemanth-cloudera, line 7453: This will be a performance killer while dropping partition stats for 10K+ partitions because you are fetching table object from the metastore for each partition via getPartValsFromName(Rawstore, catalogName, dbName, tableName, partName).
Please change this to getPartValsFromName(Table, partName) since you already fetched the table object previously.

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
```diff
@@ -576,12 +577,22 @@ public List<ColumnStatisticsObj> getTableColumnStatistics(String dbName, String
...
-  public boolean deleteTableColumnStatistics(String dbName, String tableName, String colName, String engine)
-      throws TException {
-    if (getTempTable(dbName, tableName) != null) {
-      return deleteTempTableColumnStats(dbName, tableName, colName);
+  public boolean deleteColumnStatistics(DeleteColumnStatisticsRequest req) throws TException {
+    String dbName = req.getDb_name();
+    String tableName = req.getTbl_name();
+    org.apache.hadoop.hive.metastore.api.Table table;
+    if ((table = getTempTable(dbName, tableName)) != null) {
+      List<String> colNames = req.getCol_names();
```
dengzhhu653, line 585: `colNames` might be null or empty?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/cache/CachedStore.java
```diff
@@ -47,16 +47,7 @@
...
-import org.apache.hadoop.hive.metastore.FileMetadataHandler;
-import org.apache.hadoop.hive.metastore.ObjectStore;
-import org.apache.hadoop.hive.metastore.PartFilterExprUtil;
-import org.apache.hadoop.hive.metastore.PartitionExpressionProxy;
-import org.apache.hadoop.hive.metastore.RawStore;
-import org.apache.hadoop.hive.metastore.TableType;
-import org.apache.hadoop.hive.metastore.TransactionalMetaStoreEventListener;
-import org.apache.hadoop.hive.metastore.Warehouse;
-import org.apache.hadoop.hive.metastore.HiveAlterHandler;
+import org.apache.hadoop.hive.metastore.*;
```
dengzhhu653, line 50: nit: restore the imports

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/cache/CachedStore.java
```diff
@@ -2319,6 +2333,34 @@ private Map<String, String> updatePartitionColumnStatisticsInternal(Table table,
...
+    boolean succ = rawStore.deletePartitionColumnStatistics(catName, dbName, tblName, partNames, colNames, engine);
+    // in case of event based cache update, cache is updated during commit txn.
+    if (succ && !canUseEvents) {
+      catName = normalizeIdentifier(catName);
+      dbName = normalizeIdentifier(dbName);
+      tblName = normalizeIdentifier(tblName);
+      if (!shouldCacheTable(catName, dbName, tblName)) {
+        return succ;
+      }
+      if (colNames != null) {
```
dengzhhu653, line 2351: if the `colNames` is null, our assumption is to delete all column statistics, but in CachedStore we don't follow this.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/cache/CachedStore.java
```diff
@@ -2201,6 +2192,29 @@ private static void updateTableColumnsStatsInternal(Configuration conf, ColumnSt
...
+    boolean succ = rawStore.deleteTableColumnStatistics(catName, dbName, tblName, colNames, engine);
+    // in case of event based cache update, cache is updated during commit txn
+    if (succ && !canUseEvents) {
+      catName = normalizeIdentifier(catName);
+      dbName = normalizeIdentifier(dbName);
+      tblName = normalizeIdentifier(tblName);
+      if (!shouldCacheTable(catName, dbName, tblName)) {
+        return succ;
+      }
+      if (colNames != null) {
```
dengzhhu653, line 2209: take care of the case when `colNames` is null

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
DanielZhu58: Yes.

dengzhhu653 (approved):

saihemanth-cloudera (approved):
LGTM +1

saihemanth-cloudera:
There is one failing test. Is it flaky? Can you repro in your local?

DanielZhu58:
> There is one failing test. Is it flaky? Can you repro in your local?

I ran the PR related tests locally, the tests have all passed successfully.
This failing test is about the **null sort order**, which I think is not related to this PR.
I will rerun the tests again.

PR #5579, opened 2024-12-18
HIVE-28670: Show failed status in all the latest replication history event

harshal-16:
- Once Dump/Load is failed with FAILED_ADMIN then following dumps/loads should also be flagged as same instead of SKIPPED


imkindia12 (approved):
LGTM

pudidic:
+1. Looks good to me.

pudidic (approved):

PR #5525, opened 2024-10-30
HIVE-28597: Upgrade to 8.4.3 LTS version of MySQL and 11.4 LTS version of MariaDB

KiranVelumuri:
### What changes were proposed in this pull request?

Upgrade MySQL version

### Why are the changes needed?

[HIVE-28597](https://issues.apache.org/jira/browse/HIVE-28597)

### Does this PR introduce _any_ user-facing change?

No

### Is the change a dependency upgrade?

No

### How was this patch tested?

mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=qt_database_mysql.q -pl itests/qtest -Pitests

PR #5580, opened 2024-12-19
HIVE-28671 : Upgrade MySQL connector jar version to 8.2.0

arorasimran0309:
MySQL version upgrade from 8.0.31 to 8.2.0

The current version of MySQL if 8.0.31, which has the following vulnerabilities associated with it:

Direct vulnerabilities:
The current version of MySQL connector jar is 8.0.31, which has the following vulnerabilities associated with it:

Direct vulnerabilities:
CVE-2023-22102

Vulnerabilities from dependencies:
CVE-2024-7254
CVE-2022-3510
CVE-2022-3509
CVE-2022-3171

So, this issue is to remedy this with the version upgrade as a fix.

https://dev.mysql.com/doc/relnotes/connector-j/en/news-8-2-0.html
Mysql connector/J version 8.2.0 is the smallest upgrade that fixes the CVEs and can be used against MySQL Server version 5.7 and later.
Versions 8.3.0 and above are compatible with mysql server versions 8.0 and above, and since the current version is 5.7.37 (at least as long as [#5525](https://github.com/apache/hive/pull/5525) is not merged and upgrades it to 8.4.3) upgrading mysql connecter jar version to 8.2.0 instead of 8.4.0 is the present solution.

No

Yes

Existing tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -178,7 +178,7 @@
     <log4j2.version>2.18.0</log4j2.version>
     <mariadb.version>2.5.0</mariadb.version>
     <mssql.version>6.2.1.jre8</mssql.version>
-    <mysql.version>8.0.31</mysql.version>
+    <mysql.version>8.2.0</mysql.version>
     <postgres.version>42.7.3</postgres.version>
     <oracle.version>21.3.0.0</oracle.version>
     <opencsv.version>5.9</opencsv.version>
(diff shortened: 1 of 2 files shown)


okumin (approved):
LGTM. Thanks for wrapping up the current situation and potential options. I also confirmed 8.3.0 requires MySQL 8.0.
https://dev.mysql.com/doc/relnotes/connector-j/en/news-8-3-0.html

ayushtkn (approved):
LGTM

arorasimran0309:
Thanks for the review @okumin @ayushtkn

PR #5582, opened 2024-12-19
HIVE-28668: Hive should emit fewer events for truncate table operation

saihemanth-cloudera:
Currently, we emit twice the number of events for each table/partition for truncate operation. This patch removes this redundant events.

Having fewer events help external systems that consume these events to catch up with the quantum of events faster.

No.

No.

Added unit test for the patch


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -3362,95 +3362,52 @@ private void updateStatsForTruncate(Map<String,String> props, EnvironmentContext
...
-      partition.setWriteId(writeId);
+    for (Partition partition: partitions) {
+      updateStatsForTruncate(partition.getParameters(), environmentContext);
+      if (writeId > 0) {
+        partition.setWriteId(writeId);
+      }
     }
-    alterHandler.alterPartition(ms, wh, catName, dbName, tableName, null, partition,
-        environmentContext, this, validWriteIds);
+    alterHandler.alterPartitions(ms, wh, catName, dbName, tableName, partitions, environmentContext,
```
dengzhhu653, line 3377: since we know this is a truncation op, can we just call `ms.alterPartitions(catName, dbname, name, partValsList, new_parts, writeId, writeIdList)` and create the event directly? same for truncating the non-partitioned table.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
```diff
@@ -101,7 +101,7 @@ public void setConf(Configuration conf) {
   @Override
   public void alterTable(RawStore msdb, Warehouse wh, String catName, String dbname,
       String name, Table newt, EnvironmentContext environmentContext,
-      IHMSHandler handler, String writeIdList)
+      IHMSHandler handler, String writeIdList, boolean isTruncateOp)
```
dengzhhu653, line 104: instead of adding a new `isTruncateOp`, can we put it into the environmentContext, like` environmentContext.putToProperties("__isTruncateOp__", "true");`

ql/src/java/org/apache/hadoop/hive/ql/parse/repl/load/message/TruncatePartitionHandler.java
```diff
@@ -35,17 +39,39 @@
...
-      Iterator<String> afterIterator = msg.getPtnObjAfter().getValuesIterator();
-      for (FieldSchema fs : tblObj.getPartitionKeys()) {
-        partSpec.put(fs.getName(), afterIterator.next());
+      if (MetastoreConf.getBoolVar(context.hiveConf,
+          MetastoreConf.ConfVars.NOTIFICATION_ALTER_PARTITIONS_V2_ENABLED)) {
+        AlterPartitionsMessage singleMsg = deserializer.getAlterPartitionsMessage(
+            context.dmd.getPayload());
+        tName = TableName.fromString(singleMsg.getTable(), null,
+            context.isDbNameEmpty() ? singleMsg.getDB() : context.dbName);
+        tblObj = singleMsg.getTableObj();
```
dengzhhu653, line 51: nit: can we create the `tName` from the `tblObj`?

ql/src/java/org/apache/hadoop/hive/ql/parse/repl/load/message/TruncatePartitionHandler.java
```diff
@@ -35,17 +39,39 @@
...
+        List<Map<String, String>> afterPartitionsList = singleMsg.getPartitions();
+        List<Task<?>> childTaskList = new ArrayList<>();
+        for(Map<String, String> afterIteratorMap : afterPartitionsList) {
+          Iterator<String> afterIterator = afterIteratorMap.values().iterator();
+          Map<String, String> partSpec = new LinkedHashMap<>();
+          for (FieldSchema fs : tblObj.getPartitionKeys()) {
+            partSpec.put(fs.getName(), afterIterator.next());
```
dengzhhu653, line 58: nit: `afterIterator.next` -> `afterIteratorMap.get(fs.getName())`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: We'll also call ObjectStore.alterPartitions() where the operation may not be a truncate, for e.g: L#6111 and L#6263

dengzhhu653: this operation is in a context of truncation, and we have done almost every thing in `HMSHandler`, like reset the column stats, truncate the table directory, so in my view, we don't need to call `alterHandler.alterPartitions` to perform other actions, such as fetch the partitions from db again, update the stats, etc

ql/src/java/org/apache/hadoop/hive/ql/parse/repl/load/message/TruncatePartitionHandler.java
saihemanth-cloudera: Replication tests fail because during the replication it would require source DB name(which we get it from 'context' variable).

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -3362,95 +3362,84 @@ private void updateStatsForTruncate(Map<String,String> props, EnvironmentContext
...
-            validWriteIds, writeId);
+      EnvironmentContext environmentContext = new EnvironmentContext();
+      updateStatsForTruncate(table.getParameters(), environmentContext);
+      environmentContext.putToProperties(hive_metastoreConstants.IS_TRUNCATE_OP, Boolean.TRUE.toString());
+
+      // TODO: this should actually pass thru and set writeId for txn stats.
+      if (writeId > 0) {
+        table.setWriteId(writeId);
       }
+      alterHandler.alterTable(ms, wh, catName, dbName, tableName, table,
```
dengzhhu653, line 3427: nit: cloud we also call `ms.alterTable` directly for truncating the table? so we don't need to update the column statistics in alterHandler and using the `IS_TRUNCATE_OP` to flag the operation. Rest looks good to me.

dengzhhu653 (approved):

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: Done

PR #5583, opened 2024-12-19
HIVE-28582: OOM when compiling query with many GROUP BY columns aliased multiple times

zabetak:
Copy [RelMdUniqueKeys#getProjectUniqueKeys](https://github.com/apache/calcite/blob/648a832e0b3abc0f1cd4887847bdef7c133cb383/core/src/main/java/org/apache/calcite/rel/metadata/RelMdUniqueKeys.java#L162) to HiveRelMdUniqueKeys handler to overcome the OOM till we upgrade to a Calcite version with the necessary fixes. The copy is almost exact minus a small adaptation in `Util#transform` that transforms the `colMask` bitset to a list.

The `RelMdUniqueKeys` handler has some logic that leads to an exponential increase of results (unique keys) when certain query/plan patterns appear. The problem has been addressed by CALCITE-6640 and CALCITE-6704 but those are not available in current Calcite version.

For more details see HIVE-28582.

No

No

```
mvn test -Dtest=TestMiniLlapLocalCliDriver.java -Dqfile="cbo_unique_keys_oom.q"
```


zabetak:
The test failures in `groupby_join_pushdown.q.out` are due to HIVE-28675. Once HIVE-28675 gets in the changes will no longer appear.

PR #5584, opened 2024-12-19
HIVE-28673: Fix issues in JSON SerDe implementations related to Decimal

armitage420:

Fixing jackson parser code to increase precision of decimal values during deserialisation.

Jackson tree parser incorrectly parses decimal values, that leads to data issues. [Please see jira to reproduce steps and see the wrong values]

Yes

No

My test case and current testclasses


ql/src/test/results/clientpositive/llap/json_serde3.q.out
```diff
@@ -131,7 +143,7 @@ POSTHOOK: type: QUERY
 POSTHOOK: Input: default@json_serde3_2
 #### A masked pattern was here ####
 -2	FALSE	NULL	TRUE	1.23E45	value
--2	false	NULL	true	1.23E45	value
+-2	false	NULL	true	1.23E+45	value
```
okumin, line 146: Why would this expression change? Is it a valid change?

armitage420: @okumin Thank you so much for the review!
This was not an expected change but definitely a valid one. It is a java issue as we are now using java.math.BigDecimal for float values. BigDecimal parses scientific notations and adds a '+' to our number.

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
```diff
@@ -172,7 +173,7 @@ public HiveJsonReader(ObjectInspector oi) {
   public HiveJsonReader(ObjectInspector oi, TimestampParser tsParser) {
     this.tsParser = tsParser;
     this.oi = oi;
-    this.objectMapper = new ObjectMapper();
+    this.objectMapper = new ObjectMapper().enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS);
```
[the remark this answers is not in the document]
armitage420: @Indhumathi27 Thank you so much for your review! 
The changes are required for all decimal types as jackson is not using BigDecimal even for big numbers. Hence the numbers are getting rounded off.

ql/src/test/results/clientpositive/llap/json_serde3.q.out
okumin: I am concerned that all applications using Hive, e.g., BI, can decode both formats or not, especially when they are developed not by Java. I could be too defensive.

armitage420: Your concern is understandable, so I looked into some BI tools that connect to Hive, specifically Power BI. Power BI offers DAX (Data Analysis Expressions) functions that allow users to format scientific notations in any required format, as both `E ` and `E+` are valid notations.

Additionally, any BI tool capable of decoding scientific notation with a negative exponent (e.g., `1.23E-45`) should also handle positive exponents without issues.

armitage420:
@Indhumathi27 @okumin I also added a new commit related to fix the empty(null byte) output for hcatalog serde table. Please have a look if/when you have the time! Thank You!

ql/src/test/results/clientpositive/llap/json_serde3.q.out
okumin: Now, I think it is OK.
Revisiting the test cases, I understand that `E` vs. `E+` is introduced when an underlying JSON includes a JSON number and the column is defined as `BINARY`. I believe not many tools hit the issue, as it is not a common practice to refer to a JSON number as a BINARY type. I also expect no standard convention for decoding a JSON number into BINARY.

serde/src/java/org/apache/hadoop/hive/serde2/lazy/LazyHiveDecimal.java
```diff
@@ -36,7 +36,7 @@ public class LazyHiveDecimal extends LazyPrimitive<LazyHiveDecimalObjectInspecto
 
   private final int precision;
   private final int scale;
-  private static final byte[] nullBytes = new byte[]{0x0, 0x0, 0x0, 0x0};
+  private static final byte[] nullBytes = new byte[]{'N', 'U', 'L', 'L'};
```
okumin, line 39: HiveDecimalObjectInspector can return null even though the decimal is non-null. That's because [the object inspector dynamically tweaks the scale](https://github.com/apache/hive/blob/rel/release-4.0.1/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/primitive/WritableHiveDecimalObjectInspector.java#L45-L50), and [the non-null value can be decoded as NULL](https://github.com/apache/hive/blob/rel/release-4.0.1/storage-api/src/java/org/apache/hadoop/hive/common/type/HiveDecimal.java#L1472-L1486). I guess that's why only `LazyHiveDecimal` handles NULL. In contrast, others, such as `LazyHiveIntervalDayTime`, don't have to handle NULL.
On the other hand, I would say it is not self-evident that `LazyHiveDecimal.writeUTF8` can decide how to encode NULL. That's because a specific SerDe might decide the encoding or escaping rules. I'm checking if my assumption is true or not

okumin: I have another question. Why did `org.apache.hadoop.hive.serde2.JsonSerDe` nullify too big numbers but `org.apache.hive.hcatalog.data.JsonSerDe` didn't? As far as I debug it, that's because [HiveJsonReader#optionallyWrapWritable](https://github.com/apache/hive/blob/5a9e289c8646649136b253841847e578326e41be/serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java#L254) converts a HiveDecimal and apply `enforcePrecisionScale` in the case of serde2. The HCatalog version doesn't enable `PRIMITIVE_TO_WRITABLE` and the original value, `1000000000000000000000000000000000000.00` in the case of the test case is used.

I guess we would see the consistent result if we applied `HiveDecimalUtils.enforcePrecisionScale` once on decoding a JSON number.

okumin:
I'd like to remove this change from the history if this PR doesn't change `hbase_decimal_decimal.q.out`.
image

ql/src/test/results/clientpositive/llap/json_serde3.q.out
armitage420: Just mentioning this here that, the `E` to `E+` transformation is done during the time jackson parser creates a jsonNode using json text/inputStream even before hitting the cases for `BINARY` or any other datatype to parse our particular json value.

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
```diff
@@ -55,9 +56,7 @@
 import org.apache.hadoop.hive.serde2.objectinspector.StructField;
 import org.apache.hadoop.hive.serde2.objectinspector.StructObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.PrimitiveObjectInspectorFactory;
-import org.apache.hadoop.hive.serde2.typeinfo.BaseCharTypeInfo;
-import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
-import org.apache.hadoop.hive.serde2.typeinfo.TimestampLocalTZTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.*;
```
okumin, line 59: This is the last request hopefully. We don't use a wildcard import.
https://github.com/apache/hive/blob/da272b45f7707c41713dbfd50d61f0ff285d099f/checkstyle/checkstyle.xml#L155

armitage420:
@Indhumathi27 @okumin I have addressed all the comments. Could you please have a look?

serde/src/java/org/apache/hadoop/hive/serde2/json/HiveJsonReader.java
armitage420: Fixed!

okumin (approved):
+1. I'm waiting for CI

PR #5585, opened 2024-12-20
HIVE-28669: Deadlock found when TxnStoreMutex trying to acquireLock

dengzhhu653:
Resolving the Deadlock when the back db is MySQL

By default MySQL default isolation level is REPEATABLE-READ, `for update` in this isolation will hold the gap lock, if multiple clients are trying to `for update` and then `insert` into the same gap, it cloud cause the deadlock.

A new `hive.metastore.have.multiple.leaders`, if it's false, then the same housekeeping tasks will not leverage the db mutex to block each other.

No

Testing the PR locally, querying from mysql information_schema.INNODB_TRX table, showing that trx_isolation_level of the new trx is 'READ COMMITTED'

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
@@ -26,7 +26,7 @@ import org.apache.hadoop.hive.metastore.api.*;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.metastore.metrics.MetricsConstants;
 import org.apache.hadoop.hive.metastore.metrics.PerfLogger;
-import org.apache.hadoop.hive.metastore.txn.TxnDummyMutex;
+import org.apache.hadoop.hive.metastore.txn.NoMutex;
 import org.apache.hadoop.hive.metastore.txn.entities.CompactionInfo;
 import org.apache.hadoop.hive.metastore.txn.TxnCommonUtils;
 import org.apache.hadoop.hive.metastore.txn.TxnStore;
@@ -72,7 +72,7 @@ public class Initiator extends MetaStoreCompactorThread {
       long abortedTimeThreshold = HiveConf
           .getTimeVar(conf, HiveConf.ConfVars.HIVE_COMPACTOR_ABORTEDTXN_TIME_THRESHOLD,
               TimeUnit.MILLISECONDS);
-      TxnStore.MutexAPI mutex = shouldUseMutex ? txnHandler.getMutexAPI() : new TxnDummyMutex();
+      TxnStore.MutexAPI mutex = shouldUseMutex ? txnHandler.getMutexAPI() : new NoMutex();
 
       // Make sure we run through the loop once before checking to stop as this makes testing
       // much easier.  The stop value is only for testing anyway and not used when called from
@@ -425,7 +425,7 @@ public class Initiator extends MetaStoreCompactorThread {
   }
 
   @Override
-  public void shouldUseMutex(boolean enableMutex) {
+  public void enforceMutex(boolean enableMutex) {
     this.shouldUseMutex = enableMutex;
   }
 }
--- a/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
@@ -26,6 +26,7 @@ import org.apache.hadoop.hive.metastore.api.*;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.metastore.metrics.MetricsConstants;
 import org.apache.hadoop.hive.metastore.metrics.PerfLogger;
+import org.apache.hadoop.hive.metastore.txn.TxnDummyMutex;
 import org.apache.hadoop.hive.metastore.txn.entities.CompactionInfo;
 import org.apache.hadoop.hive.metastore.txn.TxnCommonUtils;
 import org.apache.hadoop.hive.metastore.txn.TxnStore;
@@ -56,6 +57,7 @@ public class Initiator extends MetaStoreCompactorThread {
   private ExecutorService compactionExecutor;
 
   private boolean metricsEnabled;
+  private boolean shouldUseMutex = true;
 
   @Override
   public void run() {
@@ -70,6 +72,7 @@ public class Initiator extends MetaStoreCompactorThread {
       long abortedTimeThreshold = HiveConf
           .getTimeVar(conf, HiveConf.ConfVars.HIVE_COMPACTOR_ABORTEDTXN_TIME_THRESHOLD,
               TimeUnit.MILLISECONDS);
+      TxnStore.MutexAPI mutex = shouldUseMutex ? txnHandler.getMutexAPI() : new TxnDummyMutex();
 
       // Make sure we run through the loop once before checking to stop as this makes testing
       // much easier.  The stop value is only for testing anyway and not used when called from
@@ -78,13 +81,10 @@ public class Initiator extends MetaStoreCompactorThread {
         PerfLogger perfLogger = PerfLogger.getPerfLogger(false);
         long startedAt = -1;
         long prevStart;
-        TxnStore.MutexAPI.LockHandle handle = null;
-        boolean exceptionally = false;
 
         // Wrap the inner parts of the loop in a catch throwable so that any errors in the loop
         // don't doom the entire thread.
-        try {
-          handle = txnHandler.getMutexAPI().acquireLock(TxnStore.MUTEX_KEY.Initiator.name());
+        try (TxnStore.MutexAPI.LockHandle handle = mutex.acquireLock(TxnStore.MUTEX_KEY.Initiator.name())) {
           startedAt = System.currentTimeMillis();
           prevStart = handle.getLastUpdateTime();
 
@@ -174,16 +174,13 @@ public class Initiator extends MetaStoreCompactorThread {
 
           // Check for timed out remote workers.
           recoverFailedCompactions(true);
+          handle.releaseLocks(startedAt);
         } catch (InterruptedException e) {
           // do not ignore interruption requests
           return;
         } catch (Throwable t) {
           LOG.error("Initiator loop caught unexpected exception this time through the loop", t);
-          exceptionally = true;
         } finally {
-          if (handle != null) {
-            if (!exceptionally) handle.releaseLocks(startedAt); else handle.releaseLocks();
-          }
           if (metricsEnabled) {
             perfLogger.perfLogEnd(CLASS_NAME, MetricsConstants.COMPACTION_INITIATOR_CYCLE);
             updateCycleDurationMetric(MetricsConstants.COMPACTION_INITIATOR_CYCLE_DURATION, startedAt);
@@ -214,8 +211,6 @@ public class Initiator extends MetaStoreCompactorThread {
   private Database resolveDatabase(CompactionInfo ci) throws MetaException, NoSuchObjectException {
     return CompactorUtil.resolveDatabase(conf, ci.dbname);
   }
-
-
 
   @VisibleForTesting
   protected String resolveUserToRunAs(Map<String, String> cache, Table t, Partition p)
@@ -427,5 +422,10 @@ public class Initiator extends MetaStoreCompactorThread {
         }
       }
     }
+  }
+
+  @Override
+  public void shouldUseMutex(boolean enableMutex) {
+    this.shouldUseMutex = enableMutex;
   }
 }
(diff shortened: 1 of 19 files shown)


Aggarwal-Raghav:
Based on the comments on JIRA and I have tested this patch on my local and haven't seen the error stacktrace, hence
**LGTM +1 (non-binding)**

deniskuzZ:
how would that work if we have multiple HMS versions in a cluster?

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
```diff
@@ -174,16 +174,13 @@ public void run() {
 
           // Check for timed out remote workers.
           recoverFailedCompactions(true);
+          handle.releaseLocks(startedAt);
```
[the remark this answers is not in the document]
dengzhhu653: @Indhumathi27, Sorry for the late reply!
`handle.releaseLocks(startedAt);` this is for marking the starting point of previous run, if there is no exception, the Initiator will use the `startedAt` as the start to find the potential compactions:
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java#L89
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java#L117

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
```diff
@@ -78,13 +81,10 @@ public void run() {
...
         long startedAt = -1;
         long prevStart;
-        TxnStore.MutexAPI.LockHandle handle = null;
-        boolean exceptionally = false;
 
         // Wrap the inner parts of the loop in a catch throwable so that any errors in the loop
         // don't doom the entire thread.
-        try {
-          handle = txnHandler.getMutexAPI().acquireLock(TxnStore.MUTEX_KEY.Initiator.name());
+        try (TxnStore.MutexAPI.LockHandle handle = mutex.acquireLock(TxnStore.MUTEX_KEY.Initiator.name())) {
```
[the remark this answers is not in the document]
dengzhhu653: We need to call `getLastUpdateTime()` and `releaseLocks(Long timestamp)` of `TxnStore.MutexAPI.LockHandle` in this class, so not use the `AutoCloseable`.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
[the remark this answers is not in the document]
dengzhhu653: If exception happens before updating the current `startedAt`, this AUX table with startedAt will not get updated, next run the `Initiator` will use the former one(`startAt`) or -1 if there is no such record to search through db to find the potential compactions, and it will skip the queued or running candidates.

deniskuzZ:
@dengzhhu653, before HIVE-27481 we always used `READ_COMMITTED` isolation level in TxnHandler:
````
@Override
  public LockHandle acquireLock(String key) throws MetaException {
    /**
     * The implementation here is a bit kludgey but done so that code exercised by unit tests
     * (which run against Derby which has no support for select for update) is as similar to
     * production code as possible.
     * In particular, with Derby we always run in a single process with a single metastore and
     * the absence of For Update is handled via a Semaphore.  The later would strictly speaking
     * make the SQL statements below unnecessary (for Derby), but then they would not be tested.
     */
    Connection dbConn = null;
    Statement stmt = null;
    ResultSet rs = null;
    boolean needToCloseConn = true;
    try {
      try {
        String sqlStmt = sqlGenerator.addForUpdateClause("SELECT \"MT_COMMENT\", \"MT_KEY2\" FROM \"AUX_TABLE\" WHERE \"MT_KEY1\"=" + quoteString(key));
        lockInternal();
        dbConn = getDbConn(Connection.TRANSACTION_READ_COMMITTED, connPoolMutex);
        stmt = dbConn.createStatement();
        LOG.debug("About to execute SQL: {}", sqlStmt);
        rs = stmt.executeQuery(sqlStmt);
````
if that is changed to default, should we restore the `READ_COMMITTED`?

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnDummyMutex.java [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+package org.apache.hadoop.hive.metastore.txn;
+
+import org.apache.hadoop.hive.metastore.api.MetaException;
+
+/**
+ * An empty implementation of TxnStore.MutexAPI
+ */
+public class TxnDummyMutex implements TxnStore.MutexAPI {
```
deniskuzZ, line 25: should we rename to `DummyTxnStoreMutex`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/TransactionContextManager.java
```diff
@@ -54,11 +57,17 @@ public class TransactionContextManager {
    * The created transaction is wrapped into a {@link TransactionContext} which is {@link AutoCloseable} and allows using
    * the wrapper inside a try-with-resources block.
    *
-   * @param propagation The transaction propagation to use.
+   * @param transactional the transactional definition.
    */
-  public TransactionContext getNewTransaction(int propagation) {
-    TransactionContext context = new TransactionContext(realTransactionManager.getTransaction(
-        new DefaultTransactionDefinition(propagation)), this);
+  public TransactionContext getNewTransaction(Transactional transactional) {
```
deniskuzZ, line 62: why not create a parameterless method and use  Propagation.REQUIRED as default?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/AbortCompactionFunction.java
```diff
@@ -152,7 +150,7 @@ public CompactionAborter withCompactionInfo(CompactionInfo compactionInfo) {
 
     @Override
     public AbortCompactionResponseElement execute() {
-      try (TransactionContext context = jdbcResource.getTransactionManager().getNewTransaction(PROPAGATION_REQUIRED)) {
+      try (TransactionContext context = jdbcResource.getTransactionManager().getNewTransaction(null)) {
```
deniskuzZ, line 153: passing `null` is not a good practice

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/leader/LeaderElection.java
```diff
@@ -31,6 +31,11 @@
  * @param <T> the type of mutex
  */
 public interface LeaderElection<T> extends Closeable {
+  // We might have different versions of HMS, or even the same version but with different
+  // leader election methods running inside the warehouse, so it's hard to know how many HMS instances
+  // that elected as the leader. Relying on this property to tell us, default is true, means it has multiple
+  // HMS instances acting as the leader.
+  public static final String HAVE_MULTIPLE_LEADERS = "hive.metastore.have.multiple.leaders";
```
deniskuzZ, line 38: maybe better name it like `hive.metastore.enforce.aux.mutex`?

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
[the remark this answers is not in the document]
deniskuzZ: what if the leader changes here? lock won't be ever released?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/TransactionContextManager.java
dengzhhu653: this `transactional` is retrieved from the method annotation, 
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnStore.java#L281
each method might have its own definition, though we only care about the isolation level and propagation now

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/functions/AbortCompactionFunction.java
dengzhhu653: I tried to create an instance of `Transactional`, but it's hard to implement an annotation

dengzhhu653:
> @dengzhhu653, before [HIVE-27481](https://issues.apache.org/jira/browse/HIVE-27481) we always used `READ_COMMITTED` isolation level in TxnHandler:
> 
> ```
> @Override
>   public LockHandle acquireLock(String key) throws MetaException {
>     /**
>      * The implementation here is a bit kludgey but done so that code exercised by unit tests
>      * (which run against Derby which has no support for select for update) is as similar to
>      * production code as possible.
>      * In particular, with Derby we always run in a single process with a single metastore and
>      * the absence of For Update is handled via a Semaphore.  The later would strictly speaking
>      * make the SQL statements below unnecessary (for Derby), but then they would not be tested.
>      */
>     Connection dbConn = null;
>     Statement stmt = null;
>     ResultSet rs = null;
>     boolean needToCloseConn = true;
>     try {
>       try {
>         String sqlStmt = sqlGenerator.addForUpdateClause("SELECT \"MT_COMMENT\", \"MT_KEY2\" FROM \"AUX_TABLE\" WHERE \"MT_KEY1\"=" + quoteString(key));
>         lockInternal();
>         dbConn = getDbConn(Connection.TRANSACTION_READ_COMMITTED, connPoolMutex);
>         stmt = dbConn.createStatement();
>         LOG.debug("About to execute SQL: {}", sqlStmt);
>         rs = stmt.executeQuery(sqlStmt);
> ```
> 
> if that is changed to default, should we restore the `READ_COMMITTED`?

I have made the default as the `READ_COMMITTED` when calling the TxnStore's method unless it has specified the level:
https://github.com/apache/hive/pull/5585/files#diff-63125a22ca15ef76a1e6b5a52d230e9b24c816cbd15442e3ba640da2f4c167d0R113
https://github.com/apache/hive/pull/5585/files#diff-080bfb902495302473126b9827a086cf164e63d8742c3b80082469b01b405a48R66-R67

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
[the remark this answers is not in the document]
dengzhhu653: if the leader changes, the initiator will be interrupted and the `stop` is set to true:
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java#L195
the initiator will exit finally in the next run.
In any cases(db mutex or no mutex), the lock(try-with-resource) will be released as Initiator tries to acquire a new lock in the inner loop.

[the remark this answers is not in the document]
deniskuzZ: Oh, I missed that we use AutoCloseable now

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/TransactionContextManager.java
deniskuzZ: what if we keep the original signature and introduce a change in
````
context = jdbcResource.getTransactionManager().getNewTransaction(
  Optional.ofNullable(transactional).map(Transactional::propagation)
      .orElse(Propagation.REQUIRES_NEW));

....
  public TransactionContext getNewTransaction(Propagation propagation) {
    TransactionContext context = new TransactionContext(realTransactionManager.getTransaction(
        new DefaultTransactionDefinition(propagation.value())), this);
....
}
````

deniskuzZ: so, you decided to get rid of the isolation level? that's fine, but in that case, why do we need 2nd method without the propagation arg?

deniskuzZ (approved):
LGTM +1, pending tests

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/jdbc/TransactionContextManager.java
dengzhhu653: done, removed the 2nd method

PR #5586, opened 2024-12-20
HIVE-28675: Maximize the removal of redundant columns from GROUP BY clauses

zabetak:
Enhance `HiveRelFieldTrimmer` to remove the maximum number of redundant columns from the `GROUP BY` clause.

1. Generate more efficient plans by pruning as many columns as possible (less CPU/IO/network cost).
2. Avoid missing optimization opportunities by examining all candidates.

For more see HIVE-28675.

More efficient query plans.

No

```
mvn test -Dtest=TestMiniLlapLocalCliDriver.java -Dqfile="cbo_groupby_remove_key.q"
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveRelFieldTrimmer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveRelFieldTrimmer.java
@@ -502,28 +502,15 @@ public class HiveRelFieldTrimmer extends RelFieldTrimmer {
       return generateGroupSetIfCardinalitySame(aggregate, originalGroupSet, fieldsUsed);
     }
 
-    // we have set of unique key, get to the key which is same as group by key
-    ImmutableBitSet groupByUniqueKey = null;
-
+    // Find the maximum number of columns that can be removed by retaining a certain unique key
+    ImmutableBitSet columnsToRemove = ImmutableBitSet.of();
     for (ImmutableBitSet key : uniqueKeys) {
-      if (aggregate.getGroupSet().contains(key)) {
-        groupByUniqueKey = key;
-        break;
+      ImmutableBitSet removableCols = originalGroupSet.except(key).except(fieldsUsed);
+      if (aggregate.getGroupSet().contains(key) && removableCols.cardinality() > columnsToRemove.cardinality()) {
+        columnsToRemove = removableCols;
       }
     }
-
-    if (groupByUniqueKey == null) {
-      // group by keys do not represent unique keys
-      return originalGroupSet;
-    }
-
-    // we know group by key contains primary key and there is at least one column in group by which is not being used
-    // if that column is not part of key it should be removed
-    ImmutableBitSet nonKeyColumns = aggregate.getGroupSet().except(groupByUniqueKey);
-    ImmutableBitSet columnsToRemove = nonKeyColumns.except(fieldsUsed);
-    ImmutableBitSet newGroupSet = aggregate.getGroupSet().except(columnsToRemove);
-
-    return  newGroupSet;
+    return aggregate.getGroupSet().except(columnsToRemove);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveRelFieldTrimmer.java
```diff
@@ -502,28 +502,15 @@ private ImmutableBitSet generateNewGroupset(Aggregate aggregate, ImmutableBitSet
...
-    // we have set of unique key, get to the key which is same as group by key
-    ImmutableBitSet groupByUniqueKey = null;
-
+    // Find the maximum number of columns that can be removed by retaining a certain unique key
+    ImmutableBitSet columnsToRemove = ImmutableBitSet.of();
     for (ImmutableBitSet key : uniqueKeys) {
-      if (aggregate.getGroupSet().contains(key)) {
-        groupByUniqueKey = key;
-        break;
+      ImmutableBitSet removableCols = originalGroupSet.except(key).except(fieldsUsed);
```
soumyakanti3578, line 508: Is it possible to compute `except(fieldsUsed)` outside the loop? I believe changing the order of `except`s will still yield the same resulting set.

ramesh0201: Yes we can do this, but just want to be careful that if there is no uniquekey match then we should not remove any columns at all(should skip the except in the return statement below in that case)

ramesh0201: Just realized that we only update columnsToRemove when there is atleast one match. So I think yes we can move the except(fieldsUsed) outside the for loop.
Also is it efficient to loop the fieldsUsed(and then check if there is a field that is a unique key and part of aggregate keys to retain only fieldsUsed) vs looping uniquekeys? Mostly depends on size of fieldsUsed vs uniqueKeys. I will probably leave it upto you to decide on that.

zabetak: Yes, it is possible to compute `except(fieldsUsed)` outside the loop. I applied the suggestion in https://github.com/apache/hive/pull/5586/commits/e89386dc64bd4eb8a3981795b3cacd264861ee35.

zabetak: Comparing the efficiency of the iteration between `fieldsUsed` and `uniqueKeys` is not possible because the semantics are different. The `uniqueKeys` variable is a set of sets (`Set<ImmutableBitSet>`) while `fieldsUsed` is a single set (`ImmutableBitSet`). Note that a `key` is not necessarily a single column but a set of columns (composite key).

soumyakanti3578 (approved):
LGTM!

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveRelFieldTrimmer.java
ramesh0201: Thank you for your explanation. Makes sense to me.

ramesh0201:
LGTM +1

ql/src/test/queries/clientpositive/cbo_groupby_remove_key.q
```diff
@@ -0,0 +1,18 @@
...
+    UNIQUE (fname, lname) DISABLE RELY
+);
+
+EXPLAIN CBO SELECT id, COUNT(1) FROM passenger GROUP BY id, passport;
+EXPLAIN CBO SELECT passport, COUNT(1) FROM passenger GROUP BY id, passport;
+EXPLAIN CBO SELECT id, COUNT(1) FROM passenger GROUP BY id, fname, lname, passport;
+EXPLAIN CBO SELECT passport, COUNT(1) FROM passenger GROUP BY id, fname, lname, passport;
+EXPLAIN CBO SELECT fname, COUNT(1) FROM passenger GROUP BY id, fname, lname, passport;
+EXPLAIN CBO SELECT lname, COUNT(1) FROM passenger GROUP BY id, fname, lname, passport;
+EXPLAIN CBO SELECT fname, lname, COUNT(1) FROM passenger GROUP BY id, fname, lname, passport;
```
ramesh0201, line 18: In this case, is having group by fname,lname always a better plan? -- even if there is a different aggregate function?

zabetak: If we change the aggregate function the query changes as well. 

If we assume that the set of redundant columns remains the same after changing the aggregate function (e.g., `MAX(fname)`) then it's always a good idea to drop unnecessary processing so grouping on `fname, lname` is the best option. 

If the set of redundant columns changes (e.g., `MAX(passport)`) then this optimization will not trigger in the same way (`passport` column cannot be dropped).

PR #5587, opened 2024-12-22
HIVE-28677: Implement direct sql for delete table/partition column stats

wecharyu:
Improve the performance of deleting column stats for wide table.

The JDO implementation of deleting column stats needs to retrieve the column stats objects, it could be slow if the table has many columns. 

No.

No.

Add some unit tests:
```bash
mvn test -Dtest.groups= -Dtest=org.apache.hadoop.hive.metastore.TestObjectStore#testPartitionStatisticsOps -pl :hive-standalone-metastore-server
```

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
@@ -3260,7 +3260,7 @@ class MetaStoreDirectSql {
 
   public boolean deletePartitionColumnStats(String catName, String dbName, String tblName,
       String partName, String colName, String engine) throws MetaException {
-    String sqlFilter = "" + PARTITIONS + ".\"PART_NAME\" = ? ";
+    String sqlFilter = PARTITIONS + ".\"PART_NAME\" = ? ";
     List<Long> partitionIds = getPartitionIdsViaSqlFilter(catName, dbName, tblName, sqlFilter,
         Arrays.asList(partName), Collections.emptyList(), -1);
     assert(partitionIds.size() == 1);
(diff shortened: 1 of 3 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3241,6 +3241,46 @@ public void deleteColumnStatsState(long tbl_id) throws MetaException {
...
+    } catch (SQLException e) {
+      LOG.warn("Error removing table column stats. ", e);
+      return false;
+    }
+    return true;
+  }
+
+  public boolean deletePartitionColumnStats(String catName, String dbName, String tblName,
+      String partName, String colName, String engine) throws MetaException {
+    String sqlFilter = "" + PARTITIONS + ".\"PART_NAME\" = ? ";
```
Aggarwal-Raghav, line 3263: Empty String can be removed from the value of `sqlFilter`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10234,30 +10234,47 @@ protected Integer getJdoResult(GetHelper<Integer> ctx) throws MetaException, NoS
...
+    MPartition mPartition = getMPartition(catName, dbName, tableName, partName);
+    if (mPartition == null) {
+      throw new NoSuchObjectException("Partition " + partName
+          + " for which stats deletion is requested doesn't exist");
+    }
+
+    return new GetHelper<Boolean>(catName, dbName, tableName, true, true) {
+      @Override
+      protected String describeResult() {
+        return "delete prtition column stats";
```
Aggarwal-Raghav, line 10254: nit: partition spelling.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
wecharyu: Addressed.

zhangbutao (approved):
+1 LGTM

PR #5588, opened 2024-12-24
HIVE-28486: Upgrade commons-collections to commons-collections4 due t…

eshwitha:
…o EOL.

Upgrade commons-collections:commons-collections to org.apache.commons:commons-collections4 as it has reached EOL.

[https://mvnrepository.com/artifact/commons-collections/commons-collections](https://mvnrepository.com/artifact/commons-collections/commons-collections) is an EOL library and needs to be migrated to commons-collections4.


Aggarwal-Raghav:
@eshwitha , I have tried to migrate from commons-collection 3.x to 4.x  in past **but complete migration is not achievable**. IIRC, there are 2 reasons:
1. accumulo version used in hive 1.10.4 uses commons-collections-3.x. I observed UT failure in accumulo module.
2. commons-beanutils 1.9.4 also uses commons-collections-3.x. Check [BEANUTILS-532](https://issues.apache.org/jira/browse/BEANUTILS-532), based on the recent comments on the JIRA, there is a plan to release but not sure when.

with your PR also, in packaging lib dir both the version of commons-collections i.e. 3.x and 4.x are present.

pom.xml
```diff
@@ -385,15 +384,10 @@
...
-      <dependency>
-        <groupId>commons-collections</groupId>
-        <artifactId>commons-collections</artifactId>
-        <version>${commons-collections.version}</version>
-      </dependency>
       <dependency>
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-collections4</artifactId>
-        <version>${commons-collections4.version}</version>
+        <version>4.4</version>
```
Aggarwal-Raghav, line 390: Refrain from using hardcoded version

zhangbutao: @eshwitha Could you fix this comment?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
```diff
@@ -18,7 +18,7 @@
 package org.apache.hadoop.hive.metastore.txn;
 
 import com.google.common.annotations.VisibleForTesting;
-import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections4.CollectionUtils;
```
Aggarwal-Raghav, line 21: Just thinking out loud, I don't see any commons-collections-3.x dependency present in standalone-metastore module, so commons-collections-3.x must be coming from some transitive dependency (either from hadoop/atlas/commons-beanutils), now in this PR we will start shipping commons-collections-4.x in `standalone-metastore/metastore-server/target/apache-hive-metastore-4.1.0-SNAPSHOT-bin`. Can it cause any  classpath issue?

pom.xml
eshwitha: Addressed the issue and updated the PR.

PR #5590, opened 2024-12-25
HIVE-28518: Iceberg: Fix ClassCastException during in-place migration to Iceberg tables with timestamp columns

ggangadharan:

This fix improves the stability and reliability of in-place migrated Iceberg tables involving timestamp data types.

The issue occurred due to incorrect type casting in the timestamp handling logic, which caused the migrated Iceberg tables Fetch task to fail. 

No

No

Qtest - iceberg_inplace_migration_with_timestamp_column.q


Aggarwal-Raghav:
@ggangadharan , thanks for the PR. I have 1 question though.
I tried to run the q file **_without the patch_** with PARQEUT and ORC. I am seeing the difference in behaviour. **In ORC, without the patch also it is working.**  is it because of the way ORC and parquet store date/time type?
Attaching screenshot for same. 
With ORC:
ORC
With Parquet:
Parquet

ggangadharan:
Hi @Aggarwal-Raghav 

Thank you for raising this question.

Upon investigation, it appears that the issue stems from how the IcebergRecordReader interprets the timestamp column for different file formats:

- For ORC tables, the timestamp column is read as LOCALDATETIME.
- For Parquet tables, the same column is read as OFFSETDATETIME.

Due to this discrepancy, we are encountering a **ClassCastException** when working with Parquet tables.

As you mentioned, I also believe the root cause lies in the underlying file format/Iceberg level.

I’ve attached a screenshot for reference.

HIVE-28518_DEBUG


Please let me know if you need further details or if we should take any additional steps to address this. 

if it looks okay , Please review the PR.

ggangadharan:
Hi @okumin , 

Thank you for taking the time to review the pull request.

In the Iceberg Parquet table, the timestamp column is read as **LOCALDATETIME**. I’ve attached a screenshot for reference.

Screenshot 2024-12-28 at 8 18 14 AM

There is a notable difference in how the timestamp column is stored at the Parquet file format level. Specifically:

- In Iceberg Parquet tables, the timestamp column is stored as **INT64 L:TIMESTAMP(MICROS,false)** .
- In standard Parquet tables, the timestamp column is stored as **INT96** .

For clarity, I’ve also included the metadata from Parquet-tools for reference.

**As Iceberg Parquet table** 

```
file schema: table
------------------------------------------------------------------------------------------------------------------------------------------------------------
id:          OPTIONAL INT32 R:0 D:1
name:        OPTIONAL BINARY L:STRING R:0 D:1
dt:          OPTIONAL INT64 L:TIMESTAMP(MICROS,false) R:0 D:1

row group 1: RC:1 TS:112 OFFSET:4
------------------------------------------------------------------------------------------------------------------------------------------------------------
id:           INT32 SNAPPY DO:0 FPO:4 SZ:35/33/0.94 VC:1 ENC:BIT_PACKED,RLE,PLAIN ST:[min: 1, max: 1, num_nulls: 0]
name:         BINARY SNAPPY DO:0 FPO:39 SZ:44/42/0.95 VC:1 ENC:BIT_PACKED,RLE,PLAIN ST:[min: test name, max: test name, num_nulls: 0]
dt:           INT64 SNAPPY DO:0 FPO:83 SZ:39/37/0.95 VC:1 ENC:BIT_PACKED,RLE,PLAIN ST:[min: 2024-08-09T14:08:26.326107, max: 2024-08-09T14:08:26.326107, num_nulls: 0]
```

**As standard parquet table** 

```
file schema: hive_schema
------------------------------------------------------------------------------------------------------------------------------------------------------------
id:          OPTIONAL INT32 R:0 D:1
name:        OPTIONAL BINARY L:STRING R:0 D:1
dt:          OPTIONAL INT96 R:0 D:1

row group 1: RC:1 TS:137 OFFSET:4
------------------------------------------------------------------------------------------------------------------------------------------------------------
id:           INT32 UNCOMPRESSED DO:0 FPO:4 SZ:33/33/1.00 VC:1 ENC:BIT_PACKED,RLE,PLAIN ST:[min: 1, max: 1, num_nulls: 0]
name:         BINARY UNCOMPRESSED DO:0 FPO:37 SZ:42/42/1.00 VC:1 ENC:BIT_PACKED,RLE,PLAIN ST:[min: test name, max: test name, num_nulls: 0]
dt:           INT96 UNCOMPRESSED DO:79 FPO:110 SZ:62/62/1.00 VC:1 ENC:BIT_PACKED,RLE,PLAIN_DICTIONARY ST:[min: 0x78037C8D4C2E0000748B2500, max: 0x78037C8D4C2E0000748B2500, num_nulls: 0]
```

okumin:
Thanks. I also remember Hive didn't follow the regular convention on encoding TIMESTAMP. I don't have an immediate idea on how to fix it

ayushtkn:
We should try to fix the regular convention on encoding TIMESTAMP, but that might not fix the case with existing tables, For those the fix in the current PR seems ok to me

ggangadharan:
@ayushtkn Thank you for the feedback. 

Based on this, I believe the changes are in a good state to proceed unless there are further concerns.

@Aggarwal-Raghav  @okumin  Could you kindly review the code and share your feedback? Your insights would be greatly appreciated to help move this forward. If there are any questions or blockers, feel free to let me know.

Thank you for your time and support!

okumin:
Thank you! It is obvious. So, is the remaining problem to verify INT96 can be compatible with Iceberg's TIMESTAMP, which means verifying other query engines or tools can read it as a timestamp. I am trying to check it.

ggangadharan:
@okumin I have shared the output snippets from SPARK side in my previous response. Please let me know if there's anything else required from my side to help resolve this issue.

okumin:
@ggangadharan Thanks. I think you provided sufficient information. I was testing Spark and Trino, but the progress has not been so good. I faced another problem while testing them, and it's been a holiday season in my country. I would appreciate it if you could give us time

ggangadharan:
@okumin  Thanks for the update. Let me know if there's anything needed from my end. Will check and do the needful from my side.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/serde/objectinspector/IcebergTimestampObjectInspectorHive3.java
```diff
@@ -55,7 +56,13 @@ public Timestamp getPrimitiveJavaObject(Object o) {
     if (o == null) {
       return null;
     }
-    LocalDateTime time = (LocalDateTime) o;
+    LocalDateTime time;
+    if (o instanceof OffsetDateTime) {
+      OffsetDateTime odt = (OffsetDateTime) o;
+      time = odt.atZoneSameInstant(TypeInfoFactory.timestampLocalTZTypeInfo.getTimeZone()).toLocalDateTime();
+    } else {
+      time = (LocalDateTime) o;
```
zratkai, line 64: It would be safe to check if it is LocalDateTime before casting.

ggangadharan: @zratkai  Thanks for taking a look into this. 

Sure, I will make the necessary changes. Besides LocalDateTime and OffsetDateTime, are there any other classes that might be expected here? Please let me know so I can handle them appropriately.

okumin:
I have checked how we handle timestamps for the past few days, and this is the summary.
https://gist.github.com/okumin/b33876398a106068b4fd7ddd3049ae8c

As far as I tested, I didn't find any patterns where this doesn't work so far. So, I am thinking to review this and give +1. I might ask someone for one more look because I might not have comprehensive information.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/serde/objectinspector/IcebergTimestampObjectInspectorHive3.java
okumin: This should always be LocalDateTime, ideally. But the Iceberg + Parquet reader constructs OffsetDateTime. So, we should support only LocalDateTime and OffsetDateTime.

Though we expected never fail, it could be a little kinder if we had a better error message like `String.format("An unexpected type %s was passed as timestamp", o.getClass)`

okumin (approved):
@ayushtkn @zabetak While investigating the history of TIMESTAMP and related tickets, I thought you would likely be knowledgeable about this area. I'd appreciate it if you could take another look at it.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/serde/objectinspector/IcebergTimestampObjectInspectorHive3.java
ggangadharan: Thanks for the detailed update @okumin 

if it's okay. I will make the below code changes.

` public Timestamp getPrimitiveJavaObject(Object o) {
    if (o == null) {
      return null;
    }
    LocalDateTime time;
    if (o instanceof LocalDateTime) {
      time = (LocalDateTime) o;
    } else if (o instanceof OffsetDateTime) {
      OffsetDateTime odt = (OffsetDateTime) o;
      time = odt.atZoneSameInstant(TypeInfoFactory.timestampLocalTZTypeInfo.getTimeZone()).toLocalDateTime();
    } else {
      throw new ClassCastException(String.format("An unexpected type %s was passed as timestamp. " +
              "Expected LocalDateTime/OffsetDateTime", o.getClass().getName()));
    }
    return Timestamp.ofEpochMilli(time.toInstant(ZoneOffset.UTC).toEpochMilli(), time.getNano());
  }`

Please check and let me know.

ayushtkn: Should be ok

ggangadharan: Thanks for the update @ayushtkn

ggangadharan:
@ayushtkn  Updated the changes in the recent commit, and all tests have passed. Please review.

ayushtkn (approved):
LGTM

PR #5600, opened 2025-01-07
HIVE-28619: TestHivePrivilegeObjectOwnerNameAndType to run on Tez

Indhumathi27:
This test class is intended to validate the HivePrivilegeObject for queries. If the current user is changed, insert queries fail when executed on Tez due to the validation of the staging directory's owner. The staging directory is created by the system user when the test is executed in local file system mode. Since the main goal is to validate the privilege object type and count, without checking the owner name, the simplest fix is to run the tests as the system user.

Since MR is deprecated, remove the test dependency as well, by running with Tez

No

No

Test pipeline


Aggarwal-Raghav:
LGTM +1 (non-binding)

okumin (approved):
+1

> The staging directory is created by the system user when the test is executed in local file system mode.

I can agree with this statement.

> Since the main goal is to validate the privilege object type and count, without checking the owner name, the simplest fix is to run the tests as the system user.

It would be great if we could also check the name as String in `testOwnerNames`, but I think it is not easy because of the constraint of the local mode.

PR #5603, opened 2025-01-09
HIVE-28700: MRCompactor may cause data loss when performing the major…

dengzhhu653:
… compaction

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/MRCompactor.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/MRCompactor.java
@@ -571,11 +571,9 @@ public class MRCompactor implements Compactor {
           attemptId = Integer.valueOf(attemptIdString);
         }
         deltasToAttemptId.put(deltas[i].getName(), attemptId);
-        if (baseAttemptId != null) {
-          deltasToAttemptId.put(base.getName(), baseAttemptId);
-        }
       }
       if (baseAttemptId != null) {
+        deltasToAttemptId.put(base.getName(), baseAttemptId);
         deltasToAttemptId.put(base.toString(), baseAttemptId);
       }
     }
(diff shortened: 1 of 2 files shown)


ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/MRCompactor.java
```diff
@@ -571,11 +571,9 @@ public void readFields(DataInput dataInput) throws IOException {
           attemptId = Integer.valueOf(attemptIdString);
         }
         deltasToAttemptId.put(deltas[i].getName(), attemptId);
-        if (baseAttemptId != null) {
-          deltasToAttemptId.put(base.getName(), baseAttemptId);
-        }
       }
       if (baseAttemptId != null) {
+        deltasToAttemptId.put(base.getName(), baseAttemptId);
```
deniskuzZ, line 576: what if we use full path in OrcRawRecordMerger to fetch the attemptId?
````
if (deltasToAttemptId != null) {
  attemptId = deltasToAttemptId.get(mergerOptions.getBaseDir().toString());
}
````

File: ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/MRCompactor.java [hunk 2]
```diff
@@ -571,11 +571,9 @@ public void readFields(DataInput dataInput) throws IOException {
...
         deltasToAttemptId.put(base.toString(), baseAttemptId);
```
kuczoram, line 577: I am wondering if this is even needed. When this code part was added the full path was used in the map and later it was changed to use only the name. I was checking today if there is any use-case where the full path is needed but I couldn't find one. Also in the code this map is used with the directory names only. So I strongly believe that this was just accidentally left this way and it would be enough to use just the base.getName().
What Denys wrote is also a good option, just don't forget to adjust the MRCompactor.write method if you go with this solution.

deniskuzZ: nice! I think using `name` is a better approach (it would be consistent for base and deltas). Executed `TestCompactor` tests, all passed.
It seems that the code branch that was deleted, was accidentally added into a delta elem loop

dengzhhu653: done

deniskuzZ (approved):
+1, pending tests

kuczoram (approved):
+1 pending tests

PR #5604, opened 2025-01-09
HIVE-28701: OTEL: Fix race condition due to unavailability of Query Id.

ayushtkn:

Skip publishing LiveQuery until it has a QueryId allocated.
 
To avoid orphan span due to `null` queryId

No

No

Manual


tanishq-chugh (approved):
LGTM

simhadri-g (approved):

PR #5608, opened 2025-01-12
HIVE-28703: Integral Data Type Overflow Not Enforced for JsonSerDe and UDFs

armitage420:

The pull request addresses the issue of data type overflow not being enforced when using JsonSerDe in Apache Hive. The proposed changes aim to ensure strict data type validation during the insertion and loading of data into tables defined with types like TINYINT, SMALLINT, INT, and BIGINT.

This behavior can lead to data integrity issues and unexpected results.

Yes

No

Tests introduced in this PR and local machine


armitage420:
I am able to reproduce the error related to qt_database_postgres.q on master branch too(without my changes):
image

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFToByte.java
```diff
@@ -128,6 +128,11 @@ public ByteWritable evaluate(IntWritable i) {
     if (i == null) {
       return null;
     } else {
+      int value = i.get();
+      if (value < Byte.MIN_VALUE || value > Byte.MAX_VALUE) {
+        throw new IllegalArgumentException("Value out of range for Byte: " + value);
```
difin, line 133: It is better to use Hive-specific exceptions. Hive has an exception `UDFArgumentException` which seems suitable here.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFToInteger.java
```diff
@@ -146,6 +146,10 @@ public IntWritable evaluate(LongWritable i) {
     if (i == null) {
       return null;
     } else {
+      long value = i.get();
+      if (value < Integer.MIN_VALUE || value > Integer.MAX_VALUE) {
+        throw new IllegalArgumentException("Value out of range for Integer: " + value);
```
difin, line 151: Same here, generic exception.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFToShort.java
```diff
@@ -128,6 +128,10 @@ public ShortWritable evaluate(IntWritable i) {
     if (i == null) {
       return null;
     } else {
+      int value = i.get();
+      if (value < Short.MIN_VALUE || value > Short.MAX_VALUE) {
+        throw new IllegalArgumentException("Value out of range for Short: " + value);
```
difin, line 133: same, can you please use Hive specific exceptions

armitage420:
@difin Gentle reminder!

armitage420:
Seems @difin is offline.
@okumin @chinnaraolalam @ayushtkn @deniskuzZ Could you please review this PR if/when you have the time?

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFToByte.java
```diff
@@ -124,10 +125,14 @@ public ByteWritable evaluate(ShortWritable i) {
...
    */
-  public ByteWritable evaluate(IntWritable i) {
+  public ByteWritable evaluate(IntWritable i) throws UDFArgumentException {
     if (i == null) {
       return null;
     } else {
+      int value = i.get();
+      if (value < Byte.MIN_VALUE || value > Byte.MAX_VALUE) {
+        throw new UDFArgumentException("Value out of range for Byte: " + value);
+      }
```
okumin, line 135: I know this obeys the ANSI standard, but it is a breaking change for Hive. Hive's cast is generally known to be very generous, so I think we can't change the behavior without introducing a flag.

armitage420: Hey @okumin, thank you so much for the review!

Your suggestion makes sense. I have set the default value as true.

armitage420:
@okumin Gentle reminder

okumin:
@ armitage420 I'm sorry for not getting back to you sooner. I got COVID, and I'm not perfect yet. So, let me give you short feedback first.

I think "Integral Data Type Overflow Not Enforced for JsonSerDe and UDFs" is misleading because this PR is changing the fundamental semantics of Hive and affecting all type conversions, including pure `CAST(col AS X)` or non-JSON SerDe. The blast radius is too large. I believe you have to propose changing the semantics of type conversions using [the mailing list](https://hive.apache.org/community/mailinglists/) and agree with the direction with the community if you want to change everything. I guess the initiative's title will be "Introducing ANSI-compatible type conversions" or else.

PR #5615, opened 2025-01-17
HIVE-28033: Fix TestMapJoinOperator to run on Tez

Indhumathi27:

The configuration for HIVE_DEFAULT_NULLS_LAST is set to false when Tez is used. The change ensures compatibility with Tez and proper null handling during serialization.

With MapReduce (MR),  SERIALIZATION_NULL_SORT_ORDER is not set, hence the default null sorting is NULL FIRST. For Tez, the null sorting needed to be adjusted to prevent EOFException and ensure correct test execution with the generated data.

No

No

Test case

### DIFF
--- a/ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/TestMapJoinOperator.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/TestMapJoinOperator.java
@@ -18,43 +18,19 @@
 
 package org.apache.hadoop.hive.ql.exec.vector.mapjoin;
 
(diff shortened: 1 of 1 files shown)


Aggarwal-Raghav:
LGTM +1 (non-binding)

difin:
LGTM, +1

difin (approved):

okumin:
I confirmed this part is applied only to Tez and that's why.
https://github.com/apache/hive/blob/eaa8189600c74efa35cc9975eea0f81df59e5088/ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java#L524-L534

okumin (approved):
+1

PR #5616, opened 2025-01-21
HIVE-28716: Add owner information in HivePrivilegeObject for create database query

Aggarwal-Raghav:
Refer to [HIVE-28716](https://issues.apache.org/jira/browse/HIVE-28716) and [RANGER-5097](https://issues.apache.org/jira/browse/RANGER-5097) for the info

Currently we are not passing the owner information for database in HivePrivilegeObject. This object is passed to ranger and ranger is using this owner information but as it is missing, NoSuchObjectException is thrown HS2 logs. Check the above mentioned JIRA for stacktrace and debugging screenshot. 

**NOTE: There is no impact on the functionality, table/db is getting created in hive. It more of a noise in the HS2 logs coming from ranger.** 

No

No

On Cluster testing and monitoring HS2 logs

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseAnalyzer.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import org.apache.hadoop.hive.metastore.api.DataConnector;
 import org.apache.hadoop.hive.metastore.api.Database;
 import org.apache.hadoop.hive.metastore.api.DatabaseType;
+import org.apache.hadoop.hive.metastore.api.PrincipalType;
 import org.apache.hadoop.hive.ql.QueryState;
 import org.apache.hadoop.hive.ql.exec.TaskFactory;
 import org.apache.hadoop.hive.ql.ddl.DDLSemanticAnalyzerFactory.DDLType;
@@ -115,6 +116,7 @@ public class CreateDatabaseAnalyzer extends BaseSemanticAnalyzer {
       database.setRemote_dbname(remoteDbName);
     }
     database.setOwnerName(SessionState.getUserFromAuthenticator());
+    database.setOwnerType(PrincipalType.USER);
     rootTasks.add(TaskFactory.get(new DDLWork(getInputs(), getOutputs(), desc)));
     outputs.add(new WriteEntity(database, WriteEntity.WriteType.DDL_NO_LOCK));
   }
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseAnalyzer.java
@@ -33,6 +33,7 @@ import org.apache.hadoop.hive.ql.parse.ASTNode;
 import org.apache.hadoop.hive.ql.parse.BaseSemanticAnalyzer;
 import org.apache.hadoop.hive.ql.parse.HiveParser;
 import org.apache.hadoop.hive.ql.parse.SemanticException;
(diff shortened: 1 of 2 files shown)


ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseAnalyzer.java
```diff
@@ -103,6 +104,7 @@ public void analyzeInternal(ASTNode root) throws SemanticException {
       if (managedLocationUri != null) {
         database.setManagedLocationUri(managedLocationUri);
       }
+      database.setOwnerName(SessionState.getUserFromAuthenticator());
```
InvisibleProgrammer, line 107: `CreateDatabaseOperation` sets the ownername both for native and remote db type. I wonder, should we follow the same pattern?

Aggarwal-Raghav: Thanks for the review @InvisibleProgrammer, actually I was under some other assumption.

I will address this.

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseAnalyzer.java
```diff
@@ -113,6 +114,7 @@ public void analyzeInternal(ASTNode root) throws SemanticException {
       database.setType(DatabaseType.REMOTE);
       database.setRemote_dbname(remoteDbName);
     }
+    database.setOwnerName(SessionState.getUserFromAuthenticator());
```
okumin, line 117: Would it be better or more consistent if we set the owner type? I could be wrong
https://github.com/apache/hive/blob/5a9e289c8646649136b253841847e578326e41be/ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseOperation.java#L50

Aggarwal-Raghav: Thanks @okumin, for the review and a vaild question but I think there is no need to set owner type. My reasoning is, why set additional info which is not getting utilized by Ranger and post authorization the code flow will enter the CreateDatabaseOperation and there ownerType and other attributes of class will be set. We only need owner name info as it is used by ranger, it doesn't require owner type. In future if Ranger requires ownerType info then we can address it separately.

okumin: Thanks. Your statement makes sense. Let me make the last challenge.
We currently encounter the problem of Apache Ranger, and this minimal fix surely resolves the problem. I agree that this fact supports your opinion.
I'd like to add two perspectives here. First, I think adding owner information is theoretically not only for Ranger, i.e., another authorization system or another plugin such as `HiveSemanticAnalyzerHook` can use the information. Second, the context would be more explicit and less confusing with a principal type as users don't have to worry that the name is linked to a role or group. Because of the two reasons, I am guessing aligning this with `CreateDatabaseOperation` is not a bad investment.
At least, this change will likely work and resolve the problem. So, I don't have any strong objections to have this merged. Just I want to make a decision once I get your response as you are more knowledgeable with this part 👍

Aggarwal-Raghav: Sure @okumin, the first point makes sense if other authorizer or in future ranger starts using it. I will update the PR by adding `database.setOwnerType(PrincipalType.USER);`

My understanding is when we do `kinit ntid/serviceId@Realm `, even serviceId is considered as user. So setting `PrincipalType.USER` from hive should be ok!

okumin (approved):
+1. I'm waiting for CI and 24 hours

InvisibleProgrammer (approved):

PR #5618, opened 2025-01-21
HIVE-27831: Set hive.cbo.fallback.strategy to NEVER by default

ramesh0201:
Remove TEST strategy since we want to uncover and fix as many CBO problems as possible.
Remove explicit set of hive.cbo.fallback.strategy to NEVER which is now the default.

A step towards making CBO work for all queries and the default behavior

Yes, the default value for the config hive.cbo.fallback.strategy was changed to NEVER. In case of CBO error, we will not fallback to non CBO path.

No

Running all the test files modified


zabetak (changes requested):
Changes look good but some test files need to be updated.

zabetak:
I pushed https://github.com/apache/hive/pull/5618/commits/8752bd31a322e347f6be07a50a97a2cfb96ea9c2 to update the remaining .q.out files.

zabetak (approved):
LGTM, pending tests!

amansinha100 (commented):
In the PR description:
"Does this PR introduce any user-facing change?"
"No"
This is not correct since the default value is changed.  Could you update the description for completeness sake ?

ramesh0201:
I have updated the PR description to reflect the config default value change effect for the end user.

PR #5623, opened 2025-01-27
HIVE-28725: Sorting is performed when order by position is disabled when CBO is enabled

kasakrisz:
* Do not create a `HiveSortLimit` operator when `hive.orderby.position.alias` is disabled and the `Order by` clause of the query has integer constants as keys. Also ignore the constants as keys when there are both column references and constant keys.
* Do not create `HiveSortExchange` operator when both field collaction and distribution keys are all constants and `hive.orderby.position.alias` is disabled. Also ignore the constants as keys when there are both column references and constant keys.

To synchronize the Order/Sort/Distribute/Cluster by behaviors when `hive.orderby.position.alias` is disabled and CBO is enabled/disabled .

Yes, the results of queries with Order/Sort/Cluster by clauses may have different order.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=cbo_distribute_sort_cluster_by_pos.q,cbo_rp_order_distribute_sort_cluster_by_cnst.q -pl itests/qtest -Pitests
```


okumin (approved):
LGTM

PR #5624, opened 2025-01-28
HIVE-27370: support 4 bytes characters

ryukobayashi:

If a `SUBSTR` UDF has a 4-byte characters in its parameter, the behavior is different between vectorized and non-vectorized. The vectorized version handles 4-byte characters properly, but the non-vectorized version does not, so similar logic is needed.
And these fixes use vectorized logic:
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/StringSubstrColStartLen.java#L89-L130
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/StringSubstrColStart.java#L78-L109

Vectorized and non-vectorized have different results.

No.

Added pattern tests to itest for these to work correctly.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
@@ -96,53 +96,131 @@ public class UDFSubstr extends UDF implements StatEstimatorProvider {
       return r;
     }
 
-    String s = t.toString();
-    int[] index = makeIndex(pos, len, s.length());
-    if (index == null) {
+    byte[] utf8String = t.toString().getBytes();
+    populateSubstrOffsets(utf8String, pos, len);
+    if (index[0] == -1) {
       return r;
     }
 
-    r.set(s.substring(index[0], index[1]));
+    r.set(new String(utf8String, index[0], index[1]));
     return r;
   }
 
-  private int[] makeIndex(int pos, int len, int inputLen) {
-    if ((Math.abs(pos) > inputLen)) {
+  private Text evaluateInternal(Text t, int pos) {
+    r.clear();
+
+    byte[] utf8String = t.toString().getBytes();
+    int offset = getSubstrStartOffset(utf8String, pos);
+    if (offset == -1) {
+      return r;
+    }
+
+    r.set(new String(utf8String, offset, utf8String.length - offset));
+    return r;
+  }
+
+  private void populateSubstrOffsets(byte[] utf8String, int start, int len) {
+    int curIdx = -1;
+    index[0] = -1;
+    index[1] = -1;
+    int end = utf8String.length;
+
+    if (start > 0) {
+      start = start - 1;
+    } else if (start < 0) {
+      int length = 0;
+      for (int i = 0; i != end; ++i) {
+        if ((utf8String[i] & 0xc0) != 0x80) {
+          ++length;
+        }
+      }
+
+      if (-start > length) {
+        return;
+      }
+
+      start = length + start;
+    }
+
+    if (len == 0) {
+      return;
+    } else if (len > end) {
+      len = end;
+    }
+
+    int endIdx = start + len - 1;
+    for (int i = 0; i != end; ++i) {
+      if ((utf8String[i] & 0xc0) != 0x80) {
+        ++curIdx;
+        if (curIdx == start) {
+          index[0] = i;
+        } else if (curIdx - 1 == endIdx) {
+          index[1] = i - index[0];
+        }
+      }
+    }
+
+    if (index[1] == -1) {
+      index[1] = end - index[0];
+    }
+  }
+
+  private int getSubstrStartOffset(byte[] utf8String, int start) {
+    int end = utf8String.length;
+
+    if (start >= 1) {
+      start = start - 1;
+    }
+    if (start < 0) {
+      int length = 0;
+      for (int i = 0; i != end; ++i) {
+        if ((utf8String[i] & 0xc0) != 0x80) {
+          ++length;
+        }
+      }
+
+      if (-start > length) {
+        return -1;
+      }
+
+      start = length + start;
+    }
+
+    int curIdx = -1;
+    for (int i = 0; i != end; ++i) {
+      if ((utf8String[i] & 0xc0) != 0x80) {
+        ++curIdx;
+        if (curIdx == start) {
+          return i;
+        }
+      }
+    }
+
+    return -1;
+  }
+
+  public Text evaluate(Text s, IntWritable pos) {
+    if ((s == null) || (pos == null)) {
       return null;
     }
 
-    int start, end;
-
-    if (pos > 0) {
-      start = pos - 1;
-    } else if (pos < 0) {
-      start = inputLen + pos;
-    } else {
-      start = 0;
-    }
-
-    if ((inputLen - start) < len) {
-      end = inputLen;
-    } else {
-      end = start + len;
-    }
-    index[0] = start;
-    index[1] = end;
-    return index;
-  }
-
-  private final IntWritable maxValue = new IntWritable(Integer.MAX_VALUE);
-
-  // Even though we are using longs, substr can only deal with ints, so we use
-  // the maximum int value as the maxValue
-  private final LongWritable maxLongValue = new LongWritable(Integer.MAX_VALUE);
-
-  public Text evaluate(Text s, IntWritable pos) {
-    return evaluate(s, pos, maxValue);
+    return evaluateInternal(s, pos.get());
   }
 
   public Text evaluate(Text s, LongWritable pos) {
-    return evaluate(s, pos, maxLongValue);
+    if ((s == null) || (pos == null)) {
+      return null;
+    }
+
+    long longPos = pos.get();
+    // If an unsupported value is seen, we don't want to return a string
+    // that doesn't match what the user expects, so we return NULL (still
+    // unexpected, of course, but probably better than a bad string).
+    if (longPos > Integer.MAX_VALUE || longPos < Integer.MIN_VALUE) {
+      return null;
+    }
+
+    return evaluateInternal(s, (int) pos.get());
   }
 
   public BytesWritable evaluate(BytesWritable bw, LongWritable pos, LongWritable len) {
@@ -172,30 +250,59 @@ public class UDFSubstr extends UDF implements StatEstimatorProvider {
   }
 
   private BytesWritable evaluateInternal(BytesWritable bw, int pos, int len) {
-
     if (len <= 0) {
       return new BytesWritable();
     }
 
-    int[] index = makeIndex(pos, len, bw.getLength());
-    if (index == null) {
+    byte[] b = Arrays.copyOf(bw.getBytes(), bw.getLength());
+    populateSubstrOffsets(b, pos, len);
+    if (index[0] == -1) {
       return new BytesWritable();
     }
 
-    return new BytesWritable(Arrays.copyOfRange(bw.getBytes(), index[0], index[1]));
+    return new BytesWritable(arrayCopy(b, index[0], index[1]));
+  }
+
+  private BytesWritable evaluateInternal(BytesWritable bw, int pos) {
+    byte[] b = Arrays.copyOf(bw.getBytes(), bw.getLength());
+    int offset = getSubstrStartOffset(b, pos);
+    if (offset == -1) {
+      return new BytesWritable();
+    }
+
+    return new BytesWritable(arrayCopy(b, offset, bw.getLength() - offset));
   }
 
   public BytesWritable evaluate(BytesWritable bw, IntWritable pos){
-    return evaluate(bw, pos, maxValue);
+    if ((bw == null) || (pos == null)) {
+      return null;
+    }
+    return evaluateInternal(bw, pos.get());
   }
 
   public BytesWritable evaluate(BytesWritable bw, LongWritable pos){
-    return evaluate(bw, pos, maxLongValue);
+    if ((bw == null) || (pos == null)) {
+      return null;
+    }
+
+    return evaluateInternal(bw, (int) pos.get());
   }
 
   @Override
   public StatEstimator getStatEstimator() {
     return new SubStrStatEstimator();
+  }
+
+  private byte[] arrayCopy(byte[] src, int pos, int len) {
+    byte[] b = new byte[len];
+
+    int copyIdx = 0;
+    for (int srcIdx = pos; copyIdx < len; srcIdx++) {
+      b[copyIdx] = src[srcIdx];
+      copyIdx++;
+    }
+
+    return b;
   }
 
   private static class SubStrStatEstimator implements StatEstimator {
--- a/ql/src/test/queries/clientpositive/udf_substr.q
+++ b/ql/src/test/queries/clientpositive/udf_substr.q
@@ -89,3 +89,11 @@ FROM src tablesample (1 rows);
 SELECT
   substr('ABC', cast(2147483649 as bigint))
 FROM src tablesample (1 rows);
+
+--test 4-byte charactor
+set hive.vectorized.execution.enabled=false;
+SELECT
+  substr('あa🤎いiうu', 1, 3) as b1,
+  substr('あa🤎いiうu', 3) as b2,
+  substr('あa🤎いiうu', -5) as b3
+FROM src tablesample (1 rows);
--- a/ql/src/test/results/clientpositive/llap/udf_substr.q.out
+++ b/ql/src/test/results/clientpositive/llap/udf_substr.q.out
@@ -240,3 +240,20 @@ POSTHOOK: type: QUERY
 POSTHOOK: Input: default@src
 #### A masked pattern was here ####
 NULL
+PREHOOK: query: SELECT
+  substr('あa🤎いiうu', 1, 3) as b1,
+  substr('あa🤎いiうu', 3) as b2,
+  substr('あa🤎いiうu', -5) as b3
+FROM src tablesample (1 rows)
+PREHOOK: type: QUERY
+PREHOOK: Input: default@src
+#### A masked pattern was here ####
+POSTHOOK: query: SELECT
+  substr('あa🤎いiうu', 1, 3) as b1,
+  substr('あa🤎いiうu', 3) as b2,
+  substr('あa🤎いiうu', -5) as b3
+FROM src tablesample (1 rows)
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@src
+#### A masked pattern was here ####
+あa?	🤎いiうu	?いiうu


deniskuzZ:
@difin, please take a look when you have time

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -96,53 +96,131 @@ private Text evaluateInternal(Text t, int pos, int len) {
...
+    if (offset == -1) {
+      return r;
     }
 
-    int start, end;
+    r.set(new String(utf8String, offset, utf8String.length - offset));
+    return r;
+  }
+
+  private void populateSubstrOffsets(byte[] utf8String, int start, int len) {
```
deniskuzZ, line 122: try to reuse methods instead of copy&paste

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -96,53 +96,131 @@ private Text evaluateInternal(Text t, int pos, int len) {
...
+    if (index[1] == -1) {
+      index[1] = end - index[0];
     }
-    index[0] = start;
-    index[1] = end;
-    return index;
   }
 
-  private final IntWritable maxValue = new IntWritable(Integer.MAX_VALUE);
+  private int getSubstrStartOffset(byte[] utf8String, int start) {
```
deniskuzZ, line 168: same here

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: I fixed.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: I fixed.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -172,32 +170,69 @@ public BytesWritable evaluate(BytesWritable bw, IntWritable pos, IntWritable len
...
+    int copyIdx = 0;
+    for (int srcIdx = pos; copyIdx < len; srcIdx++) {
+      b[copyIdx] = src[srcIdx];
+      copyIdx++;
+    }
+
+    return b;
+  }
+
+  private int craetePos(int pos) {
```
deniskuzZ: typo: `createPos()`, maybe better rename to `adjustStartPos()`

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -172,32 +170,69 @@ public BytesWritable evaluate(BytesWritable bw, IntWritable pos, IntWritable len
...
+
+    return evaluateInternal(bw, (int) pos.get());
   }
 
   @Override
   public StatEstimator getStatEstimator() {
     return new SubStrStatEstimator();
   }
 
+  private byte[] arrayCopy(byte[] src, int pos, int len) {
```
deniskuzZ: can you use `Arrays.copyOfRange(T[] original, int from, int to)` ?

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -172,32 +170,69 @@ public BytesWritable evaluate(BytesWritable bw, IntWritable pos, IntWritable len
...
 
   private BytesWritable evaluateInternal(BytesWritable bw, int pos, int len) {
-
     if (len <= 0) {
       return new BytesWritable();
     }
 
-    int[] index = makeIndex(pos, len, bw.getLength());
-    if (index == null) {
+    byte[] b = Arrays.copyOf(bw.getBytes(), bw.getLength());
```
deniskuzZ: why do we need to create copy of bw?

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -172,32 +170,69 @@ public BytesWritable evaluate(BytesWritable bw, IntWritable pos, IntWritable len
...
+    if (index[0] == -1) {
       return new BytesWritable();
     }
 
-    return new BytesWritable(Arrays.copyOfRange(bw.getBytes(), index[0], index[1]));
+    return new BytesWritable(arrayCopy(b, index[0], index[1]));
+  }
+
+  private BytesWritable evaluateInternal(BytesWritable bw, int pos) {
+    byte[] b = Arrays.copyOf(bw.getBytes(), bw.getLength());
```
deniskuzZ: same, why do we need copy?

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -96,53 +96,51 @@ private Text evaluateInternal(Text t, int pos, int len) {
...
-    index[0] = start;
-    index[1] = end;
-    return index;
+    r.set(new String(utf8String, offset, utf8String.length - offset));
+    return r;
   }
 
-  private final IntWritable maxValue = new IntWritable(Integer.MAX_VALUE);
-
-  // Even though we are using longs, substr can only deal with ints, so we use
```
deniskuzZ, line 136: please keep this comment where you check the range

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: OK, I will rename `adjustStartPos()`

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: No, if `Arrays.copyOfRange` is used, `substr(ABC, 2, 1)` and `substr(ABC, -4, 1)` parameters cannot be supported.
https://github.com/apache/hive/blob/master/ql/src/test/queries/clientpositive/udf_substr.q

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: There was no need to copy. I fixed it.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: There was no need to copy. I fixed it.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: Sure, I will keep this comment.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
deniskuzZ: @ryukobayashi, in Java, array indices cannot be negative.
````
return Arrays.copyOfRange(src, pos, pos + len);
````

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java]: `udf_subst.q` passed with that change. please see the diff
````
Subject: [PATCH] refactor
---
Index: ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java b/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java	(revision 8d1d81e10fc4ba400e6abbf2fcfc72699ad32ebb)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java	(date 1750879762534)
@@ -18,7 +18,6 @@
 
 package org.apache.hadoop.hive.ql.udf;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 
@@ -36,6 +35,8 @@
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Text;
 
+import static java.util.Arrays.copyOfRange;
+
 /**
  * UDFSubstr.
  *
@@ -181,7 +182,7 @@
       return new BytesWritable();
     }
 
-    return new BytesWritable(arrayCopy(bw.getBytes(), index[0], index[1]));
+    return new BytesWritable(copyOfRange(bw.getBytes(), index[0], index[0] + index[1]));
   }
 
   private BytesWritable evaluateInternal(BytesWritable bw, int pos) {
@@ -192,7 +193,7 @@
       return new BytesWritable();
     }
 
-    return new BytesWritable(arrayCopy(bw.getBytes(), offset, bw.getLength() - offset));
+    return new BytesWritable(copyOfRange(bw.getBytes(), offset, bw.getLength()));
   }
 
   public BytesWritable evaluate(BytesWritable bw, IntWritable pos){
@@ -215,17 +216,6 @@
     return new SubStrStatEstimator();
   }
 
-  private byte[] arrayCopy(byte[] src, int pos, int len) {
-    byte[] b = new byte[len];
-
-    int copyIdx = 0;
-    for (int srcIdx = pos; copyIdx < len; srcIdx++) {
-      b[copyIdx] = src[srcIdx];
-      copyIdx++;
-    }
-    return b;
-  }
-
   private int adjustStartPos(int pos) {
     if (pos <= 0) {
       return pos;
@@ -244,7 +234,6 @@
       // 99 rows with 0 length
       // orig avg is 10
       // new avg is 5 (if substr(5)) ; but in reality it will stay ~10
-      Optional<Double> start = getRangeWidth(csList.get(1).getRange());
       Range startRange = csList.get(1).getRange();
       if (startRange != null && startRange.minValue != null) {
         double newAvgColLen = cs.getAvgColLen() - startRange.minValue.doubleValue();
@@ -255,7 +244,7 @@
       if (csList.size() > 2) {
         Range lengthRange = csList.get(2).getRange();
         if (lengthRange != null && lengthRange.maxValue != null) {
-          Double w = lengthRange.maxValue.doubleValue();
+          double w = lengthRange.maxValue.doubleValue();
           if (cs.getAvgColLen() > w) {
             cs.setAvgColLen(w);
           }
@@ -263,15 +252,5 @@
       }
       return Optional.of(cs);
     }
-
-    private Optional<Double> getRangeWidth(Range range) {
-      if (range != null) {
-        if (range.minValue != null && range.maxValue != null) {
-          return Optional.of(range.maxValue.doubleValue() - range.minValue.doubleValue());
-        }
-      }
-      return Optional.empty();
-    }
-
   }
 }
````

Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java]: OK, I will fix that code.

**Comment by deniskuzZ** [comment]
Hi @ryukobayashi, apologies for the slow progress on this PR. However, after noticing some obvious typos, my confidence in the changes has dropped. Overall, the code looks OK, but It needs a thorough review, which I unfortunately don’t have time for at the moment. Since @okumin is your colleague, perhaps he could help move this forward.

**Comment by okumin** [comment]
Sure. I was not confident that I should review coworker's patch to avoid bias. I will check this anyway.

File: ql/src/test/queries/clientpositive/udf_substr.q [hunk 8]
```diff
@@ -89,3 +89,11 @@ FROM src tablesample (1 rows);
...
   substr('ABC', cast(2147483649 as bigint))
 FROM src tablesample (1 rows);
+
+--test 4-byte charactor
+set hive.vectorized.execution.enabled=false;
+SELECT
+  substr('あa🤎いiうu', 1, 3) as b1,
+  substr('あa🤎いiうu', 3) as b2,
+  substr('あa🤎いiうu', -5) as b3
+FROM src tablesample (1 rows);
```
okumin, line 99: I verified the master branch can't pass this test case

```
+POSTHOOK: query: SELECT
+  substr('あa🤎いiうu', 1, 3) as b1,
+  substr('あa🤎いiうu', 3) as b2,
+  substr('あa🤎いiうu', -5) as b3
+FROM src tablesample (1 rows)
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@src
+あa?   🤎いiうu        ?いiうu
```

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -96,53 +97,51 @@ private Text evaluateInternal(Text t, int pos, int len) {
       return r;
     }
 
-    String s = t.toString();
-    int[] index = makeIndex(pos, len, s.length());
-    if (index == null) {
+    byte[] utf8String = t.toString().getBytes();
```
okumin: Why not use `Text#getBytes` or `Text#copyBytes`? Probably, `getBytes` is preferable because we don't mutate it. Note that we have to use `Text#getLength` instead of the size of the byte array.
https://github.com/apache/hadoop/blob/rel/release-3.4.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/Text.java#L116-L132

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -96,53 +97,51 @@ private Text evaluateInternal(Text t, int pos, int len) {
...
+    StringSubstrColStartLen.populateSubstrOffsets(utf8String, 0, utf8String.length, adjustStartPos(pos), len, index);
+    if (index[0] == -1) {
       return r;
     }
 
-    r.set(s.substring(index[0], index[1]));
+    r.set(new String(utf8String, index[0], index[1]));
```
okumin: I understand the previous `index[1]` holds the end index, but the current one has the length.
Let's use `new String(utf8, index[0], index[1], StandardCharsets.UTF-8)` because the current one relies on Java's configuration.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -96,53 +97,51 @@ private Text evaluateInternal(Text t, int pos, int len) {
...
+  private Text evaluateInternal(Text t, int pos) {
+    r.clear();
 
-    if (pos > 0) {
-      start = pos - 1;
-    } else if (pos < 0) {
-      start = inputLen + pos;
-    } else {
-      start = 0;
+    byte[] utf8String = t.toString().getBytes();
```
okumin: Ditto

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -96,53 +97,51 @@ private Text evaluateInternal(Text t, int pos, int len) {
...
 
-    if ((inputLen - start) < len) {
-      end = inputLen;
-    } else {
-      end = start + len;
-    }
-    index[0] = start;
-    index[1] = end;
-    return index;
+    r.set(new String(utf8String, offset, utf8String.length - offset));
```
okumin: Ditto

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -172,32 +171,59 @@ public BytesWritable evaluate(BytesWritable bw, IntWritable pos, IntWritable len
...
-
     if (len <= 0) {
       return new BytesWritable();
     }
 
-    int[] index = makeIndex(pos, len, bw.getLength());
-    if (index == null) {
+    // Even though we are using longs, substr can only deal with ints, so we use
+    // the maximum int value as the maxValue
+    StringSubstrColStartLen.populateSubstrOffsets(bw.getBytes(), 0, bw.getLength(), adjustStartPos(pos), len, index);
```
okumin: I am guessing `UDFSubstr#evaluate(Text)` is encoding-aware but ``UDFSubstr#evaluate(ByteWritable)` is not.
`StringSubstrColStartLen.populateSubstrOffsets` assumes the byte array as UTF-8 and `substrStart` and `substrLength` as the position of UTF-8. However, `UDFSubstr#evaluate(ByteWritable)` seems to handle the byte array as a pure byte array without assuming any encoding. So, we don't need to update the behavior, I think. No vectorized implementation exists for the binary type.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: Ah, OK. I will fix it.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: I will fix it.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: fixed.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: fixed

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: Ah, OK. I will revert to the original.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -96,13 +98,12 @@ private Text evaluateInternal(Text t, int pos, int len) {
...
-    String s = t.toString();
-    int[] index = makeIndex(pos, len, s.length());
-    if (index == null) {
+    StringSubstrColStartLen.populateSubstrOffsets(t.getBytes(), 0, t.getLength(), adjustStartPos(pos), len, index);
+    if (index[0] == -1) {
       return r;
     }
 
-    r.set(s.substring(index[0], index[1]));
+    r.set(new String(t.getBytes(), index[0], index[1], StandardCharsets.UTF_8));
```
okumin: Sorry. I found a better method to copy the utf8 binary. Could you use it? I slightly remember an extra UTF8 codec can be a bottleneck.
https://github.com/apache/hadoop/blob/rel/release-3.4.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/Text.java#L265-L277

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -133,16 +134,44 @@ private int[] makeIndex(int pos, int len, int inputLen) {
...
 
+  private Text evaluateInternal(Text t, int pos) {
+    r.clear();
+
+    int offset = StringSubstrColStart.getSubstrStartOffset(t.getBytes(), 0, t.getLength(), adjustStartPos(pos));
+    if (offset == -1) {
+      return r;
+    }
+
+    r.set(new String(t.getBytes(), offset, t.getLength() - offset, StandardCharsets.UTF_8));
```
okumin: ditto

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: OK, I fixed.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
ryukobayashi: fixed

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -36,6 +36,8 @@
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Text;
 
+import java.nio.charset.StandardCharsets;
```
okumin, line 39:
```suggestion
```

unused import

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFSubstr.java
```diff
@@ -36,6 +36,8 @@
...
+
```
okumin:
```suggestion
```

Unused import

okumin:
Looks good to me, but we're running additional in-house integration tests just in case.

okumin:
Update: I want to run some additional tests, but I haven't taken the time. I will update the status in 2 weeks

PR #5626, opened 2025-01-30
HIVE-28729: Apply nulls order setting in Reduce Sink operator of join branches

kasakrisz:
Use the default null order configured via `hive.default.nulls.last` in the Calcite rule `HiveInsertExchange4JoinRule.java` when creating sort keys. This rule is used only when `hive.cbo.returnpath.hiveop` is enabled.

The setting `hive.default.nulls.last` doesn't have affect to the `HiveSortExchange` operators created by this rule and the RS operators built from them.

Explain output of query plans has join operation may change.

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestMiniLlapLocalCliDriver -Dqfile=cbo_rp_null_order.q -pl itests/qtest -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
@@ -54,19 +54,6 @@ import com.google.common.collect.Sets;
  */
 public class HiveInsertExchange4JoinRule extends RelOptRule {
 
-  /** Rule that creates Exchange operators under a MultiJoin operator. */
-  public static HiveInsertExchange4JoinRule exchangeBelowMultiJoin(
-          RelFieldCollation.NullDirection defaultAscNullDirection) {
-    return new HiveInsertExchange4JoinRule(HiveMultiJoin.class, defaultAscNullDirection);
-  }
-
-
-  /** Rule that creates Exchange operators under a Join operator. */
-  public static HiveInsertExchange4JoinRule exchangeBelowJoin(
-          RelFieldCollation.NullDirection defaultAscNullDirection) {
-    return new HiveInsertExchange4JoinRule(Join.class, defaultAscNullDirection);
-  }
-
   private final RelFieldCollation.NullDirection defaultAscNullDirection;
 
   public HiveInsertExchange4JoinRule(
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
@@ -30,7 +30,6 @@ import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.Exchange;
 import org.apache.calcite.rel.core.Join;
 import org.apache.calcite.rex.RexNode;
-import org.apache.hadoop.hive.ql.util.NullOrdering;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.hive.ql.optimizer.calcite.CalciteSemanticException;
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
@@ -30,6 +30,7 @@ import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.Exchange;
 import org.apache.calcite.rel.core.Join;
 import org.apache.calcite.rex.RexNode;
+import org.apache.hadoop.hive.ql.util.NullOrdering;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.hive.ql.optimizer.calcite.CalciteSemanticException;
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
```diff
@@ -56,20 +57,28 @@
  */
 public class HiveInsertExchange4JoinRule extends RelOptRule {
 
-  protected static transient final Logger LOG = LoggerFactory
-      .getLogger(HiveInsertExchange4JoinRule.class);
+  protected static final Logger LOG = LoggerFactory.getLogger(HiveInsertExchange4JoinRule.class);
```
zabetak, line 60: The logger does not seem to be used so I guess we can drop it entirely along with the relevant imports.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
```diff
@@ -56,20 +57,28 @@
...
 
   /** Rule that creates Exchange operators under a Join operator. */
-  public static final HiveInsertExchange4JoinRule EXCHANGE_BELOW_JOIN =
-      new HiveInsertExchange4JoinRule(Join.class);
+  public static HiveInsertExchange4JoinRule exchangeBelowJoin(
+          RelFieldCollation.NullDirection defaultAscNullDirection) {
+    return new HiveInsertExchange4JoinRule(Join.class, defaultAscNullDirection);
+  }
+
+  private final RelFieldCollation.NullDirection defaultAscNullDirection;
```
zabetak, line 75: I don't see a big advantage in introducing the static factories. Since the constructor is public we could just keep that one. If we want to make the rule creation shorter then we could have a static factory with a shorter name (e.g., `of(clazz,nullDirection)`)

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
```diff
@@ -118,7 +127,7 @@ public void onMatch(RelOptRuleCall call) {
         for (int pos : joinLeafPredInfo.getProjsJoinKeysInChildSchema(i)) {
           if (!joinKeyPositions.contains(pos)) {
             joinKeyPositions.add(pos);
-            collationListBuilder.add(new RelFieldCollation(pos));
+            collationListBuilder.add(new RelFieldCollation(pos, RelFieldCollation.Direction.ASCENDING, defaultAscNullDirection));
```
zabetak, line 130: Do we always want the direction to be `ASCENDING`?

ql/src/test/queries/clientpositive/cbo_rp_null_order.q
```diff
@@ -0,0 +1,7 @@
+SET hive.cbo.returnpath.hiveop=true;
+SET hive.default.nulls.last=false;
```
zabetak, line 2: For completeness maybe we want to add an extra test with `nulls.last=true` to ensure that changes take effect.

ql/src/test/queries/clientpositive/cbo_rp_null_order.q
```diff
@@ -0,0 +1,7 @@
...
+
+CREATE TABLE t1(key int, value string);
+
+EXPLAIN CBO SELECT * FROM t1 a INNER JOIN t1 b on a.key = b.key;
```
zabetak, line 6: I think that for INNER JOIN the direction of nulls is not gonna affect correctness. Does the null direction affect correctness of outer joins?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
kasakrisz: Removed

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
kasakrisz: Removed the factory methods. Using constructor in CalcitePlanner

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInsertExchange4JoinRule.java
kasakrisz: The reason why we need these sort exchange operators is to get the data coming from join branches to be sorted because we want to perform a sort merge join.
I think Ascending is the default but it could also work with descending. I think it has an affect only on queries with limit but without order/sort by. However it doesn't matter which n rows do we pick in those cases.

ql/src/test/queries/clientpositive/cbo_rp_null_order.q
kasakrisz: We have similar tests but not with this query so I added.

ql/src/test/queries/clientpositive/cbo_rp_null_order.q
kasakrisz: I think it has an affect only on queries with limit but without order/sort by. However it doesn't matter which n rows do we pick.
We can see the null sort order in the RS properties.
```
                      Reduce Output Operator
                        key expressions: key (type: string)
                        null sort order: z
                        sort order: +
```
The join is only necessary to get the rule `HiveInsertExchange4JoinRule` to add the `HiveSortExchange` operators.

zabetak (approved):

PR #5627, opened 2025-01-31
HIVE-28732: Sorted dynamic partition optimization does not apply hive.default.nulls.last

kasakrisz:
* Currently the nulls order of RS keys in Sorted dynamic partition optimization are hardcoded to ASC NULLS FIRST and DESC NULLS LAST. This patch changes the logic to respect the value of setting `hive.default.nulls.last`.
* Fix The mapping between Iceberg null order and Hive null order: the current mapping is mixed: Iceberg `NullOrder.NULLS_FIRST` -> Hive `NullOrdering.NULLS_LAST` and vice versa.

To unify nulls ordering.

No but explain output of affected queries may change.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=dynpart_sort_optimization.q -pl itests/qtest -Pitests
```


ql/src/test/results/clientpositive/llap/materialized_view_partitioned.q.out
```diff
@@ -1154,11 +1154,11 @@ POSTHOOK: Input: default@partition_mv_3
 POSTHOOK: Input: default@partition_mv_3@key=238
 #### A masked pattern was here ####
 val_238_n	238
-val_238_n	238
 val_238	238
 val_238	238
 val_238	238
 val_238	238
+val_238_n	238
```
okumin, line 1161: Although this doesn't violate the semantics, I didn't understand why this diff showed up

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -857,7 +858,12 @@ public DynamicPartitionCtx createDPContext(
           if (sortField.sourceId() == field.fieldId()) {
             customSortPositions.add(pos);
             customSortOrder.add(sortField.direction() == SortDirection.ASC ? 1 : 0);
-            customSortNullOrder.add(sortField.nullOrder() == NullOrder.NULLS_FIRST ? 0 : 1);
+
+            NullOrdering nullOrdering = NullOrdering.NULLS_LAST;
+            if (sortField.nullOrder() == NullOrder.NULLS_FIRST) {
+              nullOrdering = NullOrdering.NULLS_FIRST;
+            }
+            customSortNullOrder.add(nullOrdering.getCode());
```
okumin, line 866: This seems to be OK as a conclusion, but I am leaving a note that came to my mind.

We added a feature to configure the sort order of Iceberg tables. I wondered if we can keep the semantics of `WRITE LOCALLY ORDERED BY a`. 
https://github.com/apache/hive/pull/5541/files#r1894857758

The null ordering of sort order specs is required. So, our DDL specifies either NULLS_FIRST or NULLS_LAST, and then `createDPContext` just follows the definition. In other words, `sortField.nullOrder()` is non-null and this change never takes effect in the real world.
https://github.com/apache/iceberg/blob/8839c9bf1f1d8c9b718f9766302ff8a2018e515f/core/src/main/java/org/apache/iceberg/SortOrderParser.java#L159-L175

okumin (approved):
LGTM

ql/src/test/results/clientpositive/llap/materialized_view_partitioned.q.out
kasakrisz: Since the query
```
SELECT * FROM partition_mv_3 where key = 238
```
does not specify ordering I think it depends on the execution engine internals.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
kasakrisz: There were two failing test without this change.
```
TestHiveIcebergInserts.testSortedInsert
TestHiveIcebergInserts.testSortedAndTransformedInsertIntoPartitionedTable
```
When I analyzed the cause of the failure of these tests I bumped into this bug: 
```
sortField.nullOrder() == NullOrder.NULLS_FIRST ? 0 : 1
```
maps `NULLS_FIRST` to `0` which actually means nulls last in Hive
https://github.com/apache/hive/blob/fa6813f725542bae127f9bec0921ebd01c946947/ql/src/java/org/apache/hadoop/hive/ql/util/NullOrdering.java#L30-L31

> We added a feature to configure the sort order of Iceberg tables. I wondered if we can keep the semantics of WRITE LOCALLY ORDERED BY a

I found that null sort order is set here when it was not specified explicitly at create table:
https://github.com/apache/hive/blob/fa6813f725542bae127f9bec0921ebd01c946947/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java#L292
and the defaults are defined at the parser level
https://github.com/apache/hive/blob/fa6813f725542bae127f9bec0921ebd01c946947/parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g#L2205-L2206

This is not affected by the setting `hive.default.nulls.last`.

>  sortField.nullOrder() is non-null and this change never takes effect in the real world.

I don't understand this part. Could you please elaborate.

okumin: I reviewed this PR, assuming we don't change the semantics of `customSortNullOrder.add(sortField.nullOrder() == NullOrder.NULLS_FIRST ? 0 : 1);`. Now, I understand 0 and 1 will be correctly inverted. Let me check...

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SortedDynPartitionOptimizer.java
```diff
@@ -635,34 +638,18 @@ public ReduceSinkOperator getReduceSinkOp(List<Integer> partitionPositions, List
...
         }
-        newSortNullOrder.addAll(customSortNullOrder);
-      }
-
-      String nullOrderStr = "";
-      for (Integer i : newSortNullOrder) {
-        if (i == 0) {
-          nullOrderStr += "a";
-        } else {
-          nullOrderStr += "z";
```
okumin, line 665: Just my note. `HiveIcebergStorageHandler` mapped NULLS_FIRST to 0 and NULLS_LAST to 1. It should have been inverted for the consistency with `NullOrdering#getCode`. But the inconsistency didn't introduce real issues because this part didn't also follow `NullOrdering#getCode`.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SortedDynPartitionOptimizer.java
```diff
@@ -635,34 +638,18 @@ public ReduceSinkOperator getReduceSinkOp(List<Integer> partitionPositions, List
...
-        if (sortNullOrder.get(0) == 0) {
-          nullOrder = 0;
-        } else {
-          nullOrder = 1;
-        }
-      }
-
-      for (Integer ignored : keyColsPosInVal) {
-        newSortNullOrder.add(nullOrder);
+        nullOrder = NullOrdering.fromCode(sortNullOrder.get(0)).getSign();
```
okumin, line 643: I wondered why we pick up only the first element of `sortOrder` and `sortNullOrder`. This question doesn't directly relate to the problem of this ticket

okumin (approved):
Approved again.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SortedDynPartitionOptimizer.java
kasakrisz: This logic was introduced by the very first implementation of `SortedDynPartitionOptimizer`.
https://issues.apache.org/jira/browse/HIVE-6455
I think the `sortOrder` is not empty only if the target table has ordering specified. Unfortunately the jira doesn't mention anything about such tables.

PR #5634, opened 2025-02-05
HIVE-28739: support restricting users to create deferred view

eubnara:

I suggest a new configuration "metastore.users.restricted_to_deferred_view" to support a feature that makes some of proxyusers to be forced to create deferred view.

In our environment, we use Impala with HiveMetastore. Since "impala" user is a proxy user, if I create a view through Impala, it will create a view, not a deferred view. (impala doesn't have impersonation support)
In our policy, we want to force users to create deferred view if there is no special reason not to create deferred view in order to follow permissions of source tables.
So I tried to exclude "impala" user from proxy user, there is some bottle neck and the change even causes impala cluster hang. I guess that with HiveMetastoreAuthorizer, impala cannot skip authorization if I exclude "impala" user from proxy user.
Also, on impala side, Ranger authorization is already applied. It is meaningless because the same hive policy applied already on impala side.
Therefore, I gave up to exclude "impala" user from proxy user.

No.

No.

Our internal Hadoop cluster.


eubnara:
.

PR #5635, opened 2025-02-05
[WIP: do not review] HIVE-28712: Fix error wrt maxParts in partition logs

arorasimran0309:
The changes include passing the config value LIMIT_PARTITION_REQUEST instead of maxparts (whose value is set to -1) in get_partitions_by_expr logging expression. This will help in reflecting the correct quantity in case the config is set and being honored with the filter/expr on the partition column

Current partition logs display max partitions = -1, which is the default value. But on changing the value of LIMIT_PARTITION_REQUEST, the set value is not displayed, instead -1 continues to be logged. 

No

No

Existing tests

Case 1: Test passing, before the fix:
image

After the fix:
image

Case 2: Test failing, before the fix:
image
image

After the fix:
image
image


arorasimran0309:
@okumin @ayushtkn could you please help to review this change.

PR #5637, opened 2025-02-07
HIVE-28744: Fix the row count estimation for INT and BIGINT types

barking-code:

Fix `INT` and `BIGINT` type column range statistics estimate values.

When `hive.stats.estimate` and `hive.stats.fetch.column.stats` are enabled and Int or bigint columns don't have col stats about ranges, incorrect ranges are used in the query plan, which returns the wrong num rows.

e.g) If `{BIGINT Type col} > INTEGER.MAX_VALUE` is in query, estimated numRows is always 0.

1. No stats for bigint type column.
image-2025-2-5_0-11-21

2. Write a query using `{BIGINT Type col} > INTEGER.MAX_VALUE`, and EXPLAIN it.
image-2025-2-5_0-9-25

No

No

```
mvn test -Dtest.output.overwrite -Dtest=TestMiniLlapLocalCliDriver -Dqfile=explainanalyze_5.q -pl itests/qtest -Pitests
```


okumin:
@barking-code Thanks for locating the issue and providing the fix. Could you update the output of the failing test cases? Please feel free to ask me if you have any questions.
https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5637/1/pipeline/580/

barking-code:
> @barking-code Thanks for locating the issue and providing the fix. Could you update the output of the failing test cases? Please feel free to ask me if you have any questions. https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5637/1/pipeline/580/

@okumin 
Hello, okumin.
I can access this page. Displaying 404 error, but I think it is an authentication or authorization issue.
Do you know how I can access this page? And do I need to add tests to pass CI tests?

Aggarwal-Raghav:
@barking-code , first login to the CI (top right hand side) before directly opening the URL mentioned in above comments. From CI, you can check which tests are failing and update/fix them accordingly.
Screenshot 2025-02-10 at 9 42 30 PM


If you want to overwrite the q.out files then you can use a command like this:
`mvn test -Dtest=<DriverName> -Dqfile=<test_name>.q -Dtest.output.overwrite`
For example: 
`mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=msck_repair_8.q -Drat.skip -Dtest.output.overwrite
`

okumin:
@barking-code Do you have a chance to update this branch? You can also take a look at a new document on how to run or update test cases.
https://hive.apache.org/development/qtest/

deniskuzZ:
Changes to the INT_TYPE range cause execution problems. Reducer is waiting for the input rows, but nothing is being sent.

deniskuzZ:
I am taking it over since there have been no updates on the PR, the fix is incomplete and breaks the code

deniskuzZ:
@kasakrisz, @okumin please take a look

kasakrisz:
@barking-code, @deniskuzZ 
```

create table t1 (a bigint) stored as orc TBLPROPERTIES ('transactional'='true');

insert into t1 values
(1),
(1704034800001),
(1704034800001),
(1704034800001),
(1704034800001),
(1704034800001),
(1704034800001),
(1704034800001),
(1704034800001);

delete from t1 where a = 1; -- makes the column stats stale

explain
select * from t1 where a >= 1704034800000;

explain
select * from t1 where a >= 10;
```

Partial plan of the first query:
```
select * from t1 where a >= 1704034800000

                TableScan
                  alias: t1
                  filterExpr: (a >= 1704034800000L) (type: boolean)
                  Statistics: Num rows: 8 Data size: 64 Basic stats: COMPLETE Column stats: COMPLETE
                  Filter Operator
                    predicate: (a >= 1704034800000L) (type: boolean)
                    Statistics: Num rows: 1 Data size: 8 Basic stats: COMPLETE Column stats: COMPLETE

```
Based on stats estimations the TS reads 8 rows and the filter forward only 1 row.
I run the query using both apache master and applying this patch but got the same plan and row count estimates.  

Partial plan of the second query:
```
select * from t1 where a >= 10

                TableScan
                  alias: t1
                  filterExpr: (a >= 10L) (type: boolean)
                  Statistics: Num rows: 8 Data size: 64 Basic stats: COMPLETE Column stats: COMPLETE
                  Filter Operator
                    predicate: (a >= 10L) (type: boolean)
                    Statistics: Num rows: 8 Data size: 64 Basic stats: COMPLETE Column stats: COMPLETE
```
This time the filter forward 8 rows bot on apache master and with this patch.

Please provide the repro steps to move this forward.
Thank you.

deniskuzZ:
@kasakrisz, please see the https://github.com/apache/hive/blob/master/ql/src/test/queries/clientpositive/explainanalyze_5.q
Due to type overflow (cast long to int), we estimate the wrong number of rows i.e 1

i've originally added the following test case, but explainanalyze_5.q replicates the issue as well https://github.com/apache/hive/pull/5637/files#diff-b48e5b1dc71b6ee41016e5bd56ba9e725d3e52f362aeb82de939e41de30cc753R477
````
set hive.stats.column.autogather=false;
set hive.stats.estimate=true;
set hive.stats.fetch.column.stats=true;

create table dot_n0(cint INT, cbigint BIGINT) stored as orc TBLPROPERTIES ('transactional'='true');

insert into table dot_n0 values
    (1,1), (2,2), (3,3), (4,4), (5,5), (-1070551680, 1704034800001), (-1070551681, 1704034800002);

explain analyze delete from dot_n0 where cint < -1070551679;
explain analyze delete from dot_n0 where cbigint >= 1704034800000;
````

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java [hunk 1]
```diff
@@ -1130,16 +1130,19 @@ private long evaluateComparator(Statistics stats, AnnotateStatsProcCtx aspCtx, E
...
+            } else if (colTypeLowerCase.equals(serdeConstants.TIMESTAMP_TYPE_NAME)) {
+              TimestampWritableV2 timestampWritable = new TimestampWritableV2(Timestamp.valueOf(boundValue));
+              value = timestampWritable.getTimestamp().toEpochSecond();
             } else {
               value = Integer.parseInt(boundValue);
             }
-            // Date is an integer internally
-            int maxValue = cs.getRange().maxValue.intValue();
+            long maxValue = cs.getRange().maxValue.longValue();
             int minValue = cs.getRange().minValue.intValue();
```
kasakrisz: Should `minValue` and `maxValue` have the same type?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
```diff
@@ -1166,40 +1169,42 @@ private long evaluateComparator(Statistics stats, AnnotateStatsProcCtx aspCtx, E
...
-            }
-            long maxValue = cs.getRange().maxValue.longValue();
-            long minValue = cs.getRange().minValue.longValue();
+          } else if (colTypeLowerCase.startsWith(serdeConstants.DECIMAL_TYPE_NAME) ||
+              colTypeLowerCase.equals(serdeConstants.BIGINT_TYPE_NAME)) {
+            BigDecimal value = new BigDecimal(boundValue);
+            BigDecimal maxValue = new BigDecimal(cs.getRange().maxValue.toString());
+            BigDecimal minValue = new BigDecimal(cs.getRange().minValue.toString());
+            int minComparison = value.compareTo(minValue);
+            int maxComparison = value.compareTo(maxValue);
```
kasakrisz, line 1178: Hive [BIGINT](https://hive.apache.org/docs/latest/languagemanual-types_27838462/#numeric-types) (8-byte signed integer, from `-9,223,372,036,854,775,808` to `9,223,372,036,854,775,807`) is represented as Java `long` in various places in the code.
Could you please move it to the long branch.

kasakrisz:
Do we have test coverage for all the affected datatypes (int, long, bigint, decimal, date, timestamp) ?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
deniskuzZ: no, minRange fits the int type

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
deniskuzZ: @kasakrisz, [BIGINT] can't be moved to the long branch as it would lead to overflow - that is what this PR is about, see the uniform distribution function:`(maxValue - minValue)`, `(value - minValue)`.  
I intentionally moved it to DECIMAL

deniskuzZ:
> Do we have test coverage for all the affected datatypes (int, long, bigint, decimal, date, timestamp) ?

added additional tests

deniskuzZ:
@okumin, @kasakrisz,  code freeze is today. Could we please move forward with the fix?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
kasakrisz: > no, minRange fits the int type

`maxValue` doesn't ? Why?

These min max values are coming from column stats if exists. Could it happen that a table column has all the values higher than `Integer.MAX_VALUE`? This branch handles TimeStamp types which are represented as long.

deniskuzZ: @kasakrisz we need to use long for `maxValue` to avoid overflow when we evaluate (maxValue-minValue). INT_TYPE: `Integer.MAX_VALUE` - `Integer.MIN_VALUE` = 2x`Integer.MAX_VALUE`
Additionally, I moved here `TIMESTAMP_TYPE` with maxValue of Long range. Note it's minValue is always positive, so there won't be overflow. 

I don't understand how the existing column statistics could fall outside the defined type ranges.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
kasakrisz: You mean this one?
```
Math.round(((double) (value - minValue) / (maxValue - minValue)) * currNumRows)
```
It seems that it is converted to double. Maybe the overflow is caused because the toDouble conversion is happening too late.

I think this could be better to avoid overflow
```
Math.round((((double)value - minValue) / ((double) maxValue - minValue)) * currNumRows)
```

`explainanalyze_5.q` doesn't seems to cover this scenario. Are there any other tests that covers?

kasakrisz:
> > Do we have test coverage for all the affected datatypes (int, long, bigint, decimal, date, timestamp) ?
> 
> I haven't changed the date & timestamp. Only int, bigint (represented as long) and decimal (see tpcds tests, more accurate row could estimation) existing q tests cover the intent of this PR

The reason why I ask is that for exmaple timestap had 
```
            long maxValue = cs.getRange().maxValue.longValue();
            long minValue = cs.getRange().minValue.longValue();
```
So the `minValue` was also `long`. In this patch it is `int`.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
kasakrisz: > I don't understand how the existing column statistics could fall outside the defined type ranges.

```
create table t1(cts timestamp);
insert into table t1 values ('2040-06-25 10:00:00.000'),('2040-06-25 20:00:00.000'),('2040-06-25 10:30:00.000');
analyze table t1 compute statistics for columns;
```
min: 2040-06-25 10:00:00.000 - epoch: 2224231200
max: 2040-06-25 20:00:00.000 - epoch: 2224267200

Both should be long because the min value `2224231200` is higher than `Integer.MAX_VALUE` (2147483647)

Probably this is only going to be a problem in 15 years in the future. :)

To avoid overflow the operands could be converted to double: `((double)maxValue-minValue)`
and similarly `((double)value - minValue)`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
okumin: I personally prefer to retain the result of the intermediate computation as an integral number

deniskuzZ:
> > > Do we have test coverage for all the affected datatypes (int, long, bigint, decimal, date, timestamp) ?
> > 
> > 
> > I haven't changed the date & timestamp. Only int, bigint (represented as long) and decimal (see tpcds tests, more accurate row could estimation) existing q tests cover the intent of this PR
> 
> The reason why I ask is that for exmaple timestap had
> 
> ```
>             long maxValue = cs.getRange().maxValue.longValue();
>             long minValue = cs.getRange().minValue.longValue();
> ```
> 
> So the `minValue` was also `long`. In this patch it is `int`.

fixed

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
deniskuzZ: explainanalyze_5.q covers that INT type, BIGINT, DECIMAL are covered by the rest
It's hard to understand why there's resistance when the bug is clearly evident.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
deniskuzZ: why would you need double here? if max can't be > long. 
_note, I've updated the type for timestamp minValue_

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
deniskuzZ: We already used BigDecimal for Decimal types, so that seemed like a right choice for BIGINT as well.
Btw, Gemini 2.5-Pro used BigDecimal for uniform distribution with Long type ranges

deniskuzZ:
not chasing this in 4.1 release anymore

deniskuzZ:
@kasakrisz, have I covered all your concerns? 
Let me know if there are any additional tests you'd like to include: https://github.com/apache/hive/blob/ecd3dfab281e647fceed18db1ed5ebe748822908/ql/src/test/queries/clientpositive/explainanalyze_6.q"

FYI, the latest commit also updated the date and timestamp ranges to future values:
https://github.com/apache/hive/pull/5637/files#diff-1b0e96106a06cddf1ca7e77f3b4cf45d849d86db4766fa7a8e9e945567f41cf3R147

ql/src/java/org/apache/hadoop/hive/ql/optimizer/stats/annotation/StatsRulesProcFactory.java
kasakrisz: > explainanalyze_5.q covers that INT type, BIGINT, DECIMAL are covered by the rest It's hard to understand why there's resistance when the bug is clearly evident.

Thank you for adding explainanalyze_6.q to cover BIGINT and Timestamp.

PR #5639, opened 2025-02-09
HIVE-28749: The default hikaricp.leakDetectionThreshold is not valid

okumin:

HicariConfig's leak detection threshold has to be smaller than the max lifetime. Our default leak detection threshold is greater than the maximum lifetime defined on the HikariCP side and is disabled by default. Please check the JIRA for details.

https://issues.apache.org/jira/browse/HIVE-28749

Make our default value effective.

No.

No.

- I verified the warning disappears with default configurations
- I tested `hikaricp.maxLifetime` was configurable in my local machine

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
@@ -17,8 +17,6 @@
  */
 package org.apache.hadoop.hive.metastore.datasource;
 
-import static java.util.concurrent.TimeUnit.MINUTES;
-
 import com.codahale.metrics.MetricRegistry;
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
@@ -42,8 +40,6 @@ public class HikariCPDataSourceProvider implements DataSourceProvider {
   private static final Logger LOG = LoggerFactory.getLogger(HikariCPDataSourceProvider.class);
 
   static final String HIKARI = "hikaricp";
-  private static final String LEAK_DETECTION_THRESHOLD = "leakDetectionThreshold";
-  private static final long DEFAULT_LEAK_DETECTION_THRESHOLD = MINUTES.toMillis(10);
 
   @Override
   public DataSource create(Configuration hdpConfig, int maxPoolSize) throws SQLException {
@@ -55,7 +51,6 @@ public class HikariCPDataSourceProvider implements DataSourceProvider {
     String passwd = DataSourceProvider.getMetastoreJdbcPasswd(hdpConfig);
 
     Properties properties = replacePrefix(DataSourceProvider.getPrefixedProperties(hdpConfig, HIKARI));
-    properties.putIfAbsent(LEAK_DETECTION_THRESHOLD, DEFAULT_LEAK_DETECTION_THRESHOLD);
 
     HikariConfig config;
     try {
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
@@ -17,6 +17,8 @@
  */
 package org.apache.hadoop.hive.metastore.datasource;
 
+import static java.util.concurrent.TimeUnit.MINUTES;
+
 import com.codahale.metrics.MetricRegistry;
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
@@ -40,9 +42,8 @@ public class HikariCPDataSourceProvider implements DataSourceProvider {
   private static final Logger LOG = LoggerFactory.getLogger(HikariCPDataSourceProvider.class);
 
   static final String HIKARI = "hikaricp";
-  private static final String CONNECTION_TIMEOUT_PROPERTY = HIKARI + ".connectionTimeout";
-  private static final String MAX_LIFETIME = HIKARI + ".maxLifetime";
-  private static final String LEAK_DETECTION_THRESHOLD = HIKARI + ".leakDetectionThreshold";
+  private static final String LEAK_DETECTION_THRESHOLD = "leakDetectionThreshold";
+  private static final long DEFAULT_LEAK_DETECTION_THRESHOLD = MINUTES.toMillis(10);
 
   @Override
   public DataSource create(Configuration hdpConfig, int maxPoolSize) throws SQLException {
@@ -53,11 +54,8 @@ public class HikariCPDataSourceProvider implements DataSourceProvider {
     String user = DataSourceProvider.getMetastoreJdbcUser(hdpConfig);
     String passwd = DataSourceProvider.getMetastoreJdbcPasswd(hdpConfig);
 
-    Properties properties = replacePrefix(
-        DataSourceProvider.getPrefixedProperties(hdpConfig, HIKARI));
-    long connectionTimeout = hdpConfig.getLong(CONNECTION_TIMEOUT_PROPERTY, 30000L);
-    long maxLifetime = hdpConfig.getLong(MAX_LIFETIME, 3600000L);
-    long leakDetectionThreshold = hdpConfig.getLong(LEAK_DETECTION_THRESHOLD, 3600000L);
+    Properties properties = replacePrefix(DataSourceProvider.getPrefixedProperties(hdpConfig, HIKARI));
+    properties.putIfAbsent(LEAK_DETECTION_THRESHOLD, DEFAULT_LEAK_DETECTION_THRESHOLD);
 
     HikariConfig config;
     try {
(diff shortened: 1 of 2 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
```diff
@@ -55,6 +56,7 @@ public DataSource create(Configuration hdpConfig, int maxPoolSize) throws SQLExc
     Properties properties = replacePrefix(
         DataSourceProvider.getPrefixedProperties(hdpConfig, HIKARI));
     long connectionTimeout = hdpConfig.getLong(CONNECTION_TIMEOUT_PROPERTY, 30000L);
+    long maxLifetime = hdpConfig.getLong(MAX_LIFETIME, 3600000L);
```
dengzhhu653, line 59: what if the max lifetime is equal to the leak detection, will the rule for leak detection not take effect as the max lifetime takes first?
To me, the default of LEAK_DETECTION_THRESHOLD is too long, we can turn down this value, what do you think?

zabetak: If we add MAX_LIFETIME then we are adding a new config toggle that was not available before. I am not against doing it but it seems like independent of the main purpose of the PR.

Regarding the value of the threshold I agree with Denys that 1h is probably too long and it makes sense to lower it. In fact, since it is a debug property maybe we should keep it disabled by default (i.e., `0`). I don't know how easy it is to pick a value that is useful in every use-case of the metastore.

zabetak: After checking a bit further the code, I get the impression that `getPrefixedProperties` block is sufficient to read any user defined Hikari values from the conf. I don't understand what's the purpose of using/adding explicit getters/setters per property.

okumin: Thanks for reviewing. I answer the questions of myself and yours first.

My question 1: What's the max lifetime?

Let's see [Configuration (knobs, baby!) in the official document](https://github.com/brettwooldridge/HikariCP?tab=readme-ov-file#gear-configuration-knobs-baby).

> This property controls the maximum lifetime of a connection in the pool. An in-use connection will never be retired, only when it is closed will it then be removed. On a connection-by-connection basis, minor negative attenuation is applied to avoid mass-extinction in the pool. We strongly recommend setting this value, and it should be several seconds shorter than any database or infrastructure imposed connection time limit. A value of 0 indicates no maximum lifetime (infinite lifetime), subject of course to the idleTimeout setting. The minimum allowed value is 30000ms (30 seconds). Default: 1800000 (30 minutes)

My question 2: What's the leak detection threshold?

> This property controls the amount of time that a connection can be out of the pool before a message is logged indicating a possible connection leak. A value of 0 means leak detection is disabled. Lowest acceptable value for enabling leak detection is 2000 (2 seconds). Default: 0

From the same document, [Configuration (knobs, baby!) in the official document](https://github.com/brettwooldridge/HikariCP?tab=readme-ov-file#gear-configuration-knobs-baby).

[HikariCP makes a warning log when the leak threshold is violated](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/ProxyLeakTask.java).

My question 3: Why did we change our default value?

HIVE-25482 says the original author faced a corner case and suggests making it configurable and configuring the default value.

My question 4: When did 1 hour become invalid?

[This is the merge commit of HIVE-25482](https://github.com/apache/hive/commit/c066d5d2b5098bb7cd4b7a5500cc076c32dccf1e). [We were using HikariCP 2.6.1 at that time](https://github.com/apache/hive/blob/c066d5d2b5098bb7cd4b7a5500cc076c32dccf1e/pom.xml#L150). The [MAX_LIFETIME of 2.6.1 is 30 minutes](https://github.com/brettwooldridge/HikariCP/blob/HikariCP-2.6.1/src/main/java/com/zaxxer/hikari/HikariConfig.java#L56) and [it already had the same validation](https://github.com/brettwooldridge/HikariCP/blob/HikariCP-2.6.1/src/main/java/com/zaxxer/hikari/HikariConfig.java#L890-L895). So, our 1 hour has been invalid since we merged HIVE-25482.

> what if the max lifetime is equal to the leak detection, will the rule for leak detection not take effect as the max lifetime takes first?

I guess it is possible, though I've not tested it.

> After checking a bit further the code, I get the impression that getPrefixedProperties block is sufficient to read any user defined Hikari values from the conf. I don't understand what's the purpose of using/adding explicit getters/setters per property.

I guess we wanted to overwrite the default value, but I'm not sure. Let me test some cases

okumin: @dengzhhu653 @zabetak 

> To me, the default of LEAK_DETECTION_THRESHOLD is too long, we can turn down this value, what do you think?
> Regarding the value of the threshold I agree with @dengzhhu653 that 1h is probably too long and it makes sense to lower it. In fact, since it is a debug property maybe we should keep it disabled by default (i.e., 0). I don't know how easy it is to pick a value that is useful in every use-case of the metastore.

It makes sense to lower the threshold.
I have no strong opinion about 0 vs moderate minutes. I'd keep it to be positive based on the report of HIVE-25482, as the parameter is not harmless anyway. 0 is also an option as we don't know a specific case where HIVE-25482 is needed.

The updated revision configures 10 minutes, cleaning up the class leveraging `getPrefixedProperties`. 0 seconds vs 10 minutes vs else could be still controversial.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
```diff
@@ -79,13 +78,10 @@ public DataSource create(Configuration hdpConfig, int maxPoolSize) throws SQLExc
     // so that the connection pool can retire the idle connection aggressively,
     // this will make Metastore more scalable especially if there is a leader in the warehouse.
     if ("mutex".equals(poolName)) {
-      int minimumIdle = Integer.valueOf(hdpConfig.get(HIKARI + ".minimumIdle", "2"));
```
zabetak, line 82: This change seems unrelated to the purpose of the PR. Moreover, it seems to remove the caching of small integers that happens when `valueOf` is used. Maybe its better to leave things as is.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
zabetak: I don't know if HIVE-25482 really wanted to set the default value to 1hr. I get the impression that HIVE-25482 was created based on the misunderstanding the the leak threshold was not configurable while in fact it was.

Since the default value of 1hr was always invalid it practically means that if we leave it at zero we are not changing anything in the current behavior. I guess the Hikaricp authors left the default to `0` because it is really use-case specific.

Anyways, I don't feel strongly about zero, 10 minutes, or another value, since when there is a problem we will most likely need to set this property again anyways. I leave the final decision to you @okumin

zabetak (approved):
Left a minor comment but otherwise LGTM! Feel free to submit once you get a green run.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
zabetak: I just seen that we pass the value to a primitive int so calling `valueOf` is completely useless.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
okumin: Sure, so let's make it zero now and respect the current behavior.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
```diff
@@ -52,10 +50,7 @@ public DataSource create(Configuration hdpConfig, int maxPoolSize) throws SQLExc
     String user = DataSourceProvider.getMetastoreJdbcUser(hdpConfig);
     String passwd = DataSourceProvider.getMetastoreJdbcPasswd(hdpConfig);
 
-    Properties properties = replacePrefix(
-        DataSourceProvider.getPrefixedProperties(hdpConfig, HIKARI));
-    long connectionTimeout = hdpConfig.getLong(CONNECTION_TIMEOUT_PROPERTY, 30000L);
-    long leakDetectionThreshold = hdpConfig.getLong(LEAK_DETECTION_THRESHOLD, 3600000L);
```
okumin, line 58: As a new unit test explains, the default connection timeout is originally 30 seconds.
We also decided to use the default leak detection threshold. It doesn't change the behavior because the current value is invalid, and the detection is disabled.

dengzhhu653 (approved):

PR #3214, opened 2022-04-15
HIVE-26144: Add keys/indexes to support highly concurrent workload

kovjanos:
### What changes were proposed in this pull request?
Missing keys/index is to be added to the HMS backend db schema

### Why are the changes needed?
On a high-concurrency test we found that backend database is doing full table scans in some cases where the table has missing key/index.

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
Integration tests for all database types:
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.hadoop.hive.metastore.dbinstall.ITestPostgres
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 13.949 s - in org.apache.hadoop.hive.metastore.dbinstall.ITestPostgres
[INFO] Running org.apache.hadoop.hive.metastore.dbinstall.ITestMssql
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 33.711 s - in org.apache.hadoop.hive.metastore.dbinstall.ITestMssql
[INFO] Running org.apache.hadoop.hive.metastore.dbinstall.ITestMysql
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 39.032 s - in org.apache.hadoop.hive.metastore.dbinstall.ITestMysql
[INFO] Running org.apache.hadoop.hive.metastore.dbinstall.ITestDerby
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.101 s - in org.apache.hadoop.hive.metastore.dbinstall.ITestDerby
[INFO] Running org.apache.hadoop.hive.metastore.dbinstall.ITestOracle
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 94.393 s - in org.apache.hadoop.hive.metastore.dbinstall.ITestOracle
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0
```

PR #5641, opened 2025-02-12
HIVE-17306: Support MySQL InnoDB Cluster

KiranVelumuri:

Introduce primary key columns to support HA for MySQL cluster. This works for MySQL 8.0.23 and above.

[HIVE-17306](https://issues.apache.org/jira/browse/HIVE-17306)

No

No

Perform schema upgrade and verify transactions are working fine, and the meta tables are getting updated as earlier.


KiranVelumuri:
@zabetak @okumin Could you please help to review? Thank you.

itests/util/src/main/java/org/apache/hadoop/hive/ql/externalDB/MariaDB.java
```diff
@@ -36,7 +36,7 @@ public String getJdbcDriver() {
     return "org.mariadb.jdbc.Driver";
   }
 
-  public String getDockerImageName() { return "mariadb:10.2"; }
+  public String getDockerImageName() { return "mariadb:11.4"; }
```
KiranVelumuri, line 40: These changes are part of MySQL upgrade in #5525. Will rebase once it is merged.

itests/util/src/main/java/org/apache/hadoop/hive/ql/externalDB/MySQLExternalDB.java
```diff
@@ -53,8 +53,8 @@ public String[] getDockerAdditionalArgs() {
     }
 
     public boolean isContainerReady(ProcessResults pr) {
-        Pattern pat = Pattern.compile("ready for connections");
+        Pattern pat = Pattern.compile("mysqld.*ready for connections.*port.*3306");
         Matcher m = pat.matcher(pr.stderr);
-        return m.find() && m.find();
+        return m.find();
```
KiranVelumuri, line 58: These changes are part of MySQL and MariaDB upgrade in #5525. Will rebase once it is merged.

standalone-metastore/metastore-server/src/main/sql/derby/hive-schema-4.1.0.derby.sql
```diff
@@ -679,14 +685,16 @@ CREATE TABLE WRITE_SET (
...
+  INV_ID BIGINT GENERATED ALWAYS AS IDENTITY (START WITH 1, INCREMENT BY 1) PRIMARY KEY
 );
 
 CREATE TABLE TXN_TO_WRITE_ID (
   T2W_TXNID bigint NOT NULL,
   T2W_DATABASE varchar(128) NOT NULL,
   T2W_TABLE varchar(256) NOT NULL,
-  T2W_WRITEID bigint NOT NULL
+  T2W_WRITEID bigint NOT NULL,
+  CONSTRAINT TXN_TO_WRITE_ID_PK PRIMARY KEY (T2W_TXNID, T2W_DATABASE, T2W_TABLE)
```
KiranVelumuri, line 697: In case of multi table insert, in the same transaction we update two or more tables. In such case to uniquely identify each row, I included DB and table as well in the primary key.

standalone-metastore/metastore-server/src/main/sql/derby/hive-schema-4.1.0.derby.sql
```diff
@@ -704,7 +713,8 @@ CREATE TABLE MIN_HISTORY_WRITE_ID (
   MH_TXNID bigint NOT NULL REFERENCES TXNS (TXN_ID),
   MH_DATABASE varchar(128) NOT NULL,
   MH_TABLE varchar(256) NOT NULL,
-  MH_WRITEID bigint NOT NULL
+  MH_WRITEID bigint NOT NULL,
+  CONSTRAINT MIN_HISTORY_WRITE_ID_PK PRIMARY KEY (MH_TXNID, MH_DATABASE, MH_TABLE, MH_WRITEID)
```
KiranVelumuri, line 717: For the same transaction ID, list of write IDs are inserted into this table. So, write ID also has to be included in the primary key.

KiranVelumuri:
image

Ref: https://www.trevorlasn.com/blog/invisible-columns-in-sql

KiranVelumuri:
@deniskuzZ @abstractdog @zhangbutao @wecharyu @dengzhhu653 Could you please help to review? Thank you.

deniskuzZ:
@KiranVelumuri, I don't understand what it solves or why you are changing the schema for all the supported DBs.

KiranVelumuri:
To implement High Availability (HA) for MySQL using the InnoDB cluster, it is a requirement that primary keys be present for all the tables. So, this PR is for adding primary keys to all such tables in Hive's schema which do not have a primary key defined. For the case of tables which cannot have unique primary keys, I added additional invisible column.

deniskuzZ:
@KiranVelumuri thanks for the explanation. It seems similar to https://github.com/apache/hive/pull/3214
cc @kovjanos

chinnaraolalam:
@KiranVelumuri Is all the supported DB's support invisible columns. Even though if syntax supports what test are performed to confirm this.

standalone-metastore/metastore-server/src/main/sql/mysql/hive-schema-4.1.0.mysql.sql
```diff
@@ -1134,6 +1143,7 @@ CREATE TABLE MIN_HISTORY_WRITE_ID (
   MH_DATABASE varchar(128) NOT NULL,
   MH_TABLE varchar(256) NOT NULL,
   MH_WRITEID bigint NOT NULL,
+  CONSTRAINT MIN_HISTORY_WRITE_ID_PK PRIMARY KEY (MH_TXNID, MH_DATABASE, MH_TABLE, MH_WRITEID),
```
chinnaraolalam, line 1146: Composite keys on more columns may effect the performance. How this is advantage than adding invisible primary key or vice versa .

standalone-metastore/metastore-server/src/main/sql/mysql/upgrade-4.0.0-to-4.1.0.mysql.sql
```diff
@@ -14,6 +14,18 @@ DROP INDEX UNIQUETABLE ON TBLS;
 CREATE UNIQUE INDEX UNIQUETABLE ON TBLS (DB_ID, TBL_NAME);
 DROP INDEX TBLS_N49 ON TBLS;
 
+-- HIVE-17306
+ALTER TABLE `COMPLETED_TXN_COMPONENTS` ADD COLUMN `INV_ID` BIGINT AUTO_INCREMENT INVISIBLE PRIMARY KEY;
```
chinnaraolalam, line 18: What about the existing rows value for newly added column

KiranVelumuri: The newly added column values are auto-generated for the existing rows.

KiranVelumuri:
> @KiranVelumuri Is all the supported DB's support invisible columns. Even though if syntax supports what test are performed to confirm this.

Not all DBs support invisible columns. Please refer to my [earlier comment](https://github.com/apache/hive/pull/5641#issuecomment-2743128332).
We have Docker tests in our Jenkins that validate fresh install and upgrade scenarios, which have all passed.

standalone-metastore/metastore-server/src/main/sql/mysql/hive-schema-4.1.0.mysql.sql
KiranVelumuri: Yes, there might be a performance impact. I have done this to add primary key wherever possible, if not then I added the invisible column.

deniskuzZ:
I am not sure if we can accept these changes. 
Looping HMS folks: @saihemanth-cloudera, @dengzhhu653

standalone-metastore/metastore-server/src/main/sql/derby/hive-schema-4.1.0.derby.sql
saihemanth-cloudera: I like the idea of having an invisible column (although not supported in derby) for all the tables as these composite primary keys are not used anyway, and maintaining a composite key becomes complex.

standalone-metastore/metastore-server/src/main/sql/mysql/hive-schema-4.1.0.mysql.sql
saihemanth-cloudera: I like the idea of having invisible columns almost everywhere as we don't explicitly use these primary keys in the application logic.

KiranVelumuri: @saihemanth-cloudera Yes, we could have invisible columns for all the missing primary key tables only for MySQL, since not all schemas support invisible columns. Should this approach be okay?

PR #5642, opened 2025-02-15
HIVE-28759: Hive Query History - records are failed to be written due to iceberg worker pools shut down

abstractdog:
Replaces iceberg's worker thread pool with a same-sized one that accepts tasks after jvm shutdown hook.

Described on jira, tldr, to make query history able to flush records on hs2 shutdown.

No.

No.

Manually tested on local MiniHS2 and Cloudera Data Warehouse, the HS2 showed that the full commit process ran successfully. Without the patch, the RejectedExecutionException was present in both scenarios.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -109,7 +109,7 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
     return OUTPUT_COMMITTER;
   }
 
-  private ExecutorService snapshotUpdateExecutor;
+  private ExecutorService workerPool;
 
   @Override
   public void setupJob(JobContext jobContext) {
@@ -225,11 +225,11 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
   }
 
   /**
-   * Receives a custom executor to be used for SnapshotUpdate.commit() operations.
-   * @param snapshotUpdateExecutor to pass to SnapshotUpdates
+   * Receives a custom workerPool to be used for SnapshotUpdate.commit() operations.
+   * @param workerPool to pass to SnapshotUpdates
    */
-  public void setSnapshotUpdateExecutor(ExecutorService snapshotUpdateExecutor) {
-    this.snapshotUpdateExecutor = snapshotUpdateExecutor;
+  public void setWorkerPool(ExecutorService workerPool) {
+    this.workerPool = workerPool;
   }
 
   /**
@@ -598,8 +598,8 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
    * @param update the SnapshotUpdate of any kind (e.g. AppendFiles, DeleteFiles, etc.)
    */
   private void commit(SnapshotUpdate<?> update) {
-    if (snapshotUpdateExecutor != null) {
-      update.scanManifestsWith(snapshotUpdateExecutor);
+    if (workerPool != null) {
+      update.scanManifestsWith(workerPool);
     }
     update.commit();
   }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -68,6 +68,7 @@ import org.apache.iceberg.ReplacePartitions;
 import org.apache.iceberg.RewriteFiles;
 import org.apache.iceberg.RowDelta;
 import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.SnapshotUpdate;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.exceptions.NotFoundException;
 import org.apache.iceberg.expressions.Expression;
@@ -107,6 +108,8 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
   public static HiveIcebergOutputCommitter getInstance() {
     return OUTPUT_COMMITTER;
   }
+
+  private ExecutorService snapshotUpdateExecutor;
 
   @Override
   public void setupJob(JobContext jobContext) {
@@ -219,6 +222,14 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
   @Override
   public void commitJob(JobContext originalContext) throws IOException {
     commitJobs(Collections.singletonList(originalContext), Operation.OTHER);
+  }
+
+  /**
+   * Receives a custom executor to be used for SnapshotUpdate.commit() operations.
+   * @param snapshotUpdateExecutor to pass to SnapshotUpdates
+   */
+  public void setSnapshotUpdateExecutor(ExecutorService snapshotUpdateExecutor) {
+    this.snapshotUpdateExecutor = snapshotUpdateExecutor;
   }
 
   /**
@@ -543,7 +554,7 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
       write.conflictDetectionFilter(filterExpr);
       write.validateNoConflictingData();
       write.validateNoConflictingDeletes();
-      write.commit();
+      commit(write);
       return;
     }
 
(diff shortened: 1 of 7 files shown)


ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
```diff
@@ -65,6 +71,43 @@ public class IcebergRepository extends AbstractRepository implements QueryHistor
...
+    Field field = Class.forName("org.apache.iceberg.util.ThreadPools").getDeclaredField("WORKER_POOL");
+    field.setAccessible(true);
+
+    Field modifiersField = Field.class.getDeclaredField("modifiers");
+    modifiersField.setAccessible(true);
+    modifiersField.setInt(field, field.getModifiers() & ~Modifier.FINAL);
+
+    int poolSize = getIcebergWorkerPoolSize();
+    LOG.info("Overriding iceberg worker pool, size: {}", poolSize);
+    field.set(null, Executors.newFixedThreadPool(poolSize,
```
deniskuzZ, line 99: this is a hack! we shouldn't touch `iceberg-worker-pool"` just for the sake of query history.

aturoczy: I also share your concern, but do you have any other recommendation how could be this issue resolvable?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -108,6 +109,8 @@ public static HiveIcebergOutputCommitter getInstance() {
     return OUTPUT_COMMITTER;
   }
 
+  private ExecutorService snapshotUpdateExecutor;
```
deniskuzZ, line 112: could we use the same naming as in iceberg? i.e. `workerPool`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -925,7 +925,7 @@ public boolean commitInMoveTask() {
   }
 
   @Override
-  public void storageHandlerCommit(Properties commitProperties, Operation operation)
+  public void storageHandlerCommit(Properties commitProperties, Operation operation, ExecutorService suppliedExecutor)
```
deniskuzZ, line 928: `executorService`

ql/src/java/org/apache/hadoop/hive/ql/exec/MoveTask.java
```diff
@@ -1157,7 +1157,7 @@ private boolean checkAndCommitNatively(MoveWork moveWork, Configuration configur
     if (storageHandlerClass != null) {
       HiveStorageHandler storageHandler = HiveUtils.getStorageHandler(configuration, storageHandlerClass);
       if (storageHandler.commitInMoveTask()) {
-        storageHandler.storageHandlerCommit(commitProperties, operation);
+        storageHandler.storageHandlerCommit(commitProperties, operation, null);
```
deniskuzZ, line 1160: please keep the original signature. add the default in the interface

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
```diff
@@ -613,15 +614,17 @@ default boolean commitInMoveTask() {
    * Commits the inserts for the non-native tables. Used in the {@link org.apache.hadoop.hive.ql.exec.MoveTask}.
    * @param commitProperties Commit properties which are needed for the handler based commit
    * @param operation the operation type
+   * @param suppliedExecutor an ExecutorService to be used by the StorageHandler (optional)
    * @throws HiveException If there is an error during commit
    */
-  default void storageHandlerCommit(Properties commitProperties, Operation operation) throws HiveException {
+  default void storageHandlerCommit(Properties commitProperties, Operation operation, ExecutorService suppliedExecutor)
```
deniskuzZ, line 620: 1. keep original signature
````
default void storageHandlerCommit(Properties commitProperties, Operation operation) {
  storageHandlerCommit(commitProperties, operation, null)
}
````
2. `executorService`

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java [hunk 6]
```diff
@@ -613,15 +614,17 @@ default boolean commitInMoveTask() {
...
+      throws HiveException {
     throw new UnsupportedOperationException();
   }
 
   @Deprecated
   default void storageHandlerCommit(Properties commitProperties, boolean overwrite) throws HiveException {
-    storageHandlerCommit(commitProperties, overwrite ? Operation.IOW : Operation.OTHER);
+    storageHandlerCommit(commitProperties, overwrite ? Operation.IOW : Operation.OTHER, null);
```
deniskuzZ, line 627: remove this

ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
```diff
@@ -53,17 +55,41 @@
...
+        (new ThreadFactoryBuilder()).setDaemon(true).setNameFormat(ICEBERG_WORKER_THREAD_NAME_FORMAT).build());
+  }
+
+  /**
+   * Calculates the pool size for the ExecutorService to be passed to the committer.
+   * The thread pool adheres to the configuration settings for Iceberg worker pools.
+   * @return pool size for iceberg worker pool
+   */
+  private int getIcebergWorkerPoolSize() {
+    String value = System.getProperty("iceberg.worker.num-threads");
```
deniskuzZ, line 87: you don't need for flush the same config as for iceberg worker pool. 2 threads should be enough

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -574,14 +585,24 @@ private void commitWrite(Table table, String branchName, Long snapshotId, long s
...
     LOG.info("Write commit took {} ms for table: {} with {} data and {} delete file(s)",
         System.currentTimeMillis() - startTime, table, results.dataFiles().size(), results.deleteFiles().size());
     LOG.debug("Added files {}", results);
   }
 
+  /**
+   * Calls the commit on the prepared SnapshotUpdate and supplies the ExecutorService if any.
+   * @param update the SnapshotUpdate of any kind (e.g. AppendFiles, DeleteFiles, etc.)
+   */
+  private void commit(SnapshotUpdate<?> update) {
```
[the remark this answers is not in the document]
deniskuzZ: this is an iceberg change commit, just implicit

ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
abstractdog: okay, using 2 threads will keep this simpler

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
deniskuzZ: TBH i would rename the method to simply `commit`. why do we need this `storageHandlerCommit` IDK

abstractdog: that's not possible I'm afraid, ThreadPools is in iceberg, hive-exec doesn't depend on that

ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
abstractdog: resolving this conversation, we moved away from this reflection hack in the meantime, fortunately

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java
deniskuzZ: that's ok, use `null` then

abstractdog:
addressed comments in https://github.com/apache/hive/pull/5642/commits/85ba8e2cceaa24ed81f3b0cb83b4c1103720ba8a

deniskuzZ (approved):
LGTM +1
pending tests

PR #5647, opened 2025-02-20
HIVE-28768: Remove hardcoded post exec hooks

Aggarwal-Raghav:
[HIVE-28768](https://issues.apache.org/jira/browse/HIVE-28768), Use HIVE_LINEAGE_STATEMENT_FILTER configuration to determine whether to compute lineage information or not and for which query type (CTAS, Create View etc) the lineage should be computed.

Currently, lineage information is computed only if few post exec hooks are configured or HIVE_LINEAGE_INFO (doesn't capture "views" related lineage) is set to true. In [HIVE-28409](https://issues.apache.org/jira/browse/HIVE-28409), fine grain control over lineage was introduced and its better to utilize that.

**Point to Ponder**: The lineage information computation will be by-default enabled for ALL query.
For more info please check the discussion [here](https://issues.apache.org/jira/browse/HIVE-28409?focusedCommentId=17927788&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17927788)

YES

NO

On local mac setup and running lineage{1...6}.q and other lineage related q files.

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -64,6 +64,7 @@ import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.time.ZoneId;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -7301,6 +7302,12 @@ public class HiveConf extends Configuration {
       }
     }
     return ret;
+  }
+
+  public static boolean shouldComputeLineage(HiveConf conf) {
+    Collection<String> lineage_filter =
+      conf.getTrimmedStringCollection(HiveConf.ConfVars.HIVE_LINEAGE_STATEMENT_FILTER.varname);
+    return !(lineage_filter.isEmpty() || lineage_filter.contains("NONE"));
   }
 
   // sync all configs from given conf
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -7305,9 +7305,10 @@ public class HiveConf extends Configuration {
   }
 
   public static boolean shouldComputeLineage(HiveConf conf) {
-    Collection<String> lineage_filter =
+    Collection<String> lineageFilter =
       conf.getTrimmedStringCollection(HiveConf.ConfVars.HIVE_LINEAGE_STATEMENT_FILTER.varname);
-    return !(lineage_filter.isEmpty() || lineage_filter.contains("NONE"));
+    return !(lineageFilter.isEmpty() || lineageFilter.contains("NONE"))
+      || conf.getBoolVar(ConfVars.HIVE_LINEAGE_INFO);
   }
 
   // sync all configs from given conf
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -3881,7 +3881,7 @@ public class HiveConf extends Configuration {
     HIVE_LINEAGE_INFO("hive.lineage.hook.info.enabled", false,
         "Whether Hive provides lineage information to hooks." +
             "Deprecated: use hive.lineage.statement.filter instead."),
-    HIVE_LINEAGE_STATEMENT_FILTER("hive.lineage.statement.filter", "ALL",
+    HIVE_LINEAGE_STATEMENT_FILTER("hive.lineage.statement.filter", "NONE",
         "Whether Hive provides lineage information to hooks for the specified statements only, " +
             "the value is a comma-separated list (ex.: CREATE_MATERIALIZED_VIEW," +
             "CREATE_TABLE,CREATE_TABLE_AS_SELECT). Possible values are: CREATE_TABLE, CREATE_TABLE_AS_SELECT, " +
--- a/data/conf/hive-site.xml
+++ b/data/conf/hive-site.xml
@@ -155,6 +155,12 @@
 </property>
 
 <property>
+  <name>hive.lineage.statement.filter</name>
+  <value>ALL</value>
+  <description>Lineage information filter</description>
+</property>
+
+<property>
   <name>hive.exec.post.hooks</name>
   <value>org.apache.hadoop.hive.ql.hooks.PostExecutePrinter</value>
   <description>Post Execute Hook for Tests</description>
--- a/data/conf/hive-site.xml
+++ b/data/conf/hive-site.xml
@@ -157,7 +157,7 @@
 <property>
   <name>hive.lineage.statement.filter</name>
   <value>ALL</value>
-  <description>Lineage information filter</description>
+  <description>Specify the types of statements for which column lineage information is generated</description>
 </property>
 
 <property>
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
@@ -90,7 +90,7 @@ public class Generator extends Transform {
   static Predicate<ParseContext> createFilterPredicateFromConf(Configuration conf) {
     Set<LineageInfoFilter> operations = new HashSet<>();
     boolean noneSpecified = false;
-    for (String valueText : conf.getStringCollection(HiveConf.ConfVars.HIVE_LINEAGE_STATEMENT_FILTER.varname)) {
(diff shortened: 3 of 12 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/Generator.java
```diff
@@ -90,7 +90,7 @@ public static Generator fromConf(HiveConf conf) {
   static Predicate<ParseContext> createFilterPredicateFromConf(Configuration conf) {
     Set<LineageInfoFilter> operations = new HashSet<>();
     boolean noneSpecified = false;
-    for (String valueText : conf.getStringCollection(HiveConf.ConfVars.HIVE_LINEAGE_STATEMENT_FILTER.varname)) {
+    for (String valueText : conf.getTrimmedStringCollection(HiveConf.ConfVars.HIVE_LINEAGE_STATEMENT_FILTER.varname)) {
```
Aggarwal-Raghav, line 93: IMO, better to use getTrimmedStringCollection method as it protects from throwing the following error if users put space while setting the value of config.
```
Error while compiling statement: FAILED: EnumConstantNotPresentException org.apache.hadoop.hive.ql.optimizer.lineage.Generator$LineageInfoFilter. CREATE_VIEW;
Query ID: raghav_20250219220810_850ffba9-62a7-40b5-95bb-ff2db6769696 (state=42000, code=40000)
```

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -7303,6 +7304,12 @@ public Map<String, String> subtree(String string) {
     return ret;
   }
 
+  public static boolean shouldComputeLineage(HiveConf conf) {
+    Collection<String> lineage_filter =
+      conf.getTrimmedStringCollection(HiveConf.ConfVars.HIVE_LINEAGE_STATEMENT_FILTER.varname);
+    return !(lineage_filter.isEmpty() || lineage_filter.contains("NONE"));
```
kasakrisz, line 7310: I think HIVE_LINEAGE_INFO also has to be checked like
```
return !(lineage_filter.isEmpty() || lineage_filter.contains("NONE")) ||
  conf.getBoolVar(HiveConf.ConfVars.HIVE_LINEAGE_INFO);
```
Or it should be removed completely since it is deprecated.
Maybe we should keep it in the next Hive release but remove it in a follow-up patch for the 2nd release.

kasakrisz:
@Aggarwal-Raghav 

> The lineage information computation will be by-default enabled for ALL query.

I think currently `org.apache.hadoop.hive.ql.hooks.PostExecutePrinter` is always added to `hive.exec.post.hooks` in `hive-site.xml` so it has the same impact.
Could you please double check?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -7303,6 +7304,12 @@ public Map<String, String> subtree(String string) {
     return ret;
   }
 
+  public static boolean shouldComputeLineage(HiveConf conf) {
+    Collection<String> lineage_filter =
```
kasakrisz, line 7308: Please change the name of `lineage_filter` to `lineageFilter`

kasakrisz:
Please check the sonarqubecloud report and try to fix the issues introduced by this patch if it makes sense.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
InvisibleProgrammer: Agree with that. It is deprecated, not removed. So it should still work.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
Aggarwal-Raghav: Ack

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
Aggarwal-Raghav: Ack

Aggarwal-Raghav:
> @Aggarwal-Raghav
> 
> > The lineage information computation will be by-default enabled for ALL query.
> 
> I think currently `org.apache.hadoop.hive.ql.hooks.PostExecutePrinter` is always added to `hive.exec.post.hooks` in `hive-site.xml` so it has the same impact. Could you please double check?

No, i don't think it is always getting added in _hive-site.xml_. I checked in _hive-default.xml.template_ (which is created for new hive dist in conf dir) and also checked in local cluster, if _hive.exec.post.hooks_ key is not set in hive-site.xml then:
```
0: jdbc:hive2://localhost:10000 (default)> set hive.exec.post.hooks;
+------------------------------------+
|                set                 |
+------------------------------------+
| hive.exec.post.hooks is undefined  |
+------------------------------------+
```

kasakrisz:
> > > @Aggarwal-Raghav
> > > > The lineage information computation will be by-default enabled for ALL query.
> > > 
> > > 
> > > I think currently `org.apache.hadoop.hive.ql.hooks.PostExecutePrinter` is always added to `hive.exec.post.hooks` in `hive-site.xml` so it has the same impact. Could you please double check?
> > 
> > 
> > No, i don't think it is always getting added in _hive-site.xml_. I checked in _hive-default.xml.template_ (which is created for new hive dist in conf dir) and also checked in local cluster, if _hive.exec.post.hooks_ key is not set in hive-site.xml then:
> > ```
> > 0: jdbc:hive2://localhost:10000 (default)> set hive.exec.post.hooks;
> > +------------------------------------+
> > |                set                 |
> > +------------------------------------+
> > | hive.exec.post.hooks is undefined  |
> > +------------------------------------+
> > ```
> 
> Yes. As I remember, PostExecutePrinter is added for qtests. That is the reason why you can see those "lineageinfo:" entries in the q.out files. They consume config files from the data/conf/... path. But they are not enabled by default.

I see. Thank you guys for looking after this. I think to keep the current behavior we should
* set the default value of `HIVE_LINEAGE_STATEMENT_FILTER` to `NONE` in the code - so in releases the lineage generation is kept turned off.
* set `HIVE_LINEAGE_STATEMENT_FILTER` to `ALL` in `hive-site.xml`-s of qtests where `org.apache.hadoop.hive.ql.hooks.PostExecutePrinter` is added. I think we should kept linage generation enabled in tests because we need the test coverage of this functionality.

Aggarwal-Raghav:
Sure @kasakrisz , thanks for suggestion. Will update the PR.

Aggarwal-Raghav:
@InvisibleProgrammer , post log4j bump to 2.24.3 change, I am seeing failure in UT in local (java.lang.NoSuchMethodError). Can you please confirm this because without the log4j, test is passing, attaching the stacktrace here:
command: 
```
mvn clean test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=lineage1.q -Dtest.output.overwrite=true -Pitests -pl itests/qtest
```
[Log4j_Stacktrace.txt](https://github.com/user-attachments/files/18945764/Log4j_Stacktrace.txt)


CC @zhangbutao

data/conf/hive-site.xml
```diff
@@ -154,6 +154,12 @@
   <description>Pre Execute Hook for Tests</description>
 </property>
 
+<property>
+  <name>hive.lineage.statement.filter</name>
+  <value>ALL</value>
+  <description>Lineage information filter</description>
```
kasakrisz, line 160: nit.
How about?
```
Specify the types of statements for which column lineage information is generated.
```

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -7303,6 +7304,13 @@ public Map<String, String> subtree(String string) {
     return ret;
   }
 
+  public static boolean shouldComputeLineage(HiveConf conf) {
```
InvisibleProgrammer, line 7307: Does it mean that we completely want to remove using POST_EXEC_HOOKS? It is a breaking change. Can we do this without a major release?

data/conf/hive-site.xml
Aggarwal-Raghav: ACK

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
Aggarwal-Raghav: Yes, @InvisibleProgrammer , it will be breaking change. I will create a doc for the same in hive-site website to help users.
 
My understanding is that next release should be major release as we have upgrade Hadoop version, Tez upgrade(not yet done but will be done before next release) and JDK17 (if possible) is planned.

InvisibleProgrammer (approved):

kasakrisz (approved):

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
kasakrisz: @Aggarwal-Raghav 
Are you planning document it here?
https://github.com/apache/hive-site

PR #5648, opened 2025-02-20
HIVE-28811: Iceberg: Partition-level column statistics support

deniskuzZ:

Add support for Iceberg's partition-level column statistics

better query planning

No

No

qtests / unit tests (TestHiveIcebergStatistics)


iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java
```diff
@@ -351,25 +384,5 @@ private void checkColStatMinMaxDistinctValue(String tableName, String colName, i
     // Check distinct
     Assert.assertEquals("distinct_count", rows.get(5)[0]);
     Assert.assertEquals(String.valueOf(distinct), rows.get(5)[1]);
-
-    shell.executeStatement("set hive.iceberg.stats.source=iceberg");
```
abstractdog, line 355: is hive.iceberg.stats.source=iceberg tested somewhere else from now?

deniskuzZ: i've added a parametrized config for that, see https://github.com/apache/hive/pull/5648/files/d645b25a042b57895156b38c792e967c598218a3#diff-85f53d8d84a6e67e63a21991e0a2fd0bd243cbb43a7fef18a791324b8c997dc7R52-R66

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerWithEngineBase.java
```diff
@@ -108,57 +108,46 @@ public abstract class HiveIcebergStorageHandlerWithEngineBase {
...
-            if (fileFormat != FileFormat.METADATA && "tez".equals(engine) && HiveVersion.min(HiveVersion.HIVE_3)) {
-              testParams.add(new Object[]{fileFormat, engine, TestTables.TestTableType.HIVE_CATALOG, true,
-                  formatVersion});
-            }
-          }
-        });
-      }
+      IntStream.of(2, 1).forEach(formatVersion -> {
+        testParams.add(new Object[]{fileFormat, HIVE_CATALOG, false, formatVersion});
+        // test for vectorization=ON in case of ORC and PARQUET format with Tez engine
```
abstractdog, line 119: nit: it can be maybe:
```
// test for vectorization=ON in case of ORC and PARQUET format
```
fortunately you removed this whole execution engine specific thing, which is cool, so we can remove it from comment too

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -1800,17 +1800,11 @@ public Table getTable(final String dbName, final String tableName, String tableM
...
-    } else {
-      String fullTableName = getFullTableName(dbName, tableName);
-      validWriteIdList = new ValidReaderWriteIdList(fullTableName, new long[0], new BitSet(), Long.MAX_VALUE);
-    }
-    return validWriteIdList;
+    long txnId = Optional.ofNullable(SessionState.get())
+      .map(ss -> ss.getTxnMgr().getCurrentTxnId()).orElse(0L);
+    
+    return (txnId > 0) ? AcidUtils.getTableValidWriteIdListWithTxnList(conf, dbName, tableName) : 
+        new ValidReaderWriteIdList();
```
abstractdog, line 1807: new ValidReaderWriteIdList() ends up with a tableName=null instead of fullTableName, isn't this changing some behavior?

ql/src/java/org/apache/hadoop/hive/ql/parse/TransformSpec.java
```diff
@@ -76,6 +90,28 @@ public String transformTypeString() {
     return transformType.name();
   }
 
+  public String toHiveExpr(Configuration conf, boolean alias) {
```
abstractdog, line 93: this method looks testable, also it's a string magic which is hard to read and understand, I would love to see a unit test with some test cases with input confs -> output string pairs

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -85,7 +85,7 @@
  **/
 public class BasicStatsNoJobTask implements IStatsProcessor {
 
-  private static transient final Logger LOG = LoggerFactory.getLogger(BasicStatsNoJobTask.class);
+  private static final Logger LOG = LoggerFactory.getLogger(BasicStatsNoJobTask.class);
```
abstractdog, line 88: good catch, I guess this rest be handled in a separate newbie patch: https://issues.apache.org/jira/browse/HIVE-28796

ql/src/java/org/apache/hadoop/hive/ql/parse/ColumnStatsSemanticAnalyzer.java
```diff
@@ -673,8 +672,8 @@ public ASTNode rewriteAST(ASTNode ast, ColumnStatsAutoGatherContext context)
     checkForPartitionColumns(colNames,
         Utilities.getColumnNamesFromFieldSchema(tbl.getPartitionKeys()));
     validateSpecifiedColumnNames(colNames);
-    if (conf.getBoolVar(ConfVars.HIVE_STATS_COLLECT_PART_LEVEL_STATS) && tbl.isPartitioned() 
-          && !tbl.hasNonNativePartitionSupport()) {
+    if (conf.getBoolVar(ConfVars.HIVE_STATS_COLLECT_PART_LEVEL_STATS) && tbl.isPartitioned()
```
abstractdog, line 675: this condition is used at least twice, does it make sense to put into a separate method?

ql/src/java/org/apache/hadoop/hive/ql/stats/ColStatsProcessor.java
```diff
@@ -226,9 +230,10 @@ public int persistColumnStats(Hive db, Table tbl) throws HiveException, MetaExce
         if (!(tbl.isMaterializedView() || tbl.isView() || tbl.isTemporary())) {
           setOrRemoveColumnStatsAccurateProperty(db, tbl, colStatDesc.getColName(), success);
         }
+      } else {
```
abstractdog, line 233: I can see this is fixed in [HIVE-27421](https://issues.apache.org/jira/browse/HIVE-27421)), and PR is told to be ready to merge, don't we want to fix this in that PR?

ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
```diff
@@ -1043,12 +1017,8 @@ public static List<ColStatistics> getTableColumnStats(
     }
     if (fetchColStats && !colStatsToRetrieve.isEmpty()) {
       try {
-        List<ColumnStatisticsObj> colStat;
-        if (table.isNonNative() && table.getStorageHandler().canProvideColStatistics(table)) {
-          colStat = table.getStorageHandler().getColStatistics(table);
-        } else {
-          colStat = Hive.get().getTableColumnStatistics(dbName, tabName, colStatsToRetrieve, false);
-        }
+        List<ColumnStatisticsObj> colStat = Hive.get().getTableColumnStatistics(
```
abstractdog, line 1020: it seems like  Hive.get().getTableColumnStatistics() is the final place where we decide whether to fetch col stats from metastore or fetch from storagehandler, which is fine if it's the final decision, but in this case this method's comment here is not accurate:
```
Get table level column statistics from metastore for needed columns
```
if we don't want to overthink, it's just:
```
Get table level column statistics for needed columns
```

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
deniskuzZ: if there is no txnId for an ACID table - that's wrong and we can return just a dummy

ql/src/java/org/apache/hadoop/hive/ql/stats/ColStatsProcessor.java
deniskuzZ: I wish it would be ready to merge, tests are failing. btw, I've reopened it, but I can't push amends to it

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerWithEngineBase.java
```diff
@@ -108,57 +108,46 @@ public abstract class HiveIcebergStorageHandlerWithEngineBase {
...
+      if (testTableType != HIVE_CATALOG) {
+        testParams.add(new Object[]{FileFormat.PARQUET, testTableType, false, 1});
       }
     }
 
     return testParams;
   }
 
   protected static TestHiveShell shell;
+  protected static String executionEngine = "tez";
```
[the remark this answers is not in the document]
deniskuzZ: didn't want to add big refactor here, but ok

ql/src/java/org/apache/hadoop/hive/ql/stats/ColStatsProcessor.java
deniskuzZ: btw, that PR has some weird logic, it calls `setColStatistics` twice
````
tbl.getStorageHandler().setColStatistics(tbl, colStats)
````

File: ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java (hunk 5 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java]: reverted from this PR

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveStorageHandler.java [hunk 10]
```diff
@@ -285,6 +286,11 @@ default List<ColumnStatisticsObj> getColStatistics(org.apache.hadoop.hive.ql.met
     return null;
   }
 
+  default AggrStats getAggrColStatsFor(org.apache.hadoop.hive.ql.metadata.Table hmsTable, List<String> colNames,
```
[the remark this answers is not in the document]
deniskuzZ: done

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerWithEngineBase.java
deniskuzZ: done

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerWithEngineBase.java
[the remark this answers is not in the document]
deniskuzZ: done

ql/src/java/org/apache/hadoop/hive/ql/parse/ColumnStatsSemanticAnalyzer.java
deniskuzZ: extracted to StatUtils

ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
deniskuzZ: done

ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
```diff
@@ -401,7 +375,7 @@ private static Statistics collectStatistics(HiveConf conf, PrunedPartitionList p
         // size is 0, aggrStats is null after several retries. Thus, we can
         // skip the step to connect to the metastore.
         if (fetchColStats && !neededColsToRetrieve.isEmpty() && !partNames.isEmpty()) {
-          aggrStats = Hive.get().getAggrColStatsFor(table.getDbName(), table.getTableName(),
+          aggrStats = Hive.get().getAggrColStatsFor(table.getDbName(), table.getTableName(),
```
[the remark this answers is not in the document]
deniskuzZ: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -612,40 +615,43 @@ private Table getTable(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
...
   }
 
-  private boolean writeColStats(ColumnStatistics tableColStats, Table tbl) {
+  private boolean writeColStats(List<ColumnStatistics> colStats, Table tbl) {
     try {
       if (!shouldRewriteColStats(tbl)) {
-        checkAndMergeColStats(tableColStats, tbl);
+        checkAndMergeColStats(colStats, tbl);
       }
       // Currently, we are only serializing table level stats.
```
difin, line 632: this comment seems to be outdated with these changes?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -722,8 +767,8 @@ public boolean canComputeQueryUsingStats(Partish partish) {
     if (hmsTable.getMetaTable() != null) {
       return false;
     }
-    if (!getStatsSource().equals(HiveMetaHook.ICEBERG) &&
-        StatsUtils.areBasicStatsUptoDateForQueryAnswering(partish.getTable(), partish.getPartParameters())) {
+    if (!getStatsSource().equals(HiveMetaHook.ICEBERG) && StatsSetupConst.areBasicStatsUptoDate(
+          partish.getPartParameters())) {
```
difin, line 771: nit: identation

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -2157,7 +2180,7 @@ private Partition getPartitionImpl(org.apache.hadoop.hive.ql.metadata.Table tabl
   public List<String> getPartitionNames(org.apache.hadoop.hive.ql.metadata.Table hmsTable,
       Map<String, String> partitionSpec) throws SemanticException {
     Table icebergTable = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
-    return IcebergTableUtil.getPartitionNames(icebergTable, partitionSpec, true);
+    return IcebergTableUtil.getPartitionNames(icebergTable, partitionSpec, false);
```
difin, line 2183: This method is called from a few locations, like `ShowPartitionsOperation`, does it work correctly with the change to include partitions for old specs? Should `ShowPartitionsOperation` show partitions for older specs?

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerTestUtils.java
```diff
@@ -126,6 +126,10 @@ static TestTables testTables(TestHiveShell shell, TestTables.TestTableType testT
     return testTableType.instance(shell.metastore().hiveConf(), temp, catalogName);
   }
 
+  static void init(TestHiveShell shell, TestTables testTables, TemporaryFolder temp) {
+    init(shell, testTables, temp, "tez");
```
difin, line 130: Should an enum be used here instead of hardcoded "tez" value?

difin: LGTM
Added a few comments and questions

difin (approved):

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: yes, [get/show]Partitions by default should return ALL table partitions

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandlerTestUtils.java
deniskuzZ: I don't think we have enum for that + I reused the same approach as in the test.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: iceberg modules have auto checkstyle validations including indentation checks, so it should be ok

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: removed

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -695,25 +700,64 @@ public List<ColumnStatisticsObj> getColStatistics(org.apache.hadoop.hive.ql.meta
...
-        .map(statsPath -> readColStats(table, statsPath))
+        .map(statsPath -> readColStats(table, statsPath, null).get(0))
         .orElse(emptyStats).getStatsObj();
     }
     return emptyStats.getStatsObj();
   }
 
-  private ColumnStatistics readColStats(Table table, Path statsPath) {
+  @Override
+  public AggrStats getAggrColStatsFor(org.apache.hadoop.hive.ql.metadata.Table hmsTable, List<String> colNames,
```
zratkai, line 710: I think some word is missing from the method name. 
For... ?

deniskuzZ: i've used existing API naming for aggr stats - `IMetaStoreClient#getAggrColStatsFor`

iceberg/iceberg-handler/src/test/queries/negative/show_partitions_negative_test.q
```diff
@@ -1,10 +0,0 @@
--- Mask the totalSize value as it can have slight variability, causing test flakiness
```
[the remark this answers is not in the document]
deniskuzZ: cause it's wrong: we should return an empty partition list for an empty table and not throw an exception. 
I've moved the test scenario into `show_partitions_test.q`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -491,23 +481,33 @@ public void appendFiles(org.apache.hadoop.hive.metastore.api.Table table, URI fr
 
   @Override
   public Map<String, String> getBasicStatistics(Partish partish) {
+    return getBasicStatistics(partish, false);
+  }
+
+  @SuppressWarnings("checkstyle:CyclomaticComplexity")
+  private Map<String, String> getBasicStatistics(Partish partish, boolean quickStats) {
```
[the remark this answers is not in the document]
deniskuzZ: `quickStats` is a part of stats that could be effortlessly computed:     

- NUM_FILES, 
- TOTAL_SIZE
- NUM_ERASURE_CODED_FILES

see  `MetaStoreServerUtils.populateQuickStats`

iceberg/iceberg-handler/src/test/results/positive/delete_iceberg_copy_on_write_partitioned.q.out
```diff
@@ -121,140 +121,140 @@ STAGE PLANS:
             Map Operator Tree:
                 TableScan
                   alias: tbl_ice
-                  Statistics: Num rows: 6 Data size: 582 Basic stats: COMPLETE Column stats: COMPLETE
+                  Statistics: Num rows: 6 Data size: 588 Basic stats: COMPLETE Column stats: PARTIAL
```
kasakrisz, line 124: What is the reason of the column stats loss?

ql/src/java/org/apache/hadoop/hive/ql/stats/ColStatsProcessor.java
```diff
@@ -141,14 +141,29 @@ private boolean constructColumnStatsFromPackedRows(Table tbl, List<ColumnStatist
...
-          List<FieldSchema> partColSchema = tbl.getPartCols();
+          List<FieldSchema> partColSchema = new ArrayList<>();
           List<String> partVals = new ArrayList<>();
-          // Iterate over partition columns to figure out partition name
-          for (int i = pos; i < pos + partColSchema.size(); i++) {
-            Object partVal = ((PrimitiveObjectInspector) fields.get(i).getFieldObjectInspector())
+          
+          if (tbl.hasNonNativePartitionSupport()) {
+            ObjectInspector inspector = fields.get(pos).getFieldObjectInspector();
+            if (inspector.getCategory() == ObjectInspector.Category.STRUCT) {
```
kasakrisz, line 149: Can `inspector.getCategory()` be anything else than `STRUCT`? What should happen if yes?

iceberg/iceberg-handler/src/test/results/positive/delete_iceberg_copy_on_write_partitioned.q.out
deniskuzZ: after the update/delete Hive doesn't recompute stats and on the next insert it only writes stats generated by that insert since prev stats is not accurate/no stats for prev snapshot. However, we had a bug and  did a fallback to HMS and used whatever was previously stored there (no snapshot id validation, similar to ACID writeIds)

ql/src/java/org/apache/hadoop/hive/ql/stats/ColStatsProcessor.java
deniskuzZ: it is always STRUCT, we compute stats per partition spec:
see `TransformSpec#toNamedStruct

ql/src/java/org/apache/hadoop/hive/ql/parse/ColumnStatsAutoGatherContext.java
```diff
@@ -122,16 +123,18 @@ public void insertAnalyzePipeline() throws SemanticException {
    */
   public void insertTableValuesAnalyzePipeline() throws SemanticException {
     // Instead of starting from analyze statement, we just generate the Select plan
-    boolean isPartitionStats = conf.getBoolVar(ConfVars.HIVE_STATS_COLLECT_PART_LEVEL_STATS) && tbl.isPartitioned();
+    boolean isPartitionStats = StatsUtils.isPartitionStats(tbl, conf);
```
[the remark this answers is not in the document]
deniskuzZ: no, we generate stats here, so we should check if we can set the stats

PR #5649, opened 2025-02-20
HIVE-28770: Fix Session Directories Persisting After HiveServer2 Shutdown

Indhumathi27:
Ensures session related directories is properly cleaned up when HiveServer2 (HS2) shuts down, by adding a shutdown hook to call close() method, which ensures cleaning session related directories.

Session related directories persist after HS2 shutdown, leading to disk bloat and performance degradation. This fix prevents unnecessary resource accumulation and improves cleanup.

No functional changes for users; it only enhances internal resource management. Temporary directories will now be cleaned up automatically.

No, this PR does not introduce any dependency upgrades.

Manually verified that session directories on /tmp and /tmp/{user_dir}/ are created and removed correctly.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
@@ -20,8 +20,6 @@ package org.apache.hadoop.hive.ql;
 
 import java.io.IOException;
 
-import org.apache.hadoop.hive.common.JavaUtils;
-import org.apache.hadoop.hive.common.ValidWriteIdList;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.ql.exec.Utilities;
 import org.apache.hadoop.hive.ql.hooks.HookContext;
@@ -92,6 +90,14 @@ public final class DriverUtils {
     }
   }
 
+  public static SessionState setUpSessionState(HiveConf conf) {
+    return setUpSessionState(conf, null);
+  }
+
+  public static SessionState setUpSessionState(HiveConf conf, String user) {
+    return setUpSessionState(conf, user, true);
+  }
+
   public static SessionState setUpSessionState(HiveConf conf, String user, boolean doStart) {
     SessionState sessionState = SessionState.get();
     if (sessionState == null) {
@@ -104,8 +110,7 @@ public final class DriverUtils {
         SessionState.start(sessionState);
       }
       SessionState.setCurrentSessionState(sessionState);
-    }
-    else {
+    } else {
       sessionState.setConf(conf);
     }
     return sessionState;
--- a/ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
@@ -90,11 +90,11 @@ public final class DriverUtils {
     }
   }
 
-  public static SessionState setUpSessionState(HiveConf conf) {
-    return setUpSessionState(conf, null);
+  public static SessionState setUpAndStartSessionState(HiveConf conf) {
+    return setUpAndStartSessionState(conf, null);
   }
 
-  public static SessionState setUpSessionState(HiveConf conf, String user) {
+  public static SessionState setUpAndStartSessionState(HiveConf conf, String user) {
     return setUpSessionState(conf, user, true);
   }
 
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java
@@ -19,6 +19,7 @@ package org.apache.hadoop.hive.ql.exec.tez;
 import java.io.IOException;
 import java.util.Map;
 
+import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.ql.DriverUtils;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.session.KillQuery;
@@ -34,6 +35,11 @@ import org.slf4j.LoggerFactory;
  */
 public class KillTriggerActionHandler implements TriggerActionHandler<TezSessionState> {
   private static final Logger LOG = LoggerFactory.getLogger(KillTriggerActionHandler.class);
+  private final HiveConf conf;
+
+  public KillTriggerActionHandler() {
+      this.conf = new HiveConf();
+  }
 
   @Override
   public void applyAction(final Map<TezSessionState, Trigger> queriesViolated) {
@@ -43,7 +49,7 @@ public class KillTriggerActionHandler implements TriggerActionHandler<TezSession
             String queryId = sessionState.getWmContext().getQueryId();
             try {
                 UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
-                DriverUtils.setUpSessionState(sessionState.getConf(), ugi.getShortUserName());
+                DriverUtils.setUpAndStartSessionState(conf, ugi.getShortUserName());
                 KillQuery killQuery = sessionState.getKillQuery();
                 // if kill query is null then session might have been released to pool or closed already
                 if (killQuery != null) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java
@@ -19,10 +19,10 @@ package org.apache.hadoop.hive.ql.exec.tez;
 import java.io.IOException;
 import java.util.Map;
 
-import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.ql.DriverUtils;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.session.KillQuery;
-import org.apache.hadoop.hive.ql.session.SessionState;
+import org.apache.hadoop.hive.ql.wm.Action;
 import org.apache.hadoop.hive.ql.wm.Trigger;
 import org.apache.hadoop.hive.ql.wm.TriggerActionHandler;
 import org.apache.hadoop.security.UserGroupInformation;
@@ -38,28 +38,24 @@ public class KillTriggerActionHandler implements TriggerActionHandler<TezSession
   @Override
   public void applyAction(final Map<TezSessionState, Trigger> queriesViolated) {
     for (Map.Entry<TezSessionState, Trigger> entry : queriesViolated.entrySet()) {
-      switch (entry.getValue().getAction().getType()) {
-      case KILL_QUERY:
-        TezSessionState sessionState = entry.getKey();
-        String queryId = sessionState.getWmContext().getQueryId();
-        try {
-          UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
-          SessionState ss = new SessionState(new HiveConf(), ugi.getShortUserName());
-          ss.setIsHiveServerQuery(true);
-          SessionState.start(ss);
-          KillQuery killQuery = sessionState.getKillQuery();
-          // if kill query is null then session might have been released to pool or closed already
-          if (killQuery != null) {
-            sessionState.getKillQuery().killQuery(queryId, entry.getValue().getViolationMsg(),
-                      sessionState.getConf());
-          }
-        } catch (HiveException|IOException e) {
-          LOG.warn("Unable to kill query {} for trigger violation", queryId);
+        if (entry.getValue().getAction().getType() == Action.Type.KILL_QUERY) {
+            TezSessionState sessionState = entry.getKey();
+            String queryId = sessionState.getWmContext().getQueryId();
+            try {
+                UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+                DriverUtils.setUpSessionState(sessionState.getConf(), ugi.getShortUserName());
+                KillQuery killQuery = sessionState.getKillQuery();
+                // if kill query is null then session might have been released to pool or closed already
+                if (killQuery != null) {
+                    sessionState.getKillQuery().killQuery(queryId, entry.getValue().getViolationMsg(),
+                            sessionState.getConf());
+                }
+            } catch (HiveException | IOException e) {
+                LOG.warn("Unable to kill query {} for trigger violation", queryId);
+            }
+        } else {
+            throw new RuntimeException("Unsupported action: " + entry.getValue());
         }
-        break;
-      default:
-        throw new RuntimeException("Unsupported action: " + entry.getValue());
-      }
     }
   }
 }
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
@@ -487,7 +487,7 @@ public class WorkloadManager extends TezSessionPoolSession.AbstractTriggerValida
             LOG.info("Invoking KillQuery for " + queryId + ": " + reason);
             try {
               UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
-              DriverUtils.setUpSessionState(conf, ugi.getShortUserName());
+              DriverUtils.setUpAndStartSessionState(conf, ugi.getShortUserName());
               kq.killQuery(queryId, reason, toKill.getConf());
               addKillQueryResult(toKill, true);
               killCtx.killSessionFuture.set(true);
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
@@ -67,6 +67,7 @@ import org.apache.hadoop.hive.metastore.api.WMFullResourcePlan;
 import org.apache.hadoop.hive.metastore.api.WMPool;
 import org.apache.hadoop.hive.metastore.api.WMPoolTrigger;
 import org.apache.hadoop.hive.metastore.api.WMTrigger;
+import org.apache.hadoop.hive.ql.DriverUtils;
 import org.apache.hadoop.hive.ql.exec.tez.AmPluginNode.AmPluginInfo;
 import org.apache.hadoop.hive.ql.exec.tez.TezSessionState.HiveResources;
 import org.apache.hadoop.hive.ql.exec.tez.UserPoolMapping.MappingInput;
@@ -486,9 +487,7 @@ public class WorkloadManager extends TezSessionPoolSession.AbstractTriggerValida
             LOG.info("Invoking KillQuery for " + queryId + ": " + reason);
             try {
               UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
-              SessionState ss = new SessionState(new HiveConf(), ugi.getShortUserName());
-              ss.setIsHiveServerQuery(true);
-              SessionState.start(ss);
+              DriverUtils.setUpSessionState(conf, ugi.getShortUserName());
               kq.killQuery(queryId, reason, toKill.getConf());
               addKillQueryResult(toKill, true);
               killCtx.killSessionFuture.set(true);
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
@@ -54,6 +54,7 @@ import org.apache.hadoop.hive.metastore.DefaultMetaStoreFilterHookImpl;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.ql.Context;
+import org.apache.hadoop.hive.ql.DriverUtils;
 import org.apache.hadoop.hive.ql.exec.ColumnInfo;
 import org.apache.hadoop.hive.ql.log.PerfLogger;
 import org.apache.hadoop.hive.ql.optimizer.calcite.HiveTypeSystemImpl;
@@ -167,12 +168,10 @@ public final class HiveMaterializedViewsRegistry {
 
     @Override
     public void run() {
-      SessionState ss = new SessionState(db.getConf());
-      ss.setIsHiveServerQuery(true); // All is served from HS2, we do not need e.g. Tez sessions
-      SessionState.start(ss);
       PerfLogger perfLogger = SessionState.getPerfLogger();
-      perfLogger.perfLogBegin(CLASS_NAME, PerfLogger.MATERIALIZED_VIEWS_REGISTRY_REFRESH);
       try {
+        DriverUtils.setUpSessionState(db.getConf());
+        perfLogger.perfLogBegin(CLASS_NAME, PerfLogger.MATERIALIZED_VIEWS_REGISTRY_REFRESH);
         if (initialized.get()) {
           for (Table mvTable : db.getAllMaterializedViewObjectsForRewriting()) {
             RelOptMaterialization existingMV = getRewritingMaterializedView(
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
@@ -170,7 +170,7 @@ public final class HiveMaterializedViewsRegistry {
     public void run() {
       PerfLogger perfLogger = SessionState.getPerfLogger();
       try {
-        DriverUtils.setUpSessionState(db.getConf());
+        DriverUtils.setUpAndStartSessionState(db.getConf());
         perfLogger.perfLogBegin(CLASS_NAME, PerfLogger.MATERIALIZED_VIEWS_REGISTRY_REFRESH);
         if (initialized.get()) {
           for (Table mvTable : db.getAllMaterializedViewObjectsForRewriting()) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
@@ -44,7 +44,12 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -341,6 +346,14 @@ public class SessionState implements ISessionAuthState {
    * It is required to exclude compaction related queries from all Ranger policies that would otherwise apply.
    */
   private boolean compaction = false;
+
+  // A thread-safe set to hold all active session states
+  private static final Set<SessionState> sessionStates = ConcurrentHashMap.newKeySet();
+
+  // Static block to add a single shutdown hook to clean up all session states
+  static {
+    ShutdownHookManager.addShutdownHook(SessionState::cleanUpAllSessionStates);
+  }
 
   public QueryState getQueryState(String queryId) {
     return queryStateMap.get(queryId);
@@ -669,8 +682,8 @@ public class SessionState implements ISessionAuthState {
   public static SessionState start(SessionState startSs) {
     start(startSs, false, null);
 
-    //add shutdown hook to clean up the session related directories and objects
-    startSs.addSessionStateShutdownHook();
+    // Register the session state in the centralized set
+    sessionStates.add(startSs);
 
     return startSs;
   }
@@ -1911,6 +1924,8 @@ public class SessionState implements ISessionAuthState {
   }
 
   public void close() throws IOException {
+    // de-register session state
+    sessionStates.remove(this);
     for (Closeable cleanupItem : cleanupItems) {
       try {
         cleanupItem.close();
@@ -1974,16 +1989,30 @@ public class SessionState implements ISessionAuthState {
     dynamicVars.clear();
   }
 
-  private void addSessionStateShutdownHook() {
-    // Added shutdown hook to close the session.
-    ShutdownHookManager.addShutdownHook(() -> {
+  private static void cleanUpAllSessionStates() {
+    ExecutorService cleanupExecutor = Executors.newCachedThreadPool();
+    try {
+      CompletableFuture<Void> allCleanupTasks = CompletableFuture.allOf(sessionStates.stream()
+              .map(sessionState -> CompletableFuture.runAsync(() -> {
+                  try {
+                      LOG.info("Closing session state: {}", sessionState.getSessionId());
+                      sessionState.close();
+                  } catch (IOException e) {
+                      throw new CompletionException(e);
+                  }
+              }, cleanupExecutor).exceptionally(e -> {
+                  LOG.error("Problem closing session state", e);
+                  return null;
+              })).toArray(CompletableFuture[]::new));
       try {
-        LOG.info("Closing session state: {}", getSessionId());
-        close();
-      } catch (IOException e) {
-        LOG.error("Problem closing session state", e);
+        allCleanupTasks.get(60, TimeUnit.SECONDS);
+      } catch (Exception e) {
+        LOG.error("Failed to close all session states", e);
       }
-    });
+    } finally {
+      // shutdown cleanup executor after all tasks finished/timeout
+      cleanupExecutor.shutdownNow();
+    }
   }
 
   private void clearReflectionUtilsCache() {
--- a/ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
@@ -110,6 +110,7 @@ import org.apache.hadoop.hive.shims.ShimLoader;
 import org.apache.hadoop.hive.shims.Utils;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.util.ReflectionUtils;
+import org.apache.hive.common.util.ShutdownHookManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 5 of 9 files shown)


ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
```diff
@@ -204,11 +205,21 @@ public void run() {
         } else {
           LOG.error("Problem connecting to the metastore when initializing the view registry", e);
         }
+      } finally {
+          cleanUpResources(ss);
```
okumin, line 209: I'm curious about what would happen if we closed `ss` using `SessionState#close`.

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
```diff
@@ -204,11 +205,21 @@ public void run() {
...
       }
       perfLogger.perfLogEnd(CLASS_NAME, PerfLogger.MATERIALIZED_VIEWS_REGISTRY_REFRESH);
     }
   }
 
+  // Method to clean up the downloaded resources
+  private void cleanUpResources(SessionState sessionState) {
+    if (sessionState != null) {
```
okumin, line 217: I believe this is always non-null.

service/src/java/org/apache/hive/service/cli/CLIService.java
```diff
@@ -133,6 +135,13 @@ private void applyAuthorizationConfigPolicy(HiveConf newHiveConf) throws HiveExc
     ss.setIsHiveServerQuery(true);
     SessionState.start(ss);
     ss.applyAuthorizationPolicy();
```
okumin, line 137: I slightly feel this `ss` can be immediately closed because this might be a hack to update `newHiveConf`

okumin (commented):
The issue is very well specified in the JIRA ticket. Thanks.
https://issues.apache.org/jira/browse/HIVE-28770

I asked you some questions because I'm still new to the parts that this PR changes. The intention and the fundamental approach looks good to me

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
```diff
@@ -204,6 +205,12 @@ public void run() {
         } else {
           LOG.error("Problem connecting to the metastore when initializing the view registry", e);
         }
+      } finally {
+          try {
+              ss.close();
```
deniskuzZ, line 210: could we make SessionState implement AutoCloseable and use try-with-resources?

service/src/java/org/apache/hive/service/cli/CLIService.java
```diff
@@ -126,13 +126,14 @@ public synchronized void init(HiveConf hiveConf) {
...
   private void applyAuthorizationConfigPolicy(HiveConf newHiveConf) throws HiveException,
-      MetaException {
+          MetaException, IOException {
     // authorization setup using SessionState should be revisited eventually, as
     // authorization and authentication are not session specific settings
     SessionState ss = new SessionState(newHiveConf);
     ss.setIsHiveServerQuery(true);
     SessionState.start(ss);
     ss.applyAuthorizationPolicy();
+    ss.close();
```
deniskuzZ, line 136: same
````
try (SessionState ss = new SessionState(newHiveConf)) {
...
}
````

Comment (deniskuzZ) [reply service/src/java/org/apache/hive/service/cli/CLIService.java]: any q test failures? maybe it is better to fix the tests and start session there?

Comment (deniskuzZ) [reply service/src/java/org/apache/hive/service/cli/CLIService.java]:
````  
/**
   * set current session to existing session object if a thread is running
   * multiple sessions - it must call this method with the new session object
   * when switching from one session to another.
   */
  public static SessionState start(SessionState startSs) {
    start(startSs, false, null);
    return startSs;
  }
````
It seems that In multiple places we just attach a new session object to the thread and don't close it. Is that true?

Comment (deniskuzZ) [reply service/src/java/org/apache/hive/service/cli/CLIService.java]: 1. DriverUtils#setUpSessionState (StatsUpdate#gatherStats, DriverUtils#runOnDriverInternal)
2. WorkloadManager#scheduleWork
3. KillTriggerActionHandler#applyAction
4. HiveMaterializedViewsRegistry.Loader#run
5. CLIService#applyAuthorizationConfigPolicy

Comment (deniskuzZ) [reply service/src/java/org/apache/hive/service/cli/CLIService.java]: @kasakrisz, do you know why in `HiveMaterializedViewsRegistry.Loader#run` we create a new SessionState object but never close it?

Comment (kasakrisz) [reply service/src/java/org/apache/hive/service/cli/CLIService.java]: `HiveMaterializedViewsRegistry.Loader` is a background thread for loading and compiling materialized views enabled for automatic query rewrite. The thread is started at HS2 start to inititalize the registry and it is also scheduled for periodic runs to update changes.
Seems that the only reason `SessionState.start` is called in this background thread is that we reach this point of code before it is called on the main thread at HS2 start up: https://issues.apache.org/jira/browse/HIVE-19691
The reason why a new `SessionState` instance is created is that we don't need TEZ session since no execution is happening in this thread: https://issues.apache.org/jira/browse/HIVE-21085

I'm thinking about whether we have to create new `SessionState` object at every refresh or it is required only at HS2 startup?

Anyway, I'm not aware of any reason cleaning of the `SessionState` was not added so I think it is a bug.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java [hunk 6]
```diff
@@ -486,15 +486,16 @@ private void scheduleWork(WmThreadSyncWork context) {
...
-              SessionState ss = new SessionState(new HiveConf(), ugi.getShortUserName());
-              ss.setIsHiveServerQuery(true);
-              SessionState.start(ss);
-              kq.killQuery(queryId, reason, toKill.getConf());
-              addKillQueryResult(toKill, true);
-              killCtx.killSessionFuture.set(true);
-              wmEvent.endEvent(toKill);
-              LOG.debug("Killed " + queryId);
-              return;
+                try (SessionState ss = new SessionState(new HiveConf(), ugi.getShortUserName())) {
```
deniskuzZ, line 489: looks legit to me. @abstractdog could you please confirm that's ok to close session object here?

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/StatsUpdater.java
```diff
@@ -91,8 +91,9 @@ public void gatherStats(CompactionInfo ci, HiveConf hiveConf,
             if (compactionQueueName != null && compactionQueueName.length() > 0) {
                 conf.set(TezConfiguration.TEZ_QUEUE_NAME, compactionQueueName);
             }
-            SessionState sessionState = DriverUtils.setUpSessionState(conf, userName, true);
-            DriverUtils.runOnDriver(conf, sessionState, sb.toString());
+            try (SessionState sessionState = DriverUtils.setUpSessionState(conf, userName, true)) {
```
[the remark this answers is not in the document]
deniskuzZ: I think so

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -108,19 +108,16 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
           VirtualColumn.FILE_PATH.getName(), fileSizePredicate == null ? "" : "and " + fileSizePredicate, orderBy);
     }
 
-    SessionState sessionState = setupQueryCompactionSession(conf, ci, tblProperties);
     String compactionTarget = "table " + HiveUtils.unparseIdentifier(compactTableName) +
-        (partSpec != null ? ", partition " + HiveUtils.unparseIdentifier(partSpec) : "");
+            (partSpec != null ? ", partition " + HiveUtils.unparseIdentifier(partSpec) : "");
 
-    try {
+    try (SessionState sessionState = setupQueryCompactionSession(conf, ci, tblProperties)) {
```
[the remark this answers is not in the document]
deniskuzZ: this -> https://github.com/apache/hive/pull/5649/files#r1993430862

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/QueryCompactor.java
```diff
@@ -153,7 +153,7 @@ void runCompactionQueries(HiveConf conf, String tmpTableName, CompactionInfo com
         LOG.error(ExceptionUtils.getStackTrace(e));
       } finally {
         //restore sessionState
-        sessionState.setCompaction(false);
+        sessionState.close();
```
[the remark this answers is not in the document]
deniskuzZ: yes, we shouldn't close

service/src/java/org/apache/hive/service/cli/CLIService.java
```diff
@@ -133,6 +134,15 @@ private void applyAuthorizationConfigPolicy(HiveConf newHiveConf) throws HiveExc
     ss.setIsHiveServerQuery(true);
     SessionState.start(ss);
     ss.applyAuthorizationPolicy();
+
+    // Add shutdown hook to close the session.
+    ShutdownHookManager.addShutdownHook(() -> {
```
abstractdog, line 139: does it make sense to introduce something like:
```
ss.closeOnExit()
```
which will be reusable on the SessionState object level

service/src/java/org/apache/hive/service/cli/CLIService.java
abstractdog: I agree that attaching a new session state to threadlocal without closing the previous one could be a problem, but creating + closing that every time properly might also be a problem from performance point of view (just taking a look at how much code resides in SesssionState.close), so wherever reusing a SessionState is possible, let's reuse, I guess having e.g. HiveMaterializedViewsRegistry as the outer scope we can store the SessionState as field reference and close it when HS2 is shut down

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
abstractdog: here, as I referred to in the case of the MV registry, what about trying a SessionState field in the workload manager, so instead of:
```
              UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
              SessionState ss = new SessionState(new HiveConf(), ugi.getShortUserName());
```
we can initialize the SessionState when the WorkloadManager is created, because neither the object returned by new HiveConf(), nor the UserGroupInformation.getCurrentUser() is supposed to change over time in HS2, it's safe to reuse SessionState

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -121,6 +121,7 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
       throw e;
     } finally {
       sessionState.setCompaction(false);
+      sessionState.closeOnExit();
```
deniskuzZ, line 124: shouldn't we set the hook when we create a session object?  i don't think that is the right place

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java
```diff
@@ -38,28 +40,27 @@ public class KillTriggerActionHandler implements TriggerActionHandler<TezSession
...
-          SessionState.start(ss);
-          KillQuery killQuery = sessionState.getKillQuery();
-          // if kill query is null then session might have been released to pool or closed already
-          if (killQuery != null) {
-            sessionState.getKillQuery().killQuery(queryId, entry.getValue().getViolationMsg(),
-                      sessionState.getConf());
-          }
-        } catch (HiveException|IOException e) {
-          LOG.warn("Unable to kill query {} for trigger violation", queryId);
+        if (Objects.requireNonNull(entry.getValue().getAction().getType()) == Action.Type.KILL_QUERY) {
```
deniskuzZ, line 43: why do you need `Objects.requireNonNull`?

````
if (Action.Type.KILL_QUERY == entry.getValue().getAction().getType()) {
..
}
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java [hunk 13]
```diff
@@ -38,28 +40,27 @@ public class KillTriggerActionHandler implements TriggerActionHandler<TezSession
...
+            TezSessionState sessionState = entry.getKey();
+            String queryId = sessionState.getWmContext().getQueryId();
+            try {
+                UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+                try (SessionState ss = new SessionState(new HiveConf(), ugi.getShortUserName())) {
```
deniskuzZ, line 48: similar to other comments, do we need to create and start new session per tez session entry?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
```diff
@@ -331,6 +334,7 @@ public void stop() throws Exception {
       stopTriggerValidator();
     }
     INSTANCE = null;
+    sessionState.close();
```
deniskuzZ, line 337: null check is missing

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/QueryCompactor.java
```diff
@@ -154,6 +154,7 @@ void runCompactionQueries(HiveConf conf, String tmpTableName, CompactionInfo com
       } finally {
         //restore sessionState
         sessionState.setCompaction(false);
+        sessionState.closeOnExit();
```
deniskuzZ, line 157: not required here

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/StatsUpdater.java
```diff
@@ -93,6 +93,7 @@ public void gatherStats(CompactionInfo ci, HiveConf hiveConf,
             }
             SessionState sessionState = DriverUtils.setUpSessionState(conf, userName, true);
             DriverUtils.runOnDriver(conf, sessionState, sb.toString());
+            sessionState.closeOnExit();
```
deniskuzZ, line 96: not required

deniskuzZ, line 96: i think it should. be done inside of `DriverUtils.setUpSessionState`

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
```diff
@@ -167,12 +168,11 @@ private Loader(Hive db) {
 
     @Override
     public void run() {
-      SessionState ss = new SessionState(db.getConf());
-      ss.setIsHiveServerQuery(true); // All is served from HS2, we do not need e.g. Tez sessions
-      SessionState.start(ss);
       PerfLogger perfLogger = SessionState.getPerfLogger();
-      perfLogger.perfLogBegin(CLASS_NAME, PerfLogger.MATERIALIZED_VIEWS_REGISTRY_REFRESH);
-      try {
+      try (SessionState ss = new SessionState(db.getConf())) {
```
deniskuzZ, line 172: should we use `DriverUtils.setUpSessionState(db.getConf(), null, true)`

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
```diff
@@ -203,12 +204,14 @@ public static WorkloadManager create(String yarnQueue, HiveConf conf, WMFullReso
...
   WorkloadManager(LlapPluginEndpointClientImpl amComm, String yarnQueue, HiveConf conf,
-      QueryAllocationManager qam, WMFullResourcePlan plan) throws ExecutionException, InterruptedException {
+      QueryAllocationManager qam, WMFullResourcePlan plan) throws ExecutionException, InterruptedException, IOException {
     this.yarnQueue = yarnQueue;
     this.conf = conf;
     this.totalQueryParallelism = determineQueryParallelism(plan);
     this.allocationManager = qam;
     this.allocationManager.setClusterChangedCallback(() -> notifyOfClusterStateChange());
+    UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+    this.sessionState = new SessionState(new HiveConf(), ugi.getShortUserName());
```
deniskuzZ, line 214:
````
DriverUtils.setUpSessionState(new HiveConf(), ugi.getShortUserName(), false)
````

Comment (deniskuzZ) [inline ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java:214]:
````
DriverUtils.setUpSessionState(new HiveConf(), ugi.getShortUserName(), false)
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java [hunk 19]
```diff
@@ -38,28 +37,23 @@ public class KillTriggerActionHandler implements TriggerActionHandler<TezSession
...
-            sessionState.getKillQuery().killQuery(queryId, entry.getValue().getViolationMsg(),
-                      sessionState.getConf());
-          }
-        } catch (HiveException|IOException e) {
-          LOG.warn("Unable to kill query {} for trigger violation", queryId);
+        if (entry.getValue().getAction().getType() == Action.Type.KILL_QUERY) {
+            TezSessionState sessionState = entry.getKey();
+            String queryId = sessionState.getWmContext().getQueryId();
+            try {
+                DriverUtils.setUpSessionState(sessionState.getConf(),true);
```
deniskuzZ, line 44: missing space between args

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
```diff
@@ -485,17 +487,15 @@ private void scheduleWork(WmThreadSyncWork context) {
             WmEvent wmEvent = new WmEvent(WmEvent.EventType.KILL);
             LOG.info("Invoking KillQuery for " + queryId + ": " + reason);
             try {
-              UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
-              SessionState ss = new SessionState(new HiveConf(), ugi.getShortUserName());
-              ss.setIsHiveServerQuery(true);
-              SessionState.start(ss);
+              sessionState.setIsHiveServerQuery(true);
```
deniskuzZ, line 490: not required, it's done inside `DriverUtils.setUpSessionState`

ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
```diff
@@ -92,6 +93,11 @@ private static void runOnDriverInternal(String query, HiveConf conf, SessionStat
     }
   }
 
+  public static SessionState setUpSessionState(HiveConf conf, boolean doStart) throws IOException {
```
deniskuzZ, line 96: i would refactor the method to accept the `Supplier<String> user`

ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
```diff
@@ -92,6 +91,11 @@ private static void runOnDriverInternal(String query, HiveConf conf, SessionStat
...
```
deniskuzZ, line 94: @Indhumathi27, it looks like `doStart` is always `true` here

ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
```diff
@@ -92,6 +91,11 @@ private static void runOnDriverInternal(String query, HiveConf conf, SessionStat
...
+    UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+    return setUpSessionState(conf, ugi.getUserName(), doStart);
```
deniskuzZ, line 96: should it be `ugi.getShortUserName`?

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
```diff
@@ -167,12 +169,10 @@ private Loader(Hive db) {
 
     @Override
     public void run() {
-      SessionState ss = new SessionState(db.getConf());
-      ss.setIsHiveServerQuery(true); // All is served from HS2, we do not need e.g. Tez sessions
-      SessionState.start(ss);
       PerfLogger perfLogger = SessionState.getPerfLogger();
-      perfLogger.perfLogBegin(CLASS_NAME, PerfLogger.MATERIALIZED_VIEWS_REGISTRY_REFRESH);
       try {
+        DriverUtils.setUpSessionState(db.getConf(), true);
```
deniskuzZ, line 174: should we pass `null` user here?

service/src/java/org/apache/hive/service/cli/CLIService.java
```diff
@@ -126,12 +127,10 @@ public synchronized void init(HiveConf hiveConf) {
...
 
   private void applyAuthorizationConfigPolicy(HiveConf newHiveConf) throws HiveException,
-      MetaException {
+      MetaException, IOException {
     // authorization setup using SessionState should be revisited eventually, as
     // authorization and authentication are not session specific settings
-    SessionState ss = new SessionState(newHiveConf);
-    ss.setIsHiveServerQuery(true);
-    SessionState.start(ss);
+    SessionState ss = DriverUtils.setUpSessionState(newHiveConf, true);
```
deniskuzZ, line 133: should we pass `null` user here?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
```diff
@@ -168,6 +168,7 @@ public class WorkloadManager extends TezSessionPoolSession.AbstractTriggerValida
   private final Thread delayedMoveThread;
   private final int delayedMoveTimeOutSec;
   private final int delayedMoveValidationIntervalSec;
+  private final SessionState sessionState;
```
deniskuzZ, line 171: why should it be an instance property?

deniskuzZ: it was good until the last force-push

deniskuzZ: please check https://github.com/apache/hive/pull/5649/files#r2000475143 as well

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
```diff
@@ -203,12 +203,13 @@ public static WorkloadManager create(String yarnQueue, HiveConf conf, WMFullReso
...
   @VisibleForTesting
   WorkloadManager(LlapPluginEndpointClientImpl amComm, String yarnQueue, HiveConf conf,
-      QueryAllocationManager qam, WMFullResourcePlan plan) throws ExecutionException, InterruptedException {
+      QueryAllocationManager qam, WMFullResourcePlan plan) throws ExecutionException, InterruptedException, IOException {
     this.yarnQueue = yarnQueue;
     this.conf = conf;
     this.totalQueryParallelism = determineQueryParallelism(plan);
     this.allocationManager = qam;
     this.allocationManager.setClusterChangedCallback(() -> notifyOfClusterStateChange());
+    DriverUtils.setUpSessionState(conf, true);
```
deniskuzZ, line 212: why did it migrate to the constructor?

ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
```diff
@@ -92,6 +90,14 @@ private static void runOnDriverInternal(String query, HiveConf conf, SessionStat
     }
   }
 
+  public static SessionState setUpSessionState(HiveConf conf) {
+    return setUpSessionState(conf, null, true);
```
deniskuzZ, line 94: ```return setUpSessionState(conf, null)```

ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
```diff
@@ -104,8 +110,7 @@ public static SessionState setUpSessionState(HiveConf conf, String user, boolean
         SessionState.start(sessionState);
       }
       SessionState.setCurrentSessionState(sessionState);
```
deniskuzZ, line 112: Do you know if that is required? i think we can drop it since it's set inside `start`

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -341,6 +345,29 @@ public enum AuthorizationMode{V1, V2};
    */
   private boolean compaction = false;
 
+  // A thread-safe set to hold all active session states
+  private static final Set<SessionState> sessionStates = Collections.newSetFromMap(new ConcurrentHashMap<>());
+
+  // Thread pool for concurrent cleanup
+  private static final ExecutorService cleanupExecutor = Executors.newCachedThreadPool();
```
deniskuzZ, line 352: Should it be a SingleThreadExecutor?

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -341,6 +345,29 @@ public enum AuthorizationMode{V1, V2};
...
+
+  // Static block to add a single shutdown hook for all session states
+  static {
+    ShutdownHookManager.addShutdownHook(new Thread(() -> {
```
deniskuzZ, line 356: can't we pass Runnable? why are you creating a new Thread?

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -341,6 +345,29 @@ public enum AuthorizationMode{V1, V2};
...
+      try {
+        cleanUpAllSessions();
```
deniskuzZ, line 358: why don't we create and destroy the threadpool inside of `cleanUpAllSessions`?

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -1969,6 +2001,24 @@ public void close() throws IOException {
...
   }
 
+  private static void deregister(SessionState ss) {
+    // remove from the sessionStates list to avoid cleaning the closed session state again
+    sessionStates.remove(ss);
+  }
+
+  private static void cleanUpAllSessions() {
+    for (SessionState sessionState : sessionStates) {
+      cleanupExecutor.submit(() -> {
```
deniskuzZ, line 2011: where do you wait for completion?

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -1969,6 +2001,24 @@ public void close() throws IOException {
     dynamicVars.clear();
   }
 
+  private static void deregister(SessionState ss) {
```
deniskuzZ, line 2004: why do you need this 1-liner method, just call `sessionStates.remove(this);`

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java
```diff
@@ -38,28 +38,24 @@ public class KillTriggerActionHandler implements TriggerActionHandler<TezSession
...
-                      sessionState.getConf());
-          }
-        } catch (HiveException|IOException e) {
-          LOG.warn("Unable to kill query {} for trigger violation", queryId);
+        if (entry.getValue().getAction().getType() == Action.Type.KILL_QUERY) {
+            TezSessionState sessionState = entry.getKey();
+            String queryId = sessionState.getWmContext().getQueryId();
+            try {
+                UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+                DriverUtils.setUpSessionState(sessionState.getConf(), ugi.getShortUserName());
```
deniskuzZ, line 46: we were passing `new HiveConf()` before

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/WorkloadManager.java
```diff
@@ -486,9 +487,7 @@ private void scheduleWork(WmThreadSyncWork context) {
             LOG.info("Invoking KillQuery for " + queryId + ": " + reason);
             try {
               UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
-              SessionState ss = new SessionState(new HiveConf(), ugi.getShortUserName());
-              ss.setIsHiveServerQuery(true);
-              SessionState.start(ss);
+              DriverUtils.setUpSessionState(conf, ugi.getShortUserName());
```
deniskuzZ, line 490: same, new HiveConf() was passed before

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java
deniskuzZ, line 46: Should we pass the `conf` to KillTriggerActionHandler constructor from TezSessionPoolManager?

deniskuzZ: Before, when we created a new session, we passed a `new HiveConf`, but now that might overwrite the current session conf. Note that in `WorkloadManager` you are using global conf and not the `TezSessionState`

deniskuzZ: What about `WorkloadManager`? does it use a separate thread to do it's job as well?

ql/src/java/org/apache/hadoop/hive/ql/DriverUtils.java
```diff
@@ -104,8 +110,7 @@ public static SessionState setUpSessionState(HiveConf conf, String user, boolean
...
-    }
-    else {
+    } else {
```
deniskuzZ, line 113: @difin, why did you add the `else` branch that overrides SessionState conf?

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -1906,6 +1923,7 @@ public void setCurrentCatalog(String currentCatalog) {
   }
 
   public void close() throws IOException {
+    sessionStates.remove(this);
```
deniskuzZ, line 1926: Please add new line between and maybe add a comment "Remove session state from the registry"

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -1969,6 +1987,38 @@ public void close() throws IOException {
...
+                  LOG.error("Error closing session state", e);
+                  return null;
+              })).toArray(CompletableFuture[]::new));
+      try {
+        allCleanupTasks.get(60, TimeUnit.SECONDS);
+      } catch (Exception e) {
+        LOG.error("Failed to close all session states", e);
+      }
+    } finally {
+      cleanupExecutor.shutdown();
```
deniskuzZ, line 2011: why do you need a graceful shutdown if at this point all tasks either completed successfully or timed out?

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -1969,6 +1987,38 @@ public void close() throws IOException {
...
+  private static void cleanUpAllSessionStates() {
+    ExecutorService cleanupExecutor = Executors.newCachedThreadPool();
+    try {
+      CompletableFuture<Void> allCleanupTasks = CompletableFuture.allOf(sessionStates.stream()
+              .map(sessionState -> CompletableFuture.runAsync(() -> {
+                  try {
+                      LOG.info("Closing session state: {}", sessionState.getSessionId());
+                      sessionState.close();
+                  } catch (IOException e) {
+                      LOG.error("Problem closing session state", e);
```
deniskuzZ, line 1999: `throw new RuntimeException(e)` here to handle in the exceptionally

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/KillTriggerActionHandler.java
```diff
@@ -38,28 +39,24 @@ public class KillTriggerActionHandler implements TriggerActionHandler<TezSession
...
-                      sessionState.getConf());
-          }
-        } catch (HiveException|IOException e) {
-          LOG.warn("Unable to kill query {} for trigger violation", queryId);
+        if (entry.getValue().getAction().getType() == Action.Type.KILL_QUERY) {
+            TezSessionState sessionState = entry.getKey();
+            String queryId = sessionState.getWmContext().getQueryId();
+            try {
+                UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+                DriverUtils.setUpSessionState(new HiveConf(), ugi.getShortUserName());
```
deniskuzZ, line 47: could we move it to an instance var and initialise in the constructor?

deniskuzZ (commented):
LGTM
@kasakrisz, @abstractdog can we merge?

ql/src/java/org/apache/hadoop/hive/ql/session/SessionState.java
```diff
@@ -1969,6 +1988,32 @@ public void close() throws IOException {
...
+  private static void cleanUpAllSessionStates() {
+    ExecutorService cleanupExecutor = Executors.newCachedThreadPool();
+    try {
+      CompletableFuture<Void> allCleanupTasks = CompletableFuture.allOf(sessionStates.stream()
+              .map(sessionState -> CompletableFuture.runAsync(() -> {
+                  try {
+                      LOG.info("Closing session state: {}", sessionState.getSessionId());
+                      sessionState.close();
+                  } catch (IOException e) {
+                      throw new RuntimeException(e);
```
deniskuzZ, line 2000: minor: @Indhumathi27, seems that `CompletionException` is recommended here instead of RuntimeException.

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMaterializedViewsRegistry.java
```diff
@@ -167,12 +168,10 @@ private Loader(Hive db) {
 
     @Override
     public void run() {
-      SessionState ss = new SessionState(db.getConf());
-      ss.setIsHiveServerQuery(true); // All is served from HS2, we do not need e.g. Tez sessions
-      SessionState.start(ss);
       PerfLogger perfLogger = SessionState.getPerfLogger();
-      perfLogger.perfLogBegin(CLASS_NAME, PerfLogger.MATERIALIZED_VIEWS_REGISTRY_REFRESH);
       try {
+        DriverUtils.setUpSessionState(db.getConf());
```
kasakrisz, line 173: nit: This refactor doesn't change the existing behavior. However the method `setUpSessionState` hides that we not just setup the SessionState but also start it. IMHO the original code emphasised this better.

A method name like `setUpAndStartSessionState` would also be better. 
WDYT?

deniskuzZ: @Indhumathi27, please do not rename the original method with `doStart` param, others 2 are fine

PR #5652, opened 2025-02-21
HIVE-28775: HiveServer2: introduce a new HA Health Check endpoint on a different p…

difin:
…ort than WebUI

Changes to deploy a new endpoint for HA health checking that listens on a different port than WebUI port on HS2 webserver.

Leader endpoint is used for HA health checking on downstream. When LDAP authentication is enabled on HS2, it requires to call leader endpoint with LDAP also. If we wanted to create a technical user for this endpoint we would need to have customers add this user into their LDAP servers which we have no control over, so this is not a good solution. Instead of that this PR proposes a new HA health check endpoint that exposes only the new route '\GET /health-ha/leader' on a different port that can be accessed without LDAP.

Yes, new http endpoint that listens on a different port.

No.

Execution of existing precommit tests + new HA tests.


dengzhhu653:
how about skipping the authentication for `/leader` or `/peers` in `LDAPAuthenticationFilter`?

difin:
> how about skipping the authentication for `/leader` or `/peers` in `LDAPAuthenticationFilter`?

If we do that and Hive is deployed on a public cloud, it will allow access to the /leader endpoints to the internet, which is not secure. By adding a new unsecured endpoint for HA health-check on a different port, that new port can be exposed internally only for health-check script which will be more secure.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -3979,6 +3979,8 @@ public static enum ConfVars {
     HIVE_SERVER2_WEBUI_BIND_HOST("hive.server2.webui.host", "0.0.0.0", "The host address the HiveServer2 WebUI will listen on"),
     HIVE_SERVER2_WEBUI_PORT("hive.server2.webui.port", 10002, "The port the HiveServer2 WebUI will listen on. This can be"
         + "set to 0 or a negative integer to disable the web UI"),
+    HIVE_SERVER2_HEALTH_HA_PORT("hive.server2.health.ha.port", 11002, "The port the HiveServer2 health-ha web app will listen on. This can be"
```
deniskuzZ, line 3982: why do we need `HA` in the naming?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +455,101 @@ public synchronized void init(HiveConf hiveConf) {
...
+  
+  private void addHAContextAttributes(HttpServer.Builder builder, HiveConf hiveConf) {
+    builder.setContextAttribute("hs2.isLeader", isLeader);
+    builder.setContextAttribute("hs2.failover.callback", new FailoverHandlerCallback(hs2HARegistry));
+    builder.setContextAttribute("hiveconf", hiveConf);
+  }
+  
+  private HttpServer.Builder initBuilder(String webHost, int port, String name, String contextPath, HiveConf hiveConf) throws IOException {
+    HttpServer.Builder builder = new HttpServer.Builder(name);
+    builder.setPort(port).setConf(hiveConf);
```
deniskuzZ, line 467: can we separate the `port` and `hiveConf` setup?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -489,13 +496,13 @@ static boolean userHasAdministratorAccess(ServletContext servletContext,
...
-  WebAppContext createWebAppContext(Builder b) {
+  WebAppContext createWebAppContext(Builder b) throws FileNotFoundException {
     WebAppContext ctx = new WebAppContext();
     setContextAttributes(ctx.getServletContext(), b.contextAttrs);
     ctx.getServletContext().getSessionCookieConfig().setHttpOnly(true);
     ctx.setDisplayName(b.name);
-    ctx.setContextPath("/");
-    ctx.setWar(appDir + "/" + b.name);
+    ctx.setContextPath(b.contextPath == null ? "/" : b.contextPath);
+    ctx.setWar(getWebAppsPath(b.name) + "/" + b.name);
```
deniskuzZ, line 505: do we need this?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -534,9 +542,110 @@ private void setupCORSFilter(Builder b) {
...
+    WebAppContext webAppContext = createWebAppContext(builder);
+    initWebAppContext(builder, webAppContext);
+    RewriteHandler rwHandler = createRewriteHandler(builder, webAppContext);
+    
+    ContextHandlerCollection portHandler = new ContextHandlerCollection();
+    ServerConnector connector = addChannelConnector(threadPool.getQueueSize(), builder);
+    portHandler.addHandler(rwHandler);
+
+    for (Pair<String, Class<? extends HttpServlet>> p : builder.servlets) {
+      addServlet(p.getFirst(), "/" + p.getFirst(), p.getSecond(), webAppContext);
```
deniskuzZ, line 562: could we use `getKey/Value` Pair method signature?

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: That existed in the original code which I refactored into another method.
Done.

common/src/java/org/apache/hive/http/HttpServer.java
difin: `ctx.setContextPath(b.contextPath == null ? "/" : b.contextPath);` is not needed, refactored to `ctx.setContextPath(b.contextPath)` because it it always set. This is needed because 2 web apps cannot have the same context path on the same web server. Health check web app has a different context path than WebUI.

`ctx.setWar(getWebAppsPath(b.name) + "/" + b.name);` is needed because each web app has a different app dir. See `service/target/hive-webapps/ `

common/src/java/org/apache/hive/http/HttpServer.java
difin: That was in the original code, fixed.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -3979,6 +3979,7 @@ public static enum ConfVars {
     HIVE_SERVER2_WEBUI_BIND_HOST("hive.server2.webui.host", "0.0.0.0", "The host address the HiveServer2 WebUI will listen on"),
     HIVE_SERVER2_WEBUI_PORT("hive.server2.webui.port", 10002, "The port the HiveServer2 WebUI will listen on. This can be"
         + "set to 0 or a negative integer to disable the web UI"),
+    HIVE_SERVER2_HA_HEALTHCHECK_PORT("hive.server2.ha.healthcheck.port", 11002, "The port the HiveServer2 health-ha web app will listen on"),
```
deniskuzZ, line 3982: Could we move this to the rest of HA configs, i.e `HIVE_SERVER2_ACTIVE_PASSIVE_HA_REGISTRY_NAMESPACE`

itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestActivePassiveHA.java
```diff
@@ -139,6 +139,72 @@ private static void setHAConfigs(Configuration conf) {
     conf.setInt(ConfVars.HIVE_ZOOKEEPER_CONNECTION_MAX_RETRIES.varname, 1);
   }
 
+  @Test(timeout = 60000)
+  public void testHealthCheck() throws Exception {
+    String instanceId1 = UUID.randomUUID().toString();
+    miniHS2_1.start(getConfOverlay(instanceId1));
```
deniskuzZ, line 145: do we need to start miniHS2 to do the port checks?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -402,90 +403,9 @@ public synchronized void init(HiveConf hiveConf) {
...
-              builder.setPAMAuthenticator(pamAuthenticator == null ? new PamAuthenticator(hiveConf) : pamAuthenticator);
-              builder.setUsePAM(true);
-            } else if (hiveConf.getBoolVar(ConfVars.HIVE_IN_TEST)) {
-              builder.setPAMAuthenticator(pamAuthenticator == null ? new PamAuthenticator(hiveConf) : pamAuthenticator);
-              builder.setUsePAM(true);
-            } else {
-              throw new IllegalArgumentException(ConfVars.HIVE_SERVER2_WEBUI_USE_SSL.varname + " has false value. It is recommended to set to true when PAM is used.");
-            }
-          }
+          HttpServer.Builder builder = initBuilder(webHost, webUIPort, "hiveserver2", "/", hiveConf);
```
deniskuzZ, line 406: Why do you need to pass constants ("hiveserver2", "/") to a private builder method?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +455,102 @@ public synchronized void init(HiveConf hiveConf) {
...
     ShutdownHookManager.addGracefulShutDownHook(() -> graceful_stop(),  timeout == 0 ? 30 : timeout);
   }
+  
+  private void addHAContextAttributes(HttpServer.Builder builder, HiveConf hiveConf) {
+    builder.setContextAttribute("hs2.isLeader", isLeader);
+    builder.setContextAttribute("hs2.failover.callback", new FailoverHandlerCallback(hs2HARegistry));
+    builder.setContextAttribute("hiveconf", hiveConf);
+  }
+  
+  private HttpServer.Builder initBuilder(String webHost, int port, String name, String contextPath, HiveConf hiveConf) throws IOException {
```
deniskuzZ, line 465: maybe `newHttpServerBuilder`?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +455,102 @@ public synchronized void init(HiveConf hiveConf) {
     // Extra time for releasing the resources if timeout sets to 0
     ShutdownHookManager.addGracefulShutDownHook(() -> graceful_stop(),  timeout == 0 ? 30 : timeout);
   }
+  
+  private void addHAContextAttributes(HttpServer.Builder builder, HiveConf hiveConf) {
```
deniskuzZ, line 459: `static`?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +455,102 @@ public synchronized void init(HiveConf hiveConf) {
...
+    builder.setContextAttribute("hs2.isLeader", isLeader);
+    builder.setContextAttribute("hs2.failover.callback", new FailoverHandlerCallback(hs2HARegistry));
+    builder.setContextAttribute("hiveconf", hiveConf);
+  }
+  
+  private HttpServer.Builder initBuilder(String webHost, int port, String name, String contextPath, HiveConf hiveConf) throws IOException {
+    HttpServer.Builder builder = new HttpServer.Builder(name);
+    builder.setPort(port);
+    builder.setConf(hiveConf);
```
deniskuzZ, line 468: wonder why we need both 
- `builder.setContextAttribute("hiveconf", hiveConf)`
- `builder.setConf(hiveConf)`?

deniskuzZ, line 468: could we move `setConf` after `setHost`

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +455,102 @@ public synchronized void init(HiveConf hiveConf) {
...
+        builder.setUsePAM(true);
+      } else {
+        throw new IllegalArgumentException(ConfVars.HIVE_SERVER2_WEBUI_USE_SSL.varname + " has false value. It is recommended to set to true when PAM is used.");
+      }
+    }
+    
+    return builder;
+  }
+  
+  private void initHAHealthChecker(HttpServer webServer, HiveConf hiveConf) throws IOException {
```
deniskuzZ, line 544: static

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +455,102 @@ public synchronized void init(HiveConf hiveConf) {
...
+    if (serviceDiscovery && activePassiveHA) {
+      String webHost = hiveConf.getVar(ConfVars.HIVE_SERVER2_WEBUI_BIND_HOST);
+      int healthCheckPort = hiveConf.getIntVar(ConfVars.HIVE_SERVER2_HA_HEALTHCHECK_PORT);
+      HttpServer.Builder builder = initBuilder(webHost, healthCheckPort, "health-ha", "/health-ha", hiveConf);
+      addHAContextAttributes(builder, hiveConf);
+      builder.addServlet("leader", HS2HAHealthChecker.class);
```
deniskuzZ, line 550: don't we already have a `leader` endpoint?
````
builder.addServlet("leader", HS2LeadershipStatus.class);
````

File: service/src/java/org/apache/hive/service/servlet/HS2HAHealthChecker.java [hunk 13]
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.hive.service.servlet;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class HS2HAHealthChecker extends HS2LeadershipStatus {
+
+  @Override
+  public void doDelete(final HttpServletRequest request, final HttpServletResponse response) throws IOException {
```
deniskuzZ, line 28: why do we need this?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -534,9 +542,110 @@ private void setupCORSFilter(Builder b) {
...
+   * @return ContextHandlerCollection - A collection of request handlers associated with the new port connector,
+   *         which includes the newly initialized web application.
+   */
+  public ContextHandlerCollection addWebApp(Builder builder) throws IOException {
+    WebAppContext webAppContext = createWebAppContext(builder);
+    initWebAppContext(builder, webAppContext);
+    RewriteHandler rwHandler = createRewriteHandler(builder, webAppContext);
+    
+    ContextHandlerCollection portHandler = new ContextHandlerCollection();
+    ServerConnector connector = addChannelConnector(threadPool.getQueueSize(), builder);
```
deniskuzZ, line 558: should we move `connector` closer to it's actual use?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -534,9 +542,110 @@ private void setupCORSFilter(Builder b) {
...
+    rwHandler.setHandler(webAppContext);
+    
+    return rwHandler;
+  }
+  
+  /**
+   * Create a channel connector for "http/https" requests and add it to the server
    */
-  Connector createChannelConnector(int queueSize, Builder b) {
+  ServerConnector addChannelConnector(int queueSize, Builder b) {
```
deniskuzZ, line 648: maybe `createAndAddChannelConnector`?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -135,9 +136,10 @@ public class HttpServer {
 
 
   private final String name;
-  private String appDir;
-  private WebAppContext webAppContext;
+  private List<WebAppContext> webAppContexts;
   private Server webServer;
+  private QueuedThreadPool threadPool;
```
deniskuzZ, line 141: why do you need this? can't you just use `builder#maxThreads`?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -604,108 +715,65 @@ void setContextAttributes(Context ctx, Map<String, Object> contextAttrs) {
...
-    // Configure web application contexts for the web server
-    ContextHandlerCollection contexts = new ContextHandlerCollection();
-    contexts.addHandler(rwHandler);
-    webServer.setHandler(contexts);
+    // Configure the global context handler
+    portHandlerWrapper = new PortHandlerWrapper();
+    webServer.setHandler(portHandlerWrapper);
 
+    // Configure the web server connector and port handler to listen on
+    ContextHandlerCollection portHandler = addWebApp(b);
```
deniskuzZ, line 742: should we do this closer to actual ref usage?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -808,14 +876,23 @@ public void addServlet(String name, String pathSpec,
     if (name != null) {
       holder.setName(name);
     }
+    webAppContexts.get(0).addServlet(holder, pathSpec);
```
deniskuzZ, line 879: can we reuse `private addServlet` here?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -808,14 +876,23 @@ public void addServlet(String name, String pathSpec,
...
+    }
     webAppContext.addServlet(holder, pathSpec);
   }
 
   public void addServlet(String name, String pathSpec, ServletHolder holder) {
     if (name != null) {
       holder.setName(name);
     }
-    webAppContext.addServlet(holder, pathSpec);
+    webAppContexts.get(0).addServlet(holder, pathSpec);
```
deniskuzZ, line 895: same - 'reuse'

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
difin: Done, and also renamed the new config to HIVE_SERVER2_ACTIVE_PASSIVE_HA_HEALTHCHECK_PORT

itests/hive-unit/src/test/java/org/apache/hive/jdbc/TestActivePassiveHA.java
difin: Yes, the test needs to bring up HS2 in Active-Passive-HA mode and check that the health check endpoint can be reached only via the dedicated port and not via the WebUI port. Other Active-Passive-HA tests also use MiniHS2

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: Added 'static'.
'new' in the name seems a little bit weird to me, are you ok with name 'createHttpServerBuilder'?

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: The method uses 2 HS2 member variables: isLeader and hs2HARegistry.
What is the benefit of making this method static?

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: This was an existing code.

"hiveconf" was used in HS2Peers servlet. I refactored it to use `HttpServer.CONF_CONTEXT_ATTRIBUTE` that is set by HttpServer.Builder.

difin: Done

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +455,102 @@ public synchronized void init(HiveConf hiveConf) {
...
+            ConfVars.HIVE_SERVER2_WEBUI_CORS_ALLOWED_HEADERS.varname + "/" +
+            " is not configured");
+      }
+      builder.setAllowedOrigins(allowedOrigins);
+      builder.setAllowedMethods(allowedMethods);
+      builder.setAllowedHeaders(allowedHeaders);
+      LOG.info("CORS enabled - allowed-origins: {} allowed-methods: {} allowed-headers: {}", allowedOrigins,
+          allowedMethods, allowedHeaders);
+    }
+    if(hiveConf.getBoolVar(ConfVars.HIVE_SERVER2_WEBUI_XFRAME_ENABLED)){
```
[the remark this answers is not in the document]
difin: Done

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +455,102 @@ public synchronized void init(HiveConf hiveConf) {
...
+    }
+    if (hiveConf.getBoolVar(ConfVars.HIVE_SERVER2_WEBUI_USE_PAM)) {
+      if (hiveConf.getBoolVar(ConfVars.HIVE_SERVER2_WEBUI_USE_SSL)) {
+        String hiveServer2PamServices = hiveConf.getVar(ConfVars.HIVE_SERVER2_PAM_SERVICES);
+        if (hiveServer2PamServices == null || hiveServer2PamServices.isEmpty()) {
+          throw new IllegalArgumentException(ConfVars.HIVE_SERVER2_PAM_SERVICES.varname + " are not configured.");
+        }
+        builder.setPAMAuthenticator(pamAuthenticator == null ? new PamAuthenticator(hiveConf) : pamAuthenticator);
+        builder.setUsePAM(true);
+      } else if (hiveConf.getBoolVar(ConfVars.HIVE_IN_TEST)) {
```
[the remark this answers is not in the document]
difin: This is an existing code.
It allows to use a `pamAuthenticator` provided by a calling test and use PAM authentication without SSL.
What should be used instead of `HIVE_IN_TEST`?

service/src/java/org/apache/hive/service/server/HiveServer2.java
difin: If we make it static it will be needed to pass 6 more variables to this method which are not unique for the health checker.

service/src/java/org/apache/hive/service/servlet/HS2HAHealthChecker.java
difin: HS2LeadershipStatus has 2 functions: 
1. Leader status (GET route)
2. Failover initiation (DELETE route)

Health check endpoint is supposed to allow only getting the leadership status. This is what is used for health checking.

The new health check endpoint is going to be not secured by LDAP and it meant for health checking only.
This is needed to disable the ability to initiate failover via health check route.

common/src/java/org/apache/hive/http/HttpServer.java
difin: Done

common/src/java/org/apache/hive/http/HttpServer.java
difin: Done

common/src/java/org/apache/hive/http/HttpServer.java
difin: `builder#maxThreads` is not always defined:

```
    threadPool = new QueuedThreadPool();
    if (b.maxThreads > 0) {
      threadPool.setMaxThreads(b.maxThreads);
    }
```

common/src/java/org/apache/hive/http/HttpServer.java
difin: I could move it only a few lines below and added a comment:

```
    // Configures the web server connector and port handler to listen on
    // Also creates and adds the web application context to the server to which the servlets will be added
    ContextHandlerCollection portHandler = createAndAddWebApp(b);

    addServlet("jmx", "/jmx", JMXJsonServlet.class);
    addServlet("conf", "/conf", ConfServlet.class);
    addServlet("stacks", "/stacks", StackServlet.class);
    addServlet("conflog", "/conflog", Log4j2ConfiguratorServlet.class);
```

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
deniskuzZ: i don't think we need `ACTIVE_PASSIVE`, it's too long and redundant

service/src/java/org/apache/hive/service/server/HiveServer2.java
deniskuzZ: It's hard to spot instance variables during review.

service/src/java/org/apache/hive/service/server/HiveServer2.java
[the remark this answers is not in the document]
deniskuzZ: Mocking techniques. We shouldn't add test code hacks into the production code. 
Since it was there before, it shouldn't be a blocker for PR, but if it's a simple/small refactor, maybe we could try to drop

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -534,6 +456,103 @@ public synchronized void init(HiveConf hiveConf) {
...
+  }
+  
+  private void initHAHealthChecker(HttpServer webServer, HiveConf hiveConf) throws IOException {
+    if (serviceDiscovery && activePassiveHA) {
+      String webHost = hiveConf.getVar(ConfVars.HIVE_SERVER2_WEBUI_BIND_HOST);
+      int healthCheckPort = hiveConf.getIntVar(ConfVars.HIVE_SERVER2_ACTIVE_PASSIVE_HA_HEALTHCHECK_PORT);
+      HttpServer.Builder builder = createHttpServerBuilder(webHost, healthCheckPort, "health-ha", 
+          "/health-ha", hiveConf, cliService, pamAuthenticator);
+      addHAContextAttributes(builder, hiveConf);
+      builder.addServlet("leader", HS2HAHealthChecker.class);
```
deniskuzZ, line 552: why do we use "leader" as the name for `endpoint`? should it be "ha-healthcheck"?

service/src/java/org/apache/hive/service/servlet/HS2HAHealthChecker.java
deniskuzZ: it's a bit confusing that we extend `HS2HAHealthChecker` from `HS2LeadershipStatus`. Should it be the other way around?

common/src/java/org/apache/hive/http/HttpServer.java
```diff
@@ -136,6 +136,11 @@ public class HttpServer {
 
 
   private final String name;
+  /**
+   * The list of web application contexts associated with the server.
+   * The first web application context at index 0 is the default context added during web server creation.
```
deniskuzZ, line 141: wouldn't it be better to define an instance var for the `default` / `root` WebAppContext`?

service/src/java/org/apache/hive/service/servlet/HS2LeadershipStatus.java
```diff
@@ -65,82 +63,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     response.flushBuffer();
   }
 
-  private void setResponseHeaders(final HttpServletResponse response) {
+  void setResponseHeaders(final HttpServletResponse response) {
```
deniskuzZ, line 66: protected ?

deniskuzZ:
> or health checking, because it returns true when the instance is a leader and false otherwise.

i am confused now. I thought that the HA health-check is supposed to check the status of a leader HS2.

difin:
> > or health checking, because it returns true when the instance is a leader and false otherwise.
> 
> i am confused now. I thought that the HA health-check is supposed to check the status of a leader HS2.

If let's say HS2 is running in HA mode and there are 2 HS2 instances, A and B, and A is a leader, then when calling the health check endpoint on A it will return true because A is a leader and when calling the health check endpoint on B it will return false because B is not a leader. If there is a leader failover and B becomes a leader then health checking on A will return false and health check on B will return true. There are several existing tests in `TestActivePassiveHA` that test it.

I don't know if in production environments the health checking script always calls the leader HS2 to check the health status or if it might call the slave HS2s. I asked someone who should know more about usage of the health check script, waiting for reply.

deniskuzZ (approved):
LGTM +1

PR #5653, opened 2025-02-22
HIVE-28766: EXECUTE IMMEDIATE with load data into table query displays ERROR on the console.

mdayakar:
HIVE-28766: EXECUTE IMMEDIATE 'load data inpath ''/tmp/test.txt'' OVERWRITE INTO table test_table' displays ERROR on the console.

When EXECUTE IMMEDIATE 'load data inpath ''/tmp/test.txt'' OVERWRITE INTO table test_table' query is executed then the load data into table query will directly send to HS2 and its get executed. After the query is executed by HS2 the control will be back to HPLSQL code in which it is trying to check the resultset metadata but in this case there is no resultset as its a load data query so it throws NPE which resulting HPLSQL error on the console even though data is successfully loaded into the table. So here we need to check whether the query result has resultset then only access its metadata.

Without the changes, beeline console shows HPLSQL error even though the data is loaded into the table successfully which confuses the user.

No

No

### DIFF
--- a/hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
+++ b/hplsql/src/main/java/org/apache/hive/hplsql/Stmt.java
@@ -1088,14 +1088,16 @@ public class Stmt {
       // Print the results
       else {
         int cols = query.columnCount();
-        while(query.next()) {
-          for(int i = 0; i < cols; i++) {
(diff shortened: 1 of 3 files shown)


kasakrisz (approved):
+1 pending tests

PR #5661, opened 2025-02-26
HIVE-28788: ObjectStore failed to commit the transaction could lead t…

dengzhhu653:
…o connection starvation

Fix the ObjectStore secondary pool connection starvation

No

No

Existing UTs

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
@@ -560,24 +560,9 @@ public class ObjectStore implements RawStore, Configurable {
    */
   @Override
   public boolean openTransaction() {
-    return openTransaction(null);
-  }
-
-  /**
-   * Opens a new one or the one already created. Every call of this function must
-   * have corresponding commit or rollback function call.
-   *
-   * @param isolationLevel The transaction isolation level. Only possible to set on the first call.
-   * @return an active transaction
-   */
-  @Override
-  public boolean openTransaction(String isolationLevel) {
     openTrasactionCalls++;
     if (openTrasactionCalls == 1) {
       currentTransaction = pm.currentTransaction();
-      if (isolationLevel != null) {
-        currentTransaction.setIsolationLevel(isolationLevel);
-      }
       currentTransaction.begin();
       transactionStatus = TXN_STATUS.OPEN;
     } else {
@@ -587,16 +572,10 @@ public class ObjectStore implements RawStore, Configurable {
         throw new RuntimeException("openTransaction called in an interior"
             + " transaction scope, but currentTransaction is not active.");
       }
-
-      // Can not change the isolation level on an already open transaction
-      if (isolationLevel != null && !isolationLevel.equals(currentTransaction.getIsolationLevel())) {
-        throw new RuntimeException("Can not set isolation level on an open transaction");
-      }
     }
 
     boolean result = currentTransaction.isActive();
-    debugLog("Open transaction: count = " + openTrasactionCalls + ", isActive = " + result + ", isolationLevel = "
-            + currentTransaction.getIsolationLevel());
+    debugLog("Open transaction: count = " + openTrasactionCalls + ", isActive = " + result);
     return result;
   }
 
@@ -677,10 +656,6 @@ public class ObjectStore implements RawStore, Configurable {
    */
   @Override
   public void rollbackTransaction() {
-    if (openTrasactionCalls < 1) {
-      debugLog("rolling back transaction: no open transactions: " + openTrasactionCalls);
-      return;
-    }
     debugLog("Rollback transaction, isActive: " + isActiveTransaction());
     try {
       if (isActiveTransaction() && transactionStatus != TXN_STATUS.ROLLBACK) {
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
@@ -141,17 +141,6 @@ public interface RawStore extends Configurable {
   boolean openTransaction();
 
   /**
-   * Opens a new one or the one already created. Every call of this function must
-   * have corresponding commit or rollback function call.
-   *
-   * @param isolationLevel The transaction isolation level. Only possible to set on the first call.
-   * @return an active transaction
-   */
-  default boolean openTransaction(String isolationLevel) {
-    throw new UnsupportedOperationException("Setting isolation level for this Store is not supported");
-  }
-
-  /**
    * if this is the commit of the first open call then an actual commit is
    * called.
    *


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -677,10 +656,6 @@ public boolean isActiveTransaction() {
    */
   @Override
   public void rollbackTransaction() {
-    if (openTrasactionCalls < 1) {
```
deniskuzZ, line 680: is this a main fix?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -560,24 +560,9 @@ public void shutdown() {
...
-
-  /**
-   * Opens a new one or the one already created. Every call of this function must
-   * have corresponding commit or rollback function call.
-   *
-   * @param isolationLevel The transaction isolation level. Only possible to set on the first call.
-   * @return an active transaction
-   */
-  @Override
-  public boolean openTransaction(String isolationLevel) {
```
deniskuzZ, line 574: we always passed `null`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
dengzhhu653: yes, we need the rollback in case of `openTrasactionCalls` is 0.
`openTrasactionCalls` is to tell when the transaction should commit, we need to very take care of `openTrasactionCalls` as it lives through the ObjectStore

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
dengzhhu653: right, we don't specify the `isolationLevel` any where

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
```diff
@@ -140,17 +140,6 @@ public interface RawStore extends Configurable {
...
-   * Opens a new one or the one already created. Every call of this function must
-   * have corresponding commit or rollback function call.
-   *
-   * @param isolationLevel The transaction isolation level. Only possible to set on the first call.
-   * @return an active transaction
-   */
-  default boolean openTransaction(String isolationLevel) {
-    throw new UnsupportedOperationException("Setting isolation level for this Store is not supported");
-  }
-
```
zabetak, line 153: `RawStore` is a user-facing (`metastore.rawstore.impl`) public interface so removing this default method is a breaking change. I guess there are not many users relying on this so it should be rather safe to drop it. However, I feel that dropping support for transactions with `isolationLevel` should have its own dedicated JIRA case unless its really tight to this fix.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
zabetak: Can we come up with any kind of unit test that reproduces the starvation or something else that demonstrates the necessity of this fix?

zabetak (commented):
The fix makes sense to me by reading the description of the JIRA but it would be nice if we could also back-it up with a test case if its not too complicated to do. Regarding the changes about support of transactions with isolation level I feel it belongs to another JIRA.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
deniskuzZ: we could create a spy for PersistenceManager and mock the currentTransaction to fail on commit

deniskuzZ (approved):

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/RawStore.java
deniskuzZ: +1

deniskuzZ: hi @pvary, do you recall why a new signature for `openTransaction` with `isolation` level param was introduced?

deniskuzZ: oh, we removed the use of it in https://github.com/apache/hive/pull/5129/files#diff-394eb2401c3389f5de729f3f1c6a423aaf0041d89d406212b14abf90bc650358L189, so should be fine to drop from the interface

deniskuzZ (approved):

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
dengzhhu653: This original can only fix the leak on primary pool, the secondary pool can still have such issue, https://github.com/dengzhhu653/hive/commit/7b4a66cf030e8570214341424bab258520abe919 can reproduce the problem

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
```diff
@@ -90,7 +94,42 @@ public DataSource create(Configuration hdpConfig, int maxPoolSize) throws SQLExc
       config.addDataSourceProperty(kv.getKey(), kv.getValue());
     }
 
-    return new HikariDataSource(initMetrics(config));
+    return new HikariDataSource(initMetrics(config)) {
+      @Override
+      public Connection getConnection() throws SQLException {
```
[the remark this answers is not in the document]
dengzhhu653: Yes, the solution may not sound elegant at first sight, but it only adds the proxy around the two connections from the secondary connection pool, the primary pool stays the same.
> we don't really know what side effects may appear if we introduce this rollback and close logic since test coverage right now is probably zero.
If an exception is happing on committing the transaction, then we know:
 1. At this point, the code reaches the end of transaction;
 2. The current operation(API call) should be failing.

so back to the solution,  we free the connection to the pool and throw the original exception to the caller, this is an expected behavior based on what we know, and the ObjectStore will rollback the current transaction(the primary connection), it should clean the current context by design.

This solution only affect the current ObjectStore, as an ObjectStore could be requested for other API calls after the issue occurs, we should ensure it do the right thing. I changed this https://github.com/dengzhhu653/hive/commit/7b4a66cf030e8570214341424bab258520abe919 a bit, it shows an insert following by the failed one(get or insert)(rollback and close logic) can get a successful run.

However the connection leak affects the whole service, it cannot serve all the requests until a restart, this is worst compared to only affecting the current ObjectStore.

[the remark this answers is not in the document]
dengzhhu653: updates: My approach still has a critical problem, the DataNucleus ID generator still holds the keys even when the connection rollback, this will cause ID conflict if multiple HMS here in the backend.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -560,24 +561,10 @@ public void shutdown() {
...
     openTrasactionCalls++;
     if (openTrasactionCalls == 1) {
       currentTransaction = pm.currentTransaction();
-      if (isolationLevel != null) {
-        currentTransaction.setIsolationLevel(isolationLevel);
-      }
+      currentTransaction.setIsolationLevel(Constants.TX_READ_COMMITTED);
```
deniskuzZ, line 567: why are we overriding the default isolation level here? it's not equivalent to the original code

dengzhhu653: reverted, the default is read-committed in datanucleus

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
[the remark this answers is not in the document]
dengzhhu653: since [HIVE-28839](https://issues.apache.org/jira/browse/HIVE-28839) was created to address the independent secondary connection leak, I reverted the changes on the data source

PR #5664, opened 2025-02-27
HIVE-28124: Do not allow non-numeric values in Hive table stats during an alter table

mszurap:
Hive should reject requests for altering table statistics which try to set non-numeric values in table properties which ought to contain numeric values. Currently these are **numFiles**, **numPartitions**, **totalSize**, **rawDataSize**, **numFilesErasureCoded**.
Worth to note that currently the "totalSize", "numFiles" and "numFilesErasureCoded" cannot be directly updated with an alter statement, still the new code verifies these and rejects the query in case the value is not a number (including an empty string). This might be useful in the future if they are going to be allowed to be changed again.

Non-numeric values in these table properties can cause other application failures, see the description in [HIVE-28124](https://issues.apache.org/jira/browse/HIVE-28124).

Queries like "alter table stat_test set TBLPROPERTIES('rawDataSize'='NaN')" are now failing with 
"SemanticException AlterTable rawDataSize failed with value NaN"

No

qtests added (both positive and negative), ran the tests successfully with "TestCliDriver".


ql/src/test/results/clientpositive/alter_tableprops_numeric.q.out
```diff
@@ -0,0 +1,52 @@
...
+#### A masked pattern was here ####
+Retention:          	0                   	 
+#### A masked pattern was here ####
+Table Type:         	EXTERNAL_TABLE      	 
+Table Parameters:	 	 
+	EXTERNAL            	TRUE                
+	STATS_GENERATED_VIA_STATS_TASK	true                
+	bucketing_version   	2                   
+#### A masked pattern was here ####
+	numFiles            	0
```
okumin, line 37: I wonder why this is not 1

okumin (commented):
@mszurap The intention and implementation look good to me.

The new tests failed on CI.
https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5664/1/pipeline/474

That's because you probably generated `.q.out` using the non-default CLI driver. The default driver is `TestMiniLlapLocalCliDriver` for positive tests and `TestNegativeLlapLocalCliDriver` for negative tests. Could you regenerate `.q.out` and push it?

We can control which drivers we use with [this file](https://github.com/apache/hive/blob/master/itests/src/test/resources/testconfiguration.properties), but the details are not explained well. [I am aware of it](https://lists.apache.org/thread/jcck8tdod3hyzf5wjzxzn075xn79st4h), but I've not updated the contributor's guide. I'd appreciate it if you could kindly regenerate the output files this time.

mszurap:
Thank you @okumin for looking into this. Right, running the test with "TestMiniLlapLocalCliDriver" places the results under the "llap" subdirectory, probably that was missing if that is used by default. Thanks for clarifying, I was not aware of that.
The result output do not differ with the TestMiniLlapLocalCliDriver, just the location. Should I remove the "ql/src/test/results/clientpositive/alter_tableprops_numeric.q.out" - if that is not really used? (to avoid confusion and duplication)
The negative output files did not change in contents and they do not have different locations, so on that part there was no change after the rerun.

ql/src/test/results/clientpositive/alter_tableprops_numeric.q.out
mszurap: Yeah, I was wondering about that too, similarly the "totalSize" also cannot be altered directly. I do not know if this is intentional or not, I did not dig myself into it.

okumin:
@mszurap Thanks.

> Should I remove the "ql/src/test/results/clientpositive/alter_tableprops_numeric.q.out" - if that is not really used? (to avoid confusion and duplication)

Yes. I think [this error](https://ci.hive.apache.org/blue/rest/organizations/jenkins/pipelines/hive-precommit/branches/PR-5664/runs/2/nodes/550/log/?start=0) is caused by the additional file. We need `ql/src/test/results/clientpositive/alter_tableprops_numeric.q.out` only when we explicitly declare that we test `ql/src/test/queries/clientpositive/alter_tableprops_numeric.q` using another driver(e.g., one of MR or Tez).
This PR does not include any MapReduce or Tez-specific changes. So, please keep only the LLAP's output.

okumin (approved):
+1. I'm waiting for CI

mszurap:
Thanks @okumin for merging the changes!

PR #5666, opened 2025-03-01
HIVE-28793: Set default expiry time for query history snapshots

abstractdog:
Set history.expire.max-snapshot-age-ms for query history table upon creation.

To let the service (introduced in [HIVE-28930](https://issues.apache.org/jira/browse/HIVE-28930)) expire snapshots of this table.

No.

No.

Final patch wasn't tested separately, but table property was tested in the early stages of the patch.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
@@ -89,6 +89,9 @@ public class IcebergRepository extends AbstractRepository implements QueryHistor
         ICEBERG_STORAGE_HANDLER);
     table.setProperty("table_type", "ICEBERG");
     table.setProperty("write.format.default", "orc");
+    // set the default max snapshot age to 1 day for query history
+    // this is applied only during table creation, it can be manually altered afterward.
+    table.setProperty("history.expire.max-snapshot-age-ms", Integer.toString(24 * 60 * 60 * 1000));
     table.setProperty(hive_metastoreConstants.META_TABLE_NAME, QUERY_HISTORY_DB_TABLE_NAME);
 
     table.setFields(schema.getFields());


ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
```diff
@@ -176,4 +187,9 @@ private void prepareConfForWrite() {
     SessionStateUtil.addCommitInfo(SessionState.getSessionConf(), tableDesc.getTableName(), jobId, 1,
         Maps.fromProperties(tableDesc.getProperties()));
   }
+
+  private void expireSnapshots() {
+    storageHandler.executeOperation(table, new AlterTableExecuteSpec<>(EXPIRE_SNAPSHOT,
+        new AlterTableExecuteSpec.ExpireSnapshotsSpec(10)));
```
ayushtkn, line 193: Do you want to hardcode ``10`` here or just pass ``null`` & let the TBLPROPERTIES handle it, this way it can be configurable also by ALTER command.

I don't think this would work the way you expect, there is some default time thing as well
```
  public static final String MAX_SNAPSHOT_AGE_MS = "history.expire.max-snapshot-age-ms";
  public static final long MAX_SNAPSHOT_AGE_MS_DEFAULT = 432000000L;
```

ayushtkn (commented):
Thanx @abstractdog will it be possible to extend a test for it as well

ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
abstractdog: thanks @ayushtkn for taking a look!
good point, I'm not familiar with EXPIRE SNAPSHOT functionality detail, so this is more like a DRAFT, let me consider

abstractdog: this seemed to work:
https://github.com/apache/hive/pull/5666/commits/4c9480ff53ab69e8a795a620d0ba8d5f9c635dec

added testing details in PR description

abstractdog: @ayushtkn : I just added a unit test for snapshot expiry testing

abstractdog:
while testing this patch, I realized that the meta hook commit process doesn't respect hiveLockEnabled, modified it, quite a small change but not related query history, please let me know if it can stay here @ayushtkn , @deniskuzZ

ayushtkn:
maybe we should split it out into a separate ticket, if it isn't related to QueryHistory

abstractdog:
> IMO we should split it out into a separate ticket, if it isn't related to QueryHistory

ack, just created https://github.com/apache/hive/pull/5687

abstractdog:
unit test passed without the locking hack (most probably it was an incomplete step that I worked around with proper configuration in the meantime)
also addressed sonarqube issues

ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/AbstractRepository.java
```diff
@@ -59,8 +60,8 @@ public void init(HiveConf conf, Schema schema) {
 
     try (Hive hive = Hive.get(conf)) {
       Database database = initDatabase(hive);
-      Table table = initTable(hive, database);
-      postInitTable(table);
+      initTable(hive, database);
+      postInitTable();
```
deniskuzZ, line 64: could we trigger `postInitTable` within the `initTable`?

ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
```diff
@@ -176,4 +193,13 @@ private void prepareConfForWrite() {
     SessionStateUtil.addCommitInfo(SessionState.getSessionConf(), tableDesc.getTableName(), jobId, 1,
         Maps.fromProperties(tableDesc.getProperties()));
   }
+
+  private void expireSnapshots() {
```
deniskuzZ, line 197: We have an `AcidHouseKeeperService`, I think we should create a similar `IcebergHouseKeeperService` and schedule periodic snapshot expiry on Iceberg tables.
You can define expiry policy for the `query-history` table and `IcebergHouseKeeperService` would take care of it

ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
```diff
@@ -101,8 +112,14 @@ protected Table createTable(Hive hive, Database db) throws HiveException {
     return table;
   }
 
+  @VisibleForTesting
+  int getSnapshotMaxAge() {
+    // expire/delete snapshots older than 1 day
+    return 24 * 60 * 60 * 1000;
```
deniskuzZ, line 118: that could be set on a table level - `history.expire.max-snapshot-age-ms`

deniskuzZ: oh, yes, that's what you do as well. (not sure if we need HiveConf for that)

abstractdog: I would not use a HiveConf, because this only applies when the table is first created, and from that point, the config is not effective (only if I implement some alter table mechanism that overcomplicates this area), so a HiveConf can make the user think the setting make its way magically any time, but in fact, the assumption is that user will ALTER the table manually

ql/src/java/org/apache/hadoop/hive/ql/queryhistory/repository/IcebergRepository.java
abstractdog: yeah, theoretically, handling them in a centralized place in metastore instead of multiple HS2s looks better from system design point of view
however, it needs a new service to be implemented, let me check

deniskuzZ: no need for a "NEW" service, simply add it to the list of `TASK_THREADS_REMOTE_ONLY`

abstractdog: took a look, my concern about this whole service is exactly the doTheMagic part: I remember we faced lots of problems with e.g. the PartitionManagementTask because it couldn't have been run in a lightweight fashion, and the common thing is that we need to iterate on all tables of all DBs to achieve what we want here: the best case scenario is that we implement a filter on ICEBERG tables (like: [listTableNamesByFilter](https://github.com/apache/hive/blob/4108ec495399800a1fe3f982697009944b03c8f2/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java#L432C16-L432C38)), kinda pushing down the filter to backend db
but even if we do that, we still need lots of code to make it as optimal as possible, just take a look at how [complicated](https://github.com/apache/hive/blob/4108ec495399800a1fe3f982697009944b03c8f2/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PartitionManagementTask.java#L98-L140) the partition management task filtering logic is to achieve the same).
So looking at the implementations so far, the "getting all iceberg tables" is the very part I'm worried about, and makes me wonder I really want to go towards that direction just the sake of this query history snapshot expiry, which is otherwise a very simple code

Focusing on the motivation, which is to handle the query history table—fully loaded and maintained by Hive—we owe it to users to optimize it as much as possible. However, the question is whether we should take on the additional burden in this PR to handle all Iceberg tables visible in the metastore, or keep the scope focused for now.

deniskuzZ: 1. you don't need to go down to partition level

abstractdog:
the generic metastore service logic has been moved to a separate jira/PR: https://github.com/apache/hive/pull/5786
this PR remains idle until we can merge that one, and here, only query_history related changes will be contributed (for example, setting the snapshot expiry table property properly)

deniskuzZ:
@abstractdog, can we close this PR as a generic solution was already merged?

abstractdog:
> @abstractdog, can we close this PR as a generic solution was already merged?

let me check if a minimal change is needed here to make the history table handled, like a table property set properly
few lines maybe

abstractdog:
@deniskuzZ , @ayushtkn : final patch here is as simple as https://github.com/apache/hive/pull/5666/commits/6b627a1f2dee9f478ba45d60aac8a2fcafbf704c
could this go to 4.1.0?

deniskuzZ (approved):
+1

PR #5668, opened 2025-03-03
[HIVE-28794]: Remove junit version from hive-metastore-tools/pom.xml to inherit version from parent pom

arorasimran0309:
Junit version upgrade in hive-metastore-tools

To bring version consistency throughout

No

Yes

Before:
image
image

After:
image
image

[dptree_junit.txt](https://github.com/user-attachments/files/19502139/dptree_junit.txt)

### DIFF
--- a/standalone-metastore/metastore-tools/pom.xml
+++ b/standalone-metastore/metastore-tools/pom.xml
@@ -30,13 +30,9 @@
     <javac.errorprone.version>2.8</javac.errorprone.version>
     <errorprone.core.version>2.3.1</errorprone.core.version>
     <picocli.version>3.1.0</picocli.version>
-    <junit.platform.runner.version>1.2.0</junit.platform.runner.version>
-    <junit.jupiter.api.version>5.2.0</junit.jupiter.api.version>
     <commons-math3.version>3.6.1</commons-math3.version>
     <jetbrain-annotation.version>16.0.2</jetbrain-annotation.version>
     <standalone.metastore.path.to.root>..</standalone.metastore.path.to.root>
-    <junit.jupiter.version>5.6.2</junit.jupiter.version>
-    <junit.vintage.version>5.6.2</junit.vintage.version>
   </properties>
   <!-- why the f*ck this pom.xml has a dependencyManagement block? -->
   <dependencyManagement>


standalone-metastore/metastore-tools/pom.xml
```diff
@@ -35,8 +35,8 @@
     <commons-math3.version>3.6.1</commons-math3.version>
     <jetbrain-annotation.version>16.0.2</jetbrain-annotation.version>
     <standalone.metastore.path.to.root>..</standalone.metastore.path.to.root>
-    <junit.jupiter.version>5.6.2</junit.jupiter.version>
-    <junit.vintage.version>5.6.2</junit.vintage.version>
+    <junit.jupiter.version>5.11.2</junit.jupiter.version>
+    <junit.vintage.version>5.11.2</junit.vintage.version>
```
okumin, line 39: The intention looks good to me. I wonder why we have to specify the version in both `standalone-metastore/pom.xml` and this file. Can we remove these properties and inherit the dependency management settings of the parent project?

ayushtkn: long long ago, there was a initiative HIVE-17159, where people wanted to make sure standalone-metastore should be independently releasable, so this is the reason these versions are duped here.

Though that initiative didn't work the way people wanted, but did leave this for us & I think going back has another set of challenges, considering it had consensus back then

okumin: Thanks for the explanation. In this case, we replicate the same property four times. I feel that the property in `standalone-metastore/metastore-tools/pom.xml` can be consolidated to that in `standalone-metastore/pom.xml`.

```
% git grep '<junit.jupiter.version>'
pom.xml:    <junit.jupiter.version>5.11.2</junit.jupiter.version>
standalone-metastore/metastore-tools/pom.xml:    <junit.jupiter.version>5.6.2</junit.jupiter.version>
standalone-metastore/pom.xml:    <junit.jupiter.version>5.11.2</junit.jupiter.version>
storage-api/pom.xml:    <junit.jupiter.version>5.11.2</junit.jupiter.version>
```

arorasimran0309: Thanks for the review and suggestions. Made the necessary changes. 
<junit.jupiter.version> and <junit.vintage.version> consolidated, to be inherited from the parent standalone-metastore/pom.xml

ayushtkn (commented):
```
    <junit.platform.runner.version>1.2.0</junit.platform.runner.version>
    <junit.jupiter.api.version>5.2.0</junit.jupiter.api.version>
```
These are also defined in this pom, I don't think they are being used, can we drop them as well?

arorasimran0309:
> ```
>     <junit.platform.runner.version>1.2.0</junit.platform.runner.version>
>     <junit.jupiter.api.version>5.2.0</junit.jupiter.api.version>
> ```
> 
> These are also defined in this pom, I don't think they are being used, can we drop them as well?

Yes, the version specifications <junit.platform.runner.version> and <junit.jupiter.api.version> are orphaned, and therefore can be dropped. Have made the required changes to clean up the codebase.

ayushtkn (approved):
LGTM

PR #5669, opened 2025-03-03
HIVE-28797: Hive Delegation Token Renewal fails for any kerberos principal user other than hive user

vikramahuja1001:
No

No

Tested on a cluster with kerberos enabled and also added a UT

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/security/TokenStoreDelegationTokenSecretManager.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/security/TokenStoreDelegationTokenSecretManager.java
@@ -105,7 +105,7 @@ public class TokenStoreDelegationTokenSecretManager extends DelegationTokenSecre
       if (info == null) {
           throw new InvalidToken("token expired or does not exist: " + identifier);
       }
-      renewIfRequired(System.currentTimeMillis(), identifier, info);
+      renewIfRequired(identifier, info);
       // we have to fetch the token again as it has been renewed and info still contains the previous renew time.
       info = this.tokenStore.getToken(identifier);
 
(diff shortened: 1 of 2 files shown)


zhangbutao:
@vikramahuja1001 Does this issue relate to HIVE-28042? 
Also cc @chinnaraolalam

vikramahuja1001:
@zhangbutao , yes this is related to HIVE-28042. So basically we were always renewing with the current user always but the renewer will not always be current user. To fix that using actually renewing the token from by getting the renewer itself.

vikramahuja1001:
@zhangbutao can the CI on this PR be retriggered?  For some reason one split is never starting in the CI.

zhangbutao:
> @zhangbutao can the CI on this PR be retriggered? For some reason one split is never starting in the CI.

Re-run the CI again  https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5669/7/pipeline

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/security/TokenStoreDelegationTokenSecretManager.java
```diff
@@ -105,7 +105,11 @@ public byte[] retrievePassword(DelegationTokenIdentifier identifier) throws Inva
       if (info == null) {
           throw new InvalidToken("token expired or does not exist: " + identifier);
       }
-      renewIfRequired(System.currentTimeMillis(), identifier, info);
+      try {
+        renewIfRequired(identifier, info);
+      } catch (InvalidToken e) {
+        LOGGER.warn("Failed to renew token: " + identifier, e);
```
zhangbutao, line 111: If the renew token fails, will the user's job fail?

vikramahuja1001: No, ongoing user's job will not fail as it does not require the token. There are 2 cases here:
1. If current time < renewal time : In this case it will never try to go for renewal.
2. current time >= renewal time : in this case it will attempt for token renewal. The renewal here is failsafe in case if the session ever requires the delegation token later in it's lifeline.

This will have no impact on the user's job.

chinnaraolalam:
Changes looks good to me +1

vikramahuja1001:
Thanks for the review @zhangbutao . Can you please merge it as well?

PR #5670, opened 2025-03-03
HIVE-28798: Bucket Map Join partially using partition transforms

okumin:

This PR updates `OpTraitsRulesProcFactory.SelectRule` to propagate bucketing information when the source is an Iceberg table, and only a subset of bucketing columns is used.

https://issues.apache.org/jira/browse/HIVE-28798

For better performance. Iceberg's transform spec allows us to bucketize multiple columns separately, e.g., stored in `/warehouse/db/table/data/key1=3/key2=5`. Hive's one encodes a set of all bucketing columns into a single integer.

No. The query plan can change, but BMJ on Iceberg has not been released yet.

No

I updated `iceberg_bucket_map_join_8.q` so that it includes various combinations.


iceberg/iceberg-handler/src/test/queries/positive/iceberg_bucket_map_join_8.q
```diff
@@ -4,47 +4,122 @@ set hive.auto.convert.join=true;
...
+(null, '1002', 'val_102', 12),
+(103, null, 'val_103', 13),
+(104, '1004', null, 14),
+(105, '1005', 'val_105', 15),
+(101, '1001', 'val_101', 16),
+(null, '1002', 'val_102', 17),
+(103, null, 'val_103', 18),
+(104, '1004', null, 19),
+(105, '1005', 'val_105', 20),
+(null, null, null, 21);
```
okumin, line 30: I added many records so that the big table detection always picks up `src_small`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/metainfo/annotation/OpTraitsRulesProcFactory.java
```diff
@@ -381,45 +381,84 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+          newBucketFunctions.add(bucketFunction);
+          break;
+        }
+        if (bucketFunction == null) {
+          // Hive's native bucketing is effective only when all the bucketing columns are used
+          newBucketColNamesList.add(new ArrayList<>());
+          newBucketFunctions.add(null);
+          break;
+        }
+        Optional<CustomBucketFunction> newBucketFunction = bucketFunction.select(retainedColumns);
```
okumin, line 420: I finally want to implement Hive's v1 and v2 bucketing as CustomBucketFunctions

ql/src/java/org/apache/hadoop/hive/ql/optimizer/metainfo/annotation/OpTraitsRulesProcFactory.java
```diff
@@ -381,45 +381,82 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
   public static class SelectRule implements SemanticNodeProcessor {
 
     // For bucket columns
-    // If all the columns match to the parent, put them in the bucket cols
+    // If the projected columns are compatible with the bucketing requirement, put them in the bucket cols
     // else, add empty list.
+    public void putConvertedColNamesForBucket(
```
ngsg: Would it be better to make it private as `getConvertedColNamesForSort`?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_bucket_map_join_8.q
```diff
@@ -4,47 +4,122 @@ set hive.auto.convert.join=true;
...
 
-SELECT *
+SELECT a.key1, a.id
 FROM srcbucket_big a
-JOIN src_small b ON a.key = b.key
+JOIN src_small b ON a.key1 = b.key1
 ORDER BY a.id;
 
--- Using a non-bucket column
+-- key2
```
ngsg, line 55: I think `key2` and `key2 & non-partition` parts could be removed, as their query plans are identical to the `key1` plans. But that's just a minor thought, and I don't have a strong opinion about it.

ngsg:
I have checked that the patch enables BMJ with a subset of bucket columns, and the patches looks good to me. I left a few minor comments.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_bucket_map_join_8.q
okumin: I am keeping `key2` because I believe it is worth testing whether Hive with Iceberg can process a non-first partition from multi-level partitioning independently.
I dropped key2 & non-partition.

okumin:
I rebased this branch since it is too obsolete. Now, CI is green.

okumin:
@ngsg Thanks for your thorough review!

PR #5672, opened 2025-03-04
HIVE-28799: Q Test: Incorrect Tag Provided in DROP Query, Should Use …

kokila-19:
…Existing Tag

Change in q test to delete an existing tag using "if exists" keyword.

### Why are the changes needed
drop tag in qtest where it is supposed to delete an existing tag.

No

No

Q test

### DIFF
--- a/iceberg/iceberg-handler/src/test/queries/positive/alter_table_create_tag.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/alter_table_create_tag.q
@@ -31,8 +31,8 @@ explain alter table iceTbl drop tag test_tag_2;
 alter table iceTbl drop tag test_tag_2;
 
 -- drop a tag with if exists
-explain alter table iceTbl drop tag if exists test_tag_3;
-alter table iceTbl drop tag if exists test_tag_3;
+explain alter table iceTbl drop tag if exists test_tag_1;
+alter table iceTbl drop tag if exists test_tag_1;
 
 -- drop a non-exist tag with if exists
 alter table iceTbl drop tag if exists test_tag_4;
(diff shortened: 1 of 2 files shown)


iceberg/iceberg-handler/src/test/queries/positive/alter_table_create_tag.q
```diff
@@ -31,8 +31,8 @@ explain alter table iceTbl drop tag test_tag_2;
 alter table iceTbl drop tag test_tag_2;
 
 -- drop a tag with if exists
-explain alter table iceTbl drop tag if exists test_tag_3;
-alter table iceTbl drop tag if exists test_tag_3;
+explain alter table iceTbl drop tag if exists test_tag_1;
+alter table iceTbl drop tag if exists test_tag_1;
```
ayushtkn, line 35: this was correct only, it is testing if the tag doesn't exist dropping the tag with if exist should make the query fail. That is intentionally using a tag which doesn't exist to test the "if exists" functionality

ayushtkn (changes requested):

iceberg/iceberg-handler/src/test/queries/positive/alter_table_create_tag.q
kokila-19: We are doing that in the next query and the comment specifically mentions that its dropping a non-existing tag.

Current queries:
1. drop an existing tag.
2. drop an existing tag with if exists (The tag used here does not exist)
3. drop a non-existing tag with if exists (The tag used here does not exist)

So, 2 and 3 are redundant. The missing case is testing the drop of existing tag using 'if exists' keyword. So I changed to the tag that exists.

```
-- drop a tag
explain alter table iceTbl drop tag test_tag_2;
alter table iceTbl drop tag test_tag_2;

-- drop a tag with if exists
explain alter table iceTbl drop tag if exists test_tag_3;  // test_tag_3 does not exist
alter table iceTbl drop tag if exists test_tag_3;

-- drop a non-exist tag with if exists
alter table iceTbl drop tag if exists test_tag_4; // test_tag_4 does not exist.

```

ayushtkn (approved):

kokila-19:
@ayushtkn can you please merge this?

PR #132, opened 2017-01-15
#HIVE-15631

Tartarus0zm:
if set hive.log.reload.variable.enable is true, and use hive.session.id in hive-log4j2.properties ,
 then will print hive.session.id on logs

PR #5674, opened 2025-03-04
HIVE-28800: Improve LongColumnStatsAggregator by preventing the overestimation of NDV

ngsg:

This patch implements a sanity-check in `LongColumnStatsAggregator`. The added check ensures that the estimated NDV is always less than or equal to the maximum number of integers within the estimated range.

If the estimated range of a long type column is `[A, B]`, then the maximum number of distinct values in the column should be `B - A + 1` (assuming no nulls in the column). Therefore, any integer that is larger than `B - A + 1` cannot be the estimated value of NDV.

No

No

The patch is tested manually by printing debug messages in bitvector-enabled HMS. Also, I checked that the existing unit tests contains invalid expected outputs.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
@@ -129,7 +129,16 @@ public class LongColumnStatsAggregator extends ColumnStatsAggregator implements
           // We have estimation, lowerbound and higherbound. We use estimation
           // if it is between lowerbound and higherbound.
           double densityAvg = densityAvgSum / partNames.size();
-          estimation = (long) ((aggregateData.getHighValue() - aggregateData.getLowValue()) / densityAvg);
+          if (densityAvg == 0) {
+            // This means each partition has exactly one value. (highValue == lowValue and NDV = 1).
+            // In this case, the NDV can be estimated as follows:
+            // NDV ~ R * (1 - (1 - 1 / R)^P) where R := highValue - lowValue + 1 and P := partNames.size().
+            // For the efficiency of computation, we just use min(R, P) instead.
+            estimation =
+                Math.min(aggregateData.getHighValue() - aggregateData.getLowValue() + 1, partNames.size());
+          } else {
+            estimation = (long) ((aggregateData.getHighValue() - aggregateData.getLowValue()) / densityAvg);
+          }
           if (estimation < lowerBound) {
             estimation = lowerBound;
           } else if (estimation > higherBound) {
@@ -163,7 +172,13 @@ public class LongColumnStatsAggregator extends ColumnStatsAggregator implements
           String partName = csp.getPartName();
           LongColumnStatsData newData = cso.getStatsData().getLongStats();
           if (useDensityFunctionForNDVEstimation) {
-            densityAvgSum += ((double) (newData.getHighValue() - newData.getLowValue())) / newData.getNumDVs();
+            if (newData.getNumDVs() == 0) {
+              // If NDV is not present, assume that the density is 0.5.
+              densityAvgSum += 0.5;
+            } else {
+              densityAvgSum +=
+                  ((double) (newData.getHighValue() - newData.getLowValue())) / newData.getNumDVs();
+            }
           }
           adjustedIndexMap.put(partName, (double) indexMap.get(partName));
           adjustedStatsMap.put(partName, cso.getStatsData());
@@ -192,7 +207,13 @@ public class LongColumnStatsAggregator extends ColumnStatsAggregator implements
               csd.setLongStats(aggregateData);
               adjustedStatsMap.put(pseudoPartName.toString(), csd);
               if (useDensityFunctionForNDVEstimation) {
-                densityAvgSum += ((double) (aggregateData.getHighValue() - aggregateData.getLowValue())) / aggregateData.getNumDVs();
+                if (aggregateData.getNumDVs() == 0) {
+                  // If NDV is not present, assume that the density is 0.5.
+                  densityAvgSum += 0.5;
+                } else {
+                  double range = aggregateData.getHighValue() - aggregateData.getLowValue();
+                  densityAvgSum += range / aggregateData.getNumDVs();
+                }
               }
               // reset everything
               pseudoPartName = new StringBuilder();
@@ -241,24 +262,10 @@ public class LongColumnStatsAggregator extends ColumnStatsAggregator implements
       columnStatisticsData.getLongStats().setHistogram(mergedKllHistogramEstimator.serialize());
     }
 
-    statsObj.setStatsData(columnStatisticsData);
+    // mutate columnStatisticsData to ensure NDV <= the number of integers in the given range.
+    adjustNDV(columnStatisticsData);
 
-    // NDV cannot exceed the number of integers within the given range.
-    long numIntInRange = columnStatisticsData.getLongStats().getHighValue()
-        - columnStatisticsData.getLongStats().getLowValue() + 1;
-    if (columnStatisticsData.getLongStats().getNumNulls() > 0) {
-      numIntInRange++;
-    }
-    if (numIntInRange < columnStatisticsData.getLongStats().getNumDVs()) {
-      LOG.debug(
-          "Adjust the estimated NDV from {} to {} as it exceeds the number of integers "
-              + "within the range of colStats: [{}, {}].",
-          columnStatisticsData.getLongStats().getNumDVs(),
-          numIntInRange,
-          columnStatisticsData.getLongStats().getLowValue(),
-          columnStatisticsData.getLongStats().getHighValue());
-      columnStatisticsData.getLongStats().setNumDVs(numIntInRange);
-    }
+    statsObj.setStatsData(columnStatisticsData);
 
     return statsObj;
   }
@@ -360,5 +367,24 @@ public class LongColumnStatsAggregator extends ColumnStatsAggregator implements
     extrapolateLongData.setHighValue(highValue);
     extrapolateLongData.setNumNulls(numNulls);
     extrapolateLongData.setNumDVs(ndv);
+  }
+
+  private void adjustNDV(ColumnStatisticsData columnStatisticsData) {
+    // NDV cannot exceed the number of integers within the given range.
+    long numIntInRange = columnStatisticsData.getLongStats().getHighValue()
+        - columnStatisticsData.getLongStats().getLowValue() + 1;
+    if (columnStatisticsData.getLongStats().getNumNulls() > 0) {
+      numIntInRange++;
+    }
+    if (numIntInRange < columnStatisticsData.getLongStats().getNumDVs()) {
+      LOG.debug(
+          "Adjust the estimated NDV from {} to {} as it exceeds the number of integers "
+              + "within the range of colStats: [{}, {}].",
+          columnStatisticsData.getLongStats().getNumDVs(),
+          numIntInRange,
+          columnStatisticsData.getLongStats().getLowValue(),
+          columnStatisticsData.getLongStats().getHighValue());
+      columnStatisticsData.getLongStats().setNumDVs(numIntInRange);
+    }
   }
 }
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
@@ -251,7 +251,7 @@ public class LongColumnStatsAggregator extends ColumnStatsAggregator implements
     }
 
     // mutate columnStatisticsData to ensure NDV <= the number of integers in the given range.
-    adjustNDV(columnStatisticsData);
+    adjustNDV(columnStatisticsData.getLongStats());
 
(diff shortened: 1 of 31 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
```diff
@@ -242,6 +242,24 @@ public ColumnStatisticsObj aggregate(List<ColStatsObjWithSourceInfo> colStatsWit
     }
 
     statsObj.setStatsData(columnStatisticsData);
+
+    // NDV cannot exceed the number of integers within the given range.
+    long numIntInRange = columnStatisticsData.getLongStats().getHighValue()
```
deniskuzZ, line 247: should we fix this as well
````
densityAvgSum += ((double) (newData.getHighValue() - newData.getLowValue())) / newData.getNumDVs();
````
since we could try to divide by 0
````
double densityAvg = densityAvgSum / partNames.size();
estimation = (long) ((aggregateData.getHighValue() - aggregateData.getLowValue()) / densityAvg);
````
````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java [hunk 2]
```diff
@@ -242,6 +242,24 @@ public ColumnStatisticsObj aggregate(List<ColStatsObjWithSourceInfo> colStatsWit
     }
 
     statsObj.setStatsData(columnStatisticsData);
+
+    // NDV cannot exceed the number of integers within the given range.
```
deniskuzZ, line 246: should we move the NDV correction logic inside of `extrapolate`?

deniskuzZ:
hi @ngsg, not sure if you saw the comments

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
ngsg: We may use a probabilistic data structure instead of an extrapolation method, so we could not embed the NDV correction logic inside `extrapolate`. I'll extract the correction logic into a separate method.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
ngsg:
```
densityAvgSum += ((double) (newData.getHighValue() - newData.getLowValue())) / newData.getNumDVs();
```
There might be an invariant such as columnStats being present implies `NDV != 0`, but it's safer to explicitly guard against division by zero. I'll update the code accordingly.

---

```
double densityAvg = densityAvgSum / partNames.size();
```
If `partNames.size()` is 0, we should return an empty ColumnStats or null, as there's nothing to aggregate. I'll check the convention and adjust the code to handle this case properly.

---

```
estimation = (long) ((aggregateData.getHighValue() - aggregateData.getLowValue()) / densityAvg);
```
It seems that `densityAvg` is 0 in some cases. I'll fix it as well.

ngsg:
@deniskuzZ I've addressed your review and updated the patch accordingly. For the division by zero concerns, I didn't modify the division by `partNames.size()` and `ndvEstimator.estimateNumDistinctValues()` because, if I understand them correctly, they are always positive. Could you please review the changes? Thank you.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
```diff
@@ -129,7 +129,16 @@ public ColumnStatisticsObj aggregate(List<ColStatsObjWithSourceInfo> colStatsWit
           // We have estimation, lowerbound and higherbound. We use estimation
           // if it is between lowerbound and higherbound.
           double densityAvg = densityAvgSum / partNames.size();
-          estimation = (long) ((aggregateData.getHighValue() - aggregateData.getLowValue()) / densityAvg);
+          if (densityAvg == 0) {
+            // This means each partition has exactly one value. (highValue == lowValue and NDV = 1).
```
deniskuzZ, line 133: 👍

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
```diff
@@ -192,7 +207,13 @@ public ColumnStatisticsObj aggregate(List<ColStatsObjWithSourceInfo> colStatsWit
               csd.setLongStats(aggregateData);
               adjustedStatsMap.put(pseudoPartName.toString(), csd);
               if (useDensityFunctionForNDVEstimation) {
-                densityAvgSum += ((double) (aggregateData.getHighValue() - aggregateData.getLowValue())) / aggregateData.getNumDVs();
+                if (aggregateData.getNumDVs() == 0) {
```
deniskuzZ, line 210: could it be `0` here as well?
https://github.com/apache/hive/pull/5674/files#diff-004a5c809d016b31028b52e5a5e591d30b446d974b0d22e386c35acfee845d59L227

deniskuzZ, line 210: could it be `0` after `ndvEstimator.estimateNumDistinctValues()`? if yes, isn't it the same here?
https://github.com/apache/hive/pull/5674/files#diff-004a5c809d016b31028b52e5a5e591d30b446d974b0d22e386c35acfee845d59R248

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
```diff
@@ -163,7 +172,13 @@ public ColumnStatisticsObj aggregate(List<ColStatsObjWithSourceInfo> colStatsWit
           String partName = csp.getPartName();
           LongColumnStatsData newData = cso.getStatsData().getLongStats();
           if (useDensityFunctionForNDVEstimation) {
-            densityAvgSum += ((double) (newData.getHighValue() - newData.getLowValue())) / newData.getNumDVs();
+            if (newData.getNumDVs() == 0) {
```
deniskuzZ, line 175: could it be `0` here as well?
https://github.com/apache/hive/pull/5674/files#diff-004a5c809d016b31028b52e5a5e591d30b446d974b0d22e386c35acfee845d59R105

deniskuzZ:
@ngsg, added a few comments, please. take a look. I think `0` check is only needed at line #132

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
ngsg: You're right; according to my previous comment, this should be also nonzero. Thank you for pointing out my mistake. I'll remove is-zero test here.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
ngsg: I write a test to validate that `newData.getNumDVs() == 0`. I check the test using LlapLocalDriver with an additional modification to `data/conf/llap/hivemetastore-site.xml`, specifically setting `metastore.stats.ndv.densityfunction` to `true`.

However, I think is-zero test is unnecessary at this point, as JVM does not throw an error even when the divisor is 0 (cf. p672 of the [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se24/jls24.pdf)).

If NDV is 0, this likely indicates that there are no rows in the given partition. In such cases, if high and low values are present(which is a weird situation), the density should be set to infinity. I don’t think we need to address this situation explicitly, as it should not occur under normal circumstances.

For documentation purposes, I attach the test I wrote below:
```
create table x_inventory (inv_quantity_on_hand int) partitioned by (inv_date_sk bigint);
insert into table x_inventory (inv_quantity_on_hand, inv_date_sk) values (0, 2000), (1, 2001), (2, 2002);

set hive.stats.ndv.algo=fm;
analyze table x_inventory compute statistics for columns;

alter table x_inventory partition (inv_date_sk = 2002) update statistics for column inv_quantity_on_hand set('numDVs'='2','numNulls'='0');
alter table x_inventory partition (inv_date_sk = 2001) update statistics for column inv_quantity_on_hand set('numDVs'='1','numNulls'='0');
alter table x_inventory partition (inv_date_sk = 2000) update statistics for column inv_quantity_on_hand set('numDVs'='0','numNulls'='0');

-- at this point, stats uses FMSketch.

set hive.stats.ndv.algo=hll;
insert into table x_inventory (inv_quantity_on_hand, inv_date_sk) values (3, 2003);

-- so there are FMSketch and HyperLogLog.

set hive.stats.column.autogather=false;
insert into table x_inventory (inv_quantity_on_hand, inv_date_sk) values (4, 2004);
-- Finally, this ensures that the colstats is incomplete and not mergable.

desc formatted x_inventory partition (inv_date_sk=2002) inv_quantity_on_hand;
desc formatted x_inventory partition (inv_date_sk=2003) inv_quantity_on_hand;
desc formatted x_inventory partition (inv_date_sk=2004) inv_quantity_on_hand;

-- This requests aggregated column stats of inv_quantity_on_hand.
explain select count(distinct inv_quantity_on_hand) from x_inventory;

```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
```diff
@@ -192,7 +207,13 @@ public ColumnStatisticsObj aggregate(List<ColStatsObjWithSourceInfo> colStatsWit
               csd.setLongStats(aggregateData);
               adjustedStatsMap.put(pseudoPartName.toString(), csd);
               if (useDensityFunctionForNDVEstimation) {
-                densityAvgSum += ((double) (aggregateData.getHighValue() - aggregateData.getLowValue())) / aggregateData.getNumDVs();
+                if (aggregateData.getNumDVs() == 0) {
+                  // If NDV is not present, assume that the density is 0.5.
+                  densityAvgSum += 0.5;
+                } else {
+                  double range = aggregateData.getHighValue() - aggregateData.getLowValue();
```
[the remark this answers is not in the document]
ngsg: I removed `range` variable and reverted the code to original version.

ngsg:
@deniskuzZ , I have checked your comments and agree that we only need to guard division-by-zero issue at `#132`, as you mentioned. I have added a comment for each case and updated the patch. Could you please confirm them? Thank you.

deniskuzZ (approved):
LGTM +1, pending tests

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
```diff
@@ -343,4 +356,23 @@ public void extrapolate(ColumnStatisticsData extrapolateData, int numParts,
     extrapolateLongData.setNumNulls(numNulls);
     extrapolateLongData.setNumDVs(ndv);
   }
+
+  private void adjustNDV(ColumnStatisticsData columnStatisticsData) {
+    // NDV cannot exceed the number of integers within the given range.
+    long numIntInRange = columnStatisticsData.getLongStats().getHighValue()
```
deniskuzZ, line 362: @ngsg , could we please rename var to `estimation`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
```diff
@@ -343,4 +356,23 @@ public void extrapolate(ColumnStatisticsData extrapolateData, int numParts,
     extrapolateLongData.setNumNulls(numNulls);
     extrapolateLongData.setNumDVs(ndv);
   }
+
+  private void adjustNDV(ColumnStatisticsData columnStatisticsData) {
```
deniskuzZ, line 360: please change param to `LongColumnStatsData columnStats` <-- `columnStatisticsData.getLongStats()`

ngsg: I changed the parameter to `LongColumnStatsData columnStats`.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/columnstats/aggr/LongColumnStatsAggregator.java
ngsg: Sure, I renamed it to `estimation`.

PR #5682, opened 2025-03-06
HIVE-28805: Re-parse the password from the original jdbc url

jjiang037:

Introduce a new regex pattern to reparse the password from the original jdbc url after getting session vars

Users can use Beeline with JDBC URLs that include passwords containing special characters

No

No

Test with replacing jdbc jar and connecting with passwords with/without special characters.

### DIFF
--- a/jdbc/src/java/org/apache/hive/jdbc/Utils.java
+++ b/jdbc/src/java/org/apache/hive/jdbc/Utils.java
@@ -540,6 +540,13 @@ public class Utils {
       }
     }
 
+    Pattern fullPasswordPattern = Pattern.compile("(?i)(?<=;|^)password=([^;]+)");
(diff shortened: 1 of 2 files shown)


saihemanth-cloudera (approved):
LGTM +1

jdbc/src/java/org/apache/hive/jdbc/Utils.java
```diff
@@ -540,6 +540,13 @@ public static JdbcConnectionParams extractURLComponents(String uri, Properties i
       }
     }
 
+    Pattern fullPasswordPattern = Pattern.compile("(?i)(?<=;|^)password=([^;]+)");
```
dengzhhu653, line 543: can we add some tests for this?

PR #5687, opened 2025-03-09
HIVE-28810: HiveIcebergMetaHook to respect table lock property

abstractdog:

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -116,9 +116,11 @@ import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.expressions.UnboundPredicate;
 import org.apache.iceberg.expressions.UnboundTerm;
 import org.apache.iceberg.hive.CachedClientPool;
+import org.apache.iceberg.hive.HiveLock;
 import org.apache.iceberg.hive.HiveSchemaUtil;
 import org.apache.iceberg.hive.HiveTableOperations;
 import org.apache.iceberg.hive.MetastoreLock;
+import org.apache.iceberg.hive.NoLock;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.mapping.MappingUtil;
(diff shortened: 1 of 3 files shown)


ayushtkn (approved):
LGTM

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -409,8 +411,12 @@ public void preAlterTable(org.apache.hadoop.hive.metastore.api.Table hmsTable, E
           context.getProperties().get(OLD_TABLE_NAME)).toString());
     }
     if (commitLock == null) {
-      commitLock = new MetastoreLock(conf, new CachedClientPool(conf, Maps.fromProperties(catalogProperties)),
-          catalogProperties.getProperty(Catalogs.NAME), hmsTable.getDbName(), hmsTable.getTableName());
+      if (HiveTableOperations.hiveLockEnabled(null, conf)) {
```
deniskuzZ, line 414: HiveIcebergMetaHook should always use MetastoreLock 
PS: `iceberg-catalog` could be dropped once iceberg Hive-4.0 support is merged)

deniskuzZ (dismissed):
-1

deniskuzZ (changes requested):
see https://github.com/apache/hive/pull/5687#discussion_r1986367661

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
abstractdog: what is interesting is that I can only see this whole hiveLockEnabled(...) thing used in testing scenarios, which applies to my case as well, see screenshot 
Screenshot 2025-03-10 at 10 40 58

this makes me think that if I want to make my query history PR snapshot work then
a) I have to override the locking behavior and bypass it in case of tests by extending the meta hook instead and use a testing one
b) or make it work with metastore locking

abstractdog: seems like b) works, I'm about to close this one soon

now I would appreciate a review on https://github.com/apache/hive/pull/5666 to finalize the first phase of query history maybe, thanks in advance!

PR #5691, opened 2025-03-12
HIVE-28817: Rebuilding a materialized view stored in Iceberg fails when schema has varchar column

kasakrisz:
In general materialized view table scan plan is just a TableScan operator.
```
HiveTableScan(table=[[default, mat1]], table:alias=[default.mat1])
```
However in some cases a `Project` is added with cast expressions. For example when the MV schema contains `varchar` typed columns.
```
HiveProject(b=[CAST($0):VARCHAR(256) CHARACTER SET "UTF-16LE"], _c1=[$1])
  HiveTableScan(table=[[default, mat1]], table:alias=[default.mat1])
```

* Use `HiveMaterializedViewUtils.extractTable` in `HiveIncrementalRelMdRowCount.source` to get the materialized view table object.
* When transforming the Materialized View rebuild AST to a multi insert AST extract all projected expression aliases from plan branch scanning the materialized view. 

Materialized view scan plan may contain a project with cast expressions. The project operator doesn't contain the MV table and an exception is thrown when querying it

No exception is thrown when rebuilding MVs with `varchar` typed column

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestIcebergLlapLocalCliDriver -Dqfile=mv_iceberg_orc8.q -pl itests/qtest-iceberg -Pitests,iceberg
```

### DIFF
--- a/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreLock.java
+++ b/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreLock.java
@@ -21,8 +21,6 @@ package org.apache.iceberg.hive;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
-import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.Executors;
@@ -33,6 +31,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.apache.hadoop.hive.metastore.LockRequestBuilder;
+import org.apache.hadoop.hive.metastore.api.DataOperationType;
 import org.apache.hadoop.hive.metastore.api.LockComponent;
 import org.apache.hadoop.hive.metastore.api.LockLevel;
 import org.apache.hadoop.hive.metastore.api.LockRequest;
@@ -42,12 +42,12 @@ import org.apache.hadoop.hive.metastore.api.LockType;
 import org.apache.hadoop.hive.metastore.api.ShowLocksRequest;
 import org.apache.hadoop.hive.metastore.api.ShowLocksResponse;
 import org.apache.hadoop.hive.metastore.api.ShowLocksResponseElement;
+import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
 import org.apache.iceberg.ClientPool;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
-import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.iceberg.util.Tasks;
 import org.apache.thrift.TException;
@@ -91,6 +91,7 @@ public class MetastoreLock implements HiveLock {
   private final long lockHeartbeatIntervalTime;
   private final ScheduledExecutorService exitingScheduledExecutorService;
   private final String agentInfo;
+  private final Configuration conf;
 
   private Optional<Long> hmsLockId = Optional.empty();
   private ReentrantLock jvmLock = null;
@@ -102,6 +103,7 @@ public class MetastoreLock implements HiveLock {
     this.fullName = catalogName + "." + databaseName + "." + tableName;
     this.databaseName = databaseName;
     this.tableName = tableName;
+    this.conf = conf;
 
     this.lockAcquireTimeout =
         conf.getLong(HIVE_ACQUIRE_LOCK_TIMEOUT_MS, HIVE_ACQUIRE_LOCK_TIMEOUT_MS_DEFAULT);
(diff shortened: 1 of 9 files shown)


ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/NonNativeAcidMaterializedViewASTBuilder.java
```diff
@@ -50,12 +50,7 @@ public void appendDeleteSelectNodes(ASTNode selectNode, String tableName) {
 
     for (int i = 0; i < selectNode.getChildCount(); ++i) {
       ASTNode selectExpr = (ASTNode) selectNode.getChild(i);
-      ASTNode expression = (ASTNode) selectExpr.getChild(0);
-      if (expression.getType() == HiveParser.DOT &&
-          expression.getChildCount() == 2 &&
-          expression.getChild(0).getType() == HiveParser.TOK_TABLE_OR_COL) {
-        selectedColumns.add(expression.getChild(1).getText());
-      }
+      selectedColumns.add(selectExpr.getChild(selectExpr.getChildCount() - 1).getText());
```
abstractdog, line 53: does this fix the same problem for types other than varchar?

iceberg/iceberg-handler/src/test/queries/positive/mv_iceberg_orc8.q
```diff
@@ -0,0 +1,22 @@
+-- MV source table has varchar column.
```
abstractdog, line 1: this qtest passes me on master without the fix, I'm running:
```
mvn install -Dtest.output.overwrite=true -Pitests,iceberg -Denforcer.skip=true -pl itests/qtest-iceberg -Dtest=TestIcebergLlapLocalCliDriver -Dqfile=mv_iceberg_orc8.q
```
how does this qtest validates the fix?

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/materialized/alter/rebuild/NonNativeAcidMaterializedViewASTBuilder.java
kasakrisz: This issue affect `string` types other than `string`. Those are converted to `string` because of Iceberg type system.

Removed the unused import

iceberg/iceberg-handler/src/test/queries/positive/mv_iceberg_orc8.q
kasakrisz: The base table must be native and the MV must be Iceberg. I updated the test.

iceberg/iceberg-handler/src/test/results/positive/mv_iceberg_orc8.q.out
```diff
@@ -0,0 +1,316 @@
...
+                      Map-reduce partition columns: _col0 (type: string)
+                      Statistics: Num rows: 1 Data size: 490 Basic stats: COMPLETE Column stats: COMPLETE
+                      value expressions: _col1 (type: bigint), _col2 (type: boolean), _col3 (type: int), _col4 (type: bigint), _col5 (type: string), _col6 (type: bigint), _col7 (type: string)
+            Execution mode: vectorized
+        Map 6 
+            Map Operator Tree:
+                TableScan
+                  alias: t1
+                  Statistics: Num rows: 2 Data size: 188 Basic stats: COMPLETE Column stats: COMPLETE
+                  Version interval from: 5632151099214777496
```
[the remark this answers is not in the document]
kasakrisz: It is no longer in the golden file.

iceberg/iceberg-handler/src/test/queries/positive/mv_iceberg_orc8.q
abstractdog: okay, it fails now with NPE, thanks, resolving this conversation

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -414,6 +414,7 @@ public void preAlterTable(org.apache.hadoop.hive.metastore.api.Table hmsTable, E
     }
 
     try {
+      conf.setLong(hive_metastoreConstants.TXN_ID, IcebergAcidUtil.getTxnId());
```
deniskuzZ, line 417: why do you need it here?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
```diff
@@ -391,4 +393,17 @@ public T build() {
     }
   }
 
+  static long getTxnId() {
```
deniskuzZ, line 396: that's not even required as it could be done as 1-liner
````
long txnId = Optional.ofNullable(SessionState.get())
    .map(ss -> ss.getTxnMgr().getCurrentTxnId()).orElse(0L);
````

**Comment by deniskuzZ** [comment]
I think we should revert this change entirely and do below instead
````
private static boolean isLockableTable(Table t, boolean isExternalEnabled, boolean isLocklessReads) {
    if (t.isTemporary()) {
      return false;
    }
    switch (t.getTableType()) {
      case MANAGED_TABLE:
      case MATERIALIZED_VIEW:
        return !(isLocklessReads && isTransactionalTable(t)) || t.isNonNative() && isExternalEnabled;
      case EXTERNAL_TABLE:
        return isExternalEnabled;
      default:
        return false;
    }
  }
````

File: iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreLock.java [hunk 6]
```diff
@@ -102,6 +103,7 @@ public MetastoreLock(Configuration conf, ClientPool<IMetaStoreClient, TException
     this.fullName = catalogName + "." + databaseName + "." + tableName;
     this.databaseName = databaseName;
     this.tableName = tableName;
+    this.conf = conf;
```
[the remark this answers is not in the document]
kasakrisz: Setting the TxnId of the Acid txn is not necessary and it would require Iceberg code change too.

I prepared a follow-up patch reverting the locking related changes and implement another approach: #5716

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
kasakrisz: Reverted in #5716

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
kasakrisz: Reverted in #5716

PR #5694, opened 2025-03-13
HIVE-28146: Add positive event type filter to the HMS notification fe…

saihemanth-cloudera:
…tch API

Added a new field 'eventTypeList' in the NotificationEventRequest

So that clients can request notificationEvents based on required event types.

No.

No.

Local testing and added unit test for the same.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
@@ -11299,12 +11299,20 @@ public class ObjectStore implements RawStore, Configurable {
       }
       if (rqst.isSetEventTypeList()) {
         filterBuilder.append(" && (");
-        for (String eventType : rqst.getEventTypeList()) {
-          parameterVals.add(eventType);
-          parameterBuilder.append(", java.lang.String para" + parameterVals.size());
-          filterBuilder.append("eventType == para" + parameterVals.size() + " || ");
+//        for (String eventType : rqst.getEventTypeList()) {
+//          parameterVals.add(eventType);
+//          parameterBuilder.append(", java.lang.String para" + parameterVals.size());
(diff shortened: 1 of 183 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -11322,6 +11322,13 @@ public NotificationEventResponse getNextNotification(NotificationEventRequest rq
         filterBuilder.setLength(filterBuilder.length() - 4); // remove the last " || "
         filterBuilder.append(") ");
       }
+      if (rqst.isSetEventTypeList()) {
+        for (String eventType : rqst.getEventTypeList()) {
+          parameterVals.add(eventType);
+          parameterBuilder.append(", java.lang.String para" + parameterVals.size());
+          filterBuilder.append(" && eventType == para" + parameterVals.size());
```
dengzhhu653, line 11329: this should be `||` instead of `&&`?

saihemanth-cloudera: You are right. Let me modify the test case to address the same.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -11322,6 +11322,16 @@ public NotificationEventResponse getNextNotification(NotificationEventRequest rq
         filterBuilder.setLength(filterBuilder.length() - 4); // remove the last " || "
         filterBuilder.append(") ");
       }
+      if (rqst.isSetEventTypeList()) {
+        filterBuilder.append(" && (");
+        for (String eventType : rqst.getEventTypeList()) {
```
dengzhhu653, line 11327: nit: leave it to you, can we use `rqst.contains(eventType)` to the `filterBuilder`?

dengzhhu653 (approved):

PR #5696, opened 2025-03-14
HIVE-28820: HMS might only creates the external directory and not the managed directory

DanielZhu58:

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -1233,56 +1233,48 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
           madeExternalDir = true;
         }
       } else {
-        if (dbMgdPath != null) {
-          try {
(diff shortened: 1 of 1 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -1233,56 +1233,48 @@ private void create_database_core(RawStore ms, final Database db)
...
-            madeManagedDir = UserGroupInformation.getLoginUser().doAs(new PrivilegedExceptionAction<Boolean>() {
-              @Override public Boolean run() throws MetaException {
-                if (!wh.isDir(dbMgdPath)) {
-                  LOG.info("Creating database path in managed directory " + dbMgdPath);
-                  if (!wh.mkdirs(dbMgdPath)) {
-                    throw new MetaException("Unable to create database managed path " + dbMgdPath + ", failed to create database " + db.getName());
-                  }
-                  return true;
+        final Path mgdPath = dbMgdPath != null ? dbMgdPath : defaultDbMgdPath;
+        try {
```
dengzhhu653, line 1237: could you create a new method for creating both the managed dir and the external dir?

PR #5699, opened 2025-03-18
HIVE-28824: Metastore should also respect the max thrift message size

dengzhhu653:
Respect the max thrift message on the server side

By default the Thrift allows the payload of message 100M at most, if current call exceeds this limit, the server will kill the connection silently, and the client will receive the exception indicating the connection is closed.

No

No

UT


service/src/java/org/apache/hive/service/cli/thrift/ThriftBinaryCLIService.java
```diff
@@ -100,7 +102,15 @@ protected void initServer() {
 
       // Server args
       int maxMessageSize = hiveConf.getIntVar(HiveConf.ConfVars.HIVE_SERVER2_THRIFT_MAX_MESSAGE_SIZE);
-      TThreadPoolServer.Args sargs = new TThreadPoolServer.Args(serverSocket).processorFactory(processorFactory)
+      TThreadPoolServer.Args sargs = new TThreadPoolServer.Args(new TServerSocket(serverSocket.getServerSocket()) {
```
deniskuzZ, line 105: why don't we reuse `MetastoreTServerSocket`, maybe we could use some generic name?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
```diff
@@ -29,8 +32,32 @@
...
+    TSocket ts = super.accept();
+    boolean keepAlive = MetastoreConf.getBoolVar(configuration, MetastoreConf.ConfVars.TCP_KEEP_ALIVE);
+    if (keepAlive) {
+      try {
+        ts.getSocket().setKeepAlive(true);
+      } catch (SocketException e) {
+        throw new TTransportException(e);
+      }
+    }
+    // for testing purpose, get the maxThriftMessageSize from the configuration for every new connection
```
deniskuzZ, line 55: we shouldn't use suboptimal design just for the sake of tests

itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/TestJdbcWithMiniKdc.java
```diff
@@ -263,6 +264,27 @@ public void testNegativeProxyAuth() throws Exception {
...
+    Statement stmt = conn.createStatement();
+    try {
+      StringBuilder createTable = new StringBuilder("create external table tesths2thriftmaxmessagesize(");
+      for (int i = 0; i < 100; i++) {
+        createTable.append("abcdefghijklmnopqrstuvwxyz").append(i).append(" string, ");
+      }
+      createTable.append(" a int)");
+      Throwable t = assertThrows(SQLException.class, () -> stmt.execute(createTable.toString())).getCause();
+      assertTrue(t instanceof TTransportException);
+      assertEquals(TTransportException.END_OF_FILE, ((TTransportException)t).getType());
```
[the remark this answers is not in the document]
dengzhhu653: No, this message only shows in HMS side(DEBUG), at client we can see "socket is closed by peer"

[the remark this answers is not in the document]
dengzhhu653: No, this message only shows in HMS side(DEBUG), at client we can see "socket is closed by peer", the `TTransportException.END_OF_FILE` can tell us there is a connection problem.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
dengzhhu653: For the sake of tests, I make the `maxThriftMessageSize` can change dynamically based on the test need, compared to initialize the value at HMS startup and add an extra method to change this limit for test only.

itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/TestJdbcWithMiniKdc.java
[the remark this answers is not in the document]
dengzhhu653: Add the check on the error message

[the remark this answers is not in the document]
deniskuzZ: is that a connection problem or a payload problem?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
```diff
@@ -29,8 +32,32 @@
  * accepted sockets.
  *
  */
-public class TServerSocketKeepAlive extends TServerSocket {
-  public TServerSocketKeepAlive(TServerSocket serverSocket) throws TTransportException {
+public class MetastoreTServerSocket extends TServerSocket {
+  private final Configuration configuration;
+
+  public MetastoreTServerSocket(TServerSocket serverSocket,
```
deniskuzZ, line 38: why pass Conf object instead of boolean `keepAlive`?

service/src/java/org/apache/hive/service/cli/thrift/ThriftBinaryCLIService.java
deniskuzZ: sorry, i did catch your point. 
If I read it correctly, the anonymous class looks similar to MetastoreTServerSocket except for the `keepAlive`.
Can't we extract it to the common module and reuse (pass keepAlive & maxMessageSize as constructor args)?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
deniskuzZ, line 38: why pass conf, instead of tcpKeepAlive var and maxThriftMessageSize?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
```diff
@@ -29,8 +32,32 @@
...
+      Configuration conf) throws TTransportException {
     super(serverSocket.getServerSocket());
+    this.configuration = Objects.requireNonNull(conf);
+  }
+
+  @Override
+  public TSocket accept() throws TTransportException {
+    TSocket ts = super.accept();
+    boolean keepAlive = MetastoreConf.getBoolVar(configuration, MetastoreConf.ConfVars.TCP_KEEP_ALIVE);
+    if (keepAlive) {
```
deniskuzZ, line 48: it doesn't seen equivalent to original code. I don't see explicit set of setKeepAlive on socket when keepAlive==true

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
```diff
@@ -29,8 +32,32 @@
...
+        ts.getSocket().setKeepAlive(true);
+      } catch (SocketException e) {
+        throw new TTransportException(e);
+      }
+    }
+    // get the limit from the configuration for every new connection
+    int maxThriftMessageSize = (int) MetastoreConf.getSizeVar(
+        configuration, MetastoreConf.ConfVars.THRIFT_METASTORE_CLIENT_MAX_MESSAGE_SIZE);
+    if (maxThriftMessageSize > 0) {
+      ts.getConfiguration().setMaxMessageSize(maxThriftMessageSize);
```
deniskuzZ, line 59: should we check if `ts.getConfiguation` != null similar to `HiveAuthUtils.configureThriftMaxMessageSize`

itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/TestJdbcWithMiniKdc.java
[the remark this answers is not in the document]
dengzhhu653: the payload causes the HMS kill the connection, the client reading from the connection will fail

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
dengzhhu653: the configuration in TSocket is not null for all cases

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
dengzhhu653: the check is performed outside of the class before the PR, https://github.com/apache/hive/pull/5699/files#diff-00e70b6958060aa36762b21bf16676f83af01c1e09b56816aecc6abe7c8ac866L634

deniskuzZ: yes, but were. did we execute `setKeepAlive` before?

dengzhhu653: it's introduced in HIVE-1410 and removed by HIVE-25098(not sure why we remove it on upgrading the thrift lib)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
dengzhhu653: 1. for changing the limit in test easily, as this class is initialized on HMS startup; 2. through configuration it can make customize the socket easily in the future, such as configuring the socket read timeout

itests/hive-minikdc/src/test/java/org/apache/hive/minikdc/TestJdbcWithMiniKdc.java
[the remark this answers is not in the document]
dengzhhu653: the connection is closed, so the client knows nothing what's happening.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
```diff
@@ -29,8 +32,32 @@
  * accepted sockets.
  *
  */
-public class TServerSocketKeepAlive extends TServerSocket {
-  public TServerSocketKeepAlive(TServerSocket serverSocket) throws TTransportException {
+public class MetastoreTServerSocket extends TServerSocket {
```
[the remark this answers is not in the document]
dengzhhu653: done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
```diff
@@ -643,7 +639,7 @@ private static ThriftServer startBinaryMetastore(int port, HadoopThriftAuthBridg
           return thread;
     });
 
-    TThreadPoolServer.Args args = new TThreadPoolServer.Args(serverSocket)
+    TThreadPoolServer.Args args = new TThreadPoolServer.Args(new MetastoreTServerSocket(serverSocket, conf))
```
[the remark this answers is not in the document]
dengzhhu653: done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
deniskuzZ: nice :)
but do we need the `if check`? can't we simply do `ts.getSocket().setKeepAlive(keepAlive);`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreTServerSocket.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Objects;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
+import org.apache.thrift.transport.TServerSocket;
+import org.apache.thrift.transport.TSocket;
+import org.apache.thrift.transport.TTransportException;
+
+/**
+ * TServerSocketKeepAlive - like TServerSocket, but will enable keepalive for
```
deniskuzZ, line 31: please update the java doc

dengzhhu653: done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
dengzhhu653: done

dengzhhu653: After some check, the default `keepAlive` in thrift 0.16 is true, https://github.com/apache/thrift/blob/0.16.0/lib/java/src/org/apache/thrift/transport/TSocket.java#L78, so `KeepAlive` is always true whether `hive.metastore.server.tcp.keepalive` is true or not based on [HIVE-1410](https://issues.apache.org/jira/browse/HIVE-1410)

Since the `KeepAlive` helps HMS prevent the half-open connections, so it would be good to enable it, so I would like to keep the [HIVE-25098](https://issues.apache.org/jira/browse/HIVE-25098).

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreTServerSocket.java
```diff
@@ -18,19 +18,37 @@
...
 /**
- * TServerSocketKeepAlive - like TServerSocket, but will enable keepalive for
- * accepted sockets.
- *
+ * Give HMS a chance to customize the accepted sockets, such as
+ * set the thrift max message size.
  */
-public class TServerSocketKeepAlive extends TServerSocket {
-  public TServerSocketKeepAlive(TServerSocket serverSocket) throws TTransportException {
+public class MetaStoreTServerSocket extends TServerSocket {
```
deniskuzZ, line 33: since it's super simple, why don't we use an anonymous class similar to ThriftBinaryCLIService?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
deniskuzZ: why don't we drop the config then?

dengzhhu653: this will remove the property in `metaVars`, I would like to leave it in a separate jira, to make the fix port to old releases easily

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreTServerSocket.java
dengzhhu653: done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreTServerSocket.java
deniskuzZ: ok

deniskuzZ (approved):
LGTM +1, pending tests

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
```diff
@@ -643,7 +584,21 @@ private static ThriftServer startBinaryMetastore(int port, HadoopThriftAuthBridg
...
     });
 
-    TThreadPoolServer.Args args = new TThreadPoolServer.Args(serverSocket)
+    TThreadPoolServer.Args args =
+        new TThreadPoolServer.Args(new TServerSocket(serverSocket.getServerSocket()) {
+          @Override
+          public TSocket accept() throws TTransportException {
+            TSocket ts = super.accept();
+            // get the limit from the configuration for every new connection
+            int maxThriftMessageSize = (int) MetastoreConf.getSizeVar(
```
deniskuzZ, line 593: you can't change HMS configs at runtime, isn't it?

dengzhhu653: In production env, yes. but in our tests HMS and the test share the same `conf` instances, so the limit can be changed

deniskuzZ: I'll leave it to you, but in my opinion since the config is static there is no need to reevaluate it on every new connection

PR #5700, opened 2025-03-18
HIVE-28828: Add index for TXNS table

wecharyu:
Add unique index for `TXNS(TXN_STATE, TXN_ID)`.

`metastore.task.threads.remote` service would run periodically, it would query for TXNS table like:
https://github.com/apache/hive/blob/2d528556b69c1ec011e83f69d2a2fdcfb78b356e/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/service/AcidOpenTxnsCounterService.java#L49-L53

The `coundOpenTxns()` could not hit index, add this index could improve the performance.

No.

No.

Comparing the backend db cpu load after update.


deniskuzZ:
`AcidOpenTxnsCounterService` is the background metrics collection service, so I don't think index should be very. critical for it. On the other side adding an index might affect the write path: open txns API

wecharyu:
@deniskuzZ We have many abort txn records that cause the `countOpenTxns()` slow. HMS supports multiple instances, so the `AcidOpenTxnsCounterService` would run frequently, it seems better to improve the performance for it.

deniskuzZ:
@wecharyu, isn't it running only on the leader instance? 
see ````HouseKeepingTasks implements LeaderElection.LeadershipStateListener````

wecharyu:
Got it, we are still running hive `3.1.2` version that does not include `LeaderElection`. Besides `countOpenTxns()` api, we also find slow query in `openTxns()`,  the new index would improve it:
```sql
select min(txn_id) from TXNS where txn_state='o';
```

deniskuzZ:
> Got it, we are still running hive `3.1.2` version that does not include `LeaderElection`. Besides `countOpenTxns()` api, we also find slow query in `openTxns()`, the new index would improve it:
> 
> ```sql
> select min(txn_id) from TXNS where txn_state='o';
> ```

It might improve the `select`, but degrade the `insert` that is more important (we need to minimize the db lock time during txn open)

wecharyu:
The index is simple enough (2 columns),  I think the write operation would not occupy lock for long time. Here is the `openTxns` benchmark for reference:
```bash
- no index
openTxns.10     6.07188 6.02161 3.81099 13.9566 30.2951
openTxns.100    10.8020 10.0459 6.05904 29.9688 26.6960
openTxns.10000  28.0985 27.4915 18.8563 56.2814 22.1813

- with index
openTxns.10     5.80826 5.98342 3.64658 16.2209 34.8030
openTxns.100    9.47621 8.04049 5.79926 19.9999 35.4419
openTxns.10000  25.5064 24.0800 19.0202 46.0265 20.8510
```

deniskuzZ:
cc @dengzhhu653, WDYT?

dengzhhu653:
We only have a very limited number of `TXN_STATE`: OPEN, COMMITTED, ABORTED. OPEN is the most common state in the production I guess, so introducing the index on `TXNS(TXN_STATE, TXN_ID)` might not make much sense per my understanding.

PR #5701, opened 2025-03-18
HIVE-28826: Beeline fails to connect with spaces in the jdbc url

mszurap:
In some environments we need to customize the javax.jdo.option.ConnectionURL, and typically with Oracle databases the JDBC connection URL might contain spaces, like

jdbc:oracle:thin:@(DESCRIPTION = (ADDRESS_LIST=...

This is not handled by beeline, so hive schema upgrades are failing.

For the Oracle JDBC driver at least, the URL escaping does not work. The proposed solution is to remove the whitespaces from the connection url before passing the url to the Commands.java:
https://github.com/mszurap/hive/blob/2d528556b69c1ec011e83f69d2a2fdcfb78b356e/beeline/src/java/org/apache/hive/beeline/Commands.java#L1521.

Spaces were not respected before in the connection url. However if spaces are needed in the connection URL (like in ";sslTrustStore=/path/to/mydir with spaces/truststore.jks" or "?tez.application.tags=A B") then the URL escaping with "%20" can be still used in those places, those are decoded properly back to spaces.

To make the upgades more robust, we should handle these situations.

No

No

Added a new beeline test "testSimpleArgsWithSpaceInUrl" to test the functionality that the spaces are now removed from the connection url ("-u" argument only).
The beeline unit tests were successful.

### DIFF
--- a/beeline/src/java/org/apache/hive/beeline/BeeLine.java
+++ b/beeline/src/java/org/apache/hive/beeline/BeeLine.java
@@ -900,6 +900,8 @@ public class BeeLine implements Closeable {
     getOpts().setScriptFile(cl.getOptionValue("f"));
 
     if (url != null) {
+      //remove white spaces in the URL in case there is any, like in "jdbc:oracle:thin:@(DESCRIPTION = (ADDRESS_LIST=..."
+      url = url.replaceAll("\\s+","");
       String hplSqlMode = Utils.parsePropertyFromUrl(url, Constants.MODE);
       if ("HPLSQL".equalsIgnoreCase(hplSqlMode)) {
         getOpts().setDelimiter("/");
--- a/beeline/src/test/org/apache/hive/beeline/TestBeelineArgParsing.java
+++ b/beeline/src/test/org/apache/hive/beeline/TestBeelineArgParsing.java
@@ -132,6 +132,16 @@ public class TestBeelineArgParsing {
   }
 
   @Test
+  public void testSimpleArgsWithSpaceInUrl() throws Exception {
+    TestBeeline bl = new TestBeeline();
+    String args[] = new String[] {"-u", "url with space", "-n", "name",
+            "-p", "password", "-d", "driver", "-a", "authType"};
+    org.junit.Assert.assertEquals(0, bl.initArgs(args));
+    Assert.assertTrue(bl.connectArgs.equals("urlwithspace name password driver"));
+    Assert.assertTrue(bl.getOpts().getAuthType().equals("authType"));
+  }
+
+  @Test
   public void testEmptyHiveConfVariable() throws Exception {
     ByteArrayOutputStream os = new ByteArrayOutputStream();
     PrintStream ops = new PrintStream(os);


mszurap:
Yeah.
- Right now, beeline fails to connect if there are spaces anywhere - because the "string".split(" ") skews the parameters
- With the fix, we remove the spaces, so that could cause problems in such situations (in the given example the directory without spaces might not exist), but as mentioned we have a manual workaround at least with urlencoding the spaces. 

I'm not sure if we need to address those very edge cases.

mszurap:
Hi @Aggarwal-Raghav , thanks for looking at this.
Not as an excuse, but even now, if there is a space anywhere in the ";sess_var_list?hive_conf_list#hive_var_list" part, the string.split will split it and this usecase breaks even now, anybody wanting that already needs to escape the spaces.

If we want to support spaces in the ";sess_var_list?hive_conf_list#hive_var_list" part too, then:
1. we either need to refactor the whole code to avoid the string splitting by space and pass the parameters in a different way to the "Commands" (haven't explored how complex is this)
2. or we use the same approach but carefully, we can skip removing the spaces after the semicolon, the question mark or the hashtag sign. Would you agree to this rather?

mszurap:
@abstractdog @Aggarwal-Raghav as mentioned, this fix does not break the current workloads as even now the spaces are not tolerated by beeline. Can you advise how should we proceed? Do we need to refactor everything just to support the spaces in the  ";sess_var_list?hive_conf_list#hive_var_list" section?

mszurap:
@nrg4878 @abstractdog do you have suggestions how to proceed further? 
Do you have objections agains removing all the spaces from the URL? (by definition, URLs cannot contain spaces, they need to be encoded).

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -900,6 +900,8 @@ private boolean connectUsingArgs(BeelineParser beelineParser, CommandLine cl) {
     getOpts().setScriptFile(cl.getOptionValue("f"));
 
     if (url != null) {
+      //remove white spaces in the URL in case there is any, like in "jdbc:oracle:thin:@(DESCRIPTION = (ADDRESS_LIST=..."
```
nrg4878, line 903: @mszurap JDBC URLs can also contain passwords where it is legal to have spaces in passwords. It would fail in such cases if we remove all spaces in the URL.
Shouldnt we argue that user should not use spaces when not needed?

mszurap: @nrg4878 let's say we do not replace all the spaces, but only those before the semicolon or question mark or hashtag (special characters are `;?#`) Like
```
    if (url != null) {
      url = removeSpacesFromURLBeforeFirstSpecialChar(url);
```
Then we need to adjust the test code to the below, which shows the problem with the spaces in the url: this will cause the command line arguments to be skewed:
```
    String args2[] = new String[] {"-u", "url with space;pw=Pw with spaces", "-n", "name",
        "-p", "password", "-d", "driver", "-a", "authType"};
    org.junit.Assert.assertEquals(0, bl.initArgs(args2));
    Assert.assertTrue(bl.connectArgs.equals("urlwithspace;pw=Pw with spaces name password driver"));
    Assert.assertTrue(bl.getOpts().getAuthType().equals("authType"));
```
the "with" becomes the username, "spaces" becomes the password, "name" becomes the driver.
This would happen even right now, in the current Hive code if someone would use spaces anywhere in the URL.

mszurap: For the reference, from a lab cluster, the connection fails if there is any space in the URL:
```
beeline -u 'jdbc:hive2://zk1:2181,zk2:2181,zk3:2181/default;password=hive with spaces;serviceDiscoveryMode=zooKeeper;user=hive;zooKeeperNamespace=hiveserver2' --verbose
...
!connect jdbc:hive2://zk1:2181,zk2:2181,zk3:2181/default;password=hive with spaces;serviceDiscoveryMode=zooKeeper;user=hive;zooKeeperNamespace=hiveserver2 hive [passwd stripped]
Connecting to jdbc:hive2://zk1:2181,zk2:2181,zk3:2181/default;password=hive
java.lang.ClassNotFoundException: hive
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
...
        at org.apache.hive.beeline.BeeLine.execCommandWithPrefix(BeeLine.java:1463)
        at org.apache.hive.beeline.BeeLine.dispatch(BeeLine.java:1502)
        at org.apache.hive.beeline.BeeLine.connectUsingArgs(BeeLine.java:922)
        at org.apache.hive.beeline.BeeLine.initArgs(BeeLine.java:804)
        at org.apache.hive.beeline.BeeLine.begin(BeeLine.java:1115)
        at org.apache.hive.beeline.BeeLine.begin(BeeLine.java:1089)
        at org.apache.hive.beeline.BeeLine.mainWithInputRedirection(BeeLine.java:547)
        at org.apache.hive.beeline.BeeLine.main(BeeLine.java:529)
...
0: jdbc:hive2://hs2 (closed)> Connection is already closed.
```
In this example, as previously explained, the username "hive" became the driver class name due to the skew in the arguments.
The fact that there are spaces in the URL immediately caused the problem with our customer too, this is the primary reason for this fix too. We cannot have spaces anywhere, not even in the passwords.

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -900,6 +900,8 @@ private boolean connectUsingArgs(BeelineParser beelineParser, CommandLine cl) {
...
+      url = url.replaceAll("\\s+","");
```
dengzhhu653, line 904: Personally I would like not to make any change on the input url, that is, the beeline can accept the url whatever the user provides. If the beeline cannot parse the url correctly, then the user should make changes to fit into the Beeline, either replacing the namespace with %20 or removing the space from the jdo connection url.

PR #5702, opened 2025-03-18
HIVE-28825: Support conversion of complex types to String

soumyakanti3578:

Support conversion of complex types to String using `cast(<value> as string)` (`GenericUDFToString.java`)

Details in https://issues.apache.org/jira/browse/HIVE-28825

No, except for the new feature.

No

```
mvn test -pl itests/qtest -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile="complex_type_to_string.q"
```


ql/src/test/queries/clientpositive/complex_type_to_string.q
```diff
@@ -0,0 +1,31 @@
+create table table1(id int, txt string, num int, flag string);
```
kasakrisz, line 1: Could you please add some test cases where complex types are nested?
```
struct {
  a map<string, array<int>>,
  b struct { c decimal(10,3), d double}
}
```

ql/src/test/queries/clientpositive/complex_type_to_string.q
```diff
@@ -0,0 +1,31 @@
...
+  (3,  NULL,   3, 'FALSE');
+insert into  table2 values
+  (1, 'one',   5, 'FALSE'),
+  (2, 'two',  14, 'TRUE'),
+  (3,  NULL,   3, 'FALSE');
+  
+select cast(array(*) as string) from table1;
+select cast(array(*) as string) from table2;
+select md5(s) csum from (select cast(array(*) as string) s from table1) t;
+select md5(s) csum from (select cast(array(*) as string) s from table2) t;
```
kasakrisz, line 16: These tests pass a `string` to the `md5` function. I don't think it is necessary to test it because we already have test for that in [udf_md5.q](https://github.com/apache/hive/blob/master/ql/src/test/queries/clientpositive/udf_md5.q)
Please let me know if there is something else that these tests cover.

Do we want to support `md5 ` with complex type parameter? I think it is out of scope of this patch.

ql/src/test/queries/clientpositive/complex_type_to_string.q
```diff
@@ -0,0 +1,31 @@
...
+  (1, 'one',   5, 'FALSE'),
+  (2, 'two',  14, 'TRUE'),
+  (3,  NULL,   3, 'FALSE');
+insert into  table2 values
+  (1, 'one',   5, 'FALSE'),
+  (2, 'two',  14, 'TRUE'),
+  (3,  NULL,   3, 'FALSE');
+  
+select cast(array(*) as string) from table1;
+select cast(array(*) as string) from table2;
```
kasakrisz, line 14: Could you please add these tests to [udf_to_string.q](https://github.com/apache/hive/blob/master/ql/src/test/queries/clientpositive/udf_to_string.q) file? I think we don't need a separate q file. WDYT?

ql/src/test/queries/clientpositive/complex_type_to_string.q
soumyakanti3578: Added nested test in udf_to_string.q

ql/src/test/queries/clientpositive/complex_type_to_string.q
soumyakanti3578: I have removed tests with md5.

ql/src/test/queries/clientpositive/complex_type_to_string.q
soumyakanti3578: Added tests for complex types in udf_to_string.q

ql/src/test/queries/clientpositive/udf_to_string.q
```diff
@@ -1,21 +1,34 @@
...
 
-SELECT CAST(CAST(-3.14 AS DOUBLE) AS STRING) FROM src tablesample (1 rows);
-SELECT CAST(CAST(-3.14 AS FLOAT) AS STRING) FROM src tablesample (1 rows);
-SELECT CAST(CAST(-3.14 AS DECIMAL(3,2)) AS STRING) FROM src tablesample (1 rows);
+SELECT CAST(CAST(-3.14 AS DOUBLE) AS STRING) FROM alltypesorc tablesample (1 rows);
+SELECT CAST(CAST(-3.14 AS FLOAT) AS STRING) FROM alltypesorc tablesample (1 rows);
+SELECT CAST(CAST(-3.14 AS DECIMAL(3,2)) AS STRING) FROM alltypesorc tablesample (1 rows);
 
-SELECT CAST('Foo' AS STRING) FROM src tablesample (1 rows);
+SELECT CAST('Foo' AS STRING) FROM alltypesorc tablesample (1 rows);
```
kasakrisz, line 18: Does the from clause relevant in these statements? It seem that we are casting constants.
Example:
```
SELECT CAST(CAST(-3.14 AS DOUBLE) AS STRING);
```
should also test `GenericUDFToString`. 

Could you please confirm?

kasakrisz (approved):

PR #5631, opened 2025-02-04
HIVE-25043: Support custom UDF in Vectorized mode

ryukobayashi:
### What changes were proposed in this pull request?

See the: https://issues.apache.org/jira/browse/HIVE-25043

### Why are the changes needed?

That want to use a custom UDF that supports Vectorized when hive.vectorized.adaptor.usage.mode=chosen is specified.

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

Remote cluster.

PR #5703, opened 2025-03-19
HIVE-28830: fix HIVE-25043 Support custom UDF in Vectorized mode

ryukobayashi:

See the https://issues.apache.org/jira/browse/HIVE-28830

The https://github.com/apache/hive/pull/5631 fix alone didn't work, so I fixed it further.

No.

No.

Remote cluster.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
@@ -277,18 +277,6 @@ import com.google.common.annotations.VisibleForTesting;
 
   private HiveVectorIfStmtMode hiveVectorIfStmtMode;
 
-  private Set<String> allowedCustomUDFs;
-
-  private Set<String> getAllowedCustomUDFs(HiveConf hiveConf) {
-    String udfs = HiveConf.getVar(hiveConf,
-        HiveConf.ConfVars.HIVE_VECTOR_ADAPTOR_CUSTOM_UDF_WHITELIST);
-    if (udfs != null && !udfs.isEmpty()) {
-      return new HashSet<>(Arrays.asList(udfs.split(",")));
-    }
-
-    return new HashSet<>();
-  }
-
   //when set to true use the overflow checked vector expressions
   private boolean useCheckedVectorExpressions;
 
@@ -310,7 +298,6 @@ import com.google.common.annotations.VisibleForTesting;
     adaptorSuppressEvaluateExceptions =
         HiveConf.getBoolVar(
             hiveConf, HiveConf.ConfVars.HIVE_VECTORIZED_ADAPTOR_SUPPRESS_EVALUATE_EXCEPTIONS);
-    this.allowedCustomUDFs = getAllowedCustomUDFs(hiveConf);
   }
 
   private void copyHiveConfVars(VectorizationContext vContextEnvironment) {
@@ -1050,7 +1037,7 @@ import com.google.common.annotations.VisibleForTesting;
                 "Could not vectorize expression (mode = " + mode.name() + "): " + exprDesc.toString()
                   + " because hive.vectorized.adaptor.usage.mode=none");
           case CHOSEN:
-            if (isNonVectorizedPathUDF(expr, mode, allowedCustomUDFs)) {
+            if (isNonVectorizedPathUDF(expr, mode)) {
               ve = getCustomUDFExpression(expr, mode);
             } else {
               throw new HiveException(
@@ -1460,7 +1447,7 @@ import com.google.common.annotations.VisibleForTesting;
    * may be implemented in the future with an optimized VectorExpression.
    */
   private static boolean isNonVectorizedPathUDF(ExprNodeGenericFuncDesc expr,
-      VectorExpressionDescriptor.Mode mode, Set<String> allowCustomUDFs) {
+      VectorExpressionDescriptor.Mode mode) {
     GenericUDF gudf = expr.getGenericUDF();
     if (gudf instanceof GenericUDFBridge) {
       GenericUDFBridge bridge = (GenericUDFBridge) gudf;
@@ -1497,8 +1484,6 @@ import com.google.common.annotations.VisibleForTesting;
     } else if (gudf instanceof GenericUDFBetween && (mode == VectorExpressionDescriptor.Mode.PROJECTION)) {
       return true;
     } else if (gudf instanceof GenericUDFConcat && (mode == VectorExpressionDescriptor.Mode.PROJECTION)) {
-      return true;
-    } else if (allowCustomUDFs.contains(gudf.getClass().getName())) {
       return true;
     }
     return false;
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
@@ -2412,6 +2412,20 @@ public class Vectorizer implements PhysicalPlanResolver {
     }
   }
 
+  private void setAllowAdditionalUDFs(HiveConf hiveConf) {
+    String[] udfs =
+        HiveConf.getTrimmedStringsVar(hiveConf, HiveConf.ConfVars.HIVE_VECTOR_ADAPTOR_CUSTOM_UDF_WHITELIST);
+    if (udfs != null) {
+      for (String udf : udfs) {
+        try {
+          supportedGenericUDFs.add(Class.forName(udf));
+        } catch (ClassNotFoundException e) {
+          LOG.warn("Additional UDF not found: " + udf);
+        }
+      }
+    }
+  }
+
   @Override
   public PhysicalContext resolve(PhysicalContext physicalContext) throws SemanticException {
 
@@ -2535,6 +2549,9 @@ public class Vectorizer implements PhysicalPlanResolver {
     isTestVectorizationSuppressExplainExecutionMode =
         HiveConf.getBoolVar(hiveConf,
             HiveConf.ConfVars.HIVE_TEST_VECTORIZATION_SUPPRESS_EXPLAIN_EXECUTION_MODE);
+
+    // Add user custom UDFs
+    setAllowAdditionalUDFs(hiveConf);
 
     // create dispatcher and graph walker
     SemanticDispatcher disp = new VectorizationDispatcher();


deniskuzZ:
@ryukobayashi , please add a test to validate the fix

ryukobayashi:
@deniskuzZ Sorry, I overlooked the comment. I added test.

okumin:
@ryukobayashi
This commit didn't add any tests. Could you please recheck your local branch?
https://github.com/apache/hive/pull/5703/commits/54b08d719640c7c88caa2afcf4d6ab3ea5d639d3

ql/src/test/queries/clientpositive/custom_udf_vectorization.q
```diff
@@ -0,0 +1,21 @@
+--! qt:dataset:src
+set hive.mapred.mode=nonstrict;
+set hive.explain.user=false;
+
+CREATE TEMPORARY FUNCTION CDATE_SUB AS 'org.apache.hadoop.hive.ql.udf.generic.GenericUDFCustomDateSub';
+
+set hive.vectorized.adaptor.custom.udf.whitelist=org.apache.hadoop.hive.ql.udf.generic.GenericUDFCustomDateSub;
+
+EXPLAIN
```
deniskuzZ, line 9: maybe better use `EXPLAIN VECTORIZED`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2411,21 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
     }
   }
 
+  private void setAllowAdditionalUDFs(HiveConf hiveConf) {
```
deniskuzZ, line 2414: please rename `addCustomUDFs`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2411,21 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
...
+    String[] udfs =
+        HiveConf.getTrimmedStringsVar(hiveConf, HiveConf.ConfVars.HIVE_VECTOR_ADAPTOR_CUSTOM_UDF_WHITELIST);
+    if (udfs != null) {
+      for (String udf : udfs) {
+        try {
+          supportedGenericUDFs.add(Class.forName(udf));
+          LOG.info("Vectorized additional custom UDF: " + udf);
```
deniskuzZ, line 2421: ````LOG.info(Registered custom UDF: {}, udf)````

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2535,6 +2550,9 @@ public PhysicalContext resolve(PhysicalContext physicalContext) throws SemanticE
         HiveConf.getBoolVar(hiveConf,
             HiveConf.ConfVars.HIVE_TEST_VECTORIZATION_SUPPRESS_EXPLAIN_EXECUTION_MODE);
 
+    // Add user custom UDFs
+    setAllowAdditionalUDFs(hiveConf);
```
deniskuzZ, line 2554: Why not do this in the `Vectorizer` constructor?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2411,21 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
     }
   }
 
+  private void setAllowAdditionalUDFs(HiveConf hiveConf) {
+    String[] udfs =
+        HiveConf.getTrimmedStringsVar(hiveConf, HiveConf.ConfVars.HIVE_VECTOR_ADAPTOR_CUSTOM_UDF_WHITELIST);
```
deniskuzZ, line 2416: Maybe add `GENERIC` to the property name, since you can't register here aggregation UDFs

ql/src/test/queries/clientpositive/custom_udf_vectorization.q
ryukobayashi: Ok, I think so too. I fixed it.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
ryukobayashi: I fixed.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
ryukobayashi: The `Vectorizer` constructor does not have the `hiveConf` set yet. This method first sets the `hiveConf`.

deniskuzZ: but you can add it

ryukobayashi: OK, I changed the call destination to the constructor.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2415,21 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
...
+  private void addCustomUDFs(HiveConf hiveConf) {
+    String[] udfs =
+        HiveConf.getTrimmedStringsVar(hiveConf, HiveConf.ConfVars.HIVE_VECTOR_ADAPTOR_CUSTOM_UDF_WHITELIST);
+    if (udfs != null) {
+      for (String udf : udfs) {
+        try {
+          supportedGenericUDFs.add(Class.forName(udf));
+          LOG.info("Registered custom UDF: {}", udf);
+        } catch (ClassNotFoundException e) {
+          LOG.warn("Additional UDF not found: {}", udf);
```
deniskuzZ, line 2427: please don't swallow the exception:
```LOG.warn("Failed to register custom UDF: {}", udf, e)````

**Review by deniskuzZ — COMMENTED** [review]
LGTM, thanks @ryukobayashi, please take a look at the remaining comments

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java (hunk 6 above)
Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java]: I fixed.

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java [hunk 7]
```diff
@@ -550,6 +551,9 @@ public Vectorizer() {
 
     // For conditional expressions
     supportedGenericUDFs.add(GenericUDFIf.class);
+
+    // Add user custom UDFs
+    addCustomUDFs(SessionState.get().getConf());
```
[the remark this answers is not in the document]
ryukobayashi: Thanks, I fixed.

ryukobayashi:
@deniskuzZ Thanks, I fixed the comment part.

deniskuzZ (approved):
+1, pending tests

deniskuzZ:
@ryukobayashi, have you checked https://issues.apache.org/jira/browse/HIVE-4961?
````
Completed working version of bridge to allow custom UDFs that are subclasses
of UDF to work in vectorized mode. This supports UDFs with evaluate() methods
that take and return boxed types (e.g. Long), Writable types (e.g. LongWritable)
and standard types (e.g. long). Generic UDFs are not supported. That will be the
subject of a future patch.

I did manual testing for a large set of UDFs taking and returning the types supported
by vectorization: tinyint, smallint, int, bigint, float, double, boolean, string, timestamp.

UDFs one argument and multiple arguments were tested. Both constant and variable arguments
were tested.
````
 Your change covers the  Generic UDFs, right?

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java [hunk 8]
```diff
@@ -550,6 +551,9 @@ public Vectorizer() {
 
     // For conditional expressions
     supportedGenericUDFs.add(GenericUDFIf.class);
+
+    // Add user custom UDFs
+    addCustomUDFs(SessionState.getSessionConf());
```
deniskuzZ, line 556: @ryukobayashi, you should only add custom UDFs when `hive.vectorized.adaptor.usage.mode`=chosen

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
```diff
@@ -1050,7 +1037,7 @@ public VectorExpression getVectorExpression(ExprNodeDesc exprDesc, VectorExpress
                 "Could not vectorize expression (mode = " + mode.name() + "): " + exprDesc.toString()
                   + " because hive.vectorized.adaptor.usage.mode=none");
           case CHOSEN:
-            if (isNonVectorizedPathUDF(expr, mode, allowedCustomUDFs)) {
+            if (isNonVectorizedPathUDF(expr, mode)) {
```
deniskuzZ, line 1040: Why doesn't it throw an exception after revert? `isNonVectorizedPathUDF()` returns true?
What if you create a UDF that extends GenericUDF, not an existing one like GenericUDFDateAdd?

ryukobayashi:
@deniskuzZ 
> Your change covers the Generic UDFs, right?
Right, we have implemented many Generic UDFs that are compatible with vectorize, but there was no way to import them.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
ryukobayashi: Yes, I will implement it.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizationContext.java
ryukobayashi: > isNonVectorizedPathUDF() returns true?

Yes, it should return true because it targets a UDF that extends GenericUDF. So, I will incorporate the implementation you proposed. The reason I created allowedCustomUDFs was because even when I implemented VectorizedExpressions so that a UDF that extends GenericUDF could be executed using vectorize, vectorize was not enabled and the UDF would be executed in normal mode.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2415,39 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
...
+
+    String[] udfs =
+        HiveConf.getTrimmedStringsVar(hiveConf, HiveConf.ConfVars.HIVE_VECTOR_ADAPTOR_CUSTOM_UDF_WHITELIST);
+    if (udfs == null) {
+      return;
+    }
+
+    HiveVectorAdaptorUsageMode hiveVectorAdaptorUsageMode =
+        HiveVectorAdaptorUsageMode.getHiveConfValue(hiveConf);
+    if (hiveVectorAdaptorUsageMode != HiveVectorAdaptorUsageMode.CHOSEN &&
```
deniskuzZ, line 2431: Why do we add custom UDFs for `HiveVectorAdaptorUsageMode.ALL` mode?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2415,39 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
...
+    }
+
+    for (String udf : udfs) {
+      try {
+        Class<?> cls = Class.forName(udf);
+        if (GenericUDF.class.isAssignableFrom(cls)) {
+          supportedGenericUDFs.add(cls);
+          LOG.info("Registered custom UDF: {}", udf);
+        } else {
+          LOG.warn("Custom UDF is not extends GenericUDF: {}", udf);
```
deniskuzZ, line 2443: Custom UDF class must inherit from the GenericUDF

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2415,39 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
...
+    HiveVectorAdaptorUsageMode hiveVectorAdaptorUsageMode =
+        HiveVectorAdaptorUsageMode.getHiveConfValue(hiveConf);
+    if (hiveVectorAdaptorUsageMode != HiveVectorAdaptorUsageMode.CHOSEN &&
+        hiveVectorAdaptorUsageMode != HiveVectorAdaptorUsageMode.ALL) {
+      return;
+    }
+
+    for (String udf : udfs) {
+      try {
+        Class<?> cls = Class.forName(udf);
```
deniskuzZ, line 2438: Should we use "JavaUtils.loadClass(udf)" ?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
ryukobayashi: I'm sure only `HiveVectorAdaptorUsageMode.CHOSEN` is no problem. I will fix.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
ryukobayashi: I will fix.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2415,38 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
...
+        if (GenericUDF.class.isAssignableFrom(cls)) {
+          supportedGenericUDFs.add(cls);
+          LOG.info("Registered custom UDF: {}", udf);
+        } else {
+          LOG.warn("{} must inherit from the GenericUDF", udf);
```
deniskuzZ, line 2442: please update the HiveConf description with that more:
````
    HIVE_VECTOR_ADAPTOR_CUSTOM_UDF_WHITELIST("hive.vectorized.adaptor.custom.udf.whitelist", "",
        "A comma-separated list of custom UDFs allowed to operate in vectorized mode " +
        "when hive.vectorized.adaptor.usage.mode is set to chosen.\n" +
        "Only Generic UDFs are supported for whitelisting; ensure that each custom UDF class extends GenericUDF"),
````

Comment (ryukobayashi) [reply ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java:2442]: Got it. will do.

File: ql/src/test/results/clientpositive/tez/custom_udf_vectorization.q.out [hunk 14]
```diff
@@ -0,0 +1,238 @@
+PREHOOK: query: CREATE TEMPORARY FUNCTION CDATE_SUB AS 'org.apache.hadoop.hive.ql.udf.generic.GenericUDFCustomDateSub'
```
deniskuzZ, line 1: @ryukobayashi, please use `TestMiniLlapLocalCliDriver` to generate the q,out, jenkins failed to find the q.out file

ryukobayashi: @deniskuzZ As I've noticed before, the following issues occur when using `TestMiniLlapLocalCliDriver`:
```
org.apache.hive.com.esotericsoftware.kryo.kryo5.KryoException: java.lang.SecurityException: UDF org.apache.hadoop.hive.ql.udf.generic.GenericUDFCustomDateSub is not allowed
```

I believe this is correct, but it is not the purpose of this time, so I have added the following fixing.
https://github.com/apache/hive/blob/d8fc50f2d2396d3a99407d0c00c18a91096c7b90/itests/src/test/resources/testconfiguration.properties#L21

deniskuzZ: @ryukobayashi, we should make it work with `TestMiniLlapLocalCliDriver` that is close to the production setup

ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java
```diff
@@ -2411,6 +2415,38 @@ private boolean getOnlyStructObjectInspectors(ReduceWork reduceWork,
     }
   }
 
+  private void addCustomUDFs(HiveConf hiveConf) {
```
deniskuzZ, line 2418: Use the session classloader and return fast when mode is CHOSEN
````
  private void addCustomUDFs(HiveConf hiveConf) {
    if (hiveConf == null) {
      return;
    }
    if (HiveVectorAdaptorUsageMode.CHOSEN != HiveVectorAdaptorUsageMode.getHiveConfValue(hiveConf)) {
      return;
    }
    String[] udfs = HiveConf.getTrimmedStringsVar(hiveConf, 
        HiveConf.ConfVars.HIVE_VECTOR_ADAPTOR_CUSTOM_UDF_WHITELIST);
    if (udfs == null) {
      return;
    }
    ClassLoader loader = Utilities.getSessionSpecifiedClassLoader();
    
    for (String udf : udfs) {
      try {
        Class<?> udfClass = Class.forName(udf, true, loader);
        if (GenericUDF.class.isAssignableFrom(udfClass)) {
          supportedGenericUDFs.add(udfClass);
        } else {
          LOG.warn("{} is not a valid GenericUDF class", udf);
        }
      } catch (ClassNotFoundException e) {
        LOG.warn("Unable to load UDF class: {}", udf, e);
      }
    }
  }
````

deniskuzZ (approved):
+1, pending tests

ryukobayashi:
@deniskuzZ Thanks for the review and merge! The ticket is still open. Can I close it?

PR #5704, opened 2025-03-19
HIVE-28831 : FS Initialisation for SerDeEncodedDataReader & VectorDeserializeOrcWriter should be based on jobConf

AnmolSun:

Similar to what is currently being done in _OrcEncodedDataReader_, FS Initialisation for _SerDeEncodedDataReader_ & _VectorDeserializeOrcWriter_ should be based on _jobConf_ rather than _daemonConf_. 

This will ensure that the _FileSystems_ are being initiated with the configurations being passed as part of _jobConf_ which could be different on a Job basis instead of LLAP IO's _daemonConf_.

No

No

Passing configs by setting them in Hive code, when using LLAP IO, these configs are not being honored by the FS being initialized by _SerDeEncodedDataReader_. This is because _SerDeEncodedDataReader_ uses LLAP IO's daemon config, rather than the _jobConf_ (which typically contains custom set of configs as well).


okumin (approved):
LGTM. I reran CI for flaky tests

PR #5706, opened 2025-03-20
HIVE-28832: Disable SharedWorkOptimization in hybridgrace_hashjoin_2

ngsg:

Disable SharedWorkOptimization(`hive.optimize.shared.work`) in `hybridgrace_hashjoin_2.q`.

After applying HIVE-26986, the CI often fails to run `hybridgrace_hashjoin_2.q` due to OOM. Since the OOM comes from the merged Map vertex, we temporarily disable SWO in order to prevent the failure of the CI.

No

No

This patch addresses the issue in the CI, so I can't test the patch in my local environment. I will check whether this patch actually prevents OOM in the CI environment.


deniskuzZ (approved):
+1

PR #5707, opened 2025-03-20
HIVE-27078: Bucket Map Join can hang if the source vertex parallelism is changed by reducer autoparallelism

mayankkunwar:
…is changed by reducer autoparallelism

Disable auto reduce parallelism for Bucket Map Join in compile time until the issue is fixed by TEZ-4603

Bucket Map Join can hang if the source vertex parallelism is changed by reducer autoparallelism

No

No

qtest


deniskuzZ:
cc @okumin, since he is an expert in that area

ql/src/java/org/apache/hadoop/hive/ql/optimizer/AbstractBucketJoinProc.java
```diff
@@ -451,6 +452,8 @@ protected void convertMapJoinToBucketMapJoin(
 
     // successfully convert to bucket map join
     desc.setBucketMapJoin(true);
+    // Disabling TEZ_AUTO_REDUCER_PARALLELISM until TEZ-4603 is fixed.
+    context.getConf().setBoolVar(HiveConf.ConfVars.TEZ_AUTO_REDUCER_PARALLELISM, false);
```
ngsg, line 456: `AbstractBucketJoinProc` is used by `SortMergeBucketMapJoinOptimizer` and `BucketMapJoinOptimizer`, which are only applied when the engine is in MapReduce mode. Therefore, I think we don't need to modify the Configuration here.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
```diff
@@ -701,6 +701,8 @@ private boolean convertJoinBucketMapJoin(JoinOperator joinOp, OptimizeTezProcCon
     }
     MapJoinDesc joinDesc = mapJoinOp.getConf();
     joinDesc.setBucketMapJoin(true);
+    // Disabling TEZ_AUTO_REDUCER_PARALLELISM until TEZ-4603 is fixed.
+    context.conf.setBoolVar(ConfVars.TEZ_AUTO_REDUCER_PARALLELISM, false);
```
ngsg, line 705: Quick question: Is the `Configuration` specific to this query? If changing it affects other queries, it could disable AutoParallelism for all of them.

ql/src/test/queries/clientpositive/bucketmapjoin_auto_reduce_parallel.q
```diff
@@ -0,0 +1,29 @@
+DROP TABLE IF EXISTS source_table2;
+DROP TABLE IF EXISTS target_table2;
```
ngsg, line 2: The `DROP TABLE ...` statements are not necessary in the qfile since CliDrivers automatically clean up all tables after each qfile test.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/AbstractBucketJoinProc.java
mayankkunwar: I see. Will remove it this conf from here. Thanks

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
mayankkunwar: The idea is to just disable this only when Bucket Map Join occurs.

ql/src/test/queries/clientpositive/bucketmapjoin_auto_reduce_parallel.q
mayankkunwar: Forgot to remove the drop statements. Will remove it.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
mayankkunwar: Maybe I need to implement ExecuteWithHookContext to restore the configuration value and use SessionState.get().getConf() to get and set the conf

ql/src/test/queries/clientpositive/bucketmapjoin_auto_reduce_parallel.q
```diff
@@ -0,0 +1,29 @@
...
+('2019-10-15', 'pipeline', '50000000000000000001598843430'), ('2020-04-01', 'pipeline', '50000000000000000002035795461'), ('2020-02-24', 'pipeline', '50000000000000000001932600185'), ('2020-04-27', 'pipeline', '50000000000000000002108160849'), ('2016-07-05', 'pipeline', '50000000000000000000054405114'), ('2020-06-02', 'pipeline', '50000000000000000002234387967'),
+('2020-08-21', 'pipeline', '50000000000000000002529168758'), ('2021-02-17', 'pipeline', '50000000000000000003158511687');
+
+set hive.auto.convert.join=true;
+set hive.optimize.dynamic.partition.hashjoin=false;
+set hive.convert.join.bucket.mapjoin.tez=true;
+set hive.vectorized.execution.enabled=false;
+
+set hive.optimize.bucketmapjoin=true;
+set hive.tez.dynamic.partition.pruning=true; -- Required for dynamic parallelism
```
ngsg, line 22: NIT: Could we remove this line as `hive.tez.dynamic.partition.pruning` is set to true by default? (Also, as far as I know, DPP makes sense when we join a table with a partitioned table.)

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
mayankkunwar: thanks for the verification

ql/src/test/queries/clientpositive/bucketmapjoin_auto_reduce_parallel.q
mayankkunwar: hive.tez.dynamic.partition.pruning is true by default, sure will remove it

ql/src/test/queries/clientpositive/bucketmapjoin_auto_reduce_parallel.q
```diff
@@ -0,0 +1,24 @@
...
+('2022-08-30', 'pipeline', '50000000000000000005905545593'), ('2022-08-16', 'pipeline', '50000000000000000005905545593'), ('2018-05-03', 'pipeline', '50000000000000000000750826355'), ('2020-01-10', 'pipeline', '50000000000000000001816579677'), ('2021-11-01', 'pipeline', '50000000000000000004269423714'), ('2017-11-07', 'pipeline', '50000000000000000000585901787'),
+('2019-10-15', 'pipeline', '50000000000000000001598843430'), ('2020-04-01', 'pipeline', '50000000000000000002035795461'), ('2020-02-24', 'pipeline', '50000000000000000001932600185'), ('2020-04-27', 'pipeline', '50000000000000000002108160849'), ('2016-07-05', 'pipeline', '50000000000000000000054405114'), ('2020-06-02', 'pipeline', '50000000000000000002234387967'),
+('2020-08-21', 'pipeline', '50000000000000000002529168758'), ('2021-02-17', 'pipeline', '50000000000000000003158511687');
+
+set hive.auto.convert.join=true;
+set hive.optimize.dynamic.partition.hashjoin=false;
+set hive.convert.join.bucket.mapjoin.tez=true;
+set hive.vectorized.execution.enabled=false;
```
okumin, line 16: We might not need `set hive.optimize.dynamic.partition.hashjoin=false;` and `set hive.convert.join.bucket.mapjoin.tez=true;` because [we use default values](https://github.com/apache/hive/blob/master/data/conf/llap/hive-site.xml).
Why do we have to disable `hive.vectorized.execution.enabled`?

ql/src/test/queries/clientpositive/bucketmapjoin_auto_reduce_parallel.q
```diff
@@ -0,0 +1,24 @@
...
+
+set hive.optimize.bucketmapjoin=true;
```
okumin, line 18: I think we can remove this line because only Hive on MR uses this parameter

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
okumin: How about removing the AUTOPARALLEL trait? Auto-reducer parallelism is enabled [only when a ReduceSinkOperator has AUTOPARALLEL](https://github.com/apache/hive/blob/2d528556b69c1ec011e83f69d2a2fdcfb78b356e/ql/src/java/org/apache/hadoop/hive/ql/parse/GenTezUtils.java#L125). So, we may remove the trait from all parent RSOs of MapJoinOperator.

The current approach will likely work. However, it might be surprising to mutate a configuration here, and it will likely disable Auto-Reducer Parallelism on all vertices.

ql/src/test/queries/clientpositive/bucketmapjoin_auto_reduce_parallel.q
mayankkunwar: okay will remove the config which is default

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
mayankkunwar: But at this point how do we know if it is a bucket map join?

mayankkunwar: Only for bucket map join we need to disable auto reduce parallelism.

mayankkunwar: @okumin I have updated the code as per your review comment

ql/src/java/org/apache/hadoop/hive/ql/parse/GenTezUtils.java
```diff
@@ -182,6 +187,30 @@ public static ReduceWork createReduceWork(
...
+        MapJoinOperator mjOp = (MapJoinOperator) childOp;
+        if (mjOp.getConf().isBucketMapJoin()) {
+          return true; // Found BMJ, no need to check further
+        }
+      }
+
+      // Recursively check children
+      if (hasBucketMapJoin(childOp)) {
+        return true;
+      }
```
okumin, line 208: Do we have to check it recursively?

```
                   TS -> BMJ
                                 ^
                                  |
TS -> RS1 -> RS2 --|
```

I currently guess we can enable auto reducer parallelism for RS1. So, we may have to check BMJ's direct parent, i.e., RS2.

mayankkunwar: Yes I had to put this recursion, as for the attached QTest, when I called the hasBucketMapJoin(reduceSink) method, found out the hierarchy as follows:
**RS[7] -> GBY[8] -> RS[11] -> MAPJOIN[49]**

And when checking on **MAPJOIN[49]** isBucketMapJoin(), it returned true. So without recursion how would we know this?

okumin: Let me understand the root cause correctly. I might be misunderstanding. Let's assume the following query.

```sql
select s.string_col, count(*)
from target_table2 t
inner join (
  select min(date_col) date_col, string_col, decimal_col
  from (
    select date_col, 'pipeline' string_col, min(decimal_col) decimal_col
    from source_table2
    where coalesce(decimal_col,'') = '50000000000000000005905545593'
    group by date_col, string_col
  ) x
  group by string_col, decimal_col
) s
on s.date_col = t.date_col AND s.string_col = t.string_col AND s.decimal_col = t.decimal_col
group by s.string_col;
```

Without HIVE-27078, auto-reducer parallelism(the green cells) is enabled on all the reducers.
image

The original patch tried to turn off `hive.tez.auto.reducer.parallelism`. Auto-parallelism would be totally disabled. It would resolve the hang issue but unrelated reducers would also be affected.
image

The current patch will likely make the following DAG. It is better, but Reducer 4 will still run without auto-reducer parallelism.
image

In my understanding, the ideal DAG should have this shape.
image

To achieve the final shape, we may judge the current ReduceWork should be auto-parallelized in the following way, but I don't care how to do. Just an example.
1. Find ReduceSinkOperators directly included in the vertex
2. Check the direct children of ReduceSinkOperators include BucketMapJoin

mayankkunwar: @okumin Thank you for your detailed explanation and the accompanying diagram. It was extremely helpful in visualizing the concept, and I now have a clear understanding of what you were trying to convey. 

But in this query execution, when I try to debug, I see that **createReduceWork** method is called only once, and for "**Reducer 4**" according to the mentioned diagram. Why this method is not called for other reducers?

So, if I remove recursion from the changes,  the query will hang.
Seems like if I remove AUTOPARALLEL for all the reducer before BucketMapJoin, the query execution do not hang.

mayankkunwar: TEZ Magic after Merge (Final Optimized DAG):
Map 3 → Merged Reducer (4 & 5) → Merged Vertex (Map 1 + Reducer 2)

ngsg: I was able to generate the above DAG using the following query. Could you try this with `TestMiniLlapLocalCliDriver`?
``` sql
create table source_table2(date_col date, string_col string, decimal_col decimal(38,0)) clustered by (decimal_col) into 7 buckets;
insert into table source_table2 values
('2022-08-30', 'pipeline', '50000000000000000005905545593'), ('2022-08-16', 'pipeline', '50000000000000000005905545593'), ('2022-09-01', 'pipeline', '50000000000000000006008686831'), ('2022-08-30', 'pipeline', '50000000000000000005992620837'), ('2022-09-01', 'pipeline', '50000000000000000005992620837'), ('2022-09-01', 'pipeline', '50000000000000000005992621067'),
('2022-08-30', 'pipeline', '50000000000000000005992621067');

create table target_table2(date_col date, string_col string, decimal_col decimal(38,0)) clustered by (decimal_col) into 7 buckets;
insert into table target_table2 values
('2017-05-17', 'pipeline', '50000000000000000000441610525'), ('2018-12-20', 'pipeline', '50000000000000000001048981030'), ('2020-06-30', 'pipeline', '50000000000000000002332575516'), ('2021-08-16', 'pipeline', '50000000000000000003897973989'), ('2017-06-06', 'pipeline', '50000000000000000000449148729'), ('2017-09-08', 'pipeline', '50000000000000000000525378314'),
('2022-08-30', 'pipeline', '50000000000000000005905545593'), ('2022-08-16', 'pipeline', '50000000000000000005905545593'), ('2018-05-03', 'pipeline', '50000000000000000000750826355'), ('2020-01-10', 'pipeline', '50000000000000000001816579677'), ('2021-11-01', 'pipeline', '50000000000000000004269423714'), ('2017-11-07', 'pipeline', '50000000000000000000585901787'),
('2019-10-15', 'pipeline', '50000000000000000001598843430'), ('2020-04-01', 'pipeline', '50000000000000000002035795461'), ('2020-02-24', 'pipeline', '50000000000000000001932600185'), ('2020-04-27', 'pipeline', '50000000000000000002108160849'), ('2016-07-05', 'pipeline', '50000000000000000000054405114'), ('2020-06-02', 'pipeline', '50000000000000000002234387967'),
('2020-08-21', 'pipeline', '50000000000000000002529168758'), ('2021-02-17', 'pipeline', '50000000000000000003158511687');

 set hive.auto.convert.join=true;

set hive.tez.auto.reducer.parallelism=true;
set hive.tez.min.partition.factor=12;
set hive.tez.max.partition.factor=50;

explain extended
select s.string_col, count(*)
from target_table2 t
inner join (
  select min(date_col) date_col, string_col, decimal_col
  from (
    select date_col, 'pipeline' string_col, min(decimal_col) decimal_col
    from source_table2
    where coalesce(decimal_col,'') = '50000000000000000005905545593'
    group by date_col, string_col
  ) x
  group by string_col, decimal_col
) s
on s.date_col = t.date_col AND s.string_col = t.string_col AND s.decimal_col = t.decimal_col
group by s.string_col;

```

mayankkunwar: Getting the following final DAG for **EXPLAIN** query:
```
      Edges:
        Map 1 <- Reducer 3 (CUSTOM_EDGE)
        Reducer 3 <- Map 2 (SIMPLE_EDGE)
```

mayankkunwar: The query mentioned by @okumin is working with or without these changes. But the query mentioned in this PR is the problematic one.

ngsg: `createReduceWork()` is called whenever it meets a new root operator that has any proceeding work(=vertex), so the number of RS operators and `createReduceWork()` calls is not equal.

In the opeator graph `RS[7] -> GBY[8] -> RS[11] -> MAPJOIN[49]`, the `GBY[8]` would trigger `createReduceWork()` since it is a root operator and it has a proceeding work `TS?-...-RS[7]`. However, `MapJoin[49]` would not trigger `createReduceWork()` as it is not a root operator; it would belong to another Map work something like `TS??-...-MapJoin[49]`.

mayankkunwar: I see, thanks! Let me add the missing changes

mayankkunwar:
@okumin @ngsg I have implemented logic to bypass the check if a reducer is present without an associated MapJoin operator. This ensures that the check is skipped in such cases, allowing auto-parallelism to remain enabled.

ql/src/test/results/clientpositive/llap/bucket_map_join_tez3.q.out
```diff
@@ -1432,7 +1432,7 @@ STAGE PLANS:
                         Map-reduce partition columns: _col0 (type: date), _col1 (type: decimal(38,0))
                         Statistics: Num rows: 1 Data size: 168 Basic stats: COMPLETE Column stats: COMPLETE
                         tag: -1
-                        auto parallelism: true
+                        auto parallelism: false
```
okumin, line 1435: Looks good.

Map 2 -> Reducer 3 -> Map 1(BMJ)

ql/src/java/org/apache/hadoop/hive/ql/parse/GenTezUtils.java
```diff
@@ -182,6 +187,41 @@ public static ReduceWork createReduceWork(
     return reduceWork;
   }
 
+  /**
+   * Checks if there is a Bucket Map Join (BMJ) following a hierarchy:
+   * ReduceSinkOperator (RS) -> ReduceSinkOperator (RS) -> MapJoinOperator (MJ).
+   * Ensures at most **two** RSOs before MJ.
```
okumin, line 193: I guess it happens only when there are precisely two RSOs. Given that Map 3 -> Reducer 4 -> Reducer 5 -> Map 1(BMJ), `createReduceWork` is invoked twice.
When we create Reducer 4, the passed RSO belongs to Map 3. The process bails out once we find the third RSO in Reducer 5.
When we create Reducer 5, we find two RSOs in Reduer 4 and Reducer 5, and the BucketMapJoin.


A vertex with BucketMapJoin is a mapper. So, basically, RS -> MJ doesn't happen.

okumin (approved):
+1 though we can fix the logic a bit. We should also address TEZ-4603

ql/src/java/org/apache/hadoop/hive/ql/parse/GenTezUtils.java
mayankkunwar: Thanks

PR #5708, opened 2025-03-20
HIVE-28836: Hive SessionState Resource directories persists if SessionState is not started

Indhumathi27:
Removed creating resource directories on initialising session state. org.apache.hadoop.hive.ql.util.ResourceDownloader#downloadResource will ensure creating directories when needed.

Previously, resource directories were created outside of SessionState.start, leading to inconsistencies and leftover directories when the session state was not properly closed. This issue was particularly noticeable with Spark SQL, as it resulted in the accumulation of unused session resource directories, consuming disk space and potentially affecting performance. 

Check [SPARK-51571](https://issues.apache.org/jira/browse/SPARK-51571) for reference

Fixing this issue in Hive itself ensures consistent behaviour across all engines that use SessionState, including Spark, making it a more robust and centralized solution.

no

no

tested in cluster and existing testcases

### DIFF
--- a/llap-server/src/java/org/apache/hadoop/hive/llap/cli/service/AsyncTaskCreateUdfFile.java
+++ b/llap-server/src/java/org/apache/hadoop/hive/llap/cli/service/AsyncTaskCreateUdfFile.java
@@ -115,7 +115,7 @@ class AsyncTaskCreateUdfFile implements Callable<Void> {
       }
     }
     for (URI srcUri : srcUris) {
-      List<URI> localUris = resourceDownloader.downloadExternal(srcUri, null, false);
+      List<URI> localUris = resourceDownloader.downloadExternal(srcUri, null);
       for(URI dst : localUris) {
         LOG.warn("Downloaded " + dst + " from " + srcUri);
       }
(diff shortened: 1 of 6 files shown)


okumin:
We have two more use cases of `ResourceDownloader`.

### AsyncTaskCreateUdfFile

[The class uses ResourceDownloader to download UDF files to upload them as one of LLAP artiracts](https://github.com/apache/hive/blob/3b562f7fefbb34ce6c191d23325d9f0c882271b2/llap-server/src/java/org/apache/hadoop/hive/llap/cli/service/AsyncTaskCreateUdfFile.java#L98). The directory is [explicitly created outside ResourceDownloader](https://github.com/apache/hive/blob/3b562f7fefbb34ce6c191d23325d9f0c882271b2/llap-server/src/java/org/apache/hadoop/hive/llap/cli/service/LlapTarComponentGatherer.java#L69-L79). This patch will unlikely cause a new issue.

### FunctionLocalizer

[The class uses ResourceDownloader to download UDFs](https://github.com/apache/hive/blob/3b562f7fefbb34ce6c191d23325d9f0c882271b2/llap-server/src/java/org/apache/hadoop/hive/llap/daemon/impl/FunctionLocalizer.java#L93). It is not created unless `ResourceDownloader#downloadResource` is used.

I slightly think we may not need to "ensureDirectory" on creating ResourceDownloader. That's because `resourceDir` is used only in `downloadResource` and the method ensures the same or more.

deniskuzZ (commented):
LGTM, but i would ask for a small cleanup. Could we please remove unused 'convertToUnix' param from `downloadResource`

deniskuzZ (approved):
LGTM +1, pending tests

PR #5709, opened 2025-03-21
HIVE-28838: Remove sensitive jdbc properties from explain plan for JdbcStorageHandler tables

InvisibleProgrammer:
It removes the following sensitive properties from tables created with JdbcStorageHandler:
- hive.sql.dbcp.username
- hive.sql.dbcp.password

Explain extended can cause a security issue if they query sys tables: Sys tables are created with JdbcStorageHandler and so that, the output of explain extended will contain sensitive jdbc connection info.

Yes, it removes the listed properties from explain extended output when a JdbcStorageHandler table is included in a query. 

No

QTest added: explain_systest_password.q


ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java
```diff
@@ -19,8 +19,7 @@
 package org.apache.hadoop.hive.ql.plan;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
-import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.META_TABLE_LOCATION;
-import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.TABLE_IS_CTAS;
+import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.*;
```
zhangbutao, line 22: Avoid using the wildcard *

ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java
```diff
@@ -1230,7 +1229,12 @@ public static Class<? extends AbstractSerDe> getDefaultSerDe() {
     return LazySimpleSerDe.class;
   }
 
-  private static final String[] FILTER_OUT_FROM_EXPLAIN = {TABLE_IS_CTAS};
+  private static final String[] FILTER_OUT_FROM_EXPLAIN = {
+          TABLE_IS_CTAS,
+          HIVE_SQL_JDBC_USERNAME,
+          HIVE_SQL_JDBC_PASSWORD,
```
zhangbutao, line 1235: I think removing `password `is enough.

ql/src/test/queries/clientpositive/explain_systest_password.q
```diff
@@ -0,0 +1,12 @@
+--! qt:sysdb
+
+select 1
+from sys.TBLS t
+    join sys.DBS d on t.DB_ID = d.DB_ID
+limit 1;
+
+explain extended
```
zhangbutao, line 8: Could we use a more simpler qtest instead?

ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java
InvisibleProgrammer: I would disagree with that. I excluded the username, password, and the JDBC URL. 
With the username, we can share some extra information about who is the admin. 
And about the JDBC URL, it provides the proper path to the metastore database.

ql/src/test/queries/clientpositive/explain_systest_password.q
InvisibleProgrammer: What would you prefer by a simpler qtest? 

It only selects a constant. And the output is limited to one row.

ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java
InvisibleProgrammer: Refactored

ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java
zratkai: I agree with you to remove all three.

zratkai (approved):
LGTM

ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java
deniskuzZ: I believe that is not a reliable fix, as an attacker could retrieve the plain user/password from a heap dump/etc. 
Why do we store that information in table properties in first place?

zhangbutao: When users query MySQL tables through Hive jdbc storage handler, users sometimes need to know which MySQL database the table comes from through JDBC URL.  Users may  execute `show create table mysql_jdbc_tbl` to check the JDBC URL.

Therefore, I think it is necessary to retain the JDBC URL & username.

deniskuzZ: another q: isn't the `SYS` schema accessed only by admins? users can only use `INFORMATION_SCHEMA` schema

deniskuzZ: > Therefore, I think it is necessary to retain the JDBC URL & username.
agree, at least it should be available to the table owners/admins

zhangbutao: > `describe formatted` / `show create table` also displays the password in plain text. I don't think it's covered here.

Yes, This PR mainly focuses on removing sensitive information in `explain`. However, it does not handle the `show create table `for jdbc storage handler tables. Currently, sensitive information can still be seen in `show create table`.

For example, if you create a mysql jdbc handler table in Hive, and then `show create table` to check its TBLPROPERTIES, you will still can see its password` "hive.sql.dbcp.password" = "passwd"`.
```
CREATE EXTERNAL TABLE jdbc_mysql_table
(
 id int
)
STORED BY 'org.apache.hive.storage.jdbc.JdbcStorageHandler'
TBLPROPERTIES (
                "hive.sql.database.type" = "MYSQL",
                "hive.sql.jdbc.driver" = "com.mysql.jdbc.Driver",
                "hive.sql.jdbc.url" = "jdbc:mysql://127.0.0.1:3306/testdb",
                "hive.sql.dbcp.username" = "root",
                "hive.sql.dbcp.password" = "passwd",
                "hive.sql.table" = "testmysql"
);
```

`show create table`  is the main use case that I am focusing on. BTW, we will check which MySQL database this table belongs to through the JDBC URL (`hive.sql.jdbc.url)`. This is also the reason why I think the JDBC URL should be retained.

ql/src/test/queries/clientpositive/explain_systest_password.q
zhangbutao: ok, it is reasonable for the `explain` case. 

If we also need to fix the "show create table" use case, then only a simple SQL command "show create table" is enough.

ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java
InvisibleProgrammer: Hi, 

This ticket came up because of an interesting behavior of explain extended in case of sys tables: 
System tables are special kind of JdbcStorageHandler tables: they are created with JdbcStorageHandler but they are not containing a jdbc connection info. Instead of that, they are using the existing connection info provided in HMS configuration. 

The interesting thing is that explain extended is a query. And for queries, Hive adds the jdbc connection info to the task (I assume because it is required to be able to run the query). 
The questionable behavior is that at sys tables, why we share the JDO connection info provided in HMS config? 

This PR addresses this issue. It is not about show create table or describe formatted at all. And, for regular JdbcStorageHandler tables, I would expect to see that information in show create table - as table is created with providing this info. And this is not the HMS metastore db connection info at all. 

So, I'm open to have a discussion or a follow up ticket to rethink what kind of information we want to share and for what kind of users. But with this PR, I want to address the case when a simple explain statement can share the HMS database connection info.

InvisibleProgrammer: I modified the qtest file to demonstrate that the problem doesn't exists for show create table or describe table - as it expected as the connection info is not part of the sys table creation.

deniskuzZ: SYS tables could be queried only by privillege users/admins, so I don't see a security risk here. 
However, it makes sense to exclude the JDO connection info from the explain plan.

standalone-metastore/metastore-common/src/gen/thrift/gen-javabean/org/apache/hadoop/hive/metastore/api/hive_metastoreConstants.java
```diff
@@ -97,4 +97,10 @@
 
   public static final java.lang.String EXPECTED_PARAMETER_VALUE = "expected_parameter_value";
 
+  public static final java.lang.String HIVE_SQL_JDBC_USERNAME = "hive.sql.dbcp.username";
```
deniskuzZ, line 100: Why do we need to define separate var for every JDBC config? 
Can't we create a list with connection properties we'd like to exclude and pass it to a filter?

InvisibleProgrammer: >  Can't we pass hive.sql regex to a filter?

It is a great question: if we add a regex pattern to filter this out, there is a chance that we can filter out something in the future that is not implemented yet. And the person who implements it, won't even know about it at all. But on the other hand, there are no huge improvements at Hive - I mean, it looks like a complete software - so I see only a little chance that it will happen. 

And also, thank you for this comment. I just now realized it is a thrift generated code. So it is definitely a bad idea to put a new constant here - it will be removed as thrift is re-generated. Let me find a proper place to store those constants.

InvisibleProgrammer:
@zhangbutao , @zratkai , @deniskuzZ , thank you for the reviews so far. Do I have any question that I didn't answer or needs more to discuss about?

Thank you

zhangbutao (approved):
+1 LGTM

I would suggest optimizing the title, and emphasizing that this PR is for the sensitive information repair related to the explain type tasks of the `sys JdbcStorageHandler` table.

ql/src/test/queries/clientpositive/explain_systest_password.q
deniskuzZ: @InvisibleProgrammer, why does it have to be a sys db?

ql/src/java/org/apache/hadoop/hive/ql/plan/PlanUtils.java
```diff
@@ -1230,7 +1233,12 @@ public static Class<? extends AbstractSerDe> getDefaultSerDe() {
     return LazySimpleSerDe.class;
   }
 
-  private static final String[] FILTER_OUT_FROM_EXPLAIN = {TABLE_IS_CTAS};
+  private static final String[] FILTER_OUT_FROM_EXPLAIN = {
+          TABLE_IS_CTAS,
+          JDBC_USERNAME,
```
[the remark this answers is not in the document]
InvisibleProgrammer: It wasn't `CreateTableLikeDesc`. However, I found it. My previous test run had two failures: 
- source.q, because it is flaky: https://github.com/apache/hive/pull/5750
- TestJdbcStorageConfigManager#testWithAllRequiredSettingsDefined. I fixed it, let's see what the pipeline says

jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java
```diff
@@ -45,6 +48,8 @@ public class JdbcStorageConfigManager {
   public static final String CONFIG_PWD_KEYSTORE = Constants.JDBC_KEYSTORE;
   public static final String CONFIG_PWD_KEY = Constants.JDBC_KEY;
   public static final String CONFIG_PWD_URI = Constants.JDBC_PASSWORD_URI;
+  public static final String CONFIG_JDBC_URL = JDBC_URL;
```
deniskuzZ, line 51: inconsistent style, don't use static imports

InvisibleProgrammer: Fixed.

jdbc-handler/src/test/java/org/apache/hive/storage/jdbc/conf/TestJdbcStorageConfigManager.java
```diff
@@ -40,9 +41,9 @@ public void testWithAllRequiredSettingsDefined() throws Exception {
     JdbcStorageConfigManager.copyConfigurationToJob(props, jobMap);
 
     assertThat(jobMap, is(notNullValue()));
-    assertThat(jobMap.size(), is(equalTo(4)));
+    assertThat(jobMap.size(), is(equalTo(3)));
     assertThat(jobMap.get(JdbcStorageConfig.DATABASE_TYPE.getPropertyName()), is(equalTo("MYSQL")));
-    assertThat(jobMap.get(JdbcStorageConfig.JDBC_URL.getPropertyName()), is(equalTo("jdbc://localhost:3306/hive")));
+    assertThat(jobMap.get(JdbcStorageConfig.JDBC_URL.getPropertyName()), is(nullValue()));
```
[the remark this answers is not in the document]
InvisibleProgrammer: It is already removing the URL. If you watch it carefully, before the change, it checked if the proper JDBC_URL is received. Now it expects null as it is removed.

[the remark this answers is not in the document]
deniskuzZ: Because your change removed it: https://github.com/apache/hive/pull/5709/files#diff-40610d6d8ad5f0c071e86342a1d86aecd602923afb8abc9be115c5f5a1c3f401R74. 
What I am saying is we shouldn't remove URL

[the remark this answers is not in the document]
InvisibleProgrammer: I added it back

jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java
```diff
@@ -66,7 +68,9 @@ public static void copyConfigurationToJob(Properties props, Map<String, String>
       if (!key.equals(CONFIG_PWD) &&
           !key.equals(CONFIG_PWD_KEYSTORE) &&
           !key.equals(CONFIG_PWD_KEY) &&
-          !key.equals(CONFIG_PWD_URI)) {
+          !key.equals(CONFIG_PWD_URI) &&
+          !key.equals(CONFIG_JDBC_USERNAME)
```
dengzhhu653, line 72: if we don't copy the user to remote job, how can the job connect to the db? i cannot see where we set the user later for the non-sys table.

deniskuzZ: I also think that would break the flow. `password` is passed through the secrets (i.e. `tableDesc.setJobSecrets(jobSecrets)`)
````
  public void configureInputJobProperties(TableDesc tableDesc, Map<String, String> jobProperties) {
      JdbcStorageConfigManager.copyConfigurationToJob(properties, jobProperties);
  }

  public void configureInputJobCredentials(TableDesc tableDesc, Map<String, String> jobSecrets) {
      JdbcStorageConfigManager.copySecretsToJob(properties, jobSecrets);
  }
````

Comment (InvisibleProgrammer) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java:72]: I did a quick check with external_jdbc_table.q: I removed the disabled comment and added an explain extended on a select that works on a jdbc table. 
It worked fine. 

I will dig a little bit deeper next week.

Comment (InvisibleProgrammer) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java:72]: Hi, I added a couple of extra steps around JdbcStorageHandler tables: `ql/src/test/queries/clientpositive/explain_jdbc.q` 

I couldn't make the test queries fail. And the sensitive information is filtered out from the explains. 

Could you please be specific about what cases you are thinking about?

Comment (deniskuzZ) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java:72]: Have you tried to create and query an external Hive table with JdbcStorageHandler? Did it work with the commented out CONFIG_JDBC_USERNAME?

Comment (InvisibleProgrammer) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java:72]: That is what I did and that is why I added the new qtests.

Comment (deniskuzZ) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java:72]: ok,  but I don't know/understand how the user is propagated.
Could you please check where we get the user when creating the jdbc connection? I don't know that part of the code.
Q-tests might not show the problem. Could you try the same on a local setup with an external postgre/mysql db as a meta-provider?

Comment (InvisibleProgrammer) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java:72]: It receives the jdbc information not from JdbcStorageHandler but from MetaStoreUtils.getTableMetadata during the query planning phase: 
Interesting thing: 

Even on master, explain analyze doesn't contain the jdbc connection info with this test: 

```
SET hive.fetch.task.conversion=none;

select * from ext_postgres_table;

explain analyze select * from ext_postgres_table;

```

Stack trace: 
```
getDeserializer:94, HiveMetaStoreUtils (org.apache.hadoop.hive.metastore)
getDeserializer:80, HiveMetaStoreUtils (org.apache.hadoop.hive.metastore)
getDeserializerFromMetaStore:357, Table (org.apache.hadoop.hive.ql.metadata)
getDeserializer:343, Table (org.apache.hadoop.hive.ql.metadata)
genTableLogicalPlan:3054, CalcitePlanner$CalcitePlannerAction (org.apache.hadoop.hive.ql.parse)
genLogicalPlan:4941, CalcitePlanner$CalcitePlannerAction (org.apache.hadoop.hive.ql.parse)
apply:1636, CalcitePlanner$CalcitePlannerAction (org.apache.hadoop.hive.ql.parse)
apply:1579, CalcitePlanner$CalcitePlannerAction (org.apache.hadoop.hive.ql.parse)
lambda$withPlanner$0:140, Frameworks (org.apache.calcite.tools)
apply:-1, 1094156964 (org.apache.calcite.tools.Frameworks$$Lambda$627)
perform:936, CalcitePrepareImpl (org.apache.calcite.prepare)
withPrepare:191, Frameworks (org.apache.calcite.tools)
withPlanner:135, Frameworks (org.apache.calcite.tools)
logicalPlan:1335, CalcitePlanner (org.apache.hadoop.hive.ql.parse)
genOPTree:587, CalcitePlanner (org.apache.hadoop.hive.ql.parse)
analyzeInternal:13170, SemanticAnalyzer (org.apache.hadoop.hive.ql.parse)
analyzeInternal:480, CalcitePlanner (org.apache.hadoop.hive.ql.parse)
analyze:336, BaseSemanticAnalyzer (org.apache.hadoop.hive.ql.parse)
analyze:224, Compiler (org.apache.hadoop.hive.ql)
compile:109, Compiler (org.apache.hadoop.hive.ql)
compile:498, Driver (org.apache.hadoop.hive.ql)
compileInternal:450, Driver (org.apache.hadoop.hive.ql)
compileAndRespond:414, Driver (org.apache.hadoop.hive.ql)
compileAndRespond:408, Driver (org.apache.hadoop.hive.ql)
compileAndRespond:126, ReExecDriver (org.apache.hadoop.hive.ql.reexec)
prepare:205, SQLOperation (org.apache.hive.service.cli.operation)
runInternal:268, SQLOperation (org.apache.hive.service.cli.operation)
run:286, Operation (org.apache.hive.service.cli.operation)
executeStatementInternal:558, HiveSessionImpl (org.apache.hive.service.cli.session)
executeStatementAsync:543, HiveSessionImpl (org.apache.hive.service.cli.session)
invoke0:-1, NativeMethodAccessorImpl (sun.reflect)
invoke:62, NativeMethodAccessorImpl (sun.reflect)
invoke:43, DelegatingMethodAccessorImpl (sun.reflect)
invoke:498, Method (java.lang.reflect)
invoke:78, HiveSessionProxy (org.apache.hive.service.cli.session)
access$000:36, HiveSessionProxy (org.apache.hive.service.cli.session)
run:63, HiveSessionProxy$1 (org.apache.hive.service.cli.session)
doPrivileged:-1, AccessController (java.security)
doAs:422, Subject (javax.security.auth)
doAs:1953, UserGroupInformation (org.apache.hadoop.security)
invoke:59, HiveSessionProxy (org.apache.hive.service.cli.session)
executeStatementAsync:-1, $Proxy125 (com.sun.proxy)
executeStatementAsync:315, CLIService (org.apache.hive.service.cli)
ExecuteStatement:652, ThriftCLIService (org.apache.hive.service.cli.thrift)
getResult:1670, TCLIService$Processor$ExecuteStatement (org.apache.hive.service.rpc.thrift)
getResult:1650, TCLIService$Processor$ExecuteStatement (org.apache.hive.service.rpc.thrift)
process:38, ProcessFunction (org.apache.thrift)
process:38, TBaseProcessor (org.apache.thrift)
process:56, TSetIpAddressProcessor (org.apache.hive.service.auth)
run:250, TThreadPoolServer$WorkerProcess (org.apache.thrift.server)
runWorker:1149, ThreadPoolExecutor (java.util.concurrent)
run:624, ThreadPoolExecutor$Worker (java.util.concurrent)
run:748, Thread (java.lang)
```

Comment (InvisibleProgrammer) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java]: Hi @deniskuzZ , is there any open question that I have to answer?

Comment (deniskuzZ) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java]: Where does the MetaStoreUtils.getTableMetadata set the user property? 
i can only find that it sets schema props: `getSchemaWithoutCols` + columns `addCols`

Comment (InvisibleProgrammer) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java]: Hi @deniskuzZ , 

There are two paths to answer your question: 
- for regular jdbc tables, it just picks that info up as other table parameters.
- for sys db, there is a special db type: METASTORE: 

MetaStoreUtils.getTableMetadata doesn't get the user property. 
But later, in a query execution, JdbcSerde is initialized: `JdbcSerde.initialize` 
And in that, it calls `JdbcStorageConfigManager.resolveMetadata` and that ads the jdbc related info to the existing props from the configuration. 
For details, see the if `if (dbType == DatabaseType.METASTORE) ` part of that class. 

The related stack trace is: 
```
checkRequiredPropertiesAreDefined:150, JdbcStorageConfigManager (org.apache.hive.storage.jdbc.conf)
convertPropertiesToConfiguration:129, JdbcStorageConfigManager (org.apache.hive.storage.jdbc.conf)
initialize:78, JdbcSerDe (org.apache.hive.storage.jdbc)
getDeserializer:97, HiveMetaStoreUtils (org.apache.hadoop.hive.metastore)
getDeserializer:80, HiveMetaStoreUtils (org.apache.hadoop.hive.metastore)
getDeserializerFromMetaStore:357, Table (org.apache.hadoop.hive.ql.metadata)
getDeserializer:343, Table (org.apache.hadoop.hive.ql.metadata)
genTableLogicalPlan:3054, CalcitePlanner$CalcitePlannerAction (org.apache.hadoop.hive.ql.parse)
genLogicalPlan:4941, CalcitePlanner$CalcitePlannerAction (org.apache.hadoop.hive.ql.parse)
apply:1636, CalcitePlanner$CalcitePlannerAction (org.apache.hadoop.hive.ql.parse)
6 hidden frames
logicalPlan:1335, CalcitePlanner (org.apache.hadoop.hive.ql.parse)
genOPTree:587, CalcitePlanner (org.apache.hadoop.hive.ql.parse)
analyzeInternal:13237, SemanticAnalyzer (org.apache.hadoop.hive.ql.parse)
analyzeInternal:480, CalcitePlanner (org.apache.hadoop.hive.ql.parse)
analyze:336, BaseSemanticAnalyzer (org.apache.hadoop.hive.ql.parse)
analyze:224, Compiler (org.apache.hadoop.hive.ql)
compile:109, Compiler (org.apache.hadoop.hive.ql)
compile:498, Driver (org.apache.hadoop.hive.ql)
compileInternal:450, Driver (org.apache.hadoop.hive.ql)
compileAndRespond:414, Driver (org.apache.hadoop.hive.ql)
compileAndRespond:408, Driver (org.apache.hadoop.hive.ql)
compileAndRespond:126, ReExecDriver (org.apache.hadoop.hive.ql.reexec)
prepare:205, SQLOperation (org.apache.hive.service.cli.operation)
runInternal:268, SQLOperation (org.apache.hive.service.cli.operation)
run:286, Operation (org.apache.hive.service.cli.operation)
executeStatementInternal:558, HiveSessionImpl (org.apache.hive.service.cli.session)
executeStatementAsync:543, HiveSessionImpl (org.apache.hive.service.cli.session)
4 hidden frames
invoke:78, HiveSessionProxy (org.apache.hive.service.cli.session)
run:63, HiveSessionProxy$1 (org.apache.hive.service.cli.session)
4 hidden frames
invoke:59, HiveSessionProxy (org.apache.hive.service.cli.session)
1 hidden frame
executeStatementAsync:315, CLIService (org.apache.hive.service.cli)
ExecuteStatement:652, ThriftCLIService (org.apache.hive.service.cli.thrift)
getResult:1670, TCLIService$Processor$ExecuteStatement (org.apache.hive.service.rpc.thrift)
3 hidden frames
process:56, TSetIpAddressProcessor (org.apache.hive.service.auth)
4 hidden frames
```

Comment (deniskuzZ) [reply jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java]: Fine by me if that was well tested

File: jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java [hunk 9]
```diff
@@ -45,6 +45,8 @@ public class JdbcStorageConfigManager {
   public static final String CONFIG_PWD_KEYSTORE = Constants.JDBC_KEYSTORE;
   public static final String CONFIG_PWD_KEY = Constants.JDBC_KEY;
   public static final String CONFIG_PWD_URI = Constants.JDBC_PASSWORD_URI;
+  public static final String CONFIG_JDBC_USERNAME = Constants.JDBC_USERNAME;
```
deniskuzZ: minor: @InvisibleProgrammer, do we need to duplicate the constants?

deniskuzZ (approved):
+1

jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java
InvisibleProgrammer: - Two qtests added
- Tested with postgres metastore and postgres jdbc tables
- Schematool generation double checked with postgres

jdbc-handler/src/main/java/org/apache/hive/storage/jdbc/conf/JdbcStorageConfigManager.java
InvisibleProgrammer: No, we don't need to duplicate. I just now see that it already exists with another name, CONFIG_USERNAME. Let me remove the duplication.

PR #5711, opened 2025-03-23
HIVE-28841: Rename hive.metastore.catalog.* to hive.metastore.iceberg.catalog.*

okumin:

Relocate the prefix of Iceberg REST Catalog properties.
https://issues.apache.org/jira/browse/HIVE-28841

Two reasons.
- The current one is not inconsistent with the constant names, e.g., ICEBERG_CATALOG_SERVLET_FACTORY
- The word "catalog" is too common and might be confusing

No. HMS Iceberg catalog has not been released

No

We have integration tests

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -1829,29 +1829,29 @@ public class MetastoreConf {
             new StringSetValidator("simple", "jwt"),
         "Property-maps servlet authentication method (simple or jwt)."
     ),
-    ICEBERG_CATALOG_SERVLET_FACTORY("hive.metastore.catalog.servlet.factory",
-            "hive.metastore.catalog.servlet.factory",
+    ICEBERG_CATALOG_SERVLET_FACTORY("metastore.iceberg.catalog.servlet.factory",
+            "hive.metastore.iceberg.catalog.servlet.factory",
             "org.apache.iceberg.rest.HMSCatalogFactory",
             "HMS Iceberg Catalog servlet factory class name."
             + "The factory needs to expose a method: "
             + "public static HttpServlet createServlet(Configuration configuration);"
     ),
-    ICEBERG_CATALOG_SERVLET_PATH("hive.metastore.catalog.servlet.path",
-        "hive.metastore.catalog.servlet.path", "iceberg",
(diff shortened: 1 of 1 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1826,29 +1826,29 @@ public enum ConfVars {
             new StringSetValidator("simple", "jwt"),
         "Property-maps servlet authentication method (simple or jwt)."
     ),
-    ICEBERG_CATALOG_SERVLET_FACTORY("hive.metastore.catalog.servlet.factory",
-            "hive.metastore.catalog.servlet.factory",
+    ICEBERG_CATALOG_SERVLET_FACTORY("metastore.iceberg.catalog.servlet.factory",
```
okumin, line 1829: I removed the "hive." prefix, assuming the metastore related parameter should start with "metastore.". I will revert it if I am wrong

SourabhBadhya: I am not sure whether this change is backward compatible.

okumin: This feature is quite new; We've not cut any release tag since we merged the original change

SourabhBadhya: Ok, but won't there be downstream users who have more frequent releases than Apache Hive and which have these configs already in their production environment. Anyways, I am not sure as well whether we should block this since Apache Hive is independent of such releases.
cc @deniskuzZ Do you think this is a good idea?

okumin: I personally think they use a snapshot version of Hive on their responsibility. Additionally, this feature doesn't work for some reason yet. In my env, NoSuchMethodError happens because of the dependency hell, and some clients can't access tables unless we support Iceberg views(our HiveCatalog doesn't support it).

SourabhBadhya:
@okumin Please rebase the changes for a green build.

okumin:
@SourabhBadhya Thanks. I rebased this PR

SourabhBadhya (approved):
LGTM +1

okumin:
@SourabhBadhya Thanks for your review!

PR #5712, opened 2025-03-24
Iceberg: WriterRegistry is not thread-safe after HIVE-26319

deniskuzZ:


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/WriterRegistry.java
```diff
@@ -36,13 +36,10 @@ public static Map<String, List<HiveIcebergWriter>> removeWriters(TaskAttemptID t
...
-    writers.putIfAbsent(taskAttemptID, Maps.newConcurrentMap());
+    Map<String, List<HiveIcebergWriter>> outputWriters =
+        writers.computeIfAbsent(taskAttemptID, k -> Maps.newConcurrentMap());
 
-    Map<String, List<HiveIcebergWriter>> writersOfTableMap = writers.get(taskAttemptID);
-    writersOfTableMap.putIfAbsent(tableName, Lists.newArrayList());
-
-    List<HiveIcebergWriter> writerList = writersOfTableMap.get(tableName);
-    writerList.add(writer);
+    outputWriters.computeIfAbsent(tableName, k -> new CopyOnWriteArrayList<>()).add(writer);
```
difin, line 42: Looks good for making the list of writers thread safe. 

Was the race condition occurring in this `registerWriter` method itself when adding a writer to the list of outputWriters or was it occurring later when some concurrent threads were working with the list of outputWriters after getting them via `removeWriters` ?

deniskuzZ: Writers registration happens during task execution and removal only after the task commit. During the task commit, we create the commit manifest file by iterating over all registered in the task writers. The issue was that some of the data files files weren't added to the task commit file.

difin: LGTM

difin (approved):

PR #5713, opened 2025-03-24
HIVE-28846: change default value of ifExist in dropCatalogRequest to …

jjiang037:
…be true

Set ifExist default value to be true in DropCatalogRequest

This is a follow-up improvement for the ticket: https://issues.apache.org/jira/browse/HIVE-28443

Set ifExist default value to be true to ensure it doesn't throw exception.

No

No

Unit test added


zhangbutao (approved):
+1 LGTM

PR #5714, opened 2025-03-24
HIVE-28847: Import of eclipse-styles.xml fails in IntelliJ IDEA 2024.3.4

zabetak:
Set the problematic alignment options to the value 16, which in Eclipse is mapped to: 
Line wrapping policy: Wrap where necessary
Indentation policy: Default indentation

To allow the code-style to work in recent versions of IntelliJ and Eclipse.

No, it only affects developpers and mostly those working in IntelliJ.

No

Manual import of the new and old style in Eclipse 2025-03 (4.35.0) and IntelliJ IDEA 2024.3.4


zhangbutao (approved):
+1 lgtm

Although I believe that very few people use this file, fixing it is well worth it. I hope we can keep all hive's code style as consistent as possible.

soumyakanti3578 (approved):

PR #5715, opened 2025-03-25
HIVE-15984: Allow ALTER TABLE DROP COLUMN

ramitg254:

The changes allow dropping column via alter table drop column 

Before, the drop column query was not there as regarding sql coverage.

Now, user can also drop column via alter table drop column and no need to use replace columns for it

No

I tested it on my local machine.


ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java
```diff
@@ -0,0 +1,62 @@
...
+public class AlterTableDropColumnAnalyser extends AbstractAlterTableAnalyzer {
+
+    public AlterTableDropColumnAnalyser(QueryState queryState) throws SemanticException {
+        super(queryState);
+    }
+
+    @Override
+    protected void analyzeCommand(TableName tableName, Map<String, String> partitionSpec, ASTNode command)
+            throws SemanticException {
+        boolean ifExists = (command.getFirstChildWithType(HiveParser.TOK_IFEXISTS)!=null);
```
soumyakanti3578, line 48: These parantheses are probably unnecessary here.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
```diff
@@ -0,0 +1,93 @@
...
+        List<FieldSchema> cols = new ArrayList<>(sd.getCols());
+        int droppingIndex = -1;
+        for (int i = 0; i < cols.size(); i++) {
+            if (cols.get(i).getName().equalsIgnoreCase(desc.getColName())) {
+                droppingIndex = i;
+                break;
+            }
+        }
+        if(droppingIndex==-1){
+            if(desc.isIfExists())return;
```
soumyakanti3578, line 87: Using braces for an `if` block even when the block consists of a single line of code is a good practice to follow :)

ql/src/test/queries/clientnegative/orc_drop_column.q
```diff
@@ -0,0 +1,4 @@
+SET hive.exec.schema.evolution=true;
+create table src_orc (key int, val string) stored as orc;
+insert into src_orc values (1, "Alice");
+alter table src_orc drop column val;
```
soumyakanti3578, line 4: Please add a newline at EOF for this and other `.q` files

soumyakanti3578 (commented):
I have reviewed this only partially. Other than specific comments, I found that the spacing is inconsistent in various places. It is a good practice to not deviate from the coding convention for better readability and maintainability. You might find this helpful: https://cwiki.apache.org/confluence/display/Hive/HowToContribute#HowToContribute-CodingConventions

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java
ramitg254: removed parenthesis

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
ramitg254: added braces

ql/src/test/queries/clientnegative/orc_drop_column.q
ramitg254: added in it and others as well

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
```diff
@@ -0,0 +1,95 @@
...
+                AlterTableUtils.isSchemaEvolutionEnabled(table, context.getConf());
+        if (isOrcSchemaEvolution) {
+            throw new HiveException(ErrorMsg.CANNOT_DROP_COLUMN, desc.getDbTableName());
+        }
+
+        if (ParquetHiveSerDe.isParquetTable(table) && AlterTableUtils.isSchemaEvolutionEnabled(table, context.getConf()) &&
+                !desc.isCascade() && table.isPartitioned()) {
+            LOG.warn("Cannot drop column from a partitioned parquet table without the CASCADE option");
+            throw new HiveException(ErrorMsg.CANNOT_DROP_COLUMN, desc.getDbTableName());
+        }
```
soumyakanti3578, line 76: nit: Do we need this `warn` if we are throwing an exception too? Leaving the `warn` is fine too, so I will leave it up to you.

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
```diff
@@ -362,6 +363,7 @@ public enum ErrorMsg {
   DBTXNMGR_REQUIRES_CONCURRENCY(10264,
       "To use DbTxnManager you must set hive.support.concurrency=true"),
   TXNMGR_NOT_ACID(10265, "This command is not allowed on an ACID table {0}.{1} with a non-ACID transaction manager", true),
+  COLUMN_NOT_FOUND(10266,"Invalid Column Name"),
```
soumyakanti3578, line 366: nit: space

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -927,6 +931,44 @@ private void handleAddColumns(org.apache.hadoop.hive.metastore.api.Table hmsTabl
...
+    if (schemaDifference.getMissingFromFirst().size() > 1) {
+      throw new MetaException("Cannot drop more than one column");
+    }
+
+    if (!schemaDifference.getMissingFromSecond().isEmpty() || !schemaDifference.getTypeChanged().isEmpty() ||
+            !schemaDifference.getCommentChanged().isEmpty() || outOfOrder != null) {
+      throw new MetaException("Unsupported operation to use DROP COLUMN for adding a column, changing a " +
+              "column type, column comment or reordering columns. Only use DROP COLUMN for dropping a column. " +
+              "For the other operations, consider using the ADD COLUMNS or CHANGE COLUMN commands.");
+    }
```
soumyakanti3578, line 956: It looks like this Exception is handling too many scenarios. It might be alright but this could be broken into more specific scenarios. What do you think?

ql/src/test/results/clientpositive/llap/orc_drop_column.q.out
```diff
@@ -0,0 +1,35 @@
...
+POSTHOOK: Output: default@src_orc
+PREHOOK: query: select * from src_orc
+PREHOOK: type: QUERY
+PREHOOK: Input: default@src_orc
+#### A masked pattern was here ####
+POSTHOOK: query: select * from src_orc
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@src_orc
+#### A masked pattern was here ####
+1
```
soumyakanti3578, line 35: In these tests it might be useful to `set hive.cli.print.header=true` to get the column names in the `q.out` files.

ql/src/java/org/apache/hadoop/hive/ql/plan/HiveOperation.java
```diff
@@ -45,6 +45,8 @@ public enum HiveOperation {
   MSCK("MSCK", HiveParser.TOK_MSCK, null, null),
   ALTERTABLE_ADDCOLS("ALTERTABLE_ADDCOLS", HiveParser.TOK_ALTERTABLE_ADDCOLS, new Privilege[]{Privilege.ALTER_METADATA},
       null),
+  ALTERTABLE_DROPCOL("ALTERTABLE_DROPCOL",HiveParser.TOK_ALTERTABLE_DROPCOL,new Privilege[]{Privilege.ALTER_METADATA},
```
soumyakanti3578, line 48: nit: space

soumyakanti3578 (commented):
Consider adding more tests as this is an important new feature:
1. Negative: Dropping a column which is not present
2. Negative: Dropping a column twice
3. Have 2-3 partitions in the partitioned tables
4. Test drop column with `show` commands like `show columns`, `show create table`
5. Test with `describe [formatted]` as they list columns too

Also any other aspect which you can think of :)

ql/src/test/results/clientpositive/llap/orc_drop_column.q.out
ramitg254: okay I'll add that

ramitg254:
added those tests

ramitg254:
> Consider adding more tests as this is an important new feature:
> 
> 1. Negative: Dropping a column which is not present
> 2. Negative: Dropping a column twice
> 3. Have 2-3 partitions in the partitioned tables
> 4. Test drop column with `show` commands like `show columns`, `show create table`
> 5. Test with `describe [formatted]` as they list columns too
> 
> Also any other aspect which you can think of :)

tests added

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
```diff
@@ -362,6 +363,8 @@ public enum ErrorMsg {
   DBTXNMGR_REQUIRES_CONCURRENCY(10264,
       "To use DbTxnManager you must set hive.support.concurrency=true"),
   TXNMGR_NOT_ACID(10265, "This command is not allowed on an ACID table {0}.{1} with a non-ACID transaction manager", true),
+  COLUMN_NOT_FOUND(10266, "Invalid Column Name"),
```
deniskuzZ, line 366: Use `INVALID_COLUMN` instead

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
```diff
@@ -362,6 +363,8 @@ public enum ErrorMsg {
...
+  DROP_COLUMN_UNCASCADED(10267, "Drop column is not supported without cascade for partitioned table. SerDe may be incompatible.", true),
```
deniskuzZ, line 367: I don't see an equivalent for `REPLACE_COLUMNS`, maybe stick with `CANNOT_DROP_COLUMN`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -927,6 +931,52 @@ private void handleAddColumns(org.apache.hadoop.hive.metastore.api.Table hmsTabl
...
   }
 
+  private void handleDropColumn(org.apache.hadoop.hive.metastore.api.Table hmsTable) throws MetaException {
+    List<FieldSchema> hmsCols = hmsTable.getSd().getCols();
+    List<FieldSchema> icebergCols = HiveSchemaUtil.convert(icebergTable.schema());
+
+    HiveSchemaUtil.SchemaDifference schemaDifference = HiveSchemaUtil.getSchemaDiff(hmsCols, icebergCols, true);
+
+    if (schemaDifference.getMissingFromFirst().size() > 1) {
+      throw new MetaException("Cannot drop more than one column");
```
deniskuzZ, line 941: Shouldn't the Parser validate this?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -927,6 +931,52 @@ private void handleAddColumns(org.apache.hadoop.hive.metastore.api.Table hmsTabl
...
+    }
+
+    if (!schemaDifference.getMissingFromFirst().isEmpty()) {
```
deniskuzZ, line 944: same

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -927,6 +931,52 @@ private void handleAddColumns(org.apache.hadoop.hive.metastore.api.Table hmsTabl
...
+
+    if (!schemaDifference.getTypeChanged().isEmpty()) {
+      throw new MetaException("Unsupported operation to use DROP COLUMN for changing a column type");
+    }
+
+    if (!schemaDifference.getCommentChanged().isEmpty()) {
+      throw new MetaException("Unsupported operation to use DROP COLUMN for changing a column comment");
+    }
+
+    if (outOfOrder != null) {
```
deniskuzZ, line 963: why is that needed?

parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
```diff
@@ -209,6 +210,13 @@ alterStatementSuffixAddCol
     ->                 ^(TOK_ALTERTABLE_REPLACECOLS columnNameTypeList restrictOrCascade?)
     ;
 
+alterStatementSuffixDropCol
+@init { gParent.pushMsg("drop column statement", state); }
+@after { gParent.popMsg(state); }
+    : KW_DROP KW_COLUMN ifExists? columnName restrictOrCascade?
+    -> ^(TOK_ALTERTABLE_DROPCOL ifExists? columnName restrictOrCascade?)
```
deniskuzZ, line 217: Do we have tests for `restrictOrCascade`?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/AlterTableType.java
```diff
@@ -30,6 +30,7 @@
 public enum AlterTableType {
   // column
   ADDCOLS("add columns"),
+  DROPCOL("drop column"),
```
deniskuzZ, line 33: Please use full name `DROP_COLUMN`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java
```diff
@@ -0,0 +1,62 @@
...
+        super(queryState);
+    }
+
+    @Override
+    protected void analyzeCommand(TableName tableName, Map<String, String> partitionSpec, ASTNode command)
+            throws SemanticException {
+        boolean ifExists = command.getFirstChildWithType(HiveParser.TOK_IFEXISTS) != null;
+        String colName = ifExists ? command.getChild(1).getText().toLowerCase() : command.getChild(0).getText().toLowerCase();
+        boolean isCascade = false;
+        if (null != command.getFirstChildWithType(HiveParser.TOK_CASCADE)) {
```
deniskuzZ, line 51: why do you need if, just assign the result from bool condition

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Operation process of dropping column.
+ */
+public class AlterTableDropColumnOperation extends AbstractAlterTableOperation<AlterTableDropColumnDesc> {
+
+    public AlterTableDropColumnOperation(DDLOperationContext context, AlterTableDropColumnDesc desc) {
+        super(context, desc);
+    }
+
+    private static final Set<String> VALID_SERIALIZATION_LIBS = ImmutableSet.of(
```
deniskuzZ, line 51: where is this used?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Operation process of dropping column.
+ */
+public class AlterTableDropColumnOperation extends AbstractAlterTableOperation<AlterTableDropColumnDesc> {
```
deniskuzZ, line 45: should it be different from `AlterTableReplaceColumnsOperation`?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnDesc.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.hadoop.hive.ql.ddl.table.AbstractAlterTableDesc;
+import org.apache.hadoop.hive.ql.ddl.table.AlterTableType;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+import org.apache.hadoop.hive.ql.plan.Explain;
+import org.apache.hadoop.hive.ql.plan.Explain.Level;
+
+/**
+ * DDL task description for ALTER TABLE ... DROP COLUMN ... command.
+ */
+@Explain(displayName = "Drop Column", explainLevels = {Level.USER, Level.DEFAULT, Level.EXTENDED})
```
deniskuzZ, line 33: i can't find any tests that validate 'EXPLAIN'

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
ramitg254: this one is added for more explainatory exception in  case of dropping columns without cascade in case of parquet

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -927,6 +931,52 @@ private void handleAddColumns(org.apache.hadoop.hive.metastore.api.Table hmsTabl
     updateSchema.commit();
   }
 
+  private void handleDropColumn(org.apache.hadoop.hive.metastore.api.Table hmsTable) throws MetaException {
```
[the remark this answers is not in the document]
ramitg254: Because handleReplaceColumns is also handling addition of columns and changing a column

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
ramitg254: this and other 3 if are handling the below exception which is present in handleReplaceColumns in an elaborative manner:
```
if (!schemaDifference.getMissingFromSecond().isEmpty() || !schemaDifference.getTypeChanged().isEmpty() ||
        !schemaDifference.getCommentChanged().isEmpty() || outOfOrder != null) {
      throw new MetaException("Unsupported operation to use REPLACE COLUMNS for adding a column, changing a " +
          "column type, column comment or reordering columns. Only use REPLACE COLUMNS for dropping columns. " +
          "For the other operations, consider using the ADD COLUMNS or CHANGE COLUMN commands.");
    }
```

parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
ramitg254: there is test for cascade added when using `desc formatted table` and `desc formatted table partition (something)` yield different results for columns present without cascade as in case of specific it also contains dropped column in case normal hive tables;

ramitg254: there is no test for restrict , i added it because it was also present in case of parsing grammer  of replace columns although for no use

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public AlterTableDropColumnAnalyser(QueryState queryState) throws SemanticException {
+        super(queryState);
+    }
+
+    @Override
+    protected void analyzeCommand(TableName tableName, Map<String, String> partitionSpec, ASTNode command)
+            throws SemanticException {
+        boolean ifExists = command.getFirstChildWithType(HiveParser.TOK_IFEXISTS) != null;
+        String colName = ifExists ? command.getChild(1).getText().toLowerCase() : command.getChild(0).getText().toLowerCase();
```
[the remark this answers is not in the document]
ramitg254: I'll change that

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java
ramitg254: I'll change that

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
ramitg254: AlterTableReplaceColumnsOperation is replacing the current columnNameTypeList with provided columnNameTypeList in the query whereas AlterTableDropColumnOperation is dropping that particular column from list of columns whose name is specified in the query.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
ramitg254: It can be removed it came by mistake from `AlterTableReplaceColumnsOperation`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnDesc.java
ramitg254: I'll modify the tests for explain

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/AlterTableType.java
ramitg254: I'll change that

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
ramitg254: I'll change that

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
ramitg254: Yeah parser will validate this I added it to be on safer side, I'll change taht

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
ramitg254: done, used already present INVALID_COLUMN_NAME in place of it

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
ramitg254: Done, removed

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/AlterTableType.java
ramitg254: Done, changed

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java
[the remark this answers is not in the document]
ramitg254: done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java
ramitg254: done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
ramitg254: done, removed

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnDesc.java
ramitg254: done , added explain statement in tests

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
deniskuzZ: Would we run into the same issue with `REPLACE COLUMNS`? now behavior won't be consistent

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
deniskuzZ: Do you know how we could then drop columns with the replace command?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
deniskuzZ: why do you care about reorder when you drop 1 column?

deniskuzZ: i would expect a much simple logic, similar to handleAddColumns
````
    Collection<FieldSchema> removedCols =
        HiveSchemaUtil.getSchemaDiff(hmsTable.getSd().getCols(), HiveSchemaUtil.convert(icebergTable.schema()), false)
            .getMissingFromSecond();
    if (!removedCols.isEmpty()) {
      transaction = icebergTable.newTransaction();
      updateSchema = transaction.updateSchema();
    }
      updateSchema.deleteColumn(removedCol.get(0).getName());
    }
    updateSchema.commit();
````

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java [hunk 22]
```diff
@@ -0,0 +1,59 @@
...
+
+    public AlterTableDropColumnAnalyser(QueryState queryState) throws SemanticException {
+        super(queryState);
+    }
+
+    @Override
+    protected void analyzeCommand(TableName tableName, Map<String, String> partitionSpec, ASTNode command)
+            throws SemanticException {
+        boolean ifExists = command.getFirstChildWithType(HiveParser.TOK_IFEXISTS) != null;
+        String colName = command.getChild((ifExists)?1:0).getText().toLowerCase();
```
deniskuzZ, line 49: Please format the code (spaes are missing)

ql/src/test/queries/clientpositive/orc_drop_column.q
```diff
@@ -0,0 +1,54 @@
+-- SORT_QUERY_RESULTS
+set hive.cli.print.header=true;
+set hive.exec.schema.evolution=false;
+
+drop table if exists tbl_orc;
+
+create external table tbl_orc(a int, b string, c int) partitioned by (d string) stored as orc;
```
deniskuzZ, line 7: do you have tests for ACID (transactional=true) tables?

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
ramitg254: Yeah it is present in AlterTableRepalceColumnsOperation as:
if (ParquetHiveSerDe.isParquetTable(table) && AlterTableUtils.isSchemaEvolutionEnabled(table, context.getConf()) &&
        !desc.isCascade() && droppingColumns && table.isPartitioned()) {
      LOG.warn("Cannot drop columns from a partitioned parquet table without the CASCADE option");
      throw new HiveException(ErrorMsg.REPLACE_CANNOT_DROP_COLUMNS, desc.getDbTableName());
    }

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
ramitg254: Yeah other 3 ifs also won't be necessary i just added them because they were present in handleRepalceColumns, I'll remove them

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
ramitg254: in case of `AlterTableReplaceColumnsOperation` it is just setting the storage descriptor columns to field schema list provided in query in place existing field schema list and in case of `AlterTableDropColumnOperation` i am making a copy of existing filed schema list and removing the column mentioned in the query from that copy list and setting the storage descriptor columns to the that copy field schema list.

ramitg254: Yeah, I am aware of how replace columns can help in dropping column so basically the main idea is the same just the difference in the query syntax leads to variation in approach

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
ramitg254: just created DROP_COLUMN_UNCASCADED for explainatory exception and nothing else

ql/src/test/queries/clientpositive/orc_drop_column.q
ramitg254: I'll add one

ramitg254: done, added orc_drop_column_acid.q under clientnegative

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnAnalyser.java
ramitg254: done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
ramitg254: done, modified to a simpler one

deniskuzZ: @ramitg254, why couldn't you reuse the code snippet shared above?

ql/src/test/queries/clientpositive/orc_drop_column.q
deniskuzZ: So drop column is not supported for ACID tables?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
```diff
@@ -0,0 +1,93 @@
...
+
+  public AlterTableDropColumnOperation(DDLOperationContext context, AlterTableDropColumnDesc desc) {
+    super(context, desc);
+  }
+
+  @Override
+  protected void doAlteration(Table table, Partition partition) throws HiveException {
+    StorageDescriptor sd = getStorageDescriptor(table, partition);
+    String serializationLib = sd.getSerdeInfo().getSerializationLib();
+    AvroSerdeUtils.handleAlterTableForAvro(context.getConf(), serializationLib, table.getTTable().getParameters());
```
deniskuzZ, line 55: what is that supposed to do? i don't see same check in replace

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
```diff
@@ -0,0 +1,93 @@
...
+
+    if ("org.apache.hadoop.hive.serde.thrift.columnsetSerDe".equals(serializationLib)) {
```
deniskuzZ, line 57: where do you check for the supported serdes?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
```diff
@@ -0,0 +1,93 @@
...
+        serializationLib.equals(OrcSerde.class.getName()) && AlterTableUtils.isSchemaEvolutionEnabled(table,
+            context.getConf());
+    if (isOrcSchemaEvolution) {
+      throw new HiveException(ErrorMsg.CANNOT_DROP_COLUMN, desc.getDbTableName());
+    }
+
+    if (ParquetHiveSerDe.isParquetTable(table) && AlterTableUtils.isSchemaEvolutionEnabled(table,
+        context.getConf()) && !desc.isCascade() && table.isPartitioned()) {
+      LOG.warn("Cannot drop column from a partitioned parquet table without the CASCADE option");
+      throw new HiveException(ErrorMsg.DROP_COLUMN_UNCASCADED, desc.getDbTableName());
```
deniskuzZ, line 72: throw the same exception in replace columns

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    List<FieldSchema> cols = new ArrayList<>(sd.getCols());
+    int droppingIndex = -1;
+    for (int i = 0; i < cols.size(); i++) {
```
deniskuzZ, line 77: could be simplified with 
````
sd.getCols().removeIf(col -> col.getName().equalsIgnoreCase(desc.getColName()));
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java [hunk 28]
```diff
@@ -947,6 +951,26 @@ private void handleAddColumns(org.apache.hadoop.hive.metastore.api.Table hmsTabl
...
+    List<FieldSchema> hmsCols = hmsTable.getSd().getCols();
+    List<FieldSchema> icebergCols = HiveSchemaUtil.convert(icebergTable.schema());
+
+    HiveSchemaUtil.SchemaDifference schemaDifference = HiveSchemaUtil.getSchemaDiff(hmsCols, icebergCols, true);
```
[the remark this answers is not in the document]
ramitg254: similar kind of lookup was done in `handleReplaceColumns` so the similar way i followed the same approach

ql/src/test/results/clientnegative/orc_drop_column_acid.q.out
```diff
@@ -0,0 +1,13 @@
...
+PREHOOK: Output: default@src_orc
+POSTHOOK: query: create table src_orc (key tinyint, val string) clustered by (val) into 2 buckets stored as orc TBLPROPERTIES ('transactional'='true')
+POSTHOOK: type: CREATETABLE
+POSTHOOK: Output: database:default
+POSTHOOK: Output: default@src_orc
+PREHOOK: query: alter table src_orc drop column val
+PREHOOK: type: ALTERTABLE_DROPCOL
+PREHOOK: Input: default@src_orc
+PREHOOK: Output: default@src_orc
+FAILED: Execution Error, return code 10259 from org.apache.hadoop.hive.ql.ddl.DDLTask. Drop column is not supported for table default.src_orc. SerDe may be incompatible.
```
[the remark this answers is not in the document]
ramitg254: `orc_replace_columns1_acid.q` displays the similar behaviour as schema evolution is automatcally set to true in case of transactional ones even if set to false explicitly leading to this behaviour

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
ramitg254: `AvroSerdeUtils.handleAlterTableForAvro` was present in `AlterTableAddColumnsOperation.java` and since it was only concerned with alter table for avro so i added it in this also

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
ramitg254: since `VALID_SERIALIZATION_LIBS` in `AlterTableReplaceColumnsOperation.java` is used to check but since most of the serde mentioned in that set require some changes which i already added from replace columns to drop column so i thought there is not a specific need to check as if a particular serde is present in `VALID_SERIALIZATION_LIBS` because it will automatically throw an error if it could not be able to do that operation as there is also not that kind of specific set like `VALID_SERIALIZATION_LIBS` is used in `AlterTableAddColumnsOperation.java`

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
ramitg254: I think it shouldn't be right thing to directly alter `sd.getCols()` as it has also `sd.setCols` and  `sd.getCols()` returns internal cols directly instead of a copy list, so changed to
```
List<FieldSchema> cols = new ArrayList<>(sd.getCols());
    boolean removed = cols.removeIf(col -> col.getName().equalsIgnoreCase(desc.getColName()));
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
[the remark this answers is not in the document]
deniskuzZ: because replace can add columns, not the drop

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
deniskuzZ: i don't follow, I was asking if we should add the same check to `replace` operation?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/column/drop/AlterTableDropColumnOperation.java
deniskuzZ: ok, in that case, i think it should be dropped from `AlterTableReplaceColumnsOperation` as well

ql/src/test/results/clientnegative/orc_drop_column_acid.q.out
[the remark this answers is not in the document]
deniskuzZ: ok, but since ACID does not support it, why are we adding ACID specific checks?
````
    if (AcidUtils.isTransactionalTable(table)) {
      setAcidDdlDesc(desc);
    }
````

**Review by deniskuzZ — COMMENTED** [review]
In general LGTM, 2 comments remained

File: ql/src/test/results/clientnegative/orc_drop_column_acid.q.out (hunk 29 above)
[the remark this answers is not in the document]
Comment (ramitg254) [reply ql/src/test/results/clientnegative/orc_drop_column_acid.q.out:13]: similar check is present in `AlterTableReplaceColumnsAnalyzer.java` so i added the same check in `AlterTableReplaceColumnsAnalyzer.java` although it is not supporting operation for acid table

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java (hunk 28 above)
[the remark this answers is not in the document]
Comment (ramitg254) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java]: done removed the bidirectional check

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java [hunk 30]
```diff
@@ -947,6 +951,23 @@ private void handleAddColumns(org.apache.hadoop.hive.metastore.api.Table hmsTabl
...
+    List<FieldSchema> hmsCols = hmsTable.getSd().getCols();
+    List<FieldSchema> icebergCols = HiveSchemaUtil.convert(icebergTable.schema());
+
+    List<FieldSchema> removedCols = HiveSchemaUtil.getSchemaDiff(icebergCols, hmsCols, false).getMissingFromSecond();
+    if (removedCols.isEmpty()) {
+      return;
+    }
+
+    transaction = icebergTable.newTransaction();
+    updateSchema = transaction.updateSchema();
```
deniskuzZ, line 964: why do you need to populate `updateSchema` var? Ideally it should be completely removed
````
transaction.updateSchema()
    .deleteColumn(removedCols.get(0).getName());
    .commit();
````

Comment (ramitg254) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java:964]: trying to keep it similar to `handleReplaceColumns`

File: ql/src/test/results/clientnegative/orc_drop_column_acid.q.out (hunk 29 above)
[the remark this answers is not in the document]
Comment (ramitg254) [reply ql/src/test/results/clientnegative/orc_drop_column_acid.q.out:13]: I'll remove it in both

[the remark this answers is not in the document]
Comment (ramitg254) [reply ql/src/test/results/clientnegative/orc_drop_column_acid.q.out:13]: done removed

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java (hunk 30 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java:964]: we should fix it as well and drop the `updateSchema` global. var

File: ql/src/test/queries/clientpositive/orc_drop_column.q (hunk 23 above)
Comment (deniskuzZ) [reply ql/src/test/queries/clientpositive/orc_drop_column.q:7]: @ramitg254, build failed: https://ci.hive.apache.org/job/hive-precommit/job/PR-5715/28/testReport/junit/org.apache.hadoop.hive.ql.lockmgr/TestDbTxnManager2/Testing___split_10___PostProcess___testReplaceColumnsBlocking/
Seems replace columns should be supported for ACID... ???

Comment (ramitg254) [reply ql/src/test/queries/clientpositive/orc_drop_column.q:7]: i think it is due to removal of
```
if (AcidUtils.isTransactionalTable(table)) {
      setAcidDdlDesc(desc);
    }
```
as ACID tables are not supported for dropping columns case of replace columns and other things like changing order and addition of columns through it are supported so need to revert the changes of  `AlterTableReplaceColumnsAnalyzer.java`

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java (hunk 30 above)
Comment (ramitg254) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java]: dropped for the `handleDropColumn` but I think its necessary for `handleReplaceColumns` due to following loop:
```
for (FieldSchema droppedCol : schemaDifference.getMissingFromFirst()) {
      updateSchema.deleteColumn(droppedCol.getName());
    }
```

ramitg254: Done

ql/src/test/queries/clientpositive/orc_drop_column.q
ramitg254: Done now it is passing

deniskuzZ:
> What about Drop Column Cascade Command?

If someone is willing to contribute would be great. 
Cascade option might be useful to update the views definitions and stored procedures referencing the column. 
However, that is out of scope of this PR

PR #5716, opened 2025-03-25
HIVE-28817: Rebuilding a materialized view stored in Iceberg fails when schema has varchar column - ADDENDUM

kasakrisz:
* Revert setting the transactionId of an Acid transaction in Iceberg job configurations
* Materialized views stored by Iceberg and the MV metadata is stored in HMS doesn't need a write lock because
   the locking is done by DbTxnManager.acquireMaterializationRebuildLock()
* Acquire materialization rebuild lock when rebuilding an MV stored by Iceberg

Avoid deadlock when rebuilding M stored by Iceberg with native base tables

No.

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestIcebergLlapLocalCliDriver -Dqfile=mv_iceberg_orc8.q -pl itests/qtest-iceberg -Pitests
```


deniskuzZ (approved):
LGTM +1, pending tests

PR #5717, opened 2025-03-28
HIVE-28853: SharedWorkOptimizer should always consider MapJoin operators' InMemoryDataSize.

ngsg:

The patch modifies `SharedWorkOptimizer.extractSharedOptimizationInfoForRoot` to consider MapJoin's inMemoryDataSize.

Current SharedWorkOptimizer may produce a vertex with a large number of MapJoins, potentially causing the MapJoin HashTables to exceed available memory. This patch prevents such misbehaviour.

No

No

I tested the patch by qfile test.

### DIFF
--- a/ql/src/test/results/clientpositive/llap/sharedwork_mapjoin_datasize_check.q.out
+++ b/ql/src/test/results/clientpositive/llap/sharedwork_mapjoin_datasize_check.q.out
@@ -502,8 +502,8 @@ POSTHOOK: Input: default@src
 POSTHOOK: Input: default@src1
 #### A masked pattern was here ####
 OPTIMIZED SQL: SELECT `t4`.`$f0`, `t11`.`$f1`, `t11`.`$f2`, `t4`.`$f0` AS `$f00`, `t4`.`$f1` AS `$f10`, `t4`.`$f2` AS `$f20`, `t17`.`$f0` AS `$f01`, `t17`.`$f1` AS `$f11`, `t17`.`$f2` AS `$f21`, `t23`.`$f0` AS `$f02`, `t23`.`$f1` AS `$f12`, `t23`.`$f2` AS `$f22`
-FROM (SELECT `t0`.`key` AS `$f0`, SUM(`t0`.`"*"`) / COUNT(`t0`.`"*"`) AS `$f1`, SUM(`t2`.`value`) AS `$f2`
-FROM (SELECT `key`, 2 * CAST(`value` AS DOUBLE) AS `*`
+FROM (SELECT `t0`.`key` AS `$f0`, SUM(`t0`.`b`) / COUNT(`t0`.`b`) AS `$f1`, SUM(`t2`.`value`) AS `$f2`
+FROM (SELECT `key`, 2 * CAST(`value` AS DOUBLE) AS `b`
 FROM `default`.`src`
 WHERE `value` > 1000001 AND `key` IS NOT NULL) AS `t0`
 INNER JOIN (SELECT `key`, `value`
@@ -518,16 +518,16 @@ INNER JOIN (SELECT `key`, `value`
 FROM `default`.`src1`
 WHERE `value` > 1000000 AND `key` IS NOT NULL) AS `t8` ON `t6`.`key` = `t8`.`key`
 GROUP BY `t6`.`key`) AS `t11` ON `t4`.`$f0` = `t11`.`$f0`
-INNER JOIN (SELECT `t13`.`key` AS `$f0`, SUM(`t13`.`"*"`) / COUNT(`t13`.`"*"`) AS `$f1`, SUM(`t15`.`value`) AS `$f2`
-FROM (SELECT `key`, 3 * CAST(`value` AS DOUBLE) AS `*`
+INNER JOIN (SELECT `t13`.`key` AS `$f0`, SUM(`t13`.`b`) / COUNT(`t13`.`b`) AS `$f1`, SUM(`t15`.`value`) AS `$f2`
+FROM (SELECT `key`, 3 * CAST(`value` AS DOUBLE) AS `b`
 FROM `default`.`src`
 WHERE `value` > 1000002 AND `key` IS NOT NULL) AS `t13`
 INNER JOIN (SELECT `key`, `value`
 FROM `default`.`src1`
 WHERE `value` > 1000002 AND `key` IS NOT NULL) AS `t15` ON `t13`.`key` = `t15`.`key`
 GROUP BY `t13`.`key`) AS `t17` ON `t4`.`$f0` = `t17`.`$f0`
-INNER JOIN (SELECT `t19`.`key` AS `$f0`, SUM(`t19`.`"*"`) / COUNT(`t19`.`"*"`) AS `$f1`, SUM(`t21`.`value`) AS `$f2`
-FROM (SELECT `key`, 4 * CAST(`value` AS DOUBLE) AS `*`
+INNER JOIN (SELECT `t19`.`key` AS `$f0`, SUM(`t19`.`b`) / COUNT(`t19`.`b`) AS `$f1`, SUM(`t21`.`value`) AS `$f2`
+FROM (SELECT `key`, 4 * CAST(`value` AS DOUBLE) AS `b`
 FROM `default`.`src`
 WHERE `value` > 1000003 AND `key` IS NOT NULL) AS `t19`
 INNER JOIN (SELECT `key`, `value`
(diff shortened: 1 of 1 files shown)


ql/src/test/queries/clientpositive/sharedwork_mapjoin_datasize_check.q
```diff
@@ -0,0 +1,66 @@
+--! qt:dataset:src
+--! qt:dataset:src1
+
+set hive.auto.convert.join=true;
+set hive.llap.mapjoin.memory.oversubscribe.factor=0;
+set hive.auto.convert.join.noconditionaltask.size=500;
+
+-- The InMemoryDataSize of MapJoin is 280. Therefore, SWO should not merge 2 TSs reading src
+-- as the sum of InMemoryDataSize of 2 unmerged MapJoin exceeds 500.
```
SourabhBadhya, line 9: What is the expected change in the Explain plan which reduces the number of map joins or use a different join for performing this query? What was happening before?

SourabhBadhya (commented):
Minor comment.

ql/src/test/queries/clientpositive/sharedwork_mapjoin_datasize_check.q
ngsg: Before the patch, `SharedWorkOptimizer` merges two TableScan operators, which results in fewer Map vertices in the explained plan.

I also attached the Tez vertex dependency from the original qfile output for your understanding:
```
      Edges:
        Map 1 <- Map 4 (BROADCAST_EDGE), Reducer 5 (BROADCAST_EDGE)
        Reducer 2 <- Map 1 (SIMPLE_EDGE), Reducer 3 (BROADCAST_EDGE)
        Reducer 3 <- Map 1 (SIMPLE_EDGE)
        Reducer 5 <- Map 4 (SIMPLE_EDGE)
```

SourabhBadhya: @ngsg So now it doesn't merge these map join operators due to increase in join's memory size?

ngsg: Yes, that's correct.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1427,12 +1426,29 @@ private static SharedResult extractSharedOptimizationInfoForRoot(ParseContext pc
...
+      // Accumulate InMemoryDataSize of unmerged MapJoin operators.
+      Set<Operator<?>> opsWork1 = findWorkOperators(optimizerCache, retainableTsOp);
+      for (Operator<?> op : opsWork1) {
+        if (op instanceof MapJoinOperator) {
+          MapJoinOperator mop = (MapJoinOperator) op;
+          dataSize = StatsUtils.safeAdd(dataSize, mop.getConf().getInMemoryDataSize());
+          maxDataSize = mop.getConf().getMemoryMonitorInfo().getAdjustedNoConditionalTaskSize();
+        }
+      }
+      Set<Operator<?>> opsWork2 = findWorkOperators(optimizerCache, discardableTsOp);
```
SourabhBadhya, line 1439: Why are we considering discardable table scan operator for datasize here?

ngsg: At this point, we detect that the two TableScan operators are identical (i.e., they can be merged). However, at least one of them has multiple child operators, preventing us from merging their descendant operators. As a result, the total memory consumed by the MapJoin operator in the merged vertex should account for both descendant MapJoin operators of the retainable and discardable TableScan operators.

For example, if we merge TS1-{Join1, Join2} and TS2-{Join3, Join4}, the merged operator graph would be TS1-{Join1, Join2, Join3, Join4}, and we need to consider Join3 and Join4, which come from the discarded TableScan operator TS2.

SourabhBadhya (approved):
LGTM +1
@kasakrisz Can you please take a second look.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1474,8 +1489,29 @@ private static SharedResult extractSharedOptimizationInfoForRoot(ParseContext pc
...
+            maxDataSize = mop.getConf().getMemoryMonitorInfo().getAdjustedNoConditionalTaskSize();
+          }
+        }
+        Set<Operator<?>> opsWork2 = findWorkOperators(optimizerCache, currentOp2);
+        for (Operator<?> op : opsWork2) {
+          if (op instanceof MapJoinOperator) {
+            MapJoinOperator mop = (MapJoinOperator) op;
+            dataSize = StatsUtils.safeAdd(dataSize, mop.getConf().getInMemoryDataSize());
+            maxDataSize = mop.getConf().getMemoryMonitorInfo().getAdjustedNoConditionalTaskSize();
+          }
```
kasakrisz, line 1511: This logic already exists in lines 1430-1446. Could you please extract it to a method and remove the variable `bailOut` and call that method.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1427,12 +1426,29 @@ private static SharedResult extractSharedOptimizationInfoForRoot(ParseContext pc
...
-      return new SharedResult(retainableOps, discardableOps, discardableInputOps,
-          dataSize, maxDataSize);
+
+      // Accumulate InMemoryDataSize of unmerged MapJoin operators.
+      Set<Operator<?>> opsWork1 = findWorkOperators(optimizerCache, retainableTsOp);
+      for (Operator<?> op : opsWork1) {
+        if (op instanceof MapJoinOperator) {
+          MapJoinOperator mop = (MapJoinOperator) op;
+          dataSize = StatsUtils.safeAdd(dataSize, mop.getConf().getInMemoryDataSize());
+          maxDataSize = mop.getConf().getMemoryMonitorInfo().getAdjustedNoConditionalTaskSize();
```
kasakrisz, line 1436: I debugged the test 
```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=sharedwork_mapjoin_datasize_check.q -pl itests/qtest -Dmaven.surefire.debug -Pitests
```
but this code was never hit.

Maybe I'm missing something but could you please share how the new test `sharedwork_mapjoin_datasize_check.q` covers this part?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
```diff
@@ -1427,12 +1426,29 @@ private static SharedResult extractSharedOptimizationInfoForRoot(ParseContext pc
...
+        }
+      }
+      Set<Operator<?>> opsWork2 = findWorkOperators(optimizerCache, discardableTsOp);
+      for (Operator<?> op : opsWork2) {
+        if (op instanceof MapJoinOperator) {
+          MapJoinOperator mop = (MapJoinOperator) op;
+          dataSize = StatsUtils.safeAdd(dataSize, mop.getConf().getInMemoryDataSize());
+          maxDataSize = mop.getConf().getMemoryMonitorInfo().getAdjustedNoConditionalTaskSize();
```
kasakrisz, line 1444: `maxDataSize` is overwriten every time a `MapJoinOperator` is found. Also in line 1436. What if the first `MapJoinOperator` conf has a bigger number and the 2nd one has a smaller? In that case the smaller number is kept. Is this expected?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
ngsg: I checked the test and found that I made a mistake in the qfile. The second select query was intended to cover this part, but the current version does not cover it properly. I'll update the qfile in the next commit to ensure that the intended scenario is properly covered. Thank you for pointint it out.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/SharedWorkOptimizer.java
ngsg: The original code in `extractSharedOptimizationInfo`, which I used as a reference, overwrites it for every MapJoinOperator. I followed the same approach in `extractSharedOptimizationInfoForRoot` for consistency, but I'm not sure the current version is expected, and I think that it would be better to replace this overwriting logic with max() (StatsUtils version if available). I'll update the code in both `extractSharedOptimizationInfoForRoot` and `extractSharedOptimizationInfo` in the next commit.

ngsg:
@kasakrisz , I have updated the patch according to your review. Could you review the changes again?

kasakrisz (approved):

kasakrisz:
@ngsg
The test sharedwork_mapjoin_datasize_check failed in the last PTest run.
Could you please check it?

ngsg:
@kasakrisz , I have updated the qfile output. It looks like I forgot to compile the project after rebase. Thanks for pointing that out.

kasakrisz:
@ngsg 
Could you please rebase this patch to master and check the test failures.

PR #5719, opened 2025-03-28
HIVE-28855: VectorGroupByOperator should use getGroupByMemoryUsage() instead of getMemoryThreshold().

ngsg:

Modify VectorGroupByOperator to use getGroupByMemoryUsage() instead of getMemoryThreshold().

Both `GroupByOperator` and `VectorGroupByOperator` flush the HashTable when:
(1) The process’s memory usage becomes too high.
(2) The estimated HashTable size exceeds its computed limit (`maxHashTblMemory` in both classes).

For (1), `GroupByOperator` compares memory usage against `hive.map.aggr.hash.force.flush.memory.threshold``, while `VectorGroupByOperator` monitors GC canary.
For (2), both operators compute `maxHashTblMemory` and compare it with the estimated HashTable size. However, they use different configurations: `GroupByOperator` relies on `hive.map.aggr.hash.percentmemory`, while `VectorGroupByOperator` incorrectly uses `hive.map.aggr.hash.force.flush.memory.threshold`.

Since `GroupByOperator` already uses `THRESHOLD` for condition (1), `VectorGroupByOperator` should use `PERCENT` for condition (2) to maintain consistency.
 
If a user manages GroupBy's HashTable size using `hive.map.aggr.hash.force.flush.memory.threshold`, one should adjust the configuration file accordingly.

No

This patch does not include a test since it only corrects a configuration misinterpretation.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorGroupByOperator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorGroupByOperator.java
@@ -155,7 +155,7 @@ public class VectorGroupByOperator extends Operator<GroupByDesc>
 
   private transient long maxMemory;
 
-  private float memoryThreshold;
+  private float hashTableMemoryPercentage;
 
   private boolean isLlap = false;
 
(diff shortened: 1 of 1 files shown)


okumin:
I also remember I saw this odd inconsistency between GroupByOperator and VectorGroupByOperator. I'm recalling it and double-checking this PR.
By the way, I can see TestVectorGroupByOperator failing.
https://ci.hive.apache.org/blue/rest/organizations/jenkins/pipelines/hive-precommit/branches/PR-5719/runs/1/nodes/433/log/?start=0

ngsg:
@okumin, thanks for looking into this issue.
I checked `TestVectorGroupByOperator` and replaced `setMemoryThreshold` with `setGroupByMemoryUsage` as the test limits the size of HashTables, not the size of JVM process.

okumin:
I also double-checked the content in the PR description. As you say, GroupByOperator and VectorGroupByOperator have some inconsistencies.
- The source of the maximum memory for the operators; the non-vectorized version uses getMaxMemoryAvailable both for LLAP and Tez, and the vectorized version uses getMaxMemoryAvailable for LLAP and max JVM heap for Tez
- `hive.map.aggr.hash.force.flush.memory.threshold` vs `hive.map.aggr.hash.percentmemory` for hash table memory(numEntries * width)
- The metrics of total memory pressure; the non-vectorized version uses used JVM heap and the vectorized version uses a SoftReference
- Only the vectorized version has a threshold for simple count

This PR addresses the second problem. I think we can also standardize the first one. I have no idea about the third and fourth ones, as they seem to be intentional.

okumin (approved):
+1. I am rerunning Jenkins

ngsg:
> This PR addresses the second problem. I think we can also standardize the first one. I have no idea about the third and fourth ones, as they seem to be intentional.

I agree that the third and fourth issues seem to be intended behaviour. Regarding the first one, I believe the non-vectorized version might be more appropriate; however, I’m not entirely certain since I’m not fully familiar with the memory management policy of Tez. It might be helpful to open a dedicated JIRA to discuss and address the first problem.

PR #5721, opened 2025-03-28
HIVE-28170: Implement drop stats

soumyakanti3578:

Implements drop stats for columns in partitioned and unpartitioned tables

https://issues.apache.org/jira/browse/HIVE-28170

Yes, users can now drop stats for columns

No

```
mvn test -pl itests/qtest -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile="drop_stats_for_columns.q,drop_stats_for_columns_partitioned.q,drop_histogram_stats_for_columns.q"
mvn test -pl itests/qtest -Pitests -Dtest=TestNegativeLlapCliDriver -Dtest.output.overwrite=true -Dqfile="drop_stats_for_columns_iceberg.q"
```


ql/src/java/org/apache/hadoop/hive/ql/ddl/table/misc/columnstats/AlterTableDropColumnStatisticsAnalyzer.java
```diff
@@ -0,0 +1,71 @@
...
+  public AlterTableDropColumnStatisticsAnalyzer(QueryState queryState) throws SemanticException {
+    super(queryState);
+  }
+
+  @Override
+  protected void analyzeCommand(TableName tableName, Map<String, String> partitionSpec, ASTNode command)
+      throws SemanticException {
+    Table table = getTable(tableName);
+    
+    if (DDLUtils.isIcebergTable(table)) {
```
kasakrisz, line 54: Please don't use `isIcebergTable` because in the compiler it is too specific. I don't understand why is this method exists in the first place. It is an architectural mistake.

IIUC we provide the drop stats function for storage types that store the column stats in HMS. So how about excluding the ones which can store them by themselves?
```
if (table.isNonNative() && table.getStorageHandler().canSetColStatistics(table))
```

There is also a configuration which controls where stats are stored in case of Iceberg:
https://github.com/apache/hive/blob/0e10ff30292265f00a7e7fb9c93f310f4349ac56/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java#L2238-L2239

What should we do when it is set to `metasore`? Maybe we should allow deleting? WYT?

ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnStatsDropTask.java
```diff
@@ -0,0 +1,59 @@
...
+  @Override
+  public int execute() {
+    try {
+      getHive()
+          .deleteColumnStatistics(work.getDbName(), work.getTableName(), work.getPartName(), work.getColNames());
+      return 0;
+    } catch (Exception e) {
+      setException(e);
+      LOG.info("Failed to drop column stats in metastore", e);
+      return ErrorMsg.getErrorMsg(e.getMessage()).getErrorCode();
```
kasakrisz, line 46: What is the error code in this case?
What type of exceptions do we expect?
Should we create a unique error?

ql/src/test/queries/clientpositive/drop_stats_for_columns.q
```diff
@@ -0,0 +1,36 @@
+create table t1(i int, j int, k int);
+
+alter table t1 update statistics set('numRows'='10000', 'rawDataSize'='18000');
+alter table t1 update statistics for column i set('numDVs'='2500','numNulls'='50','highValue'='1000','lowValue'='0');
```
kasakrisz, line 4: How does this update works?
Especially if some properties are not specified? Do they removed?
```
alter table t1 update statistics for column i set();
```

ql/src/test/results/clientpositive/llap/drop_histogram_stats_for_columns.q.out
```diff
@@ -0,0 +1,307 @@
...
+c                   	double              	                    
+	 	 
+# Detailed Table Information	 	 
+Database:           	default             	 
+#### A masked pattern was here ####
+Retention:          	0                   	 
+#### A masked pattern was here ####
+Table Type:         	MANAGED_TABLE       	 
+Table Parameters:	 	 
+	COLUMN_STATS_ACCURATE	{\"BASIC_STATS\":\"true\",\"COLUMN_STATS\":{\"a\":\"true\",\"b\":\"true\",\"c\":\"true\"}}
```
kasakrisz, line 230: Why are column stats accurate after drop?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/misc/columnstats/AlterTableDropColumnStatisticsAnalyzer.java
soumyakanti3578: For now I think I will replace `isIcebergTable` with the condition that you suggested, since it is too specific in the compiler. Along the same lines, I think using the conf `HIVE_ICEBERG_STATS_SOURCE` might be too specific too.

Maybe we should rename this property to remove `ICEBERG`, and update the values  to `metastore` and `other` too? But this is beyond the scope of this PR.

ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnStatsDropTask.java
soumyakanti3578: I think this is catching any `Exception` that may come from HMS.
I can create a new error code for this and add it to the `HiveException` in `Hive#deleteColumnStatistics`

ql/src/test/queries/clientpositive/drop_stats_for_columns.q
soumyakanti3578: > How does this update works?

This is handled by [AlterTableUpdateColumnStatistictAnalyzer](https://github.com/apache/hive/blob/b9ada6d8b30b325c91e6ab88a9deb7bb98b8a738/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/misc/columnstats/AlterTableUpdateColumnStatistictAnalyzer.java). Also checkout [ColumnStatsUpdateTask](https://github.com/apache/hive/blob/b9ada6d8b30b325c91e6ab88a9deb7bb98b8a738/ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnStatsUpdateTask.java), where we construct column stats in `constructColumnStatsFromInput()` and then build a `SetPartitionsStatsRequest` to persist the stats in HMS.

> alter table t1 update statistics for column i set();

This throws a `ParseException`:
```
org.apache.hadoop.hive.ql.parse.ParseException: line 2:50 cannot recognize input near ')' '<EOF>' '<EOF>' in table properties list
```
`set()` cannot be empty as it expects `tableProperties` ([AlterClauseParser.g](https://github.com/apache/hive/blob/b9ada6d8b30b325c91e6ab88a9deb7bb98b8a738/parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g#L258))
```
alterStatementSuffixUpdateStatsCol[boolean partition]
@init { gParent.pushMsg("update column statistics", state); }
@after { gParent.popMsg(state); }
    : KW_UPDATE KW_STATISTICS KW_FOR KW_COLUMN? colName=identifier KW_SET tableProperties (KW_COMMENT comment=StringLiteral)?
    -> {partition}? ^(TOK_ALTERPARTITION_UPDATECOLSTATS $colName tableProperties $comment?)
    ->              ^(TOK_ALTERTABLE_UPDATECOLSTATS $colName tableProperties $comment?)
    ;
```

ql/src/test/results/clientpositive/llap/drop_histogram_stats_for_columns.q.out
soumyakanti3578: If I understand correctly, we need to do this in HS2 - after `getMSC().deleteColumnStatistics(request);` returns successfully in `Hive#deleteColumnStatistics()`.

I think we can probably use `StatsSetupConst#removeColumnStatsState(Map<String, String> params, List<String> colNames)` to update the column stats for the list of columns in the ALTER statement. However, for partitioned tables, when partition info is not provided in the ALTER statement, we need to fetch ALL partitions (using `List<Partition> getPartitions(Table tbl)` in Hive.java), and then for each Partition, we need to call `StatsSetupConst#removeColumnStatsState`.

Do you think there is a better and more efficient way to do this?

soumyakanti3578: I think my analysis above is not correct. Columns stats are accurate even after dropping stats because the `Table`'s parameters still has the stats marked as accurate. I am trying to figure out how to update the parameters in HMS so that `getTable()` API in `Hive.java` can return `Table` with updated parameters.

ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnStatsDropTask.java
kasakrisz: Does the exceptions coming from HMS are general errors (like i/o error) or there are exceptions related to the drop stats functionality. If the latter is true I think we should introduce an error code for those ones and `GENERIC_ERROR` is fine otherwise .

ql/src/test/queries/clientpositive/drop_stats_for_columns.q
kasakrisz: Thanks for checking this. My idea was to use an empty set to remove stats but let's go with the drop statistics syntax.

ql/src/test/results/clientpositive/llap/drop_histogram_stats_for_columns.q.out
kasakrisz: Yes, maybe this should be handled by HMS. AFAIK stats autogather also update them using alter table HMS api call. 
https://github.com/apache/hive/blob/db041fd7b71a15a8bd1a997b0908a02e8124d7ac/ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java#L306

ql/src/java/org/apache/hadoop/hive/ql/exec/ColumnStatsDropTask.java
soumyakanti3578: I think we can just use `GENERIC_ERROR` then. I will make the changes.

PR #5722, opened 2025-03-30
HIVE-28851: HiveIcebergMetaHook acquires an HMS lock, regardless of the config and operations

owenmonn:

HiveIcebergMetaHook acquires the HMS lock only when necessary, depending on the operationType and the config(`engine.hive.lock-enabled`).

Stat tasks fail during insert queries on Iceberg tables.
How to reproduce:
```
set hive.txn.ext.locking.enabled=true;
set hive.stats.autogather=true;

create table ice_t (i int) stored by iceberg tblproperties ('engine.hive.lock-enabled'='false');

insert into ice_t values (1);
```
StatsTask Fail:
```
2025-03-27T12:54:00,304 INFO [Thread-108] stats.BasicStatsTask: [Warning] could not update stats.Failed with exception Unable to alter table. org.apache.iceberg.hive.LockException: Timed out after 182005 ms waiting for lock on default.ice_t
 ```
HIVE_LOCKS Table:
```
+----------------+---------+----------+---------------+--------------+-----+---------------------+
| HL_LOCK_EXT_ID | HL_DB   | HL_TABLE | HL_LOCK_STATE | HL_LOCK_TYPE | ... | HL_BLOCKEDBY_EXT_ID |
+----------------+---------+----------+---------------+--------------+-----+---------------------+
|            110 | default | ice_t    | a             | w            | ... |                NULL |
|            111 | default | ice_t    | w             | x            | ... |                 110 |
+----------------+---------+----------+---------------+--------------+-----+---------------------+
```

The Stat tasks of the Iceberg table will succeed

No

Unit test added

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -3121,7 +3121,7 @@ public class HiveConf extends Configuration {
     TXN_MERGE_INSERT_X_LOCK("hive.txn.xlock.mergeinsert", false,
         "Ensures MERGE INSERT operations acquire EXCLUSIVE / EXCL_WRITE lock for transactional tables.\n" +
         "If enabled, prevents duplicates when MERGE statements are executed in parallel transactions."),
-    TXN_WRITE_X_LOCK("hive.txn.xlock.write", true,
+    TXN_WRITE_X_LOCK("hive.txn.xlock.write", false,
         "Manages concurrency levels for ACID resources. Provides better level of query parallelism by enabling " +
         "shared writes and write-write conflict resolution at the commit step." +
         "- If true - exclusive writes are used:\n" +
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -3130,7 +3130,7 @@ public class HiveConf extends Configuration {
         "  - INSERT acquires SHARED_READ locks\n" +
         "- If false - shared writes, transaction is aborted in case of conflicting changes:\n" +
         "  - INSERT OVERWRITE acquires EXCL_WRITE locks\n" +
-        "  - INSERT/UPDATE/DELETE acquire SHARED_READ locks"),
+        "  - INSERT/UPDATE/DELETE acquire SHARED_WRITE locks"),
     HIVE_TXN_STATS_ENABLED("hive.txn.stats.enabled", true,
         "Whether Hive supports transactional stats (accurate stats for transactional tables)"),
     HIVE_TXN_ACID_DIR_CACHE_DURATION("hive.txn.acid.dir.cache.duration",
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -55,6 +55,7 @@ import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
 import org.apache.hadoop.hive.metastore.partition.spec.PartitionSpecProxy;
 import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
+import org.apache.hadoop.hive.ql.QueryState;
 import org.apache.hadoop.hive.ql.ddl.misc.sortoder.SortFieldDesc;
 import org.apache.hadoop.hive.ql.ddl.misc.sortoder.SortFields;
 import org.apache.hadoop.hive.ql.ddl.table.AlterTableType;
@@ -70,6 +71,7 @@ import org.apache.hadoop.hive.ql.parse.PartitionTransform;
 import org.apache.hadoop.hive.ql.parse.TransformSpec;
 import org.apache.hadoop.hive.ql.plan.ExprNodeDesc;
 import org.apache.hadoop.hive.ql.plan.ExprNodeGenericFuncDesc;
+import org.apache.hadoop.hive.ql.plan.HiveOperation;
 import org.apache.hadoop.hive.ql.session.SessionState;
 import org.apache.hadoop.hive.ql.session.SessionStateUtil;
 import org.apache.hadoop.hive.ql.util.NullOrdering;
@@ -115,10 +117,13 @@ import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.expressions.UnboundPredicate;
 import org.apache.iceberg.expressions.UnboundTerm;
+import org.apache.iceberg.hadoop.ConfigProperties;
 import org.apache.iceberg.hive.CachedClientPool;
+import org.apache.iceberg.hive.HiveLock;
 import org.apache.iceberg.hive.HiveSchemaUtil;
 import org.apache.iceberg.hive.HiveTableOperations;
 import org.apache.iceberg.hive.MetastoreLock;
+import org.apache.iceberg.hive.NoLock;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.mapping.MappingUtil;
@@ -193,7 +198,7 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
   private Transaction transaction;
   private AlterTableType currentAlterTableOp;
   private boolean createHMSTableInHook = false;
-  private MetastoreLock commitLock;
+  private HiveLock commitLock;
 
   private enum FileFormat {
     ORC("orc"), PARQUET("parquet"), AVRO("avro");
@@ -409,8 +414,7 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
           context.getProperties().get(OLD_TABLE_NAME)).toString());
     }
     if (commitLock == null) {
-      commitLock = new MetastoreLock(conf, new CachedClientPool(conf, Maps.fromProperties(catalogProperties)),
-          catalogProperties.getProperty(Catalogs.NAME), hmsTable.getDbName(), hmsTable.getTableName());
+      commitLock = lockObject(hmsTable);
     }
 
     try {
@@ -420,6 +424,32 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
       commitLock.unlock();
       throw new MetaException(StringUtils.stringifyException(e));
     }
+  }
+
+  private HiveLock lockObject(org.apache.hadoop.hive.metastore.api.Table hmsTable) {
+    if (hiveLockEnabled(hmsTable, conf)) {
+      return new MetastoreLock(conf, new CachedClientPool(conf, Maps.fromProperties(catalogProperties)),
+          catalogProperties.getProperty(Catalogs.NAME), hmsTable.getDbName(), hmsTable.getTableName());
+    } else {
+      return new NoLock();
+    }
+  }
+
+  private static boolean hiveLockEnabled(org.apache.hadoop.hive.metastore.api.Table hmsTable, Configuration conf) {
+    if (SessionStateUtil.getQueryState(conf).map(QueryState::getHiveOperation)
+        .filter(opType -> HiveOperation.ANALYZE_TABLE == opType)
+        .isPresent()) {
+      return false;
+    }
+
+    if (hmsTable.getParameters().containsKey(TableProperties.HIVE_LOCK_ENABLED)) {
+      // We know that the property is set, so default value will not be used,
+      return Boolean.parseBoolean(
+          hmsTable.getParameters().getOrDefault(TableProperties.HIVE_LOCK_ENABLED, "false"));
+    }
+
+    return conf.getBoolean(
+        ConfigProperties.LOCK_HIVE_ENABLED, TableProperties.HIVE_LOCK_ENABLED_DEFAULT);
   }
 
   private void doPreAlterTable(org.apache.hadoop.hive.metastore.api.Table hmsTable, EnvironmentContext context)
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -117,7 +117,6 @@ import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.expressions.ResidualEvaluator;
 import org.apache.iceberg.expressions.UnboundPredicate;
 import org.apache.iceberg.expressions.UnboundTerm;
-import org.apache.iceberg.hadoop.ConfigProperties;
 import org.apache.iceberg.hive.CachedClientPool;
 import org.apache.iceberg.hive.HiveLock;
 import org.apache.iceberg.hive.HiveSchemaUtil;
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java [hunk 1]
```diff
@@ -130,6 +130,28 @@ public void testStatsWithInsert() {
...
+
+    shell.setHiveSessionValue(HiveConf.ConfVars.HIVE_STATS_AUTOGATHER.varname, true);
+    shell.setHiveSessionValue(HiveConf.ConfVars.HIVE_TXN_EXT_LOCKING_ENABLED.varname, true);
+    testTables.createTable(shell, identifier.name(), HiveIcebergStorageHandlerTestUtils.CUSTOMER_SCHEMA,
+        PartitionSpec.unpartitioned(), fileFormat, ImmutableList.of(), 1,
+        ImmutableMap.of(TableProperties.HIVE_LOCK_ENABLED, "false"));
+
+    if (testTableType != TestTables.TestTableType.HIVE_CATALOG) {
+      // If the location is set and we have to gather stats, then we have to update the table stats now
+      shell.executeStatement("ANALYZE TABLE " + identifier + " COMPUTE STATISTICS FOR COLUMNS");
```
deniskuzZ, line 145: why do you need this?

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java
```diff
@@ -130,6 +130,28 @@ public void testStatsWithInsert() {
...
 
+  @Test
+  public void testStatsWithPessimisticLockInsert() {
+    TableIdentifier identifier = TableIdentifier.of("default", "customers");
+
+    shell.setHiveSessionValue(HiveConf.ConfVars.HIVE_STATS_AUTOGATHER.varname, true);
+    shell.setHiveSessionValue(HiveConf.ConfVars.HIVE_TXN_EXT_LOCKING_ENABLED.varname, true);
+    testTables.createTable(shell, identifier.name(), HiveIcebergStorageHandlerTestUtils.CUSTOMER_SCHEMA,
+        PartitionSpec.unpartitioned(), fileFormat, ImmutableList.of(), 1,
+        ImmutableMap.of(TableProperties.HIVE_LOCK_ENABLED, "false"));
```
deniskuzZ, line 141: Could you create a parametrized test and check `HIVE_LOCK_ENABLED`=`true` as well?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -422,6 +426,32 @@ public void preAlterTable(org.apache.hadoop.hive.metastore.api.Table hmsTable, E
...
+      return new MetastoreLock(conf, new CachedClientPool(conf, Maps.fromProperties(catalogProperties)),
+          catalogProperties.getProperty(Catalogs.NAME), hmsTable.getDbName(), hmsTable.getTableName());
+    } else {
+      return new NoLock();
+    }
+  }
+
+  private static boolean hiveLockEnabled(org.apache.hadoop.hive.metastore.api.Table hmsTable, Configuration conf) {
+    if (SessionStateUtil.getQueryState(conf).map(QueryState::getHiveOperation)
+        .filter(opType -> HiveOperation.ANALYZE_TABLE == opType)
```
deniskuzZ, line 440: i think it should be changed to `HiveOperation.QUERY == opType`

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java
owenmonn: It's not necessary. I've fixed it to only test HiveCatalog. [8dbe2f0](https://github.com/apache/hive/pull/5722/commits/8dbe2f08dda9e734774413225fb64f8c8220cf7e)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
owenmonn: I agree. I've fixed it. [a0d4c38](https://github.com/apache/hive/pull/5722/commits/a0d4c3830fe5c4c3676af0df054e0f5e6b5ce2db)

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java
owenmonn: Are you referring to a test in which `HIVE_LOCK_ENABLED`=`true` cause a query to fail due to a conflict between the pessimistic lock and the commit lock at the move task stage?

owenmonn: I would like to clarify something.
Even after this fix, when HIVE_LOCK_ENABLED is set to true, the STATS TASK (HiveIcebergMetaHook) does not acquire the HMS lock during the INSERT query, but the MOVE TASK (HiveTableOperations) still acquires the HMS lock. This leads to a deadlock with pessimistic lock.

Given this, I understand that if you want to use pessimistic locking with Iceberg tables, you should set HIVE_LOCK_ENABLED to false. If not, it seems that modifications might also be necessary in HiveTableOperations.

Could you please confirm if my understanding is correct?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -422,6 +426,32 @@ public void preAlterTable(org.apache.hadoop.hive.metastore.api.Table hmsTable, E
...
+    if (hiveLockEnabled(hmsTable, conf)) {
+      return new MetastoreLock(conf, new CachedClientPool(conf, Maps.fromProperties(catalogProperties)),
+          catalogProperties.getProperty(Catalogs.NAME), hmsTable.getDbName(), hmsTable.getTableName());
+    } else {
+      return new NoLock();
+    }
+  }
+
+  private static boolean hiveLockEnabled(org.apache.hadoop.hive.metastore.api.Table hmsTable, Configuration conf) {
+    if (SessionStateUtil.getQueryState(conf).map(QueryState::getHiveOperation)
```
InvisibleProgrammer, line 439: If I understood it correctly, the evaluation order of deciding if locking is enabled is: session level -> table properties -> config file. 

Am I correct? 
The other question is: What is the user experience if the enable locking, disable locking on a session level, forget about it and use a table when locking is explicitly enabled on table level? 
Should we give at least a warning in that case just to be able to troubleshoot it?

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java
deniskuzZ: you are right. 
In that case I think we should add a fallback to `HiveIcebergStorageHandler#getLockType` and when `HIVE_LOCK_ENABLED=true` log WARN "Pessimistic locking can't be enabled when HIVE_LOCK_ENABLED=true" and return SHARED_READ lock

deniskuzZ:
````
  public LockType getLockType(WriteEntity writeEntity) {
    // Materialized views stored by Iceberg and the MV metadata is stored in HMS doesn't need write locking because
    // the locking is done by DbTxnManager.acquireMaterializationRebuildLock()
    if (TableType.MATERIALIZED_VIEW == writeEntity.getTable().getTableType()) {
      return LockType.SHARED_READ;
    }
    if (IcebergTableUtil.hiveLockEnabled(writeEntity.getTable().getTTable())) {
      LOG.warn("Pessimistic locking cannot be enabled when HIVE_LOCK_ENABLED is set to true");
      return LockType.SHARED_READ;
    }
    if (WriteEntity.WriteType.INSERT_OVERWRITE == writeEntity.getWriteType()) {
      return LockType.EXCL_WRITE;
    }
    return LockType.SHARED_WRITE;
  }
````

Comment (owenmonn) [reply iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java:142]: b663804
It seems like everything is working well now. I will add some test code.

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java (hunk 4 above)
Comment (owenmonn) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java]: After the additional [commit](https://github.com/apache/hive/pull/5722/commits/b663804786cb616ace3aafbc0955c7e702109f8d),
If the situation you're concerned about still exists, could you please inform me about it via a script?

File: iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java (hunk 2 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java]: awesome! thank you, @owenmonn!

File: iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java [hunk 5]
```diff
@@ -52,14 +53,20 @@ public class TestHiveIcebergStatistics extends HiveIcebergStorageHandlerWithEngi
...
+  @Parameterized.Parameters(name = "fileFormat={0}, catalog={1}, isVectorized={2}, " +
+      "formatVersion={3}, statsSource={4}, isLockEnabled={5}")
   public static Collection<Object[]> parameters() {
     Collection<Object[]> baseParams = HiveIcebergStorageHandlerWithEngineBase.parameters();
 
     Collection<Object[]> testParams = Lists.newArrayList();
     for (String statsSource : new String[]{"iceberg", "metastore"}) {
-      for (Object[] params : baseParams) {
-        testParams.add(ArrayUtils.add(params, statsSource));
+      for (boolean isLockEnabled : new boolean[]{true, false}) {
```
deniskuzZ, line 66: That would cause all other tests to run twice. Maybe better to create 2 tests?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -3121,7 +3121,7 @@ public static enum ConfVars {
     TXN_MERGE_INSERT_X_LOCK("hive.txn.xlock.mergeinsert", false,
         "Ensures MERGE INSERT operations acquire EXCLUSIVE / EXCL_WRITE lock for transactional tables.\n" +
         "If enabled, prevents duplicates when MERGE statements are executed in parallel transactions."),
-    TXN_WRITE_X_LOCK("hive.txn.xlock.write", true,
+    TXN_WRITE_X_LOCK("hive.txn.xlock.write", false,
```
deniskuzZ, line 3124: @owenmonn, could you please update https://github.com/apache/hive/pull/5722/files#diff-b7bbe8545a21ec7d7e9cfe40ef66444789e332996aaa9e7f1430dbe4822a2c9cR3133
There is a typo:  INSERT/UPDATE/DELETE acquire `SHARED_WRITE` locks

owenmonn: [6b2c912](https://github.com/apache/hive/pull/5722/commits/6b2c91258d1961ca6c163b6e6a49cfd8a7b526fa)
I've fixed it.

iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergStatistics.java
```diff
@@ -130,6 +131,43 @@ public void testStatsWithInsert() {
     checkColStatMinMaxValue(identifier.name(), "customer_id", 0, 5);
   }
 
+  @Test
+  public void testStatsWithPessimisticLockInsert() {
```
deniskuzZ, line 135: The test code is exactly the same. Please extract it to a private method with `enableStatsAutogather` param

owenmonn: [503172f](https://github.com/apache/hive/pull/5722/commits/503172fde0dc90fe6ca83f084323cf2c2c4e1aa5)
i extract private method(getTableIdentifierWithPessimisticLock)!

deniskuzZ (approved):
LGTM, pending tests

owenmonn:
@deniskuzZ 

Thanks for the review! The tests seem to be pending due to the "This workflow requires approval from a maintainer" requirement.

Could you please approve the workflow?

owenmonn:
[3567afc](https://github.com/apache/hive/pull/5722/commits/3567afc905a2f5f14ee47e651b99ded1b6a15ddc)

The tests `TestTxnCommands#testParallelTruncateAnalyzeStats` and `TestTxnCommandsWithSplitUpdateAndVectorization#testParallelTruncateAnalyzeStats` were failing due to the default value change of `hive.txn.xlock.write` to `false`.

When `hive.txn.xlock.write=false`, the `ANALYZE` query throws a `RuntimeException` if it attempts to acquire a `SHARED_READ` lock after the `TRUNCATE` query has already acquired `EXCL_WRITE` lock, contrary to the expected test results.

This commit adjusts `hive.txn.xlock.write` to `true` in test to address this issue.

deniskuzZ:
> [3567afc](https://github.com/apache/hive/pull/5722/commits/3567afc905a2f5f14ee47e651b99ded1b6a15ddc)
> 
> The tests `TestTxnCommands#testParallelTruncateAnalyzeStats` and `TestTxnCommandsWithSplitUpdateAndVectorization#testParallelTruncateAnalyzeStats` were failing due to the default value change of `hive.txn.xlock.write` to `false`.
> 
> When `hive.txn.xlock.write=false`, the `ANALYZE` query throws a `RuntimeException` if it attempts to acquire a `SHARED_READ` lock after the `TRUNCATE` query has already acquired `EXCL_WRITE` lock, contrary to the expected test results.
> 
> This commit adjusts `hive.txn.xlock.write` to `true` in test to address this issue.

`hive.txn.xlock.write` doesn't change the locking behavior used for `TRUNCATE` or `ANALYZE`. 
`TRUNCATE` takes an `EXCLUSIVE` lock and `ANALYZE` - `SHARED_READ`

owenmonn:
>hive.txn.xlock.write doesn't change the locking type needed for TRUNCATE or ANALYZE.
TRUNCATE takes an EXCLUSIVE lock and ANALYZE - SHARED_READ
However, it activates the ZeroWaitRead optimization that allows SHARED_READ to fail fast and not wait for EXCLUSIVE lock to be released.

You're right. Enabling `ZeroWaitRead` causes existing tests to fail. To address this, I've disabled ZeroWaitRead in the existing tests by setting `hive.txn.xlock.write=true`.

test failure log:
```
java.lang.RuntimeException: java.lang.RuntimeException: analyze table mm_table compute statistics for columns failed: 
(responseCode = 10, 
 errorMessage = FAILED: Error in acquiring locks: Locks on the underlying objects cannot be acquired, retry after some time. 
 LockResponse(lockid:10, state:NOT_ACQUIRED, errorMessage:Unable to acquire read lock due to an existing exclusive lock 
{lockid:9,  intLockId:1, txnid:9, db:default, table:mm_table, partition:null, state:ACQUIRED, type:EXCLUSIVE}), 
 hiveErrorCode = 40000,  SQLState = 42000,  exception = Locks on the underlying objects cannot be acquired, retry after some time. 
```

PR #5728, opened 2025-04-01
HIVE-28859: Upgrade commons-logging to 1.3.5 to fix CVEs

tanishq-chugh:
Upgrading commons-logging to 1.3.5

To fix multiple CVEs

No

Yes
[dpn_2.txt](https://github.com/user-attachments/files/19552812/dpn_2.txt)

Built hive locally with this patch and ran some manual queries

### DIFF
--- a/storage-api/pom.xml
+++ b/storage-api/pom.xml
@@ -28,7 +28,6 @@
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <commons-logging.version>1.3.5</commons-logging.version>
     <guava.version>22.0</guava.version>
     <hadoop.version>3.4.1</hadoop.version>
     <junit.version>4.13.2</junit.version>
--- a/storage-api/pom.xml
+++ b/storage-api/pom.xml
@@ -28,7 +28,7 @@
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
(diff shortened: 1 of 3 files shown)


storage-api/pom.xml
```diff
@@ -28,7 +28,7 @@
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <commons-logging.version>1.1.3</commons-logging.version>
+    <commons-logging.version>1.3.5</commons-logging.version>
```
Aggarwal-Raghav, line 31: We can remove this commons-logging property from storage-api.

tanishq-chugh: This has been addressed in [35e558c](https://github.com/apache/hive/pull/5728/commits/35e558ca7b8c076a59cde13c6a5d7315ee20a6fa)

ayushtkn (commented):
Where is commons-logging being used in Hive? I see [HIVE-24691](https://issues.apache.org/jira/browse/HIVE-24691)/[HIVE-20019](https://issues.apache.org/jira/browse/HIVE-20019) banned it

Aggarwal-Raghav:
Yes, commons-logging can't used as a direct dependency https://github.com/apache/hive/blob/95db41cf6dff065796d47c6910bcb7c2e85856d0/pom.xml#L1695-L1700

But commons-logging-1.1.3 is getting packaged as commons-logging direct dependency exists in standalone-metastore and https://github.com/apache/hive/blob/master/standalone-metastore/pom.xml#L341-L345
and in its submodules.

PR #5731, opened 2025-04-01
HIVE-28861 fixed issue: collection added to itself

dk2k:


dk2k:
@SourabhBadhya can you please take a look? I have no idea why checks failed

PR #5732, opened 2025-04-01
HIVE-28862 Throwable not thrown in TransactionalValidationListener

dk2k:


dk2k:
What can I do to get it fixed and merged?

PR #5733, opened 2025-04-01
HIVE-28863 Removed unused StringBuilder in ColumnPrunerProcFactory

dk2k:


dk2k:
Why do the check failed?

PR #5736, opened 2025-04-02
HIVE-28866: Upgrade netty-codec-http to fix CVE-2024-29025

tanishq-chugh:
Upgrade a single instance of compile time netty-codec-http 

To fix CVE-2024-29025

No

Yes
[dpn_netty_119.txt](https://github.com/user-attachments/files/19574998/dpn_netty_119.txt)

Ran manual queries after local hive build with patch

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -187,7 +187,7 @@
     <mockito-core.version>3.4.4</mockito-core.version>
     <mockito-inline.version>4.11.0</mockito-inline.version>
     <mina.version>2.0.0-M5</mina.version>
-    <netty.version>4.1.116.Final</netty.version>
+    <netty.version>4.1.119.Final</netty.version>
     <netty3.version>3.10.5.Final</netty3.version>
     <!-- used by druid storage handler -->
     <pac4j-saml.version>4.5.8</pac4j-saml.version>
--- a/pom.xml
+++ b/pom.xml
@@ -431,6 +431,16 @@
         <version>${netty.version}</version>
       </dependency>
       <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-handler</artifactId>
+        <version>${netty.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-transport-native-epoll</artifactId>
+        <version>${netty.version}</version>
+      </dependency>
+      <dependency>
         <groupId>jakarta.jms</groupId>
         <artifactId>jakarta.jms-api</artifactId>
         <version>${jms.version}</version>
--- a/standalone-metastore/pom.xml
+++ b/standalone-metastore/pom.xml
@@ -92,6 +92,7 @@
     <libthrift.version>0.16.0</libthrift.version>
     <log4j2.version>2.24.3</log4j2.version>
     <mockito-core.version>3.4.4</mockito-core.version>
+    <netty.version>4.1.116.Final</netty.version>
     <orc.version>1.9.4</orc.version>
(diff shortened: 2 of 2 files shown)


standalone-metastore/pom.xml
```diff
@@ -173,6 +174,11 @@
         <artifactId>metrics-json</artifactId>
         <version>${dropwizard.version}</version>
       </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-all</artifactId>
```
Aggarwal-Raghav, line 179: I think the goal is for the transitive dependencies to have higher version of netty? 

Also, I see older version of netty _(can be seen in your attached dependency tree)_ in `packaging/target/apache-hive-4.1.0-SNAPSHOT-bin/apache-hive-4.1.0-SNAPSHOT-bin/lib` coming from zookeeper 3.8.4 and has CVE's as well https://mvnrepository.com/artifact/io.netty/netty-handler/4.1.105.Final
```
netty-handler-4.1.105.Final.jar
netty-transport-native-epoll-4.1.105.Final.jar
```

standalone-metastore/pom.xml
```diff
@@ -92,6 +92,7 @@
     <libthrift.version>0.16.0</libthrift.version>
     <log4j2.version>2.24.3</log4j2.version>
     <mockito-core.version>3.4.4</mockito-core.version>
+    <netty.version>4.1.116.Final</netty.version>
```
ayushtkn, line 95: can we move to `4.1.119.Final`
https://mvnrepository.com/artifact/io.netty/netty-all/4.1.119.Final

standalone-metastore/pom.xml
tanishq-chugh: Have added these deps as well in dependencyManagement in [bd96998](https://github.com/apache/hive/pull/5736/commits/bd96998e7b2385e05610bb8b81957ee14bcd0560) , so these transitive dependencies are of the same versions.

standalone-metastore/pom.xml
tanishq-chugh: Checking it out.

pom.xml
```diff
@@ -430,6 +430,16 @@
         <artifactId>netty-all</artifactId>
         <version>${netty.version}</version>
       </dependency>
+      <dependency>
```
ayushtkn, line 433: Why are we adding dependency is scope of upgrade? I don't see any exclusions either which could lead it this?

tanishq-chugh: We are adding these specifications in dependencyManagement, coz other deps were bringing in older versions of these and with this addition we'll get only 4.1.119.Final. 
We are addressing @Aggarwal-Raghav comment with this [https://github.com/apache/hive/pull/5736#discussion_r2025071034](https://github.com/apache/hive/pull/5736#discussion_r2025071034)

Aggarwal-Raghav: > We are adding these specifications in dependencyManagement, coz other deps were bringing in older versions of these transitively and with this addition we'll get only 4.1.119.Final. We are addressing @Aggarwal-Raghav comment with this [#5736 (comment)](https://github.com/apache/hive/pull/5736#discussion_r2025071034)

@tanishq-chugh , can you please provide the sonatype report URL for this? I don't see netty-codec-http coming up in my nexus scan in my org.
Regarding this approach to add dependency of higher version is not a good idea, IMO, because if higher version of netty is not backward compatible with older version (required by transitive dependency) then we are prone to NoSuchMethod or NoClassFound errors at runtime. This is similar to the avro issue https://github.com/apache/hive/pull/5593 that we last discussed.

PR #5739, opened 2025-04-02
HIVE-28868: Print AM hostname on console

abstractdog:
Print AM hostname on the console.

Because on DAG submission, the user ends up seeing an application id, which is hard to track back to a Tez AM / coordinator, making the investigation hard. Imagine a long-running query on a cluster that utilizes 30+ Tez AMs, and in case of e.g. a kubernetes environment, logical hostnames are used, like TezAM1, TezAM2...by knowing only the DAG id and application id, there will always be an extra step to locate the AM, which could be painful if the app id <-> hostname pair is only printed once a few hours (days?) ago, if it logged at all (need to be aware that the application can be alive since "forever").

Yes, on DAG submission, not only the AM app id, but the app master's hostname will be printed.

No.

Tested on cluster, check screenshot:
https://issues.apache.org/jira/secure/attachment/13075767/Screenshot%202025-04-02%20at%2014.53.49.png

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezTask.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezTask.java
@@ -268,8 +268,8 @@ public class TezTask extends Task<TezWork> {
         // Log all the info required to find the various logs for this query
         String dagId = this.dagClient.getDagIdentifierString();
         String appId = this.dagClient.getSessionIdentifierString();
-        LOG.info("HS2 Host: [{}], Query ID: [{}], Dag ID: [{}], DAG App ID: [{}]", ServerUtils.hostname(), queryId,
-            dagId, appId);
+        LOG.info("HS2 Host: [{}], Query ID: [{}], Dag ID: [{}], DAG App ID: [{}], DAG App address: [{}]",
+            ServerUtils.hostname(), queryId, dagId, appId, runtimeContext.getAmHostName());
         LogUtils.putToMDC(LogUtils.DAGID_KEY, dagId);
         this.jobID = dagId;
         runtimeContext.setDagId(dagId);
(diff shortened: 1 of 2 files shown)


ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezTask.java
```diff
@@ -268,8 +268,8 @@ public int execute() {
         // Log all the info required to find the various logs for this query
         String dagId = this.dagClient.getDagIdentifierString();
         String appId = this.dagClient.getSessionIdentifierString();
-        LOG.info("HS2 Host: [{}], Query ID: [{}], Dag ID: [{}], DAG App ID: [{}]", ServerUtils.hostname(), queryId,
-            dagId, appId);
+        LOG.info("HS2 Host: [{}], Query ID: [{}], Dag ID: [{}], DAG App ID: [{}], DAG App address: [{}]",
+            ServerUtils.hostname(), queryId, dagId, appId, runtimeContext.getAmHostName());
```
deniskuzZ, line 272: can't we use `session.getSession().getAmHost()`?

deniskuzZ: other thing, `Ref<TezSessionState> sessionRef` magic is confusing and redundant

abstractdog: makes sense, getAmHost works, I changed to it
also removed an unused TezClient import 

regarding the confusing sessionRef, agree, but I'm not about to touch that magic at the moment, created [HIVE-28877](https://issues.apache.org/jira/browse/HIVE-28877)

abstractdog:
thanks for the reviews @deniskuzZ, @kokila-19 so far, please check if the latest patch can be merged

PR #5742, opened 2025-04-02
HIVE-28874 Incorrect comparison in LlapFixedRegistryImpl

dk2k:


llap-client/src/java/org/apache/hadoop/hive/llap/registry/impl/LlapFixedRegistryImpl.java
```diff
@@ -257,7 +257,7 @@ public List<LlapServiceInstance> getAllInstancesOrdered(boolean consistentIndexe
       Collections.sort(list, new Comparator<LlapServiceInstance>() {
         @Override
         public int compare(LlapServiceInstance o1, LlapServiceInstance o2) {
-          return o2.getWorkerIdentity().compareTo(o2.getWorkerIdentity());
+          return o1.getWorkerIdentity().compareTo(o2.getWorkerIdentity());
```
SourabhBadhya, line 260: Doesn't this change the sorting order? I think now you will see it in ascending order when compared to descending order.

DmKryukov: Previously there was no order (and inconsistency instead) as o2 was compared to o2 and they are always equal.
Ascending Order (o1.compareTo(o2))
Descending Order (o2.compareTo(o1))
Please let me know if descending order is expected

dk2k:
dear maintainers, please review

llap-client/src/java/org/apache/hadoop/hive/llap/registry/impl/LlapFixedRegistryImpl.java
abstractdog: Considering that the comparison was incorrect and FixedServiceInstanceSet failed to comply with any sorting requirements (unlike DynamicServiceInstanceSet, which uses a TreeMap to maintain order), it doesn't really matter whether we fix the sorting or remove it entirely.

However, since the interface implies that some ordering is expected, it's reasonable to fix it now using a simple ascending order, which aligns better with intuitive expectations and brings consistency to the implementation.

abstractdog:
> dear maintainers, please review

left a comment above, this is fine to be merged, can you please rebase your patch on master to get a green build?

dk2k:
@abstractdog  please merge

PR #5743, opened 2025-04-02
HIVE-28875 Numbers should be compared with equals(), not with ==

dk2k:
Comparison with == and != checks memory adresses
equals() check values


aturoczy (commented):
As I see, you have made several PR for equals change. Please wait for @ayushtkn and @deniskuzZ  recommendation about the style. If it meets the community standard, I think this PR's could be 1 "big" does not need to separate to 10+ small.

SourabhBadhya:
`equals()` is used to check the object content, `==` does a comparision on the object reference. Using `equals()` is the right programming paradigm.

I did a small test for confirming the same for Double datatype - There is a difference - 
```
Double do1 = 9.0;
Double do2 = 9.00;
System.out.println(do1 == do2); // Returns false
System.out.println(do1.equals(do2)); // Returns true
```

I agree with @aturoczy comment. Instead of creating small PRs for handling each of these minor details its better to handle it in a single PR. Its always better to group similar ones and post aggregated PRs on them.

dk2k:
@SourabhBadhya @aturoczy I didn't mean to overwhelm the repo with PRs. But still I prefer to have several atomic PRs because it's easier to complete each PR. As for equals(): it has some specifics for Strings and numbers. If both strings are interned (which I didn't check), it's ok to compare their values using ==

SourabhBadhya:
@dk2k Please rebase your PR for the tests to pass.

dk2k:
@SourabhBadhya I can merge master branch. Is it what you need?

SourabhBadhya:
@dk2k Yes you can do that as well.

dk2k:
@SourabhBadhya I merged

SourabhBadhya:
@dk2k There is a qtest which is failing - windowing_udaf.q due to this change.
Please evaluate whether modifying the corresponding q.out file is OK here.
https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5743/6/tests

dk2k:
@SourabhBadhya  please restart the checks

SourabhBadhya:
@dk2k Please merge master branch due to recent Calcite upgrade.

dk2k:
@SourabhBadhya please check

ql/src/test/results/clientpositive/llap/windowing_udaf.q.out
```diff
@@ -505,7 +505,7 @@ alice brown	25.258749999999996
 alice brown	25.529374999999998
 alice brown	25.63012987012987
 alice brown	26.472439024390237
-alice brown	27.27881720430106
+alice brown	27.100638297872322
```
SourabhBadhya, line 508: https://github.com/apache/hive/pull/5743/files#diff-3ed1e47d0499aed223f40aaacf7d3b5d17ec6c6e63d3f986209c0c217bcdb513R930
I am seeing references of window function using the `isEqualPrimitive()` in its function call stack. I am assuming this to be the reason why this value is changing.

dk2k:
@SourabhBadhya tests are green. Is it good to merge?

PR #5744, opened 2025-04-02
HIVE-28876 Strings should be compared with equals(), not with ==

dk2k:
Comparison with == and != checks memory adresses
equals() check values

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/repl/util/ReplUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/repl/util/ReplUtils.java
@@ -361,8 +361,8 @@ public class ReplUtils {
   }
 
   private static String getMetricStageName(String stageName, ReplicationMetricCollector metricCollector) {
-    if( stageName == "REPL_DUMP" || stageName == "REPL_LOAD" || stageName == "ATLAS_DUMP" || stageName == "ATLAS_LOAD"
-            || stageName == "RANGER_DUMP" || stageName == "RANGER_LOAD" || stageName == "RANGER_DENY"){
+    if( "REPL_DUMP".equals(stageName) || "REPL_LOAD".equals(stageName) || "ATLAS_DUMP".equals(stageName) || "ATLAS_LOAD".equals(stageName)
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/exec/repl/util/ReplUtils.java
```diff
@@ -361,8 +361,8 @@ public static boolean isCreateOperation(DDLOperation<?> ddlOperation) {
   }
 
   private static String getMetricStageName(String stageName, ReplicationMetricCollector metricCollector) {
-    if( stageName == "REPL_DUMP" || stageName == "REPL_LOAD" || stageName == "ATLAS_DUMP" || stageName == "ATLAS_LOAD"
-            || stageName == "RANGER_DUMP" || stageName == "RANGER_LOAD" || stageName == "RANGER_DENY"){
+    if( "REPL_DUMP".equals(stageName) || "REPL_LOAD".equals(stageName) || "ATLAS_DUMP".equals(stageName) || "ATLAS_LOAD".equals(stageName)
```
aturoczy, line 364: I do not have to much sentiment if the string is equals or ==. Both are acceptable.  I would chime @deniskuzZ and @ayushtkn to tell which is the most commonly used style in Hive.

aturoczy (changes requested):
Please fill out the PR description accordingly.

ql/src/java/org/apache/hadoop/hive/ql/exec/repl/util/ReplUtils.java
kasakrisz: In Java the `==` operator compares only the references not the actual content of the objects so this change is needed here.

harshal-16: Old code also works because Java uses constant folding at compile time for optimization.
and this stageName comes from private final String STAGE_NAME = "REPL_LOAD";
but to make it consistent with non-final fields, .equals() is a better option.

You can check the first answer from this.

https://stackoverflow.com/questions/8581824/does-java-compiler-include-string-constant-folding

PR #5745, opened 2025-04-03
HIVE-28852: Enhanced HMSHandler partition logs for get_partitions expression calls

shivjha30:
Additional log statements have been added to capture the details of the get_partitions_by_expr function calls, including input parameters and the results of the execution. This will help in tracking the function's behavior and identifying issues more effectively.

For get_partition_by_expr and get_partitions_spec_by_expr it is not logging the expression/filter 
Adding the expression along with other data can ease the process of reading HMS logs in future

NO

NO

Checking the outcome in Hive.logs
image

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
@@ -3340,12 +3340,8 @@ public class ObjectStore implements RawStore, Configurable {
       final String order, final int maxParts) throws MetaException, NoSuchObjectException {
     final String defaultPartitionName = getDefaultPartitionName(defaultPartName);
     final boolean isEmptyFilter = exprBytes.length == 1 && exprBytes[0] == -1;
-    ExpressionTree tmp = null;
-    if (!isEmptyFilter) {
-      tmp = PartFilterExprUtil.makeExpressionTree(expressionProxy, exprBytes,
-          getDefaultPartitionName(defaultPartName), conf);
-    }
-    final ExpressionTree exprTree = tmp;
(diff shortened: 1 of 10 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -3342,8 +3342,8 @@ public List<String> listPartitionNames(final String catName, final String dbName
     final boolean isEmptyFilter = exprBytes.length == 1 && exprBytes[0] == -1;
     ExpressionTree tmp = null;
```
kokila-19, line 3343: I think we can remove the **ExpressionTree tmp = null;**  as it is not used anymore as you have removed it in the if condition.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -3342,8 +3342,8 @@ public List<String> listPartitionNames(final String catName, final String dbName
...
     if (!isEmptyFilter) {
-      tmp = PartFilterExprUtil.makeExpressionTree(expressionProxy, exprBytes,
-          getDefaultPartitionName(defaultPartName), conf);
+      String exprString = PartFilterExprUtil.getExpressionString(expressionProxy, exprBytes, getDefaultPartitionName(defaultPartName), conf);
+      ExpressionTree exprTree = ExpressionTree.fromString(exprString);
     }
     final ExpressionTree exprTree = tmp;
```
kokila-19, line 3348: you can remove this as well as it is used further down in 
**final ExpressionTree exprTree = exprString.isEmpty() ? ExpressionTree.EMPTY_TREE : ExpressionTree.fromString(exprString);**

PR #5746, opened 2025-04-03
HIVE-28878: Query with window function and redundant GROUP BY key fails during compilation

kasakrisz:
When generating the CBO plan of a query which contains window function calls Hive first creates a `HiveProject` contains the window function calls. Alongside a `RowResolver` object is created based on the `RowResolver` of the input operator of this new `HiveProject`. The source operator might contain more complex expressions not just column references and this information is not carried to the new `RowResolver`. This patch adds the missing setting.

Although having redundant Group by keys doesn't change the semantics of the query it is not forbidden. However Hive throws `SemanticsException` with the message ambiguous column reference. This is caused by that the RowResolver created by the window function handler in the CBO doesn't expect more complex expressions than a column reference.

Such queries will compile and run. 

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestMiniLlapLocalCliDriver -Dqfile=cbo_windowing_aggregate.q -pl itests/qtest -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
@@ -4180,6 +4180,7 @@ public class CalcitePlanner extends SemanticAnalyzer {
 
       // 3. Construct new Row Resolver with everything from below.
       RowResolver out_rwsch = new RowResolver();
+      out_rwsch.setExprResolver(inputRR.getIsExprResolver());
       if (!RowResolver.add(out_rwsch, inputRR)) {
         LOG.warn(ERROR_MESSAGE_DUPLICATES_DETECTED);
       }
--- a/ql/src/test/queries/clientpositive/cbo_windowing_aggregate.q
+++ b/ql/src/test/queries/clientpositive/cbo_windowing_aggregate.q
@@ -1,0 +1,17 @@
+create table t1 (a int);
+
+explain cbo
+select
+  a * 2 as mul1,
+  a * 2 as mul2,
+  row_number() over (order by a * 2)
+from t1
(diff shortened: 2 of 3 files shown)


simhadri-g (approved):
LGTM +1

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -4180,6 +4180,7 @@ private RelNode genSelectForWindowing(QB qb, RelNode srcRel, HashSet<ColumnInfo>
 
       // 3. Construct new Row Resolver with everything from below.
       RowResolver out_rwsch = new RowResolver();
+      out_rwsch.setExprResolver(inputRR.getIsExprResolver());
```
mdayakar, line 4183: Should we set it to 'true' always instead of getting the value from inputRR...? If inputRR contains 'false' for 'isExprResolver' then we face same issue again right.

kasakrisz: IIUC there are two types of RowResolver: one can store expressions and the other one can't. This affects the ambiguity check. For the one which does not allow expression but columns only we need the ambiguity check. For the one which allows expressions like group by key duplicates are allowed even though it hasn't got an impact on the semantics.

mdayakar: ok got your point, in the current flow `inputRR` related to group by for which `isExprResolver` is always true as per code flow so I feel the code changes are fine.

mdayakar (approved):

mdayakar:
LGTM +1

PR #5748, opened 2025-04-06
HIVE-28765: Iceberg: Incorrect partition statistics on time travel + partition evolution

okumin:

https://issues.apache.org/jira/browse/HIVE-28765

We use not partition stats but table stats when we access a complicated table with tagging or branching.

We would like to improve the query plan for a time-travel read with tagging and branching.

No

No

Updated an existing query file.


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -2149,12 +2149,16 @@ public List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table hm
   }
 
   public boolean isPartitioned(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
-    if ((hmsTable.getAsOfVersion() != null || hmsTable.getAsOfTimestamp() != null) &&
-            hasUndergonePartitionEvolution(hmsTable) ||
```
okumin, line 2153: I used `IcebergTableUtil.getTableSnapshot` to avoid duplicate knowledge of the snapshot configurations. I can move it back to the original style if we prefer it.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -2149,12 +2149,16 @@ public List<Partition> getPartitions(org.apache.hadoop.hive.ql.metadata.Table hm
...
-        !hmsTable.getTTable().isSetId()) {
+    if (!hmsTable.getTTable().isSetId()) {
       return false;
     }
     Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
+    Snapshot snapshot = IcebergTableUtil.getTableSnapshot(table, hmsTable);
+    boolean isTimeTravel = snapshot != null && table.currentSnapshot() != null &&
```
deniskuzZ, line 2157: could we use `equals` here?
````
  public boolean isPartitioned(org.apache.hadoop.hive.ql.metadata.Table hmsTable) {
    if (!hmsTable.getTTable().isSetId()) {
      return false;
    }
    Table table = IcebergTableUtil.getTable(conf, hmsTable.getTTable());
    Snapshot snapshot = IcebergTableUtil.getTableSnapshot(table, hmsTable);

    boolean isTimeTravel = snapshot != null && snapshot.equals(table.currentSnapshot());
    if (isTimeTravel && hasUndergonePartitionEvolution(hmsTable)) {
      return false;
    }
    return table.spec().isPartitioned();
  }
````

okumin: I agree that it is simpler and tolerant with unexpected snapshots, such as having the same snapshot id but different timestamps.
https://github.com/apache/iceberg/blob/c338323b2e9cd8a862fdf328ceacc1b59ea6275a/core/src/main/java/org/apache/iceberg/BaseSnapshot.java#L320-L332

I updated it. I kept the variable name, isTimeTravel, assuming reading a different snapshot using tag/branch sounds like a time-travel query.

deniskuzZ: forked branch could be ahead of main, so I wouldn't call it time-travel + you can't rollback to a branch/tag

deniskuzZ (approved):
+1, pending tests

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
okumin: I grabbed your point. I applied a slight modification to the variable name.
https://github.com/apache/hive/pull/5748/commits/1ac16b2ae6880228fcb76acb520e51b1d1b74306

okumin:
@deniskuzZ Thanks for reviewing this pull request!

PR #5749, opened 2025-04-07
HIVE-28881: Ambiguous column reference is not detected in GROUP BY query.

kasakrisz:
Check for ambiguity when building expressions of group by keys.

Ambiguous column references are not detected.

Yes, exception is thrown when ambiguous column references are found in group by key expressions.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestNegativeLlapLocalCliDriver -Dqfile=cbo_ambiguous_colref_in_gby.q -pl itests/qtest -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
@@ -3800,26 +3800,21 @@ public class CalcitePlanner extends SemanticAnalyzer {
         groupByOutputRowResolver.setIsExprResolver(true);
 
         if (hasGrpByAstExprs) {
-          try {
-            groupByInputRowResolver.setCheckForAmbiguity(true);
-            // 4. Construct GB Keys (ExprNode)
-            for (int i = 0; i < groupByNodes.size(); ++i) {
-              ASTNode groupByNode = groupByNodes.get(i);
-              Map<ASTNode, RexNode> astToRexNodeMap = genAllRexNode(
-                      groupByNode, groupByInputRowResolver, cluster.getRexBuilder());
-              RexNode groupByExpression = astToRexNodeMap.get(groupByNode);
-              if (groupByExpression == null) {
-                throw new CalciteSemanticException("Invalid Column Reference: " + groupByNode.dump(),
-                        UnsupportedFeature.Invalid_column_reference);
-              }
-
-              addToGBExpr(groupByOutputRowResolver, groupByInputRowResolver, groupByNode,
-                      groupByExpression, groupByExpressions, outputColumnNames);
+          groupByInputRowResolver.setCheckForAmbiguity(true);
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -3800,19 +3800,25 @@ private RelNode genGBLogicalPlan(QB qb, RelNode srcRel) throws SemanticException
...
 
-            addToGBExpr(groupByOutputRowResolver, groupByInputRowResolver, groupByNode,
-                groupByExpression, groupByExpressions, outputColumnNames);
+              addToGBExpr(groupByOutputRowResolver, groupByInputRowResolver, groupByNode,
+                      groupByExpression, groupByExpressions, outputColumnNames);
+            }
+          }
+          finally {
+            groupByInputRowResolver.setCheckForAmbiguity(false);
           }
```
soumyakanti3578, line 3822: I was wondering if we can avoid the extra indentation caused by `try-finally` here. If we don't use `try-finally` the only concerning thing that I can think of is that `groupByInputRowResolver.setCheckForAmbiguity(false);` will not run anymore in case we throw the `CalciteSemanticException`. 

But do we need to set the boolean anyway if we are throwing the Exception? 

If not, then it might be better to just remove `try-finally`; then this will be a small two-line change.

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -3800,19 +3800,25 @@ private RelNode genGBLogicalPlan(QB qb, RelNode srcRel) throws SemanticException
...
-                groupByNode, groupByInputRowResolver, cluster.getRexBuilder());
-            RexNode groupByExpression = astToRexNodeMap.get(groupByNode);
-            if (groupByExpression == null) {
-              throw new CalciteSemanticException("Invalid Column Reference: " + groupByNode.dump(),
-                  UnsupportedFeature.Invalid_column_reference);
-            }
+          try {
+            groupByInputRowResolver.setCheckForAmbiguity(true);
+            // 4. Construct GB Keys (ExprNode)
+            for (int i = 0; i < groupByNodes.size(); ++i) {
```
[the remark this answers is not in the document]
kasakrisz: replaced to foreach

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
kasakrisz: Your are right, an exception is causing a CBO failure anyway.
I removed the try-finally

soumyakanti3578 (approved):
LGTM, approved.

SourabhBadhya (approved):
LGTM +1

PR #5750, opened 2025-04-07
HIVE-28884: Decouple source.q test from SRC dataset

zabetak:
Update the SQL script to simply run two very basic SQL queries that don't require access to any data.

The purpose of the test is to verify that `SOURCE` CLI command (introduced in HIVE-1415) launches a SQL script but the content of the actual script is not really important.

1. Avoid intermittent failures in CI and when the test runs solo.
2. Speedup test and reduce resource usage.

No

No

```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=source.q -pl itests/qtest -Pitests
```


data/files/source.txt
```diff
@@ -1,4 +1,2 @@
-EXPLAIN
-SELECT x.* FROM SRC x;
-
-SELECT x.* FROM SRC x;
+SELECT 'Executed SQL query 1 using SOURCE command';
+SELECT 'Executed SQL query 2 using SOURCE command';
```
soumyakanti3578, line 2: nit: It probably doesn't matter, but earlier we were testing EXPLAIN too. If you think it is needed, please add an EXPLAIN statement too.

soumyakanti3578 (approved):
Thanks for the patch @zabetak as this was failing intermittently in other builds. I was thinking of creating a patch with `--! qt:dataset:src` added to `source.q`, but this is clearly better :)

data/files/source.txt
InvisibleProgrammer: You are right. It doesn't matter. As @zabetak wrote: 
The goal of the test source.q is to test if the command source works at all.

InvisibleProgrammer (approved):
Thank you for fixing it. I just wasted a couple of hours trying to figure out why this test fails: 
- It failed at my PR
- And now I understand why it didn't work locally. 

The other solution could be to initialize the dataset itself. But your solution is more elegant - as it removes extra, unnecessary statements like actually querying tables. 
I just wanted to open a PR that fixes the issue. But with this, I won't do that.

okumin (approved):
+1. I'm curious why this test sometimes successfully finished

simhadri-g (approved):

zabetak:
@okumin Without the changes the test could finish successfully when it run together (and after) another test that loads the SRC dataset using `qt:dataset:src` command.

PR #5751, opened 2025-04-07
HIVE-28885: Disable unnecessary annotation processing

zabetak:
1. Disable annotation processors from all modules that do not need them.
2. Decouple immutables processor from errorprone processor and move immutables processor in iceberg-catalog module since it is only user there.

Except iceberg, all the other modules do not need annotation processors. However, the log4j-core dependency contains an annotation processor and this has impact on the build leading to maven WARNING messages and the deactivation of compile avoidance.
```
[WARNING] The following annotation processors were found on the classpath:
[org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor], provided by /home/someuser/.m2/repository/org/apache/logging/log4j/log4j-core/2.24.3/log4j-core-2.24.3.jar
Compile avoidance has been deactivated.
Please use the maven-compiler-plugin version 3.5 or above and use the <annotationProcessorPaths> configuration element to declare the processors instead.
If you did not intend to use the processors above (e.g. they were leaked by a dependency), you can use the <proc>none</proc> option to disable annotation processing.

For more information see https://gradle.com/help/maven-extension-compile-avoidance.
```

Apart from compile avoidance the use of annotation processing can have other side effects during the build and it can even lead to cryptic compilation failures.

No

No

1. Checked that build passes with itests and errorProne profile activated.

```
mvn clean install -DskipTests -Pitests
mvn clean install -DskipTests -Pitests,errorProne
```
2. Verified that annotation processor warnings are gone
3. Introduced some errorProne bugs (e.g., use toString on array) in iceberg catalog module and confirmed that build fails when errorProne profile is active.
```java
long[] someArray = new long[1];
someArray.toString();
```

### DIFF
--- a/iceberg/pom.xml
+++ b/iceberg/pom.xml
@@ -305,23 +305,19 @@
             <artifactId>maven-compiler-plugin</artifactId>
             <configuration>
               <fork>true</fork>
+              <proc combine.self="override"/>
               <showWarnings>true</showWarnings>
               <compilerArgs>
                 <arg>-XDcompilePolicy=simple</arg>
                 <arg>-Xplugin:ErrorProne -Xep:BanSerializableRead:WARN</arg>
                 <arg>-J-Xbootclasspath/p:${settings.localRepository}/com/google/errorprone/javac/${google.errorprone.javac.version}/javac-${google.errorprone.javac.version}.jar</arg>
               </compilerArgs>
-              <annotationProcessorPaths>
+              <annotationProcessorPaths combine.children="append">
                 <path>
                   <groupId>com.google.errorprone</groupId>
                   <artifactId>error_prone_core</artifactId>
                   <version>${google.errorprone.version}</version>
                 </path>
-                <dependency>
-                  <groupId>org.immutables</groupId>
-                  <artifactId>value</artifactId>
-                  <version>${immutables.value.version}</version>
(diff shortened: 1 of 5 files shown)


iceberg/pom.xml
```diff
@@ -305,23 +305,19 @@
...
                 <arg>-J-Xbootclasspath/p:${settings.localRepository}/com/google/errorprone/javac/${google.errorprone.javac.version}/javac-${google.errorprone.javac.version}.jar</arg>
               </compilerArgs>
-              <annotationProcessorPaths>
+              <annotationProcessorPaths combine.children="append">
                 <path>
                   <groupId>com.google.errorprone</groupId>
                   <artifactId>error_prone_core</artifactId>
                   <version>${google.errorprone.version}</version>
                 </path>
-                <dependency>
```
deniskuzZ, line 320: we should probably move the `error_prone_annotations` to iceberg-catalog as well

deniskuzZ:
@zabetak, could you please reopen? changes LGTM

zabetak:
@deniskuzZ I couldn't reopen this PR but rebased and created https://github.com/apache/hive/pull/6221 . I will monitor the test run to see if anything else is needed. I don't remember if there were issues with it.

PR #5753, opened 2025-04-08
HIVE-28781: Hive shouldn't disable IPv6 stack.

difin:

Hive used to disable IPv6 by setting this config in some services and in maven surefire plugin (for tests): `java.net.preferIPv4Stack=true`. When this config is `true`, IPv6 is completely disabled. This should be changed to `false` to support dual stack, IPv4 and IPv6.

To add support for IPv6 in Hive.

No

No

Hive pre-commit tests.

### DIFF
--- a/hcatalog/src/test/e2e/templeton/deployers/config/webhcat/webhcat-site.xml
+++ b/hcatalog/src/test/e2e/templeton/deployers/config/webhcat/webhcat-site.xml
@@ -95,7 +95,7 @@
 
     <property>
         <name>templeton.controller.mr.child.opts</name>
-        <value> -Xmx64m -Djava.net.preferIPv4Stack=true</value>
+        <value> -Xmx64m</value>
     </property>
 
     <property>
--- a/testutils/hadoop
+++ b/testutils/hadoop
@@ -69,6 +69,7 @@ fi
 HADOOP_OPTS="$HADOOP_OPTS -Dhadoop.policy.file=$HADOOP_POLICYFILE"
 
 # Disable ipv6 as it can cause issues
+# TODO: change to -Djava.net.preferIPv4Stack=false after Hadoop adds support for IPv6
 HADOOP_OPTS="$HADOOP_OPTS -Djava.net.preferIPv4Stack=true"
 
 COMMAND=$1
--- a/hcatalog/src/test/e2e/templeton/deployers/config/webhcat/webhcat-site.updateConfig.xml
+++ b/hcatalog/src/test/e2e/templeton/deployers/config/webhcat/webhcat-site.updateConfig.xml
@@ -97,7 +97,7 @@
 
     <property>
         <name>templeton.controller.mr.child.opts</name>
-        <value> -Xmx64m -Djava.net.preferIPv4Stack=true</value>
+        <value> -Xmx64m</value>
     </property>
 
     <property>
--- a/hcatalog/webhcat/svr/src/main/config/webhcat-default.xml
+++ b/hcatalog/webhcat/svr/src/main/config/webhcat-default.xml
@@ -193,7 +193,7 @@
 
   <property>
     <name>templeton.controller.mr.child.opts</name>
-    <value>-server -Xmx256m -Djava.net.preferIPv4Stack=true</value>
+    <value>-server -Xmx256m</value>
     <description>Java options to be passed to templeton controller map task.
         The default value of mapreduce child "-Xmx" (heap memory limit)
         might be close to what is allowed for a map task.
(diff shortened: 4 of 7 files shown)


aturoczy (changes requested):

hcatalog/src/test/e2e/templeton/deployers/config/webhcat/webhcat-site.xml
```diff
@@ -95,7 +95,7 @@
 
     <property>
         <name>templeton.controller.mr.child.opts</name>
-        <value> -Xmx64m -Djava.net.preferIPv4Stack=true</value>
+        <value> -Xmx64m -Djava.net.preferIPv4Stack=false</value>
```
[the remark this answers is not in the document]
SourabhBadhya: From what I saw in the web, if this flag is set to true then we disable IPv6. But does it enable it when set to false?
https://stackoverflow.com/questions/9882357/how-to-set-java-net-preferipv4stack-true-at-runtime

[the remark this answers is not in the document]
difin: The `java.net.preferIPv4Stack` variable name is misleading, in my opinion. 
It controls if IPv6 stack is enabled or disabled.
When `true`, IPv6 stack is completely disabled and a java component that starts with this config won't work in IPv6-only environment.
When `false`, both IPv4 and IPv6 stacks are available.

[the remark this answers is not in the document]
aturoczy: If there is no argument what is the default? I guess without argument it should be available for Ipv4 and Ipv6? Does it need for java explicitly say it is enabled or disabled?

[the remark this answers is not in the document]
difin: The default is `false` starting from Java 8.
Correct, there is no need to set it if we want the default behavior.

aturoczy (approved):

hcatalog/src/test/e2e/templeton/deployers/config/webhcat/webhcat-site.xml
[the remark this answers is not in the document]
aturoczy: +1 from me :)

testutils/hadoop
```diff
@@ -69,6 +69,7 @@ fi
 HADOOP_OPTS="$HADOOP_OPTS -Dhadoop.policy.file=$HADOOP_POLICYFILE"
 
 # Disable ipv6 as it can cause issues
+# TODO: change to -Djava.net.preferIPv4Stack=false after Hadoop adds support for IPv6
 HADOOP_OPTS="$HADOOP_OPTS -Djava.net.preferIPv4Stack=true"
```
okumin, line 73: Can we assume libraries provided by Hadoop can handle IPv6?

difin: No, Hadoop presently doesn't support IPv6

okumin: I guess this PR assumes the environment where Hive is deployed without YARN or HDFS. Just to clarify, I still wonder if shared or client libraries used in Hive, such as hadoop-common or File System API, correctly work with IPv6. Is the answer probably yes?

difin: You are right, this PR has to wait for Hadoop to add IPv6 support.
Hadoop currently doesn't support IPv6, and I haven't found info about any hadoop libraries to support IPv6.

ayushtkn (commented):
@difin AFAIK Djava.net.preferIPv4Stack=true forces Java to use IPv4 over IPv6, even when IPv6 is available.

As of now Hadoop doesn't support IPV6 & moreover we didn't even test Hive fully to see if it works completely f9 with IPV6.

Moreover even if it works f9, don't you think this would be a behaviour change for Hive, like post upgrade suddenly the Hive machines will start preferring IPv6, I believe this would break compat, we should let users decide if they want to go with IPV6 or not. We can just claim from Hive, yes, IPV6 is now supported in Hive & to enable you can remove these flags from the launch scripts or so

difin:
Hi @ayushtkn ,
If we have `java.net.preferIPv4Stack=false` and `java.net.preferIPv6Addresses=false` (which is BTW the default in Java), there won't be behavior change in Hive because it doesn't make IPv6 preferred over IPv4. 

- `java.net.preferIPv4Stack=false` --> makes both IPv4 and IPv6 available (when `true`, IPv6 is completely disabled).
- `java.net.preferIPv6Addresses=false` --> IPv4 addresses are preferred over IPv4.

The only time Java would prefer to use IPv6 with these settings is when a DNS name has only IPv6 mapping.
If Hive precommit tests pass with this change on a dual-stack machine, would you agree with these changes?

ayushtkn:
Sounds good then, if the CI is green on Dual stack machine, Then it is good with me

testutils/hadoop
okumin: I understand your point.

okumin: I checked the documents of Java 8 and 17 just in case.
- https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html
- https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/doc-files/net-properties.html

difin:
> Sounds good then, if the CI is green on Dual stack machine, Then it is good with me

Hi @ayushtkn, I executed all unit tests and q-tests on my local Linux VM, and there were no IPv6-related failures.

ayushtkn (approved):
LGTM

PR #5756, opened 2025-04-09
HIVE-27424: Display dependency:tree in GitHub actions

zabetak:
1. Avoid forcing contributors to upload the dependency tree, which can also be wrong or obsolete. 
2. Avoid having to check and fill an extra field on every submitted PR.

No

No

Inspect output of the macOs (JDK 8) action and verify that dependency tree is shown in the logs.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -37,13 +37,6 @@ If possible, please also clarify if this is a user-facing change compared to the
 If no, write 'No'.
 -->
 
-### Is the change a dependency upgrade?
-<!--
-If yes, please attach a file with output from mvn dependency:tree to validate a complete upgrade of dependency.
-If no, write 'No'.
--->
-
-
 ### How was this patch tested?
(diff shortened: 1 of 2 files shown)


soumyakanti3578 (approved):
LGTM.

Do you think it is a good idea to keep the question: `Is the change a dependency upgrade?` in the template anyway?
Answer could be just a yes or no.

zabetak:
Thanks for the review @soumyakanti3578 ! I wanted to get rid of the question in the template cause I don't feel it helps. If the change is an upgrade is obvious from the summary/diff. Why do you think we should keep it?

okumin (approved):
👍

SourabhBadhya (approved):
LGTM +1

aturoczy (approved):

zabetak:
Thanks for the reviews @soumyakanti3578 @okumin @SourabhBadhya @aturoczy !

PR #5757, opened 2025-04-09
HIVE-27291: Constant reduction in CBO does not work for UNIX_TIMESTAMP

kasakrisz:
* Make `UNIX_TIMESTAMP` not a dynamic function to enable Calcite evaluate it compile time.
* Make `UNIX_TIMESTAMP` `HiveSqlFunction`. `HiveSqlFunction` has a property `isRuntimeConstant` which helps identify functions not allowed in materialized view definitions which are enabled for automatic query rewrite.

If `UNIX_TIMESTAMP` is used in a partition filter expression, the expression is pushed to HMS to load all partitions that satisfy the filter. However, in this case, `UNIX_TIMESTAMP` cannot be evaluated because `SessionState` returns null. This results in a `NullPointerException`, which causes a CBO failure on the HS2 side.

No, but more optimizations might be applied because of CBO.

No.

```
mvn test -Dtest.output.overwrite -DskipSparkTests -Dtest=TestMiniLlapLocalCliDriver -Dqfile=materialized_view_rewrite_12.q,partition_unix_timestamp.q -pl itests/qtest -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveUnixTimestampSqlOperator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveUnixTimestampSqlOperator.java
@@ -29,13 +29,9 @@ import org.apache.calcite.sql.type.ReturnTypes;
  */
 public class HiveUnixTimestampSqlOperator {
   public static final SqlFunction INSTANCE =
-      new SqlFunction("UNIX_TIMESTAMP", SqlKind.OTHER_FUNCTION, ReturnTypes.BIGINT, null,
+      new HiveSqlFunction("UNIX_TIMESTAMP", SqlKind.OTHER_FUNCTION, ReturnTypes.BIGINT, null,
           OperandTypes.or(OperandTypes.NILADIC,
               OperandTypes.or(OperandTypes.STRING, OperandTypes.TIMESTAMP, OperandTypes.DATE),
-              OperandTypes.STRING_STRING), SqlFunctionCategory.NUMERIC) {
-        @Override
-        public boolean isDynamicFunction() {
-          return true;
-        }
+              OperandTypes.STRING_STRING), SqlFunctionCategory.NUMERIC, true, true) {
       };
 }
--- a/ql/src/test/queries/clientpositive/cbo_join_transitive_pred_loop_1.q
+++ b/ql/src/test/queries/clientpositive/cbo_join_transitive_pred_loop_1.q
@@ -1,5 +1,8 @@
 set hive.optimize.join.disjunctive.transitive.predicates.pushdown=false;
 
+-- HiveProject(month=[CAST(202503):INTEGER])
+--! qt:replace:/(.*month=\[CAST\()\d+(\).*)/$1#Masked#$2/
+
 CREATE TABLE test1 (act_nbr string);
 CREATE TABLE test2 (month int);
 CREATE TABLE test3 (mth int, con_usd double);
--- a/ql/src/test/results/clientpositive/llap/cbo_filter_proj_transpose_noinputref.q.out
+++ b/ql/src/test/results/clientpositive/llap/cbo_filter_proj_transpose_noinputref.q.out
@@ -18,9 +18,7 @@ unix_timestamp(void) is deprecated. Use current_timestamp instead.
 unix_timestamp(void) is deprecated. Use current_timestamp instead.
 unix_timestamp(void) is deprecated. Use current_timestamp instead.
 unix_timestamp(void) is deprecated. Use current_timestamp instead.
-unix_timestamp(void) is deprecated. Use current_timestamp instead.
-unix_timestamp(void) is deprecated. Use current_timestamp instead.
-unix_timestamp(void) is deprecated. Use current_timestamp instead.
+Warning: Shuffle Join MERGEJOIN[11][tables = [$hdt$_0, $hdt$_1]] in Stage 'Reducer 2' is a cross product
 PREHOOK: query: EXPLAIN CBO SELECT c.m
 FROM (
   SELECT substr(from_unixtime(unix_timestamp(), 'yyyy-MM-dd'), 1, 1) as m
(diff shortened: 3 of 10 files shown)


ql/src/test/results/clientpositive/llap/cbo_filter_proj_transpose_noinputref.q.out
```diff
@@ -42,13 +40,11 @@ POSTHOOK: Input: default@test1
...
 CBO PLAN:
-HiveProject(m=[$0])
-  HiveJoin(condition=[=($0, $1)], joinType=[inner], algorithm=[none], cost=[not available])
-    HiveProject(m=[substr(FROM_UNIXTIME(UNIX_TIMESTAMP(), _UTF-16LE'yyyy-MM-dd':VARCHAR(2147483647) CHARACTER SET "UTF-16LE"), 1, 1)])
-      HiveFilter(condition=[=(substr(FROM_UNIXTIME(UNIX_TIMESTAMP(), _UTF-16LE'yyyy-MM-dd':VARCHAR(2147483647) CHARACTER SET "UTF-16LE"), 1, 1), _UTF-16LE'2')])
-        HiveProject(DUMMY=[0])
-          HiveTableScan(table=[[default, test1]], table:alias=[test1])
-    HiveProject($f0=[substr(FROM_UNIXTIME(UNIX_TIMESTAMP(), _UTF-16LE'yyyy-MM-dd':VARCHAR(2147483647) CHARACTER SET "UTF-16LE"), 1, 1)])
-      HiveFilter(condition=[AND(=($0, substr(FROM_UNIXTIME(UNIX_TIMESTAMP(), _UTF-16LE'yyyy-MM-dd':VARCHAR(2147483647) CHARACTER SET "UTF-16LE"), 1, 1)), =(substr(FROM_UNIXTIME(UNIX_TIMESTAMP(), _UTF-16LE'yyyy-MM-dd':VARCHAR(2147483647) CHARACTER SET "UTF-16LE"), 1, 1), _UTF-16LE'2'))])
+HiveProject(m=[CAST(_UTF-16LE'2':VARCHAR(2147483647) CHARACTER SET "UTF-16LE"):VARCHAR(2147483647) CHARACTER SET "UTF-16LE"])
```
kasakrisz, line 43: The constant `2` is the first digit of the current year. I didn't mask it because it will only change in the next millennium. :)

ql/src/test/queries/clientpositive/cbo_join_transitive_pred_loop_1.q
```diff
@@ -1,5 +1,8 @@
 set hive.optimize.join.disjunctive.transitive.predicates.pushdown=false;
 
+-- HiveProject(month=[CAST(202503):INTEGER])
+--! qt:replace:/(.*month=\[CAST\()\d+(\).*)/$1#Masked#$2/
```
zabetak, line 4: Instead of regex replace I think it will be safer and more robust to new plan changes to use `hive.test.currenttimestamp`.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveUnixTimestampSqlOperator.java
```diff
@@ -29,13 +29,9 @@
...
   public static final SqlFunction INSTANCE =
-      new SqlFunction("UNIX_TIMESTAMP", SqlKind.OTHER_FUNCTION, ReturnTypes.BIGINT, null,
+      new HiveSqlFunction("UNIX_TIMESTAMP", SqlKind.OTHER_FUNCTION, ReturnTypes.BIGINT, null,
           OperandTypes.or(OperandTypes.NILADIC,
               OperandTypes.or(OperandTypes.STRING, OperandTypes.TIMESTAMP, OperandTypes.DATE),
-              OperandTypes.STRING_STRING), SqlFunctionCategory.NUMERIC) {
-        @Override
-        public boolean isDynamicFunction() {
-          return true;
-        }
```
zabetak, line 39: Not sure how much we rely on the notion of dynamic function but this PR changes the current behavior. Before the changes we were returning true and after this change we will return false.

zabetak (approved):
LGTM, some minor comments/observations.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveUnixTimestampSqlOperator.java
kasakrisz: Yes, you are right, but this is one of the goal of this patch. Otherwise `UNIX_TIMESTAMP` doesn't evaluated during constant folding by Calcite.

Caclcite 1.25
https://github.com/apache/calcite/blob/68b02dfd4af15bc94a91a0cd2a30655d04439555/core/src/main/java/org/apache/calcite/rel/rules/ReduceExpressionsRule.java#L1094-L1099

From Calcite 1.28 treating dynamic functions as constant or not is configurable
https://issues.apache.org/jira/browse/CALCITE-2736

This is a rule level config though.

ql/src/test/queries/clientpositive/cbo_join_transitive_pred_loop_1.q
kasakrisz: IMHO the presence of setting like `hive.test.currenttimestamp` is not safe. Test code should not be shipped with production code.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveUnixTimestampSqlOperator.java
zabetak: I guess we can't do anything else at the moment. If necessary we can revisit the choice once we upgrade to 1.28+

ql/src/test/queries/clientpositive/cbo_join_transitive_pred_loop_1.q
zabetak: I am not a big fan of such settings either but since we have them we could as well use them. Anyways, don't feel strongly about it so pick whatever option you feel its best.

PR #5758, opened 2025-04-09
HIVE-28782: Ensure Hive constructs URI's with IPv6 literals safely.

difin:

Added proper escaping of IPv6 literals where Hive constructs URI's with a host and port.

When an IPv6 address and a port are part of a URI (like `<scheme>://<ip>:<port>`/, IPv6 must be surrounded by square brackets, because it may contain ":" internally.

No

No

Hive pre-commit tests.


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/common/IPStackUtils.java
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.hive.common;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class IPStackUtils {
```
ayushtkn, line 23: We are copying some code over here, can't we keep everything at one place, either of the jars would have dependency of one another. I think hive-common has a dependency of metastore-common, so everything can be kept in metastore-common?
https://github.com/apache/hive/blob/db041fd7b71a15a8bd1a997b0908a02e8124d7ac/common/pom.xml#L46-L49

difin: Yes, it's much better to have have everything in one place. I moved the utils class to metastore-common.

standalone-metastore/metastore-common/pom.xml
```diff
@@ -333,6 +333,12 @@
       <artifactId>slf4j-simple</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <version>${junit.jupiter.version}</version>
+      <scope>test</scope>
+    </dependency>
```
ayushtkn, line 341: Why is this kicking in here? the junit depedency in past have led to tests being silently ignored, can you check if the total number of tests that ran didn't reduce. I suspect they did

ayushtkn (commented):
Thanx @difin , can you check the Junit thing. rest everything LGTM

standalone-metastore/metastore-common/pom.xml
difin: I needed this after moving `IPStackUtilsTest` class from `common` to `metastore-common` because tests were using some methods from junit5. Reverted this change in pom.xml and adapted the tests.

ayushtkn (approved):
LGTM

PR #5763, opened 2025-04-10
HIVE-28886: CalcitePlanner throws NullPointerException when column name case in ColumnStatisticsObj differs from the table schema

deniskuzZ:

Transform the column name in ColumnStatsObj to lowercase

NullPointerException Fix

No

No

q test

### DIFF
--- a/iceberg/iceberg-handler/src/test/results/positive/col_stats.q.out
+++ b/iceberg/iceberg-handler/src/test/results/positive/col_stats.q.out
@@ -111,12 +111,12 @@ POSTHOOK: type: QUERY
 POSTHOOK: Input: default@tbl_ice_puffin
 POSTHOOK: Output: default@tbl_ice_puffin
 POSTHOOK: Output: default@tbl_ice_puffin
-PREHOOK: query: analyze table tbl_ice_puffin  compute statistics for columns
+PREHOOK: query: analyze table tbl_ice_puffin compute statistics for columns
 PREHOOK: type: ANALYZE_TABLE
 PREHOOK: Input: default@tbl_ice_puffin
 PREHOOK: Output: default@tbl_ice_puffin
 PREHOOK: Output: hdfs://### HDFS PATH ###
-POSTHOOK: query: analyze table tbl_ice_puffin  compute statistics for columns
+POSTHOOK: query: analyze table tbl_ice_puffin compute statistics for columns
 POSTHOOK: type: ANALYZE_TABLE
 POSTHOOK: Input: default@tbl_ice_puffin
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/stats/ColumnStatisticsObjTranslator.java
```diff
@@ -59,7 +59,7 @@ public static ColumnStatisticsObj readHiveColumnStatistics(String columnName, St
       List<ColumnStatsField> columnStatsFields, int start, List<? extends StructField> fields,
       List<Object> values) throws HiveException {
     ColumnStatisticsObj statsObj = new ColumnStatisticsObj();
-    statsObj.setColName(columnName);
+    statsObj.setColName(columnName.toLowerCase());
```
aturoczy, line 62: How this fix the lower case? Somewhere is a compare the column name and in other places could be capital?

deniskuzZ: it's always in lowercase in Hive schema, but not in ColumnStatisticsObj

aturoczy (approved):

SourabhBadhya (approved):
LGTM +1 (pending tests)

iceberg/iceberg-handler/src/test/results/positive/col_stats.q.out
```diff
@@ -459,24 +469,43 @@ Stage-0
...
           Output:["_col0","_col1","_col2"]
         <-Map 1 [SIMPLE_EDGE] vectorized
           SHUFFLE [RS_6]
-            Select Operator [SEL_5] (rows=9 width=95)
+            Select Operator [SEL_5] (rows=6 width=192)
               Output:["_col0","_col1","_col2"]
-              TableScan [TS_0] (rows=9 width=95)
-                default@tbl_ice,tbl_ice,Tbl:COMPLETE,Col:COMPLETE,Output:["a","b","c"]
+              TableScan [TS_0] (rows=6 width=192)
+                default@tbl_ice_puffin,tbl_ice_puffin,Tbl:COMPLETE,Col:PARTIAL,Output:["a","b","c"]
```
kasakrisz, line 479: The column stats state changed from `complete` to `partial`. Is this expected?

deniskuzZ: yes, that is expected as we compute stats for 2 columns out of 3. BTW, the original q test had a typo and did a validation on a wrong table

PR #5766, opened 2025-04-11
HIVE-28848:Remove DFS_URI auth from ALTER_PARTITION if there is no ch…

rtrivedi12:
…ange in partition location

This change skips the DFS_URI authorization if there is no change in the PARTITION location in the ALTER_PARTITION event.

If the old partition and new partition Uri are the same or the old partition location is empty, then skip adding DFS_URI for authorization.

The LOAD DATA command invokes the ALTER_PARTITION event to update stats for the newly loaded partition. ALTER_PARTITION authorizes DFS_URI for the partition location even when the partition locations do not change. This requires the Hadoop SQL  policy for DFS_URI to be defined for the partition location for LOAD DATA commands.

Yes, after this change, users will no longer need the DFS_URI policy for target table partition location.

No

Manual testing


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -6051,8 +6051,9 @@ private void alter_partition_core(String catName, String db_name, String tbl_nam
     Exception ex = null;
     try {
       Table table = getMS().getTable(catName, db_name, tbl_name, null);
+      Partition oldPartition = getMS().getPartition(catName, db_name, tbl_name, part_vals);
```
saihemanth-cloudera, line 6054: nit: we fetch oldPart at L6062, can you see if you can eliminate this redundant call?

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/AlterPartitionEvent.java
```diff
@@ -76,10 +76,15 @@ private List<HivePrivilegeObject> getOutputHObjs() {
...
     Partition newPartition = event.getNewPartition();
+    Partition oldPartition = event.getOldPart();
+
     String    newUri       = (newPartition != null) ? getSdLocation(newPartition.getSd()) : "";
+    String    oldUri       = (oldPartition != null) ? getSdLocation(oldPartition.getSd()) : "";
 
-    if (StringUtils.isNotEmpty(newUri)) {
-        ret.add(getHivePrivilegeObjectDfsUri(newUri));
+    // Skip DFS_URI auth if old and new partition uri are empty or same
+    if (StringUtils.isNotEmpty(newUri) && StringUtils.isNotEmpty(oldUri) && !StringUtils.equalsIgnoreCase(oldUri,
```
wecharyu, line 85: Do we need `DFS_URI` auth when `newUri` is not empty but `oldUri` is empty? Also consider add unit test for this change.

dengzhhu653: change the condition to `StringUtils.isNotEmpty(newUri) & !StringUtils.equalsIgnoreCase(oldUri, newUri)` ?

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/AlterPartitionEvent.java
```diff
@@ -76,10 +76,15 @@ private List<HivePrivilegeObject> getOutputHObjs() {
     ret.add(getHivePrivilegeObject(event.getTable()));
 
     Partition newPartition = event.getNewPartition();
+    Partition oldPartition = event.getOldPart();
```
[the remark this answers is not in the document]
rtrivedi12: @deniskuzZ Both Partition and Table objects in `PreAlterPartitionEvent` are  thrift table and thrift partition and would need to be translated to org.apache.hadoop.hive.ql.metadata.Table and org.apache.hadoop.hive.ql.metadata.Partition. But I think this makes sense to avoid modification in PreAlterPartitionEvent and HMSHandler.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/AlterPartitionEvent.java
rtrivedi12: @wecharyu @dengzhhu653  We want to skip DFS_URI auth when newUri is not empty but oldUri is empty like in case of stats update during LOAD PARTITION, it would have empty old uri and non-empty new URI.
But I see your point this condition would be anyways covered by `!StringUtils.equalsIgnoreCase(oldUri, newUri)`.

wecharyu: Not quite understand why not need check the newUri auth when oldUri is empty but newUri is not.

> But I see your point this condition would be anyways covered by !StringUtils.equalsIgnoreCase(oldUri, newUri).

But the check would fail-fast by `StringUtils.isNotEmpty(oldUri)` in your current implementation.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
rtrivedi12: I removed this redundant call from `HMSHandler`.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/AlterPartitionEvent.java
rtrivedi12: @wecharyu I have updated the condition to `StringUtils.isNotEmpty(newUri) & !StringUtils.equalsIgnoreCase(oldUri, newUri)` now.  And this patch proposes skipping DFS_URI authorization if there is no change in partition location. So, if new URI is non-empty and old uri is empty , ALTER PARTITION  might be triggered for other HiveOperationType like ALTERPARTITION_FILEFORMAT, ALTERPARTITION_SERDEPROPERTIES which does not require DFS_URI auth.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/AlterPartitionEvent.java
```diff
@@ -76,10 +79,17 @@ private List<HivePrivilegeObject> getOutputHObjs() {
     ret.add(getHivePrivilegeObject(event.getTable()));
 
     Partition newPartition = event.getNewPartition();
-    String    newUri       = (newPartition != null) ? getSdLocation(newPartition.getSd()) : "";
+    String oldUri = "";
 
-    if (StringUtils.isNotEmpty(newUri)) {
-        ret.add(getHivePrivilegeObjectDfsUri(newUri));
+    if (event.getOldPartVals() != null && event.getOldPartVals().size() > 0) {
```
dengzhhu653, line 84: should change `null` to `tmpPart.getValues` in this line: https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java#L6184?

ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java
```diff
@@ -696,6 +700,107 @@ public void testU_DropDataConnector_authorizedUser() {
...
+      List<HivePrivilegeObject> tableOutputs = outputs.stream()
+          .filter(o -> o.getType() == HivePrivilegeObject.HivePrivilegeObjectType.DFS_URI)
+          .collect(Collectors.toList());
+
+      assertEquals("Should have one DFS_URI privilege object", 1, tableOutputs.size());
+      HivePrivilegeObject DFSUriObj = tableOutputs.get(0);
+
+      assertEquals("Output object should be a table",
+          HivePrivilegeObject.HivePrivilegeObjectType.DFS_URI,
+          DFSUriObj.getType());
```
dengzhhu653: this should always be a `DFS_URI` obj, can we validate the real location instead?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/events/PreAlterPartitionEvent.java
```diff
@@ -18,14 +18,14 @@
 
 package org.apache.hadoop.hive.metastore.events;
 
-import java.util.List;
```
dengzhhu653: nit: please revert the change if possible

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/AlterPartitionEvent.java
rtrivedi12: We don't have any changes in `HMSHandler`  as part of this change. IIUC  `null` is sent for oldPartition values and `tmpPart` is new partition here https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java#L6184

rtrivedi12:
@dengzhhu653 Thanks for the review!  I have addressed review comments.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/AlterPartitionEvent.java
dengzhhu653: The oldPartition values are the same as the tmpPart values in this method

dengzhhu653 (approved):

dengzhhu653:
The `split-15`  is failing constantly, could we check it please? I will close the PR and retry the test pipeline

PR #5769, opened 2025-04-12
HIVE-28903: Skip deleting archived path when drop partition/table

wecharyu:
When drop archived partition/table, only the original path of partition should be considered as a deletion candidate

Hadoop archived path is not supported to be deleted, in current implement the drop_partition/table calls would drop metadata successfully but throw exception when deleting the archived paths.

No.

No.

- Add a qtest
```bash
mvn test -pl itests/qtest -Pitests -Dtest=org.apache.hadoop.hive.cli.TestMiniLlapLocalCliDriver -Dqfile=archive_drop.q
```
- Test locally by sql:
```sql
-- create a table
create table tbl (id int) partitioned by (dt date, region string);

SET hive.execution.engine=mr;
insert overwrite table tbl partition (dt='2025-01-01', region='us') select 1;
insert overwrite table tbl partition (dt='2025-01-01', region='sg') select 2;

-- change to managed table
ALTER TABLE tbl SET TBLPROPERTIES ('EXTERNAL'='FALSE');

-- archive partitions
SET hive.archive.enabled=true;
ALTER TABLE tbl ARCHIVE PARTITION (dt='2025-01-01');

-- drop partition
alter table tbl drop partition (dt='2025-01-01', region='us');

-- drop table
drop table tbl;
```

### DIFF
--- a/ql/src/test/queries/clientpositive/archive_drop.q
+++ b/ql/src/test/queries/clientpositive/archive_drop.q
@@ -1,0 +1,24 @@
+set hive.mapred.mode=nonstrict;
+set hive.archive.enabled = true;
+
+create database test_db;
+
+create table test_db.test_tbl (id int, name string) partitioned by (dt date, hr string);
+
+insert overwrite table test_db.test_tbl partition (dt='2025-04-01', hr='11') select 1, 'tom';
+insert overwrite table test_db.test_tbl partition (dt='2025-04-01', hr='12') select 2, 'jerry';
+insert overwrite table test_db.test_tbl partition (dt='2025-04-01', hr='13') select 3, 'spike';
+
+show partitions test_db.test_tbl;
+
+alter table test_db.test_tbl archive partition (dt='2025-04-01');
+
+show partitions test_db.test_tbl;
+
+alter table test_db.test_tbl drop partition (dt='2025-04-01',hr='12');
+
+show partitions test_db.test_tbl;
+
+select * from test_db.test_tbl;
+
+drop table test_db.test_tbl;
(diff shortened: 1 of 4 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -3179,8 +3179,7 @@ private void deleteDataExcludeCmroot(Path path, boolean ifPurge, boolean shouldE
         wh.deleteDir(path, true, ifPurge, shouldEnableCm);
       }
     } catch (Exception e) {
-      LOG.error("Failed to delete directory: " + path +
-          " " + e.getMessage());
+      LOG.error("Failed to delete directory: " + path, e);
```
SourabhBadhya, line 3182: nit: Its better to use `{}` instead of concatenation in logs

SourabhBadhya (commented):
Please consider adding tests for this.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: Addressed.

ql/src/test/queries/clientpositive/archive_drop.q
```diff
@@ -0,0 +1,24 @@
...
+
+show partitions test_db.test_tbl;
+
+alter table test_db.test_tbl drop partition (dt='2025-04-01',hr='12');
+
+show partitions test_db.test_tbl;
+
+select * from test_db.test_tbl;
+
+drop table test_db.test_tbl;
```
SourabhBadhya, line 24: @wecharyu Can we do a list of the archived partition path, so that the files are intact after the SQL statements?

wecharyu: Addressed.

SourabhBadhya (approved):
LGTM +1 (pending tests)

PR #5770, opened 2025-04-14
HIVE-28898: Include table ownership details in getTableMetas() for Ra…

jjiang037:
…ngerHiveAuthorizer

Passing table ownership information to RangerHiveAuthorizer when ranger starts using getTableMetas()

Ranger would fetch ownership information of table if hms doesn't provide initially.

No

No

Unit test

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/filtercontext/TableFilterContext.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/filtercontext/TableFilterContext.java
@@ -20,22 +20,23 @@
 
 package org.apache.hadoop.hive.ql.security.authorization.plugin.metastore.filtercontext;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.hadoop.hive.metastore.api.TableMeta;
 import org.apache.hadoop.hive.ql.security.authorization.plugin.HiveOperationType;
 import org.apache.hadoop.hive.ql.security.authorization.plugin.HivePrivilegeObject;
 import org.apache.hadoop.hive.ql.security.authorization.plugin.HivePrivilegeObject.HivePrivilegeObjectType;
 import org.apache.hadoop.hive.ql.security.authorization.plugin.HivePrivilegeObject.HivePrivObjectActionType;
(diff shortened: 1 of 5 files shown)


Aggarwal-Raghav:
@jjiang037, thanks for the PR.

Quick question, how is it different from [HIVE-27285](https://issues.apache.org/jira/browse/HIVE-27285), IIRC for both database and Table/View we are passing owner information in hivePrivObject.

jjiang037:
Hi @Aggarwal-Raghav, thanks for the review and question!

You're right—HIVE-27285 introduced ownership info (ownerName, ownerType) in TableMeta and updated filterTableMetas to work with it. But it didn’t use that info within the TableFilterContext.

This PR (HIVE-28898) addresses that gap:

Wires ownership into TableFilterContext: Adds a createFromTableMetas factory method that converts TableMeta to Table and copies ownerName and ownerType.

Populates HivePrivilegeObjects properly: getInputHObjs now includes owner info when building the list, enabling plugins like Ranger to use it.

Ensures correct flow in HMSHandler: Updates get_all_tables to fetch TableMeta (with ownership) before filtering, so it propagates down properly.

In short: HIVE-27285 exposed ownership data; this PR makes sure it’s actually used during filtering and authorization.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/filtercontext/TableFilterContext.java
```diff
@@ -53,36 +54,56 @@ public TableFilterContext(String dbName, List<String> tableNames) {
...
+  public HiveMetaStoreAuthzInfo getAuthzContext() {
+    HiveMetaStoreAuthzInfo ret = new HiveMetaStoreAuthzInfo(preEventContext, HiveOperationType.QUERY, getInputHObjs(), getOutputHObjs(), null);
     return ret;
   }
 
   private List<HivePrivilegeObject> getInputHObjs() {
-    LOG.debug("==> TableFilterContext.getOutputHObjs()");
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("==> TableFilterContext.getOutputHObjs()");
+    }
```
saihemanth-cloudera, line 86: nit: You don't need this if condition for debug log.

dengzhhu653 (approved):

saihemanth-cloudera (approved):

PR #5776, opened 2025-04-19
HIVE-28891: Fix spotbugs issues in hive-shims with spotbugs-maven-plugin 4.8.6.6

ramitg254:

Upgrade spotbugs-maven-plugin to 4.8.6.6

Upgraded plugin introduces new rules to identify potential bugs.

No

Locally, running spotbugs check and fixing the bugs, followed by running basic manual test on hive by building locally and running queries.


shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,8 +210,20 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
     this.aclStatus = aclStatus;
   }
 
+  public static HadoopFileStatus createInstance (Configuration conf, FileSystem fs, Path file) throws IOException {
```
okumin, line 213:
```suggestion
  public static HadoopFileStatus createInstance(Configuration conf, FileSystem fs, Path file) throws IOException {
```

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,8 +210,20 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
...
+    FileStatus fileStatus = fs.getFileStatus(file);
+    return new HadoopFileStatus(conf, fs, fileStatus, file);
+  }
+
   public FileStatus getFileStatus() {
-    return fileStatus;
+    Path symlink;
+    try {
+      symlink= fileStatus.getSymlink();
```
okumin, line 221:
```suggestion
      symlink = fileStatus.getSymlink();
```

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,8 +210,20 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
...
+    }catch (IOException e){
```
okumin, line 222:
```suggestion
    } catch (IOException e) {
```

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,8 +210,20 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
...
+      symlink=null;
```
okumin, line 223:
```suggestion
      symlink = null;
```

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,8 +210,20 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
...
+    }
+    FileStatus other=fileStatus;
```
okumin, line 225:
```suggestion
    FileStatus other = fileStatus;
```

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,8 +210,20 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
...
+    return new FileStatus(other.getLen(), other.isDirectory(), other.getReplication(), other.getBlockSize(), other.getModificationTime(), other.getAccessTime(), other.getPermission(), other.getOwner(), other.getGroup(), symlink, other.getPath());
```
okumin, line 226: The length of a single line has to be smaller than or equal to 120 in our coding style.

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,8 +210,20 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
...
   }
 
+  public static HadoopFileStatus createInstance (Configuration conf, FileSystem fs, Path file) throws IOException {
+    FileStatus fileStatus = fs.getFileStatus(file);
+    return new HadoopFileStatus(conf, fs, fileStatus, file);
+  }
+
   public FileStatus getFileStatus() {
-    return fileStatus;
+    Path symlink;
```
okumin, line 219: I would like to know the intention of this change

okumin (commented):
I'm still reviewing this PR. I found that SonarCloud reported some issues. Could you please resolve ones that this PR should address?
https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=5776&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
ramitg254: i used this parameterised constructor to prevent "EI_EXPOSE_REP"

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
ramitg254: basically this is one of the parameter i am going to pass in the parameterised constructor of fileStatus to prevent `"EI_EXPOSE_REP"` and symlink can throw exception so try catch is used below

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
ramitg254: I'll break it into 2-3 lines

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
ramitg254: done

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
ramitg254: done

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
ramitg254: done

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
ramitg254: done

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
ramitg254: done

ramitg254:
> I'm still reviewing this PR. I found that SonarCloud reported some issues. Could you please resolve ones that this PR should address? https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=5776&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

regarding the issues flagged by sonarcloud under maintainability issues focussing on exceptions thrown, I haven't made changes to those, they were already present in the codebase i just wrapped them inside a wrapper function in some cases which is to solve spotbugs warnings which is now scanned by sonarcloud and it is flagging them now although I think they are meaningless

ramitg254:
> I'm still reviewing this PR. I found that SonarCloud reported some issues. Could you please resolve ones that this PR should address? https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=5776&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

and regarding the reliability ones the first one is flagging for re-interrupt for an interruption exception which is handled in factory method of that constructor but sonarcloud is not smart enough to understand that and the other one which is for not closing a closable object inside a createInstance factory method which should not be done because that method is used in place of constructor to create object so it does not make sense to close object inside that method

shims/0.23/src/main/java/org/apache/hadoop/hive/shims/Hadoop23Shims.java
```diff
@@ -774,26 +772,28 @@ public FutureDataInputStreamBuilder openFile(Path path) throws IOException, Unsu
...
-        }
-      };
+          @Override
+          public LocatedFileStatus next() throws IOException {
+            LocatedFileStatus result = stats.next();
+            return new LocatedFileStatus(ProxyFileSystem23.super.swizzleFileStatus(result, false),
+                result.getBlockLocations());
+          }
+        };
+      } catch (IOException e) {
```
okumin, line 794: Can we remove this try-catch because this method throws an IOException?

shims/0.23/src/main/java/org/apache/hadoop/hive/shims/Hadoop23Shims.java
```diff
@@ -1334,10 +1335,23 @@ public static class HdfsEncryptionShim implements HadoopShims.HdfsEncryptionShim
...
     private final Configuration conf;
 
-    public HdfsEncryptionShim(URI uri, Configuration conf) throws IOException {
+    public static HdfsEncryptionShim createInstance(URI uri, Configuration conf) throws IOException {
+      HdfsAdmin hadmin = null;
+      KeyProvider keyP = null;
+      try {
+        hadmin = new HdfsAdmin(uri, conf);
+        keyP = hadmin.getKeyProvider();
+      } catch (IOException e) {
```
okumin, line 1344: Can we remove this try-catch?

shims/0.23/src/main/java/org/apache/hadoop/mapred/WebHCatJTShim23.java
```diff
@@ -51,19 +48,36 @@ public class WebHCatJTShim23 implements WebHCatJTShim {
...
+        }
+      });
+    } catch (InterruptedException ex) {
+      eMessage.add(ex.getMessage());
+      flag[0] = -2;
+      this.jc = null;
+    } catch (IOException ex) {
+      flag[0] = -1;
+      this.jc = null;
     }
```
okumin, line 81: Can we pull out the error-handling from the constructor like this one?

```
  public static WebHCatJTShim createInstance(final Configuration conf, final UserGroupInformation ugi)
      throws IOException {
    JobClient jc;
    try {
      jc = ugi.doAs((PrivilegedExceptionAction<JobClient>) () -> {
        //create this in doAs() so that it gets a security context based passed in 'ugi'
        return new JobClient(conf);
      });
      throw new RuntimeException("Failed to create JobClient", e);
    }
    return new WebHCatJTShim23(conf, jc);
  }

  private WebHCatJTShim23(final Configuration conf, final JobClient jc) {
    this.conf = new Configuration(conf);
    this.jc = jc;
  }
```

shims/common/pom.xml
```diff
@@ -106,6 +106,10 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-storage-api</artifactId>
+    </dependency>
```
okumin, line 112: Do we need this one?

shims/common/src/main/java/org/apache/hadoop/fs/ProxyFileSystem.java
```diff
@@ -207,21 +199,25 @@ public FileStatus[] listStatus(Path f) throws IOException {
...
+          return orig.hasNext();
+        }
+
+        @Override
+        public FileStatus next() throws IOException {
+          FileStatus ret = orig.next();
+          return swizzleFileStatus(ret, false);
+        }
+      };
+    } catch (IOException e) {
```
okumin, line 218: Can we remove this try-catch?

shims/common/src/main/java/org/apache/hadoop/hive/shims/ShimLoader.java
```diff
@@ -87,7 +87,9 @@ public static HadoopShims getHadoopShims() {
         }
       }
     }
-    return hadoopShims;
+    synchronized (ShimLoader.class) {
+      return loadShims(HADOOP_SHIM_CLASSES, HadoopShims.class);
+    }
```
okumin, line 92: I think we intentionally return a cached `hadoopShims` for performance. I guess we need to mark it as a false positive.

- https://github.com/spotbugs/spotbugs/issues/1747
- https://github.com/apache/hbase/pull/4787/files

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
```diff
@@ -185,10 +185,27 @@ public float getProgress() throws IOException {
      * A generic RecordReader that can hand out different recordReaders
      * for each chunk in the CombineFileSplit.
      */
-    public CombineFileRecordReader(JobConf job, CombineFileSplit split,
-        Reporter reporter,
-        Class<RecordReader<K, V>> rrClass)
-        throws IOException {
+    public static <K,V> CombineFileRecordReader<K,V> createInstance(JobConf job, CombineFileSplit split,
+                                                                    Reporter reporter,
+                                                                    Class<RecordReader<K, V>> rrClass) throws IOException{
```
okumin, line 190:
```suggestion
                                                                    Class<RecordReader<K, V>> rrClass) throws IOException {
```

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
okumin: Thanks. In my current impression, it might be safer to keep using the original status with an annotation. That's because `FileStatus` is heavily extended by various classes, and it is not self-evident that we can clone it as the parent class.
https://github.com/apache/hadoop/blob/rel/release-3.4.1/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/protocol/HdfsNamedFileStatus.java

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
ramitg254: done

shims/common/src/main/java/org/apache/hadoop/hive/shims/ShimLoader.java
ramitg254: done returned `hadoopShims` and suppressed the warning

shims/common/src/main/java/org/apache/hadoop/fs/ProxyFileSystem.java
ramitg254: yeah, done removed

shims/common/pom.xml
ramitg254: this i added because there was no dependency for suppressing warning annotation for hive-shims-common so i used one from hive-storage-api because other ones were leading to cyclic dependency

shims/0.23/src/main/java/org/apache/hadoop/mapred/WebHCatJTShim23.java
ramitg254: Thanks this one is good approach i added this one instead of the hack i was using 😅

shims/0.23/src/main/java/org/apache/hadoop/hive/shims/Hadoop23Shims.java
ramitg254: done removed

shims/0.23/src/main/java/org/apache/hadoop/hive/shims/Hadoop23Shims.java
ramitg254: done removed

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,6 +211,12 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
     this.aclStatus = aclStatus;
   }
 
+  public static HadoopFileStatus createInstance(Configuration conf, FileSystem fs, Path file) throws IOException {
+    FileStatus fileStatus = fs.getFileStatus(file);
+    return new HadoopFileStatus(conf, fs, fileStatus, file);
+  }
+
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "intended")
```
ramitg254, line 219: done returned the fileStatus itself and suppressed the warning

shims/common/src/main/java/org/apache/hadoop/hive/io/HdfsUtils.java
```diff
@@ -211,6 +211,12 @@ public HadoopFileStatus(Configuration conf, FileSystem fs, Path file) throws IOE
...
   public FileStatus getFileStatus() {
     return fileStatus;
   }
```
ramitg254, line 222: done returned fileStatus itself and suppressed the warning

shims/0.23/src/main/java/org/apache/hadoop/hive/shims/Hadoop23Shims.java
```diff
@@ -1334,10 +1331,19 @@ public static class HdfsEncryptionShim implements HadoopShims.HdfsEncryptionShim
 
     private final Configuration conf;
 
-    public HdfsEncryptionShim(URI uri, Configuration conf) throws IOException {
+    public static HdfsEncryptionShim createInstance(URI uri, Configuration conf) throws IOException {
+      HdfsAdmin hadmin = null;
+      KeyProvider keyP = null;
```
okumin, line 1336: I think we don't need to initialize it as null.

shims/0.23/src/main/java/org/apache/hadoop/mapred/WebHCatJTShim23.java
```diff
@@ -37,10 +37,7 @@
 import java.io.IOException;
 import java.net.URI;
 import java.security.PrivilegedExceptionAction;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
okumin, line 40: If I remember correctly, we don't use a wildcard import.
https://github.com/apache/hive/blob/be59bd0b445096e431c18bbfa0d682c78ba18b76/checkstyle/checkstyle.xml#L152-L155

You can run checkstyle with `mvn checkstyle:checkstyle -DskipTests -pl shims/0.23`, and see the result in the target directory. It is unrealistic to fix all issues, but it would be great if we can keep the codestyle consistent as much as possible especially for new changes.

```
cat shims/0.23/target/checkstyle-result.xml | grep -A 1 WebHCatJTShim23
<file name="/Users/shohei.okumiya/ghq/github.com/okumin/hive/shims/0.23/src/main/java/org/apache/hadoop/mapred/WebHCatJTShim23.java">
<error line="40" severity="warning" message="Using the &apos;.*&apos; form of import should be avoided - java.util.*." source="com.puppycrawl.tools.checkstyle.checks.imports.AvoidStarImportCheck"/>
```

shims/common/pom.xml
okumin: That makes sense... Can we move this class to shims-common so that both shims-common and 0.23 can share it? Adding the whole dependency just for the annotation might confuse maintainers or the chaos might cause a future problem.
https://github.com/apache/hive/blob/master/shims/0.23/src/main/java/org/apache/hadoop/util/SuppressFBWarnings.java

okumin (commented):
Thanks for handling. Hopefully, this is the last comment.

ramitg254:
shims/common/src/main/java/org/apache/hadoop/util/SuppressFBWarnings.java
added this in place of dependency

shims/0.23/src/main/java/org/apache/hadoop/mapred/WebHCatJTShim23.java
ramitg254: changed to individual imports

shims/0.23/src/main/java/org/apache/hadoop/hive/shims/Hadoop23Shims.java
ramitg254: removed null

shims/common/pom.xml
ramitg254: [shims/common/src/main/java/org/apache/hadoop/util/SuppressFBWarnings.java](https://github.com/apache/hive/pull/5776/files#diff-0f95d1dce726d2eec066cf58c509df20f4787698ff8360715265492e97bd4a6e)
added this in place of dependency

shims/0.23/src/main/java/org/apache/hadoop/hive/shims/Hadoop23Shims.java
```diff
@@ -1334,10 +1331,19 @@ public static class HdfsEncryptionShim implements HadoopShims.HdfsEncryptionShim
 
     private final Configuration conf;
 
-    public HdfsEncryptionShim(URI uri, Configuration conf) throws IOException {
+    public static HdfsEncryptionShim createInstance(URI uri, Configuration conf) throws IOException {
+      HdfsAdmin hadmin;
+      KeyProvider keyP;
+      hadmin = new HdfsAdmin(uri, conf);
+      keyP = hadmin.getKeyProvider();
```
okumin, line 1338:
```suggestion
      HdfsAdmin hadmin = new HdfsAdmin(uri, conf);
      KeyProvider keyP = hadmin.getKeyProvider();
```

Thanks for the update! In this case, we may not have to separate assignments from declarations.

shims/common/src/main/java/org/apache/hadoop/util/SuppressFBWarnings.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.util;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.CLASS)
+public @interface SuppressFBWarnings {
```
okumin, line 25: Can we remove a copy from shims/0.23?
https://github.com/apache/hive/blob/master/shims/0.23/src/main/java/org/apache/hadoop/util/SuppressFBWarnings.java

I assume shims/0.23 inherits shims/common here. If my assumption is wrong, we can keep duplicates at this point.

shims/0.23/src/main/java/org/apache/hadoop/hive/shims/Hadoop23Shims.java
ramitg254: I didn't see that, I'll update

ramitg254: done updated 👍

shims/common/src/main/java/org/apache/hadoop/util/SuppressFBWarnings.java
ramitg254: yeah it can be removed , done 👍

okumin (approved):
Approved. I'm waiting for CI + 24 hours

okumin:
@ramitg254 Thanks for resolving the bunch of new violations!

PR #5780, opened 2025-04-21
HIVE-28916: Fix unbalannced calls in ObjectStore rollback transaction

wecharyu:
Fix the bug in `ObjectStore#openTransaction` and `ObjectStore#rollbackTransaction`.

Currently `rollbackTransaction()` always reset `openTransactionCalls` to 0, but it would cause unbalanced issue in nested transaction calls, for example:
```bash
openTransaction()
  // new function1
  {
    openTransaction()
    commitTransaction() # place1
  }

  // new function2
  {
    openTransaction()
    commitTransaction()
  }
commitTransaction() # place2
```
If the new function1 rollbackTransaction in place1, then it would get unbalanced exception in place2.

No.

Add a unit test:
```bash
mvn test -Dtest.groups= -Dtest=org.apache.hadoop.hive.metastore.TestObjectStore#testNestedTransaction -pl :hive-standalone-metastore-server
```

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -914,23 +914,24 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
         throw new InvalidObjectException("You must specify a path for the catalog");
       }
 
-      RawStore ms = getMS();
       Path catPath = new Path(catalog.getLocationUri());
       boolean madeDir = false;
       Map<String, String> transactionalListenersResponses = Collections.emptyMap();
-      try {
-        firePreEvent(new PreCreateCatalogEvent(this, catalog));
-        if (!wh.isDir(catPath)) {
-          if (!wh.mkdirs(catPath)) {
-            throw new MetaException("Unable to create catalog path " + catPath +
-                ", failed to create catalog " + catalog.getName());
-          }
-          madeDir = true;
+      firePreEvent(new PreCreateCatalogEvent(this, catalog));
+      if (!wh.isDir(catPath)) {
+        if (!wh.mkdirs(catPath)) {
+          throw new MetaException("Unable to create catalog path " + catPath +
+              ", failed to create catalog " + catalog.getName());
         }
+        madeDir = true;
+      }
+
+      RawStore ms = getMS();
+      try {
+        ms.openTransaction();
         // set the create time of catalog
         long time = System.currentTimeMillis() / 1000;
         catalog.setCreateTime((int) time);
-        ms.openTransaction();
         ms.createCatalog(catalog);
 
         // Create a default database inside the catalog
@@ -964,7 +965,6 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
               transactionalListenersResponses, ms);
         }
       }
-      success = true;
     } catch (AlreadyExistsException|InvalidObjectException|MetaException e) {
       ex = e;
       throw e;
@@ -983,12 +983,11 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     GetCatalogResponse oldCat = null;
 
     try {
+      ms.openTransaction();
       oldCat = get_catalog(new GetCatalogRequest(rqst.getName()));
       // Above should have thrown NoSuchObjectException if there is no such catalog
       assert oldCat != null && oldCat.getCatalog() != null;
       firePreEvent(new PreAlterCatalogEvent(oldCat.getCatalog(), rqst.getNewCat(), this));
-
-      ms.openTransaction();
       ms.alterCatalog(rqst.getName(), rqst.getNewCat());
 
       if (!transactionalListeners.isEmpty()) {
@@ -1929,9 +1928,8 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     boolean success = false;
     Map<String, String> transactionalListenersResponses = Collections.emptyMap();
     try {
-      firePreEvent(new PreCreateDataConnectorEvent(connector, this));
-
       ms.openTransaction();
+      firePreEvent(new PreCreateDataConnectorEvent(connector, this));
       ms.createDataConnector(connector);
 
       if (!transactionalListeners.isEmpty()) {
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -1134,8 +1134,6 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     } catch (NoSuchObjectException e) {
       if (!ifExists) {
         throw new NoSuchObjectException(e.getMessage());
-      } else {
-        ms.rollbackTransaction();
       }
     } finally {
       if (success) {
@@ -1213,6 +1211,7 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
     boolean madeExternalDir = false;
     boolean isReplicated = isDbReplicationTarget(db);
     Map<String, String> transactionalListenersResponses = Collections.emptyMap();
+    boolean openTransaction = false;
     try {
       firePreEvent(new PreCreateDatabaseEvent(db, this));
       //reinstate location uri for metastore db.
@@ -1286,7 +1285,7 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
         }
       }
 
-      ms.openTransaction();
+      openTransaction = ms.openTransaction();
       ms.createDatabase(db);
 
       if (!transactionalListeners.isEmpty()) {
@@ -1299,7 +1298,9 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
       success = ms.commitTransaction();
     } finally {
       if (!success) {
-        ms.rollbackTransaction();
+        if (openTransaction) {
+          ms.rollbackTransaction();
+        }
 
         if (db.getCatalogName() != null && !db.getCatalogName().
             equals(Warehouse.DEFAULT_CATALOG_NAME)) {
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
@@ -141,6 +141,22 @@ public class HiveAlterHandler implements AlterHandler {
       throw new InvalidOperationException(errMsg);
     }
 
+    // Switching tables between catalogs is not allowed.
+    if (!catName.equalsIgnoreCase(newt.getCatName())) {
+      throw new InvalidOperationException("Tables cannot be moved between catalogs, old catalog" +
+          catName + ", new catalog " + newt.getCatName());
+    }
+
+    boolean rename = false;
+    // check if table with the new name already exists
+    if (!newTblName.equals(name) || !newDbName.equals(dbname)) {
+      if (msdb.getTable(catName, newDbName, newTblName, null) != null) {
+        throw new InvalidOperationException("new table " + newDbName
+            + "." + newTblName + " already exists");
+      }
+      rename = true;
+    }
+
     Path srcPath = null;
     FileSystem srcFs;
     Path destPath = null;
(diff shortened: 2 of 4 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -3252,7 +3252,7 @@ public Partition getPartitionWithAuth(String catName, String dbName, String tblN
       openTransaction();
       MPartition mpart = getMPartition(catName, dbName, tblName, partVals, null);
       if (mpart == null) {
-        commitTransaction();
+        success = commitTransaction();
```
deniskuzZ, line 3255: why not just move `commitTransaction()` to finally and remove the flag completely? this is a simple select, there is nothing to rollback

wecharyu: Addressed.

dengzhhu653:
Some thoughts:
-  at `place1`, the `openTrasactionCalls` should be 2, so `commitTransaction` is just changing the `openTrasactionCalls` to 1, we don't reach the end of the transaction here.
-  if we rollback at the `place1`, doesn't it mean we get an unexpected exception here, why do we continue to execute to `place2`?

wecharyu:
> if we rollback at the place1, doesn't it mean we get an unexpected exception here, why do we continue to execute to place2?

@dengzhhu653  There are some cases where rollback transaction does not need fail the api calls, for example `create_xxx` will always check if it already exists by `get_xxx` that throw expected `NoSuchObjectException`, in this case we should continue other queries.

dengzhhu653:
> > if we rollback at the place1, doesn't it mean we get an unexpected exception here, why do we continue to execute to place2?
> 
> @dengzhhu653 There are some cases where rollback transaction does not need fail the api calls, for example `create_xxx` will always check if it already exists by `get_xxx` that throw expected `NoSuchObjectException`, in this case we should continue other queries.

Per my understanding, this shouldn't happen, imaging we rollback the transaction at place1, then the ObjectStore `openTrasactionCalls` is reset to 0, if we continue to execute, then means we have one `openTransaction` and two `commitTransaction`, the `commitTransaction` at place2 should throw the exception:
```
if (openTrasactionCalls <= 0) {
      RuntimeException e = new RuntimeException("commitTransaction was called but openTransactionCalls = "
          + openTrasactionCalls + ". This probably indicates that there are unbalanced " +
          "calls to openTransaction/commitTransaction");
      LOG.error("Unbalanced calls to open/commit Transaction", e);
      throw e;
 }
```
the Metastore client call should get this exception, but our tests work fine.

wecharyu:
> Per my understanding, this shouldn't happen, imaging we rollback the transaction at place1, then the ObjectStore 
 openTrasactionCalls is reset to 0, if we continue to execute, then means we have one openTransaction and two commitTransaction, the commitTransaction at place2 should throw the exception:
> ```java
> if (openTrasactionCalls <= 0) {
>      RuntimeException e = new RuntimeException("commitTransaction was called but openTransactionCalls = "
>          + openTrasactionCalls + ". This probably indicates that there are unbalanced " +
>          "calls to openTransaction/commitTransaction");
>      LOG.error("Unbalanced calls to open/commit Transaction", e);
>      throw e;
> }
> ```
>the Metastore client call should get this exception, but our tests work fine.

Yes it works now because current code does not put `get_xxx` and `create_xxx` into the same transaction, it would fail if we    openTransaction at the top of a create_xxx api.

Additionally, the failed tests in this PR reveal following unbalanced calls that are hidden by current implementation:
- `commitTransaction()` twice for a `openTransaction()`
- `rollbackTransaction()` even it does not call `openTransaction()`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -1299,7 +1297,9 @@ private void create_database_core(RawStore ms, final Database db)
       success = ms.commitTransaction();
     } finally {
       if (!success) {
-        ms.rollbackTransaction();
+        if (openTransaction) {
```
deniskuzZ, line 1300: why are you checking for an `openTransaction` flag here? I don't see the same check in other places. 
````
if (isActiveTransaction() && transactionStatus != TXN_STATUS.ROLLBACK) {
    currentTransaction.rollback();
}
````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java [hunk 3]
```diff
@@ -1928,9 +1928,8 @@ private void create_dataconnector_core(RawStore ms, final DataConnector connecto
     boolean success = false;
     Map<String, String> transactionalListenersResponses = Collections.emptyMap();
     try {
-      firePreEvent(new PreCreateDataConnectorEvent(connector, this));
-
       ms.openTransaction();
+      firePreEvent(new PreCreateDataConnectorEvent(connector, this));
```
deniskuzZ, line 1932: why the change?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -983,12 +983,11 @@ public void alter_catalog(AlterCatalogRequest rqst) throws TException {
     GetCatalogResponse oldCat = null;
 
     try {
+      ms.openTransaction();
```
deniskuzZ, line 986: why do we need to open txn before the `PreAlterCatalogEvent`? are those transactional as well?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
```diff
@@ -191,6 +184,10 @@ public void alterTable(RawStore msdb, Warehouse wh, String catName, String dbnam
             TableName.getQualified(catName, dbname, name) + " doesn't exist");
       }
 
+      String expectedKey = environmentContext != null && environmentContext.getProperties() != null ?
```
deniskuzZ, line 187: why move non-transactional logic under the txn boundary?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -629,13 +622,13 @@ public boolean commitTransaction() {
...
-    openTrasactionCalls--;
-    debugLog("Commit transaction: count = " + openTrasactionCalls + ", isactive "+ currentTransaction.isActive());
 
-    if ((openTrasactionCalls == 0) && currentTransaction.isActive()) {
+    debugLog("Commit transaction: count = " + openTrasactionCalls + ", isactive "+ currentTransaction.isActive());
+    if ((openTrasactionCalls == 1) && currentTransaction.isActive()) {
       transactionStatus = TXN_STATUS.COMMITED;
       currentTransaction.commit();
     }
+    openTrasactionCalls--;
```
deniskuzZ, line 631: what if commit throws RuntimeException, db txn would be closed, but openTrasactionCalls won't be 0

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -662,7 +655,7 @@ public void rollbackTransaction() {
         currentTransaction.rollback();
       }
     } finally {
-      openTrasactionCalls = 0;
+      openTrasactionCalls--;
```
deniskuzZ, line 658: afyter rollback openTrasactionCalls should be 0

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -3254,31 +3247,38 @@ protected List<Partition> getJdoResult(GetHelper<List<Partition>> ctx) throws Me
...
       throws NoSuchObjectException, MetaException, InvalidObjectException {
-    boolean success = false;
-    try {
-      openTransaction();
-      MPartition mpart = getMPartition(catName, dbName, tblName, partVals, null);
-      if (mpart == null) {
-        commitTransaction();
-        throw new NoSuchObjectException("partition values="
-            + partVals.toString());
+    return new GetHelper<Partition>(catName, dbName, tblName, false, true) {
```
deniskuzZ, line 3250: why refactor?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: In `create_database_core` it may throw exception before calling `openTransaction()`, so it may not need call `rollbackTransaction()` in finally block.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
wecharyu: ditto, `firePreEvent` may throw exception and not call `openTransaction()`.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -561,17 +561,10 @@ public void shutdown() {
   @Override
   public boolean openTransaction() {
     openTrasactionCalls++;
-    if (openTrasactionCalls == 1) {
+    if (currentTransaction == null || !currentTransaction.isActive()){
```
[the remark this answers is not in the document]
wecharyu: To open transaction if it's not opened. It would reopen a new transaction if has already called rollback.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
wecharyu: If commit throw RuntimeException, it should call rollback in the finally block of previous layer where the openTransactionCalls would be decreased.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
wecharyu: Our unit test could not pass if is 0 here, the key point seems that if we should support such usage:
```bash
openTransaction()
  // new function1
  {
    openTransaction()
    commitTransaction() # place1
  }

  // new function2
  {
    openTransaction()
    commitTransaction()
  }
commitTransaction() # place2
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
wecharyu: It may throw exception, so we need such logic:
```java
try {
  openTransaction();

  success = commitTransaction();
} finally {
  if (!success) rollbackTransaction();
}
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
deniskuzZ: won't `isActiveTransaction()` return false?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
deniskuzZ: What is the point? We open a txn, throw an exception in `firePreEvent`, and then abort txn?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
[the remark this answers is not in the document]
deniskuzZ: if the interior txn was rolled back, should we rollback everything?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
deniskuzZ: that is what we support now, isn't ? the inner transactions are somewhat similar to @Transactional(REQUIRED) annotation

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
deniskuzZ: does it even run in txn? where do we open it?

deniskuzZ:
@wecharyu, sorry for abandoning the review! @dengzhhu653 we should probably revisit this PR

PR #5782, opened 2025-04-23
HIVE-28921: Add API and implementation for retrieving database object

jjiang037:

Add API and implementation for retrieving database object

Currently, HMS only provides APIs to retrieve database names (get_databases and get_all_databases), but these don't return the full database objects with their properties. This forces clients to make additional calls to retrieve database properties when needed. The new API allows retrieving database objects directly with a single call, which is more efficient and reduces network overhead.

No

Unit test added, regression test


standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
```diff
@@ -438,6 +438,15 @@ struct Database {
   13: optional string remote_dbname
 }
 
+struct GetDatabaseObjectsRequest {
+  1: required string catalogName,
```
dengzhhu653, line 442: nit: should `catalogName` be `optional`

dengzhhu653:
+1, please address the conflicts

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
jjiang037: Resolved it in the commit: 9d8fe91023c6d7798b20c083c77e36bba629e858

dengzhhu653 (approved):
+1 pending tests

PR #5784, opened 2025-04-24
HIVE-28923: Clean up dummy RawStores in the unit test

dengzhhu653:
every time the RawStore introduces a new method, all of them needs to implement it, however this is meaningless as no where would this new method be used.

no

current tests


SourabhBadhya (approved):
LGTM +1

PR #5786, opened 2025-04-25
HIVE-28930: Implement a metastore service that expires iceberg table snapshots periodically

abstractdog:
This patch introduces a metastore task as a MetastoreTaskThread that can expire snapshots of iceberg tables periodically according to configuration: catalog name, database pattern, table pattern. The configuration was inspired by the partition management task.

Patch contents:
1. IcebergHouseKeeperService + TestIcebergHouseKeeperService unit test
2. added the new task class (ICEBERG_TABLE_SNAPSHOT_EXPIRY_SERVICE_CLASS ) to the default housekeeping threads
3. MiniHS2 changes: withHouseKeepingThreads (for manual testing)
4. changing to keepJdbcUri=true in a [call](https://github.com/apache/hive/pull/5786/files#diff-9ede0c7f5d06a7ea25a0db82dcc9892865d148f93b511ff9dfe39266400e5f8cR424), otherwise in remote metastore mode, 2 different derby databases are used, leading to exotic problems
5. Generalized TableFetcher, which is a basically a table filter builder, originally in the PartitionManagementTask, completely reused + TestTableFetcher unit test

This service could act as a convenient helper to maintain iceberg tables, which otherwise need explicit hive ql statements by the user.

No.

Unit tests added.

Manual testing is also possible, as the patch adds MiniHS2 capability and fixes to run metastore tasks in remote mode, example command:
```
mvn clean install -Dtest=StartMiniHS2Cluster -DminiHS2.clusterType=llap -DminiHS2.conf="target/testconf/llap/hive-site.xml"  -DminiHS2.run=true -DminiHS2.usePortsFromConf=true -pl itests/hive-unit -Pitests -pl itests/util -DminiHS2.clusterType=LOCALFS_ONLY -DminiHS2.isMetastoreRemote=true -DminiHS2.withHouseKeepingThreads=true
```


abstractdog:
@deniskuzZ : this is the reusable, general part of the iceberg table maintenance service (no query history bits can be found here), I would appreciate a review in the future once you have time for that

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergUtil.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class HiveIcebergUtil {
+
+  private HiveIcebergUtil() {
+  }
+
+  public static ExecutorService getDeleteExecutorService(String completeName, int numThreads) {
```
deniskuzZ, line 31: Could this be moved to `IcebergTableUtil`? Also maybe rename to `newDeleteThreadPool`

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -0,0 +1,148 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+public class TableFetcher {
```
deniskuzZ, line 35: Was it extracted as it from `PartitionManagementTask`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Deletes snapshots of an Iceberg table, using the number of threads defined by the
+   * Hive config HIVE_ICEBERG_EXPIRE_SNAPSHOT_NUMTHREADS.
+   * This is largely equivalent to the HiveIcebergStorageHandler.expireSnapshotWithDefaultParams method.
+   *
+   * @param icebergTable the iceberg Table reference
+   */
+  private void expireSnapshotsForTable(Table icebergTable) {
+    LOG.info("Expire snapshots for: {}", icebergTable);
```
deniskuzZ, line 99: does that print fully-qualified table name?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergUtil.java
```diff
@@ -0,0 +1,39 @@
...
+    AtomicInteger deleteThreadsIndex = new AtomicInteger(0);
```
deniskuzZ, line 32: what does this counter do, isn't it always 0

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -0,0 +1,131 @@
...
+    int numThreads = conf.getInt(HiveConf.ConfVars.HIVE_ICEBERG_EXPIRE_SNAPSHOT_NUMTHREADS.varname,
+        HiveConf.ConfVars.HIVE_ICEBERG_EXPIRE_SNAPSHOT_NUMTHREADS.defaultIntVal);
+
+    ExecutorService deleteExecutorService = null;
+    try {
+      if (numThreads > 0) {
+        LOG.info("Executing expire snapshots on iceberg table {} with {} threads", icebergTable.name(), numThreads);
+        deleteExecutorService = HiveIcebergUtil.getDeleteExecutorService(icebergTable.name(), numThreads);
+      }
+      if (deleteExecutorService != null) {
```
deniskuzZ, line 111: can it be null?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -0,0 +1,131 @@
...
+  @Override
+  public void run() {
+    LOG.debug("Running IcebergHouseKeeperService...");
+
+    String catalogName = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.ICEBERG_TABLE_EXPIRY_CATALOG_NAME);
+    String dbPattern = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.ICEBERG_TABLE_EXPIRY_DATABASE_PATTERN);
+    String tablePattern = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.ICEBERG_TABLE_EXPIRY_TABLE_PATTERN);
+
+    try (IMetaStoreClient msc = new HiveMetaStoreClient(conf)) {
+      // TODO: Future improvement – modify TableFetcher to return HMS Table API objects directly,
```
deniskuzZ, line 62: makes sence

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -0,0 +1,131 @@
...
+      // avoiding the need for subsequent msc.getTable calls to fetch each matched table individually
+      List<TableName> tables = getTableFetcher(msc, catalogName, dbPattern, tablePattern).getTables();
```
deniskuzZ, line 64: can we add a cache for iceberg `Table` obj  to avoid frequent requests?

deniskuzZ (commented):
In general LGTM, minor comments

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
abstractdog: exactly, I'm expecting TestPartitionManagement to test this thoroughly, or at least to a certain point that the filter conditions were preserved while refactoring

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
abstractdog: yes, in case numThreads == 0 is explicitly set, ExpireSnapshots is not supplied with an executor service, in which case if uses the default one:
```
  RemoveSnapshots(TableOperations ops) {
...
    this.deleteExecutorService = DEFAULT_DELETE_EXECUTOR_SERVICE;
...
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
abstractdog: created HIVE-28952 and changing the comment to
```
// TODO: HIVE-28952 – modify ...
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
deniskuzZ: oh yes, I see now, sorry

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergUtil.java
deniskuzZ: my bad

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
abstractdog: Maybe I could implement a simple time-based cache, where a table record is evicted once a day. This might be a good trade-off between cache efficiency and stale object risk.

I’d prefer not to make this period configurable, as it feels like too much of an implementation detail leaking into the configuration surface.

Are you okay with that?

abstractdog:
> In general LGTM, minor comments

thanks for the review so far, addressed your comments in https://github.com/apache/hive/pull/5786/commits/dd76d541867b24a2547269b602610cbca45d1802

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
deniskuzZ: sure, maybe something similar to `CacheTableHelper`
````
Caffeine.newBuilder()
  .maximumSize(1000)
  .softValues()
  .expireAfterAccess(10, TimeUnit.MINUTES)
  .build()
````

Comment (abstractdog) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java]: okay, let me apply these:
```
 .maximumSize(1000)
  .softValues()
```
regarding expiry, 1 day still makes more sense to me

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java:87]: note, above sninnet was just an example

Comment (abstractdog) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java:87]: okay, so actually I used the Caffeine cache too with expireAfterAccess, and I just added maximumSize + softValues, know I think maximumSize=1000 still makes sense (why not), but softValues is discouraged unless I know what I'm doing, and here, I cannot tell if it's needed

```
Specifies that each value (not key) stored in the cache should be wrapped in a SoftReference (by default, strong references are used). Softly-referenced objects will be garbage-collected in a globally least-recently-used manner, in response to memory demand.
Warning: in most circumstances it is better to set a per-cache maximum size instead of using soft references. You should only use this method if you are very familiar with the practical consequences of soft references.
```

so I think, this is the one that makes sense, do you agree?
```
  private final Cache<TableName, Table> tableCache = Caffeine.newBuilder()
      .maximumSize(1000)
      .expireAfterWrite(1, TimeUnit.DAYS)
      .build();
```

File: standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java [hunk 8]
```diff
@@ -0,0 +1,148 @@
...
+  // tableFilter built from the input tablePattern and custom table conditions
+  @VisibleForTesting
+  String tableFilter;
+
+  private TableFetcher(Builder builder) {
+    this.client = builder.client;
+    if ("*".equalsIgnoreCase(builder.catalogName)) {
+      LOG.warn("Invalid wildcard '*' parameter for catalogName, exact catalog name is expected instead of regexp");
+    }
+    this.catalogName = Optional.ofNullable(builder.catalogName).orElse("hive");
```
okumin, line 57: We might use `Warehouse.DEFAULT_CATALOG_NAME`, or we just throw a NPE using `this.catalogName = Objects.requireNonNull(builder.catalogName)`.

okumin (commented):
Impressive work

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -916,6 +918,22 @@ public enum ConfVars {
     HMS_HANDLER_PROXY_CLASS("metastore.hmshandler.proxy", "hive.metastore.hmshandler.proxy",
         METASTORE_RETRYING_HANDLER_CLASS,
         "The proxy class name of HMSHandler, default is RetryingHMSHandler."),
+    ICEBERG_TABLE_EXPIRY_INTERVAL("metastore.iceberg.table.expiry.interval",
+        "hive.metastore.iceberg.table.expiry.interval", 3600, TimeUnit.SECONDS,
+        "Time interval describing how often the iceberg table expiry service runs."),
+    ICEBERG_TABLE_EXPIRY_CATALOG_NAME("metastore.iceberg.table.expiry.catalog.name",
```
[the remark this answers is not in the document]
abstractdog: unfortunately, there is no catalog pattern based retrieval in the metastore client, so this needs further work, are you fine with creating a follow-up ticket to address this?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -916,6 +918,22 @@ public enum ConfVars {
...
+        "hive.metastore.iceberg.table.expiry.catalog.name", "hive",
+        "Iceberg table expiry service looks for tables under the specified catalog name"),
+    ICEBERG_TABLE_EXPIRY_DATABASE_PATTERN("metastore.iceberg.table.expiry.database.pattern",
+        "hive.metastore.iceberg.table.expiry.database.pattern", "none",
+        "Iceberg table expiry service searches for tables using the specified database pattern. " +
+            "By default, the pattern is set to 'none', which results in no matches (this is intentional" +
```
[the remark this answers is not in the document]
abstractdog: ack, as long as empty string leads to the same empty result set, I'm double-checking

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -0,0 +1,166 @@
...
+        LOG.info("Executing expire snapshots on iceberg table {} with {} threads", icebergTable.name(), numThreads);
+        deleteExecutorService = IcebergTableUtil.newDeleteThreadPool(icebergTable.name(), numThreads);
+      }
+      if (deleteExecutorService != null) {
+        expireSnapshots.executeDeleteWith(deleteExecutorService);
+      }
+      expireSnapshots.commit();
+    } finally {
+      if (deleteExecutorService != null) {
+        deleteExecutorService.shutdown();
```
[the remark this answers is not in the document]
abstractdog: ack, let me do

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
abstractdog: I like DEFAULT_CATALOG_NAME

abstractdog:
@okumin : addressed your comments, let me know if I can proceed with this patch

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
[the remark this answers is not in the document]
okumin: Yes, I am, of course.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -916,6 +918,23 @@ public enum ConfVars {
...
+    ICEBERG_TABLE_EXPIRY_CATALOG_NAME("metastore.iceberg.table.expiry.catalog.name",
+        "hive.metastore.iceberg.table.expiry.catalog.name", "hive",
+        "Iceberg table expiry service looks for tables under the specified catalog name"),
+    ICEBERG_TABLE_EXPIRY_DATABASE_PATTERN("metastore.iceberg.table.expiry.database.pattern",
+        "hive.metastore.iceberg.table.expiry.database.pattern", "",
+        "Iceberg table expiry service searches for tables using the specified database pattern. " +
+            "By default, the pattern is set to empty string, which results in no matches (this is intentional" +
+            "to avoid expensive metastore calls unless explicitly configured by the user)."),
+    ICEBERG_TABLE_EXPIRY_TABLE_PATTERN("metastore.iceberg.table.expiry.table.pattern",
+        "hive.metastore.iceberg.table.expiry.table.pattern", "none",
```
okumin: Can this be the empty string, too?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -0,0 +1,158 @@
...
+   * This is largely equivalent to the HiveIcebergStorageHandler.expireSnapshotWithDefaultParams method.
+   *
+   * @param icebergTable the iceberg Table reference
+   */
+  private void expireSnapshotsForTable(Table icebergTable) {
+    ExpireSnapshots expireSnapshots = icebergTable.expireSnapshots();
+    if (deleteExecutorService != null) {
+      expireSnapshots.executeDeleteWith(deleteExecutorService);
+    }
+    expireSnapshots.commit();
```
okumin, line 137: I would skip not only `ExpireSnapshots#executeDeleteWith` but also `icebergTable.expireSnapshots()` and `ExpireSnapshots#commit`. That's because the Iceberg library might implement some side effects in the future, especially on commit.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -0,0 +1,158 @@
...
+      List<TableName> tables = getTableFetcher(msc, catalogName, dbPattern, tablePattern).getTables();
+
+      LOG.debug("{} candidate tables found", tables.size());
+
+      for (TableName table : tables) {
+        expireSnapshotsForTable(getIcebergTable(table, msc));
+      }
+    } catch (Exception e) {
+      LOG.error("Exception while running iceberg expiry service on catalog/db/table: {}/{}/{}",
+          catalogName, dbPattern, tablePattern, e);
```
okumin: We may move the try-catch block within the for-loop because `ExpireSnapshots#commit` can fail due to a conflict.
https://github.com/apache/iceberg/blob/apache-iceberg-1.9.1/api/src/main/java/org/apache/iceberg/PendingUpdate.java#L43-L55

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
abstractdog: yeah

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
abstractdog: do you mean something like this?

```
    if (deleteExecutorService != null) {
      ExpireSnapshots expireSnapshots = icebergTable.expireSnapshots();
      expireSnapshots.executeDeleteWith(deleteExecutorService);
      expireSnapshots.commit();
    }
```
so, basically, setting hive.iceberg.expire.snapshot.numthreads=0 to completely turn this housekeeper service off?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
abstractdog: makes sense, currently, if a table fails to be committed, none of the remaining tables will be attempted, which is not the desired behavior

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
okumin: Sorry, I thought `executeDeleteWith ` would perform the side effect. Now, I understand it just configures the executor. We don't need to make any changes. My bad
https://github.com/apache/iceberg/blob/apache-iceberg-1.9.1/api/src/main/java/org/apache/iceberg/ExpireSnapshots.java#L87-L100

abstractdog:
thanks again @okumin for the comments, addressed them in https://github.com/apache/hive/pull/5786/commits/4c5686a239ff4b7936c30d5f14c3712c237f69b5

abstractdog:
> LGTM 👍

thanks @okumin !
I just realized I left a compilation issue :D getTables() can also throw exception, so I changed the outer catch to a more generic one: https://github.com/apache/hive/pull/5786/commits/18d5358f35a55d2d801dedb2722fc3c47feead66
plese, let me know if you see any problems with that, otherwise I'm assuming that your approval still holds

okumin:
@abstractdog It sounds fine to me. The build failed. Please take a look and fix it.

```
[INFO] 878 goals, 878 executed
Error:  Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.44.4:check (validate) on project hive-iceberg-handler: The following files had format violations:
Error:      src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
Error:          @@ -30,7 +30,6 @@
Error:           import·org.apache.hadoop.hive.metastore.IMetaStoreClient;
Error:           import·org.apache.hadoop.hive.metastore.MetastoreTaskThread;
Error:           import·org.apache.hadoop.hive.metastore.api.GetTableRequest;
Error:          -import·org.apache.hadoop.hive.metastore.api.MetaException;
Error:           import·org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
Error:           import·org.apache.hadoop.hive.metastore.conf.MetastoreConf;
Error:           import·org.apache.hadoop.hive.metastore.txn.NoMutex;
Error:  Run 'mvn spotless:apply' to fix these violations.
Error:  -> [Help 1]
```

abstractdog:
> @abstractdog It sounds fine to me. The build failed. Please take a look and fix it.
> 
> ```
> [INFO] 878 goals, 878 executed
> Error:  Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.44.4:check (validate) on project hive-iceberg-handler: The following files had format violations:
> Error:      src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
> Error:          @@ -30,7 +30,6 @@
> Error:           import·org.apache.hadoop.hive.metastore.IMetaStoreClient;
> Error:           import·org.apache.hadoop.hive.metastore.MetastoreTaskThread;
> Error:           import·org.apache.hadoop.hive.metastore.api.GetTableRequest;
> Error:          -import·org.apache.hadoop.hive.metastore.api.MetaException;
> Error:           import·org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
> Error:           import·org.apache.hadoop.hive.metastore.conf.MetastoreConf;
> Error:           import·org.apache.hadoop.hive.metastore.txn.NoMutex;
> Error:  Run 'mvn spotless:apply' to fix these violations.
> Error:  -> [Help 1]
> ```

eh, sorry, committing one more

deniskuzZ (approved):
+1, pending tests

PR #5789, opened 2025-04-28
HIVE-25948: Optimize Iceberg writes by directing records either to Clustered or Fanout writer

deniskuzZ:

Make Iceberg writer configurable

Write optimization

No

Cluster + iceberg_merge_files.q

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -456,7 +456,7 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
 
     for (JobContext jobContext : jobContexts) {
       JobConf conf = jobContext.getJobConf();
-      table = Optional.ofNullable(table).orElse(Catalogs.loadTable(conf, catalogProperties));
+      table = Optional.ofNullable(table).orElseGet(() -> Catalogs.loadTable(conf, catalogProperties));
       branchName = conf.get(InputFormatConfig.OUTPUT_TABLE_SNAPSHOT_REF);
       snapshotId = getSnapshotId(outputTable.table, branchName);
 
@@ -857,7 +857,6 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
           .onFailure((output, exc) -> LOG.warn("Failed to retrieve merge input file for the table {}", output, exc))
           .run(output -> {
             for (JobContext jobContext : outputs.get(output)) {
-              LOG.info("Cleaning job for jobID: {}, table: {}", jobContext.getJobID(), output);
               Table table = output.table;
               FileSystem fileSystem = new Path(table.location()).getFileSystem(jobConf);
               String jobLocation = generateJobLocation(table.location(), jobConf, jobContext.getJobID());
@@ -897,14 +896,14 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
    * @return Returns the list of file statuses of the output files in the jobContexts
    * @throws IOException Throws IOException
    */
-  public List<ContentFile> getOutputContentFiles(List<JobContext> jobContexts) throws IOException {
+  public List<ContentFile<?>> getOutputContentFiles(List<JobContext> jobContexts) throws IOException {
     Multimap<OutputTable, JobContext> outputs = collectOutputs(jobContexts);
     JobConf jobConf = jobContexts.get(0).getJobConf();
 
     ExecutorService fileExecutor = fileExecutor(jobConf);
     ExecutorService tableExecutor = tableExecutor(jobConf, outputs.keySet().size());
 
-    Collection<ContentFile> files = new ConcurrentLinkedQueue<>();
+    Collection<ContentFile<?>> files = new ConcurrentLinkedQueue<>();
     try {
       Tasks.foreach(outputs.keySet())
           .suppressFailureWhenFinished()
@@ -912,7 +911,6 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
           .onFailure((output, exc) -> LOG.warn("Failed to retrieve merge input file for the table {}", output, exc))
           .run(output -> {
             for (JobContext jobContext : outputs.get(output)) {
-              LOG.info("Cleaning job for jobID: {}, table: {}", jobContext.getJobID(), output);
               Table table = output.table;
               String jobLocation = generateJobLocation(table.location(), jobConf, jobContext.getJobID());
               // list jobLocation to get number of forCommit files
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergSerDe.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergSerDe.java
@@ -25,7 +25,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Collectors;
-import javax.annotation.Nullable;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.ql.Context;
@@ -33,7 +32,6 @@ import org.apache.hadoop.hive.ql.security.authorization.HiveCustomStorageHandler
 import org.apache.hadoop.hive.serde2.AbstractSerDe;
 import org.apache.hadoop.hive.serde2.ColumnProjectionUtils;
 import org.apache.hadoop.hive.serde2.SerDeException;
-import org.apache.hadoop.hive.serde2.SerDeStats;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.StructObjectInspector;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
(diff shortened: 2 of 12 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -921,7 +921,8 @@ public DynamicPartitionCtx createDPContext(
     List<Function<List<ExprNodeDesc>, ExprNodeDesc>> customSortExprs = Lists.newLinkedList();
     dpCtx.setCustomSortExpressions(customSortExprs);
 
-    if (table.spec().isPartitioned()) {
+    if (table.spec().isPartitioned() &&
+          hiveConf.getIntVar(ConfVars.HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD) == 1) {
```
SourabhBadhya, line 925: Whats the idea of adding this conf?

deniskuzZ: To add sort by partitions expressions only when `HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD` optimization is enabled (i.e >= 0). 
When it's disabled, we don't expect/need input to be sorted and can use FanoutWriter

deniskuzZ: The below reducer isn't required when HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD is disabled
````
Reducer 2 
            Execution mode: vectorized, llap
            Reduce Operator Tree:
              Select Operator
                expressions: VALUE._col0 (type: int), VALUE._col1 (type: string), KEY._col2 (type: string)
                outputColumnNames: _col0, _col1, _col2
                File Output Operator
                  compressed: false
                  Dp Sort State: PARTITION_SORTED
                  Statistics: Num rows: 5 Data size: 890 Basic stats: COMPLETE Column stats: COMPLETE
                  table:
                      input format: org.apache.iceberg.mr.hive.HiveIcebergInputFormat
                      output format: org.apache.iceberg.mr.hive.HiveIcebergOutputFormat
                      serde: org.apache.iceberg.mr.hive.HiveIcebergSerDe
                      name: default.ice_dst
````

Comment (SourabhBadhya) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java]: @deniskuzZ Makes sense. This is a static solution however, there could be places wherein we can allow using FanoutWriters dynamically. 
Consider these cases - 
Inserting into / overwrite a single partition / finite low number of partitions. (We can do this now)

Anyways I am completely fine with this approach. Just want to highlight a possible optimization which can be done in some scenarios. Let me know if I am wrong in this optimization and also does it make sense to do this specifically.

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java]: `HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD` supports the flexibility
1) -1 ,1 is static, used to turn on/off the optimization
2) 0, however, does the decision based on `SortedDynPartitionOptimizer#shouldDo`, however, it's not yet adopted for iceberg

Comment (SourabhBadhya) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java]: @deniskuzZ Makes sense. This is a superior approach than what I suggested.

**Comment by deniskuzZ** [comment]
HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD = 1 
````
sourcePath = /warehouse/orc_part_source/.hive-staging_hive_2025-05-14_07-15-29_561_3761875106510575150-1/-ext-10001
> ls -la /warehouse/orc_part_source/.hive-staging_hive_2025-05-14_07-15-29_561_3761875106510575150-1
-tmp.-ext-10001
````
`MoveTask.moveFile()` doesn't do anything since the path doesn't exist, however, when

HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD = -1 
````
sourcePath = /warehouse/orc_part_source/.hive-staging_hive_2025-05-14_05-57-33_012_3401175743628391916-1/-ext-10001
> ls -la /warehouse/orc_part_source/.hive-staging_hive_2025-05-14_05-57-33_012_3401175743628391916-1/-ext-10001
total 0
 -tmp.HIVE_UNION_SUBDIR_1
 -tmp.HIVE_UNION_SUBDIR_2
````
`MoveTask.moveFile()` wipes the table directory completely

@SourabhBadhya, any ideas? 

problematic query from `iceberg_merge_files.q`
````
insert into table orc_part_source select * from orc_part_source where ds = 102 union all select * from orc_part_source where ds = 103;
````

**Comment by SourabhBadhya** [comment]
@deniskuzZ , when I worked on the merge task feature, I never got to see a dependency on this config - HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD. This mainly seems to be a dependency on the directory structure which can impact merge task execution. Was merge task executed here?

**Comment by deniskuzZ** [comment]
@SourabhBadhya, according to the comments in the Qfile, that is just a data setup phase before the actual merge task.
If I replace `UNION ALL` with `UNION` - queries complete successfully. 
I am starting to suspect `HIVE-23891: UNION ALL and multiple task attempts can cause file duplication`

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/GenMapRedUtils.java [hunk 2]
```diff
@@ -1368,7 +1368,7 @@ public static void createMRWorkForMergingFiles(FileSinkOperator fsInput,
     // NOTE: we should gather stats in MR1 rather than MR2 at merge job since we don't
     // know if merge MR2 will be triggered at execution time
     MoveWork dummyMv = null;
-    if (srcMmWriteId == null) {
+    if (srcMmWriteId == null && !fsInputDesc.isIcebergTable()) {
```
deniskuzZ, line 1371: @SourabhBadhya does it look valid to you? 
What I noticed is that we create some dummy moveWork that tries to move an empty `.hive-staging../-tmp.-ext` that doesn't exist. 
In case of `UNION ALL` we create UNION subdirs under  `.hive-staging../-ext` that cause the root table folder wipe.

SourabhBadhya: Yeah, moveTask for Iceberg is handled in a different way. Its ok for now, however ideally this should be handled on the move side. Was there any mergeWork created here?

deniskuzZ: yes, dummy LoadFileWork

SourabhBadhya: Honestly, I am fine with this. It's an overkill to define a function from Storage handler or custom interface just to see if it's an Iceberg table.

deniskuzZ: `fsInputDesc` has similar methods for ACID and insert-only tables

deniskuzZ:
hi @SourabhBadhya, please approve if it looks good to you, thanks!

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -897,22 +896,21 @@ public List<FileStatus> getOutputFiles(List<JobContext> jobContexts) throws IOEx
...
-    Collection<ContentFile> files = new ConcurrentLinkedQueue<>();
+    Collection<ContentFile<?>> files = new ConcurrentLinkedQueue<>();
     try {
       Tasks.foreach(outputs.keySet())
           .suppressFailureWhenFinished()
           .executeWith(tableExecutor)
           .onFailure((output, exc) -> LOG.warn("Failed to retrieve merge input file for the table {}", output, exc))
           .run(output -> {
             for (JobContext jobContext : outputs.get(output)) {
-              LOG.info("Cleaning job for jobID: {}, table: {}", jobContext.getJobID(), output);
```
SourabhBadhya, line 915: Any reason for removing this log?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergSerDe.java
```diff
@@ -126,41 +124,41 @@ public void initialize(@Nullable Configuration configuration, Properties serDePr
...
-    // Here we ensure that SortedDynPartitionOptimizer will kick in and do the sorting.
-    // TODO: remove once we have both Fanout and ClusteredWriter available: HIVE-25948
-    HiveConf.setIntVar(configuration, HiveConf.ConfVars.HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD, 1);
-    HiveConf.setVar(configuration, HiveConf.ConfVars.DYNAMIC_PARTITIONING_MODE, "nonstrict");
+    if (!IcebergTableUtil.isFanoutEnabled(Maps.fromProperties(serDeProperties))) {
+      // ClusteredWriter requires that records are ordered by partition keys.
+      // Here we ensure that SortedDynPartitionOptimizer will kick in and do the sorting.
+      HiveConf.setIntVar(conf, HiveConf.ConfVars.HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD, 1);
+    }
+    HiveConf.setVar(conf, HiveConf.ConfVars.DYNAMIC_PARTITIONING_MODE, "nonstrict");
```
SourabhBadhya, line 148: Why add this config?

deniskuzZ: user could change it + it was there in original version

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
deniskuzZ: yes, it's misleading, there is no cleanup here

SourabhBadhya (approved):
LGTM +1

PR #5790, opened 2025-04-28
HIVE-28655: Implement HMS Related Drop Stats Changes Part2 (param COLUMN_STAT_ACCURATE related changes)

DanielZhu58:


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7431,6 +7431,13 @@ public boolean delete_column_statistics_req(DeleteColumnStatisticsRequest req) t
       if (!isPartitioned || req.isTableLevel()) {
         ret = rawStore.deleteTableColumnStatistics(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, colNames, engine);
         if (ret) {
+          Map<String, String> parameters = table.getParameters();
```
dengzhhu653, line 7434: can we just remove the deleted `columns` from the COLUMN_STATS_ACCURATE? using the `StatsSetupConst.removeColumnStatsState`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7460,6 +7467,14 @@ public boolean delete_column_statistics_req(DeleteColumnStatisticsRequest req) t
...
             for (String partName : partNames) {
               List<String> partVals = getPartValsFromName(table, partName);
+              Partition partition = rawStore.getPartition(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, partVals);
+              Map<String, String> partParams = partition.getParameters();
+              if (partParams == null) {
+                partParams = new HashMap<>();
+              }
+              partParams.put("COLUMN_STATS_ACCURATE", "false");
+              partition.setParameters(partParams);
+              rawStore.alterPartition(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, partVals, partition, null);
```
dengzhhu653, line 7477: alter the partitions in bulk?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10214,6 +10214,8 @@ public boolean deletePartitionColumnStatistics(String catName, String dbName, St
     }
     dbName = org.apache.commons.lang3.StringUtils.defaultString(dbName, Warehouse.DEFAULT_DATABASE_NAME);
     catName = normalizeIdentifier(catName);
+    long tableID = directSql.getTableId(dbName, tableName);
+    directSql.deleteColumnStatsState(tableID);
```
wecharyu, line 10218: what if direct sql is not allowed? can we move it to `getSqlResult()`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7460,6 +7467,8 @@ public boolean delete_column_statistics_req(DeleteColumnStatisticsRequest req) t
               .collect(Collectors.toList()) : colNames) {
             for (String partName : partNames) {
               List<String> partVals = getPartValsFromName(table, partName);
+              Partition partition = rawStore.getPartition(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, partVals);
+              Map<String, String> partParams = partition.getParameters();
```
wecharyu, line 7471: The `partParams` seems no used.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7440,6 +7440,13 @@ public boolean delete_column_statistics_req(DeleteColumnStatisticsRequest req) t
             }
             events.add(new DeleteTableColumnStatEvent(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, colName, engine, this));
           }
+          if (colNames == null){
+            // remove all column names in parameter COLUMN_STATS_ACCURATE
+            StatsSetupConst.clearColumnStatsState(table.getParameters());
+          } else {
+            // remove the deleted column names in parameter COLUMN_STATS_ACCURATE
+            StatsSetupConst.removeColumnStatsState(table.getParameters(), colNames);
+          }
```
dengzhhu653, line 7449: Need to persist the changed parameters by `rawStore.alterTable()`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7460,13 +7467,22 @@ public boolean delete_column_statistics_req(DeleteColumnStatisticsRequest req) t
...
               }
               events.add(new DeletePartitionColumnStatEvent(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName,
                   partName, partVals, colName, engine, this));
+              if (colNames == null) {
+                // remove all column names in parameter COLUMN_STATS_ACCURATE
+                StatsSetupConst.clearColumnStatsState(partParams);
+              } else {
+                // remove the deleted column names in parameter COLUMN_STATS_ACCURATE
+                StatsSetupConst.removeColumnStatsState(partParams, colNames);
+              }
```
dengzhhu653, line 7485: need to persist the partition parameters after resetting the column stats

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3235,13 +3274,34 @@ public void deleteColumnStatsState(long tbl_id) throws MetaException {
...
     }
   }
 
+  public void deleteTblColumnStatsState(long tbl_id) throws MetaException {
+    String queryText;
+    switch (dbType.dbType) {
+      // delete the table parameter COLUMN_STATS_ACCURATE
+      default:
+        // @formatter:off
+        queryText = ""
```
dengzhhu653, line 3291: if this is the only query for all db types, why not remove the `switch`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3235,13 +3274,34 @@ public void deleteColumnStatsState(long tbl_id) throws MetaException {
         // @formatter:on
     }
 
+    executeWithArray(query.getInnerQuery(), null, queryText);
```
dengzhhu653, line 3277: the queryText would be executed twice, can we remove this?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3211,6 +3211,45 @@ public void lockDbTable(String tableName) throws MetaException {
     }
   }
 
+  public Long getTableId(String dbName, String tableName) throws MetaException {
```
dengzhhu653, line 3214: A simple `MTable.getId` call can serve the same purpose I think

dengzhhu653:
any ideas? cc @saihemanth-cloudera @soumyakanti3578

ql/src/test/results/clientpositive/llap/alter_table_column_stats.q.out
```diff
@@ -1105,7 +1105,6 @@ Retention:          	0
 #### A masked pattern was here ####
 Table Type:         	MANAGED_TABLE       	 
 Table Parameters:	 	 
-	COLUMN_STATS_ACCURATE	{\"BASIC_STATS\":\"true\"}
```
soumyakanti3578: I see that we are removing `COLUMN_STATS_ACCURATE` here, however, in the corresponding q file, I see this comment:
```
-- rename a partition should not change its table, partition, and column stats
alter table statsdb1.testpart1 partition (part = 'part1') rename to partition (part = 'part11');
describe formatted statsdb1.testpart1;
```
I just want us to be sure that this is what we intend to do as I see several other similar changes just below which were due to renaming/replacing columns. And if the tests don't make sense any more maybe we should consider updating the tests.

ql/src/test/results/clientpositive/llap/alter_table_column_stats.q.out
```diff
@@ -1476,7 +1471,7 @@ Database:           	statsdb1
 Table:              	testpart1           	 
 #### A masked pattern was here ####
 Partition Parameters:	 	 
-	COLUMN_STATS_ACCURATE	{\"BASIC_STATS\":\"true\",\"COLUMN_STATS\":{\"col1\":\"true\",\"col2\":\"true\"}}
+	COLUMN_STATS_ACCURATE	{}
```
soumyakanti3578: Here we are printing empty `COLUMN_STATS_ACCURATE`. If it is easy to remove them, maybe we should do that?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/common/StatsSetupConst.java
```diff
@@ -374,8 +374,11 @@ public static void removeColumnStatsState(Map<String, String> params, List<Strin
       return;
     }
     ColumnStatsAccurate stats = parseStatsAcc(params.get(COLUMN_STATS_ACCURATE));
-    colNames.forEach(colName ->
-        stats.columnStats.remove(colName.toLowerCase()));
+    colNames.forEach(colName -> {
+      if (colName != null) {
```
saihemanth-cloudera, line 378: Can a column name be null?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -7502,6 +7502,15 @@ public boolean delete_column_statistics_req(DeleteColumnStatisticsRequest req) t
...
       }
+      // on the table level, partially delete(update) table level parameter COLUMN_STATS_ACCURATE
+      if (colNames == null || colNames.isEmpty()){
+        // remove all column names in parameter COLUMN_STATS_ACCURATE
+        StatsSetupConst.clearColumnStatsState(table.getParameters());
+      } else {
+        // remove the deleted column names in parameter COLUMN_STATS_ACCURATE
+        StatsSetupConst.removeColumnStatsState(table.getParameters(), colNames);
+      }
+      rawStore.alterTable(parsedDbName[CAT_NAME], parsedDbName[DB_NAME], tableName, table, null);
```
saihemanth-cloudera: I don't like the idea of doing an alter operation just to update 'COLUMN_STATS_ACCURATE' field.

dengzhhu653: we might need to add a new interface in RawStore, example `void updateTableParam(long tableId, String paramKey, String newParamValue)` to update the table's parameters, then in `MetaStoreDirectSql`, use `update TABLE_PARAMS set PARAM_VALUE = new_json_string where TBL_ID = table_id and PARAM_KEY='COLUMN_STATS_ACCURATE'` to update the column status

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3288,6 +3291,190 @@ public List<Void> run(List<String> input) throws Exception {
     return true;
   }
 
+  // a helper function which will firstly get the current COLUMN_STATS_ACCURATE parameter on table level
+  // secondly convert the JSON String into map, and update the information in it, and convert it back to JSON
+  // thirdly update the COLUMN_STATS_ACCURATE parameter with the new value on table level using directSql
+  public long updateColumnStatsAccurateForTable(Table table, List<String> droppedCols) throws MetaException {
+    String currentValue = table.getParameters().get("COLUMN_STATS_ACCURATE");
+    if (currentValue == null) return 0;
```
soumyakanti3578, line 3299: Consider adding braces for better maintainability and consistency.

```suggestion
    if (currentValue == null) {
        return 0;
    }
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3288,6 +3291,190 @@ public List<Void> run(List<String> input) throws Exception {
     return true;
   }
 
+  // a helper function which will firstly get the current COLUMN_STATS_ACCURATE parameter on table level
+  // secondly convert the JSON String into map, and update the information in it, and convert it back to JSON
+  // thirdly update the COLUMN_STATS_ACCURATE parameter with the new value on table level using directSql
+  public long updateColumnStatsAccurateForTable(Table table, List<String> droppedCols) throws MetaException {
```
soumyakanti3578, line 3297: nit: I don't see the `long` return value getting used anywhere. I see that it's required because of the method `updateTableParam`, but could this be handled in a better way?

If you want to document what the method does, please use the correct javadoc format `/** ... */`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3288,6 +3291,190 @@ public List<Void> run(List<String> input) throws Exception {
...
+          statsMap.remove("COLUMN_STATS");
+        } else {
+          // Remove only the dropped columns
+          for (String col : droppedCols) {
+            columnStats.remove(col.toLowerCase());
+          }
+          if (columnStats.isEmpty()) {
+            statsMap.remove("COLUMN_STATS");
+          }
+        }
```
soumyakanti3578, line 3326: I think you are trying to remove all column stats when either `droppedCols` is `null`, or when it's size is zero or equal to the size of the map `columnStats`, right?

In that case update `removeAll` to
```
boolean removeAll = droppedCols == null || droppedCols.isEmpty() || droppedCols.size() == columnStats.size();
```
You can then remove the block at the end:
```
          if (columnStats.isEmpty()) {
            statsMap.remove("COLUMN_STATS");
          }
```
This assumes that `List<String> droppedCols` has unique columns.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3288,6 +3291,190 @@ public List<Void> run(List<String> input) throws Exception {
...
+        effectiveColNames = new ArrayList<>();
+      } else {
+        effectiveColNames = table.getSd().getCols().stream()
+                .map(f -> f.getName().toLowerCase())
+                .collect(Collectors.toList());
+      }
+    } else {
+      effectiveColNames = colNames.stream().map(String::toLowerCase).collect(Collectors.toList());
+    }
+    List<String> finalColNames = effectiveColNames;
```
soumyakanti3578, line 3362: We don't need `finalColNames` here as we are not updating `effectiveColNames` after this point.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3288,6 +3291,190 @@ public List<Void> run(List<String> input) throws Exception {
...
+              Map<String, Object> statsMap = mapper.readValue(
+                      currentValue, new TypeReference<Map<String, Object>>() {});
+              Object columnStatsObj = statsMap.get("COLUMN_STATS");
+
+              boolean changed = false;
+              if (columnStatsObj instanceof Map) {
+                Map<String, String> columnStats = (Map<String, String>) columnStatsObj;
+                for (String col : finalColNames) {
+                  if (columnStats.remove(col) != null) {
+                    changed = true;
```
soumyakanti3578, line 3395: `break` early when `changed = true;`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3288,6 +3291,190 @@ public List<Void> run(List<String> input) throws Exception {
...
+                  }
+                }
+
+                if (columnStats.isEmpty()) {
+                  statsMap.remove("COLUMN_STATS");
+                  changed = true;
+                }
```
soumyakanti3578, line 3402: It would be better to check this before running the `for` loop.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3288,6 +3291,190 @@ public List<Void> run(List<String> input) throws Exception {
...
+              }
+
+              if (!statsMap.containsKey("COLUMN_STATS")) {
+                if (statsMap.remove("BASIC_STATS") != null) {
+                  changed = true;
+                }
+              }
```
soumyakanti3578, line 3409: Can this be checked right after initializing `statsMap`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -3288,6 +3291,190 @@ public List<Void> run(List<String> input) throws Exception {
...
+
+
+
+
+
+
+
+
+
+
```
soumyakanti3578, line 3477: Please remove extra space.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -10280,7 +10280,11 @@ protected String describeResult() {
       }
       @Override
       protected Boolean getSqlResult(GetHelper<Boolean> ctx) throws MetaException {
-        return directSql.deletePartitionColumnStats(catName, dbName, tableName, partNames, colNames, engine);
+        if (directSql.deletePartitionColumnStats(catName, dbName, tableName, partNames, colNames, engine)){
+          directSql.updateColumnStatsAccurateForPartitions(catName, dbName, getTable(), partNames, colNames);
+          return true;
```
soumyakanti3578, line 10285: Can we return the method call directly here? Or do we want to return `true` even when `updateColumnStatsAccurateForPartitions` returns `false`?

soumyakanti3578 (changes requested):

PR #5791, opened 2025-04-29
Fix flaky test in TestLazyBinaryColumnarSerDe by using LinkedHashMap

seltepu:
…nerNulls()

This pull request fixes a flaky test in TestLazyBinaryColumnarSerDe.java. The test was using a HashMap to store values, which does not guarantee iteration order. This nondeterministic order caused the test to sometimes fail when the output order changed. The fix replaces HashMap with LinkedHashMap to ensure a consistent iteration order and deterministic test results.

The test fails nondeterministically because HashMap does not guarantee iteration order. This means the serialized output can vary between runs, leading to intermittent assertion failures. Using LinkedHashMap ensures the order is preserved, making the test reliable.

No.

The test was run multiple times using the NonDex plugin to randomize internal order.

Before the fix, the test failed intermittently, as shown in the attached failing-test.log. After the fix, the test passed consistently across multiple runs, as shown in the attached passing-test.log.

Reproduction steps:

1. Run the test with NonDex:

mvn edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.apache.hadoop.hive.serde2.columnar.TestLazyBinaryColumnarSerDe#testSerDeInnerNulls -DnondexRuns=10

2. Observe that before the fix, the test fails intermittently but after the fix, it always passes.

### DIFF
--- a/serde/src/test/org/apache/hadoop/hive/serde2/columnar/TestLazyBinaryColumnarSerDe.java
+++ b/serde/src/test/org/apache/hadoop/hive/serde2/columnar/TestLazyBinaryColumnarSerDe.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.TreeMap;
+import java.util.LinkedHashMap;
 
 import org.junit.Assert;
 import junit.framework.TestCase;
@@ -165,7 +166,7 @@ public class TestLazyBinaryColumnarSerDe {
     outerStruct.mByte  = 101;
     outerStruct.mShort = 2002;
     outerStruct.mInt = 3003;
-    outerStruct.mLong = 4004l;
+    outerStruct.mLong = ddd;
     outerStruct.mFloat = 5005.01f;
     outerStruct.mDouble = 6006.001d;
     outerStruct.mString = "";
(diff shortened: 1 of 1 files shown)


serde/src/test/org/apache/hadoop/hive/serde2/columnar/TestLazyBinaryColumnarSerDe.java
```diff
@@ -165,7 +166,7 @@ public void testLazyBinaryColumnarSerDeWithEmptyBinary() throws SerDeException {
     outerStruct.mByte  = 101;
     outerStruct.mShort = 2002;
     outerStruct.mInt = 3003;
-    outerStruct.mLong = 4004l;
+    outerStruct.mLong = 40041;
```
okumin, line 169: Probably, the original value is intended. Or we can use an uppercase.

```suggestion
    outerStruct.mLong = 4004L;
```

serde/src/test/org/apache/hadoop/hive/serde2/columnar/TestLazyBinaryColumnarSerDe.java
```diff
@@ -234,7 +235,7 @@ public void testSerDeInnerNulls() throws SerDeException {
     outerStruct.mArray = new ArrayList<InnerStruct>(2);
     outerStruct.mArray.add(is1);
     outerStruct.mArray.add(is2);
-    outerStruct.mMap = new HashMap<String, InnerStruct>();
+    outerStruct.mMap = new LinkedHashMap<String, InnerStruct>(); //fixed line
```
okumin, line 238: I believe we don't need the additional comment.

PR #5792, opened 2025-04-30
HIVE-28935: Iceberg: fix partition filtering condition in compaction query

difin:

Using partition spec id and partition values for condition on partitions in compaction query instead of partition hash 

Partition hash does not guarantee uniqueness. it is not meant to serve as a unique key, but rather a mean for consistent distribution.

No

New and existing q-tests.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergFilterFactory.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergFilterFactory.java
@@ -228,6 +228,32 @@ public class HiveIcebergFilterFactory {
     }
   }
 
+  public static Object convertPartitionLiteral(Object literal, TransformSpec transform) {
+    if (transform == null) {
+      return literal;
+    }
+    try {
+      switch (transform.getTransformType()) {
+        case YEAR:
+          return parseYearToTransformYear(literal.toString());
+        case MONTH:
+          return parseMonthToTransformMonth(literal.toString());
+        case HOUR:
+          return parseHourToTransformHour(literal.toString());
+        case DAY:
+        case TRUNCATE:
+        case BUCKET:
+        case IDENTITY:
+          return literal;
+        default:
+          throw new UnsupportedOperationException("Unknown transform: " + transform.getTransformType());
+      }
+    } catch (NumberFormatException | DateTimeException | IllegalStateException e) {
+      throw new RuntimeException(
+          String.format("Unable to parse value '%s' as '%s' transform value", literal.toString(), transform));
+    }
+  }
+
   private static final int ICEBERG_EPOCH_YEAR = 1970;
   private static final int ICEBERG_EPOCH_MONTH = 1;
 
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
@@ -44,6 +44,7 @@ import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
 import org.apache.hadoop.hive.ql.Context;
 import org.apache.hadoop.hive.ql.QueryState;
 import org.apache.hadoop.hive.ql.io.sarg.SearchArgument;
+import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.metadata.HiveUtils;
 import org.apache.hadoop.hive.ql.parse.AlterTableExecuteSpec;
 import org.apache.hadoop.hive.ql.parse.SemanticException;
@@ -524,9 +525,10 @@ public class IcebergTableUtil {
   }
 
 
-  public static PartitionSpec getPartitionSpec(Table icebergTable, String partitionPath) throws MetaException {
+  public static PartitionSpec getPartitionSpec(Table icebergTable, String partitionPath)
+      throws MetaException, HiveException {
     if (icebergTable == null || partitionPath == null || partitionPath.isEmpty()) {
-      throw new IllegalArgumentException("Table and partitionPath must not be null or empty.");
+      throw new HiveException("Table and partitionPath must not be null or empty.");
     }
 
     // Extract field names from the path: "field1=val1/field2=val2" → [field1, field2]
@@ -540,7 +542,7 @@ public class IcebergTableUtil {
           return specFieldNames.equals(fieldNames);
         })
         .findFirst() // Supposed to be only one matching spec
-        .orElse(null);
+        .orElseThrow(() -> new HiveException("No matching partition spec found for partition path: " + partitionPath));
   }
 
   public static TransformSpec getTransformSpec(Table table, String transformName, int sourceId) {
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
@@ -21,7 +21,6 @@ package org.apache.iceberg.mr.hive;
 
 import java.io.IOException;
 import java.time.ZoneId;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -37,8 +36,10 @@ import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.common.type.TimestampTZ;
 import org.apache.hadoop.hive.common.type.TimestampTZUtil;
 import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.ColumnStatisticsObj;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
+import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
 import org.apache.hadoop.hive.ql.Context;
 import org.apache.hadoop.hive.ql.QueryState;
@@ -79,6 +80,7 @@ import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.mr.Catalogs;
 import org.apache.iceberg.mr.InputFormatConfig;
 import org.apache.iceberg.relocated.com.google.common.collect.FluentIterable;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.types.Conversions;
 import org.apache.iceberg.types.Type;
@@ -522,15 +524,13 @@ public class IcebergTableUtil {
   }
 
 
-  public static Integer getPartitionSpecId(Table icebergTable, String partitionPath) {
+  public static PartitionSpec getPartitionSpec(Table icebergTable, String partitionPath) throws MetaException {
     if (icebergTable == null || partitionPath == null || partitionPath.isEmpty()) {
       throw new IllegalArgumentException("Table and partitionPath must not be null or empty.");
     }
 
     // Extract field names from the path: "field1=val1/field2=val2" → [field1, field2]
-    List<String> fieldNames = Arrays.stream(partitionPath.split("/"))
-        .map(s -> s.split("=")[0])
-        .collect(Collectors.toList());
+    List<String> fieldNames = Lists.newArrayList(Warehouse.makeSpecFromName(partitionPath).keySet());
 
     return icebergTable.specs().values().stream()
         .filter(spec -> {
@@ -539,7 +539,6 @@ public class IcebergTableUtil {
               .collect(Collectors.toList());
           return specFieldNames.equals(fieldNames);
         })
-        .map(PartitionSpec::specId)
         .findFirst() // Supposed to be only one matching spec
         .orElse(null);
   }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
@@ -21,6 +21,7 @@ package org.apache.iceberg.mr.hive;
 
 import java.io.IOException;
 import java.time.ZoneId;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -520,25 +521,27 @@ public class IcebergTableUtil {
     }
   }
 
-  public static long getPartitionHash(Table icebergTable, String partitionPath) throws IOException {
-    PartitionsTable partitionsTable = (PartitionsTable) MetadataTableUtils
-        .createMetadataTableInstance(icebergTable, MetadataTableType.PARTITIONS);
 
-    try (CloseableIterable<FileScanTask> fileScanTasks = partitionsTable.newScan().planFiles()) {
-      return FluentIterable.from(fileScanTasks)
-          .transformAndConcat(task -> task.asDataTask().rows())
-          .transform(row -> {
-            StructProjection data = row.get(IcebergTableUtil.PART_IDX, StructProjection.class);
-            PartitionSpec spec = icebergTable.specs().get(row.get(IcebergTableUtil.SPEC_IDX, Integer.class));
-            PartitionData partitionData = IcebergTableUtil.toPartitionData(data,
-                Partitioning.partitionType(icebergTable), spec.partitionType());
-            String path = spec.partitionToPath(partitionData);
-            return Maps.immutableEntry(path, data);
-          })
-          .filter(e -> e.getKey().equals(partitionPath))
-          .transform(e -> IcebergAcidUtil.computeHash(e.getValue()))
-          .get(0);
+  public static Integer getPartitionSpecId(Table icebergTable, String partitionPath) {
+    if (icebergTable == null || partitionPath == null || partitionPath.isEmpty()) {
+      throw new IllegalArgumentException("Table and partitionPath must not be null or empty.");
     }
+
+    // Extract field names from the path: "field1=val1/field2=val2" → [field1, field2]
+    List<String> fieldNames = Arrays.stream(partitionPath.split("/"))
+        .map(s -> s.split("=")[0])
+        .collect(Collectors.toList());
+
+    return icebergTable.specs().values().stream()
+        .filter(spec -> {
+          List<String> specFieldNames = spec.fields().stream()
+              .map(PartitionField::name)
+              .collect(Collectors.toList());
+          return specFieldNames.equals(fieldNames);
+        })
+        .map(PartitionSpec::specId)
+        .findFirst() // Supposed to be only one matching spec
+        .orElse(null);
   }
 
   public static TransformSpec getTransformSpec(Table table, String transformName, int sourceId) {
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
@@ -30,6 +30,7 @@ import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
 import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.CompactionType;
+import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.txn.entities.CompactionInfo;
 import org.apache.hadoop.hive.ql.Context.RewritePolicy;
 import org.apache.hadoop.hive.ql.DriverUtils;
@@ -46,6 +47,7 @@ import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
 import org.apache.hive.iceberg.org.apache.orc.storage.common.TableName;
 import org.apache.iceberg.PartitionField;
+import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Partitioning;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.hive.HiveSchemaUtil;
@@ -132,14 +134,18 @@ public class IcebergQueryCompactor extends QueryCompactor  {
       HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.PARTITION.name());
       conf.set(IcebergCompactionService.PARTITION_PATH, new Path(ci.partName).toString());
 
-      int specId = IcebergTableUtil.getPartitionSpecId(icebergTable, ci.partName);
+      PartitionSpec spec;
+      try {
+        spec = IcebergTableUtil.getPartitionSpec(icebergTable, ci.partName);
+      } catch (MetaException e) {
+        throw new HiveException(e);
+      }
       String partitionPredicate = buildPartitionPredicate(ci, icebergTable);
 
       compactionQuery = String.format("INSERT OVERWRITE TABLE %1$s SELECT * FROM %1$s WHERE %2$s IN " +
-          "(SELECT FILE_PATH FROM %1$s.FILES WHERE %3$s AND SPEC_ID = %7$d) AND %6$s = %7$d %4$s %5$s",
-      compactTableName, VirtualColumn.FILE_PATH.getName(), partitionPredicate,
-      fileSizePredicate == null ? "" : "AND " + fileSizePredicate, orderBy,
-      VirtualColumn.PARTITION_SPEC_ID.getName(), specId);
+          "(SELECT FILE_PATH FROM %1$s.FILES WHERE %3$s AND SPEC_ID = %4$d) %5$s %6$s",
+      compactTableName, VirtualColumn.FILE_PATH.getName(), partitionPredicate, spec.specId(),
+      fileSizePredicate == null ? "" : "AND " + fileSizePredicate, orderBy);
     }
     return compactionQuery;
   }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
@@ -20,7 +20,6 @@
 package org.apache.iceberg.mr.hive.compaction;
 
 import java.io.IOException;
-import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
@@ -134,12 +133,13 @@ public class IcebergQueryCompactor extends QueryCompactor  {
       conf.set(IcebergCompactionService.PARTITION_PATH, new Path(ci.partName).toString());
 
       PartitionSpec spec;
+      String partitionPredicate;
       try {
         spec = IcebergTableUtil.getPartitionSpec(icebergTable, ci.partName);
+        partitionPredicate = buildPartitionPredicate(ci, spec);
       } catch (MetaException e) {
         throw new HiveException(e);
       }
-      String partitionPredicate = buildPartitionPredicate(ci, spec);
 
       compactionQuery = String.format("INSERT OVERWRITE TABLE %1$s SELECT * FROM %1$s WHERE %2$s IN " +
           "(SELECT FILE_PATH FROM %1$s.FILES WHERE %3$s AND SPEC_ID = %4$d) %5$s %6$s",
@@ -149,10 +149,8 @@ public class IcebergQueryCompactor extends QueryCompactor  {
     return compactionQuery;
   }
 
-  private String buildPartitionPredicate(CompactionInfo ci, PartitionSpec spec) {
-    Map<String, String> partSpecMap = new LinkedHashMap<>();
-    Warehouse.makeSpecFromName(partSpecMap, new Path(ci.partName), null);
-
+  private String buildPartitionPredicate(CompactionInfo ci, PartitionSpec spec) throws MetaException {
+    Map<String, String> partSpecMap = Warehouse.makeSpecFromName(ci.partName);
     Map<String, PartitionField> partitionFieldMap = spec.fields().stream()
         .collect(Collectors.toMap(PartitionField::name, Function.identity()));
 
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
@@ -22,6 +22,9 @@ package org.apache.iceberg.mr.hive.compaction;
 import java.io.IOException;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
@@ -35,13 +38,21 @@ import org.apache.hadoop.hive.ql.metadata.Hive;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.metadata.HiveUtils;
 import org.apache.hadoop.hive.ql.metadata.VirtualColumn;
+import org.apache.hadoop.hive.ql.parse.TransformSpec;
 import org.apache.hadoop.hive.ql.session.SessionState;
 import org.apache.hadoop.hive.ql.txn.compactor.CompactorContext;
 import org.apache.hadoop.hive.ql.txn.compactor.QueryCompactor;
+import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
 import org.apache.hive.iceberg.org.apache.orc.storage.common.TableName;
+import org.apache.iceberg.PartitionField;
+import org.apache.iceberg.Partitioning;
 import org.apache.iceberg.Table;
+import org.apache.iceberg.hive.HiveSchemaUtil;
+import org.apache.iceberg.mr.hive.HiveIcebergFilterFactory;
 import org.apache.iceberg.mr.hive.IcebergTableUtil;
 import org.apache.iceberg.mr.hive.compaction.evaluator.CompactionEvaluator;
+import org.apache.iceberg.types.Types;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -58,13 +69,33 @@ public class IcebergQueryCompactor extends QueryCompactor  {
 
     HiveConf conf = new HiveConf(context.getConf());
     CompactionInfo ci = context.getCompactionInfo();
-    String partSpec = ci.partName;
+    String compactionQuery = buildCompactionQuery(context, compactTableName, conf);
+
+    SessionState sessionState = setupQueryCompactionSession(conf, ci, tblProperties);
+    String compactionTarget = "table " + HiveUtils.unparseIdentifier(compactTableName) +
+        (ci.partName != null ? ", partition " + HiveUtils.unparseIdentifier(ci.partName) : "");
+
+    try {
+      DriverUtils.runOnDriver(conf, sessionState, compactionQuery);
+      LOG.info("Completed compaction for {}", compactionTarget);
+      return true;
+    } catch (HiveException e) {
+      LOG.error("Failed compacting {}", compactionTarget, e);
+      throw e;
+    } finally {
+      sessionState.setCompaction(false);
+    }
+  }
+
+  private String buildCompactionQuery(CompactorContext context, String compactTableName, HiveConf conf)
+      throws HiveException {
+    CompactionInfo ci = context.getCompactionInfo();
     org.apache.hadoop.hive.ql.metadata.Table table = Hive.get(conf).getTable(context.getTable().getDbName(),
         context.getTable().getTableName());
     Table icebergTable = IcebergTableUtil.getTable(conf, table.getTTable());
-    String compactionQuery;
     String orderBy = ci.orderByClause == null ? "" : ci.orderByClause;
     String fileSizePredicate = null;
+    String compactionQuery;
 
     if (ci.type == CompactionType.MINOR) {
       long fileSizeInBytesThreshold = CompactionEvaluator.getFragmentSizeBytes(table.getParameters());
@@ -76,7 +107,7 @@ public class IcebergQueryCompactor extends QueryCompactor  {
       conf.setBoolVar(ConfVars.HIVE_VECTORIZATION_ENABLED, false);
     }
 
-    if (partSpec == null) {
+    if (ci.partName == null) {
       if (!icebergTable.spec().isPartitioned()) {
         HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.FULL_TABLE.name());
         compactionQuery = String.format("insert overwrite table %s select * from %<s %2$s %3$s", compactTableName,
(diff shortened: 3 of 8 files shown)


difin:
@okumin @deniskuzZ 

This works except for one case and I’m not sure how we can do it.

The case where it doesn't work:

Let’s say you have a table with partition evolution and the partition specs are:
```
Spec_Id, Partition Schema
0, {field_a: String}
1,. {field_a: String, b: String}
```

And let’s say there are 2 partitions on disk:
```
{a=NULL}
{a=NULL, b=NULL}

```

When calculating partition hash, we pass Iceberg partition struct, which contain fields from all partition specs, so we send these inputs to computeHash methods for the 2 different partitions:
```
{a=NULL, b=NULL}
{a=NULL, b=NULL}
```
and the partition hash becomes the same for both partitions..

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
```diff
@@ -243,14 +247,49 @@ public static long getDeleteFilePosition(Record rec) {
...
+        // Fallback to object's string representation
+        hasher.putLong(Objects.hash(val));
+      }
+    }
+
+    HashCode hashCode = hasher.hash();
+    return hashCode.asLong();
+  }
+
   public static long computeHash(StructLike struct) {
```
deniskuzZ, line 285: why don't we use Iceberg's 
````
Types.StructType partitionType = Partitioning.partitionType(table);
JavaHash.forType(partitionType).hash(struct)
````

Comment (difin) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java:285]: I re-implemented with Iceberg's code as you suggested, and tested.
The q-test results with bucketing and partition evolution with nulls were even worse than the original approach.
It leads to collisions with 0 and null.

I pushed this code version to a separate branch: https://github.com/difin/hive/commit/a03bf02eeebdee7e2a16400603b85fc8ed12c6fd

Here is a portion from the q.out file with duplicates:

```
PREHOOK: query: SELECT * FROM default.srcbucket_big ORDER BY id
 PREHOOK: type: QUERY
 PREHOOK: Input: default@srcbucket_big
 @@ -316,10 +316,52 @@ NULL	val_102	2
 105	val_105	5
 NULL	NULL	6
 101	val_101	7
 101	val_101	7
 101	val_101	7
 101	val_101	7
 101	val_101	7
 101	val_101	7
  ...
```

Iceberg hashing function uses the same `Objects::hashCode` method which was I'm attempting to get rid of in this PR because it leads to collisions 

```
  static <T> JavaHash<T> forType(Type type) {
    switch (type.typeId()) {
      case STRING:
        return JavaHashes.strings();
      case STRUCT:
        return JavaHashes.struct(type.asStructType());
      case LIST:
        return JavaHashes.list(type.asListType());
      default:
        return Objects::hashCode;
    }
  }
```

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java [hunk 2]
```diff
@@ -243,14 +247,49 @@ public static long getDeleteFilePosition(Record rec) {
     return rec.get(DELETE_FILE_META_COLS.get(MetadataColumns.ROW_POSITION), Long.class);
   }
 
+  private static long hashObjectArray(Object[] values) {
+    Hasher hasher = Hashing.murmur3_128().newHasher();
+
+    for (Object val : values) {
+      if (val == null) {
+        // Unique constant for null
+        hasher.putInt(0xDEADBEEF);
```
SourabhBadhya, line 256: `0xDEADBEEF`
What is this value?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
```diff
@@ -243,14 +247,49 @@ public static long getDeleteFilePosition(Record rec) {
     return rec.get(DELETE_FILE_META_COLS.get(MetadataColumns.ROW_POSITION), Long.class);
   }
 
+  private static long hashObjectArray(Object[] values) {
```
SourabhBadhya, line 250: AFAIK, partition hash is not written anywhere, hence I am assuming this will not have backward compatible issues.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
```diff
@@ -243,14 +247,49 @@ public static long getDeleteFilePosition(Record rec) {
...
+        hasher.putByte((Byte) val);
+      } else if (val instanceof Character) {
+        hasher.putChar((Character) val);
+      } else if (val instanceof Double) {
+        hasher.putDouble((Double) val);
+      } else if (val instanceof Float) {
+        hasher.putFloat((Float) val);
+      } else {
+        // Fallback to object's string representation
+        hasher.putLong(Objects.hash(val));
```
SourabhBadhya, line 277: Consider adding tests for this case since we are using long type for these cases.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
difin: > The number 0xDEADBEEF is a hexadecimal literal often used in programming as a sentinel, debug marker, or placeholder value. It is interpreted as -559038737 in Java

This is a unique constant for null to avoid 0 and null contribute the same hashing value.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
difin: I updated my fix with a smaller change that also works and this change is no longer part of the fix.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
```diff
@@ -206,10 +207,11 @@ public static int parseSpecId(Record rec) {
     return rec.get(FILE_READ_META_COLS.get(MetadataColumns.SPEC_ID), Integer.class);
   }
 
-  public static long computePartitionHash(Record rec) {
+  public static long computePartitionHash(Table table, Record rec) {
     StructProjection part = rec.get(FILE_READ_META_COLS.get(PARTITION_STRUCT_META_COL), StructProjection.class);
+    Types.StructType partitionType = Partitioning.partitionType(table);
     // we need to compute a hash value for the partition struct so that it can be used as a sorting key
-    return computeHash(part);
+    return computeHash(partitionType).hash(part);
```
okumin, line 214: Disclaimer: This is an impression of myself who started looking at the change 1 min ago. I might change my stance tomorrow

It seems strange that the value of a digest affects the result set. I would understand if the problem statement were "hash collisions are causing performance issues"

iceberg/iceberg-handler/src/test/queries/positive/iceberg_minor_compaction_bucket.q
```diff
@@ -0,0 +1,56 @@
...
+('l', null, null);
+
+desc formatted default.srcbucket_big;
+SELECT * FROM default.srcbucket_big ORDER BY id;
+
+alter table srcbucket_big compact 'minor' and wait;
+show compactions order by 'partition';
+
+desc formatted default.srcbucket_big;
+SELECT * FROM default.srcbucket_big ORDER BY id;
```
okumin, line 56: Can we add the following query before and after the compaction? I expect data files with obsolete specs to be rewritten by the current partition spec. The following query will assert our expectations regardless of whether my expectations are correct.

```sql
select `partition`, spec_id, record_count
from default.srcbucket_big.partitions
order by `partition`, spec_id, record_count;
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergAcidUtil.java
difin: Partition hash value is assumed to uniquely identify a partition and this assumption is used in constructing Iceberg compaction query.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_minor_compaction_bucket.q
difin: Added the query, but please consider that compaction is expected to compact data only if partitions meet compaction criteria. if the compaction criteria are not met data files with obsolete specs can remain as is.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -88,28 +89,30 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
       conf.set(IcebergCompactionService.PARTITION_PATH, new Path(partSpec).toString());
 
       Map<String, String> partSpecMap = new LinkedHashMap<>();
       Warehouse.makeSpecFromName(partSpecMap, new Path(partSpec), null);
 
-      compactionQuery = String.format("insert overwrite table %1$s select * from %1$s where %2$s=%3$d " +
-              "and %4$s is not null %5$s %6$s", compactTableName, VirtualColumn.PARTITION_HASH.getName(), partitionHash,
-          VirtualColumn.FILE_PATH.getName(), fileSizePredicate == null ? "" : "and " + fileSizePredicate, orderBy);
+      compactionQuery = String.format("INSERT OVERWRITE TABLE %1$s SELECT * FROM %1$s " +
+              "WHERE %2$s LIKE '%%/%3$s/data/%4$s/%%' AND %5$s = %6$d %7$s %8$s",
```
okumin, line 108: `'%%/%3$s/data/%4$s/%%'` assumes a strict file structure. In case that users performed the in-place migration, the file name can't follow the convention. It sounds acceptable at this point.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_minor_compaction_bucket.q
okumin: Thanks. Looks good.

Before: 12 records

```
{"key_bucket":0,"key_bucket_8":null}	0	3
{"key_bucket":3,"key_bucket_8":null}	0	1
{"key_bucket":null,"key_bucket_8":0}	1	1
{"key_bucket":null,"key_bucket_8":3}	1	1
{"key_bucket":null,"key_bucket_8":4}	1	2
{"key_bucket":null,"key_bucket_8":null}	1	4
```

After: 12 records

```
{"key_bucket":null,"key_bucket_8":0}	1	2
{"key_bucket":null,"key_bucket_8":3}	1	2
{"key_bucket":null,"key_bucket_8":4}	1	4
{"key_bucket":null,"key_bucket_8":null}	1	4
```

okumin (approved):
👍

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -520,7 +521,8 @@ public static List<String> getPartitionNames(Table table, Map<String, String> pa
     }
   }
 
-  public static long getPartitionHash(Table icebergTable, String partitionPath) throws IOException {
+  public static Pair<Integer, StructProjection> getPartitionSpecIdAndStruct(Table icebergTable, String partitionPath)
```
deniskuzZ, line 524: minor, `getPartitionStructWithSpecId`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,16 +106,46 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
         throw new HiveException(ErrorMsg.COMPACTION_NO_PARTITION);
       }
     } else {
-      long partitionHash = IcebergTableUtil.getPartitionHash(icebergTable, partSpec);
+      Pair<Integer, StructProjection> partSpecPair =
+          IcebergTableUtil.getPartitionSpecIdAndStruct(icebergTable, partSpec);
+      int partitionSpecId = partSpecPair.getKey();
```
deniskuzZ, line 111: use simply `specId` and `partition`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,16 +106,46 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
         throw new HiveException(ErrorMsg.COMPACTION_NO_PARTITION);
       }
     } else {
-      long partitionHash = IcebergTableUtil.getPartitionHash(icebergTable, partSpec);
+      Pair<Integer, StructProjection> partSpecPair =
+          IcebergTableUtil.getPartitionSpecIdAndStruct(icebergTable, partSpec);
```
deniskuzZ, line 110: why do you need to list `PARTITIONS` on the HS2 Worker? Can't we combine it with the main query?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,16 +106,46 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
       HiveConf.setVar(conf, ConfVars.REWRITE_POLICY, RewritePolicy.PARTITION.name());
       conf.set(IcebergCompactionService.PARTITION_PATH, new Path(partSpec).toString());
 
       Map<String, String> partSpecMap = new LinkedHashMap<>();
       Warehouse.makeSpecFromName(partSpecMap, new Path(partSpec), null);
 
-      compactionQuery = String.format("insert overwrite table %1$s select * from %1$s where %2$s=%3$d " +
-              "and %4$s is not null %5$s %6$s", compactTableName, VirtualColumn.PARTITION_HASH.getName(), partitionHash,
-          VirtualColumn.FILE_PATH.getName(), fileSizePredicate == null ? "" : "and " + fileSizePredicate, orderBy);
+      Types.StructType partitionType = Partitioning.partitionType(icebergTable);
```
deniskuzZ, line 122: can we extract this logic into an iceberg `CompactionQueryBuilder`? see `CompactionQueryBuilderFactory`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,16 +106,46 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
+      String partitionCondition = IntStream.range(0, partitionType.fields().size())
+          .mapToObj(i -> {
+            Types.NestedField field = partitionType.fields().get(i);
+            String name = "`partition`." + HiveUtils.unparseIdentifier(field.name());
+            String type = field.type().toString().equalsIgnoreCase("long") ? "bigint" :
+                field.type().toString();
+
+            Object value = partitionValues.get(i, Object.class);
+            if (type.equalsIgnoreCase("date") && value != null) {
+              value = String.format("date_add('1970-01-01', %s)", value);
```
deniskuzZ, line 132: should we use `HiveValueConverter` here?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,16 +106,46 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
 
-      compactionQuery = String.format("insert overwrite table %1$s select * from %1$s where %2$s=%3$d " +
-              "and %4$s is not null %5$s %6$s", compactTableName, VirtualColumn.PARTITION_HASH.getName(), partitionHash,
-          VirtualColumn.FILE_PATH.getName(), fileSizePredicate == null ? "" : "and " + fileSizePredicate, orderBy);
+      Types.StructType partitionType = Partitioning.partitionType(icebergTable);
+      String partitionCondition = IntStream.range(0, partitionType.fields().size())
+          .mapToObj(i -> {
+            Types.NestedField field = partitionType.fields().get(i);
+            String name = "`partition`." + HiveUtils.unparseIdentifier(field.name());
+            String type = field.type().toString().equalsIgnoreCase("long") ? "bigint" :
```
deniskuzZ, line 127: can we use `HiveSchemaUtil.convertToTypeString`?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution2.q
```diff
@@ -29,23 +29,23 @@ use ice_comp;
 create table ice_orc (
     first_name string, 
     last_name string,
-    dept_id bigint
+    is_remote boolean
```
deniskuzZ, line 32: why change the test?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition.q
```diff
@@ -29,19 +29,19 @@ create table ice_orc_wo_evo (
     first_name string, 
     last_name string
  )
-partitioned by (dept_id bigint, 
+partitioned by (dept_id decimal(10, 2),
```
deniskuzZ, line 32: dept_id can't be a decimal, why change here?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition_with_evolution.q
```diff
@@ -29,42 +29,42 @@ create table ice_orc (
     last_name string,
     registration_date date
  )
-partitioned by (dept_id bigint, 
+partitioned by (dept_id float,
```
deniskuzZ, line 32: same

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition_with_evolution2.q
```diff
@@ -28,37 +28,37 @@ set hive.merge.tezfiles=true;
 create table ice_orc (
     a string
  )
-partitioned by (b bigint)
+partitioned by (b double)
```
deniskuzZ, line 31: if you need to test all supported types maybe add a separate test for that and use alter command to change the type.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
difin: Improved the fix to not assume that files are places under data folder.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Fixed

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
difin: Listing partitions meta-table to find the specId and partition struct that correspond to the given partition name in human readable format and then to build a condition on files meta-table with constraints on `files.partition` struct values. I don't see how all this logic can be implemented in an SQL in the main query, do you?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
difin: There is a JIRA in the backlog for this: https://issues.apache.org/jira/browse/HIVE-28731

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution2.q
difin: There are many iceberg compaction q-tests that have partition fields with 'bigint' types, changed this test to test with boolean type. This way iceberg compaction q-tests test more cases, isn't it a good thing?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition.q
difin: Same, to increase test coverage of partition field data types in iceberg compaction. Yes, `dept_id` wouldn't likely be a decimal in a real world, but it is only a field name which doesn't affect anything?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition_with_evolution2.q
difin: I remember on another PR you mentioned that there are already too many q-tests and precommit runs for too long time. These changes of data types do not change what these tests test, but increase coverage on different partition key data types. Are you ok if I correct field names to make them logically suitable to their new data types and leave the data type changes?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
deniskuzZ: `It is done to find the specId and partition struct that correspond to the given partition name` - why can't we do this directly on FILES table?
````
select FILE_PATH from default.ice_orc.FILES where `partition` = named_struct("event_src_trunc","AAA");
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java (hunk 11 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java:124]: can we at least move it into construct sql method

File: iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition_with_evolution2.q (hunk 17 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition_with_evolution2.q:31]: i think, we should create an integration test similar to `CompactorOnTezTest `

File: iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition_with_evolution.q (hunk 16 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition_with_evolution.q]: closing as duplicate issue

File: iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition.q (hunk 15 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition.q]: closing as duplicate issue

File: iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution2.q (hunk 14 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_partition_evolution2.q:32]: closing as duplicate issue

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java (hunk 10 above)
Comment (difin) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java]: > why can't we do this directly on FILES table? 
Because we need to know the values that need to be used in the `named_struct`, but IcebergQueryCompactor gets partition name in the human readable format and partition values needs to be retrieved and converted.

An example from existing Iceberg compaction q-test: we compact a table with partition spec `spec(truncate(3, event_src), month(event_time))`. 
IcebergQueryCompactor gets `ci.partName` = `event_src_trunc=BBB/event_time_month=2024-08`.
The query on the `files` table needs to be 
`select FILE_PATH from default.ice_orc.files where `partition` = named_struct("event_src_trunc","AAA",event_time_month,655);`

`2024-08` needs to be converted to `655`.

I am listing partitions table to find the spec id and partition struct for the given `ci.partName` and applying conversions to get the values that can be used in the condition on `partition` field from `files` metatable.

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java]: > To do that I am listing partitions table to find the spec id and the partition struct for the given ci.partName and applying conversions to get the values that can be used in the condition on partition field from files metastable.
`partition` struct in the PARTITIONS table == `partition` struct in the `FILES` table, so I still don't understand why we can't directly look up the `FILES` table?

Comment (difin) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java]: > partition struct in the PARTITIONS table == partition struct in the FILES table
Correct

> partition struct in the PARTITIONS table == partition struct in the FILES table, so I still don't understand why we can't directly look up the FILES table?
We don't have the values that can be used in SQL conditions on the partition struct.

We do not lookup the `partitions` table using SQL. We lookup using Iceberg API to get the raw values that after transformations can be used in SQL on partition struct.

> From the above example, what do we pass in partitionPath: event_src_trunc=BBB/event_time_month=2024-08? Based on that can't we construct a proper named_struct filter?
No, we can't construct based on human readable values.
In this example, `ci.partName` = `event_src_trunc=BBB/event_time_month=2024-08`
but struct in files metadata table is  `named_struct("event_src_trunc","AAA",event_time_month,655)`

`event_time_month` is int with value of 655.

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java (hunk 11 above)
Comment (difin) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java:124]: Done

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java [hunk 18]
```diff
@@ -96,16 +108,39 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
         throw new HiveException(ErrorMsg.COMPACTION_NO_PARTITION);
       }
     } else {
-      long partitionHash = IcebergTableUtil.getPartitionHash(icebergTable, partSpec);
+      Pair<Integer, StructProjection> partSpecPair =
+          IcebergTableUtil.getPartitionStructWithSpecId(icebergTable, partSpec);
+      int specId = partSpecPair.getKey();
+      StructProjection partition = partSpecPair.getValue();
+
+      HiveConf.setBoolVar(conf, ConfVars.HIVE_CONVERT_JOIN, false);
```
okumin, line 116: Why is this needed? I'm just curious, and it's acceptable if we have reasons.

okumin (approved):

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
difin: When enabled, compaction query is failing with the following exception in `convertJoinOpMapJoinOp`:

INSERT OVERWRITE TABLE default.ice_orc SELECT * FROM default.ice_orc
WHERE FILE__PATH IN (SELECT FILE_PATH FROM default.ice_orc.FILES WHERE partition.event_src_trunc='AAA' AND SPEC_ID=0) AND PARTITION__SPEC__ID = 0

```
INFO  : Compiling command(queryId=hive_20250507190138_8a53bfd0-9ed8-4951-9aa3-548e57b2ba93): INSERT OVERWRITE TABLE default.ice_orc SELECT d.* FROM default.ice_orc d, default.ice_orc.files f
WHERE d.FILE__PATH = f.FILE_PATH and f.partition.event_src_trunc='AAA' AND f.SPEC_ID=0 AND d.PARTITION__SPEC__ID = 0
INFO  : No Stats for default@ice_orc, Columns: event_id, event_src, event_time
INFO  : No Stats for default@ice_orc, Columns: file_path, partition, spec_id
INFO  : No Stats for default@ice_orc, Columns: partition
ERROR : FAILED: NullPointerException null
java.lang.NullPointerException
	at org.apache.hadoop.hive.ql.plan.ExprNodeDescUtils.indexOf(ExprNodeDescUtils.java:77)
	at org.apache.hadoop.hive.ql.plan.ExprNodeDescUtils.indexOf(ExprNodeDescUtils.java:72)
	at org.apache.hadoop.hive.ql.optimizer.MapJoinProcessor.getMapJoinDesc(MapJoinProcessor.java:1311)
	at org.apache.hadoop.hive.ql.optimizer.MapJoinProcessor.convertJoinOpMapJoinOp(MapJoinProcessor.java:584)
	at org.apache.hadoop.hive.ql.optimizer.ConvertJoinMapJoin.convertJoinMapJoin(ConvertJoinMapJoin.java:1348)
	at org.apache.hadoop.hive.ql.optimizer.ConvertJoinMapJoin.process(ConvertJoinMapJoin.java:208)
	at org.apache.hadoop.hive.ql.lib.DefaultRuleDispatcher.dispatch(DefaultRuleDispatcher.java:90)
	at org.apache.hadoop.hive.ql.lib.DefaultGraphWalker.dispatchAndReturn(DefaultGraphWalker.java:105)
	at org.apache.hadoop.hive.ql.lib.DefaultGraphWalker.dispatch(DefaultGraphWalker.java:89)
	at org.apache.hadoop.hive.ql.lib.ForwardWalker.walk(ForwardWalker.java:74)
	at org.apache.hadoop.hive.ql.lib.DefaultGraphWalker.startWalking(DefaultGraphWalker.java:120)
	at org.apache.hadoop.hive.ql.parse.TezCompiler.runStatsDependentOptimizations(TezCompiler.java:485)
	at org.apache.hadoop.hive.ql.parse.TezCompiler.optimizeOperatorPlan(TezCompiler.java:218)
	at org.apache.hadoop.hive.ql.parse.TaskCompiler.compile(TaskCompiler.java:178)
	at org.apache.hadoop.hive.ql.parse.SemanticAnalyzer.analyzeInternal(SemanticAnalyzer.java:13763)
	at org.apache.hadoop.hive.ql.parse.CalcitePlanner.analyzeInternal(CalcitePlanner.java:489)
	at org.apache.hadoop.hive.ql.parse.BaseSemanticAnalyzer.analyze(BaseSemanticAnalyzer.java:327)
	at org.apache.hadoop.hive.ql.Compiler.analyze(Compiler.java:227)
	at org.apache.hadoop.hive.ql.Compiler.compile(Compiler.java:108)
	at org.apache.hadoop.hive.ql.Driver.compile(Driver.java:203)
	at org.apache.hadoop.hive.ql.Driver.compileInternal(Driver.java:651)
	at org.apache.hadoop.hive.ql.Driver.compileAndRespond(Driver.java:596)
	at org.apache.hadoop.hive.ql.Driver.compileAndRespond(Driver.java:590)
	at org.apache.hadoop.hive.ql.reexec.ReExecDriver.compileAndRespond(ReExecDriver.java:126)
	at org.apache.hive.service.cli.operation.SQLOperation.prepare(SQLOperation.java:209)
	at org.apache.hive.service.cli.operation.SQLOperation$BackgroundWork$1.run(SQLOperation.java:340)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/javax.security.auth.Subject.doAs(Subject.java:423)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1899)
	at org.apache.hive.service.cli.operation.SQLOperation$BackgroundWork.run(SQLOperation.java:362)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
difin: Done, now not scanning `partitions` table anymore to get partition struct and spec id.
The only thing I did differently is the condition on struct fields.
instead of `where `partition` = named_struct('field1','value1','field2','value2')` I implemented 
`where `partition`.`field1` = 'value1' and `partition`.`field2` = 'value2'

because this is equivalent and the first variant doesn't work - the values in named_struct condition need to be of exact Iceberg types, it requires type casting, I bumped into cases where it was not evident which type to use.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_major_compaction_single_partition_with_evolution2.q
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -520,25 +521,27 @@ public static List<String> getPartitionNames(Table table, Map<String, String> pa
...
-          .filter(e -> e.getKey().equals(partitionPath))
-          .transform(e -> IcebergAcidUtil.computeHash(e.getValue()))
-          .get(0);
+  public static Integer getPartitionSpecId(Table icebergTable, String partitionPath) {
+    if (icebergTable == null || partitionPath == null || partitionPath.isEmpty()) {
+      throw new IllegalArgumentException("Table and partitionPath must not be null or empty.");
     }
+
+    // Extract field names from the path: "field1=val1/field2=val2" → [field1, field2]
+    List<String> fieldNames = Arrays.stream(partitionPath.split("/"))
```
deniskuzZ, line 531: can we use `Warehouse.makeSpecFromName(partitionPath).keySet()`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -520,25 +521,27 @@ public static List<String> getPartitionNames(Table table, Map<String, String> pa
...
-            PartitionSpec spec = icebergTable.specs().get(row.get(IcebergTableUtil.SPEC_IDX, Integer.class));
-            PartitionData partitionData = IcebergTableUtil.toPartitionData(data,
-                Partitioning.partitionType(icebergTable), spec.partitionType());
-            String path = spec.partitionToPath(partitionData);
-            return Maps.immutableEntry(path, data);
-          })
-          .filter(e -> e.getKey().equals(partitionPath))
-          .transform(e -> IcebergAcidUtil.computeHash(e.getValue()))
-          .get(0);
+  public static Integer getPartitionSpecId(Table icebergTable, String partitionPath) {
```
deniskuzZ, line 525: `getSpecIdByPartition(Table icebergTable, String partitionName)` ?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
deniskuzZ: @difin, why check `PARTITION__SPEC__ID = 0` when you already have a `SPEC_ID=0` filter?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,31 +127,46 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
     }
+    return compactionQuery;
+  }
 
-    SessionState sessionState = setupQueryCompactionSession(conf, ci, tblProperties);
-    String compactionTarget = "table " + HiveUtils.unparseIdentifier(compactTableName) +
-        (partSpec != null ? ", partition " + HiveUtils.unparseIdentifier(partSpec) : "");
+  private String buildPartitionPredicate(CompactionInfo ci, Table icebergTable) {
+    Map<String, String> partSpecMap = new LinkedHashMap<>();
+    Warehouse.makeSpecFromName(partSpecMap, new Path(ci.partName), null);
```
deniskuzZ, line 149: why not use simpler method?
````
Map<String, String> partSpecMap = Warehouse.makeSpecFromName(ci.partName))
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java [hunk 22]
```diff
@@ -96,31 +127,46 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
-    } catch (HiveException e) {
-      LOG.error("Failed compacting {}", compactionTarget, e);
-      throw e;
-    } finally {
-      sessionState.setCompaction(false);
-    }
+    Map<String, PartitionField> partitionFieldMap = IcebergTableUtil.getPartitionFields(icebergTable, false)
+        .stream().collect(Collectors.toMap(PartitionField::name, Function.identity()));
+
+    Types.StructType partitionType = Partitioning.partitionType(icebergTable);
```
deniskuzZ, line 154: why are we adding partition fields for all specs?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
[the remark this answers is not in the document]
difin: `Partitioning.partitionType(icebergTable)` returns all partition fields, from all partition specs, which is needed for us, but `getPartitionSpec` will return only one spec. I changed the method `getPartitionSpecId` to return `PartitionSpec` to make it more generic for reuse, but didn't replace `partitionType` and `fields` parts.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
difin: Changed the method to return `PartitionSpec`, but it cannot replace `Partitioning.partitionType(icebergTable)`. Because the later returns all partition fields, from all partition specs, which is needed for us, but `getPartitionSpec` returns only one spec, it doesn't have all the fields.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
difin: Fixed

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
deniskuzZ: Why do we need all specs? We find the spec conforming to the partition pattern, don't we?

difin: Because `partition` field in `files` metadata table contains partition struct with all partition fields, from all partition specs, so the condition on partition struct should include all fields.

difin: `partition` field in `files` metadata table contains partition struct with all partition fields, from all partition specs, so the condition on partition struct should include all fields, for that we need `Partitioning.partitionType(icebergTable)`

fields we can get from our spec, because all fields not in the spec need to have null values - done.

deniskuzZ: so what, don't you construct the predicate to only check the relevant fields from the relevant PartitionSpec?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,31 +128,50 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
 
-      Map<String, String> partSpecMap = new LinkedHashMap<>();
-      Warehouse.makeSpecFromName(partSpecMap, new Path(partSpec), null);
+      PartitionSpec spec;
+      try {
+        spec = IcebergTableUtil.getPartitionSpec(icebergTable, ci.partName);
+      } catch (MetaException e) {
+        throw new HiveException(e);
+      }
+      String partitionPredicate = buildPartitionPredicate(ci, spec);
```
deniskuzZ, line 142: Where do you check for null? if you don't need maybe throw exception in getPartitionSpec?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,31 +128,50 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
     }
+    return compactionQuery;
+  }
 
-    SessionState sessionState = setupQueryCompactionSession(conf, ci, tblProperties);
-    String compactionTarget = "table " + HiveUtils.unparseIdentifier(compactTableName) +
-        (partSpec != null ? ", partition " + HiveUtils.unparseIdentifier(partSpec) : "");
+  private String buildPartitionPredicate(CompactionInfo ci, PartitionSpec spec) {
+    Map<String, String> partSpecMap = new LinkedHashMap<>();
+    Warehouse.makeSpecFromName(partSpecMap, new Path(ci.partName), null);
```
deniskuzZ, line 154: can we do `Map<String, String> partSpecMap = Warehouse.makeSpecFromName(ci.partName)`?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergFilterFactory.java
```diff
@@ -228,6 +228,32 @@ private static Object convertLiteral(Object literal, TransformSpec transform) {
     }
   }
 
+  public static Object convertPartitionLiteral(Object literal, TransformSpec transform) {
```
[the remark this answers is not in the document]
difin: Because for the `DAY`, `TRUNCATE` and `BUCKET` transforms `convertLiteral()` method makes wrong conversions which cause exceptions. I added a Java integration test that tests all the transforms.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergQueryCompactor.java
```diff
@@ -96,31 +128,50 @@ public boolean run(CompactorContext context) throws IOException, HiveException,
...
-    }
+    Map<String, PartitionField> partitionFieldMap = spec.fields().stream()
+        .collect(Collectors.toMap(PartitionField::name, Function.identity()));
+
+    Types.StructType partitionType = spec.partitionType();
+    return  partitionType.fields().stream().map(field -> {
+      String value = partSpecMap.get(field.name());
+      String literal = "NULL";
+
+      if (value != null && !value.equals("null")) {
```
deniskuzZ, line 164: can it be "null" ?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergFilterFactory.java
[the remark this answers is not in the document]
deniskuzZ:
````
private static Object convertLiteral(Object literal, TransformSpec transform) {
    if (transform == null) {
      return literal;
    }
    try {
      switch (transform.getTransformType()) {
        case YEAR:
        case MONTH:
        case DAY:
        case HOUR:
          return convertTimeLiteral(literal));
        case TRUNCATE:
          return Transforms.truncate(transform.getTransformParam().get()).bind(Types.StringType.get())
              .apply(literal.toString());
        case BUCKET:
          return Transforms.bucket(transform.getTransformParam().get()).bind(Types.StringType.get())
              .apply(literal.toString());
        case IDENTITY:
          return literal;
        default:
          throw new UnsupportedOperationException("Unknown transform: " + transform.getTransformType());
      }
    } catch (NumberFormatException | DateTimeException | IllegalStateException e) {
      throw new RuntimeException(
          String.format("Unable to parse value '%s' as '%s' transform value", literal.toString(), transform));
    }
  }
````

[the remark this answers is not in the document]
deniskuzZ: ok

deniskuzZ (approved):
LGTM, pending tests

PR #5795, opened 2025-04-30
HIVE-28910: Remove redundant IS NOT NULL predicates when expanding SEARCH

zabetak:
Better plans

Existing tests


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/SearchTransformer.java
```diff
@@ -128,8 +132,23 @@ public Shuttle(final RexBuilder rexBuilder) {
         } else {
           return call;
         }
+      case IS_NULL:
+      case IS_NOT_NULL:
+      case CASE:
+      case COALESCE:
+        // Everything below must be transformed with the UNKNOWN handler
```
asolimando, line 139: Can you elaborate on why you can't use the outermost handler here?

In the Jira ticket you seem towards the opposite direction, that is taking into account that in `WHERE` clauses filters obey to the `unknownAsFalse` semantics.

I haven't reviewed the Calcite upgrade which introduced some of the code you touch here, so maybe it should be clear from the context.

zabetak: There are some operators with special semantics that we need to retain the 3-valued logic even if we are inside the `WHERE` clause. 

Example: `COALESCE(SEARCH(...), true)`
1. If SEARCH -> true then COALESCE -> true
2. If SEARCH -> false then COALESCE -> false
3. If SEARCH -> null then COALESCE -> true

If we consider `unknownAsFalse` inside the COALESCE then we will incorrectly drop the `IS NOT NULL` predicate during the expansion and the COALESCE will fall into the third case returning `true` instead of `false`.

Basically, if the SEARCH is below/inside an operator that is sensitive to null values we cannot safely perform the simplification.

Note that `RexSimplify` also changes the default unknown semantics for certain operators (see `simplifyIs2`, `simplifyCoalesce`, etc.).

asolimando: I suspected something similar, thanks for the clarification.

Since this is a rather deep technical detail that most people not familiar with CBO will have problems with, I'd suggest to have a more explicit comment like "// These operator kinds require the UNKNOWN handler to correctly handle NULL values" or something similar.

I know it's still a draft PR but for the rest LGTM already, feel free to ping me for a review if needed.

zabetak: I added more comments and also made the approach a bit more defensive in the sense that if SEARCH happens to be under any operator other than AND/OR we don't perform the simplification.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInBetweenExpandRule.java
```diff
@@ -152,7 +153,7 @@ private RexInBetweenExpander(RexBuilder rexBuilder) {
     public RexNode visitCall(final RexCall call) {
       switch (call.getKind()) {
       case SEARCH: {
-        return new SearchTransformer<>(rexBuilder, call).transform().accept(this);
+        return new SearchTransformer<>(rexBuilder, call, RexUnknownAs.UNKNOWN).transform().accept(this);
```
kasakrisz, line 156: Is this rule expand in and between predicates in Filter operator? Should `RexUnknownAs.FALSE` be used in that case?

zabetak: The RexInBetweenExpander is also used for Project and Join operators so if we want to change the unknown context we should probably generalize a bit this class. Not sure if it will make a huge difference in the plans but I can try it out if you want.

kasakrisz: I'm ok with a follow-up jira/PR and merge this one.

zabetak:
I added two commits to sync with master and resolve conflicts. If it comes green I will merge the PR tomorrow.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveInBetweenExpandRule.java
zabetak: Raised https://github.com/apache/hive/pull/5906 as follow-up. Let's see how that goes.

zabetak:
Thanks for the reviews @kasakrisz and @asolimando !

PR #5798, opened 2025-04-30
HIVE-28938: Error in LATERAL VIEW with non native tables due to prese…

soumyakanti3578:
…nce of incorrect virtual columns in RowResolver

1. Refactored virtual column adding logic in `CalcitePlanner::genTableLogicalPlan`
2. Removed `obtainTableType(Table tabMetaData)` as the logic inside it was insufficient for classifying tables
3. Also removed `enum TableType` containing types `NATIVE`, `DRUID`, and `JDBC`.
4. Added methods `isDruidTable` and `isJdbcTable` in `Table` as an alternative to the enum types. `isNonNative` is already present in `Table`.

Explained in detail in https://issues.apache.org/jira/browse/HIVE-28938, but in short, iceberg tables were getting misclassified as NATIVE tables leading to addition of wrong virtual columns in RowResolver.

No

```
mvn test -pl itests/qtest-iceberg -Piceberg -Pitests -Dtest=TestIcebergCliDriver -Dtest.output.overwrite=true -Dqfile="iceberg_multiple_lateral_views.q"
```


amansinha100:
> Explained in detail in https://issues.apache.org/jira/browse/HIVE-28938, but in short, iceberg tables were getting misclassified as NATIVE tables leading to addition of wrong virtual columns in RowResolver.

The nature of the code change in the PR is not limited to lateral views, so I am curious is this a wider issue for Iceberg tables ?  Can you test with nested subqueries in the FROM clause, nested regular views, nested materialized views ?

soumyakanti3578:
This is indeed a wider issue. All ICEBERG tables were classified as NATIVE tables in `CalcitePlanner::genTableLogicalPlan`, however, the side effect was only visible because the lateral view references all virtual columns too, as shown in the Jira. This is also visible when we do an `explain formatted select * from <table>` as shown below by running:
```
create external table test (a int, b int, c int, d struct<e:string, f:int>) stored by iceberg;
explain formatted select * from test;
```
We see NATIVE and NON NATIVE virtual columns in the output:
```
{"CBOPlan":"{
  \"rels\": [
    {
      \"id\": \"0\",
      \"relOp\": \"org.apache.hadoop.hive.ql.optimizer.calcite.reloperators.HiveTableScan\",
      \"table\": [
        \"default\",
        \"test\"
      ],
      \"table:alias\": \"test\",
      \"inputs\": [],
      \"rowCount\": 1.0,
      \"avgRowSize\": 248.0,
      \"rowType\": {
        \"fields\": [
          {
            \"type\": \"INTEGER\",
            \"nullable\": true,
            \"name\": \"a\"
          },
          {
            \"type\": \"INTEGER\",
            \"nullable\": true,
            \"name\": \"b\"
          },
          {
            \"type\": \"INTEGER\",
            \"nullable\": true,
            \"name\": \"c\"
          },
          {
            \"fields\": {
              \"fields\": [
                {
                  \"type\": \"VARCHAR\",
                  \"nullable\": true,
                  \"precision\": 2147483647,
                  \"name\": \"e\"
                },
                {
                  \"type\": \"INTEGER\",
                  \"nullable\": true,
                  \"name\": \"f\"
                }
              ],
              \"nullable\": true
            },
            \"nullable\": true,
            \"name\": \"d\"
          },
          {
            \"type\": \"BIGINT\",
            \"nullable\": true,
            \"name\": \"BLOCK__OFFSET__INSIDE__FILE\"
          },
          {
            \"type\": \"VARCHAR\",
            \"nullable\": true,
            \"precision\": 2147483647,
            \"name\": \"INPUT__FILE__NAME\"
          },
          {
            \"fields\": {
              \"fields\": [
                {
                  \"type\": \"BIGINT\",
                  \"nullable\": true,
                  \"name\": \"writeid\"
                },
                {
                  \"type\": \"INTEGER\",
                  \"nullable\": true,
                  \"name\": \"bucketid\"
                },
                {
                  \"type\": \"BIGINT\",
                  \"nullable\": true,
                  \"name\": \"rowid\"
                }
              ],
              \"nullable\": true
            },
            \"nullable\": true,
            \"name\": \"ROW__ID\"
          },
          {
            \"type\": \"BOOLEAN\",
            \"nullable\": true,
            \"name\": \"ROW__IS__DELETED\"
          },
          {
            \"type\": \"INTEGER\",
            \"nullable\": true,
            \"name\": \"PARTITION__SPEC__ID\"
          },
          {
            \"type\": \"BIGINT\",
            \"nullable\": true,
            \"name\": \"PARTITION__HASH\"
          },
          {
            \"type\": \"VARCHAR\",
            \"nullable\": true,
            \"precision\": 2147483647,
            \"name\": \"FILE__PATH\"
          },
          {
            \"type\": \"BIGINT\",
            \"nullable\": true,
            \"name\": \"ROW__POSITION\"
          },
          {
            \"type\": \"VARCHAR\",
            \"nullable\": true,
            \"precision\": 2147483647,
            \"name\": \"PARTITION__PROJECTION\"
          },
          {
            \"type\": \"BIGINT\",
            \"nullable\": true,
            \"name\": \"SNAPSHOT__ID\"
          }
        ],
        \"nullable\": false
      },
      \"colStats\": [
        {
          \"name\": \"a\",
          \"ndv\": 1,
          \"minValue\": -9223372036854775808,
          \"maxValue\": 9223372036854775807
        },
        {
          \"name\": \"b\",
          \"ndv\": 1,
          \"minValue\": -9223372036854775808,
          \"maxValue\": 9223372036854775807
        },
        {
          \"name\": \"c\",
          \"ndv\": 1,
          \"minValue\": -9223372036854775808,
          \"maxValue\": 9223372036854775807
        },
        {
          \"name\": \"d\",
          \"ndv\": 1
        }
      ]
    },
    {
      \"id\": \"1\",
      \"relOp\": \"org.apache.hadoop.hive.ql.optimizer.calcite.reloperators.HiveProject\",
      \"fields\": [
        \"a\",
        \"b\",
        \"c\",
        \"d\"
      ],
      \"exprs\": [
        {
          \"input\": 0,
          \"name\": \"$0\"
        },
        {
          \"input\": 1,
          \"name\": \"$1\"
        },
        {
          \"input\": 2,
          \"name\": \"$2\"
        },
        {
          \"input\": 3,
          \"name\": \"$3\"
        }
      ],
      \"rowCount\": 1.0
    }
  ]
  }","optimizedSQL":"SELECT `a`, `b`, `c`, `d`
FROM `default`.`test`","cboInfo":"Plan optimized by CBO.","STAGE DEPENDENCIES":{"Stage-0":{"ROOT STAGE":"TRUE"}},"STAGE PLANS":{"Stage-0":{"Fetch Operator":{"limit:":"-1","Processor Tree:":{"TableScan":{"alias:":"test","columns:":["a","b","c","d"],"database:":"default","table:":"test","isTempTable:":"false","OperatorId:":"TS_0","children":{"Select Operator":{"expressions:":"a (type: int), b (type: int), c (type: int), d (type: struct<e:string,f:int>)","columnExprMap:":{"_col0":"a","_col1":"b","_col2":"c","_col3":"d"},"outputColumnNames:":["_col0","_col1","_col2","_col3"],"OperatorId:":"SEL_1","children":{"ListSink":{"OperatorId:":"LIST_SINK_3"}}}}}}}}}}
```

ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java
```diff
@@ -1364,4 +1365,16 @@ public SourceTable createSourceTable() {
...
     return sourceTable;
   }
+  
+  public boolean isDruidTable() {
+    return isNonNative() && getStorageHandler().toString().equals(Constants.DRUID_HIVE_STORAGE_HANDLER_ID);
+  }
+  
+  public boolean isJdbcTable() {
+    return isNonNative() && getStorageHandler().toString().equals(Constants.JDBC_HIVE_STORAGE_HANDLER_ID);
+  }
```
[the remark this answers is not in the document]
soumyakanti3578: For #1, I can add a null check like before.

> Could you please share how this refactor is related to collecting the virtual columns?
Wrong virtual columns were being added because `obtainTableType` was incorrectly classifying iceberg tables as NATIVE. So I wanted to remove/fix the method. Moreover, I noticed that the enum `TableType` is only used in a very few places, and was only returned by this method `obtainTableType`. If we remove the method, then there's no real reason to keep the enum.

So I decided to move the table type logic like:
```
if (storageHandlerStr
            .equals(Constants.DRUID_HIVE_STORAGE_HANDLER_ID)) {
          return TableType.DRUID;
        }
```
to their respective methods in Table.java. I thought this was a better way because we already have methods like `isNonNative` in Table.java.

While this is strictly not needed for collecting virtual columns, I thought obtainTableType and enum TableType are really not needed, and decided to remove them. Let me know if you want me to revert these changes or move them somewhere else. I will add a null check if we decide to keep them here.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java
```diff
@@ -1364,4 +1364,8 @@ public SourceTable createSourceTable() {
     sourceTable.setDeletedCount(0L);
     return sourceTable;
   }
+  
+  public boolean isNative() {
+    return !isNonNative();
+  }
```
zabetak, line 1370: nit: the method is trivial; I guess we can leave without it.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -12193,6 +12188,22 @@ samplePredicate, true, new SampleDesc(ts.getNumerator(),
...
+      virtualColumns.addAll(tab.getStorageHandler().acidVirtualColumns());
+    }
+    if (tab.isNonNative() && tab.getStorageHandler().areSnapshotsSupported() &&
+        isBlank(tab.getMetaTable())) {
+      virtualColumns.add(VirtualColumn.SNAPSHOT_ID);
+    }
+    
+    return virtualColumns;
+  }
+
```
zabetak, line 12206: nit: I like the refactoring although I feel that the method may have better encapsulation if moved inside `Table`.

zabetak (approved):
LGTM, pending tests! Left some nit comments but can be merged even without addressing those.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java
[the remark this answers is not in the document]
kasakrisz: Let's just focus on collecting the virtual columns. I like the new method `getVirtualColumns`

kasakrisz (approved):

hbase-handler/src/test/results/positive/hbase_queries.q.out
```diff
@@ -151,14 +151,14 @@ STAGE PLANS:
               predicate: UDFToDouble(key) is not null (type: boolean)
               Statistics: Num rows: 1 Data size: 4 Basic stats: COMPLETE Column stats: NONE
               Select Operator
-                expressions: key (type: int)
+                expressions: UDFToDouble(key) (type: double)
```
amansinha100, line 154: I am curious why HBase (and Kudu in another file) column type changed from int to double. The stats above shows 4 bytes which indicates integer.

soumyakanti3578:
@zabetak @kasakrisz @amansinha100 
I have addressed all the comments and the tests are finally passing. 
Please review and merge if everything looks good.

hbase-handler/src/test/results/positive/hbase_queries.q.out
amansinha100: The new plan with the pushed down cast 'key' to double on both sides of the join looks good to me, although I don't know why the extraneous virtual columns in the past would have prevented the push down since the cast is on a user column.  But that's a separate topic.

amansinha100 (approved):
+1

zabetak (approved):

zabetak:
Thanks for the PR @soumyakanti3578 and for the reviews @amansinha100 and @kasakrisz !

PR #5801, opened 2025-05-05
HIVE-28924: Drop hive.optimize.join.disjunctive.transitive.predicates.pushdown and related logic

zabetak:
1. Remove code that is not actively used by existing tests.
2. Drop redundant config property that is not released yet.

No, cause anyways the disjunctive predicate pushdown currently happens irrespective the value of the property.

```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile_regex="cbo_join_transitive_pred.*"
mvn test -Dtest=TestTezTPCDS30TBPerfCliDriver -Dtest.output.overwrite
```


zabetak:
@asolimando You are very familiar with all the history behind this logic of disjunctive predicate pushdown so if you have time to take a look it would be very much appreciated!

asolimando (approved):
LGTM, glad to see that the Calcite upgrade managed to lift this limitation, great work for noticing it, Stamatis!

zabetak:
Many thanks for the review @asolimando !

PR #5803, opened 2025-05-06
HIVE-28944: Mask File, data sizes in tests

kokila-19:
Data size, table size, and file size values frequently change during upgrades or similar activities, so we are masking them in qtests.

 To avoid the need for continual updates and maintenance of these values, we are masking them. This will help reduce overhead and improve maintainability.

No

qtest

### DIFF
--- a/itests/util/src/main/java/org/apache/hadoop/hive/ql/QOutProcessor.java
+++ b/itests/util/src/main/java/org/apache/hadoop/hive/ql/QOutProcessor.java
@@ -44,13 +44,17 @@ import org.apache.hadoop.hive.ql.qoption.QTestReplaceHandler;
 public class QOutProcessor {
   public static final String PATH_HDFS_REGEX = "(hdfs://)([a-zA-Z0-9:/_\\-\\.=])+";
   public static final String PATH_HDFS_WITH_DATE_USER_GROUP_REGEX =
-      "([a-z]+) ([a-z]+)([ ]+)([0-9]+) ([0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}) "
+      "([a-z]+) ([a-z]+)(\\s+)([1-9]\\d*) (\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}) "
           + PATH_HDFS_REGEX;
+  public static final String PATH_HDFS_WITH_DATE_USER_GROUP_ZERO_SIZE_REGEX =
+          "([a-z]+) ([a-z]+)(\\s+)(0) (\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}) "
+                  + PATH_HDFS_REGEX;
 
   public static final String HDFS_MASK = "### HDFS PATH ###";
   public static final String HDFS_DATE_MASK = "### HDFS DATE ###";
   public static final String HDFS_USER_MASK = "### USER ###";
   public static final String HDFS_GROUP_MASK = "### GROUP ###";
+  public static final String HDFS_SIZE_MASK = "### SIZE ###";
 
   public static final String MASK_PATTERN = "#### A masked pattern was here ####";
   public static final String PARTIAL_MASK_PATTERN = "#### A PARTIAL masked pattern was here ####";
@@ -67,6 +71,9 @@ public class QOutProcessor {
   private static final PatternReplacementPair MASK_LINEAGE = new PatternReplacementPair(
       Pattern.compile("POSTHOOK: Lineage: .*"),
       "POSTHOOK: Lineage: ###Masked###");
+
+  public static final String FILE_SIZE_MASK = "$1#Masked#";
+  public static final String DATA_SIZE_MASK = "$1#Masked#$2";
   
   private FsType fsType = FsType.LOCAL;
 
@@ -353,10 +360,14 @@ public class QOutProcessor {
       {
         add(toPatternPair("(pblob|s3.?|swift|wasb.?).*hive-staging.*",
             "### BLOBSTORE_STAGING_PATH ###"));
-        add(toPatternPair(PATH_HDFS_WITH_DATE_USER_GROUP_REGEX, String.format("%s %s$3$4 %s $6%s",
-            HDFS_USER_MASK, HDFS_GROUP_MASK, HDFS_DATE_MASK, HDFS_MASK)));
+        add(toPatternPair(PATH_HDFS_WITH_DATE_USER_GROUP_REGEX, String.format("%s %s$3%s %s $6%s",
+            HDFS_USER_MASK, HDFS_GROUP_MASK, HDFS_SIZE_MASK, HDFS_DATE_MASK, HDFS_MASK)));
+        add(toPatternPair(PATH_HDFS_WITH_DATE_USER_GROUP_ZERO_SIZE_REGEX, String.format("%s %s$3$4 %s $6%s",
+                HDFS_USER_MASK, HDFS_GROUP_MASK, HDFS_DATE_MASK, HDFS_MASK)));
         add(toPatternPair(PATH_HDFS_REGEX, String.format("$1%s", HDFS_MASK)));
-        add(toPatternPair("(.*totalSize\\s*=*\\s*)\\d+\\s*(.*)", "$1#Masked#$2"));
+        add(toPatternPair("(.*totalSize\\s*=*\\s*)\\d+\\s*(.*)", DATA_SIZE_MASK));
+        add(toPatternPair("(.*File raw data size:*\\s*)\\d+\\s*bytes(.*)", DATA_SIZE_MASK));
+        add(toPatternPair("((?:totalFileSize|maxFileSize|minFileSize):\\s*)(?!0$)\\d+", FILE_SIZE_MASK));
       }
     };
   }
--- a/iceberg/iceberg-handler/src/test/queries/positive/query_iceberg_metadata_of_partitioned_table.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/query_iceberg_metadata_of_partitioned_table.q
@@ -1,7 +1,8 @@
 -- SORT_QUERY_RESULTS
 -- Mask the file size values as it can have slight variability, causing test flakiness
 --! qt:replace:/("file_size_in_bytes":)\d+/$1#Masked#/
---! qt:replace:/("total-files-size":)\d+/$1#Masked#/
+--! qt:replace:/("added-files-size":")\d+/$1#Masked#/
+--! qt:replace:/("total-files-size":")\d+/$1#Masked#/
 --! qt:replace:/((ORC|PARQUET|AVRO)\s+\d+\s+)\d+/$1#Masked#/
 -- Mask iceberg version
 --! qt:replace:/("iceberg-version":")(\w+\s\w+\s\d+\.\d+\.\d+\s\(\w+\s\w+\))/$1#Masked#/
--- a/iceberg/iceberg-handler/src/test/results/positive/query_iceberg_metadata_of_partitioned_table.q.out
+++ b/iceberg/iceberg-handler/src/test/results/positive/query_iceberg_metadata_of_partitioned_table.q.out
@@ -258,7 +258,7 @@ POSTHOOK: query: select summary from default.ice_meta_3.snapshots
 POSTHOOK: type: QUERY
 POSTHOOK: Input: default@ice_meta_3
 POSTHOOK: Output: hdfs://### HDFS PATH ###
-{"added-data-files":"7","added-records":"15","added-files-size":"1378","changed-partition-count":"7","total-records":"15","total-files-size":"1378","total-data-files":"7","total-delete-files":"0","total-position-deletes":"0","total-equality-deletes":"0","iceberg-version":"#Masked#"}
+{"added-data-files":"7","added-records":"15","added-files-size":"#Masked#","changed-partition-count":"7","total-records":"15","total-files-size":"#Masked#","total-data-files":"7","total-delete-files":"0","total-position-deletes":"0","total-equality-deletes":"0","iceberg-version":"#Masked#"}
 PREHOOK: query: select summary['changed-partition-count'] from default.ice_meta_2.snapshots
 PREHOOK: type: QUERY
 PREHOOK: Input: default@ice_meta_2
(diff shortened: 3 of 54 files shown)


itests/util/src/main/java/org/apache/hadoop/hive/ql/QOutProcessor.java
```diff
@@ -353,10 +354,17 @@ private ArrayList<Pair<Pattern, String>> initPatternWithMaskComments() {
       {
         add(toPatternPair("(pblob|s3.?|swift|wasb.?).*hive-staging.*",
             "### BLOBSTORE_STAGING_PATH ###"));
-        add(toPatternPair(PATH_HDFS_WITH_DATE_USER_GROUP_REGEX, String.format("%s %s$3$4 %s $6%s",
-            HDFS_USER_MASK, HDFS_GROUP_MASK, HDFS_DATE_MASK, HDFS_MASK)));
+        add(toPatternPair(PATH_HDFS_WITH_DATE_USER_GROUP_REGEX, String.format("%s %s$3%s %s $6%s",
+            HDFS_USER_MASK, HDFS_GROUP_MASK, HDFS_SIZE_MASK, HDFS_DATE_MASK, HDFS_MASK)));
         add(toPatternPair(PATH_HDFS_REGEX, String.format("$1%s", HDFS_MASK)));
         add(toPatternPair("(.*totalSize\\s*=*\\s*)\\d+\\s*(.*)", "$1#Masked#$2"));
+        add(toPatternPair("(.*HDFS_BYTES_READ:*\\s*)\\d+\\s*(.*)", "$1#Masked#$2"));
```
deniskuzZ, line 361: i am not sure we should mask HDFS_BYTES_READ, CACHE_HIT_BYTES, CACHE_MISS_BYTES completely. We might miss issues in the future because of that

kokila-19: Considering if we should not mask it due to possibility of not detecting the issues, then we can only overwrite this for upgrades.

deniskuzZ: that's ok

kokila-19: I have updated the PR, please review it.
Thanks

ql/src/test/queries/clientpositive/explainanalyze_acid_with_direct_insert.q
```diff
@@ -3,6 +3,7 @@ set hive.txn.manager=org.apache.hadoop.hive.ql.lockmgr.DbTxnManager;
 set hive.acid.direct.insert.enabled=true;
 
 -- SORT_QUERY_RESULTS
+-- MASK_DATA_SIZE
```
okumin, line 6: Do we need this one?

okumin: Please ignore. I didn't know the feature

ql/src/test/results/clientpositive/llap/explainanalyze_acid_with_direct_insert.q.out
```diff
@@ -47,21 +47,21 @@ STAGE PLANS:
                 TableScan
                   alias: _dummy_table
                   Row Limit Per Split: 1
-                  Statistics: Num rows: 1/1 Data size: 10 Basic stats: COMPLETE Column stats: COMPLETE
+                  Statistics: Num rows: 1/1 Data size: ###Masked### Basic stats: COMPLETE Column stats: COMPLETE
```
okumin, line 50: I guess this means the data size changes when we use JDK 17. Do we know why? In my rough impression, we will likely get the same number, as the data size stands for raw data size = memory size.

hbase-handler/src/test/results/negative/cascade_dbdrop.q.out
```diff
@@ -19,8 +19,8 @@ POSTHOOK: type: CREATETABLE
 POSTHOOK: Output: database:hbasedb
 POSTHOOK: Output: hbaseDB@hbase_table_0
 Found 2 items
-drwxr-xr-x   - ### USER ### ### GROUP ###          0 ### HDFS DATE ### hdfs://### HDFS PATH ###
-drwxr-xr-x   - ### USER ### ### GROUP ###          0 ### HDFS DATE ### hdfs://### HDFS PATH ###
+drwxr-xr-x   - ### USER ### ### GROUP ###          ### SIZE ### ### HDFS DATE ### hdfs://### HDFS PATH ###
```
deniskuzZ, line 22: @kokila-19 can we not mask 0 size?

itests/util/src/main/java/org/apache/hadoop/hive/ql/QOutProcessor.java
```diff
@@ -67,6 +68,9 @@ public class QOutProcessor {
   private static final PatternReplacementPair MASK_LINEAGE = new PatternReplacementPair(
       Pattern.compile("POSTHOOK: Lineage: .*"),
       "POSTHOOK: Lineage: ###Masked###");
+
+  public static final String MASK_1 = "$1#Masked#";
```
deniskuzZ, line 72: can we use some meaningful names like: DATA_SIZE_MASK & FILE_SIZE_MASK?

itests/util/src/main/java/org/apache/hadoop/hive/ql/QOutProcessor.java
```diff
@@ -353,10 +357,14 @@ private ArrayList<Pair<Pattern, String>> initPatternWithMaskComments() {
...
             "### BLOBSTORE_STAGING_PATH ###"));
-        add(toPatternPair(PATH_HDFS_WITH_DATE_USER_GROUP_REGEX, String.format("%s %s$3$4 %s $6%s",
-            HDFS_USER_MASK, HDFS_GROUP_MASK, HDFS_DATE_MASK, HDFS_MASK)));
+        add(toPatternPair(PATH_HDFS_WITH_DATE_USER_GROUP_REGEX, String.format("%s %s$3%s %s $6%s",
+            HDFS_USER_MASK, HDFS_GROUP_MASK, HDFS_SIZE_MASK, HDFS_DATE_MASK, HDFS_MASK)));
         add(toPatternPair(PATH_HDFS_REGEX, String.format("$1%s", HDFS_MASK)));
-        add(toPatternPair("(.*totalSize\\s*=*\\s*)\\d+\\s*(.*)", "$1#Masked#$2"));
+        add(toPatternPair("(.*totalSize\\s*=*\\s*)\\d+\\s*(.*)", MASK_2));
+        add(toPatternPair("(.*File raw data size:*\\s*)\\d+\\s*bytes(.*)", MASK_2));
+        add(toPatternPair("(totalFileSize:\\s*)\\d+", MASK_1));
```
deniskuzZ, line 365: can we combine multiple patterns with or? ((total|max|min)FileSize:\\s*)\\d+

ql/src/test/results/clientpositive/llap/alter_table_location.q.out
```diff
@@ -24,9 +24,9 @@ columns:struct columns { i32 a, string b}
 partitioned:false
 partitionColumns:
 totalNumberFiles:0
-totalFileSize:0
-maxFileSize:0
-minFileSize:0
+totalFileSize:#Masked#
```
deniskuzZ, line 27: why mask 0?

ql/src/test/results/clientpositive/llap/exim_00_nonpart_empty.q.out
```diff
@@ -71,9 +71,9 @@ columns:struct columns { i32 dep_id}
 partitioned:false
 partitionColumns:
 totalNumberFiles:0
-totalFileSize:0
-maxFileSize:0
-minFileSize:0
+totalFileSize:#Masked#
```
deniskuzZ, line 74: same

ql/src/test/results/clientpositive/llap/exim_04_evolved_parts.q.out
```diff
@@ -184,9 +184,9 @@ columns:struct columns { i32 emp_id, string emp_name, string emp_dob, string emp
 partitioned:true
 partitionColumns:struct partition_columns { string emp_country, string emp_state}
 totalNumberFiles:0
-totalFileSize:0
-maxFileSize:0
-minFileSize:0
+totalFileSize:#Masked#
```
deniskuzZ, line 187: same

ql/src/test/results/clientpositive/llap/explainanalyze_acid_with_direct_insert.q.out
```diff
@@ -47,21 +47,21 @@ STAGE PLANS:
...
                   Select Operator
                     expressions: array(const struct(1),const struct(2),const struct(3),const struct(4)) (type: array<struct<col1:int>>)
                     outputColumnNames: _col0
-                    Statistics: Num rows: 1/1 Data size: 72 Basic stats: COMPLETE Column stats: COMPLETE
+                    Statistics: Num rows: 1/1 Data size: ###Masked### Basic stats: COMPLETE Column stats: COMPLETE
```
deniskuzZ, line 54: I don't think we should mask stats value, that could lead to hidden issues, better to update the sizes

ql/src/test/results/clientpositive/llap/show_tables.q.out
```diff
@@ -358,9 +358,9 @@ columns:struct columns { i32 a}
 partitioned:false
 partitionColumns:
 totalNumberFiles:0
-totalFileSize:0
-maxFileSize:0
-minFileSize:0
+totalFileSize:#Masked#
```
deniskuzZ, line 361: why mask 0

ql/src/test/results/clientpositive/llap/explainanalyze_acid_with_direct_insert.q.out
deniskuzZ: orc version was upgraded

kokila-19: True, ORC 1.9.x supports both JDK 8 and JDK 17.

**Background:**
Previously, Hive was using Hadoop 3.3.6, while ORC 1.9.5, when compiled with JDK 17, was bringing in hadoop-client-api 3.3.5 (as defined [here](https://github.com/apache/orc/blob/c6904de5310ca8e6bcf77ca6fe2fc2335624e774/java/shims/pom.xml#L96)).

This mismatch caused version conflicts, particularly with Hadoop shaded class paths, which led to hadoop-client-api being excluded from the build to avoid runtime errors. However, this dependency might still be required at runtime by orc-shims to access Hadoop APIs.

Now that Hive has upgraded Hadoop to 3.4.1, it makes sense to align the ORC version as well. Starting from ORC 2.0.3, ORC supports Hadoop 3.4.1.

Additionally, since we are upgrading to JDK 17, adopting ORC 2.0.3+ will help avoid runtime compatibility issues.

okumin: Thanks for presenting the background. It's something I should be aware of, and I appreciate your dedicated work.

kokila-19:
@deniskuzZ @okumin Thanks for the comments and I have addressed them.

itests/util/src/main/java/org/apache/hadoop/hive/ql/QOutProcessor.java
```diff
@@ -67,6 +71,9 @@ public class QOutProcessor {
   private static final PatternReplacementPair MASK_LINEAGE = new PatternReplacementPair(
       Pattern.compile("POSTHOOK: Lineage: .*"),
       "POSTHOOK: Lineage: ###Masked###");
+
+  public static final String FILE_SIZE_MASK = "$1#Masked#";
+  public static final String MASK_SIZE_PATTERN = "$1#Masked#$2";
```
deniskuzZ, line 76: `DATA_SIZE_MASK`?

ql/src/test/results/clientpositive/llap/explainanalyze_acid_with_direct_insert.q.out
deniskuzZ: @kokila-19, can we fallback to ZLIB with some Hive/ORC configs?

itests/util/src/main/java/org/apache/hadoop/hive/ql/QOutProcessor.java
```diff
@@ -44,13 +44,17 @@
 public class QOutProcessor {
   public static final String PATH_HDFS_REGEX = "(hdfs://)([a-zA-Z0-9:/_\\-\\.=])+";
   public static final String PATH_HDFS_WITH_DATE_USER_GROUP_REGEX =
-      "([a-z]+) ([a-z]+)([ ]+)([0-9]+) ([0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}) "
+      "([a-z]+) ([a-z]+)([ ]+)((?!0\\s)[0-9]+) ([0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}) "
           + PATH_HDFS_REGEX;
+  public static final String PATH_HDFS_WITH_DATE_USER_GROUP_ZERO_SIZE_REGEX =
```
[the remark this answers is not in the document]
kokila-19: Because of the way we add mask with patterns
```
add(toPatternPair(PATH_HDFS_WITH_DATE_USER_GROUP_REGEX, String.format("%s %s$3%s %s $6%s",
    HDFS_USER_MASK, HDFS_GROUP_MASK, HDFS_SIZE_MASK, HDFS_DATE_MASK, HDFS_MASK)));
```
Here, we can't distinguish between 0 and non-zero values to decide whether to use HDFS_SIZE_MASK or not. So , I had to add another regex to make it work

ql/src/test/results/clientpositive/llap/explainanalyze_acid_with_direct_insert.q.out
simhadri-g: I too think its better to correct the data size. 
* We can infer quite a bit of info from the actual value and catch issues more easily(such as, if predicate push down is resulting in less data being read?)
* Commit history will also be preserved which will allow us to  track which commit modified the data size (Like the orc upgrade mentioned in previous comment by okumin)

ql/src/test/results/clientpositive/llap/explainanalyze_acid_with_direct_insert.q.out
kokila-19: We can configure it using orc.compress, but that's a table-level property in Hive. I'll look into setting a global default and where it should be defined.
If this can be done, I will take up testing ZSTD separately.

okumin (approved):
I checked all files and gave this +1!

PR #5804, opened 2025-05-06
HIVE-28909: Modify McolumnDescriptor definition in package.jdo to sup…

saihemanth-cloudera:
…port JDK17 + JDO6.0

For Hive to support JDK17 + JDO6.0, select/update/delete queries of the table are failing to fetch column information of the table. The DataNucleus6.0 is unable to map the Mcolumn Descriptor definition to the underlying table schema due to the presence of CLOB type field for the join condition in the package.jdo file. This patch addresses the issue by avoiding the CLOB column in the join condition.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
@@ -21,6 +21,7 @@
  */
 package org.apache.hadoop.hive.metastore.model;
 
+import javax.jdo.annotations.NotPersistent;
 import java.util.ArrayList;
 import java.util.stream.Collectors;
 import java.util.List;
@@ -31,29 +32,36 @@ import java.util.List;
  * A wrapper around a list of columns.
  */
 public class MColumnDescriptor {
-  private List<MColumn> columns;
+  private List<MColumn> fields;
+
+  @NotPersistent
+  private List<MFieldSchema> columns;
   private long id;
 
   public MColumnDescriptor() {}
 
   public MColumnDescriptor(List<MFieldSchema> cols) {
-    columns = cols.stream().map(schema ->
+    fields = cols.stream().map(schema ->
         new MColumn(schema.getName(), schema.getType(), schema.getComment()))
             .collect(Collectors.toList());
+    columns = new ArrayList<>();
   }
 
   public List<MColumn> getFields() {
-    return columns;
+    return fields;
   }
 
   public List<MFieldSchema> getCols() {
-    List<MFieldSchema> schemas = new ArrayList<>();
-    if (getFields() != null) {
-      schemas = getFields().stream().map(column ->
-          new MFieldSchema(column.getName(), column.getType(), column.getComment()))
-              .collect(Collectors.toList());
+    if (!columns.isEmpty()) {
+      return columns;
     }
-    return schemas;
+    columns = new ArrayList<>();
+    if (fields != null) {
+      columns = fields.stream().map(column ->
+              new MFieldSchema(column.getName(), column.getType(), column.getComment()))
+          .collect(Collectors.toList());
+    }
+    return columns;
   }
 
   public long getId() {
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
@@ -21,6 +21,7 @@
  */
 package org.apache.hadoop.hive.metastore.model;
 
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -29,23 +30,37 @@ import java.util.List;
  * A wrapper around a list of columns.
  */
 public class MColumnDescriptor {
-  private List<MFieldSchema> cols;
+  private List<MColumn> columns;
+  private long id;
 
   public MColumnDescriptor() {}
 
-  /**
-   *
-   * @param cols
-   */
   public MColumnDescriptor(List<MFieldSchema> cols) {
-    this.cols = cols;
+    columns = new ArrayList<>(cols.size());
+    for (MFieldSchema schema : cols) {
+      columns.add(new MColumn(schema.getName(), schema.getType(), schema.getComment()));
+    }
+  }
+
+  public List<MColumn> getColumns() {
+    return columns;
+  }
+
+  public void setColumns(List<MColumn> columns) {
+    this.columns = columns;
   }
 
   public List<MFieldSchema> getCols() {
-    return cols;
+    List<MFieldSchema> schemas = new ArrayList<>();
+    if (getColumns() != null) {
+      for (MColumn column : getColumns()) {
+        schemas.add(new MFieldSchema(column.getName(), column.getType(), column.getComment()));
+      }
+    }
+    return schemas;
   }
 
-  public void setCols(List<MFieldSchema> cols) {
-    this.cols = cols;
+  public long getId() {
+    return id;
   }
 }
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
@@ -44,7 +44,6 @@ public class MColumnDescriptor {
     fields = cols.stream().map(schema ->
         new MColumn(schema.getName(), schema.getType(), schema.getComment()))
             .collect(Collectors.toList());
-    columns = new ArrayList<>();
   }
 
   public List<MColumn> getFields() {
@@ -52,7 +51,7 @@ public class MColumnDescriptor {
   }
 
   public List<MFieldSchema> getCols() {
-    if (!columns.isEmpty()) {
+    if (columns != null) {
       return columns;
     }
     columns = new ArrayList<>();
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
@@ -22,6 +22,7 @@
 package org.apache.hadoop.hive.metastore.model;
 
 import java.util.ArrayList;
+import java.util.stream.Collectors;
 import java.util.List;
 
 /**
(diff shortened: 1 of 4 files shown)


dengzhhu653 (approved):
+1 pending tests

standalone-metastore/metastore-server/src/main/resources/datanucleus-log4j.properties
```diff
@@ -30,5 +30,5 @@ log4j.category.DataNucleus.Transaction=INFO, A1
 log4j.category.DataNucleus.Datastore=DEBUG, A1
 log4j.category.DataNucleus.ValueGeneration=DEBUG, A1
 
-log4j.category.DataNucleus.Enhancer=INFO, A1
+log4j.category.DataNucleus.Enhancer=DEBUG, A1
```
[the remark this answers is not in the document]
saihemanth-cloudera: Not required. Changing it back.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
```diff
@@ -29,23 +30,37 @@
...
+    for (MFieldSchema schema : cols) {
+      columns.add(new MColumn(schema.getName(), schema.getType(), schema.getComment()));
+    }
+  }
+
+  public List<MColumn> getColumns() {
+    return columns;
+  }
+
+  public void setColumns(List<MColumn> columns) {
```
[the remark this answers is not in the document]
saihemanth-cloudera: We don't need this

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
```diff
@@ -29,23 +30,33 @@
...
+  }
+
+  public List<MColumn> getColumns() {
+    return columns;
   }
 
   public List<MFieldSchema> getCols() {
-    return cols;
+    List<MFieldSchema> schemas = new ArrayList<>();
+    if (getColumns() != null) {
```
deniskuzZ, line 51: Wonder why we need the double-conversion? Why not store the `List<MFieldSchema> cols` passed through the constructor?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
```diff
@@ -29,23 +30,33 @@
...
   public MColumnDescriptor() {}
 
-  /**
-   *
-   * @param cols
-   */
   public MColumnDescriptor(List<MFieldSchema> cols) {
-    this.cols = cols;
+    columns = new ArrayList<>(cols.size());
+    for (MFieldSchema schema : cols) {
```
deniskuzZ, line 40: minor: why not use stream API?
````
columns = cols.stream()
  .map(schema -> new MColumn(schema.getName(), schema.getType(), schema.getComment())
  .collect(Collectors.toList())
````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java [hunk 5]
```diff
@@ -29,23 +30,33 @@
...
+      columns.add(new MColumn(schema.getName(), schema.getType(), schema.getComment()));
+    }
+  }
+
+  public List<MColumn> getColumns() {
```
deniskuzZ, line 45: Could we come up with a better naming to distinguish 2: `getColumns()` & `getCols()`?
keep `getCols()`, but rename `getColumns()` to `getFields` or similar

standalone-metastore/metastore-server/src/main/resources/package.jdo
```diff
@@ -364,32 +380,13 @@
...
-            <field name="type">
-              <column name="TYPE_NAME" length="32672" jdbc-type="VARCHAR" allows-null="false"/>
-            </field>
-            <field name="comment">
-              <column name="COMMENT" length="256" jdbc-type="VARCHAR" allows-null="true"/>
-            </field>
-          </embedded>
-        </element>
+    <class name="MColumnDescriptor" identity-type="application" table="CDS" detachable="true">
+      <field name="id" primary-key="true" value-strategy="native">
```
okumin, line 384: The native strategy sounds good, though I'm not a specialist of JDO.
https://www.datanucleus.org/products/accessplatform_4_2/jdo/value_generation.html

standalone-metastore/metastore-server/src/main/resources/package.jdo
```diff
@@ -364,32 +380,13 @@
...
+        <column name="CD_ID" jdbc-type="BIGINT" />
+      </field>
+      <field name="columns" mapped-by="cd">
+        <order column="INTEGER_IDX"/>
+        <collection element-type="MColumn" dependent-element="true" />
```
okumin, line 389: I referred to this document to know how JDO with collections works.
https://www.datanucleus.org/products/accessplatform_4_1/jdo/orm/one_to_many_collection.html

standalone-metastore/metastore-server/src/main/resources/package.jdo
```diff
@@ -115,6 +115,22 @@
       </field>
     </class>
 
+    <class name="MColumn" identity-type="application" table="COLUMNS_V2" detachable="true" objectid-class="MColumn$PK">
```
okumin, line 118: I understand we need `objectid-class` because the PK is composed of two columns.
https://www.datanucleus.org/products/accessplatform_4_2/jdo/application_identity.html#primarykey

okumin (commented):
This pull request is probably good, but this is the first time for me to see JDO

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
saihemanth-cloudera: JDK17 doesn't understand MfieldSchema because of the CLOB column in the join statement. So now need two representations 1) MfieldSchema in the application layer 2) Mcolumn in the JDO layer for conversion to/from the database.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
saihemanth-cloudera: ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
saihemanth-cloudera: ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
deniskuzZ: i understand that, but why do you need to reconstruct the `getCols() ` representation
{code}
  private List<MColumn> fields;
  private long id;

  private List<MFieldSchema> cols;

  public MColumnDescriptor(List<MFieldSchema> cols) {
    this.cols = cols;
    
    this.fields = cols.stream().map(schema ->
        new MColumn(schema.getName(), schema.getType(), schema.getComment()))
  }

  public List<MColumn> getFields() {
    return fields;
  }

  public List<MFieldSchema> getCols() {
     return cols
  }
{code}

deniskuzZ: I understand this, but why do we need to reconstruct the diff representations?
````
public class MColumnDescriptor {

  private List<MFieldSchema> cols;
  
  private List<MColumn> fields;
  private long id;

  public MColumnDescriptor() {}

  public MColumnDescriptor(List<MFieldSchema> cols) {
    this.cols = cols;
    fields = cols.stream().map(schema ->
        new MColumn(schema.getName(), schema.getType(), schema.getComment()))
  }

  public List<MColumn> getFields() {
    return fields;
  }

  public List<MFieldSchema> getCols() {
    if (cols != null) {
      return cols;
    }
    cols = new ArrayList<>();
    if (fields != null) {
      cols = getFields().stream().map(column ->
          new MFieldSchema(column.getName(), column.getType(), column.getComment()))
    }
    return cols;
  }
````

Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java]: Maintaining a single `fields` looks enough, this `public MColumnDescriptor(List<MFieldSchema> cols)` is called only from the HMS side, and the DataNucleus serializes only the `fields`.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java]: @dengzhhu653, but the getCols() is executed in the loop in application code leading to unnessesary repetitive fields transform to List<MFieldSchema>

Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java]: Per my understanding, the `fields` will be filled when DataNucleus reads from db(or on calling `getFields`).

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java]: my concern here is not about `fields` but their transformed representation returned by `getCols()`

File: standalone-metastore/metastore-server/src/main/resources/package.jdo [hunk 9]
```diff
@@ -384,10 +384,13 @@
       <field name="id" primary-key="true" value-strategy="native">
         <column name="CD_ID" jdbc-type="BIGINT" />
       </field>
-      <field name="columns" mapped-by="cd">
+      <field name="fields" mapped-by="cd">
         <order column="INTEGER_IDX"/>
         <collection element-type="MColumn" dependent-element="true" />
       </field>
+      <field name="columns" persistence-modifier="none">
```
deniskuzZ, line 391: @saihemanth-cloudera, why do we need this?

saihemanth-cloudera: All the fields declared in JDO are persistent by default. i.e.., Data Nucleus assumes that there is mapping of this field to a column in back end db (here it assumes that there is 1-1 mapping of columns field to a column in CDS table). It explicitly say that this field is only used in application layer and that there is no such column in the DB, we need to declare it as persistence-modifier="none".
`@NotPersistant` annotation mentioned in Java is not taking into effect unless we mention it in `.jdo` class. So I left the annotation for future reference to the developers.

saihemanth-cloudera: without `default-fetch-group="false"` DataNucleus tries to read this field and it fails, so I had to include it.

deniskuzZ: weird, see https://www.datanucleus.org/products/accessplatform_4_1/jdo/fields_properties.html

Aggarwal-Raghav: @deniskuzZ  @okumin nit: better to check the doc https://www.datanucleus.org/products/accessplatform_6_0/ i.e 6_0 instead of 4_1 which is for datanucleus 4.x. Hoping the most of the things are same in docs, haven't verified myself :-)

deniskuzZ: same thing
https://www.datanucleus.org/products/accessplatform_6_0/jdo/mapping.html

> Making a field/property non-persistent

If you have a field/property that you don’t want to persist, just mark it’s persistence-modifier as none, like this

````
<class name="mydomain.MyClass">
    <field name="unimportantField" persistence-modifier="none"/>
</class>
````

Comment (saihemanth-cloudera) [reply standalone-metastore/metastore-server/src/main/resources/package.jdo]: tested and verified. It works.

File: standalone-metastore/metastore-server/src/main/resources/package.jdo [hunk 10]
```diff
@@ -364,33 +380,15 @@
...
-        </element>
+    <class name="MColumnDescriptor" identity-type="application" table="CDS" detachable="true">
+      <field name="id" primary-key="true" value-strategy="native">
+        <column name="CD_ID" jdbc-type="BIGINT" />
+      </field>
+      <field name="fields" mapped-by="cd">
+        <order column="INTEGER_IDX"/>
+        <collection element-type="MColumn" dependent-element="true" />
       </field>
+      <field name="columns" persistence-modifier="none"></field>
```
deniskuzZ, line 391: nit: why not use a self-closing tag as in the example?

PR #5811, opened 2025-05-12
HIVE-28949: TestDbNotificationListener#createDatabase fails intermittently due to different createTime

zabetak:
Ignore user defined database creation time from equality comparison.

Avoid intermittent test failures and stabilize CI.

The test is problematic cause it sets explicitly the creation time and it expects the same to be returned in the notification message. However, the HMS always overrides the `createTime` (expected behavior according to tests added by HIVE-21077) field during the creation of the database so the initial value is lost. Slight delays (more than 1 second) between the initialization of the test and the creation of the database are causing the test to fail. The test is more likely to fail in CI where delays are more common.

No

Adding delays (e.g., `Thread.sleep(2000)`) before `ms.createDatabase` call and running the test.

### DIFF
--- a/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java
+++ b/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java
@@ -344,7 +344,6 @@ public class TestDbNotificationListener
     String dbDescription = "no description";
     Database db = new Database(dbName, dbDescription, dbLocationUri, emptyParameters);
     db.setOwnerName("test_user");
-    db.setCreateTime(startTime);
     msClient.createDatabase(db);
 
     // Read notification from metastore
@@ -362,7 +361,7 @@ public class TestDbNotificationListener
     // Parse the message field
     CreateDatabaseMessage createDbMsg = md.getCreateDatabaseMessage(event.getMessage());
     assertEquals(dbName, createDbMsg.getDB());
-    assertEquals(db, createDbMsg.getDatabaseObject());
+    assertEqualsSansTime(db, createDbMsg.getDatabaseObject());
 
     // Verify the eventID was passed to the non-transactional listener
     MockMetaStoreEventListener.popAndVerifyLastEventId(EventType.CREATE_DATABASE, firstEventId + 1);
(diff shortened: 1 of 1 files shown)


okumin (approved):
I confirmed the issue would happen on the master branch.

```diff
% git diff             
diff --git a/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java b/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java
index de37e9f270..f5685c3bae 100644
--- a/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java
+++ b/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java
@@ -345,6 +345,7 @@ public void createDatabase() throws Exception {
     Database db = new Database(dbName, dbDescription, dbLocationUri, emptyParameters);
     db.setOwnerName("test_user");
     db.setCreateTime(startTime);
+    Thread.sleep(2000);
     msClient.createDatabase(db);
 
     // Read notification from metastore
```

The tail of outputs.

```
java.lang.AssertionError: 
Expected :Database(name:createdb, description:no description, locationUri:..., parameters:{}, ownerName:test_user, catalogName:hive, createTime:1747207149)
Actual   :Database(name:createdb, description:no description, locationUri:..., parameters:{}, ownerName:test_user, catalogName:hive, createTime:1747207151
```

With this patch, the test succeeds regardless of existence the pause.

```diff
% git diff
diff --cc itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java
index ae588d8b93,f5685c3bae..0000000000
--- a/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java
+++ b/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/listener/TestDbNotificationListener.java
@@ -344,6 +344,7 @@ public void createDatabase() throws Exception {
     String dbDescription = "no description";
     Database db = new Database(dbName, dbDescription, dbLocationUri, emptyParameters);
     db.setOwnerName("test_user");
+    Thread.sleep(2000);
     msClient.createDatabase(db);
 
     // Read notification from metastore
```

zabetak:
Thanks for the review @okumin !

PR #5812, opened 2025-05-13
HIVE-28953: Remove HiveMetaStoreClientPreCatalog

dengzhhu653:
HiveMetaStoreClient is aware of the default catalog set in the configuration, making HiveMetaStoreClientPreCatalog unnecessary in the test, and each time we introduce a new method in IMetaStoreClient, we have to implement the body in HiveMetaStoreClientPreCatalog, which is meaningless.

No


dengzhhu653:
> Can we confirm that `HiveMetaStoreClientPreCatalog `will not be used by any other downstream open-source components?
> 
> IMO, `HiveMetaStoreClientPreCatalog` to be of no use, so I think it can be safely deleted.

This class resides in our test package, which I think no downstream project should use this in their base, will check Impala and Iceberg.

okumin:
I assume the class is really not needed.

I'd like to remove the name also from the comment in HiveMetaStore.java.

```sh
% git grep HiveMetaStoreClientPreCatalog
standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java:   * This method is called indirectly by HiveMetastoreClient and HiveMetaStoreClientPreCatalog
standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java:   * change in HiveMetastoreClient and HiveMetaStoreClientPreCatalog.
standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientPreCatalog.java:public class HiveMetaStoreClientPreCatalog implements IMetaStoreClient, AutoCloseable {
standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientPreCatalog.java:  static final protected Logger LOG = LoggerFactory.getLogger(HiveMetaStoreClientPreCatalog.class);
standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientPreCatalog.java:  public HiveMetaStoreClientPreCatalog(Configuration conf) throws MetaException {
standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientPreCatalog.java:  public HiveMetaStoreClientPreCatalog(Configuration conf, HiveMetaHookLoader hookLoader) throws MetaException {
standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientPreCatalog.java:  public HiveMetaStoreClientPreCatalog(Configuration conf, HiveMetaHookLoader hookLoader, Boolean allowEmbedded)
standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientPreCatalog.java:      HiveMetaStoreClientPreCatalog.class.getClassLoader(),
standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestCatalogNonDefaultSvr.java:    return new HiveMetaStoreClientPreCatalog(conf);
standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestCatalogOldClient.java:    return new HiveMetaStoreClientPreCatalog(conf);
```

dengzhhu653:
I checked the master/main branch of Iceberg and Impala, they are not using this class.

zhangbutao (approved):
+1 LGTM

okumin (approved):
+1

PR #5813, opened 2025-05-14
HIVE-28789: Session-level metrics from Tez sessions

abstractdog:
The proposal aims to implement a mechanism in HS2 to retrieve metrics from Tez sessions, with the actual implementation abstracted at the TezSession/TezSessionState level.

The code contains:
1. core logic: TezSessionPoolManagerMetrics.java + unit test
2. DAGStatusObserver: interface by which a tez session object is notified by the DAG status changes, hence it's able to provide metrics accumulated from the DAG status: this is important to achieve the minimum viable product (even if in the long term, the sessions might be directly queried through the unmanaged Tez AMs)
3. session-related objects to be adapted to the DAG state -> metrics flow

A new interface on the Tez session (currently TezSessionState) makes it clear that a session might be able to expose metrics about itself (e.g. task metrics), which could be adapted to the managed Yarn sessions as well as the more modern, unmanaged Tez sessions ([TEZ-3991](https://issues.apache.org/jira/browse/TEZ-3991)).

No.

1. Manual testing: tested on a Cloudera Data Warehouse, where the corresponding implementation of an external tez session was present. The expected outcome was an aggregated value of tez session metrics when queried from the HS2 metrics endpoint.
4. Unit testing: unit test for the core logis is added to the PR.


ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManagerMetrics.java
```diff
@@ -0,0 +1,259 @@
...
+  private final TaskBacklogRatioMetric[] taskBacklogRatioDurations = Arrays.stream(TRACKED_TASK_BACKLOG_RATIOS)
+      .map(
+          ratio -> new TaskBacklogRatioMetric(String.format(TEZ_SESSION_METRIC_TASK_BACKLOG_RATIO_PRESENT_SINCE, ratio),
+              ratio))
+      .toArray(TaskBacklogRatioMetric[]::new);
+
+
+  public TezSessionPoolManagerMetrics(TezSessionPoolManager poolManager) {
+    this.poolManager = poolManager;
+    this.metrics = MetricsFactory.getInstance();
```
ayushtkn, line 84: should we `init` first?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManagerMetrics.java
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  public TezSessionPoolManagerMetrics start(HiveConf conf) {
+    if (metrics == null) {
+      LOG.warn(
+          "Metrics are not enabled, cannot start TezSessionPoolManagerMetrics (check {})",
+          HiveConf.ConfVars.HIVE_SERVER2_METRICS_ENABLED.varname);
+      return this;
+    }
+    this.mainExecutor = Executors.newSingleThreadScheduledExecutor(
```
ayushtkn, line 108: just thinking should this be `daemon`?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManagerMetrics.java
```diff
@@ -0,0 +1,259 @@
...
+    for (TaskBacklogRatioMetric m : taskBacklogRatioDurations) {
+      m.createGauge(metrics);
+    }
+    clearMetrics(); //initial empty values for gauges
+    return this;
+  }
+
+  public void stop() {
+    if (this.mainExecutor != null) {
+      mainExecutor.shutdownNow();
```
ayushtkn, line 127: should this be ``shutdown`` or you want ``shutdownNow()`` only, moreover should we set `mainExecutor`=`null` or a check that it is already not shutdown or terminated. in case someone does stop twice or stop post shutdown

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManagerMetrics.java
```diff
@@ -0,0 +1,259 @@
...
+      return;
+    }
+
+    AtomicDouble currentRunningTasksCount = new AtomicDouble();
+    AtomicDouble currentPendingTasksCount = new AtomicDouble();
+    List<Future<?>> collectTasks = new ArrayList<>();
+
+    // run on maximum 10 threads
+    ExecutorService collectorExecutor = Executors.newFixedThreadPool(Math.min(sessions.size(), 10),
+        new ThreadFactoryBuilder().setNameFormat("TezSessionPoolManagerMetrics collector thread - " + "#%d").build());
```
ayushtkn, line 147: you can avoid concat & directly put
```
"TezSessionPoolManagerMetrics collector thread - #%d"
```

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolSession.java
```diff
@@ -31,6 +32,7 @@
 import org.apache.hadoop.hive.ql.session.SessionState.LogHelper;
 import org.apache.hadoop.hive.registry.impl.TezAmInstance;
 import org.apache.tez.dag.api.TezException;
+import org.apache.tez.dag.api.client.DAGStatus;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
```
ayushtkn, line 38: unrelated changes

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -4093,6 +4093,9 @@ public static enum ConfVars {
         "This flag is used in HiveServer2 to enable a user to use HiveServer2 without\n" +
         "turning on Tez for HiveServer2. The user could potentially want to run queries\n" +
         "over Tez without the pool of sessions."),
+    HIVE_SERVER2_TEZ_SESSIONS_METRICS_COLLECTION_INTERVAL_SECONDS(
+        "hive.server2.tez.sessions.metrics.collection.interval.seconds", 10,
+        "Interval for collecting metrics from Tez sessions."),
```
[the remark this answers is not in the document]
abstractdog: makes sense, I'll do

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DAGStatusObserver.java
```diff
@@ -0,0 +1,24 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package org.apache.hadoop.hive.ql.exec.tez;
+
+import org.apache.tez.dag.api.client.DAGStatus;
+
+public interface DAGStatusObserver {
+  void update(DAGStatus dagStatus);
+}
```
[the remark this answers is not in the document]
abstractdog: do you feel it's overengineering to separate this behavior into an interface that might be reused later?
considering alternatives: if I don't use an interface here, I need to create a TezSessionState field in TezJobMonitor to be notified about the DAG state changes, would it be better here?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManagerMetrics.java
abstractdog: I don't think so, if it's not initialized by default, then it means HS2 is not configured to utilize metrics, in this case, this whole thing isn't expected to be used
this is what I handle by:
```
  public TezSessionPoolManagerMetrics start(HiveConf conf) {
    if (metrics == null) {
      LOG.warn(
          "Metrics are not enabled, cannot start TezSessionPoolManagerMetrics (check {})",
          HiveConf.ConfVars.HIVE_SERVER2_METRICS_ENABLED.varname);
      return this;
    }
```

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManagerMetrics.java
abstractdog: yes, let me do

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManagerMetrics.java
abstractdog: hm, shutdown would be better, yes, as I don't need this call to attempt to stop actively executing tasks (what's shutdownNow is for), actually, I don't care about the running tasks here on a HS2 shutdown
regarding mainExecutor=null, it makes sense to simply make stop() idempotent

abstractdog:
addressed comments in https://github.com/apache/hive/pull/5813/commits/9521e2e9753c4b7fb02fee92a7774fd5ee04bc3b

ayushtkn (approved):
Thanx @abstractdog for the update.
Changes LGTM

PR #5821, opened 2025-05-21
HIVE-28969: Use shallow clone with limited depth to reduce .git directory size

zabetak:
1. Reduce the disk space occupied by the .git directory in CI
2. Speed up checkout times

Pull-requests are now limited to 50 commits (value of depth parameter). When the number of commits exceeds this number the CI run will abort since it will not be able to merge the PR branch with master. Affected PRs can either temporarily increase the value of the depth parameter or squash some commits together.

Manually inspect output of `du` command in pipeline logs

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -226,7 +226,7 @@ jobWrappers {
             $class: 'GitSCM',
             branches: scm.branches,
             doGenerateSubmoduleConfigurations: scm.doGenerateSubmoduleConfigurations,
-            extensions: scm.extensions,
+            extensions: [ cloneOption(honorRefspec: true, depth: 50, noTags: true, shallow: true) ],
             userRemoteConfigs: scm.userRemoteConfigs + [[
               name: 'origin',
               refspec: scm.userRemoteConfigs[0].refspec+ " +refs/heads/${CHANGE_TARGET}:refs/remotes/origin/target",


zabetak:
The disk usage for .git folder before and after the changes can be found in CI logs :

In the latest PR run:
```
[2025-06-19T13:55:44.769Z] + du -h --max-depth=1
...
[2025-06-19T13:55:45.293Z] 100M	./.git
```
In the latest [master run](https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/master/2569/pipeline/355/):
```
[2025-06-22T05:34:15.764Z] + du -h --max-depth=1
...
[2025-06-22T05:34:16.807Z] 580M	./.git
```

tanishq-chugh (approved):
LGTM +1

Jenkinsfile
```diff
@@ -226,7 +226,7 @@ jobWrappers {
             $class: 'GitSCM',
             branches: scm.branches,
             doGenerateSubmoduleConfigurations: scm.doGenerateSubmoduleConfigurations,
-            extensions: scm.extensions,
+            extensions: [ cloneOption(honorRefspec: true, depth: 50, noTags: true, shallow: true) ],
```
ayushtkn, line 229: It should be good in most cases.
@zabetak I have a small question, say in some rare scenrios like where we chase major upgrades or so, like hadoop or jdk upgrade kind of places, where the commits in PR can go beyond 50.

So, in that case does my build break?
& say if I force Push to reduce back the number down to 49 will it start working, if later, I believe that should be mentioned somewhere in the PR template or so.

If nothing breaks in such a scenario, I am good with this change

zabetak: Exactly, if the PR has more than 50 commits then the build will break on checkout cause it will not be able to merge the PR branch with the master branch since it will not find a common parent. I will update the PR template accordingly.

zabetak: The PRs that need to keep more than 50 commits can also increase explicitly the depth parameter temporarily in their runs. Added this note also on the PR template.

zabetak:
Thanks everyone for the reviews and valuable feedback!

PR #5822, opened 2025-05-21
HIVE-28960: Compaction Stats updater does not collect column stats when hive.stats.autogather is true

kasakrisz:
* Decouple Compaction Stats Updater from `hive.stats.autogather` setting
* Check stats status when generating the analyze command.

`hive.stats.autogather` does not guarantee that stats are maintained by DML statements hence it is not identical with the statement
```
analyze table <table name> compute statistics for columns <column list>;
```
There are various reasons when executing a DML statement the stats become stale. Example: delete statements mark the column stats stale because the delta stats can not be calculated from the deleted records.

No.

```
mvn test -Dtest=TestCrudCompactorOnTez#testMajorCompactionUpdateMissingColumnStats -pl itests/hive-unit -Pitests
```


ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/StatsUpdater.java
```diff
@@ -81,10 +83,14 @@ public void gatherStats(HiveConf hiveConf, CompactionInfo ci, Map<String, String
                 sb.append(")");
             }
             sb.append(" compute statistics");
-            if (!conf.getBoolVar(HiveConf.ConfVars.HIVE_STATS_AUTOGATHER) && ci.isMajorCompaction()) {
-                List<String> columnList = msc.findColumnsWithStats(CompactionInfo.compactionInfoToStruct(ci));
+            if (ci.isMajorCompaction()) {
+                List<String> columnList = msc.findColumnsWithStats(CompactionInfo.compactionInfoToStruct(ci)).stream()
+                        .filter(columnName -> !StatsSetupConst.areColumnStatsUptoDate(tableProperties, columnName))
```
deniskuzZ, line 88: maybe
````
boolean isStatsUpToDate = .. `columnList.stream.allMatch(columnName -> StatsSetupConst.areColumnStatsUptoDate(tableProperties, columnName))`
````

File: ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/StatsUpdater.java [hunk 2]
```diff
@@ -81,10 +83,14 @@ public void gatherStats(HiveConf hiveConf, CompactionInfo ci, Map<String, String
...
+                        .collect(Collectors.toList());
                 if (!columnList.isEmpty()) {
                     sb.append(" for columns ").append(String.join(",", columnList));
+                } else if (StatsSetupConst.areBasicStatsUptoDate(tableProperties)) {
```
deniskuzZ, line 92: same here

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/StatsUpdater.java
kasakrisz: > Maybe use allMatch
With the filter
```
.filter(columnName -> !StatsSetupConst.areColumnStatsUptoDate(tableProperties, columnName))
```
only the columns without stats collected and updated.

> Would it work for the partitioned table ?
Thanks for mentioning this; added partition support.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/StatsUpdater.java
kasakrisz: Added partition support.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/StatsUpdater.java
deniskuzZ: > only the columns without stats collected and updated.

oh, I missed that

deniskuzZ (approved):
LGTM +1, pending tests

itests/hive-unit/src/test/java/org/apache/hadoop/hive/ql/txn/compactor/TestCrudCompactorOnTez.java
```diff
@@ -3665,4 +3670,41 @@ public void testFallbackForMergeCompactionWhenDeleteDeltaPresent() throws Except
     verify(primary.get(), times(1)).run(any());
     verify(secondary.get(), times(1)).run(any());
   }
+
+  @Test
+  public void testMajorCompactionUpdateMissingColumnStats() throws Exception {
+    executeStatementOnDriver("create table " + TABLE1 + "(a int, b varchar(128), c float) stored as orc TBLPROPERTIES " +
```
deniskuzZ, line 3676: call `drop if exists` first

PR #5823, opened 2025-05-21
HIVE-28961: Respect partition limit in alter_table_req for partitioned tables

wecharyu:
1. Calls `get_partitions_req` api to get all partitions to enable partition limit check
2. Avoid calls `get_partitions_req` if not `dataWasMoved` 

Get all partitions of a table is a heavy operation that could affect both HMS server and backend DB, `alter_table` should also check and limit the fetched partition number.

Yes, alter table cascade to large partitions may fail after this PR.

Add a unit test:
```bash
mvn test -Dtest.groups= -Dtest=org.apache.hadoop.hive.metastore.TestRemoteHiveMetaStore#testAlterTableCascadeExceedsPartitionLimits -pl :hive-standalone-metastore-server
```


Aggarwal-Raghav:
@wecharyu, thanks for the PR. I have 1 question, a bit out of context of this PR. Can we make this https://github.com/apache/hive/blob/2495898ae937de2bfd8fe72c63eec2ae905c908c/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java#L421 run in batch as done in case of rename table https://github.com/apache/hive/blob/2495898ae937de2bfd8fe72c63eec2ae905c908c/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java#L375

Aggarwal-Raghav:
Changes LGTM +1

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
```diff
@@ -352,26 +354,29 @@ public void alterTable(RawStore msdb, Warehouse wh, String catName, String dbnam
...
-            part.setTableName(newTblName);
-          }
           // Do not verify stats parameters on a partitioned table.
           msdb.alterTable(catName, dbname, name, newt, null);
           int partitionBatchSize = MetastoreConf.getIntVar(handler.getConf(),
               MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX);
           String catalogName = catName;
           // alterPartition is only for changing the partition location in the table rename
           if (dataWasMoved) {
+            PartitionsRequest req = new PartitionsRequest(newDbName, newTblName);
```
zhangbutao, line 364: > Avoid calls get_partitions_req if not dataWasMoved

Please give more context about this change.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
```diff
@@ -352,26 +354,29 @@ public void alterTable(RawStore msdb, Warehouse wh, String catName, String dbnam
...
+            req.setCatName(catName);
+            req.setMaxParts((short) -1);
+            parts = handler.get_partitions_req(req).getPartitions();
+            for (Partition part : parts) {
```
deniskuzZ, line 368: please add new line above the loop

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
```diff
@@ -408,7 +413,10 @@ public List<Void> run(List<Partition> input) throws Exception {
 
           if (runPartitionMetadataUpdate) {
             if (cascade || retainOnColRemoval) {
-              parts = msdb.getPartitions(catName, dbname, name, -1);
+              PartitionsRequest req = new PartitionsRequest(dbname, name);
+              req.setCatName(catName);
+              req.setMaxParts((short) -1);
```
deniskuzZ, line 421: @wecharyu, where do we check the partition limit? Don't we disable it with -1?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
wecharyu: When `dataWasMoved` is false, we do not need get all partitions and update the location of partitions.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
wecharyu: We use `HMSHandler#get_partitions_req(req)` to check the limit, set max as -1 will not disable the limit checking.

deniskuzZ (approved):
LGTM +1

deniskuzZ:
@wecharyu looks like tests failed with
````
Unexpected exception, expected<org.apache.hadoop.hive.metastore.api.InvalidOperationException> but was<org.apache.hadoop.hive.metastore.api.MetaException>
````
&
````
Caused by: org.apache.hadoop.hive.metastore.api.InvalidOperationException: Unable to change partition or table. Check metastore logs for detailed stack.The following columns have types incompatible with the existing columns in their respective positions :
val
````

PR #5826, opened 2025-05-22
HIVE-28968: In CI use dedicated project-local .m2 directory for storing maven artifacts

zabetak:
Previously the maven artifacts were stored under .git/.m2 directory. In terms of functionality the exact location (.git/.m2 vs .m2) does not matter. However, putting the artifacts under the .git folder can lead into confusion and false conclusions for people not very familiar with the specificities of the CI.

For instance, someone who is not gonna look into detail on what is present under the .git directory may think that the .git data for the project are huge. Using commands such as `du -h --max-depth=1`, which is also used in some places by the CI, can endorse such false impressions.

Moreover, the .m2 directory is a more common place for storing the maven artifacts in CI/CD pipelines.

No

* Ensure tests run fine
* Inspect CI logs especially those related to `du -h --max-depth=1` command

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -93,11 +93,12 @@ set -x
 export USER="`whoami`"
 export MAVEN_OPTS="-Xmx2g"
 export -n HIVE_CONF_DIR
-cp $SETTINGS .git/settings.xml
-OPTS=" -s $PWD/.git/settings.xml -B -Dtest.groups= "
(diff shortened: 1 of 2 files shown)


zhangbutao (approved):
+1 LGTM

This change makes sense.

zabetak:
Thanks for the review @zhangbutao !

PR #5012, opened 2024-01-17
WIP: pipeline tests 

akshat0395:
### What changes were proposed in this pull request?

Hive compile time and run time support for JDK 17

PR #5828, opened 2025-05-27
HIVE-28643: Fix "Error getting slot value" for Java 12 and above

vrozov:
Use `getDeclaredFields0` to try getting `slot` `Field` if `getDeclaredField` fails. Fallback to the old non deterministic behavior if `slot` field is not accessible.

Java 12 and above filters out private fields of the `Field` class causing "Error getting slot value".

No

Verified manually that the error is gone on Java 17. Using existing test for the rest.

### DIFF
--- a/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
+++ b/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java
@@ -19,6 +19,8 @@
 package org.apache.hadoop.hive.serde2.objectinspector;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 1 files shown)


kokila-19:
@vrozov 
Slot value was used as non-deterministic result would result in flaky test. Falling back to it won't be a good idea. 

There is an alternative provided for this so I have already fixed it which is present JDK 17 PR , currently in WIP which will be rebased to the main PR soon. 
https://github.com/apache/hive/pull/5012/commits/94c3bb6cb42f4ea5811ce254c66c4250a6b93433

vrozov:
@kokila-19 

> Slot value was used as non-deterministic result would result in flaky test. Falling back to it won't be a good idea.

1. Most of the time there will be no fallback. `Field.class.getDeclaredField(name)` will work in Java 8 and `getDeclaredFields0.invoke` will work in Java 12 and above.
2. The fallback is provided as the last resort option to the old not deterministic behavior.

> Using the MethodHandles privateLookupIn , slot can be accessed.

It requires JDK 9 and above. Do you suggest that support for JDK 8 is completely dropped in 4.1?

vrozov:
> Yes. We can't support JDK8 and JDK17 together.

It is possible to compile to Java 8 byte code and test both with Java 8 and any other version of Java including 17. IMO, dropping support for Java 8 should be done as part of major version upgrade.

> one example is "--add-opens" which won't be recognized by JDK8 because it does not know such an option.

The option applies only at runtime and may be provided for Java 17 only if necessary

> Also, most of the dependencies that are upgraded use jdk17 specifically.

Which dependencies? All current dependencies are Java 8 compatible.

deniskuzZ:
We are dropping support for jdk8 in `Hive-4.1`. jdk17 would be required for the compilation.

ayushtkn:
We are on JDK-21 now. @vrozov can you check if the problem still persists, if not lets close this

PR #5830, opened 2025-05-29
HIVE-28973: Metastore service in Docker fails to start unless --verbose is explicitly set

kokila-19:
…se is explicitly set

This commit avoids appending the --verbose flag when it's not explicitly set, preventing invalid arguments from being passed during metastore startup.

When VERBOSE is not set, an empty string argument ("") is passed to the startup command, which causes a parsing error. 

No

**Docker commands:**
**Without Verbose:**
docker run -d -p 9083:9083 --env SERVICE_NAME=metastore --name metastore-standalone apache/hive:${HIVE_VERSION}
**With Verbose:**
docker run -d -p 9083:9083 --env SERVICE_NAME=metastore --env VERBOSE="true" --name metastore-standalone apache/hive:${HIVE_VERSION}

### DIFF
--- a/packaging/src/docker/entrypoint.sh
+++ b/packaging/src/docker/entrypoint.sh
@@ -22,14 +22,18 @@ set -x
 : "${DB_DRIVER:=derby}"
 
 SKIP_SCHEMA_INIT="${IS_RESUME:-false}"
-[[ $VERBOSE = "true" ]] && VERBOSE_MODE="--verbose" || VERBOSE_MODE=""
+[[ $VERBOSE = "true" ]] && VERBOSE_MODE="--verbose"
 
 function initialize_hive {
   COMMAND="-initOrUpgradeSchema"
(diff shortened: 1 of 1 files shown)


okumin:
With the patch, all commands are very successful.

```sh
% docker run --rm --env SERVICE_NAME=metastore --name metastore-standalone apache/hive:4.1.0-SNAPSHOT
% docker run --rm --env SERVICE_NAME=metastore --env VERBOSE="true" --name metastore-standalone apache/hive:4.1.0-SNAPSHOT
% docker run --rm --env SERVICE_NAME=hiveserver2 --name hiveserver2 apache/hive:4.1.0-SNAPSHOT
```

okumin:
@kokila-19 Thanks for reporting and fixing the issue!

kokila-19:
Thank you @okumin for the review and merge.

PR #5831, opened 2025-05-29
HIVE-28978:Remove Json-Path from Hive

devaspatikrishnatri:
json-smart comes as a part of json-path , and the latest release of json-path was 2.9.0 in Jan 2024 , which brings in json-smart 2.5.0.This drops json-path explicitly from hive as it is not used.

fix CVEs

No

Built locally, relying on pre-commits.


okumin:
Hmm.
https://github.com/json-path/JsonPath/pull/1030

okumin:
Reviewing. I verified `net.minidev:json-smart:jar:2.5.0` and `net.minidev:accessors-smart:jar:2.5.0` come from the json-path library on the master branch.
https://github.com/apache/hive/actions/runs/15371374316/job/43250990748

```
[INFO] |  +- com.jayway.jsonpath:json-path:jar:2.9.0:runtime
[INFO] |  |  \- net.minidev:json-smart:jar:2.5.0:runtime
[INFO] |  |     \- net.minidev:accessors-smart:jar:2.5.0:runtime
```

okumin:
I verified this patch upgrades all places using `net.minidev:json-smart` and `net.minidev:accessors-smart`. It also upgrade `org.ow2.asm:asm:jar`.
https://github.com/apache/hive/actions/runs/15340007646/job/43164237620?pr=5831

```
[INFO] +- com.jayway.jsonpath:json-path:jar:2.9.0:runtime
[INFO] +- net.minidev:json-smart:jar:2.5.2:runtime
[INFO] |  \- net.minidev:accessors-smart:jar:2.5.2:runtime
[INFO] |     \- org.ow2.asm:asm:jar:9.7.1:runtime
```

okumin:
Nashorn uses `asm` and Nashorn was included by the JDK upgrade.
https://github.com/apache/hive/commit/413069ea0faeb893fe78faa215b08a70ece80595#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8

```
[INFO] |  +- org.openjdk.nashorn:nashorn-core:jar:15.4:compile
[INFO] |  |  +- org.ow2.asm:asm:jar:7.3.1:compile
[INFO] |  |  +- org.ow2.asm:asm-commons:jar:7.3.1:compile
[INFO] |  |  |  \- org.ow2.asm:asm-analysis:jar:7.3.1:compile
[INFO] |  |  +- org.ow2.asm:asm-tree:jar:7.3.1:compile
[INFO] |  |  \- org.ow2.asm:asm-util:jar:7.3.1:compile
```

I wonder if we are allowed to include Nashorn or if we can remove it. Anyway, I'm asking it first.
https://lists.apache.org/thread/yyy1fckg2bjvs309bcvszbcr9jtqxvn2

zabetak (commented):
Why should we do this in Hive and not in Calcite which is the dependency that brings in transitively json-path?

devaspatikrishnatri:
@zabetak 
I think hive-exec directly references json-path.
https://github.com/apache/hive/blob/81c02a72e88dee8bf28a88b23401e9c3f1fad4a7/ql/pom.xml#L853
But I can try this upgrade in calcite as well.

pom.xml
```diff
@@ -1298,6 +1299,18 @@
         <artifactId>json-path</artifactId>
         <version>${json-path.version}</version>
```
zabetak: Instead of upgrading json-smart, I was wondering if we could drop json-path completely? Have you explored this option?

zabetak: The dep comes from Calcite and in the context of Hive I get the impression that we don't use that code path so maybe we can get rid of this dep especially since json-path does not seem to be actively maintained.

devaspatikrishnatri: I will try it out on local and see how it goes.

zabetak: The build will probably pass but try also to run some tests to see that they don't fail.

devaspatikrishnatri:
Have tried removing json-path completely from hive.
[DepTree](https://issues.apache.org/jira/secure/attachment/13076824/mvn_dependency_tree1.txt).
Now this dependency comes only from Calcite(1.33) at a version of 2.7.0.
If pre-commits pass and this is indeed not required then maybe we can drop this dependency from hive completely and just keep the upgrades on calcite side , as after Jan 2024 there does not seem to be any new version of json-path.

zabetak:
@devaspatikrishnatri By removing the json-path, I was thinking actually excluding it from calcite and not having it at all in the Hive dependency tree.

devaspatikrishnatri:
@zabetak Let me try that. I was under wrong impression.

devaspatikrishnatri:
[Dep Tree](https://issues.apache.org/jira/secure/attachment/13076915/mvn_dependency_tree5.txt).
This now completely removes it from Hive , waiting for pre-commits.

devaspatikrishnatri:
@zabetak I do not think we can completely drop json-path , there are multiple test failures like this 
"missing com/jayway/jsonpath/spi/mapper/MappingProvider".

We can remove all direct references from hive and keep it coming from caclite , and maintain this dependency there itself.
Or
We can keep updating the json-path and json-smart versions explicitly in hive.

devaspatikrishnatri:
@zabetak Any suggestions on which of the above two is the preferred approach ?

zabetak:
@devaspatikrishnatri The way it seems is that we have a problem in Calcite and by pinning the dependency in Hive we are making it a Hive problem as well. Probably the best would be to let Calcite it deal with its transitive dependencies. Anyways pinning version is in general risky especially when we start touching transitively L+3 (calcite->json-path->json-smart) dependencies. 

Summing up I am leaning towards removing all direct references of json-path from hive.

devaspatikrishnatri:
Got it . Will do the needful.
[Dep-Tree](https://issues.apache.org/jira/secure/attachment/13077052/mvn_dependency_tree_final_4.txt) after dropping json-path from hive.

arorasimran0309:
@devaspatikrishnatri is there a plan to fix failed tests for this change.

PR #5832, opened 2025-05-29
HIVE-28967: Support multiple partitions in RELOAD event

saihemanth-cloudera:
To support multiple partitions in the Reload event

Clients like Impala can use this fire multiple partitions in a single RELOAD event

No

Local


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -9826,7 +9826,7 @@ public FireEventResponse fire_listener_event(FireEventRequest rqst) throws TExce
       catName = rqst.isSetCatName() ? rqst.getCatName() : getDefaultCatalog(conf);
       dbName = rqst.getDbName();
       tblName = rqst.getTableName();
-      List<String> partitionVals = rqst.getPartitionVals();
```
dengzhhu653, line 9829: should we also consider the old client in case `rqst.getPartitionVals()` is not null?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/messaging/ReloadMessage.java
```diff
@@ -40,7 +40,7 @@ protected ReloadMessage() {
      *
      * @return The Json format of Partition object if the table is partitioned else return null.
      */
-    public abstract Partition getPtnObj() throws Exception;
+    public abstract Iterable<Partition> getPartitionObjs() throws Exception;
```
dengzhhu653, line 43: might introduce the backward problem for this API?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: Ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/messaging/ReloadMessage.java
saihemanth-cloudera: True, clients consuming this event will have to update their code.

saihemanth-cloudera: https://issues.apache.org/jira/browse/IMPALA-14089 - Impala's grammar is changed to take in multiple partitions.

dengzhhu653: per my understanding, this reload event is only for Impala, right?

saihemanth-cloudera: yes correct

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/events/ReloadEvent.java
```diff
@@ -67,10 +68,13 @@ public ReloadEvent(String catName, String db, String table, List<String> partVal
                 this.tableObj.getParameters().putAll(tblParams);
             }
             if (partVals != null) {
-                this.ptnObj = handler.get_partition(MetaStoreUtils.prependNotNullCatToDbName(catName, db),
-                        table, partVals);
+                this.ptns = new ArrayList<>();
+                for(List<String> partVal : partVals) {
+                    this.ptns.add(handler.get_partition(MetaStoreUtils.prependNotNullCatToDbName(catName, db),
```
dengzhhu653: how about use `get_partitions_by_names_req` to get all partitions in a round

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -9840,9 +9845,9 @@ public FireEventResponse fire_listener_event(FireEventRequest rqst) throws TExce
                 .get(MetaStoreEventListenerConstants.DB_NOTIFICATION_EVENT_ID_KEY_NAME)));
       } else {
         String msg = "Reload event id not generated for ";
-        if (event.getPartitionObj() != null) {
-          msg += "partition " + Arrays
-                  .toString(event.getPartitionObj().getValues().toArray()) + " of ";
+        if (event.getPartitions() != null) {
```
dengzhhu653, line 9848: if the partitions are large, will the message pollute the whole log file? if we really need it, how about only log the part_vals?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/events/ReloadEvent.java
saihemanth-cloudera: ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: These partitions spec comes directly from end-user, so this won't big (usually 5-10 partitions). We need entire partition object to identify the self event.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/events/ReloadEvent.java
```diff
@@ -23,24 +23,28 @@
...
 import org.apache.hadoop.hive.metastore.api.ClientCapability;
+import org.apache.hadoop.hive.metastore.api.GetPartitionsByNamesRequest;
 import org.apache.hadoop.hive.metastore.api.GetTableRequest;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
 import org.apache.thrift.TException;
+//import org.datanucleus.store.types.wrappers.ArrayList;
```
dengzhhu653: nit: you may need to remove this commented line, leave it to you to decide.

PR #5835, opened 2025-06-01
HIVE-28972: HMS performace degradation post HIVE-28909 for alter query

dengzhhu653:
HMS performace degradation post HIVE-28909 for alter table cascade, with this PR the alter query in this Jira runs ~30X faster against the master, and 5X faster without the HIVE-28909 on my local. This also saves the entries stored in back db, before (with or without HIVE-28909) it has 4 million entries in COLUMNS_V2, now it's only about ~800 records.

Why the performance degradation post HIVE-28909?
With HIVE-28909, the column is inserted to the db one by one per partition, before that the partition's columns are added in a batch. With this PR, all the partitions refer to the same MColumnDescriptor the table holds, so no more column is inserted per partition.

No

Current unit tests

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
@@ -407,6 +407,8 @@ public class HiveAlterHandler implements AlterHandler {
               MetastoreConf.getBoolVar(handler.getConf(), MetastoreConf.ConfVars.COLSTATS_RETAIN_ON_COLUMN_REMOVAL);
 
           if (runPartitionMetadataUpdate) {
+            // Don't validate table-level stats for a partitoned table.
+            msdb.alterTable(catName, dbname, name, newt, null);
             if (cascade || retainOnColRemoval) {
               parts = msdb.getPartitions(catName, dbname, name, -1);
               for (Partition part : parts) {
@@ -431,8 +433,6 @@ public class HiveAlterHandler implements AlterHandler {
               TableName tableName = new TableName(catName, dbname, name);
               msdb.deleteAllPartitionColumnStatistics(tableName, writeIdList);
             }
-            // Don't validate table-level stats for a partitoned table.
-            msdb.alterTable(catName, dbname, name, newt, null);
           } else {
             LOG.warn("Alter table not cascaded to partitions.");
             msdb.alterTable(catName, dbname, name, newt, writeIdList);
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
@@ -409,6 +409,7 @@ public class HiveAlterHandler implements AlterHandler {
           if (runPartitionMetadataUpdate) {
             // Don't validate table-level stats for a partitoned table.
             msdb.alterTable(catName, dbname, name, newt, null);
+
             if (cascade || retainOnColRemoval) {
               parts = msdb.getPartitions(catName, dbname, name, -1);
               for (Partition part : parts) {
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
@@ -5188,7 +5188,7 @@ public class ObjectStore implements RawStore, Configurable {
     catName = normalizeIdentifier(catName);
     name = normalizeIdentifier(name);
     dbname = normalizeIdentifier(dbname);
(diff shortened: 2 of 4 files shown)


Aggarwal-Raghav:
@dengzhhu653, Thanks for the quick fix and the explanation, I will verify this in my env as well.

Aggarwal-Raghav:
> @dengzhhu653, Thanks for the quick fix and the explanation, I will verify this in my env as well.

Yes, the fix is indeed working. **It took 81.864 seconds to perform the alter table cascade**, which is significantly faster.

dengzhhu653:
> > @dengzhhu653, Thanks for the quick fix and the explanation, I will verify this in my env as well.
> 
> Yes, the fix is indeed working. **It took 81.864 seconds to perform the alter table cascade**, which is significantly faster.

On my local laptop the alter finished in 66s, compared to 370s without HIVE-28909. Have you tested the using a new table created from scratch, i.e, create table -> msck repair partition-> alter cascade?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/model/MColumnDescriptor.java
```diff
@@ -44,6 +44,7 @@ public MColumnDescriptor(List<MFieldSchema> cols) {
     fields = cols.stream().map(schema ->
         new MColumn(schema.getName(), schema.getType(), schema.getComment()))
             .collect(Collectors.toList());
+    fields.forEach(column -> column.setCd(this));
```
deniskuzZ: minor: why not do this via constructor?
````
new MColumn(schema.getName(), schema.getType(), this, schema.getComment()))
````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java [hunk 2]
```diff
@@ -5188,7 +5188,7 @@ private Partition alterPartitionNoTxn(String catName, String dbname,
     catName = normalizeIdentifier(catName);
     name = normalizeIdentifier(name);
     dbname = normalizeIdentifier(dbname);
-    MPartition newp = convertToMPart(newPart, table, false);
+    MPartition newp = convertToMPart(newPart, table, true);
```
deniskuzZ: can we drop `useTableCD` arg as it is always true now?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
```diff
@@ -407,6 +407,8 @@ public List<Void> run(List<Partition> input) throws Exception {
               MetastoreConf.getBoolVar(handler.getConf(), MetastoreConf.ConfVars.COLSTATS_RETAIN_ON_COLUMN_REMOVAL);
 
           if (runPartitionMetadataUpdate) {
+            // Don't validate table-level stats for a partitoned table.
+            msdb.alterTable(catName, dbname, name, newt, null);
             if (cascade || retainOnColRemoval) {
```
deniskuzZ, line 413: please add a new line before if section

Aggarwal-Raghav:
> > > @dengzhhu653, Thanks for the quick fix and the explanation, I will verify this in my env as well.
> > 
> > 
> > Yes, the fix is indeed working. **It took 81.864 seconds to perform the alter table cascade**, which is significantly faster.
> 
> On my local laptop the alter finished in 66s, compared to 370s without [HIVE-28909](https://issues.apache.org/jira/browse/HIVE-28909). Have you tested the using a new table created from scratch, i.e, create table -> msck repair partition-> alter cascade?

yes, tested against master branch, with the same steps mentioned in repro. The difference might be because of config differences between our environment. I am using very basic configs https://github.com/Aggarwal-Raghav/hive-mac-setup/blob/main/hive/hive-site.xml

I think we can ignore small time delta, unless you think otherwise 😅

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
dengzhhu653: done

dengzhhu653:
> > > > @dengzhhu653, Thanks for the quick fix and the explanation, I will verify this in my env as well.
> > > 
> > > 
> > > Yes, the fix is indeed working. **It took 81.864 seconds to perform the alter table cascade**, which is significantly faster.
> > 
> > 
> > On my local laptop the alter finished in 66s, compared to 370s without [HIVE-28909](https://issues.apache.org/jira/browse/HIVE-28909). Have you tested the using a new table created from scratch, i.e, create table -> msck repair partition-> alter cascade?
> 
> yes, tested against master branch, with the same steps mentioned in repro. The difference might be because of config differences between our environment. I am using very basic configs https://github.com/Aggarwal-Raghav/hive-mac-setup/blob/main/hive/hive-site.xml
> 
> I think we can ignore small time delta, unless you think otherwise 😅

Thank you for the validation, I'm just wondering if I was missing anything.

deniskuzZ (approved):
+1, pending tests

PR #5837, opened 2025-06-02
HIVE-28075: Vectorized DayOFWeek returns inconsistent results for non…

tanishq-chugh:
…-UTC timezones

Fix the issue of UDF dayofweek returning inconsistent results for non-UTC time zones.

Currently, the Calender object being used in VectorUDFDayOfWeekString.java is initialised to use UTC timezone by default. So, the date string passed to the same is represented in UTC timezone irrespective of the timezone set in hive/JVM.

No, it is fixing a bug.

Manual. Added a qtest as well.


ql/src/test/queries/clientpositive/udf_dayofweek_with_tz.q
```diff
@@ -0,0 +1,31 @@
...
+set hive.vectorized.execution.enabled=true;
+
+SELECT fund_code,
+ test_date,
+ dayofweek(test_date) AS SR,
+ CASE
+     WHEN dayofweek(test_date) = 1 THEN 7
+     ELSE dayofweek(test_date) - 1
+ END AS week_day
+FROM dayOfWeek_test;
```
okumin, line 31: Is it expected that the result sets are identical on my machine without this patch?

```
SEC016210079    2023-04-13      5       4
PREHOOK: query: SELECT fund_code,
 test_date,
 dayofweek(test_date) AS SR,
 CASE
     WHEN dayofweek(test_date) = 1 THEN 7
     ELSE dayofweek(test_date) - 1
 END AS week_day
FROM dayOfWeek_test
PREHOOK: type: QUERY
PREHOOK: Input: default@dayofweek_test
POSTHOOK: query: SELECT fund_code,
 test_date,
 dayofweek(test_date) AS SR,
 CASE
     WHEN dayofweek(test_date) = 1 THEN 7
     ELSE dayofweek(test_date) - 1
 END AS week_day
FROM dayOfWeek_test
POSTHOOK: type: QUERY
POSTHOOK: Input: default@dayofweek_test
SEC016210079    2023-04-13      5       4
```

tanishq-chugh: Hi @okumin 
I don't think so. I was able to repro the issue with the same qtest running with TestCliDriver without the VectorUDFDayOfWeekString.java file change from this patch.

okumin: I used TestMiniLlapLocalCliDriver, and it is something, which LLAP can't reproduce. I made sure the issue happens with TestCliDriver. Thanks

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/VectorUDFDayOfWeekString.java
```diff
@@ -36,8 +36,7 @@ public final class VectorUDFDayOfWeekString extends VectorUDFTimestampFieldStrin
   private static final long serialVersionUID = 1L;
 
   private transient final SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd");
-  private transient final Calendar calendar = Calendar.getInstance(
-      TimeZone.getTimeZone("UTC"));
+  private transient final Calendar calendar = Calendar.getInstance();
 
   public VectorUDFDayOfWeekString(int colNum, int outputColumnNum) {
```
okumin, line 35: Do you think we can use the same logic as the non-vectorized version?
https://github.com/apache/hive/blob/de994e5c556930d1761601a61d6941c69fa9e153/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFDayOfWeek.java#L68-L72

tanishq-chugh: Yes, that would be much better. It would be more consistent and we'll be using Hive's DateParser instead as well. 
Thanks for catching this !
Have made the change in [aca0ed4](https://github.com/apache/hive/pull/5837/commits/aca0ed41e43ddafe3184b1f7bf117fd5c9e15ca2)

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/VectorUDFDayOfWeekString.java
```diff
@@ -49,15 +41,16 @@ public VectorUDFDayOfWeekString() {
 
   @Override
   protected long getField(byte[] bytes, int start, int length) throws ParseException {
-    Date date = null;
     try {
       String decoded = Text.decode(bytes, start, length);
-      date = format.parse(decoded);
+      Date date = DateParser.parseDate(decoded.toString());
+      if (date == null) {
+        throw new ParseException("Unable to parse date string: " + decoded, 0);
```
okumin: I honestly don't know the exact security policy of Hive, and I may be overthinking it. When I examined the relevant codes, I found that no piece had exposed the argument or row. Is it possible that we should not expose `decoded`? Say a malicious user submits `SELECT DAY_OF_WEEK(credit_card_number) FROM pii_table`. A card number can be stored in HS2's log files or a client system, such as Apache Airflow. I'm considering what semantics are desirable. Related:
https://issues.apache.org/jira/browse/HIVE-20644

tanishq-chugh: Yes, this could have led to a potential CVE. i'll remove the `decoded` string being exposed in the exception.
Thanks for catching this @okumin !

tanishq-chugh: Made the change in [720ff34](https://github.com/apache/hive/pull/5837/commits/720ff345754fd1d387870bc82d96b5e48cb5eae3)

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/VectorUDFDayOfWeekString.java
```diff
@@ -20,25 +20,17 @@
...
+import org.apache.hadoop.hive.common.type.Date;
+import org.apache.hive.common.util.DateParser;
 
 /**
  * Expression to get day of week.
  * Extends {@link VectorUDFTimestampFieldString}
  */
 public final class VectorUDFDayOfWeekString extends VectorUDFTimestampFieldString {
 
-  private static final long serialVersionUID = 1L;
```
okumin, line 36: Sorry, please restore this variable because this class is serializable.

okumin (approved):
I added a minor comment about `serialVersionUID`. Other than that, looks good!

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/expressions/VectorUDFDayOfWeekString.java
tanishq-chugh: Added back in [14bd5a6](https://github.com/apache/hive/pull/5837/commits/14bd5a68899f585e3fd87c7477db9f4d10c77953)

okumin: Thanks. I will merge this after the 24-hour rule is satisfied.

tanishq-chugh: Sure, Thanks @okumin !

okumin:
Merged. @tanishq-chugh Thanks for addressing the data integrity problem 👍

PR #5838, opened 2025-06-02
HIVE-25495: Upgrade JLine to version 3

abstractdog:


abstractdog:
opening new PR to check precommits
already known failure to be checked: TestHplSqlViaBeeLine

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -1147,7 +1154,15 @@ public int begin(String[] args, InputStream inputStream, boolean keepHistory) th
     } catch (Exception e) {
       // ignore
     }
-    return execute(reader, false);
+    return startListening();
```
zratkai: Why not calling execute(lineReader, false); ?

beeline/src/java/org/apache/hive/beeline/Commands.java
```diff
@@ -291,7 +291,7 @@ public boolean dropall(String line) {
       return beeLine.error(beeLine.loc("no-current-connection"));
     }
     try {
-      if (!(beeLine.getConsoleReader().readLine(beeLine.loc("really-drop-all")).equals("y"))) {
+      if (!(beeLine.getLineReader().readLine(beeLine.loc("really-drop-all")).equals("y"))) {
```
zratkai, line 292: Can we add it as a private static final field?

beeline/src/java/org/apache/hive/beeline/BeeLine.java
abstractdog: makes sense, originally, startListening encapsulated more logic, now it's just a single line, let me change

beeline/src/java/org/apache/hive/beeline/Commands.java
abstractdog: hm, I haven't touched constants with this PR, but as it's already huge, this can also be taken care of
btw, e.g. "really-drop-all" and all other strings used in beeline.loc are used only once in the Commands class, does it make sense to refactor all of them to contants @zratkai ? there are around 40 of them, I mean, we would end up having 40 private static final in Commands that are used only once and used only in this class

beeline/pom.xml
```diff
@@ -114,6 +114,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-hdfs</artifactId>
+      <scope>test</scope>
+    </dependency>
```
ayushtkn: Why is hdfs being added as a dependency here? It is already there at L194

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -938,8 +945,8 @@ private boolean connectUsingArgs(BeelineParser beelineParser, CommandLine cl) {
     String propertyFile = cl.getOptionValue("property-file");
     if (propertyFile != null) {
       try {
-        this.consoleReader = new ConsoleReader();
-      } catch (IOException e) {
+        this.lineReader = LineReaderBuilder.builder().build();
+      } catch (IOError e) {
         handleException(e);
```
ayushtkn: `IOException` has changed to `IOError`, can this `handleException` handle it?

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -1270,6 +1278,7 @@ public HS2ConnectionFileParser getHiveSiteHS2ConnectionFileParser() {
     return new HiveSiteHS2ConnectionFileParser();
   }
 
+  @VisibleForTesting
   int runInit() {
```
ayushtkn: this is used in `Commands.java` which is prod code & that too requires this visibility, I don't think we need ``VisibleForTesting`` here

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -1360,16 +1369,18 @@ private int executeFile(String fileName) {
...
-                       : ConsoleReader.NULL_MASK;
+                       : LineReaderImpl.NULL_MASK;
 
+    String line;
     while (!exit) {
       try {
         // Execute one instruction; terminate on executing a script if there is an error
         // in silent mode, prevent the query and prompt being echoed back to terminal
-        String line = (getOpts().isSilent() && getOpts().getScriptFile() != null) ? reader
+        line = (getOpts().isSilent() && getOpts().getScriptFile() != null) ? reader
```
ayushtkn: Why you need to extract this? It works the previous way as well

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -1385,7 +1396,17 @@ private int execute(ConsoleReader reader, boolean exitOnError) {
         } else if (line != null) {
           lastExecutionResult = ERRNO_OK;
         }
-
+      } catch (EndOfFileException t) {
```
ayushtkn: nit 
can you change the variable name to `eof`, with `t` it just give me `Throwable` vibes

beeline/src/java/org/apache/hive/beeline/BeeLineDummyTerminal.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public BeeLineDummyTerminal() {
+    this(true);
+  }
+
+  public BeeLineDummyTerminal(boolean isBeeLine) {
+    super(isBeeLine);
+  }
+
+  protected Terminal buildTerminal(InputStream inputStream) throws IOException {
```
ayushtkn, line 44: nit
can you add the `override` annotation

beeline/src/java/org/apache/hive/beeline/BeeLineOpts.java
```diff
@@ -152,11 +153,17 @@ public String get(String envVar) {
...
+      Terminal terminal = TerminalBuilder.terminal();
+      if (terminal.getWidth() > 0) {
+        maxWidth = terminal.getWidth();
+      }
+      if (terminal.getHeight() > 0) {
+        maxHeight = terminal.getHeight();
+      }
+      terminal.close();
+    } catch(IOException e) {
+      // nothing to do
```
ayushtkn: We should atleast log something here

beeline/src/java/org/apache/hive/beeline/BeeLineOpts.java
```diff
@@ -152,11 +153,17 @@ public String get(String envVar) {
...
-      maxHeight = terminal.getHeight();
+    try {
+      Terminal terminal = TerminalBuilder.terminal();
+      if (terminal.getWidth() > 0) {
+        maxWidth = terminal.getWidth();
+      }
+      if (terminal.getHeight() > 0) {
+        maxHeight = terminal.getHeight();
+      }
+      terminal.close();
```
ayushtkn, line 164: Curious where was the terminal closed earlier?

beeline/src/java/org/apache/hive/beeline/BeeLineOpts.java
```diff
@@ -195,12 +202,11 @@ public File saveDir() {
 
 
   @Override
-  public int complete(String buf, int pos, List cand) {
+  public void complete(LineReader reader, ParsedLine line, List<Candidate> candidates) {
```
ayushtkn, line 205: this looks like some behaviour change to me, can we not return the values which were returned earlier, supposedly 0 & -1

beeline/src/java/org/apache/hive/beeline/Commands.java
```diff
@@ -1103,16 +1103,29 @@ public String handleMultiLineCmd(String line) throws IOException {
...
-        extra = beeLine.getConsoleReader().readLine(prompt.toString());
+
+      try {
+        if (beeLine.getOpts().isSilent() && beeLine.getOpts().getScriptFile() != null) {
+          extra = beeLine.getLineReader().readLine(null, mask);
+        } else {
+          extra = beeLine.getLineReader().readLine(prompt.toString());
+        }
+      } catch (EndOfFileException t) {
+        /*
```
ayushtkn: maybe change t to eof and better to put a debug log here

beeline/src/java/org/apache/hive/beeline/Commands.java
```diff
@@ -1701,6 +1713,24 @@ public boolean connect(Properties props) throws IOException {
...
+   * behavior. This is typically used for input prompts such as username/password.
+   *
+   * @param prompt the prompt message displayed to the user
+   * @param mask the character used to mask the input, if any
+   * @return the full line read, or the partial input captured before the EndOfFileException was thrown
+   */
+  private String readLineWithPrompt(String prompt, Character mask) {
+    try {
+      return beeLine.getLineReader().readLine(prompt, mask);
+    } catch (EndOfFileException e) {
```
ayushtkn: eof

beeline/src/java/org/apache/hive/beeline/DatabaseConnection.java
```diff
@@ -79,24 +79,7 @@ void setCompletions(boolean skipmeta) throws SQLException, IOException {
...
-            if (Character.isWhitespace(c)) {
-              return true;
-            }
-            return !(Character.isLetterOrDigit(c))
-                && c != '_'
-                && extraNameCharacters.indexOf(c) == -1;
-          }
-        },
-        new SQLCompleter(SQLCompleter.getSQLCompleters(beeLine, skipmeta)));
+    sqlCompleter = new ArgumentCompleter(new SQLCompleter(SQLCompleter.getSQLCompleters(beeLine, skipmeta)));
```
ayushtkn, line 78: there is a variable above `extraNameCharacters` that goes unused with this change, that can be dropped

beeline/src/java/org/apache/hive/beeline/DatabaseConnection.java
```diff
@@ -79,24 +79,7 @@ void setCompletions(boolean skipmeta) throws SQLException, IOException {
             : getDatabaseMetaData().getExtraNameCharacters();
 
     // setup the completer for the database
-    sqlCompleter = new ArgumentCompleter(
-        new ArgumentCompleter.AbstractArgumentDelimiter() {
-          // delimiters for SQL statements are any
-          // non-letter-or-number characters, except
-          // underscore and characters that are specified
-          // by the database to be valid name identifiers.
```
ayushtkn, line 87: are these case being taken care, I think in the new code Uses the default delimiter logic from JLine’s ArgumentCompleter:
Usually splits on whitespace and some basic punctuation & doesn’t account for underscores or custom DB identifier characters.

That is what my little research says, can you double check once

beeline/src/java/org/apache/hive/beeline/TableNameCompletor.java
```diff
@@ -38,11 +41,11 @@ class TableNameCompletor implements Completer {
   }
 
   @Override
-  public int complete(String buf, int pos, List cand) {
-    if (beeLine.getDatabaseConnection() == null) {
-      return -1;
+  public void complete(LineReader reader, ParsedLine line, List<Candidate> candidates) {
```
ayushtkn, line 44: can we return int & maintain the method signature, this is a public method, changing signature of a public method afaik is an incompatible change

beeline/src/java/org/apache/hive/beeline/cli/HiveCli.java
```diff
@@ -32,11 +32,15 @@ public static void main(String[] args) throws IOException {
...
-    beeLine = new BeeLine(false);
+    beeLine = createBeeline();
     try {
       return beeLine.begin(cmd, inputStream);
     } finally {
       beeLine.close();
     }
   }
+
+  protected BeeLine createBeeline() {
```
ayushtkn: is protected needed or default access works?

beeline/src/test/org/apache/hive/beeline/TestBeelineArgParsing.java
```diff
@@ -70,7 +70,7 @@ public TestBeelineArgParsing(String connectionString, String driverClazzName, St
     this.defaultSupported = defaultSupported;
   }
 
-  public class TestBeeline extends BeeLine {
+  public static class TestBeeline extends BeeLine {
```
ayushtkn: this just seems some style fix, can be avoided in this scope

beeline/src/test/org/apache/hive/beeline/cli/HiveCliForTest.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package org.apache.hive.beeline.cli;
+
+import org.apache.hive.beeline.BeeLine;
+import org.apache.hive.beeline.BeeLineDummyTerminal;
+
+public class HiveCliForTest extends HiveCli {
+
+  protected BeeLine createBeeline() {
```
ayushtkn: add `override` annotation

beeline/src/test/org/apache/hive/beeline/cli/TestHiveCli.java
```diff
@@ -80,7 +81,7 @@ public void testCommentStripping() {
 
   @Test
   public void testSetPromptValue() {
-    verifyCMD("set hive.cli.prompt=MYCLI;SHOW\nTABLES;", "MYCLI> ", err, null,
+    verifyCMD("set hive.cli.prompt=MYCLI;SHOW\nTABLES;\n", "MYCLI> ", err, null,
```
ayushtkn: why this \n thing has kicked in? is it something that will bother the end user in prod as well. Moreover rather than modifying each call with \n can we handle it in `verifyCMD` method itself?

cli/pom.xml
```diff
@@ -121,6 +121,10 @@
       <artifactId>hadoop-mapreduce-client-core</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-1.2-api</artifactId>
+    </dependency>
```
ayushtkn: where is this being used?

itests/util/pom.xml
```diff
@@ -122,6 +122,10 @@
       <artifactId>hive-llap-server</artifactId>
       <classifier>tests</classifier>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-jdbc</artifactId>
+    </dependency>
```
ayushtkn: why?

itests/hive-unit/src/test/java/org/apache/hive/beeline/hs2connection/BeelineWithHS2ConnectionFileTestBase.java
```diff
@@ -75,19 +76,19 @@ public abstract class BeelineWithHS2ConnectionFileTestBase {
   public String transportMode =  null;
 
 
-  protected class TestBeeLine extends BeeLine {
+  protected class BeelineWithConfigFileManager extends BeeLineDummyTerminal {
```
ayushtkn, line 78: remove the import for beeline above it goes unused after this change

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestBeelinePasswordOption.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.hive.beeline;
 
+import static org.junit.Assert.assertEquals;
```
ayushtkn, line 20: this is unused

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestBeeLineWithArgs.java
```diff
@@ -175,13 +173,40 @@ static String testCommandLineScript(List<String> argList, InputStream inputStrea
...
     String output = os.toString("UTF8");
     return output;
   }
 
+  /*
+   * Creates a BeeLineDummyTerminalFromFile instance that skips the main execution code path
+   * if an init script has already run. Since this unit test class executes scripts in init (-i) or file (-f) mode,
+   * we can avoid using waiting hacks and simply return from execute() with the original init result.
+   */
+  private static BeeLineDummyTerminalFromFile getBeeLineDummyTerminal() {
```
ayushtkn: is this required for JLIne3, or is it some optimisation kicking in here?

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
```diff
@@ -609,47 +610,32 @@ public static Completer[] getCommandCompleter() {
       candidateStrings.add(s.toLowerCase());
     }
 
-    StringsCompleter strCompleter = new StringsCompleter(candidateStrings);
-
-    // Because we use parentheses in addition to whitespace
-    // as a keyword delimiter, we need to define a new ArgumentDelimiter
-    // that recognizes parenthesis as a delimiter.
```
ayushtkn, line 616: I think the new code doesn't handle this, did you try this case?

cli/src/test/org/apache/hadoop/hive/cli/TestCliDriverMethods.java
```diff
@@ -206,24 +204,28 @@ private PrintStream headerPrintingTestDriver(Schema mockSchema) throws CommandPr
...
-
+    Completer[] completers = CliDriver.getCommandCompleter();
+    assertEquals(2, completers.length);
+    assertTrue(completers[0] instanceof ArgumentCompleter);
+    assertNotNull(completers[1]);
+
+    final List<Candidate> candidates1 = new ArrayList<>();
+    candidates1.add(new Candidate(")"));
+    completers[1].complete(null, CliDriver.getDefaultParser().parse("fdsdfsdf", 0), candidates1);
+    assertEquals(")", candidates1.get(0).value());
```
ayushtkn: can we change these `get(0)` with `getFirst()` here & below

cli/src/test/org/apache/hadoop/hive/cli/TestCliDriverMethods.java
```diff
@@ -206,24 +204,28 @@ private PrintStream headerPrintingTestDriver(Schema mockSchema) throws CommandPr
...
+    System.out.printf("--- --> %s%n", candidates2.get(0).value());
+    assertTrue(candidates2.get(0).value().endsWith("length("));
+
+    final List<Candidate> candidates3 = new ArrayList<>();
+    completers[0].complete(null, CliDriver.getDefaultParser().parse("set f", 0), candidates3);
+    assertEquals("set", candidates3.get(0).value());
   }
 
+
+
```
ayushtkn: nit
avoid

common/src/java/org/apache/hive/common/util/MatchingStringsCompleter.java
```diff
@@ -0,0 +1,73 @@
...
+    this(Arrays.asList(strings));
+  }
+
+  public MatchingStringsCompleter(Iterable<String> strings) {
+    strings.forEach(candidateStrings::add);
+  }
+
+  public MatchingStringsCompleter(Candidate... candidates) {
+    Arrays.stream(candidates).map(Candidate::value).forEach(candidateStrings::add);
+  }
```
ayushtkn: seems unused

itests/hive-unit/src/test/java/org/apache/hive/beeline/BeeLineDummyTerminalFromFile.java
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hive.beeline;
+
+import org.apache.hive.beeline.BeeLineDummyTerminal;
```
ayushtkn: this ain't required

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -1496,13 +1548,12 @@ boolean dispatch(String line) {
     }
 
     line = HiveStringUtils.removeComments(line);
+    line = line.trim();
 
-    if (line.trim().length() == 0) {
+    if (line.length() == 0) {
```
ayushtkn: since you are touching it, can you change it to line.isEmpty()

beeline/src/java/org/apache/hive/beeline/Commands.java
```diff
@@ -48,6 +48,7 @@
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.ListIterator;
```
ayushtkn: unused

beeline/src/java/org/apache/hive/beeline/BeeLine.java
abstractdog:
```
this.lineReader = LineReaderBuilder.builder().build();
```
doesn't throw a checked exception at all, will remove the whole try/catch

beeline/src/java/org/apache/hive/beeline/BeeLine.java
abstractdog: yeah, most probably a leftover of earlier work

beeline/src/java/org/apache/hive/beeline/BeeLine.java
abstractdog: absolutely, ack!

beeline/src/java/org/apache/hive/beeline/BeeLineOpts.java
abstractdog: right, it wasn't really handled, I'll take care of closing the terminals since they might use native resources

beeline/src/java/org/apache/hive/beeline/BeeLineOpts.java
abstractdog: apparently, since Jline3 a Completer doesn't return a value, we cannot preserve that behavior:
https://github.com/jline/jline3/blob/778edccd4c1d863c2d3a89b9fc387c38203c7009/reader/src/main/java/org/jline/reader/Completer.java#L61

beeline/src/java/org/apache/hive/beeline/DatabaseConnection.java
abstractdog: yes, apparently this way we would lose parsing logic, I just found jline3 commit which removed the original overridden delimiter:
https://github.com/jline/jline3/commit/8e09ee5fad9973ca6359ddae8056bf6d0614fd7d

apparently, the same implementation has been moved to the DefaultParser, so as I'm already instantiating one, I can simply reuse the old logic by overriding DefaultParser.isDelimiterChar()

beeline/src/java/org/apache/hive/beeline/DatabaseConnection.java
abstractdog: reused this according to another comment in this area (around instantiating DefaultParser)

beeline/src/java/org/apache/hive/beeline/TableNameCompletor.java
abstractdog: this implements Jline3's completer, which changed its signature, we cannot do anything else than following
fortunately beeline doesn't expose this return value, as it's part of jline internals

beeline/src/java/org/apache/hive/beeline/cli/HiveCli.java
abstractdog: works with package-protected

cli/pom.xml
abstractdog: not sure, was part of the patch that I picked up earlier, removing it and we'll see

itests/util/pom.xml
abstractdog: hm, it compiles without this, maybe a leftover (when I implemented custom Beeline subclasses)

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestBeelinePasswordOption.java
abstractdog: I would rather change to static import to stay inline with the rest of the class

beeline/src/test/org/apache/hive/beeline/cli/TestHiveCli.java
abstractdog: good catch, it's definitely something I wanted to get back to :) it's turned out that I haven't handled the EndOfLineException properly, which is currently being fixed, making all these extra line feeds unnecessary (hence properly getting back to the old Jline2 behavior)

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
abstractdog: similarly to the beeline case, this is now handled in DefaultParser which is implemented as
```
  static DefaultParser getDefaultParser() {
    return new DefaultParser() {
      @Override
      public boolean isDelimiterChar(CharSequence buffer, int pos) {
        // Because we use parentheses in addition to whitespace
        // as a keyword delimiter, we need to define a new ArgumentDelimiter
        // that recognizes parenthesis as a delimiter.
        final char c = buffer.charAt(pos);
        return (Character.isWhitespace(c) || c == '(' || c == ')' ||
                c == '[' || c == ']');
      }
    };
  }
```
in this class

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestBeeLineWithArgs.java
abstractdog: it's because of the refactor that came along with the upgrade
thanks for pointing this out, I'm now rewriting the test classes to become simpler is possible
I'll let you know when all the comments are addressed

itests/hive-unit/src/test/java/org/apache/hive/beeline/BeeLineDummyTerminalFromFile.java
abstractdog: the whole class could have been removed, fortunately

abstractdog:
@ayushtkn : all your comments have been addressed, precommits turned green again

beeline/src/java/org/apache/hive/beeline/BeeLine.java
```diff
@@ -1394,58 +1409,153 @@ private int execute(ConsoleReader reader, boolean exitOnError) {
...
+    }
+    return inputStream;
+  }
+
+  protected Terminal buildTerminal(InputStream inputStream) throws IOException {
+    Terminal terminal;
+    if (inputStream != null) { // typically when there is a file script to read from
+      terminal = TerminalBuilder.builder().streams(inputStream, getErrorStream()).build();
+    } else { // no input stream, normal operation: proper behavior needs a system terminal
+      // system terminal can only be created with system streams
```
ayushtkn: nit ``system terminal`` is coming twice here, first in the first line then again in the second line

beeline/src/test/org/apache/hive/beeline/cli/TestHiveCli.java
```diff
@@ -206,7 +207,7 @@ public void testVariablesForSource() {
 
   @Test
   public void testErrOutput() {
-    verifyCMD("show tables;set system:xxx=5;set system:yyy=${system:xxx};\nlss;",
+    verifyCMD("show tables;set system:xxx=5;set system:yyy=${system:xxx};\nlss;\n",
```
ayushtkn: why \n kicks in here?

abstractdog: eh, leftover, removing :D

beeline/src/java/org/apache/hive/beeline/BeeLine.java
abstractdog: fixed in https://github.com/apache/hive/pull/5838/commits/c130f4d6f74087bf965a4462f7ec04d04faf0841

beeline/src/test/org/apache/hive/beeline/cli/TestHiveCli.java
abstractdog: fixed in https://github.com/apache/hive/pull/5838/commits/c130f4d6f74087bf965a4462f7ec04d04faf0841

abstractdog:
> One minor question, rest LGTM

thanks a lot @ayushtkn , fixed!

PR #5839, opened 2025-06-02
HIVE-28987: Iceberg: A faulty query predicate can compromise transaction isolation

deniskuzZ:

Enable conflict detection filter only when plan contains 1 TableScan Operator for a table

BugFix

No

Cluster + locally
https://issues.apache.org/jira/secure/attachment/13076866/iceberg_atomic_merge_update.q

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/session/SessionStateUtil.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/session/SessionStateUtil.java
@@ -25,10 +25,18 @@ import com.google.common.collect.Maps;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.ql.QueryState;
+import org.apache.hadoop.hive.ql.exec.UDFArgumentException;
+import org.apache.hadoop.hive.ql.plan.ExprNodeDesc;
+import org.apache.hadoop.hive.ql.plan.ExprNodeDescUtils;
+import org.apache.hadoop.hive.ql.plan.ExprNodeGenericFuncDesc;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class SessionStateUtil {
+  private static final Logger LOG = LoggerFactory.getLogger(SessionStateUtil.class);
 
   private static final String COMMIT_INFO_PREFIX = "COMMIT_INFO.";
+  private static final String CONFLICT_DETECTION_FILTER = "conflictDetectionFilter.";
   public static final String DEFAULT_TABLE_LOCATION = "defaultLocation";
 
   private SessionStateUtil() {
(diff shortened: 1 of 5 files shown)


deniskuzZ:
@simhadri-g, do you recall we have you combined using `and` predicates from all the TableScan operators in order to construct the final iceberg commit filter?

deniskuzZ:
@kasakrisz, @zabetak isn't it possible to find a top level table scan operator for a specific table based on DAG plan?

simhadri-g:
> redicates from all the TableScan operators in order to construct the final iceberg commit filter?

Yes, this was the same logic spark used to build the conflict detection filter as well.
https://github.com/apache/iceberg/blob/main/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java#L399-L410

deniskuzZ:
> > redicates from all the TableScan operators in order to construct the final iceberg commit filter?
> 
> Yes, this was the same logic spark used to build the conflict detection filter as well. https://github.com/apache/iceberg/blob/main/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java#L399-L410

Thanks @simhadri-g , I have no idea how Spark manages/constructs the `filterExpressions`, but based on the code, it seems Spark has a single TS operator (root TS ???). 
Maybe you can check with @vrozov and help us do the same in Hive.  
````
scan.filterExpressions()
````
In the current PR, since I don't have any better ideas how to identify root TS and construct the final `filterExpressions`, I am disabling the `conflict detection filter` when DAG has multiple TS operators for the write table. 
cc @okumin

**Comment by simhadri-g** [comment]
Sure, Denys. 
But, can you please check if this issue happens when with tez.local.mode is set to false?  I had seen something similar when tez.local.mode was enabled https://github.com/apache/hive/pull/5188

**Comment by deniskuzZ** [comment]
> Sure, Denys. But, can you please check if this issue happens when with tez.local.mode is set to false? I had seen something similar when tez.local.mode was enabled #5188

The problem isn't tied to `tez.local.mode`; it arises from generating an invalid filter when subqueries reference the target table

**Comment by kasakrisz** [comment]
> @kasakrisz, @zabetak isn't it possible to find a top level table scan operator for a specific table based on DAG plan?

@deniskuzZ 
I'm not fully understand the functionality introduced by https://issues.apache.org/jira/browse/HIVE-27653 but I try to answer your question: All TS operators are the starting operators of a plan. There is a field in SemanticAnalyzer which contains all of them: https://github.com/apache/hive/blob/4193a6db6243bfd8a5a2a3d142bef5acdb02e389/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java#L364

The key in the map is the alias of the table scanned by the TS. Depending on the query a table might be scanned by more than one TS but with different aliases. Shared work optimizer merges them in most cases but there are cases when it is not possible.

**Comment by deniskuzZ** [comment]
> > @kasakrisz, @zabetak isn't it possible to find a top level table scan operator for a specific table based on DAG plan?
> 
> @deniskuzZ I'm not fully understand the functionality introduced by https://issues.apache.org/jira/browse/HIVE-27653 but I try to answer your question: All TS operators are the starting operators of a plan. There is a field in SemanticAnalyzer which contains all of them:
> 
> https://github.com/apache/hive/blob/4193a6db6243bfd8a5a2a3d142bef5acdb02e389/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java#L364
> 
> The key in the map is the alias of the table scanned by the TS. Depending on the query a table might be scanned by more than one TS but with different aliases. Shared work optimizer merges them in most cases but there are cases when it is not possible.

what is the best place to check the TS predicates post SWO? atm we are constructing filters before SWO, however, this could be improved in a follow up PR

**Comment by kasakrisz** [comment]
> what is the best place to check the TS predicates post SWO? atm we are constructing filters before SWO, however, this could be improved in a follow up PR

Top operator map also can be found in `ParseContext`
```
pCtx.getTopOps()
```
It is referencing the same map in `SemanticAnalyzer.topOps`

After `SemanticAnalyzer.compilePlan ` all optimizations should be completed https://github.com/apache/hive/blob/af59f241a07fb9960412529be71c6bfa1fcd042e/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java#L13331

File: ql/src/java/org/apache/hadoop/hive/ql/session/SessionStateUtil.java [hunk 1]
```diff
@@ -99,12 +107,34 @@ public static void addCommitInfo(
...
+    String key = CONFLICT_DETECTION_FILTER + tableName;
+
+    getConflictDetectionFilter(conf, tableName)
+        .ifPresentOrElse(prevFilterExpr -> {
+            if (!prevFilterExpr.isSame(filterExpr)) {
+              ExprNodeDesc disjunction = null;
+              try {
+                disjunction = ExprNodeDescUtils.disjunction(prevFilterExpr, filterExpr);
+              } catch (UDFArgumentException e) {
+                LOG.warn(e.getMessage());
```
kasakrisz: This doesn't print the stack trace.

deniskuzZ: fixed

kasakrisz (approved):
+1 pending tests

PR #5840, opened 2025-06-03
HIVE-28954: CI fails intermittently due to ephemeral-storage exhaustion

zabetak:
Introduce ephemeral storage request and limit for HDB container based on current usage and cluster capacity.

Based on recent runs the HDB container, which executes the tests, consumes 10Gi to 15Gi of ephemeral storage. To ensure that pods are scheduled correctly to the GKE nodes that have the necessary capacity we should add an explicit resource request.

Moreover, to avoid malfunctioning PRs/pods affect the overall health of the cluster we set the resource limit to 20Gi that is reasonably high to permit precommits to run fine and can also guard against accidental changes that may cause disk spikes.

No

Monitor requests/limits through the GKE console.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -136,6 +136,8 @@ def hdbPodTemplate(closure) {
         resourceLimitCpu: '8000m',
         resourceRequestMemory: '6400Mi',
         resourceLimitMemory: '12000Mi',
+        resourceRequestEphemeralStorage: '10Gi',
+        resourceLimitEphemeralStorage: '20Gi',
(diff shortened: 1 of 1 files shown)


kokila-19:
LGTM +1
There is one test failure which is flaky , re-triggering should pass the tests.

kokila-19 (approved):

aturoczy (approved):

zabetak:
Thanks everyone for the reviews! Much appreciated :)

PR #5841, opened 2025-06-03
HIVE-28976 : Enhance Commit message in notification_log  to correctly filter events during incremental replication

harshal-16:
Details:
  * Added write ID, database name into commit & abort message
  * Updated filter in CommitTxnHandler to utilise it during replication dump process

Testing:
  * Added test cases
  * Tested on YCloud

  * In commit & abort message in notification_log will have additional write IDs and databases details to be used by replication. 
  * During Replication dump, if hive.repl.filter.transactions is true then, it will use these details there are some DDLs which needs new write ID but it does not add an entry into txn_write_notification_log table which was only source of truth earlier to figure out whether that txn was read / write related. 

  * In presence of hive.repl.filter.transactions property, replication will fail on the DR side post 11 days if user runs Truncate operation (or any other which implements DDLDescWithWriteId). Reason being it generates alloc_write_id event in notification_log table. In presence of above filter on source side we want to filter out commit_txn which are associated with read operation or it's not related to database in replication. 
  * To decide this we rely on another historical table txn_write_notification_log, now, in insert and other DML operatioins it does get an entry. But for truncate / DDLDescWithWriteId it doesn't get any entry into that which leads to skipping the commit operation on the source and on the DR side, we open the implicit txn if alloc_write_id event is present and we wait for commit / abort to come from the source
  * This results into txn being open for a long time and then it eventually times out and adds repl_incompatible flag on the database on the DR side
  * So, now there are 2 options to address this issue. Either add an dummy entry into txn_write_notification_log, which may open up some other issue and we will have to keep on monitoring the code change because if hive adds any new command which implements DDLDescWithWriteId then it can break the replication again in future
  * The 2nd option is to enhance the commit message, which we are storing in the notifiation_log and add some additional context, which replication dump can use to figure it out.

  * No

  * Added test cases
  * Tested on actual cluster

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
@@ -81,6 +81,7 @@ import org.apache.hadoop.hive.metastore.datasource.DataSourceProvider;
 import org.apache.hadoop.hive.metastore.datasource.DataSourceProviderFactory;
 import org.apache.hadoop.hive.metastore.events.AbortTxnEvent;
 import org.apache.hadoop.hive.metastore.events.AcidWriteEvent;
+import org.apache.hadoop.hive.metastore.events.CommitTxnEvent;
 import org.apache.hadoop.hive.metastore.events.ListenerEvent;
 import org.apache.hadoop.hive.metastore.messaging.EventMessage;
 import org.apache.hadoop.hive.metastore.metrics.Metrics;
@@ -90,6 +91,7 @@ import org.apache.hadoop.hive.metastore.txn.entities.CompactionState;
 import org.apache.hadoop.hive.metastore.txn.entities.LockInfo;
 import org.apache.hadoop.hive.metastore.txn.entities.MetricsInfo;
 import org.apache.hadoop.hive.metastore.txn.entities.TxnStatus;
+import org.apache.hadoop.hive.metastore.txn.entities.TxnWriteDetails;
 import org.apache.hadoop.hive.metastore.txn.jdbc.commands.*;
 import org.apache.hadoop.hive.metastore.txn.jdbc.functions.*;
 import org.apache.hadoop.hive.metastore.txn.jdbc.queries.*;
@@ -129,6 +131,7 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiPredicate;
+import java.util.stream.Collectors;
 
 import static org.apache.hadoop.hive.metastore.txn.TxnUtils.getEpochFn;
 import static org.apache.hadoop.hive.metastore.utils.MetaStoreUtils.getDefaultCatalog;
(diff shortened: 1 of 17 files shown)


standalone-metastore/src/test/java/org/apache/hadoop/hive/metastore/events/TestCommitTxnEventWithDbAndWriteId.java
```diff
@@ -0,0 +1,75 @@
...
+        Assert.assertTrue(actual.isEmpty());
+
+        List actualWriteIds = commitTxnMsg.getWriteIds();
+        Assert.assertTrue(actualWriteIds.remove(1L));
+        Assert.assertTrue(actualWriteIds.remove(2L));
+        Assert.assertTrue(actualWriteIds.isEmpty());
+
+        Assert.assertEquals(999L, commitTxnMsg.getTxnId().longValue());
+    }
+
```
pudidic, line 73: Could you remove the blank lines?

deniskuzZ:
@harshal-16, please fill out all the details in the PR summary

pudidic (approved):
+1. Looks good to me.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
```diff
@@ -513,23 +516,53 @@ public long getTargetTxnId(String replPolicy, long sourceTxnId) throws MetaExcep
...
     List<Long> txnIds = rqst.getTxn_ids();
     TxnErrorMsg txnErrorMsg = TxnErrorMsg.NONE;
     if (rqst.isSetErrorCode()) {
       txnErrorMsg = TxnErrorMsg.getTxnErrorMsg(rqst.getErrorCode());
     }
+    HashMap<Long, List<TxnWriteDetails>> txnWriteDetailsMap = new HashMap<>();
+    if (transactionalListeners != null) {
+      //Find the write details for this transaction.
+      //Doing it here before the metadata tables are updated below.
+      for(Long txnId : txnIds)
```
deniskuzZ, line 563: @harshal-16, that is very inefficient, why not pass the txnIds list directly to the backend db?
cc @dengzhhu653

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
```diff
@@ -513,23 +516,53 @@ public long getTargetTxnId(String replPolicy, long sourceTxnId) throws MetaExcep
 
   @Override
   public void abortTxn(AbortTxnRequest rqst) throws NoSuchTxnException, MetaException, TxnAbortedException {
+    List<TxnWriteDetails> txnWriteDetails = new ArrayList<>();
+    if (transactionalListeners != null) {
+      //Find the write details for this transaction.
+      //Doing it here before the metadata tables are updated below.
+      txnWriteDetails = getWriteIdsForTxnID(rqst.getTxnid());
```
deniskuzZ, line 523: @harshal-16, why not pass the `tableWriteIds` from the `DbTxnManager` and avoid the additional DB request?

deniskuzZ:
@harshal-16, why didn't you remove `GetTxnDbsUpdatedHandler`? `getTxnDbsUpdated` is no longer used

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
harshal-16: ACK

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
harshal-16: can you please explain this a bit more? like how

harshal-16:
> @harshal-16, why didn't you remove `GetTxnDbsUpdatedHandler`? `getTxnDbsUpdated` is no longer used

ACK

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
deniskuzZ: let me check this further, but why not reuse existing code instead of adding `GetWriteIdsForTxnIDHandler`?
````
HMSHandler.class

@Override
  public void abort_txn(AbortTxnRequest rqst) throws TException {
    List<WriteEventInfo> writeEventInfo =
        getMS().getAllWriteEventInfo(rqst.getTxnid(), null, null);
    rqst.setWriteEventInfo(writeEventInfo);
    
    getTxnHandler().abortTxn(rqst);
````
@harshal-16 is that sufficient for you?

deniskuzZ: few notes:

- no need to fetch writeIds for readOnly transaction:`DbTxnManager#txnType`!
- if abort is not manually triggered, `DbTxnManager#tableWriteIds` has all you need, just add this to the `AbortTxnRequest`.

harshal-16: This was the main issue 
  List<WriteEventInfo> writeEventInfo =
        getMS().getAllWriteEventInfo(rqst.getTxnid(), null, null);
    rqst.setWriteEventInfo(writeEventInfo);

- this uses txn_write_notification_log table to get the information, which has entry only in case of DML operations which generates the hdfs files
- In case of truncate / DDLDescWithWriteId doesn't have any entry into that table resulting in 0 rows in case of getAllWriteEventInfo

deniskuzZ: @harshal-16, can't we update the `getAllWriteEventInfo` to return info even for truncate case (use TXN_TO_WRITE_ID) instead of adding duplicate functionality?

harshal-16: getAllWriteEventInfo calls the txn_write_notification_log table, which has much more information than just the write ID. i.e., tableObj, a file list for a given write operation.
For truncate, I wanted to add a dummy entry into txn_write_notification_log so getAllWriteEventInfo returns the value. But later on we figured out that there are much more DDL operations that require write ID (all those that implement DDLDescWithWriteId)
So, checking the txn_to_write_id table and enhancing the commit message in the notification_log table was the only viable option.

deniskuzZ: @harshal-16, can we pass the DbTxnManager#tableWriteIds to the `AbortTxnRequest` and call the `GetWriteIdsForTxnIDHandler` only when the list is empty?

harshal-16: This might be possible, but i will have to check.

PR #5842, opened 2025-06-03
HIVE-28988: Handle materialized view commit txn during hive acid repl

shivaom02:
  * Commit txn of alter mv command was getting populated by write event infos.
  * This was causing a repl failure on load.
  * This is now fixed.

Check [HIVE-28988](https://issues.apache.org/jira/browse/HIVE-28988)

### Does this PR introduce any user-facing change?
No, just fixes a bug

Added itest. Will check CI result.


harshal-16 (approved):

ql/src/java/org/apache/hadoop/hive/ql/parse/repl/dump/events/CommitTxnHandler.java
```diff
@@ -128,6 +130,20 @@ private List<WriteEventInfo> getAllWriteEventInfo(Context withinContext) throws
               })));
   }
 
+  private List<WriteEventInfo> getAllWriteEventInfoExceptMV(List<WriteEventInfo> writeEventInfoList) {
```
deniskuzZ, line 133: do we even need to store the MV write events and do the filter later, as done here?

pudidic:
+1 Looks good to me.

deniskuzZ:
@pudidic, why was the PR merged when it has an open comment?

ql/src/java/org/apache/hadoop/hive/ql/parse/repl/dump/events/CommitTxnHandler.java
harshal-16: It is required for https://github.com/apache/hive/commit/f82d780d6d5db116a19ebef846ce302d36cf2d4d
As Materialised view is not supported by replication, we need a way to differentiate between insert and materialised view rebuild kind of event 
you can check the truth table in CommitTxnHandler in 
(https://github.com/apache/hive/commit/f82d780d6d5db116a19ebef846ce302d36cf2d4d#diff-d8c044001331ab3f52b23cf8ff3394a2b941c8848f7711e1d69f4b4cdd703cd9)
We want to replay the commit for insert events but want to skip for the MV related events

deniskuzZ: thanks @harshal-16 for replying! Going forward, please try to address open questions before merging.
If MV is not supported for the replication, why do we need to store MV related events at all? Do we use them elsewhere?

PR #5843, opened 2025-06-04
HIVE-28990: Collecting the non-native summary can be slow to raise th…

dengzhhu653:
…e timeout

TestMetaTookTaskMetadataSummary

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -1905,6 +1905,8 @@ public class MetastoreConf {
         "Only collect the non-native table's summary that has been updated/changed in configured recent days."),
     METADATA_SUMMARY_MAX_NONNATIVE_TABLES("hive.metatool.summary.maxNonNativeTables", "hive.metatool.summary.maxNonNativeTables", "",
         "The maximum non-native tables allowed per table type during collecting the summary."),
+    METADATA_SUMMARY_NONNATIVE_THREADS("hive.metatool.summary.nonnative.threads", "hive.metatool.summary.nonnative.threads", 20,
+        "Number of threads to be allocated for MetaToolTaskMetadataSummary for collecting the non-native table's summary."),
 
     // These are all values that we put here just for testing
     STR_TEST_ENTRY("test.str", "hive.test.str", "defaultval", "comment"),
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java
@@ -20,6 +20,7 @@ package org.apache.hadoop.hive.metastore.tools.metatool;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ArrayListMultimap;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonArray;
@@ -169,6 +170,7 @@ public class MetaToolTaskMetadataSummary extends MetaToolTask {
   Pair<MetaSummarySchema, List<MetadataTableSummary>> obtainAndFilterSummary() throws MetaException {
     Deadline.registerIfNot(taskTimeout);
     boolean isTimerStarted = false;
+    ExecutorService service = null;
     try {
       isTimerStarted = Deadline.startTimer("obtainAndFilterSummary");
       MetaToolObjectStore objectStore = getObjectStore();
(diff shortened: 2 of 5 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1905,6 +1905,8 @@ public enum ConfVars {
         "Only collect the non-native table's summary that has been updated/changed in configured recent days."),
     METADATA_SUMMARY_MAX_NONNATIVE_TABLES("hive.metatool.summary.maxNonNativeTables", "hive.metatool.summary.maxNonNativeTables", "",
         "The maximum non-native tables allowed per table type during collecting the summary."),
+    METADATA_SUMMARY_NONNATIVE_THREADS("hive.metatool.summary.nonnative.threads", "hive.metatool.summary.nonnative.threads", 20,
+        "Number of threads to be allocated for MetaToolTaskMetadataSummary for collecting the non-native table's summary."),
```
saihemanth-cloudera, line 1909: I feel the thread count is too high for an operation that would run in the background and that is not run very often. Thoughts?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java
```diff
@@ -183,25 +185,45 @@ Pair<MetaSummarySchema, List<MetadataTableSummary>> obtainAndFilterSummary() thr
...
+                tableIds.size());
+            if (nThreads > 1) {
+              service = Executors.newFixedThreadPool(nThreads,
+                  new ThreadFactoryBuilder().setDaemon(true).setNameFormat("MetaToolTaskMetadataSummary #%d").build());
+            }
+          }
           for (MetadataTableSummary summary : tableSummaries) {
-            if (tableIds.contains(summary.getTableId())) {
+            if (!tableIds.contains(summary.getTableId())) {
+              filteredSummary.put(summary, null);
```
saihemanth-cloudera: Instead of a Map with value type void, we could have this variable type as `Set<MetadataTableSummary>`.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java
```diff
@@ -183,25 +185,45 @@ Pair<MetaSummarySchema, List<MetadataTableSummary>> obtainAndFilterSummary() thr
...
                   summary.getDbName(), summary.getTblName());
-              summaryHandler.appendSummary(tableName, summary);
-            } else {
-              filteredSummary.put(summary, null);
-            }
-            // If there is an exception while collecting the summary, remove it
-            if (summary.isDropped()) {
-              filteredSummary.put(summary, null);
+              Runnable task = () -> {
+                summaryHandler.appendSummary(tableName, summary);
```
saihemanth-cloudera, line 210: If there is an exception collecting the summary, should we log the same in the Summary report rather than ignoring it? so that the end-user is aware of it?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java
```diff
@@ -183,25 +185,45 @@ Pair<MetaSummarySchema, List<MetadataTableSummary>> obtainAndFilterSummary() thr
...
+              if (service != null) {
+                futures.add(service.submit(task));
+              } else {
+                task.run();
+              }
             }
           }
+          // Waiting for the result before closing the MetaSummaryHandler
+          for (Future<?> future : futures) {
+            future.get();
```
saihemanth-cloudera, line 225: Should we consider a timeout for this task? Threads could be blocked at the back end and the metadata summary tool can run for hours(I have seen this happen in a customer env).

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
dengzhhu653: I was inspired by `hive.metastore.fshandler.threads`, which default value is 15. The thread is used for loading the Iceberg table: reading the metadata files, where it might take time over thousands of Iceberg tables, leading to timeout. The actual threads is min(number of Iceberg tables, hive.metatool.summary.nonnative.threads)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java
dengzhhu653: My idea is to try the best to collect a `correct` summary, if we just ignore the exception(imaging the table is dropped when comes to collect), this will give a wrong insight of the table, which could be misleading.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java
dengzhhu653: the timeout is treated outside this method, 
https://github.com/apache/hive/blob/5c64b95c41bcd556b0f0d21648ac782d15cf14c8/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java#L117-L118
If the task is over the time limit, the main thread will be timeout and exit

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/tools/metatool/MetaToolTaskMetadataSummary.java
dengzhhu653: yes, i will change that

saihemanth-cloudera (approved):

saihemanth-cloudera:
Pending green tests

PR #5845, opened 2025-06-04
HIVE-28735: Propagated the exception properly from hash table-loader threads to main thread

Paramvir109:

Need to properly propagate the exception which can occur while parallel loading of fast Hash tables by the threads.
In the current scenario the thread is silently dying causing incorrect(lesser number)of HT entries to be reported which is giving incorrect results

It fixes bug caused by https://issues.apache.org/jira/browse/HIVE-25149 that hides actual exceptions and gives incorrect results.  Example query is added in the jira : https://issues.apache.org/jira/browse/HIVE-28735

No

No

Tested using the sample query. 
Hive 3 
```
Caused by: java.lang.OutOfMemoryError: Java heap space
    at org.apache.hadoop.hive.ql.exec.vector.mapjoin.fast.VectorMapJoinFastLongHashTable.expandAndRehash(VectorMapJoinFastLongHashTable.java:166)
    at org.apache.hadoop.hive.ql.exec.vector.mapjoin.fast.VectorMapJoinFastLongHashTable.add(VectorMapJoinFastLongHashTable.java:100)
    at org.apache.hadoop.hive.ql.exec.vector.mapjoin.fast.VectorMapJoinFastLongHashTable.adaptPutRow(VectorMapJoinFastLongHashTable.java:91)
    at org.apache.hadoop.hive.ql.exec.vector.mapjoin.fast.VectorMapJoinFastLongHashMap.putRow(VectorMapJoinFastLongHashMap.java:147)
    at org.apache.hadoop.hive.ql.exec.vector.mapjoin.fast.VectorMapJoinFastTableContainer.putRow(VectorMapJoinFastTableContainer.java:184)
    at org.apache.hadoop.hive.ql.exec.vector.mapjoin.fast.VectorMapJoinFastHashTableLoader.load(VectorMapJoinFastHashTableLoader.java:130)
    at org.apache.hadoop.hive.ql.exec.MapJoinOperator.loadHashTableInternal(MapJoinOperator.java:388)
    at org.apache.hadoop.hive.ql.exec.MapJoinOperator.loadHashTable(MapJoinOperator.java:457)
    at org.apache.hadoop.hive.ql.exec.MapJoinOperator.lambda$initializeOp$0(MapJoinOperator.java:241)
    ... 3 more
```

Before the fix : Wrong results on each run
After the fix : Getting same OOM exception as in hive 3

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
@@ -18,11 +18,15 @@
 package org.apache.hadoop.hive.ql.exec.vector.mapjoin.fast;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.LongAccumulator;
@@ -148,8 +152,9 @@ public class VectorMapJoinFastHashTableLoader implements org.apache.hadoop.hive.
     }
   }
 
-  private void submitQueueDrainThreads(VectorMapJoinFastTableContainer vectorMapJoinFastTableContainer)
+  private List<Future<?>> submitQueueDrainThreads(VectorMapJoinFastTableContainer vectorMapJoinFastTableContainer)
       throws InterruptedException, IOException, SerDeException {
+    List<Future<?>> futures = new ArrayList<>();
     for (int partitionId = 0; partitionId < numLoadThreads; partitionId++) {
       int finalPartitionId = partitionId;
       this.loadExecService.submit(() -> {
@@ -157,10 +162,12 @@ public class VectorMapJoinFastHashTableLoader implements org.apache.hadoop.hive.
           LOG.info("Partition id {} with Queue size {}", finalPartitionId, loadBatchQueues[finalPartitionId].size());
           drainAndLoadForPartition(finalPartitionId, vectorMapJoinFastTableContainer);
         } catch (IOException | InterruptedException | SerDeException | HiveException e) {
+          LOG.error("Caught error while starting HT threads: " + e.getMessage(), e);
           throw new RuntimeException("Failed to start HT Load threads", e);
         }
       });
     }
+    return futures;
   }
 
   private void drainAndLoadForPartition(int partitionId, VectorMapJoinFastTableContainer tableContainer)
@@ -271,7 +278,7 @@ public class VectorMapJoinFastHashTableLoader implements org.apache.hadoop.hive.
 
         tableContainer.setSerde(null, null); // No SerDes here.
         // Submit parallel loading Threads
-        submitQueueDrainThreads(tableContainer);
+        List<Future<?>> futures = submitQueueDrainThreads(tableContainer);
 
         long receivedEntries = 0;
         long startTime = System.currentTimeMillis();
@@ -310,6 +317,13 @@ public class VectorMapJoinFastHashTableLoader implements org.apache.hadoop.hive.
 
         if (!loadExecService.awaitTermination(2, TimeUnit.MINUTES)) {
           throw new HiveException("Failed to complete the hash table loader. Loading timed out.");
+        }
+        for (Future<?> f : futures) {
+          try {
+            f.get(); // This will throw if the thread threw
+          } catch (ExecutionException e) {
+            throw new HiveException("One of the loader threads failed", e.getCause());
+          }
         }
         batchPool.clear();
         LOG.info("Total received entries: {} Threads {} HT entries: {}", receivedEntries, numLoadThreads, totalEntries.get());


ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
```diff
@@ -105,6 +110,23 @@ public void init(ExecMapperContext context, MapredContext mrContext,
     this.htLoadCounter = tezContext.getTezProcessorContext().getCounters().findCounter(counterGroup, counterName);
   }
 
+  @VisibleForTesting
+  protected void initHTLoadingServiceForTest(Configuration conf, long estKeyCount, ExecutorService loadExecService) {
```
deniskuzZ: @Paramvir109, that is production code; it shouldn't contain any test-specific logic. Please use mocking techniques instead

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
```diff
@@ -148,19 +170,29 @@ public void resetBeforeOffer(HashTableElementBatch elementBatch) {
...
+    return submitQueueDrainThreads(vectorMapJoinFastTableContainer);
+  }
+
+  private List<Future<?>> submitQueueDrainThreads(VectorMapJoinFastTableContainer vectorMapJoinFastTableContainer)
       throws InterruptedException, IOException, SerDeException {
+    List<Future<?>> futures = new ArrayList<>();
     for (int partitionId = 0; partitionId < numLoadThreads; partitionId++) {
       int finalPartitionId = partitionId;
-      this.loadExecService.submit(() -> {
+      Future<?> future = this.loadExecService.submit(() -> {
```
deniskuzZ: please use `CompletableFuture` instead
````
List<CompletableFuture<Void>> loaderTasks = new ArrayList<>();
...................

CompletableFuture<Void> asyncTask = CompletableFuture.runAsync(() -> {
  try {
    LOG.info("Partition id {} with Queue size {}", finalPartitionId, loadBatchQueues[finalPartitionId].size());
    drainAndLoadForPartition(finalPartitionId, vectorMapJoinFastTableContainer);
  } catch (IOException | InterruptedException | SerDeException | HiveException e) {
    throw new RuntimeException("Failed to start HT Load thread", e);
  }
}, loadExecService);

.....................

loaderTasks.add(asyncTask);

try {
  CompletableFuture.allOf(loaderTasks.toArray(new CompletableFuture[0]))
    .get(2, TimeUnit.MINUTES);
} catch (TimeoutException e) {
  throw new HiveException("Failed to complete the hash table loader. Loading timed out.");
} catch (ExecutionException e) {
  throw new HiveException("One of the loader threads failed", e.getCause());
  Thread.currentThread().interrupt();
} finally {
  if (loadExecService != null && !loadExecService.isTerminated()) {
    loadExecService.shutdownNow();
  }
}
````

Comment (Paramvir109) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java]: Ack, I'll update it.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java (hunk 1 above)
Comment (Paramvir109) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java]: Hey @deniskuzZ, The method which I need to test - `submitQueueDrainThreads` is a private method hence I'm making a protected testing method which will actually call the private method. And `initHTLoadingService` method which is responsible for initialising all the class members(private) required for this test also has a private access in the class. I won't be able to mock private class members. I can use reflection to set the private class members but it won't be a good idea I believe

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java]: please make the method package-private and use it in a test. Please do not add any test-specific functions in the production code.

Comment (Paramvir109) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java]: Made the changes @deniskuzZ , Just added a protected getter for a private variable. Please review

File: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java [hunk 3]
```diff
@@ -105,7 +114,7 @@ public void init(ExecMapperContext context, MapredContext mrContext,
     this.htLoadCounter = tezContext.getTezProcessorContext().getCounters().findCounter(counterGroup, counterName);
   }
 
-  private void initHTLoadingService(long estKeyCount) {
+  protected void initHTLoadingService(long estKeyCount) {
```
deniskuzZ, line 117: why did you change to `protected` and not `package-private`?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
deniskuzZ: lgtm

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Test
+    public void testSubmitQueueDrainThreads_FutureGetThrowsExecutionException() throws
+            IOException, InterruptedException, SerDeException, ExecutionException, HiveException {
+        doThrow(new InterruptedException("Simulated interruption")).when(vectorMapJoinFastHashTableLoader)
+                .drainAndLoadForPartition(anyInt(), any(VectorMapJoinFastTableContainer.class));
+        vectorMapJoinFastHashTableLoader.initHTLoadingService(1048577);
+        List<CompletableFuture<Void>> loaderTasks = vectorMapJoinFastHashTableLoader.submitQueueDrainThreads(mockTableContainer);
+        assertEquals(2, loaderTasks.size());
+        vectorMapJoinFastHashTableLoader.getLoadExecService().shutdown();
```
deniskuzZ: i don't think that is a good way to repro the issue: 

- why prepare mock `drainAndLoadForPartition` and not even trigger it? 

Instead, execute direct `getLoadExecService().shutdown()` and `CompletableFuture.allOf`.

- your test should focus on `ExecutionException` and not the `InterruptedException` (based on PR description)

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java
```diff
@@ -0,0 +1,119 @@
...
+    private ProcessorContext mockProcessorContext;
+
+    @Mock
+    private TezCounters mockTezCounters;
+
+    @Mock
+    private TezCounter mockTezCounter;
+
+    @Mock
+    private MapJoinDesc joinDesc;
```
deniskuzZ, line 66: do you really need all these mocks?

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
deniskuzZ: @Paramvir109 , why did you change to `protected` and not `package-private`?

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java
Paramvir109: I need hconf object to be set. Since we don't want to change/add any testing related methods to the actual class, these mocks are required so that I can call the constructor for [VectorMapJoinFastHashTableLoader](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java#L86). Other approach would be to add a new constructor in the class which just initialises the hconf

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java
Paramvir109: @deniskuzZ , when we call the submitQueueDrainThreads method - it will internally call [drainAndLoadForPartition](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java#L158) and this is what I want to test. If any error occurs in drainAndLoadForPartition, it will get caught and RuntimeException will be thrown which will be propagated as ExecutionException. 
In my repro sceanrio the code flow was reaching till [here](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java#L160) but not being propagated properly.

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
Paramvir109: Makes sense, since any it won't be accessed outside the package. I'll update it

Paramvir109: Updated the PR. Can you please review @deniskuzZ ?

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java
deniskuzZ: [here](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java#L160) would be wrapped in `ExecutionException`. 
Your test is checking for `InterruptedException` which was already covered in https://github.com/apache/hive/pull/5845/files#diff-573e1d79d59124631df039ba5fedfd037457f3e3c3b8fb5e349b4086191bc197L330

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
```diff
@@ -307,9 +320,19 @@ public void load(MapJoinTableContainer[] mapJoinTables,
...
-
-        if (!loadExecService.awaitTermination(2, TimeUnit.MINUTES)) {
+        try {
+          CompletableFuture.allOf(loaderTasks.toArray(new CompletableFuture[0]))
+                  .get(2, TimeUnit.MINUTES);
+        } catch (TimeoutException e) {
           throw new HiveException("Failed to complete the hash table loader. Loading timed out.");
+        } catch (ExecutionException e) {
+          throw new HiveException("One of the loader threads failed", e.getCause());
+        } catch (InterruptedException e) {
```
deniskuzZ: you already have handling for `InterruptedException` and `loadExecService` shutdown [here](https://github.com/apache/hive/pull/5845/files#diff-573e1d79d59124631df039ba5fedfd037457f3e3c3b8fb5e349b4086191bc197R353-R363)
no need to duplicate

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java
deniskuzZ: that is fine, i wasn't sure all this mocks are neeeded

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
```diff
@@ -83,6 +88,10 @@ public VectorMapJoinFastHashTableLoader() {
 
   }
 
+  ExecutorService getLoadExecService() {
```
deniskuzZ, line 91: remove this

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
```diff
@@ -307,9 +320,19 @@ public void load(MapJoinTableContainer[] mapJoinTables,
         LOG.info("Finished loading the queue for input: {} waiting {} minutes for TPool shutdown", inputName, 2);
         addQueueDoneSentinel();
         loadExecService.shutdown();
```
deniskuzZ: remove loadExecService.shutdown()

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
deniskuzZ: please check the patch
````
Subject: [PATCH] cleanup
---
Index: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java	(revision 7e356e06eee218e077586a52811a42449196152d)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java	(date 1749805780973)
@@ -88,10 +88,6 @@
 
   }
 
-  ExecutorService getLoadExecService() {
-    return loadExecService;
-  }
-
   public VectorMapJoinFastHashTableLoader(TezContext context, Configuration hconf, MapJoinOperator joinOp) {
     this.tezContext = context;
     this.hconf = hconf;
@@ -114,7 +110,7 @@
     this.htLoadCounter = tezContext.getTezProcessorContext().getCounters().findCounter(counterGroup, counterName);
   }
 
-   void initHTLoadingService(long estKeyCount) {
+  void initHTLoadingService(long estKeyCount) {
     if (estKeyCount < VectorMapJoinFastHashTable.FIRST_SIZE_UP) {
       // Avoid many small HTs that will rehash multiple times causing GCs
       this.numLoadThreads = 1;
@@ -157,8 +153,8 @@
     }
   }
 
-   List<CompletableFuture<Void>> submitQueueDrainThreads(VectorMapJoinFastTableContainer vectorMapJoinFastTableContainer)
-          throws InterruptedException, IOException, SerDeException {
+  List<CompletableFuture<Void>> submitQueueDrainThreads(VectorMapJoinFastTableContainer vectorMapJoinFastTableContainer)
+          throws InterruptedException {
     List<CompletableFuture<Void>> loaderTasks = new ArrayList<>();
     for (int partitionId = 0; partitionId < numLoadThreads; partitionId++) {
       int finalPartitionId = partitionId;
@@ -319,20 +315,11 @@
 
         LOG.info("Finished loading the queue for input: {} waiting {} minutes for TPool shutdown", inputName, 2);
         addQueueDoneSentinel();
-        loadExecService.shutdown();
         try {
           CompletableFuture.allOf(loaderTasks.toArray(new CompletableFuture[0]))
-                  .get(2, TimeUnit.MINUTES);
-        } catch (TimeoutException e) {
-          throw new HiveException("Failed to complete the hash table loader. Loading timed out.");
-        } catch (ExecutionException e) {
-          throw new HiveException("One of the loader threads failed", e.getCause());
-        } catch (InterruptedException e) {
-          Thread.currentThread().interrupt();
-        } finally {
-          if (loadExecService != null && !loadExecService.isTerminated()) {
-            loadExecService.shutdownNow();
-          }
+              .get(2, TimeUnit.MINUTES);
+        } catch (ExecutionException | TimeoutException e) {
+          throw new HiveException("Failed to complete the hash table loader.", e.getCause());
         }
         batchPool.clear();
         LOG.info("Total received entries: {} Threads {} HT entries: {}", receivedEntries, numLoadThreads, totalEntries.get());
````

File: ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java (hunk 4 above)
Comment (Paramvir109) [reply ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java]:
``` 
ExecutionException thrown = assertThrows(ExecutionException.class, () -> {
            CompletableFuture.allOf(loaderTasks.toArray(new CompletableFuture[0]))
                    .get(2, TimeUnit.MINUTES);
        });
```
The above code block is checking for ExecutionException only

```
 assertInstanceOf(InterruptedException.class, cause.getCause());
```
The above assertion is just extra assertion which tells the underlying cause of ExecutionException. I can remove it too

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
Paramvir109: Ack, I'll update in the next revision

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
Paramvir109: I need this so I can actually shutdown the service in tearDown method after running unit test. Since loadExecService is a private member of the class, I thought to make a package-private getter for the same so that I can access it in unit test. 
Other approach will be to change access modifier of loadExecService to package-private

deniskuzZ: `drainAndLoadForPartition` in the finally should shutdown the service

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java
deniskuzZ: You should test the production code `drainAndLoadForPartition`, configure `drainAndLoadForPartition` to throw the exception, and not test the TEST

ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java
Paramvir109: Hey @deniskuzZ . Thanks for the review and sorry for replying a bit late.
service is shutdown [here](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java#L338) in the `load()` but since I can't test the load() method (As the original OOM scenario is reproducible only with specific data under some low memory conditions given in the jira) via unit test therefore I need this package private getter to actually close the service in my unit test

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/TestVectorMapJoinFastHashTableLoader.java
Paramvir109: I can't test the drainAndLoadForPartition() as the original OOM scenario is reproducible only with specific data under some low memory conditions given in the jira and I'll have to mimic the values for [vectorMapJoinFastTableContainer](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/mapjoin/fast/VectorMapJoinFastHashTableLoader.java#L151) to give OOM error which is not possible since the issue can't be reproduced via a unit test. Therefore I'm causing OOM error whenever drainAndLoadForPartition() is called and my part of the code will check if that OOM error can be propagated as RuntimeException in the main thread

PR #5847, opened 2025-06-06
HIVE-28993: use lower case datanucleus.plugin.pluginRegistryBundleCheck

vrozov:
use lower case `datanucleus.plugin.pluginregistrybundlecheck` and deprecate prior `datanucleus.plugin.pluginRegistryBundleCheck`.

After upgrade to 6.x datanucleus does not recognize old settings as it was changed to all lower case.

No

Using existing tests


simhadri-g (approved):
Adding a link to the datanucleus change for reference: https://github.com/datanucleus/datanucleus-core/commit/3e5b661000d11b1c1920db8ad6b05ec2f31ee217#diff-b9ebd71adfcb98d965225c373e44910dc01d1dd8c37a3e4d6d906d58f108e120R28 


LGTM pending green test run.
Thanks for the PR!

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -739,9 +739,10 @@ public enum ConfVars {
     DATANUCLEUS_INIT_COL_INFO("datanucleus.rdbms.initializeColumnInfo",
         "datanucleus.rdbms.initializeColumnInfo", "NONE",
         "initializeColumnInfo setting for DataNucleus; set to NONE at least on Postgres."),
-    DATANUCLEUS_PLUGIN_REGISTRY_BUNDLE_CHECK("datanucleus.plugin.pluginRegistryBundleCheck",
-        "datanucleus.plugin.pluginRegistryBundleCheck", "LOG",
-        "Defines what happens when plugin bundles are found and are duplicated [EXCEPTION|LOG|NONE]"),
+    DATANUCLEUS_PLUGIN_REGISTRY_BUNDLE_CHECK("datanucleus.plugin.pluginRegistryBundleCheck".toLowerCase(),
```
deniskuzZ, line 744: 1. why use `toLowerCase()` instead of just lowercasing the string?
2. hiveName remained the same, why mark it as deprecated?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1963,6 +1964,18 @@ public enum ConfVars {
       this.description = description;
     }
 
+    ConfVars(String varname, String hiveName, String defaultVal, boolean caseSensitive,
```
deniskuzZ, line 1989: can we reuse code in the overloaded constructors instead of copy&paste?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
vrozov: 1. Same as datanucleus. Please see https://github.com/datanucleus/datanucleus-core/blob/6dc3ad2e63b74efdd3a3c4d3f318a983048e13fc/src/main/java/org/datanucleus/PropertyNames.java#L25.
2. The old key whether it is `hiveName` or metastore `varname` has no effect, so both are deprecated.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
vrozov: it follows the same pattern as other `ConfVars` ctors that do not reuse overloaded constructor.

vrozov:
@deniskuzZ IMO, it will be good to revisit of how much testing was done during datanucleus major version upgrade and review https://www.datanucleus.org/products/accessplatform_6_0/migration.html.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: that's not a pattern, but a bad coding style

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: :) interesting style. 
Hive already creates plenty of HiveConf obj instances that take a significant part of Heap, so i am not sure it's a good idea to add an overhead by creating unnecessary objects.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
vrozov: agree, but that needs to be refactored in a separate PR that will take care of all ctors.

deniskuzZ: ok, agree

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
vrozov: HiveConf holds over 1K entries and avoiding one extra mapping is not going to buy much, IMO. Avoiding just one extra HiveConf object is going to save much more :)

deniskuzZ: true, but in any case, why add more redundancy?

PR #5850, opened 2025-06-08
[WIP] HIVE-28997: Replace nashorn-core

okumin:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,6 @@
     <hamcrest.version>1.3</hamcrest.version>
     <hbase.version>2.5.6-hadoop3</hbase.version>
     <hppc.version>0.7.2</hppc.version>
-    <nashorn.version>15.4</nashorn.version>
     <!-- required for logging test to avoid including hbase which pulls disruptor transitively -->
     <disruptor.version>3.3.7</disruptor.version>
     <hikaricp.version>4.0.3</hikaricp.version>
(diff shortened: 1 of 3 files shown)


ayushtkn (commented):
I have created a ticket HIVE-28984, to track this initiative. I am more inclined on using graalvm as linked in the ticket. I believe jol-core is GPLv2 + Classpath Exception, so graalvm might be better

okumin:
@ayushtkn I currently assume we use nashorn-core to use `jdk.nashorn.internal.ir.debug.ObjectSizeCalculator.getObjectSize`, which retrieves the deep size of a Java object, in a unit test. It's a kind of accident that `ObjectSizeCalculator.getObjectSize` exists in nashorn-core, and I assume we do not need an alternative JS runtime, but need an alternative object size calculator. Or we need both if we really need a JS runtime somewhere.

okumin:
I started feeling that this code path uses Nashorn 🤕 
https://github.com/apache/hive/blob/88dc98331244cea5a4d4247a91df6ccbd9296b6a/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java#L314

In that case, we will likely need GraalJS. I'm not sure if GraalJS has an alternative to `getObjectSize`, but at least we can mark Nashorn as a test dependency. I will wait for the result of HIVE-28984 at this point.

PR #5852, opened 2025-06-10
HIVE-28994: Implement direct sql for getAllDatabases and getDatabases in RawStore

shuyouZZ:
Implement direct sql for getAllDatabases and getDatabases in RawStore.

Improve the performance of getAllDatabases and getDatabases.

No

```
mvn test -Dtest.groups= -Dtest=org.apache.hadoop.hive.metastore.TestObjectStore#testDatabaseOps -pl :hive-standalone-metastore-server
```

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
@@ -463,6 +463,7 @@ class MetaStoreDirectSql {
               .map(String::trim)
               .filter(s -> !s.isEmpty())
               .distinct()
+              .map(String::toLowerCase)
               .toList();
       if (!validPatterns.isEmpty()) {
         queryText.append(" AND (");


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -1002,8 +1002,37 @@ public boolean dropDatabase(String catName, String dbname)
...
+      }
+
+      @Override
+      protected List<String> getSqlResult(GetHelper<List<String>> ctx) throws MetaException {
+        return directSql.getDatabases(catName, pattern);
+      }
+
+      @Override
+      protected List<String> getJdoResult(GetHelper<List<String>> ctx) throws MetaException {
+        return getGetDatabasesViaJdo(catName, pattern);
```
kokila-19: getGetDatabasesViaJdo has redundant get in the method name which can be removed

shuyouZZ: > getGetDatabasesViaJdo has redundant get in the method name which can be removed

thanks, have fix it.

okumin:
CI failed probably because of this change.

shuyouZZ:
> CI failed probably because of this change.

I can not open the continuous-integration page, it shows Not Found.
The previous execution of CI was successful.

okumin:
@shuyouZZ You can sign in the Jenkins with a GitHub account

shuyouZZ:
> @shuyouZZ You can sign in the Jenkins with a GitHub account

thanks, let me try

deniskuzZ:
is that PR abandoned?

shuyouZZ:
> is that PR abandoned?

Sorry for taking so long to reply. I still hope to continue working on this PR, but I'm not sure why this build failed.

okumin:
@shuyouZZ Although this might not be the direct reason, you have to rebase the topic branch. The CI says it was tested with JDK 17, but the upstream uses JDK 21
https://github.com/apache/hive/actions/runs/15722082143/job/44304757814?pr=5852

shuyouZZ:
> @shuyouZZ Although this might not be the direct reason, you have to rebase the topic branch. The CI says it was tested with JDK 17, but the upstream uses JDK 21 https://github.com/apache/hive/actions/runs/15722082143/job/44304757814?pr=5852

Ok. Let me try force rebase.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -993,8 +993,37 @@ public boolean dropDatabase(String catName, String dbname)
...
+    try {
+      return getDatabasesInternal(catName, pattern);
+    } catch (NoSuchObjectException e) {
+      throw new MetaException(ExceptionUtils.getStackTrace(e));
+    }
+  }
+
+  private List<String> getDatabasesInternal(String catName, String pattern)
+          throws MetaException, NoSuchObjectException {
+    return new GetHelper<List<String>>(catName, null, pattern, true, true) {
```
okumin, line 1006: I'd say two potential updates on this line.
First, I suspect this should be `new GetHelper<>(catname, null, null, true, true)` because GetHelper can unlikely handle a pattern. @dengzhhu653 We would like your advice on this hypothesis.
Second, if `GetHelper` accepts a pattern, this should be `GetHelper<>(catName, pattern, null, true, true)` because the given pattern is not that of tables but that of databases.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -449,6 +449,53 @@ public Database getDatabase(String catName, String dbName) throws MetaException{
...
+              .distinct()
+              .map(String::toLowerCase)
+              .toList();
+      if (!validPatterns.isEmpty()) {
+        queryText.append(" AND (");
+        for (int i = 0; i < validPatterns.size(); i++) {
+          if (i > 0) queryText.append(" OR ");
+          queryText.append("\"NAME\" LIKE ?");
+          params.add(validPatterns.get(i).replace('*', '%'));
+        }
```
okumin, line 474: Does this loop retain the original semantics? I can see the original implementation assumes `pattern` is a regular expression, which is not identical to that of SQL's `LIKE`.
https://github.com/apache/hive/blob/411d3a49cb7a2720907cf0342f5f77f53dbfed3d/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java#L2027-L2051

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -449,6 +449,53 @@ public Database getDatabase(String catName, String dbName) throws MetaException{
...
+        queryText.append(")");
+      }
+    }
+
+    queryText.append(" ORDER BY \"NAME\" ASC");
+
```
okumin, line 480: We might want to instantiate and reuse the query text.

```java
String properVariableName = queryText.toString();
```

PR #5856, opened 2025-06-10
HIVE-28984: Replace nashorn-core with graalvm which is compatible wit…

maheshrajus:
…h ASF license

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
     <hamcrest.version>1.3</hamcrest.version>
     <hbase.version>2.5.6-hadoop3</hbase.version>
     <hppc.version>0.7.2</hppc.version>
-    <nashorn.version>15.4</nashorn.version>
+    <graalvm.version>23.0.8</graalvm.version>
     <!-- required for logging test to avoid including hbase which pulls disruptor transitively -->
     <disruptor.version>3.3.7</disruptor.version>
     <hikaricp.version>4.0.3</hikaricp.version>
@@ -420,9 +420,14 @@
         <version>${commons-math3.version}</version>
       </dependency>
       <dependency>
-        <groupId>org.openjdk.nashorn</groupId>
-        <artifactId>nashorn-core</artifactId>
-        <version>${nashorn.version}</version>
+        <groupId>org.graalvm.js</groupId>
+        <artifactId>js-scriptengine</artifactId>
+        <version>${graalvm.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.graalvm.js</groupId>
+        <artifactId>js</artifactId>
+        <version>${graalvm.version}</version>
       </dependency>
       <dependency>
         <groupId>io.jsonwebtoken</groupId>
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.hive.ql.metadata;
 
+import com.oracle.truffle.js.scriptengine.GraalJSScriptEngine;
 import org.apache.hadoop.hive.metastore.api.AlreadyExistsException;
 import org.apache.hadoop.hive.metastore.api.GetPartitionsFilterSpec;
 import org.apache.hadoop.hive.metastore.api.GetPartitionsRequest;
@@ -28,11 +29,8 @@ import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.PartitionFilterMode;
 import org.apache.hadoop.hive.metastore.api.PartitionListComposingSpec;
 import org.apache.hadoop.hive.metastore.api.PartitionSpec;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.graalvm.polyglot.Context;
 
-import javax.script.ScriptEngine;
-import javax.script.ScriptEngineManager;
 import javax.script.ScriptException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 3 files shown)


ql/src/test/org/apache/hadoop/hive/ql/exec/tez/TestVectorMapJoinFastHashTable.java
```diff
@@ -44,6 +44,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@org.junit.Ignore("Replace nashorn-core with graalvm script engine")
```
ayushtkn: I think no need to ignore the test takes care of ignoring if nashorn isn't in classpath

maheshrajus: OK, fixed

maheshrajus: This test case using "jdk.nashorn.internal.ir.debug.ObjectSizeCalculator" for calculating the object size.
For correcting this test case we have jira https://issues.apache.org/jira/browse/HIVE-28997 to address it. thx !

pom.xml
```diff
@@ -420,9 +420,14 @@
...
-        <artifactId>nashorn-core</artifactId>
-        <version>${nashorn.version}</version>
+        <groupId>org.graalvm.js</groupId>
+        <artifactId>js-scriptengine</artifactId>
+        <version>${graalvm.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.graalvm.js</groupId>
+        <artifactId>js</artifactId>
+        <version>${graalvm.version}</version>
```
okumin, line 430: Both libraries are licensed as UPL. It's fine as it's classified as [Category A](https://www.apache.org/legal/resolved.html).
- https://mvnrepository.com/artifact/org.graalvm.js/js-scriptengine/23.0.8
- https://mvnrepository.com/artifact/org.graalvm.js/js/23.0.8

ql/src/java/org/apache/hadoop/hive/ql/metadata/PartitionTree.java
```diff
@@ -258,21 +254,20 @@ List<Partition> getPartitionsByFilter(final String filter) throws MetaException
...
-    for (Map.Entry<String, Partition> entry : parts.entrySet()) {
-      se.put("partitionName", entry.getKey());
-      se.put("values", entry.getValue().getValues());
-      try {
-        if ((Boolean)se.eval(filter)) {
-          result.add(entry.getValue());
+    try (GraalJSScriptEngine se = GraalJSScriptEngine.create(null,
+            Context.newBuilder().allowExperimentalOptions(true)
+                    .option("js.nashorn-compat", "true")
+                    .allowAllAccess(true))) {
```
okumin, line 260: I'm checking this document.
https://www.graalvm.org/latest/reference-manual/js/NashornMigrationGuide/

I guess `allowAllAccess(true)` is redundant: https://github.com/oracle/graaljs/blob/release/graal-vm/23.0/graal-js/src/com.oracle.truffle.js.scriptengine/src/com/oracle/truffle/js/scriptengine/GraalJSScriptEngine.java#L346

maheshrajus: @okumin thanks for the review.
i checked and debug the above code. so here we are setting _builder.allowAllAccess(true)_ in the method **updateForNashornCompatibilityMode()**. But this method call when contextConfig is null. But in our current case we are setting flags related to nashorn compatibility("js.nashorn-compat") in Context config and calling method GraalJSScriptEngine.create(). So we will not hit the null case(see code below) and the method(updateForNashornCompatibilityMode) will not call. So here setting "**allowAllAccess(true)**" in Context builder is ok in my opinion. Please let me know your opinion on same.  thank you !

```
        if (contextConfig == null) {      --> in our current case this will not be NULL.
            contextConfigToUse = Context.newBuilder(new String[]{"js"}).allowExperimentalOptions(true);
            contextConfigToUse.option("js.syntax-extensions", "true");
            contextConfigToUse.option("js.load", "true");
            contextConfigToUse.option("js.print", "true");
            contextConfigToUse.option("js.global-arguments", "true");
            contextConfigToUse.option("js.charset", "UTF-8");
            if (NASHORN_COMPATIBILITY_MODE) {
                updateForNashornCompatibilityMode(contextConfigToUse);
            } else if (Boolean.getBoolean("graaljs.insecure-scriptengine-access")) {
                updateForScriptEngineAccessibility(contextConfigToUse);
            }
        }
```
```
    private static void updateForNashornCompatibilityMode(Context.Builder builder) {
        builder.allowAllAccess(true);
        builder.allowHostAccess(NASHORN_HOST_ACCESS);
        builder.useSystemExit(true);
    }
```

okumin: You'll be right. What if `.allowAllAccess(true)` is removed? The document doesn't mention the option.

```
try (Context context = Context.newBuilder().allowExperimentalOptions(true).option("js.nashorn-compat", "true").build()) {
      context.eval("js", "print(__LINE__)");
  }
```

If GraalJS works with Hive without the option, it's better from the point of view of security.
https://www.graalvm.org/sdk/javadoc/org/graalvm/polyglot/Context.Builder.html#allowAllAccess(boolean)

maheshrajus: yeah, i removed property .allowAllAccess(true) and checked it previously, we are getting error "failed due to: Unknown identifier: get". So in our case we need to pass this field. thanks !

maheshrajus:
@okumin @ayushtkn Can you please review and approve the changes? thanks !

ayushtkn (approved):
LGTM

PR #5857, opened 2025-06-10
HIVE-29001: Remove duplicate mockito dependency in hive-common pom causing maven warnings

Aggarwal-Raghav:
Check [HIVE-29001](https://issues.apache.org/jira/browse/HIVE-29001), duplicate dependency of mockito throwing maven warning during build

To have a clean maven build output :-)

NO

Will see CI output

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -255,12 +255,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-      <version>${mockito-core.version}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>


common/pom.xml
```diff
@@ -254,12 +254,6 @@
       <artifactId>tempus-fugit</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
```
Aggarwal-Raghav, line 259: The same dependency is present at 
https://github.com/apache/hive/blob/39c88d7c1742b6ab0c87d58d333d81aefedeba3d/common/pom.xml#L362

Aggarwal-Raghav:
CC @ayushtkn.

Aggarwal-Raghav:
@ayushtkn, quick question unrelated to the PR.
For log4j 2.24.3, the changes to remove packages attribute due to LOG4J2-3644 was done only in [data/conf/hive-log4j2.properties](https://github.com/apache/hive/pull/5636/files#diff-d003b0ed1a14718be707a0b4e196ca3c2f9dc7adad2ca10d4868946aac53a178). 

But in hive tarball which we ship, log4j properties is picked from [_common/src/main/resources/hive-log4j2.properties_](https://github.com/apache/hive/blob/39c88d7c1742b6ab0c87d58d333d81aefedeba3d/packaging/src/main/assembly/bin.xml#L418) which doesn't contain the log4j cleanup changes. Should we address this?

Reason why I am asking is, if i use default log4j properties present in tarball then on starting hive services i get the following warning:

image

CC @InvisibleProgrammer

PR #5860, opened 2025-06-12
HIVE-29007 Fix NPE in TezSessionPoolManager

zratkai:
Fix an NPE.

Fix an NPE.

No.

Manually.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManager.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManager.java
@@ -119,7 +119,7 @@ public class TezSessionPoolManager extends TezSessionPoolSession.AbstractTrigger
     if (resourcePlan != null) {
       Collection<String> appliedTriggers = updateTriggers(resourcePlan);
       LOG.info("Updated tez session pool manager with triggers {} from active resource plan: {}",
-          appliedTriggers, resourcePlan.getPlan().getName());
+          appliedTriggers, resourcePlan.getPlan() == null ? "null" : resourcePlan.getPlan().getName());
     }
     metrics.start(conf);
   }


kasakrisz (approved):
+1 pending tests

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManager.java
```diff
@@ -119,7 +119,7 @@ public void startPool(HiveConf conf, final WMFullResourcePlan resourcePlan) thro
     if (resourcePlan != null) {
       Collection<String> appliedTriggers = updateTriggers(resourcePlan);
       LOG.info("Updated tez session pool manager with triggers {} from active resource plan: {}",
-          appliedTriggers, resourcePlan.getPlan().getName());
+          appliedTriggers, resourcePlan == null || resourcePlan.getPlan() == null ? "null" : resourcePlan.getPlan().getName());
```
Indhumathi27: resourcePlan == null check won't be applicable as it is already in the if block line:119

zratkai: Good catch, refactored.

zratkai: Good catch, refactored.

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/TezSessionPoolManager.java
```diff
@@ -119,7 +119,7 @@ public void startPool(HiveConf conf, final WMFullResourcePlan resourcePlan) thro
     if (resourcePlan != null) {
       Collection<String> appliedTriggers = updateTriggers(resourcePlan);
       LOG.info("Updated tez session pool manager with triggers {} from active resource plan: {}",
-          appliedTriggers, resourcePlan.getPlan().getName());
+          appliedTriggers, resourcePlan.getPlan() == null ? "null" : resourcePlan.getPlan().getName());
```
ayushtkn, line 122: Are we sure `resourcePlan` can't be `null`?
Looking at the Hiveserver2 code, it does check for `resourcePlan` as well
https://github.com/apache/hive/blob/98c0a1a047e9c32b8418eee5c53d9413c0ac428d/service/src/java/org/apache/hive/service/server/HiveServer2.java#L933

and 
https://github.com/apache/hive/blob/98c0a1a047e9c32b8418eee5c53d9413c0ac428d/service/src/java/org/apache/hive/service/server/HiveServer2.java#L952

zratkai: In line No. 119 we already check that:
if (resourcePlan != null) {...

PR #5862, opened 2025-06-12
HIVE-29008: CI logs do not display information about test execution

zabetak:
1. Disable maven's quiet mode (-q) and use default logging for running tests
2. Remove Surefire's logger configuration since defaults should work just fine

Display INFO logs from the tests that are running

No, only CI output is affected.

Manually inspect logs from the CI run.


zhangbutao:
BTW, Will this test execution log output increase the duration of the CI execution? I hope not.

zabetak:
Thanks for the review @zhangbutao ! 

I posted some details comments about execution time and disk space under HIVE-29008. Let's continue the discussion there if needed.

PR #5866, opened 2025-06-13
HIVE-29013: [ICEBERG] Data type info is not correct in Error message for tinyint, smallint

arorasimran0309:
This is to fix the wrong data type logged in the error, when table is created with smallint, tinyint datatypes

wrong data type in error message

Yes

Existing tests, and one updated test testCreateTableWithNotSupportedTypes()


arorasimran0309:
Thanks for the review @deniskuzZ

PR #5868, opened 2025-06-13
HIVE-29033: ORC reader should not assume that TimestampColumnVector is in UTC time zone

vrozov:
While setting `TimestampWritableV2` result from `TimestampColumnVector` use either UTC or local time zone depending on `TimestampColumnVector.isUTC()` flag.

While Hive assumes that `isUTC` is always true, Spark uses local time zone

No

Using existing tests in Hive and Spark

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
@@ -20,8 +20,11 @@ package org.apache.hadoop.hive.ql.io;
 
 import com.google.common.collect.Lists;
 
+import org.apache.hadoop.hive.common.type.Timestamp;
 import org.apache.hadoop.hive.llap.DebugUtils;
 
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.Arrays;
 
 import org.slf4j.Logger;
@@ -518,7 +521,8 @@ public abstract class BatchToRowReader<StructType, UnionType>
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      result.set(Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
+          tcv.isUTC() ? ZoneOffset.UTC : ZoneId.systemDefault()));
       return result;
     } else {
       return null;
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
@@ -23,8 +23,6 @@ import com.google.common.collect.Lists;
 import org.apache.hadoop.hive.common.type.Timestamp;
 import org.apache.hadoop.hive.llap.DebugUtils;
 
-import java.time.ZoneId;
-import java.time.ZoneOffset;
 import java.util.Arrays;
 
 import org.slf4j.Logger;
@@ -520,9 +518,7 @@ public abstract class BatchToRowReader<StructType, UnionType>
       } else {
         result = (TimestampWritableV2) previous;
       }
-      TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.set(Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
-          tcv.isUTC() ? ZoneOffset.UTC : ZoneId.systemDefault()));
+      result.set(Timestamp.from((TimestampColumnVector) vector, row));
       return result;
     } else {
       return null;
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
@@ -18,12 +18,15 @@
 package org.apache.hadoop.hive.ql.io.orc;
 
 import java.io.IOException;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.common.type.Timestamp;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.ql.exec.vector.BytesColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
@@ -72,6 +75,12 @@ public class RecordReaderImpl extends org.apache.orc.impl.RecordReaderImpl
       batch = this.schema.createRowBatchV2();
     } else {
       batch = this.schema.createRowBatch();
+    }
+    // Workaround for ORC not initializing TimestampColumnVector properly
+    for (int i = 0; i < batch.numCols; ++i) {
+      if (batch.cols[i] instanceof TimestampColumnVector) {
+        ((TimestampColumnVector) batch.cols[i]).setIsUTC(fileReader.isUTC());
+      }
     }
     rowInBatch = 0;
   }
@@ -458,7 +467,8 @@ public class RecordReaderImpl extends org.apache.orc.impl.RecordReaderImpl
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      result.set(Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
+          tcv.isUTC() ? ZoneOffset.UTC : ZoneId.systemDefault()));
       return result;
     } else {
       return null;
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
@@ -18,8 +18,6 @@
 package org.apache.hadoop.hive.ql.io.orc;
 
 import java.io.IOException;
-import java.time.ZoneId;
-import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -464,9 +462,7 @@ public class RecordReaderImpl extends org.apache.orc.impl.RecordReaderImpl
       } else {
         result = (TimestampWritableV2) previous;
       }
-      TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.set(Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
-          tcv.isUTC() ? ZoneOffset.UTC : ZoneId.systemDefault()));
+      result.set(Timestamp.from((TimestampColumnVector) vector, row));
       return result;
     } else {
       return null;
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
@@ -65,23 +65,16 @@ public class RecordReaderImpl extends org.apache.orc.impl.RecordReaderImpl
   private final VectorizedRowBatch batch;
   private int rowInBatch;
   private long baseRow;
+  private final boolean useUTC;
 
   protected RecordReaderImpl(ReaderImpl fileReader,
     Reader.Options options, final Configuration conf) throws IOException {
     super(fileReader, options);
+    useUTC = fileReader.isUTC();
     final boolean useDecimal64ColumnVectors = conf != null && HiveConf.getVar(conf,
       HiveConf.ConfVars.HIVE_VECTORIZED_INPUT_FORMAT_SUPPORTS_ENABLED).equalsIgnoreCase("decimal_64");
-    if (useDecimal64ColumnVectors){
-      batch = this.schema.createRowBatchV2();
-    } else {
-      batch = this.schema.createRowBatch();
-    }
     // Workaround for ORC not initializing TimestampColumnVector properly
-    for (int i = 0; i < batch.numCols; ++i) {
-      if (batch.cols[i] instanceof TimestampColumnVector) {
-        ((TimestampColumnVector) batch.cols[i]).setIsUTC(fileReader.isUTC());
-      }
-    }
+    batch = createRowBatch(useDecimal64ColumnVectors);
     rowInBatch = 0;
   }
 
@@ -100,7 +93,12 @@ public class RecordReaderImpl extends org.apache.orc.impl.RecordReaderImpl
   }
 
   public VectorizedRowBatch createRowBatch(boolean useDecimal64) {
-    return useDecimal64 ? this.schema.createRowBatchV2() : this.schema.createRowBatch();
+    TimestampColumnVector.setUseUTC(useUTC);
+    try {
+      return useDecimal64 ? this.schema.createRowBatchV2() : this.schema.createRowBatch();
+    } finally {
+      TimestampColumnVector.setUseUTC(null);
+    }
   }
 
   @Override
--- a/storage-api/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampColumnVector.java
+++ b/storage-api/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampColumnVector.java
@@ -40,6 +40,15 @@ import org.apache.hive.common.util.SuppressFBWarnings;
  * using the scratch timestamp, and then perhaps update the column vector row with a result.
  */
 public class TimestampColumnVector extends ColumnVector {
+  private static final ThreadLocal<Boolean> useUTC = ThreadLocal.withInitial(() -> Boolean.FALSE);
+
+  public static void setUseUTC(Boolean isUTC) {
+    if (isUTC != null) {
+      useUTC.set(isUTC);
+    } else {
+      useUTC.remove();
+    }
+  }
 
   /*
    * The storage arrays for this column vector corresponds to the storage of a Timestamp:
@@ -85,7 +94,7 @@ public class TimestampColumnVector extends ColumnVector {
 
     scratchWritable = null;     // Allocated by caller.
 
-    isUTC = false;
+    isUTC = useUTC.get();
   }
 
   /**
--- a/common/src/java/org/apache/hadoop/hive/common/type/Timestamp.java
+++ b/common/src/java/org/apache/hadoop/hive/common/type/Timestamp.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.hive.common.type;
 
+import org.apache.hadoop.hive.ql.exec.vector.TimestampColumnVector;
 import org.apache.hive.common.util.SuppressFBWarnings;
 
 import java.time.DateTimeException;
@@ -235,6 +236,11 @@ public class Timestamp implements Comparable<Timestamp> {
     return new Timestamp(LocalDateTime
         .ofInstant(Instant.ofEpochMilli(epochMilli), ZoneOffset.UTC)
         .withNano(nanos));
+  }
+
+  public static Timestamp from(TimestampColumnVector tcv, int row) {
+    return Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
+        tcv.isUTC() ? ZoneOffset.UTC : ZoneId.systemDefault());
   }
 
   public void setNanos(int nanos) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java
@@ -35,6 +35,10 @@ public class ReaderImpl extends org.apache.orc.impl.ReaderImpl
 
   private final ObjectInspector inspector;
 
+  boolean isUTC() {
+    return useUTCTimestamp;
+  }
+  
   @Override
   public ObjectInspector getObjectInspector() {
     return inspector;


vrozov:
@deniskuzZ Please review.

simhadri-g:
@ayushtkn, @zhangbutao and @deniskuzZ . This change seems to be needed for hive 4 upgrade in spark, can we please included this in 4.1.0 ?

Thanks!

zhangbutao:
> @ayushtkn, @zhangbutao and @deniskuzZ . This change seems to be needed for hive 4 upgrade in spark, can we please included this in 4.1.0 ?
> 
> Thanks!

@simhadri-g I don't quite understand the context of this PR. Will this issue block the upgrade of Hive4 dependencies in Spark?

From the description `While Hive assumes that isUTC is always true, Spark uses local time zone`, I'm not sure if this can be defined as a problem. 
BTW, Can this issue be fixed from the Spark side?

I'm not very familiar with this piece of code. If someone has a thorough understanding of the fix, i am ok to include it to 4.1.0. But  personally i prefer to incorporate this code into the next version to avoid any possible impact for the release 4.1.0.

Thanks.

vrozov:
@zhangbutao
> Will this issue block the upgrade of Hive4 dependencies in Spark?

Yes, it is a blocking issue as it introduces regression into Apache Spark.

> BTW, Can this issue be fixed from the Spark side?

No, I explored multiple options and the issue is not fixable on the Spark side as the problem is during conversion from `TimestampColumnVector` to `TimestampWritableV2` on the Hive side.

ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
```diff
@@ -518,7 +521,8 @@ public static TimestampWritableV2 nextTimestamp(ColumnVector vector,
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      result.set(Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
+          tcv.isUTC() ? ZoneOffset.UTC : ZoneId.systemDefault()));
```
deniskuzZ: isn't `TimestampColumnVector.isUTC()` always false in BatchToRowReader?

ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
```diff
@@ -73,6 +76,12 @@ protected RecordReaderImpl(ReaderImpl fileReader,
     } else {
       batch = this.schema.createRowBatch();
     }
+    // Workaround for ORC not initializing TimestampColumnVector properly
+    for (int i = 0; i < batch.numCols; ++i) {
+      if (batch.cols[i] instanceof TimestampColumnVector) {
+        ((TimestampColumnVector) batch.cols[i]).setIsUTC(fileReader.isUTC());
```
deniskuzZ: do we need the same workaround for non-orc readers like parquet?

ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
vrozov: @deniskuzZ I don't see why it is always the case, can you please explain? Even if it is always `false` now, would it be safer to handle both options?

deniskuzZ: `TimestampColumnVector.setIsUTC` is only called from the `org.apache.hadoop.hive.ql.io.orc.WriterImpl` and the default `isUTC`=false. So readers should be expecting the Timestamp values in UTC.

deniskuzZ: sorry, I was wrong, it's set in `org.apache.orc.impl.TreeReaderFactory`
````
public void nextVector(ColumnVector previousVector, boolean[] isNull, int batchSize, FilterContext filterContext, TypeReader.ReadPhase readPhase) throws IOException {
            TimestampColumnVector result = (TimestampColumnVector)previousVector;
            result.changeCalendar(this.fileUsesProleptic, false);
            super.nextVector(previousVector, isNull, batchSize, filterContext, readPhase);
            result.setIsUTC(this.context.getUseUTCTimestamp());
            if (filterContext.isSelectedInUse()) {
                this.nextVector(result, isNull, filterContext, batchSize);
            } else {
                this.nextVector(result, isNull, batchSize);
            }
        }
````

File: ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java (hunk 2 above)
Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java]: I think that it is only necessary for ORC readers. Note that ORC `TimestampTreeReader` sets the UTC flag in https://github.com/apache/orc/blob/29c1b37955e3423681304c6e9eb3fe7884ba57bc/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L1364, so the workaround is necessary because `TimestampFromAnyIntegerTreeReader` and similar ORC `ConvertTreeReader` don't set the flag correctly.

File: ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java (hunk 1 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: i would expect vectorized scenario handled same way, without additional hacks

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: correct, note that for Hive, `context.getUseUTCTimestamp()` is `true` while Spark sets it to `false`.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: @vrozov, can you refactor
````
 protected RecordReaderImpl(ReaderImpl fileReader,
    Reader.Options options, final Configuration conf) throws IOException {
    super(fileReader, options);
    ...
    context = new ReaderContext(fileReader.isUTC());
}

private static class ReaderContext  {
  private boolean useUTCTimestamp = false;

  private ReaderContext(boolean useUTCTimestamp) {
    this.useUTCTimestamp = useUTCTimestamp;
  }

  public boolean getUseUTCTimestamp() {
    return this.useUTCTimestamp;
  }
}

static Object nextValue(ColumnVector vector,
                          int row,
                          TypeDescription schema,
                          ReaderContext context,
                          Object previous) {

      case TIMESTAMP:
        return nextTimestamp(vector, row, context, previous); 
}

context.getUseUTCTimestamp()? ZoneOffset.UTC : ZoneId.systemDefault()
````

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: attached the diff for `RecordReaderImpl` with extra cleanup. try applying to the original version of the class  without your changes
````
Subject: [PATCH] refactor
---
Index: ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java	(date 1750795481725)
@@ -18,12 +18,15 @@
 package org.apache.hadoop.hive.ql.io.orc;
 
 import java.io.IOException;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.hadoop.hive.common.type.Timestamp;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.ql.exec.vector.BytesColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
@@ -53,15 +56,13 @@
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Text;
 import org.apache.orc.TypeDescription;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class RecordReaderImpl extends org.apache.orc.impl.RecordReaderImpl
                               implements RecordReader {
-  static final Logger LOG = LoggerFactory.getLogger(RecordReaderImpl.class);
   private final VectorizedRowBatch batch;
   private int rowInBatch;
   private long baseRow;
+  private ReaderContext context;
 
   protected RecordReaderImpl(ReaderImpl fileReader,
     Reader.Options options, final Configuration conf) throws IOException {
@@ -73,10 +74,23 @@
     } else {
       batch = this.schema.createRowBatch();
     }
+    context = new ReaderContext(fileReader.isUTC());
     rowInBatch = 0;
   }
 
-  /**
+  private static class ReaderContext  {
+    private boolean useUTCTimestamp;
+
+    private ReaderContext(boolean useUTCTimestamp) {
+      this.useUTCTimestamp = useUTCTimestamp;
+    }
+
+    public boolean getUseUTCTimestamp() {
+      return this.useUTCTimestamp;
+    }
+  }
+
+    /**
    * If the current batch is empty, get a new one.
    * @return true if we have rows available.
    * @throws IOException
@@ -135,10 +149,10 @@
       }
       for(int i=0; i < numberOfChildren; ++i) {
         result.setFieldValue(i, nextValue(batch.cols[i], rowInBatch,
-            children.get(i), result.getFieldValue(i)));
+            children.get(i), context, result.getFieldValue(i)));
       }
     } else {
-      previous = nextValue(batch.cols[0], rowInBatch, schema, previous);
+      previous = nextValue(batch.cols[0], rowInBatch, schema, context, previous);
     }
     rowInBatch += 1;
     return previous;
@@ -446,6 +460,7 @@
 
   static TimestampWritableV2 nextTimestamp(ColumnVector vector,
                                            int row,
+                                           ReaderContext context,
                                            Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -458,7 +473,8 @@
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      result.set(Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
+          context.getUseUTCTimestamp() ? ZoneOffset.UTC : ZoneId.systemDefault()));
       return result;
     } else {
       return null;
@@ -468,6 +484,7 @@
   static OrcStruct nextStruct(ColumnVector vector,
                               int row,
                               TypeDescription schema,
+                              ReaderContext context,
                               Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -485,7 +502,7 @@
       StructColumnVector struct = (StructColumnVector) vector;
       for(int f=0; f < numChildren; ++f) {
         result.setFieldValue(f, nextValue(struct.fields[f], row,
-            childrenTypes.get(f), result.getFieldValue(f)));
+            childrenTypes.get(f), context, result.getFieldValue(f)));
       }
       return result;
     } else {
@@ -496,6 +513,7 @@
   static OrcUnion nextUnion(ColumnVector vector,
                             int row,
                             TypeDescription schema,
+                            ReaderContext context,
                             Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -511,7 +529,7 @@
       UnionColumnVector union = (UnionColumnVector) vector;
       byte tag = (byte) union.tags[row];
       result.set(tag, nextValue(union.fields[tag], row, childrenTypes.get(tag),
-          result.getObject()));
+          context, result.getObject()));
       return result;
     } else {
       return null;
@@ -521,6 +539,7 @@
   static ArrayList<Object> nextList(ColumnVector vector,
                                     int row,
                                     TypeDescription schema,
+                                    ReaderContext context,
                                     Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -541,14 +560,14 @@
       TypeDescription childType = schema.getChildren().get(0);
       while (idx < length && idx < oldLength) {
         result.set(idx, nextValue(list.child, offset + idx, childType,
-            result.get(idx)));
+            context, result.get(idx)));
         idx += 1;
       }
       if (length < oldLength) {
         result.subList(length,result.size()).clear();
       } else if (oldLength < length) {
         while (idx < length) {
-          result.add(nextValue(list.child, offset + idx, childType, null));
+          result.add(nextValue(list.child, offset + idx, childType, context, null));
           idx += 1;
         }
       }
@@ -561,6 +580,7 @@
   static Map<Object,Object> nextMap(ColumnVector vector,
                                     int row,
                                     TypeDescription schema,
+                                    ReaderContext context,
                                     Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -573,7 +593,7 @@
       TypeDescription valueType = schema.getChildren().get(1);
       LinkedHashMap<Object,Object> result;
       if (previous == null || previous.getClass() != LinkedHashMap.class) {
-        result = new LinkedHashMap<Object,Object>(length);
+        result = new LinkedHashMap<>(length);
       } else {
         result = (LinkedHashMap<Object,Object>) previous;
         // I couldn't think of a good way to reuse the keys and value objects
@@ -581,8 +601,8 @@
         result.clear();
       }
       for(int e=0; e < length; ++e) {
-        result.put(nextValue(map.keys, e + offset, keyType, null),
-                   nextValue(map.values, e + offset, valueType, null));
+        result.put(nextValue(map.keys, e + offset, keyType, context, null),
+                   nextValue(map.values, e + offset, valueType, context, null));
       }
       return result;
     } else {
@@ -593,47 +613,29 @@
   static Object nextValue(ColumnVector vector,
                           int row,
                           TypeDescription schema,
+                          ReaderContext context,
                           Object previous) {
-    switch (schema.getCategory()) {
-      case BOOLEAN:
-        return nextBoolean(vector, row, previous);
-      case BYTE:
-        return nextByte(vector, row, previous);
-      case SHORT:
-        return nextShort(vector, row, previous);
-      case INT:
-        return nextInt(vector, row, previous);
-      case LONG:
-        return nextLong(vector, row, previous);
-      case FLOAT:
-        return nextFloat(vector, row, previous);
-      case DOUBLE:
-        return nextDouble(vector, row, previous);
-      case STRING:
-        return nextString(vector, row, previous);
-      case CHAR:
-        return nextChar(vector, row, schema.getMaxLength(), previous);
-      case VARCHAR:
-        return nextVarchar(vector, row, schema.getMaxLength(), previous);
-      case BINARY:
-        return nextBinary(vector, row, previous);
-      case DECIMAL:
-        return nextDecimal(vector, row, previous);
-      case DATE:
-        return nextDate(vector, row, previous);
-      case TIMESTAMP:
-        return nextTimestamp(vector, row, previous);
-      case STRUCT:
-        return nextStruct(vector, row, schema, previous);
-      case UNION:
-        return nextUnion(vector, row, schema, previous);
-      case LIST:
-        return nextList(vector, row, schema, previous);
-      case MAP:
-        return nextMap(vector, row, schema, previous);
-      default:
-        throw new IllegalArgumentException("Unknown type " + schema);
-    }
+      return switch (schema.getCategory()) {
+          case BOOLEAN -> nextBoolean(vector, row, previous);
+          case BYTE -> nextByte(vector, row, previous);
+          case SHORT -> nextShort(vector, row, previous);
+          case INT -> nextInt(vector, row, previous);
+          case LONG -> nextLong(vector, row, previous);
+          case FLOAT -> nextFloat(vector, row, previous);
+          case DOUBLE -> nextDouble(vector, row, previous);
+          case STRING -> nextString(vector, row, previous);
+          case CHAR -> nextChar(vector, row, schema.getMaxLength(), previous);
+          case VARCHAR -> nextVarchar(vector, row, schema.getMaxLength(), previous);
+          case BINARY -> nextBinary(vector, row, previous);
+          case DECIMAL -> nextDecimal(vector, row, previous);
+          case DATE -> nextDate(vector, row, previous);
+          case TIMESTAMP -> nextTimestamp(vector, row, context, previous);
+          case STRUCT -> nextStruct(vector, row, schema, context, previous);
+          case UNION -> nextUnion(vector, row, schema, context, previous);
+          case LIST -> nextList(vector, row, schema, context, previous);
+          case MAP -> nextMap(vector, row, schema, context, previous);
+          default -> throw new IllegalArgumentException("Unknown type " + schema);
+      };
   }
 
   /* Routines for copying between VectorizedRowBatches */
@@ -789,19 +791,11 @@
     castedDestination.noNulls = castedSource.noNulls;
     if (source.isRepeating) {
       castedDestination.isNull[0] = castedSource.isNull[0];
-      for(int c=0; c > castedSource.fields.length; ++c) {
-        copyColumn(castedDestination.fields[c], castedSource.fields[c], 0, 1);
-      }
     } else {
       if (!castedSource.noNulls) {
         for (int r = 0; r < length; ++r) {
           castedDestination.isNull[r] = castedSource.isNull[sourceOffset + r];
         }
-      } else {
-        for (int c = 0; c > castedSource.fields.length; ++c) {
-          copyColumn(castedDestination.fields[c], castedSource.fields[c],
-              sourceOffset, length);
-        }
       }
     }
   }
@@ -829,14 +823,10 @@
           castedDestination.tags[r] = castedSource.tags[sourceOffset + r];
         }
       } else {
-        for(int r=0; r < length; ++r) {
+        for (int r=0; r < length; ++r) {
           castedDestination.tags[r] = castedSource.tags[sourceOffset + r];
         }
       }
-      for(int c=0; c > castedSource.fields.length; ++c) {
-        copyColumn(castedDestination.fields[c], castedSource.fields[c],
-            sourceOffset, length);
-      }
     }
   }
 
@@ -954,7 +944,6 @@
    * @param destination the batch to copy into
    * @param source the batch to copy from
    * @param sourceStart the row number to start from in the source
-   * @return the number of rows copied
    */
   void copyIntoBatch(VectorizedRowBatch destination,
                      VectorizedRowBatch source,

````

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: Sorry, but I don't see why is it necessary to use a different flag. `TimestampColumnVector` already has the flag that identifies whether it is in UTC zone or local zone.

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > i would expect vectorized scenario handled same way, without additional hacks

I don't follow. The additional workaround is for type conversion only.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: the way you pass the UTC flag is hacky and not scalable for future use. updated the diff with all the changes, please take a look
````
Subject: [PATCH] refactor
---
Index: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java	(date 1750801652028)
@@ -18,22 +18,15 @@
 
 package org.apache.hadoop.hive.ql.exec.vector;
 
-import java.util.concurrent.TimeUnit;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 
 import org.apache.hadoop.hive.common.type.Timestamp;
-import org.apache.hadoop.hive.serde2.io.DateWritableV2;
 import org.apache.hadoop.hive.serde2.io.HiveIntervalDayTimeWritable;
 import org.apache.hadoop.hive.serde2.io.TimestampWritableV2;
 
 public final class TimestampUtils {
 
-  static final long MILLISECONDS_PER_SECOND = TimeUnit.SECONDS.toMillis(1);
-  static final long NANOSECONDS_PER_MILLISECOND = TimeUnit.MILLISECONDS.toNanos(1);
-
-  public static long daysToNanoseconds(long daysSinceEpoch) {
-    return DateWritableV2.daysToMillis((int) daysSinceEpoch) * NANOSECONDS_PER_MILLISECOND;
-  }
-
   public static TimestampWritableV2 timestampColumnVectorWritable(
       TimestampColumnVector timestampColVector, int elementNum,
       TimestampWritableV2 timestampWritable) {
@@ -63,4 +56,9 @@
     }
     return o.toString();
   }
+
+  public static Timestamp timestamp(TimestampColumnVector tcv, int row) {
+    return Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
+        tcv.isUTC() ? ZoneOffset.UTC : ZoneId.systemDefault());
+  }
 }
Index: ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java b/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java	(date 1750801652033)
@@ -44,6 +44,7 @@
 import org.apache.hadoop.hive.ql.exec.vector.MapColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.StructColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.TimestampColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.TimestampUtils;
 import org.apache.hadoop.hive.ql.exec.vector.UnionColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatchCtx;
@@ -518,7 +519,7 @@
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      result.set(TimestampUtils.timestamp(tcv, row));
       return result;
     } else {
       return null;
Index: ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java	(date 1750801652025)
@@ -36,6 +36,7 @@
 import org.apache.hadoop.hive.ql.exec.vector.MapColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.StructColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.TimestampColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.TimestampUtils;
 import org.apache.hadoop.hive.ql.exec.vector.UnionColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.hadoop.hive.serde2.io.ByteWritable;
@@ -53,15 +54,13 @@
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Text;
 import org.apache.orc.TypeDescription;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class RecordReaderImpl extends org.apache.orc.impl.RecordReaderImpl
                               implements RecordReader {
-  static final Logger LOG = LoggerFactory.getLogger(RecordReaderImpl.class);
   private final VectorizedRowBatch batch;
   private int rowInBatch;
   private long baseRow;
+  private ReaderContext context;
 
   protected RecordReaderImpl(ReaderImpl fileReader,
     Reader.Options options, final Configuration conf) throws IOException {
@@ -73,10 +72,23 @@
     } else {
       batch = this.schema.createRowBatch();
     }
+    context = new ReaderContext(fileReader.isUTC());
     rowInBatch = 0;
   }
 
-  /**
+  private static class ReaderContext  {
+    private boolean useUTCTimestamp;
+
+    private ReaderContext(boolean useUTCTimestamp) {
+      this.useUTCTimestamp = useUTCTimestamp;
+    }
+
+    public boolean getUseUTCTimestamp() {
+      return this.useUTCTimestamp;
+    }
+  }
+
+    /**
    * If the current batch is empty, get a new one.
    * @return true if we have rows available.
    * @throws IOException
@@ -135,10 +147,10 @@
       }
       for(int i=0; i < numberOfChildren; ++i) {
         result.setFieldValue(i, nextValue(batch.cols[i], rowInBatch,
-            children.get(i), result.getFieldValue(i)));
+            children.get(i), context, result.getFieldValue(i)));
       }
     } else {
-      previous = nextValue(batch.cols[0], rowInBatch, schema, previous);
+      previous = nextValue(batch.cols[0], rowInBatch, schema, context, previous);
     }
     rowInBatch += 1;
     return previous;
@@ -446,6 +458,7 @@
 
   static TimestampWritableV2 nextTimestamp(ColumnVector vector,
                                            int row,
+                                           ReaderContext context,
                                            Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -458,7 +471,8 @@
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      tcv.setIsUTC(context.getUseUTCTimestamp());
+      result.set(TimestampUtils.timestamp(tcv, row));
       return result;
     } else {
       return null;
@@ -468,6 +482,7 @@
   static OrcStruct nextStruct(ColumnVector vector,
                               int row,
                               TypeDescription schema,
+                              ReaderContext context,
                               Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -485,7 +500,7 @@
       StructColumnVector struct = (StructColumnVector) vector;
       for(int f=0; f < numChildren; ++f) {
         result.setFieldValue(f, nextValue(struct.fields[f], row,
-            childrenTypes.get(f), result.getFieldValue(f)));
+            childrenTypes.get(f), context, result.getFieldValue(f)));
       }
       return result;
     } else {
@@ -496,6 +511,7 @@
   static OrcUnion nextUnion(ColumnVector vector,
                             int row,
                             TypeDescription schema,
+                            ReaderContext context,
                             Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -511,7 +527,7 @@
       UnionColumnVector union = (UnionColumnVector) vector;
       byte tag = (byte) union.tags[row];
       result.set(tag, nextValue(union.fields[tag], row, childrenTypes.get(tag),
-          result.getObject()));
+          context, result.getObject()));
       return result;
     } else {
       return null;
@@ -521,6 +537,7 @@
   static ArrayList<Object> nextList(ColumnVector vector,
                                     int row,
                                     TypeDescription schema,
+                                    ReaderContext context,
                                     Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -541,14 +558,14 @@
       TypeDescription childType = schema.getChildren().get(0);
       while (idx < length && idx < oldLength) {
         result.set(idx, nextValue(list.child, offset + idx, childType,
-            result.get(idx)));
+            context, result.get(idx)));
         idx += 1;
       }
       if (length < oldLength) {
         result.subList(length,result.size()).clear();
       } else if (oldLength < length) {
         while (idx < length) {
-          result.add(nextValue(list.child, offset + idx, childType, null));
+          result.add(nextValue(list.child, offset + idx, childType, context, null));
           idx += 1;
         }
       }
@@ -561,6 +578,7 @@
   static Map<Object,Object> nextMap(ColumnVector vector,
                                     int row,
                                     TypeDescription schema,
+                                    ReaderContext context,
                                     Object previous) {
     if (vector.isRepeating) {
       row = 0;
@@ -573,7 +591,7 @@
       TypeDescription valueType = schema.getChildren().get(1);
       LinkedHashMap<Object,Object> result;
       if (previous == null || previous.getClass() != LinkedHashMap.class) {
-        result = new LinkedHashMap<Object,Object>(length);
+        result = new LinkedHashMap<>(length);
       } else {
         result = (LinkedHashMap<Object,Object>) previous;
         // I couldn't think of a good way to reuse the keys and value objects
@@ -581,8 +599,8 @@
         result.clear();
       }
       for(int e=0; e < length; ++e) {
-        result.put(nextValue(map.keys, e + offset, keyType, null),
-                   nextValue(map.values, e + offset, valueType, null));
+        result.put(nextValue(map.keys, e + offset, keyType, context, null),
+                   nextValue(map.values, e + offset, valueType, context, null));
       }
       return result;
     } else {
@@ -593,47 +611,29 @@
   static Object nextValue(ColumnVector vector,
                           int row,
                           TypeDescription schema,
+                          ReaderContext context,
                           Object previous) {
-    switch (schema.getCategory()) {
-      case BOOLEAN:
-        return nextBoolean(vector, row, previous);
-      case BYTE:
-        return nextByte(vector, row, previous);
-      case SHORT:
-        return nextShort(vector, row, previous);
-      case INT:
-        return nextInt(vector, row, previous);
-      case LONG:
-        return nextLong(vector, row, previous);
-      case FLOAT:
-        return nextFloat(vector, row, previous);
-      case DOUBLE:
-        return nextDouble(vector, row, previous);
-      case STRING:
-        return nextString(vector, row, previous);
-      case CHAR:
-        return nextChar(vector, row, schema.getMaxLength(), previous);
-      case VARCHAR:
-        return nextVarchar(vector, row, schema.getMaxLength(), previous);
-      case BINARY:
-        return nextBinary(vector, row, previous);
-      case DECIMAL:
-        return nextDecimal(vector, row, previous);
-      case DATE:
-        return nextDate(vector, row, previous);
-      case TIMESTAMP:
-        return nextTimestamp(vector, row, previous);
-      case STRUCT:
-        return nextStruct(vector, row, schema, previous);
-      case UNION:
-        return nextUnion(vector, row, schema, previous);
-      case LIST:
-        return nextList(vector, row, schema, previous);
-      case MAP:
-        return nextMap(vector, row, schema, previous);
-      default:
-        throw new IllegalArgumentException("Unknown type " + schema);
-    }
+      return switch (schema.getCategory()) {
+          case BOOLEAN -> nextBoolean(vector, row, previous);
+          case BYTE -> nextByte(vector, row, previous);
+          case SHORT -> nextShort(vector, row, previous);
+          case INT -> nextInt(vector, row, previous);
+          case LONG -> nextLong(vector, row, previous);
+          case FLOAT -> nextFloat(vector, row, previous);
+          case DOUBLE -> nextDouble(vector, row, previous);
+          case STRING -> nextString(vector, row, previous);
+          case CHAR -> nextChar(vector, row, schema.getMaxLength(), previous);
+          case VARCHAR -> nextVarchar(vector, row, schema.getMaxLength(), previous);
+          case BINARY -> nextBinary(vector, row, previous);
+          case DECIMAL -> nextDecimal(vector, row, previous);
+          case DATE -> nextDate(vector, row, previous);
+          case TIMESTAMP -> nextTimestamp(vector, row, context, previous);
+          case STRUCT -> nextStruct(vector, row, schema, context, previous);
+          case UNION -> nextUnion(vector, row, schema, context, previous);
+          case LIST -> nextList(vector, row, schema, context, previous);
+          case MAP -> nextMap(vector, row, schema, context, previous);
+          default -> throw new IllegalArgumentException("Unknown type " + schema);
+      };
   }
 
   /* Routines for copying between VectorizedRowBatches */
@@ -789,19 +789,11 @@
     castedDestination.noNulls = castedSource.noNulls;
     if (source.isRepeating) {
       castedDestination.isNull[0] = castedSource.isNull[0];
-      for(int c=0; c > castedSource.fields.length; ++c) {
-        copyColumn(castedDestination.fields[c], castedSource.fields[c], 0, 1);
-      }
     } else {
       if (!castedSource.noNulls) {
         for (int r = 0; r < length; ++r) {
           castedDestination.isNull[r] = castedSource.isNull[sourceOffset + r];
         }
-      } else {
-        for (int c = 0; c > castedSource.fields.length; ++c) {
-          copyColumn(castedDestination.fields[c], castedSource.fields[c],
-              sourceOffset, length);
-        }
       }
     }
   }
@@ -829,14 +821,10 @@
           castedDestination.tags[r] = castedSource.tags[sourceOffset + r];
         }
       } else {
-        for(int r=0; r < length; ++r) {
+        for (int r=0; r < length; ++r) {
           castedDestination.tags[r] = castedSource.tags[sourceOffset + r];
         }
       }
-      for(int c=0; c > castedSource.fields.length; ++c) {
-        copyColumn(castedDestination.fields[c], castedSource.fields[c],
-            sourceOffset, length);
-      }
     }
   }
 
@@ -954,7 +942,6 @@
    * @param destination the batch to copy into
    * @param source the batch to copy from
    * @param sourceStart the row number to start from in the source
-   * @return the number of rows copied
    */
   void copyIntoBatch(VectorizedRowBatch destination,
                      VectorizedRowBatch source,
Index: ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java	(date 1750794090926)
@@ -83,7 +83,9 @@
     return new RecordReaderImpl(this, options, conf);
   }
 
-
+  boolean isUTC() {
+    return useUTCTimestamp;
+  }
 
   @Override
   public RecordReader rows(boolean[] include) throws IOException {
````

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > the way you pass the UTC flag is hacky and not scalable for future use. updated the diff with all the changes, please take a look

The PR uses `TimestampColumnVector` existing `isUTC` flag. It does not introduce any new flags. The bug is that Hive currently ignores that flag (incorrectly) and always assumes that it is `true` even when it is `false`. The proper long term solution is to use the existing flag and initialize it correctly, IMO.  I think that introducing a parallel flag is not the right long term solution.

Please also see https://github.com/apache/orc/pull/2300

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: I'm not questioning that Hive currently ignores the UTC flag and always treats it as true. My concern is about how the flag is being set.
`TreeReaderFactory` is using `ReaderContext` to propagate the flag into the `TimestampColumnVector` vector.  
```` 
result.setIsUTC(this.context.getUseUTCTimestamp());
````
Rather than passing the context to the vector, we loop through all column vectors, searching for `TimestampColumnVector` and setting the flag manually in the RecordReaderImpl constructor.
While I agree that apache/orc#2300 is a proper fix, can we do the refactor in the meantime.  
Additionally, it would be nice to reuse the timestamp conversion function to avoid code duplication

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: I still don't see why significant refactoring and propagating flag using `context` is a better solution. What is wrong with the iterating over columns? The PR approach works with existing way how `TimestampTreeReader` sets `isUTC` flag in `TimestampColumnVector` and also aligns with the https://github.com/apache/orc/pull/2300 changes. Once Hive upgrades to ORC with the fix, the workaround in the PR can be kept in place or removed, while suggested refactoring becomes obsolete.

> Additionally, it would be nice to reuse the timestamp conversion function to avoid code duplication

I can make that change, if that the only remaining concern with the PR.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > The PR approach works with existing way how TimestampTreeReader sets isUTC flag in TimestampColumnVector

i don't think so 
````
public void nextVector(...) {
  ....
  result.setIsUTC(this.context.getUseUTCTimestamp());
  ....
}
````
Here is an update: no new classes or method signature changes (except static to private). 
Note: patch might seem big, but it just removes the dead code as well
````
Subject: [PATCH] refactor
---
Index: ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampUtils.java	(date 1750842094416)
@@ -18,22 +18,15 @@
 
 package org.apache.hadoop.hive.ql.exec.vector;
 
-import java.util.concurrent.TimeUnit;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 
 import org.apache.hadoop.hive.common.type.Timestamp;
-import org.apache.hadoop.hive.serde2.io.DateWritableV2;
 import org.apache.hadoop.hive.serde2.io.HiveIntervalDayTimeWritable;
 import org.apache.hadoop.hive.serde2.io.TimestampWritableV2;
 
 public final class TimestampUtils {
 
-  static final long MILLISECONDS_PER_SECOND = TimeUnit.SECONDS.toMillis(1);
-  static final long NANOSECONDS_PER_MILLISECOND = TimeUnit.MILLISECONDS.toNanos(1);
-
-  public static long daysToNanoseconds(long daysSinceEpoch) {
-    return DateWritableV2.daysToMillis((int) daysSinceEpoch) * NANOSECONDS_PER_MILLISECOND;
-  }
-
   public static TimestampWritableV2 timestampColumnVectorWritable(
       TimestampColumnVector timestampColVector, int elementNum,
       TimestampWritableV2 timestampWritable) {
@@ -63,4 +56,9 @@
     }
     return o.toString();
   }
+
+  public static Timestamp getValue(TimestampColumnVector tcv, int row) {
+    return Timestamp.ofEpochSecond(Math.floorDiv(tcv.time[row], 1000L), tcv.nanos[row],
+        tcv.isUTC() ? ZoneOffset.UTC : ZoneId.systemDefault());
+  }
 }
Index: ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java b/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java	(date 1750842094422)
@@ -44,6 +44,7 @@
 import org.apache.hadoop.hive.ql.exec.vector.MapColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.StructColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.TimestampColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.TimestampUtils;
 import org.apache.hadoop.hive.ql.exec.vector.UnionColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatchCtx;
@@ -518,7 +519,7 @@
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      result.set(TimestampUtils.getValue(tcv, row));
       return result;
     } else {
       return null;
Index: ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java	(date 1750842094411)
@@ -36,6 +36,7 @@
 import org.apache.hadoop.hive.ql.exec.vector.MapColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.StructColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.TimestampColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.TimestampUtils;
 import org.apache.hadoop.hive.ql.exec.vector.UnionColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.hadoop.hive.serde2.io.ByteWritable;
@@ -53,19 +54,20 @@
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Text;
 import org.apache.orc.TypeDescription;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+
+import static org.apache.orc.impl.TreeReaderFactory.ReaderContext;
 
 public class RecordReaderImpl extends org.apache.orc.impl.RecordReaderImpl
                               implements RecordReader {
-  static final Logger LOG = LoggerFactory.getLogger(RecordReaderImpl.class);
   private final VectorizedRowBatch batch;
+  private final ReaderContext context;
   private int rowInBatch;
   private long baseRow;
 
   protected RecordReaderImpl(ReaderImpl fileReader,
-    Reader.Options options, final Configuration conf) throws IOException {
+      Reader.Options options, final Configuration conf) throws IOException {
     super(fileReader, options);
+    context = (new ReaderContext()).useUTCTimestamp(fileReader.isUTC());
     final boolean useDecimal64ColumnVectors = conf != null && HiveConf.getVar(conf,
       HiveConf.ConfVars.HIVE_VECTORIZED_INPUT_FORMAT_SUPPORTS_ENABLED).equalsIgnoreCase("decimal_64");
     if (useDecimal64ColumnVectors){
@@ -444,7 +446,7 @@
     }
   }
 
-  static TimestampWritableV2 nextTimestamp(ColumnVector vector,
+  private TimestampWritableV2 nextTimestamp(ColumnVector vector,
                                            int row,
                                            Object previous) {
     if (vector.isRepeating) {
@@ -458,14 +460,15 @@
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      tcv.setIsUTC(context.getUseUTCTimestamp());
+      result.set(TimestampUtils.getValue(tcv, row));
       return result;
     } else {
       return null;
     }
   }
 
-  static OrcStruct nextStruct(ColumnVector vector,
+  private OrcStruct nextStruct(ColumnVector vector,
                               int row,
                               TypeDescription schema,
                               Object previous) {
@@ -493,7 +496,7 @@
     }
   }
 
-  static OrcUnion nextUnion(ColumnVector vector,
+  private OrcUnion nextUnion(ColumnVector vector,
                             int row,
                             TypeDescription schema,
                             Object previous) {
@@ -518,7 +521,7 @@
     }
   }
 
-  static ArrayList<Object> nextList(ColumnVector vector,
+  private ArrayList<Object> nextList(ColumnVector vector,
                                     int row,
                                     TypeDescription schema,
                                     Object previous) {
@@ -558,7 +561,7 @@
     }
   }
 
-  static Map<Object,Object> nextMap(ColumnVector vector,
+  private Map<Object,Object> nextMap(ColumnVector vector,
                                     int row,
                                     TypeDescription schema,
                                     Object previous) {
@@ -573,7 +576,7 @@
       TypeDescription valueType = schema.getChildren().get(1);
       LinkedHashMap<Object,Object> result;
       if (previous == null || previous.getClass() != LinkedHashMap.class) {
-        result = new LinkedHashMap<Object,Object>(length);
+        result = new LinkedHashMap<>(length);
       } else {
         result = (LinkedHashMap<Object,Object>) previous;
         // I couldn't think of a good way to reuse the keys and value objects
@@ -590,50 +593,31 @@
     }
   }
 
-  static Object nextValue(ColumnVector vector,
+  private Object nextValue(ColumnVector vector,
                           int row,
                           TypeDescription schema,
                           Object previous) {
-    switch (schema.getCategory()) {
-      case BOOLEAN:
-        return nextBoolean(vector, row, previous);
-      case BYTE:
-        return nextByte(vector, row, previous);
-      case SHORT:
-        return nextShort(vector, row, previous);
-      case INT:
-        return nextInt(vector, row, previous);
-      case LONG:
-        return nextLong(vector, row, previous);
-      case FLOAT:
-        return nextFloat(vector, row, previous);
-      case DOUBLE:
-        return nextDouble(vector, row, previous);
-      case STRING:
-        return nextString(vector, row, previous);
-      case CHAR:
-        return nextChar(vector, row, schema.getMaxLength(), previous);
-      case VARCHAR:
-        return nextVarchar(vector, row, schema.getMaxLength(), previous);
-      case BINARY:
-        return nextBinary(vector, row, previous);
-      case DECIMAL:
-        return nextDecimal(vector, row, previous);
-      case DATE:
-        return nextDate(vector, row, previous);
-      case TIMESTAMP:
-        return nextTimestamp(vector, row, previous);
-      case STRUCT:
-        return nextStruct(vector, row, schema, previous);
-      case UNION:
-        return nextUnion(vector, row, schema, previous);
-      case LIST:
-        return nextList(vector, row, schema, previous);
-      case MAP:
-        return nextMap(vector, row, schema, previous);
-      default:
-        throw new IllegalArgumentException("Unknown type " + schema);
-    }
+      return switch (schema.getCategory()) {
+          case BOOLEAN -> nextBoolean(vector, row, previous);
+          case BYTE -> nextByte(vector, row, previous);
+          case SHORT -> nextShort(vector, row, previous);
+          case INT -> nextInt(vector, row, previous);
+          case LONG -> nextLong(vector, row, previous);
+          case FLOAT -> nextFloat(vector, row, previous);
+          case DOUBLE -> nextDouble(vector, row, previous);
+          case STRING -> nextString(vector, row, previous);
+          case CHAR -> nextChar(vector, row, schema.getMaxLength(), previous);
+          case VARCHAR -> nextVarchar(vector, row, schema.getMaxLength(), previous);
+          case BINARY -> nextBinary(vector, row, previous);
+          case DECIMAL -> nextDecimal(vector, row, previous);
+          case DATE -> nextDate(vector, row, previous);
+          case TIMESTAMP -> nextTimestamp(vector, row, previous);
+          case STRUCT -> nextStruct(vector, row, schema, previous);
+          case UNION -> nextUnion(vector, row, schema, previous);
+          case LIST -> nextList(vector, row, schema, previous);
+          case MAP -> nextMap(vector, row, schema, previous);
+          default -> throw new IllegalArgumentException("Unknown type " + schema);
+      };
   }
 
   /* Routines for copying between VectorizedRowBatches */
@@ -789,19 +773,11 @@
     castedDestination.noNulls = castedSource.noNulls;
     if (source.isRepeating) {
       castedDestination.isNull[0] = castedSource.isNull[0];
-      for(int c=0; c > castedSource.fields.length; ++c) {
-        copyColumn(castedDestination.fields[c], castedSource.fields[c], 0, 1);
-      }
     } else {
       if (!castedSource.noNulls) {
         for (int r = 0; r < length; ++r) {
           castedDestination.isNull[r] = castedSource.isNull[sourceOffset + r];
         }
-      } else {
-        for (int c = 0; c > castedSource.fields.length; ++c) {
-          copyColumn(castedDestination.fields[c], castedSource.fields[c],
-              sourceOffset, length);
-        }
       }
     }
   }
@@ -829,14 +805,10 @@
           castedDestination.tags[r] = castedSource.tags[sourceOffset + r];
         }
       } else {
-        for(int r=0; r < length; ++r) {
+        for (int r=0; r < length; ++r) {
           castedDestination.tags[r] = castedSource.tags[sourceOffset + r];
         }
       }
-      for(int c=0; c > castedSource.fields.length; ++c) {
-        copyColumn(castedDestination.fields[c], castedSource.fields[c],
-            sourceOffset, length);
-      }
     }
   }
 
@@ -954,7 +926,6 @@
    * @param destination the batch to copy into
    * @param source the batch to copy from
    * @param sourceStart the row number to start from in the source
-   * @return the number of rows copied
    */
   void copyIntoBatch(VectorizedRowBatch destination,
                      VectorizedRowBatch source,
Index: ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java	(revision fac17feed39473c4219f3b3a2c63aeed4fc92701)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/orc/ReaderImpl.java	(date 1750794090926)
@@ -83,7 +83,9 @@
     return new RecordReaderImpl(this, options, conf);
   }
 
-
+  boolean isUTC() {
+    return useUTCTimestamp;
+  }
 
   @Override
   public RecordReader rows(boolean[] include) throws IOException {
````

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > i don't think so
```
public void nextVector(...) {
  ....
  result.setIsUTC(this.context.getUseUTCTimestamp());
  ....
}
```
I don't follow.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: I've attached here the `TreeReaderFactory` snippet that shows how UTC flag is passed to the column vector.  So your statement is not valid

> The PR approach works with existing way how TimestampTreeReader sets isUTC flag in TimestampColumnVector

Please apply the suggested changes to get the fix in until the orc fix is released. Note: now you'll have to cherry-pick the fix into the 4.1 branch

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: In the absence of tests, this PR would have been declined by most of committers.

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > In the absence of tests, this PR would have been declined by most of committers.

Which tests? There are plenty of tests that exercise ORC files and `TimestampWritableV2`. With the initial revision of the PR several tests failed (please see Jenkins build history).

If there are specific tests that needs to be added, please explain. 

Note that it will also block Spark upgrade to Hive 4.x till fix is implemented.

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > Please apply the suggested changes to get the fix in until the orc fix is released. Note: now you'll have to cherry-pick the fix into the 4.1 branch

I disagree with that solution and don't think that it is correct. You can open a separate PR and validate that it works both for Hive and Spark.

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > I've attached here the TreeReaderFactory snippet that shows how UTC flag is passed to the column vector. So your statement is not valid

There are no changes to `TimestampTreeReader` in the PR and how `isUTC` flag is set there, so the statement "The PR approach works with existing way how `TimestampTreeReader` sets isUTC flag in `TimestampColumnVector`" is valid. It is necessary to set `isUTC` flag in `TimestampColumnVector` on initialization or write, not on read and especially not to use a different flag that may be different from what `TimestampColumnVector` uses.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: 1. You've introduced functionality to support non-UTC timezones - where are the corresponding tests?
2. If you disagree, that's entirely up to you; we’ll proceed with the release without it.
3. `TimestampTreeReader` sets the flag directly on the `TimestampColumnVector`; it doesn't iterate over all column vectors as you implemented. Instead, it uses the `ReaderContext` to retrieve and propagate the UTC flag. I’d prefer not to come back to this again.

If you’re determined to proceed with your current approach, I’m afraid I won’t be able to assist further.

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: 1. As I mentioned local time zone was tested using Spark unit tests. I don't see how this can be done inside Hive (that hardcodes UTC time zone), but if you have a suggestion I am open to it.
2. I am not open to implement changes that I consider to be incorrect and not maintainable in the long run. It is your changes and if you consider them correct, why not to open a PR?
3. It is up to Hive PMC member to proceed with or without changes. I'll cast my (non-binding) vote once RC is available. Also, it is likely that Spark committers will also request fix to Spark regressions caused by the Hive behavior change between 2.3.10 and 4.x if the fix is not implemented one way or another in Hive 4.1.x
4. I never said that `TimestampTreeReader` uses the same approach as what is implemented in `RecordReaderImpl`. I said that it works with (and actually without) those changes. That change is required for TimestampFromXXXTreeReaders, not for `TimestampTreeReader`. And once https://github.com/apache/orc/pull/2300 is fixed, `TimestampFromXXXTreeReaders` will work with and without changes in `RecordReaderImpl`.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: PS: I have doubts that it works. We have `UTC` hardcoded in `ReaderOptions`:
https://github.com/apache/hive/blob/765ce86bb2066fbc7632db8ccf518cfa78c41ace/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcFile.java#L71
````
this.useUTCTimestamp = options.getUseUTCTimestamp();
fileReader.isUTC() would be always true
````

[HIVE-20007](https://issues.apache.org/jira/browse/HIVE-20007): Hive should carry out timestamp computations in UTC

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > You've introduced functionality to support non-UTC timezones 

To be correct, I have not introduced that functionality. Hive 2.x uses local time zone. ORC readers and `TimestampColumnVector` already support both UTC and local time zone as well. The only functionality that this PR targets is supporting proper conversion from `TimestampColumnVector` to `TimestampWritableV2` both for local and UTC time zones.

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > PS: I have doubts that it works. We have UTC hardcoded in ReaderOptions:

Please see "As I mentioned local time zone was tested using Spark unit tests. I don't see how this can be done inside Hive (that hardcodes UTC time zone), but if you have a suggestion I am open to it."

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: 1. You've opened a PR for Hive, not Spark.
2. Likewise. Please note that it's generally a bad practice to open a PR for an issue that's already being worked on by someone else.
3. The problem is also in ORC, so it's not solely a Hive issue. Without the hack in `RecordReaderImpl` I'd be fine with merging, but it wouldn't actually address your issue. 
Additionally, this isn't a regression, as Hive-3 behaved the same way. 
5. The approach using `ReaderContext` would also be compatible with the fix proposed in [ORC#2300](https://github.com/apache/orc/pull/2300).

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: Created a [draft](https://github.com/apache/hive/pull/5900), I'll try to add tests later. In my opinion, that is much cleaner solution.
Note, I am not planning on merging it.

Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: 1. It passes all Hive tests (that use UTC). As I said I am open to introduce new test cases assuming that other changes in the PR are OK and the PR is rejected only because of missing unit tests for `isUTC=false`.
2. It is bad ASF policy to reject PR without giving -1 and detailed technical explanation. It is OK to open a second PR that uses completely different approach.
3. The issue that I need to fix is purely in Hive. The problem is that without more changes in `RecordReaderImpl` few Hive tests also break, so that extra fix is necessary for Hive.
4. OK, it is regression in Hive 3.x compared to Hive 2.x. Note that Spark upgrades from 2.3.10 to 4.x, so for Spark it is a regression.
5. I don't argue whether or not your changes are compatible with https://github.com/apache/orc/pull/2300. They are simply orthogonal to the way how `TimestampColumnVector` works.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java:525]: > The only functionality that this PR targets is supporting proper conversion from TimestampColumnVector to TimestampWritableV2 both for local and UTC time zones.

Ok, then why do we need [this](https://github.com/apache/hive/pull/5868/files#diff-68582a8fd3dc2d4df104561d5ddc00f46001ca8aa0dbde493a621a12de37916fR79-R84)?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > Please see "As I mentioned local time zone was tested using Spark unit tests. I don't see how this can be done inside Hive (that hardcodes UTC time zone), but if you have a suggestion I am open to it."

Why do you even use Hive's RecordReader?

**Review by deniskuzZ — CHANGES_REQUESTED** [review]
Your solution has a limitation - it doesn't handle structs, such as a `StructColumnVector` that includes a `TimestampColumnVector` among its fields.

Similarly, `ListColumnVector` and `MapColumnVector` are not handled by the current approach either.

-1

**Comment by vrozov** [comment]
> Your solution has a limitation - it doesn't handle structs, such as a StructColumnVector that includes a TimestampColumnVector among its fields.

That can be easily fixed. Note that your solution assumes that ORC readers use exclusively `fileReader.readerOptions()` to set `isUTC` flag in `TimestampColumnVector` and this is not always the case (as you can see from https://github.com/apache/orc/pull/2300). Another issue with your PR is that all that propagation of `fileReader.readerOptions()` to `RecordReaderImpl` is not necessary (obsolete) once https://github.com/apache/orc/pull/2300 is fixed.

**Comment by deniskuzZ** [comment]
> That can be easily fixed. 

A few more loops and type checks? 

> Note that your solution assumes that ORC readers use exclusively `fileReader.readerOptions()` to set `isUTC` flag in `TimestampColumnVector`

Just to clarify, your solution does not make use of the `useUTCTimestamp` flag configured through readerOptions()? 

> Another issue with your PR is that all that propagation of `fileReader.readerOptions()` to `RecordReaderImpl` is not necessary (obsolete) once [apache/orc#2300](https://github.com/apache/orc/pull/2300) is fixed.

Are you saying your solution doesn't encounter this issue?

LOL

File: ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java (hunk 1 above)
Comment (vrozov) [reply ql/src/java/org/apache/hadoop/hive/ql/io/BatchToRowReader.java]: > > Please see "As I mentioned local time zone was tested using Spark unit tests. I don't see how this can be done inside Hive (that hardcodes UTC time zone), but if you have a suggestion I am open to it."
> 
> Why do you even use Hive's RecordReader?

I guess you mean why does Spark use Hive `RecordReader`?

**Comment by vrozov** [comment]
> A few more loops and type checks?

What is wrong with type check or a loop? Maybe you can explain your concern with that.

**Comment by vrozov** [comment]
> Just to clarify, your solution does not make use of the useUTCTimestamp flag configured through readerOptions()?

My solution uses `TimestampColumnVector.isUTC` flag. It does not rely on the way how the flag was set. It is supposed to be set by `TimestampTreeReader` or `TimestampFromXXXTreeReader` when `TimestampColumnVector` sets time of `TimestampColumnVector` and it could be local or UTC time, not while getting time of `TimestampColumnVector`. If you know time zone of the time in `TimestampColumnVector` without using `TimestampColumnVector.isUTC` flag, the flag should not be introduced in the first place.

**Comment by vrozov** [comment]
> Are you saying your solution doesn't encounter this issue?

The obsolete code in my solution is limited to 4 lines of code in a single location, don't you see the difference with the way how your solution affects multiple places and multiple APIs?

File: storage-api/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampColumnVector.java [hunk 3]
```diff
@@ -40,6 +40,15 @@
  * using the scratch timestamp, and then perhaps update the column vector row with a result.
  */
 public class TimestampColumnVector extends ColumnVector {
+  private static final ThreadLocal<Boolean> useUTC = ThreadLocal.withInitial(() -> Boolean.FALSE);
```
deniskuzZ, line 43: > The obsolete code in my solution is limited to 4 lines of code in a single location, don't you see the difference with the way how your solution affects multiple places and multiple APIs?

Just how many more hacks are you planning to add? You're contradicting your own claims.

deniskuzZ:
> My solution uses `TimestampColumnVector.isUTC` flag. It does not rely on the way how the flag was set.

Oh really, is that why you've added a ThreadLocal?

Doesn't  [here](https://github.com/apache/orc/pull/2300/files#diff-7dd7108785d69b8ff6238d9b8e24b2a50965839be8e176d0970f6b2455cdb105R1536) set the flag while reading the next vector?

deniskuzZ:
@vrozov, here's another suggestion: 
Either use the `maven-dependency-plugin` to replace Hive's classes with your fixes in Spark. That should help unblock the upgrade. Or wait for ORC release https://github.com/apache/orc/pull/2300

storage-api/src/java/org/apache/hadoop/hive/ql/exec/vector/TimestampColumnVector.java
vrozov: LOL. I did not know that  usage of `java.lang.ThreadLocal` class is now considered to be a hack. What other `java.lang` classes are considered to be a hack as well (just curious)? Have you checked Hive code for the  existing usage of the `ThreadLocal` class? Are those also hacks?

deniskuzZ: Know its limitations: this approach fails if vector processing is executed on a different thread. Since I’ve already shared my thoughts on your change, I think we can leave it here.

vrozov: > Know its limitations: this approach fails if vector processing is executed on a different thread.

Exactly where that different thread is introduced? Is it in Hive or ORC?

vrozov:
> Either use the maven-dependency-plugin to replace Hive's classes with your fixes in Spark. 

I don't see how that will work, but I will delegate to Spark community to decide on how to proceed once Hive 4.1.0 is released and I am able to publish my PR.

vrozov:
> Doesn't [here](https://github.com/apache/orc/pull/2300/files#diff-7dd7108785d69b8ff6238d9b8e24b2a50965839be8e176d0970f6b2455cdb105R1536) set the flag while reading the next vector?

Please take a closer look. In that code `nextVector()` reads `longColVector` and updates (writes) `timestampColVector` using `useUtc` flag and `longColVector`.

PR #5869, opened 2025-06-14
HIVE-29017: Restore ability to prevent metastore metrics startup queries

cnauroth:

Restore configuration `metastore.initial.metadata.count.enabled` to prevent metastore metrics startup queries

For releases as late as Hive 2.3.10, there was support for a `metastore.initial.metadata.count.enabled` configuration property. This could be set to `false` to prevent running queries to determine the counts of existing databases, tables and partitions. In large-scale ephemeral/serverless deployments with many HiveMetaStore processes starting up, this can be a source of unnecessary load spikes on the database. It appears this property was accidentally removed during the Standalone Metastore separation work, specifically in the [HIVE-17307](https://issues.apache.org/jira/browse/HIVE-17307) metrics refactoring.

Yes, this restores the ability to control configuration for whether or not to run these metrics queries.

Manual testing.

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -942,6 +942,9 @@ public class MetastoreConf {
     INIT_HOOKS("metastore.init.hooks", "hive.metastore.init.hooks", "",
         "A comma separated list of hooks to be invoked at the beginning of HMSHandler initialization. \n" +
             "An init hook is specified as the name of Java class which extends org.apache.riven.MetaStoreInitListener."),
+    INIT_METADATA_COUNT_ENABLED("metastore.initial.metadata.count.enabled",
+        "hive.metastore.initial.metadata.count.enabled", true,
+        "Enable a metadata count at metastore startup for metrics."),
     INTEGER_JDO_PUSHDOWN("metastore.integral.jdo.pushdown",
         "hive.metastore.integral.jdo.pushdown", false,
         "Allow JDO query pushdown for integral partition columns in metastore. Off by default. This\n" +
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -10055,7 +10055,8 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
 
   @VisibleForTesting
   void updateMetrics() throws MetaException {
-    if (Metrics.getRegistry() != null) {
+    if (Metrics.getRegistry() != null &&
+        MetastoreConf.getBoolVar(conf, ConfVars.INIT_METADATA_COUNT_ENABLED)) {
       LOG.info("Begin calculating metadata count metrics.");
       Metrics.getOrCreateGauge(MetricsConstants.TOTAL_TABLES).set(getMS().getTableCount());
       Metrics.getOrCreateGauge(MetricsConstants.TOTAL_PARTITIONS).set(getMS().getPartitionCount());


cnauroth:
@dengzhhu653 and @nrg4878 , this is another request to review bug fix changes in a similar area around the metastore metrics. Thank you.

dengzhhu653:
I'm not sure if we need this property since HIVE-27692, we don't run these count queries on embedded HMS client.

cnauroth:
> I'm not sure if we need this property since [HIVE-27692](https://issues.apache.org/jira/browse/HIVE-27692), we don't run these count queries on embedded HMS client.

Thanks for the reply, @dengzhhu653 . I still have a use for this configuration, even after [HIVE-27692](https://issues.apache.org/jira/browse/HIVE-27692). I have a customer whose architecture involves starting many ephemeral clusters (potentially thousands concurrently), running a few jobs, and then deleting the clusters. Each cluster runs the HiveMetaStore process connected to a centralized database. This is not embedded HMS, but it triggers the same kind of unnecessary load pattern. I would like to be able to disable these queries when we upgrade to Hive 4.

dengzhhu653:
> > I'm not sure if we need this property since [HIVE-27692](https://issues.apache.org/jira/browse/HIVE-27692), we don't run these count queries on embedded HMS client.
> 
> Thanks for the reply, @dengzhhu653 . I still have a use for this configuration, even after [HIVE-27692](https://issues.apache.org/jira/browse/HIVE-27692). I have a customer whose architecture involves starting many ephemeral clusters (potentially thousands concurrently), running a few jobs, and then deleting the clusters. Each cluster runs the HiveMetaStore process connected to a centralized database. This is not embedded HMS, but it triggers the same kind of unnecessary load pattern. I would like to be able to disable these queries when we upgrade to Hive 4.

What do you think? cc @deniskuzZ @nrg4878 @saihemanth-cloudera

cnauroth:
@ayushtkn , thank you for reviewing.

@dengzhhu653 , what are your thoughts on committing this, based on the +1 from Ayush?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -942,6 +942,9 @@ public enum ConfVars {
     INIT_HOOKS("metastore.init.hooks", "hive.metastore.init.hooks", "",
         "A comma separated list of hooks to be invoked at the beginning of HMSHandler initialization. \n" +
             "An init hook is specified as the name of Java class which extends org.apache.riven.MetaStoreInitListener."),
+    INIT_METADATA_COUNT_ENABLED("metastore.initial.metadata.count.enabled",
+        "hive.metastore.initial.metadata.count.enabled", true,
+        "Enable a metadata count at metastore startup for metrics."),
```
saihemanth-cloudera, line 947: Shouldn't this be disabled by default in Hive 4.x+?

ayushtkn: Sai, I checked the 3.x code, it was true then as well
https://github.com/apache/hive/blob/bdf3285bc250609c1b31a52e7732be8c043b8817/standalone-metastore/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java#L539-L541

& if you check the current code
```
 if (Metrics.getRegistry() != null &&
        MetastoreConf.getBoolVar(conf, ConfVars.INIT_METADATA_COUNT_ENABLED)) {
```
earlier if ``Metrics.getRegistry() `` is not ``null`` this was always enabled, but if set `false` as `default`, it won't preserve the old behaviour.

So, I think `true` as default is fine, we are just allowing an option to disable this if they want, as of now it is always enabled.

Did I miss something here?

saihemanth-cloudera: My bad! I thought this was disabled by default.

cnauroth:
Thank you for the review and commit everyone!

PR #5870, opened 2025-06-14
HIVE-28998: Support Iceberg REST Catalog without authentication

okumin:
This PR would add a new authentication mode, NONE, to the Iceberg REST Catalog API. The currently supported SIMPLE and JWT are not central authentication. It is inherited from another Hive servlet. It appears that NONE and OAuth2 are widely used.

To acquire test users.

No. The feature has not been shipped.

Unit and integration tests

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -1868,8 +1868,8 @@ public class MetastoreConf {
             " positive value will be used as-is."
     ),
     ICEBERG_CATALOG_SERVLET_AUTH("metastore.iceberg.catalog.servlet.auth",
-        "hive.metastore.iceberg.catalog.servlet.auth", "jwt", new StringSetValidator("simple", "jwt"),
-        "HMS Iceberg Catalog servlet authentication method (simple or jwt)."
+        "hive.metastore.iceberg.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt"),
+        "HMS Iceberg Catalog servlet authentication method (none, simple, or jwt)."
     ),
     ICEBERG_CATALOG_CACHE_EXPIRY("metastore.iceberg.catalog.cache.expiry",
         "hive.metastore.iceberg.catalog.cache.expiry", -1,
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
@@ -24,6 +24,7 @@ import java.util.concurrent.SynchronousQueue;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.common.ZKDeRegisterWatcher;
 import org.apache.hadoop.hive.common.ZooKeeperHiveHelper;
+import org.apache.hadoop.hive.metastore.ServletSecurity.AuthType;
 import org.apache.hadoop.hive.metastore.api.Database;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore;
@@ -420,7 +421,8 @@ public class HiveMetaStore extends ThriftHiveMetastore {
     processor = new ThriftHiveMetastore.Processor<>(handler);
     LOG.info("Starting DB backed MetaStore Server with generic processor");
     boolean jwt = MetastoreConf.getVar(conf, ConfVars.THRIFT_METASTORE_AUTHENTICATION).equalsIgnoreCase("jwt");
-    ServletSecurity security = new ServletSecurity(conf, jwt);
+    AuthType authType = jwt ? AuthType.JWT : AuthType.SIMPLE;
+    ServletSecurity security = new ServletSecurity(authType, conf);
     Servlet thriftHttpServlet = security.proxy(new TServlet(processor, protocolFactory));
 
     boolean directSqlEnabled = MetastoreConf.getBoolVar(conf, ConfVars.TRY_DIRECT_SQL);
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java
@@ -78,25 +78,25 @@ import java.util.Optional;
  * </p>
  */
 public class ServletSecurity {
+  public enum AuthType {
+    NONE, SIMPLE, JWT;
+
+    public static AuthType fromString(String type) {
+      return AuthType.valueOf(type.toUpperCase());
+    }
+  }
+
   private static final Logger LOG = LoggerFactory.getLogger(ServletSecurity.class);
   static final String X_USER = MetaStoreUtils.USER_NAME_HTTP_HEADER;
   private final boolean isSecurityEnabled;
-  private final boolean jwtAuthEnabled;
+  private final AuthType authType;
   private final Configuration conf;
   private JWTValidator jwtValidator = null;
 
-  public ServletSecurity(String authType, Configuration conf) {
-    this(conf, isAuthJwt(authType));
-  }
-
-  public ServletSecurity(Configuration conf, boolean jwt) {
+  public ServletSecurity(AuthType authType, Configuration conf) {
     this.conf = conf;
     this.isSecurityEnabled = UserGroupInformation.isSecurityEnabled();
-    this.jwtAuthEnabled = jwt;
-  }
-
-  private static boolean isAuthJwt(String authType) {
-    return "jwt".equalsIgnoreCase(authType);
+    this.authType = authType;
   }
 
   /**
@@ -104,7 +104,7 @@ public class ServletSecurity {
    * @throws ServletException if the jwt validator creation throws an exception
    */
   public void init() throws ServletException {
-    if (jwtAuthEnabled && jwtValidator == null) {
+    if (authType == AuthType.JWT && jwtValidator == null) {
       try {
         jwtValidator = new JWTValidator(this.conf);
       } catch (Exception e) {
@@ -155,10 +155,14 @@ public class ServletSecurity {
    * @return a servlet instance or null if security initialization fails
    */
   public HttpServlet proxy(HttpServlet servlet) {
+    if (authType == AuthType.NONE) {
+      LOG.warn("Servlet security is disabled for {}", servlet);
+      return servlet;
+    }
     try {
       init();
     } catch (ServletException e) {
-      LOG.error("Unable to proxy security for servlet {}", servlet.toString(), e);
+      LOG.error("Unable to proxy security for servlet {}", servlet, e);
       return null;
     }
     return new ProxyServlet(servlet);
@@ -206,10 +210,12 @@ public class ServletSecurity {
       String userFromHeader = extractUserName(request, response);
       // Temporary, and useless for now. Here only to allow this to work on an otherwise kerberized
       // server.
-      if (isSecurityEnabled || jwtAuthEnabled) {
+      if (isSecurityEnabled || authType == AuthType.JWT) {
         LOG.info("Creating proxy user for: {}", userFromHeader);
         clientUgi = UserGroupInformation.createProxyUser(userFromHeader, UserGroupInformation.getLoginUser());
       } else {
+        // Unreachable in the case of NONE
+        Preconditions.checkState(authType == AuthType.SIMPLE);
         LOG.info("Creating remote user for: {}", userFromHeader);
         clientUgi = UserGroupInformation.createRemoteUser(userFromHeader);
       }
@@ -237,13 +243,15 @@ public class ServletSecurity {
 
   private String extractUserName(HttpServletRequest request, HttpServletResponse response)
       throws HttpAuthenticationException {
-    if (!jwtAuthEnabled) {
+    if (authType == AuthType.SIMPLE) {
       String userFromHeader = request.getHeader(X_USER);
       if (userFromHeader == null || userFromHeader.isEmpty()) {
         throw new HttpAuthenticationException("User header " + X_USER + " missing in request");
       }
       return userFromHeader;
     }
+    // Unreachable in the case of NONE
+    Preconditions.checkState(authType == AuthType.JWT);
     String signedJwt = extractBearerToken(request, response);
     if (signedJwt == null) {
       throw new HttpAuthenticationException("Couldn't find bearer token in the auth header in the request");
--- a/standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
+++ b/standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
@@ -23,6 +23,7 @@ import java.util.TreeMap;
 import javax.servlet.http.HttpServlet;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.metastore.ServletSecurity;
+import org.apache.hadoop.hive.metastore.ServletSecurity.AuthType;
 import org.apache.hadoop.hive.metastore.ServletServerBuilder;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf.ConfVars;
@@ -99,7 +100,7 @@ public class HMSCatalogFactory {
    */
   private HttpServlet createServlet(Catalog catalog) {
     String authType = MetastoreConf.getVar(configuration, ConfVars.ICEBERG_CATALOG_SERVLET_AUTH);
-    ServletSecurity security = new ServletSecurity(authType, configuration);
+    ServletSecurity security = new ServletSecurity(AuthType.fromString(authType), configuration);
     return security.proxy(new HMSCatalogServlet(new HMSCatalogAdapter(catalog)));
   }
 
--- a/standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestRESTCatalogJwtAuth.java
+++ b/standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestRESTCatalogJwtAuth.java
@@ -34,7 +34,7 @@ import org.junit.jupiter.api.extension.RegisterExtension;
 class TestRESTCatalogJwtAuth extends BaseRESTCatalogTests {
   @RegisterExtension
   private static final HiveRESTCatalogServerExtension REST_CATALOG_EXTENSION =
-      new HiveRESTCatalogServerExtension(AuthType.JWT);
+      HiveRESTCatalogServerExtension.builder(AuthType.JWT).build();
 
   @Override
   protected Map<String, String> getDefaultClientConfiguration() throws Exception {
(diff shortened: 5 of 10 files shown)


standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestHMSCatalogAuthJWT.java
```diff
@@ -0,0 +1,287 @@
...
+    MOCK_JWKS_SERVER.stubFor(get("/jwks")
+        .willReturn(ok()
+            .withBody(Files.readAllBytes(JWT_JWKS_FILE.toPath()))));
+    super.setUp();
+  }
+
+  @After
+  @Override
+  public void tearDown() throws Exception {
+    super.tearDown();
```
deniskuzZ: why override?

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestHMSCatalogAuthJWT.java
```diff
@@ -0,0 +1,287 @@
...
+  @Override
+  public void setUp() throws Exception {
+    conf = MetastoreConf.newMetastoreConf();
+    MetastoreConf.setVar(conf, MetastoreConf.ConfVars.ICEBERG_CATALOG_SERVLET_AUTH, "jwt");
+    MetastoreConf.setVar(conf, MetastoreConf.ConfVars.THRIFT_METASTORE_AUTHENTICATION_JWT_JWKS_URL,
+        "http://localhost:" + MOCK_JWKS_SERVER_PORT + "/jwks");
+    MOCK_JWKS_SERVER.stubFor(get("/jwks")
+        .willReturn(ok()
+            .withBody(Files.readAllBytes(JWT_JWKS_FILE.toPath()))));
+    super.setUp();
```
deniskuzZ: can we pass `conf` in `setUp()` for more clarify

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestHMSCatalogAuthNone.java
```diff
@@ -36,45 +39,41 @@
...
+    MetastoreConf.setVar(conf, MetastoreConf.ConfVars.ICEBERG_CATALOG_SERVLET_AUTH, "none");
+    super.setUp();
   }
-  
+
   @After
   @Override
   public void tearDown() throws Exception {
-      super.tearDown();
+    super.tearDown();
```
deniskuzZ: same, no need to override

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestHMSCatalogAuthNone.java
```diff
@@ -36,45 +39,41 @@
...
 
   @Test
   public void testCreateNamespaceHttp() throws Exception {
     String ns = "nstesthttp";
     // list namespaces
     URL url = new URL("http://hive@localhost:" + catalogPort + "/"+catalogPath+"/v1/namespaces");
-    String jwt = generateJWT();
     // check namespaces list (ie 0)
-    Object response = clientCall(jwt, url, "GET", null);
+    Object response = clientCall(url, "GET", Collections.emptyMap(), null);
```
deniskuzZ: is that the only difference? 
if yes, can't we move the tests into base class and override just the client function call?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java
```diff
@@ -78,33 +78,37 @@
...
-  public ServletSecurity(String authType, Configuration conf) {
-    this(conf, isAuthJwt(authType));
-  }
-
-  public ServletSecurity(Configuration conf, boolean jwt) {
+  public ServletSecurity(AuthType authType, Configuration conf) {
     this.conf = conf;
     this.isSecurityEnabled = UserGroupInformation.isSecurityEnabled();
-    this.jwtAuthEnabled = jwt;
+    this.authType = authType;
```
deniskuzZ, line 99: why not do `this.authType = AuthType.fromString(authType)` here?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java
```diff
@@ -78,33 +78,37 @@
...
   }
 
-  private static boolean isAuthJwt(String authType) {
-    return "jwt".equalsIgnoreCase(authType);
+  public ServletSecurity(Configuration conf, boolean jwt) {
+    this(jwt ? AuthType.JWT : AuthType.SIMPLE, conf);
```
deniskuzZ: this looks confusing, can we get rid of this constructor?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java
```diff
@@ -156,6 +160,10 @@ public String getServletInfo() {
    * @return a servlet instance or null if security initialization fails
    */
   public HttpServlet proxy(HttpServlet servlet) {
+    if (authType == AuthType.NONE) {
+      LOG.info("No security enabled, not proxying servlet {}", servlet.toString());
```
deniskuzZ: minor: `LOG.warn("Servlet security is disabled {}, servlet.toString());`

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1867,7 +1867,7 @@ public enum ConfVars {
             " positive value will be used as-is."
     ),
     ICEBERG_CATALOG_SERVLET_AUTH("metastore.iceberg.catalog.servlet.auth",
-        "hive.metastore.iceberg.catalog.servlet.auth", "jwt", new StringSetValidator("simple", "jwt"),
+        "hive.metastore.iceberg.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt"),
```
deniskuzZ, line 1871: can we update the [HMS config in docker](https://github.com/apache/hive/blob/2826a6e6dfaa5ff3c0f3e077b55b2994c98f60c4/standalone-metastore/metastore-server/src/docker/conf/metastore-site.xml#L41) to NONE as well

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestHMSCatalogAuthNone.java
okumin: Yes. I'm thinking of replacing all tests with the compatibility kit before applying cosmetic changes on this test cases. Basically, the current coverage is too small.
https://issues.apache.org/jira/browse/HIVE-29019

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: OK. I am considering if we should support SIMPLE and JWT. They are likely inherited from `hive.metastore.properties.servlet.auth`.

deniskuzZ: That comes from `PROPERTIES_SERVLET_AUTH` - another servlet that has 0 usages. I don't understand why we can't have 1 generic config

okumin: I think we should separate one at least for Iceberg. OAuth2 is not used in Hive's endpoints, but the Iceberg community prefers it. Another table format might prefer SPNEGO or something else if we support it. Anyway, a set of auth methods relies on the product, the community, and the client implementation.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
```diff
@@ -431,7 +432,8 @@ public void setThreadFactory(ThreadFactory threadFactory) {
     processor = new ThriftHiveMetastore.Processor<>(handler);
     LOG.info("Starting DB backed MetaStore Server with generic processor");
     boolean jwt = MetastoreConf.getVar(conf, ConfVars.THRIFT_METASTORE_AUTHENTICATION).equalsIgnoreCase("jwt");
-    ServletSecurity security = new ServletSecurity(conf, jwt);
+    AuthType authType = jwt ? AuthType.JWT : AuthType.SIMPLE;
```
deniskuzZ, line 424: why do we need these hacks? can we drop the jwt var completely?

I think we should replace `ICEBERG_CATALOG_SERVLET_AUTH` & `PROPERTIES_SERVLET_AUTH`
with
`METASTORE_REST_CATALOG_AUTH("metastore.rest-catalog.authentication"` since these servlets are deployed within the same servlet container.
````
String authType = MetastoreConf.getVar(configuration, ConfVars.METASTORE_REST_CATALOG_AUTH);
ServletSecurity security = new ServletSecurity(AuthType.fromString(authType), configuration);
`

okumin: This line configures [an HTTP server for HiveMetaStore with Thrift over HTTP](https://github.com/apache/hive/blob/174ff56b77b212bf51ee01587de9cf22e77f2dd3/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java#L966-L977), not for Iceberg REST or Properties REST. This change keeps the original behavior. We can't fully encode `hive.metastore.authentication` there, as it doesn't have SIMPLE.

As for the necessity of consolidation, I am currently biased toward separation. Let's say we introduced OAuth 2.0 with OIDC for the Iceberg REST catalog, and a user wanted to use it. Can all PROPERTIES clients immediately use OAuth? I am a little skeptical that different HTTP endpoints can share the same authentication method even if they are on the same machine.

deniskuzZ: ok, how about 
````
String authType = MetastoreConf.getVar(configuration, ConfVars.THRIFT_METASTORE_AUTHENTICATION);

ServletSecurity security = new ServletSecurity(
  AuthType.JWT.name().equalsIgnoreCase(authType) ? AuthType.JWT : AuthType.SIMPLE, 
  configuration);
````

deniskuzZ: i find it bizarre to support different authentication methods for multiple endpoints served by the same servlet in a production cluster

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: `hive.metastore.iceberg.catalog.servlet.auth` config rename has caused confusion. It's not meant to be specific to Iceberg - it's intended for use by the HMS RestCatalog (aka UnityCatalog) serving federated metadata (including Hive external and ACID tables)."

deniskuzZ (approved):
+1, thanks @okumin for this feature, let's get it in :)

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestHMSCatalogAuthNone.java
deniskuzZ: @okumin meanwhile could we please do the following refactor using parametrized test and RESTCatalog client
````
@Parameters(name = "authType = {0}")
public static List<Object> authType() {
  return Arrays.asList(AuthType.JWT, AuthType.NONE);
}

@Parameter(0)
public ServletSecurity.AuthType authType; 
protected RESTCatalog restCatalog;
````

in setUp (use RESTCatalog client instead of manually constructed http requests)
````
    MetastoreConf.setVar(conf, MetastoreConf.ConfVars.ICEBERG_CATALOG_SERVLET_AUTH, authType.name());
    
    restCatalog = initCatalog(catalog.name());
  }

  private RESTCatalog initCatalog(String catalogName) throws Exception {
    URI catalogUri = URI.create("http://localhost:" + catalogPort + "/" + catalogPath);
    RESTCatalog catalog = new RESTCatalog();

    Map<String, String> properties = Maps.newHashMap();
    properties.put(CatalogProperties.URI, catalogUri.toString());
    if (authType == ServletSecurity.AuthType.JWT) {
      properties.put("token", generateJWT());
    }
    catalog.initialize(catalogName, properties);
    return catalog;
  }
````
refactored Test snippet
````
@Test
  public void testTableAPI() throws Exception {
    Schema schema = getTestSchema();
    final String tblName = "tbl_" + Integer.toHexString(RND.nextInt(65536));
    final TableIdentifier TBL = TableIdentifier.of(DB_NAME, tblName);
    String location = temp.newFolder(TBL.toString()).toString();

    restCatalog
        .buildTable(TBL, schema)
        .withLocation(location)
        .create();

    Table table = catalog.loadTable(TBL);
    Assert.assertEquals(location, table.location());

    // rename table
    final String rtblName = "TBL_" + Integer.toHexString(RND.nextInt(65536));
    final TableIdentifier rtableIdent = TableIdentifier.of(DB_NAME, rtblName);

    restCatalog.renameTable(TBL, rtableIdent);
    table = catalog.loadTable(rtableIdent);
    Assert.assertEquals(
        String.join(".", catalog.name(),
        rtableIdent.toString()), table.name());
  }
````

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: I'd like to follow your take if you have such a context. I'd be more than happy if you had the summary of the vision :) > it's intended for use by the HMS RestCatalog (aka UnityCatalog) serving federated metadata (including Hive external and ACID tables)

Anyway, I would like the conclusion here. This is the summary I have.

| Service | AuthN supported by clients | users |
|-|-|-|
| HMS(HTTP) | NONE(probably, like a SIMPLE), JWT | Hive, etc. |
| Properties API | SIMPLE, JWT | (Who uses it?) |
| Iceberg REST | NONE, OAuth 2.0 | Icebrerg clients |
| UnityCatalog | OAuth 2.0? | Databricks, etc. |

What should we support in `metastore.rest-catalog.authentication`? Having the union is likely a conclusion. I'm concerned that some clients can't support an arbitrary protocol because some of them are out of our control.

deniskuzZ: hi @okumin, 
The idea was to expose the metadata management functionality via REST in HMS. 
Currently we are focused only on exposing iceberg tables, however, we could extend it with the support of Hive managed format like ACID. 
Another thing is federation. Register multiple catalog connectors in HMS (keep connection details with credentials in HMS DB) and expose it via same Unified/Iceberg Rest API.

UnityCatalog already supports a Managed catalog and ability to integrate with the Foreign catalogs like Glue, HMS, Iceberg Rest Catalogs.

| Service | AuthN supported by clients | users |
|-|-|-|
| HMS (HTTP over Thrift) | SIMPLE, JWT | Hive, etc. |
| HMS (REST) | NONE, SIMPLE, JWT, OAuth 2.0 | HMS RestCatalog clients (Iceberg, etc..)  |

The HMS RestCatalog would handle the proxying of metadata requests to either the HMSHandler, Iceberg's HiveCatalog or one of the registered external catalogs, using the stored connection details and credentials.

@okumin, does it make sense? 

As far as I know, the Properties API isn't used by anyone. That was one of my concerns, given the lack of a clear vision or usage patterns. I felt like it's yet another key-value store bicycle without a defined purpose.

zhangbutao: It was a very good discussion. This discussion involved the further development direction and positioning of HMS in the future. 

First, I think we should write a detailed proposal and discuss it clearly in the community email. Because I feel that many people are not clear about the positioning of the HMS rest catalog, which may cause various misunderstandings. 

Returning to the current issue, I basically agree with Denys' statement that HMS(Rest) in the future will be similar to the Unity Catalog, capable of serving as an Iceberg rest catalog server (its primary function), and also able to register other catalogs (such as Hive tables from other HMS).

okumin: @deniskuzZ @zhangbutao Thanks. I also think we should discuss the excellent vision on the mailing list. I want to support it. However, one point remains unclear to me.
Unity Catalog provides two types of endpoints: One is Unity REST API(`/api/2.1/unity-catalog`) and the other is Iceberg REST API(`/api/2.1/unity-catalog/iceberg-rest`). [Unity REST API is likely to authenticate a user with Personal Access Token](https://docs.databricks.com/aws/en/external-access/unity-rest). [Iceberg REST API is likely to authenticate a user with OAuth or PAT](https://docs.databricks.com/aws/en/external-access/iceberg). I understand we want the equivalent of the Unity REST; Let's call it HMS v2 REST.
I don't understand why we can use the consistent auth across the endpoints. HMS v2 REST clients will be able to use SIMPLE(i.e., a client sends a user name via `x-actor-username` header) if we implement the client so. However, as a reality, [most Iceberg REST clients probably do not support it](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/auth/AuthManagers.java). I think Trino has no method to use SIMPLE and JWT.
If I were saying something strange, I would follow the recommendation now.

deniskuzZ: I'll try to create a draft with the proposal, so we can collaborate more effectively. 
You might be right and we should consider the possibility of supporting multiple endpoints, each with its own authentication. I haven't thought much about it.
FYI, we are using JWT downstream to enable Snowflake integration with HMS Iceberg RestCatalog.

okumin: This is a good discussion, but I started thinking it does not need to be a blocker to make the current Iceberg REST API minimally viable. I separated the debate in the following ticket.
https://issues.apache.org/jira/browse/HIVE-29038

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/TestHMSCatalogAuthNone.java
okumin: It's okay if it's a blocker, but I'm removing the current test case in this pull request. I'll do it if we want to merge HIVE-28998 first, and the change is a must.
https://github.com/apache/hive/pull/5887

deniskuzZ: Let's merge https://github.com/apache/hive/pull/5887 first.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: Yes, 100%. That's beyond the scope of this PR

deniskuzZ:
hi @okumin, could you please rebase, or if you are busy, i can help. we need this in 4.1

okumin:
I probably think the consensus is that this is not a blocker, but I try to rebase this pr promptly.

okumin:
@deniskuzZ I rebased it

deniskuzZ (approved):
+1, pending tests

PR #5871, opened 2025-06-14
HIVE-29016: Disable caching on the Iceberg REST Catalog

okumin:
Remove caching on the Iceberg REST API server. It's mainly because the current code didn't work with caching correctly, as stated in HIVE-29016.

Also, I believe the REST Catalog service is not as mature as considering caching. Caching objects without proper consideration can be harmful, such as when users update an Iceberg table with both Thrift and REST APIs in parallel.

Remove caching.

No. This feature is not yet shipped.

Unit test


okumin:
@deniskuzZ If the +1 stands for the approval, I will merge this pull request

deniskuzZ:
@okumin, I'd like to loop in @henrib and @dengzhhu653 to reach a consensus

henrib:
> @okumin, I'd like to loop in @henrib and @dengzhhu653 to reach a consensus

Thanks @deniskuzZ :-) 
Before we remove the code, doesn't disabling caching by setting the property hive.metastore.catalog.cache.expiry=-1  a valid workaround for the initial 'Trino creates table' issue ? I'd much rather having caching work by improving tx handling  - and since we are in HMS, observe and invalidate cache entries in case of changes - than assume caching is not needed (loading these manifests is not without latency and IO rate can be quite demanding with many concurrent users).

May be worth trying overriding HMSCachingCatalog:buildTable():
`
  @Override
  public Catalog.TableBuilder buildTable(TableIdentifier identifier, Schema schema) {
    super.tableCache.invalidate(identifier);
    return nsCatalog.buildTable(identifier, schema);
  }
`

deniskuzZ:
good idea with setting the property `hive.metastore.catalog.cache.expiry`=-1. This we can merge quickly

okumin:
@henrib What client did you use while testing the feature?
I suppose it is not a Trino issue but the Hive's REST catalog issue, which means "hive.metastore.catalog.cache.expiry > 0" does not work with another system such as Spark. If my understanding is correct, I suppose we don't need to keep the current implementation. When caching is really necessary, we will implement a new mechanism with robust integration tests, and add the configuration back. I assume adding something is easy and removing or renaming exinsting ones are relatively hard

henrib:
@okumin , setting the property hive.metastore.catalog.cache.expiry=-1 in the configuration disables the cache which solves the issue of creating/updating tables through the catalog triggering an error. The actual coding solution to that error might just be overriding buildTable() as mentioned above. Removing the caching code will just make querying - which is expected to be the most common operation - slower and less efficient; caching is already necessary at scale.
If overriding buildTable() seems too complicated or not sufficient, changing the default value for the property is enough to solve the issue for users updating tables through catalog without impeding querying usage for the most common case.

okumin:
OK. Although I haven't tested the proposed solution, it's not a bad idea to disable it at this point. Is this expected?
https://github.com/apache/hive/pull/5878

PR #5872, opened 2025-06-14
HIVE-29012: NPE in ExplainTask when protologging posthook is enabled

Aggarwal-Raghav:
Check [HIVE-29012](https://issues.apache.org/jira/browse/HIVE-29012) for the stacktrace. 

When protologging hook is enabled, it stores the query plan for a query as well. The conf object in ExplainTask is null for the "Hive Hook Proto Log Writer" thread.

No

Will see CI output + local single node setup/cluster testing

### DIFF
--- a/ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
@@ -21,13 +21,22 @@
 package org.apache.hadoop.hive.ql.hooks;
 
 import static java.util.Collections.singletonList;
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertArrayEquals;
+import static org.mockito.Mockito.mock;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
-
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.ql.CompilationOpContext;
+import org.apache.hadoop.hive.ql.exec.FetchTask;
+import org.apache.hadoop.hive.ql.exec.TableScanOperator;
 import org.apache.hadoop.hive.ql.hooks.proto.HiveHookEvents;
+import org.apache.hadoop.hive.ql.parse.ExplainConfiguration;
+import org.apache.hadoop.hive.ql.plan.ExplainWork;
+import org.apache.hadoop.hive.ql.plan.FetchWork;
+import org.apache.hadoop.hive.ql.plan.TableDesc;
 import org.json.JSONObject;
 import org.junit.Test;
 
@@ -49,6 +58,15 @@ public class TestHiveHookEventProtoPartialBuilder {
     assertArrayEquals(event1.toByteArray(), event2.toByteArray());
   }
 
+  @Test
+  public void testOtherInfoQueryPlan() {
+    String stageIdRearrange = HiveConf.ConfVars.HIVE_STAGE_ID_REARRANGE.defaultStrVal;
+    HiveHookEvents.HiveHookEventProto event = buildIn2Steps(mock(JSONObject.class), getWork(), stageIdRearrange);
+    // Without HIVE-29012, OTHERINFO will have CONF initially i.e size is 1.
+    // With HIVE-29012, OTHERINFO should have QUERY as well i.e size is 2.
+    assert(event.getOtherInfo(1).hasKey());
+  }
+
   private HiveHookEvents.HiveHookEventProto buildWithOtherInfo(JSONObject json) {
     return HiveHookEvents.HiveHookEventProto
             .newBuilder()
@@ -66,6 +84,10 @@ public class TestHiveHookEventProtoPartialBuilder {
   }
 
   private HiveHookEvents.HiveHookEventProto buildIn2Steps(JSONObject json) {
+    return buildIn2Steps(json, null, null);
+  }
+
+  private HiveHookEvents.HiveHookEventProto buildIn2Steps(JSONObject json, ExplainWork work, String stageIdRearrange) {
     HiveHookEvents.HiveHookEventProto.Builder builder = HiveHookEvents.HiveHookEventProto
             .newBuilder()
             .setEventType(HiveProtoLoggingHook.EventType.QUERY_SUBMITTED.name())
@@ -77,6 +99,24 @@ public class TestHiveHookEventProtoPartialBuilder {
             .setExecutionMode(TEZ);
     Map<HiveProtoLoggingHook.OtherInfoType, JSONObject> otherInfo = new HashMap<>();
     otherInfo.put(HiveProtoLoggingHook.OtherInfoType.CONF, json);
-    return new HiveHookEventProtoPartialBuilder(builder, null, otherInfo, null, null).build();
+    return new HiveHookEventProtoPartialBuilder(builder, work, otherInfo, null, stageIdRearrange).build();
+  }
+
+  private ExplainWork getWork() {
+    CompilationOpContext cCtx = new CompilationOpContext();
+    TableScanOperator scanOp = new TableScanOperator(cCtx);
+
+    FetchWork taskWork = new FetchWork(mock(Path.class), mock(TableDesc.class));
+    taskWork.setSource(scanOp);
+
+    // Create a FetchTask with the FetchWork
+    FetchTask task = new FetchTask();
+    task.setWork(taskWork);
+
+    ExplainWork work = new ExplainWork();
+    work.setConfig(new ExplainConfiguration());
+    work.setRootTasks(List.of(task));
+
+    return work;
   }
 }
--- a/ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
@@ -22,11 +22,13 @@ package org.apache.hadoop.hive.ql.hooks;
 
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertArrayEquals;
-import static org.mockito.Mockito.mock;
+import static org.junit.Assert.assertEquals;
 
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.ql.CompilationOpContext;
@@ -54,17 +56,19 @@ public class TestHiveHookEventProtoPartialBuilder {
     json.put("key2", "value2");
     json.put("key3", "value3");
     HiveHookEvents.HiveHookEventProto event1 = buildWithOtherInfo(json);
-    HiveHookEvents.HiveHookEventProto event2 = buildIn2Steps(json);
+    HiveHookEvents.HiveHookEventProto event2 = buildIn2Steps(json, null);
     assertArrayEquals(event1.toByteArray(), event2.toByteArray());
   }
 
   @Test
   public void testOtherInfoQueryPlan() {
-    String stageIdRearrange = HiveConf.ConfVars.HIVE_STAGE_ID_REARRANGE.defaultStrVal;
-    HiveHookEvents.HiveHookEventProto event = buildIn2Steps(mock(JSONObject.class), getWork(), stageIdRearrange);
-    // Without HIVE-29012, OTHERINFO will have CONF initially i.e size is 1.
-    // With HIVE-29012, OTHERINFO should have QUERY as well i.e size is 2.
-    assert(event.getOtherInfo(1).hasKey());
+    HiveHookEvents.HiveHookEventProto event = buildIn2Steps(new JSONObject(), createExplainWork());
+    Set<String> expectedOutput = Set.of("CONF", "QUERY");
+    Set<String> actualOutput =
+        event.getOtherInfoList().stream()
+            .map(HiveHookEvents.MapFieldEntry::getKey)
+            .collect(Collectors.toSet());
+    assertEquals(expectedOutput, actualOutput);
   }
 
   private HiveHookEvents.HiveHookEventProto buildWithOtherInfo(JSONObject json) {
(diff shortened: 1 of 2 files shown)


Aggarwal-Raghav:
Based on the debugging, the _Task#initialize()_ method creates conf object in HiveServer2-Handler-Pool thread and when the post hook for protologging is triggered, it created a new thread (Hive Hook Proto Log Writer) which get the query plan and at that time, _ExplainTask#outputPlan()_ has conf as null which causes NPE.

https://github.com/apache/hive/blob/af59f241a07fb9960412529be71c6bfa1fcd042e/ql/src/java/org/apache/hadoop/hive/ql/exec/ExplainTask.java#L1006

Aggarwal-Raghav:
@zabetak, can you please help with the review as the changes are around HIVE-22173's code.

ql/src/java/org/apache/hadoop/hive/ql/exec/ExplainTask.java
```diff
@@ -1003,6 +1003,9 @@ public JSONObject outputPlan(Object work, PrintStream out,
       Operator<? extends OperatorDesc> operator =
         (Operator<? extends OperatorDesc>) work;
       final int visitCnt = operatorVisits.merge(operator, 1, Integer::sum);
+      if (conf == null && this.work != null && this.work.getFetchTask() != null) {
+          conf = this.work.getFetchTask().getConf();
+      }
```
zabetak: The initial NPE may be avoided but it is not guaranteed that `this.work.getFetchTask().getConf()` will be not null so technically the code could still hit NPE. Furthermore, I see at least 9 references to `conf` inside this class that do not account for the nullability of the field so fixing one of those seems somewhat incomplete. 

Moreover, this kind of fallback strategy to obtain a conf object from the internal fetch task else does not seem to be a widespread pattern so not sure what's the reasoning behind it.

Since the problem seems to be triggered mainly by `HiveProtoLoggingHook` I was wondering if we should rather ensure that the respective class passes a non-null configuration object even if that is the default one (`new HiveConf()`).

Another alternative would be to do something like the following:
```java
final int limit = conf == null ? ConfVars.HIVE_EXPLAIN_NODE_VISIT_LIMIT.defaultIntVal
          : conf.getIntVar(ConfVars.HIVE_EXPLAIN_NODE_VISIT_LIMIT);
```
that definitely prevents the NPE no matter what, but it is still a localized fix.

zabetak (changes requested):
Thanks for the PR @Aggarwal-Raghav ! I left some questions/suggestions for the proposed fix. 

For completeness, it would be nice to also add a unit test possibly somewhere in `TestHiveProtoLoggingHook` or `TestHiveHookEventProtoPartialBuilder`. Some fixes are trivial but writting also tests about them increases test coverage and guards against future regressions.

ql/src/java/org/apache/hadoop/hive/ql/exec/ExplainTask.java
Aggarwal-Raghav: Thanks for the review and suggestions @zabetak and will try to come up with a UT as well. 

**My understanding:**
1. I also though of passing the non-null config from `HiveProtoLoggingHook` class  but it would require some method parameter/argument changes across classes and might need additional check if prologging is disabled. So didn't gave a try 🙁.
2. The reason behind using `this.work.getFetchTask().getConf()` is that I wanted to use the conf which was present in ExplainTask without propogating it from other classes.
3. As you said that there are 9 references of conf but while debugging only _ExplainTask#outputPlan()_ having `conf.getIntVar(ConfVars.HIVE_EXPLAIN_NODE_VISIT_LIMIT)` is getting triggered. So I think handling, the localized place should be enough.
5. Using `ConfVars.HIVE_EXPLAIN_NODE_VISIT_LIMIT.defaultIntVal`, as you suggested, can work as using the default value shouldn't have any impact on OtherInfoType.CONF. Will test and confirm.

zabetak: Regarding 1., I was thinking that changing just a bit HiveHookEventProtoPartialBuilder should work to avoid this NPE:
```java
private JSONObject getExplainJSON(ExplainWork explainWork) throws Exception {
    ExplainTask explain = (ExplainTask) TaskFactory.get(explainWork, new HiveConf());
    return explain.getJSONPlan(null, explainWork, stageIdRearrange);
}
```
This approach seems better than 4, although it doesn't eliminate the potential NPE from other code paths.

The limit could also become part of the `ExplainConfiguration` but I guess this change would have bigger impact.

Anyways, there are many more options, so I will leave the final decision up to you.

Aggarwal-Raghav:
Just info: [HIVE-27240](https://issues.apache.org/jira/browse/HIVE-27240), was very similar to this issue based on stacktrace, which shows the need for UT :-)

Aggarwal-Raghav:
Hi @zabetak, apologies for late reply, just wanted to let you know that I'm working on this PR and found some new issues like [HIVE-29053](https://issues.apache.org/jira/browse/HIVE-29053) along the way when coming up with UT.

Writing UT for this PR is not straightforward and **need your opinion on this**. There are 2 things to note:

1. In case of NPE, its just log the error and continue as intended, which means using `TestHiveHookEventProtoPartialBuilder`   or `TestHiveProtoLoggingHook` won't cut it IMO (tried with mockito as well, didn't work).
2. Writes/flush of proto data are not consistent based on ScheduledThreadPoolExecutor https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/hooks/HiveProtoLoggingHook.java#L277,
one way to force flush is to stop the HS2.

What I/m thinking is, writing a itest/hive-unit integration test as below code. Because of [2], I'm not able to make qfile work as the proto data is written post test completition.
```
CREATE TABLE tbl1 (id INT, name STRING);
INSERT INTO tbl1 VALUES (1, 'Alice'), (2, 'Bob');
SELECT * FROM tbl1;

msck repair table sys.proto_hive_query_data;

select count(*) from sys.proto_hive_query_data where lower(trim(regexp_replace(get_json_object(otherinfo[1].value, '$.queryText'), '\n|\\\s+', ' '))) = "select * from tbl1";
```

Because of NPE the QUERY Object will not be part of proto data so without fix the output should be 0, with fix it should be 1.

zabetak:
The end to end approach (2.) seems a bit of an overkill. Even if the NPE is caught and logged you still have some missing entries (e.g., `queryPlan`) in the `otherInfo` map. You could probably base your test on constructing a simple `ExplainWork` object and reason about missing entries in the info map. Would that work as a unit test?

Aggarwal-Raghav:
> The end to end approach (2.) seems a bit of an overkill. Even if the NPE is caught and logged you still have some missing entries (e.g., `queryPlan`) in the `otherInfo` map. You could probably base your test on constructing a simple `ExplainWork` object and reason about missing entries in the info map. Would that work as a unit test?

Thanks for the suggestions/help @zabetak. I have come up with a test case highlighting the problem observed in local cluster. Can you please review and confirm if its up to the standards?

Aggarwal-Raghav:
Test failure `org.apache.hadoop.hive.metastore.security.TestZookeeperTokenStoreSSLEnabled`, Seems irrelevant to change.

ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
```diff
@@ -77,6 +99,24 @@ private HiveHookEvents.HiveHookEventProto buildIn2Steps(JSONObject json) {
...
     otherInfo.put(HiveProtoLoggingHook.OtherInfoType.CONF, json);
-    return new HiveHookEventProtoPartialBuilder(builder, null, otherInfo, null, null).build();
+    return new HiveHookEventProtoPartialBuilder(builder, work, otherInfo, null, stageIdRearrange).build();
+  }
+
+  private ExplainWork getWork() {
+    CompilationOpContext cCtx = new CompilationOpContext();
+    TableScanOperator scanOp = new TableScanOperator(cCtx);
+
+    FetchWork taskWork = new FetchWork(mock(Path.class), mock(TableDesc.class));
```
zabetak: Do we really need mocking? Path and TableDesc should be easy to create no?

ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
```diff
@@ -77,6 +99,24 @@ private HiveHookEvents.HiveHookEventProto buildIn2Steps(JSONObject json) {
             .setExecutionMode(TEZ);
     Map<HiveProtoLoggingHook.OtherInfoType, JSONObject> otherInfo = new HashMap<>();
     otherInfo.put(HiveProtoLoggingHook.OtherInfoType.CONF, json);
-    return new HiveHookEventProtoPartialBuilder(builder, null, otherInfo, null, null).build();
+    return new HiveHookEventProtoPartialBuilder(builder, work, otherInfo, null, stageIdRearrange).build();
+  }
+
+  private ExplainWork getWork() {
```
zabetak: This is not a getter but rather a factory so it would make more sense to name it accordingly. For instance:
* createExplainWork
* emptyExplainWork
* emptyWork
* ...

ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
```diff
@@ -77,6 +99,24 @@ private HiveHookEvents.HiveHookEventProto buildIn2Steps(JSONObject json) {
...
+    CompilationOpContext cCtx = new CompilationOpContext();
+    TableScanOperator scanOp = new TableScanOperator(cCtx);
+
+    FetchWork taskWork = new FetchWork(mock(Path.class), mock(TableDesc.class));
+    taskWork.setSource(scanOp);
+
+    // Create a FetchTask with the FetchWork
```
zabetak: The comment does not add much value; its pretty obvious that we create a task with some work.

ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
zabetak: The method can also be made static.

ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
```diff
@@ -49,6 +58,15 @@ public void testEquality() {
     assertArrayEquals(event1.toByteArray(), event2.toByteArray());
   }
 
+  @Test
+  public void testOtherInfoQueryPlan() {
+    String stageIdRearrange = HiveConf.ConfVars.HIVE_STAGE_ID_REARRANGE.defaultStrVal;
+    HiveHookEvents.HiveHookEventProto event = buildIn2Steps(mock(JSONObject.class), getWork(), stageIdRearrange);
```
zabetak: Why mock a JSONObject it's equally simple to do `new JSONObject()`?

ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
```diff
@@ -66,6 +84,10 @@ private HiveHookEvents.HiveHookEventProto buildWithOtherInfo(JSONObject json) {
   }
 
   private HiveHookEvents.HiveHookEventProto buildIn2Steps(JSONObject json) {
+    return buildIn2Steps(json, null, null);
+  }
```
zabetak: This is not production code so we don't care much about retaining backward compatibility. Let's just keep the new method.

ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
```diff
@@ -77,6 +99,24 @@ private HiveHookEvents.HiveHookEventProto buildIn2Steps(JSONObject json) {
             .setExecutionMode(TEZ);
     Map<HiveProtoLoggingHook.OtherInfoType, JSONObject> otherInfo = new HashMap<>();
     otherInfo.put(HiveProtoLoggingHook.OtherInfoType.CONF, json);
-    return new HiveHookEventProtoPartialBuilder(builder, null, otherInfo, null, null).build();
+    return new HiveHookEventProtoPartialBuilder(builder, work, otherInfo, null, stageIdRearrange).build();
```
zabetak: Since stageIdRearrange receives only one value during these tests consider setting directly here.

ql/src/test/org/apache/hadoop/hive/ql/hooks/TestHiveHookEventProtoPartialBuilder.java
```diff
@@ -49,6 +58,15 @@ public void testEquality() {
     assertArrayEquals(event1.toByteArray(), event2.toByteArray());
   }
 
+  @Test
+  public void testOtherInfoQueryPlan() {
+    String stageIdRearrange = HiveConf.ConfVars.HIVE_STAGE_ID_REARRANGE.defaultStrVal;
+    HiveHookEvents.HiveHookEventProto event = buildIn2Steps(mock(JSONObject.class), getWork(), stageIdRearrange);
+    // Without HIVE-29012, OTHERINFO will have CONF initially i.e size is 1.
+    // With HIVE-29012, OTHERINFO should have QUERY as well i.e size is 2.
+    assert(event.getOtherInfo(1).hasKey());
```
zabetak: I believe that using a java assertion here was not intentional; please replace this with a junit assertion. 

In addition, consider using `assertEquals` with the full `List` instead of assertTrue/False. The failure message will be much more descriptive if something breaks. 

Moreover, instead of putting the expectations in a comment (i.e., CONF + QUERY) just create manually the expected List and base your assertion there.

zabetak (changes requested):
Thanks for your effort @Aggarwal-Raghav !! The prod changes LGTM. 

I just have some small (mostly nit) comments regarding the test. There are two best-practices that I wanted to highlight here:
* Avoid mocking when its easy/simple to create real objects
* Avoid using java `assert` for tests

In some cases, "best-practices" can be subjective but I feel the above are applicable/beneficial in every project.

Aggarwal-Raghav:
@zabetak , I have addressed the review comments. can you please take a look?

zabetak (approved):
LGTM, will merge it soon.

zabetak:
Thanks for the PR @Aggarwal-Raghav !

PR #5879, opened 2025-06-17
HIVE-28667: Error initializing field trimmer instance when starting HS2

zabetak:
1. Change `trimFields(JdbcRel)` to `trimFields(JdbcSort)` to break the ambiguity and avoid the error. 
2. Make `CalcitePlanner#HIVE_REL_NODE_CLASSES` public mainly to facilitate testing. In fact, it makes sense to expose this information somewhere so why not in CalcitePlanner.

Avoid the ambiguity exception during HS2 startup. For more details see commit message or HIVE-28667.

No. There is no change in behavior since anyways the trimmer cannot be called with a JdbcSort operator.

```
mvn test -Dtest=TestHiveRelFieldTrimmer -pl ql
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile_regex=".*jdbc.*" -pl itests/qtest -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/RelFieldTrimmer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/RelFieldTrimmer.java
@@ -19,7 +19,6 @@ package org.apache.hadoop.hive.ql.optimizer.calcite.rules;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import org.apache.calcite.adapter.jdbc.JdbcConvention;
-import org.apache.calcite.adapter.jdbc.JdbcRel;
 import org.apache.calcite.adapter.jdbc.JdbcRules.JdbcAggregate;
 import org.apache.calcite.adapter.jdbc.JdbcRules.JdbcAggregateRule;
 import org.apache.calcite.adapter.jdbc.JdbcRules.JdbcFilter;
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/RelFieldTrimmer.java
```diff
@@ -587,6 +576,13 @@ public TrimResult trimFields(
     return result(relBuilder.build(), inputMapping);
   }
 
+  public TrimResult trimFields(
+      JdbcSort rel,
```
ngsg, line 580: If I understand correctly, `trimFields` for `JdbcSort` is now hiding `trimFields` for `Sort`. Then do we need `trimFields` for `JdbcUnion` to hide the one for `SetOp` as well?

zabetak: In fact almost all `trimFields` methods for `Jdbc` operators are pretty much useless so I am planning to remove them in HIVE-29025. Here, I just wanted a minimal fix that breaks the ambiguity but even with the new code these method will never be called.

ngsg: I understand. Thanks for the explanation.

ngsg:
The patch looks good to me, +1(non-biding).

okumin (approved):
+1

I verified that the error message is still displayed with the master branch, and it is never shown with this patch.

zabetak:
Many thanks for the reviews @ngsg @okumin !

PR #5880, opened 2025-06-17
HIVE-29027: Move to JDK-21

ayushtkn:
**Tasks**
- [X] **Hive Dev Box to support JDK-21** --> Built custom image with jdk-21 support & pushed to [DockerHub](https://hub.docker.com/r/ayushtkn/hive-dev-box/tags)
- [X] **DockerFiles to use JDK-21** -> Bump JDK version in docker images
- [X] **System.setSecurityManager removal in JDK-21** --> Replaced System.exit() & setSecurityManager with ExitUtil (Hadoop Solution) Moves to HIVE-29056
- [X] **Removal of Thread.stop()** -> Thread.stop is removed in JDK-21 --> Replaced with interrupt & will thread handle (Replaced in a deprecated test only method otherwise interrupt ain't a direct replacement of stop)
- [X] **Removal of Reflections hack** -> JDK-21 is more strict it doesn't let hack the Final fields -> test class had field Final removed that final from test class & dropped the hack
- [X] **JWT tests modifying System env** --> Not supported in JDK-21, tried some hacks to prevent not dropping those tests, may not be exact copy but better than dropping
- [X] **Decimal/Float precession change in JDK-21** --> JDK-21 had changes around the precession, leads to query output regen, haven't found any hack to use the old behaviour, as of now can't help it but adapt. https://inside.java/2022/09/23/quality-heads-up/
https://bugs.openjdk.org/browse/JDK-4511638
- [X] **Kudu tests failing** -> Tests failing since Kudu uses internal java classes removed in JDK-21 raised [KUDU-3670](https://issues.apache.org/jira/browse/KUDU-3670). --> Replaced the util which runs the Kudu Cluster with directly running the Kudu cluster


aturoczy:
Instant approve :)

ayushtkn:
Down to some 24 Failures, of which 2 flaky (they passed in previous build). 22 are all from Kudu Handler. Have created a ticket [KUDU-3670](https://issues.apache.org/jira/browse/KUDU-3670).

deniskuzZ:
Trino is already on JDK-24, doesn't even compile with lower versions

ayushtkn:
Green Build with JDK-21 :-)

deniskuzZ:
134 files changed 🥇

simhadri-g:
Awesome, will this be in hive v5.0 ? :)

deniskuzZ:
@ayushtkn, did you grep for
````
    <maven.compiler.source>8</maven.compiler.source>
    <maven.compiler.target>8</maven.compiler.target>
````
see https://issues.apache.org/jira/browse/HIVE-29018

**Comment by ayushtkn** [comment]
I didn't knew something was on 8 :-) 
I just pushed upgrading it to 21, ran tests locally in that module, they were passing, rest will see
```
[INFO] Results:
[INFO] 
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0

```

File: standalone-metastore/metastore-rest-catalog/pom.xml [hunk 1]
```diff
@@ -21,8 +21,8 @@
   <name>Hive Metastore REST Catalog</name>
   <properties>
     <standalone.metastore.path.to.root>..</standalone.metastore.path.to.root>
-    <maven.compiler.source>8</maven.compiler.source>
-    <maven.compiler.target>8</maven.compiler.target>
+    <maven.compiler.source>21</maven.compiler.source>
```
deniskuzZ: please drop them, they are already defined in `standalone-metastore` root pom.
it was just a copy&paste

deniskuzZ:
@ayushtkn, what are the plans here? should we wait with the merge until 4.1.0 cut-off ?

ayushtkn:
That is what I am thinking, hold it till branch 4.1.0 is cut & then chase this for master branch

pan3793:
I believe changes of `*.java` are also applicable to Java 17, can those changes go branch-4.1 too? then 4.1.x should be compatible with both Java 17 & 21

aturoczy:
@ayushtkn is there any blocker (beside +1's) to merge this change?

ayushtkn:
@aturoczy nothing from my side as such

aturoczy:
https://tenor.com/hu/view/yes-gif-22712908

cnauroth (approved):
+1 (non-binding). Great work @ayushtkn !

abstractdog:
Nice work, @ayushtkn!

Does ExitUtil work without the JDK 21 upgrade? If so, how about creating a separate ticket for it and marking it as a blocking dependency for this one? It already has standalone value, and splitting it out would make this PR much clearer.

This applies to all the changes here that are compatible with earlier JDK versions, but since the ExitUtil changes are quite substantial, I’d prefer to extract them into a separate patch.

ayushtkn:
>  If so, how about creating a separate ticket for it and marking it as a blocking dependency for this one? It already has standalone value, and splitting it out would make this PR much clearer.

Thanx @abstractdog for the review. The ExitUtils change is now separated

Jenkinsfile
```diff
@@ -129,7 +129,7 @@ def sonarAnalysis(args) {
 def hdbPodTemplate(closure) {
   podTemplate(
   containers: [
-    containerTemplate(name: 'hdb', image: 'wecharyu/hive-dev-box:executor', ttyEnabled: true, command: 'tini -- cat',
+    containerTemplate(name: 'hdb', image: 'ayushtkn/hive-dev-box:executor', ttyEnabled: true, command: 'tini -- cat',
```
abstractdog, line 132: note: this definitely needs a follow-up (I won't block this change), because we still haven't taken care of adapting the hive-dev-box, so now we have different images on different branches, still no version-controlled Dockerfile + image in an official apache docker repo

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestRemoteHiveMetastoreWithHttpJwt.java
```diff
@@ -160,88 +116,53 @@ private static void setupMockServer() throws Exception {
...
-    try {
-      Database createdDb = new Database();
-      createdDb.setName(dbName);
-      client.createDatabase(createdDb);
-      Database dbFromServer = client.getDatabase(dbName);
-      assertEquals(dbName, dbFromServer.getName());
-    } finally {
-      try {
+
+    SystemLambda.withEnvironmentVariable("HMS_JWT", validJwtToken).execute(() -> {
```
abstractdog: did some research, system-stubs is told to be more junit-like and more maintained: https://github.com/webcompere/system-stubs (MIT licensed similarly to SystemLambda), would you consider using it?

abstractdog:
PR looks good to me—I'm open to giving a +1.

The only potentially debatable part is the precision change, which results in many q.out diffs. However, looking at the underlying Java change:
🔗 https://github.com/openjdk/jdk/pull/3402/files
which was eventually committed here:
🔗 https://github.com/openjdk/jdk/commit/72bcf2aa03d53b0f68eb07a902575b4e8628d859

The fact that even Java’s golden files changed is pretty telling—so it is what it is.
For the Hive release, we can note that this is due to [JDK-4511638](https://bugs.openjdk.org/browse/JDK-4511638).


assuming that @ayushtkn will consider my comment regarding systemlambda vs. system stubs, LGTM

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestRemoteHiveMetastoreWithHttpJwt.java
ayushtkn: Thanx @abstractdog for the pointers, I have replaced it as you suggested :-)

abstractdog (approved):
LGTM +1
Would you be so kind, @ayushtkn, to wait another 24 hours in case anyone has objections?

PR #5882, opened 2025-06-20
HIVE-29035: Fixing cache handling for REST catalog

henrib:
- add logic check on cache to ensure returned table is not stale

Adding logic to verify if a cached table actually points to the latest known version

Cache handling for REST Iceberg catalog may return stale tables.

No

Unit tests


standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -24,30 +24,45 @@
...
+    super(catalog, false, expiration, Ticker.systemTicker());
     nsCatalog = catalog;
   }
 
   public CATALOG hmsUnwrap() {
     return nsCatalog;
   }
 
+  public void invalidateTable(String dbName, String tableName) {
+    super.invalidateTable(TableIdentifier.of(dbName, tableName));
```
saihemanth-cloudera: I think we can call l#60 here?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -24,30 +24,45 @@
...
+  }
+
+  @Override
+  public void invalidateTable(TableIdentifier tableIdentifier) {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Avoid invalidating table: {}", tableIdentifier);
+    }
```
saihemanth-cloudera: We can then do super.invalidateTable() here? Also, why not do `nsCatalog.invalidateTable()` here?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: This version of invalidateTable() is called by the listener and we do want the invalidation performed here. The override at l#60 does nothing on purpose; we do want all invalidations performed through the listener.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: The idea is to try and have all actual invalidation performed through the listener; this override voids the default call on purpose.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSEventListener.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.hadoop.hive.metastore.events.DropTableEvent;
+import org.apache.hadoop.hive.metastore.events.ReloadEvent;
+import org.apache.iceberg.catalog.Catalog;
+import org.slf4j.Logger;
+
+/**
+ * IcebergEventListener is a Hive Metastore event listener that invalidates the cache
+ * of the HMSCachingCatalog when certain events occur, such as altering or dropping a table.
+ */
+public class HMSEventListener extends MetaStoreEventListener {
```
okumin: I'm new to this API. Does this work even when we set up HMS with HA with `metastore.thrift.uri.selection=RANDOM`?

henrib: Good question, I would hope so, let me try and seek a definite answer.

henrib: HMS only triggers MetaStoreEventListener for 'local' events so HA would need to disable the cache (for now). A potential solution would be to actively poll events from each HMS instance (need a dedicated thread that polls the NotificationEventRequest and follows the same invalidation logic).

okumin: As I shared it with Jira, HiveCatalog might access two types of remote resources.
- (A) Pointers from table names to the metadata location in the backend RDBMS
- (B) The content of metadata.json on Amazon S3

I'm wondering if we can update HiveCatalog to allow caching (B). The metadata.json is immutable. So, we can safely cache it.
I'd say RDMBS for Iceberg pointers is fast enough, such as 10k RPS.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
```diff
@@ -35,19 +38,33 @@
...
-  private static final String SERVLET_ID_KEY = "metastore.in.test.iceberg.catalog.servlet.id";
+ private static final String SERVLET_ID_KEY = "metastore.in.test.iceberg.catalog.servlet.id";
 
-  private final Configuration configuration;
-  private final int port;
-  private final String path;
+  /**
+   * Convenience soft reference to last catalog.
+   */
+  protected static final AtomicReference<Reference<Catalog>> catalogRef = new AtomicReference<>();
```
deniskuzZ: remove it

henrib: This is the only way I found to pass the catalog instance to the event listener. This is a partial solution (non HA); the HA version would require polling from the notifications through a thread and there would still be a propagation delay and potential stale entries in 'distributed' caches; a performance trade-off...
Shall I close this ? No point in continuing if you've already reached hard '-1'.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSEventListener.java
henrib: Wouldn't caching the file-system layer from the table layer (ie B) somehow be redundant ? The (distributed) fs and UGI already cache some info, dont they ?
I understand that caching the Table object itself comes with a propagation/staleness trade-off in HA mode but can't this be a valid one ? Any user transaction will run on a given snapshot, not necessarily the last one...

okumin: Let me try to explain my intention. As you know, the location of the metadata top is stored in `metadata_location`. The most important point is `file:/opt/hive/data/warehouse/test/metadata/00000-165e08e2-9cea-47b1-9eef-45921f238fe2.metadata.json` is immutable. Whenever we update the table, the path will change.

```
| CREATE EXTERNAL TABLE `test`(                      |
|   `id` int)                                        |
| ROW FORMAT SERDE                                   |
|   'org.apache.iceberg.mr.hive.HiveIcebergSerDe'    |
| STORED BY                                          |
|   'org.apache.iceberg.mr.hive.HiveIcebergStorageHandler'  |
|                                                    |
| LOCATION                                           |
|   'file:/opt/hive/data/warehouse/test'             |
| TBLPROPERTIES (                                    |
|   'TRANSLATED_TO_EXTERNAL'='TRUE',                 |
|   'bucketing_version'='2',                         |
|   'current-schema'='{"type":"struct","schema-id":0,"fields":[{"id":1,"name":"id","required":false,"type":"int"}]}',  |
|   'external.table.purge'='TRUE',                   |
|   'format-version'='2',                            |
|   'iceberg.orc.files.only'='false',                |
|   'metadata_location'='file:/opt/hive/data/warehouse/test/metadata/00000-165e08e2-9cea-47b1-9eef-45921f238fe2.metadata.json',  |
```

Therefore, the following code does not include the staleness problem.

```java
Map<String, org.apache.iceberg.Table> cache = new HashMap(); // We should use Guava's cache or something
cache.put("file:/opt/hive/data/warehouse/test/metadata/00000-165e08e2-9cea-47b1-9eef-45921f238fe2.metadata.json", icebergTable);
```

On `loadTable`, we will use the cache in the following way.

```java
org.apache.hadoop.hive.metastore.api.Table hmsTable = rdbms.getTable(tableName);
String metadataLocation = hmsTable.getParameters().get("metadata_location"); // this might be wrong
org.apache.iceberg.Table cachedTable = cache.get(metaedataLocation);
if (cachedTable != null) {
  return cachedTable;
}
// We will access S3, and then build the org.apache.iceberg.Table
```

This is suboptimal because we access the RDBMS every time. However, we can eliminate S3 latency without staleness issues. I haven't run the PoC, so I'm not sure if it really works.

okumin: If we cache the RDBMS entry, I think it can be on the pure HMS side. I haven't checked if it's possible or not.
P.S. This means, if we can cache `table_type=ICEBERG`, we can probably cache other types of tables. Iceberg REST can be just one user of the caching mechanism if we can push it into the normal HMS side. If it is impossible, probably, it could be a sign that we should not cache `table_type=ICEBERG`

henrib: Ok, got it; by using the metadata location as table cache key & peeking it in the RDBMS each time we loadTable, we avoid accessing a stale version of the table. We can still actively remove known stales from the cache and have some TTL/expiry to avoid keeping tables accessed once in the cache.

henrib: Actually, with the same logic and reusing the existing code, on loadTable() we can verify that if a table is cached, its location is the same as the one known in Hive DB.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
henrib: I've removed this and changed the logic in the latest iteration to check whether a cached table still points to the most up to date location (reusing @okumin idea).

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -35,24 +36,22 @@
...
 /**
  * Class that wraps an Iceberg Catalog to cache tables.
  */
 public class HMSCachingCatalog extends CachingCatalog implements SupportsNamespaces, ViewCatalog {
+  private static final Logger LOG = LoggerFactory.getLogger(HMSCachingCatalog.class);
   private final HiveCatalog hiveCatalog;
   
   public HMSCachingCatalog(HiveCatalog catalog, long expiration) {
-    super(catalog, true, expiration, Ticker.systemTicker());
+    super(catalog, false, expiration, Ticker.systemTicker());
```
okumin, line 58: I'm curious about pros vs cons of the case sensitivity here.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -63,6 +62,27 @@ public List<Namespace> listNamespaces(Namespace nmspc) throws NoSuchNamespaceExc
...
+        LOG.debug("Table {} has no location, returning cached table without location", canonicalIdentifier);
+      } else if (!location.equals(cachedTable.location())) {
+        LOG.debug("Cached table {} has a different location than the one in the catalog: {} != {}",
+                 canonicalIdentifier, cachedTable.location(), location);
+      } else {
+        LOG.debug("Returning cached table: {}", canonicalIdentifier);
+        return cachedTable;
+      }
+      // Invalidate the cached table if the location is different
+      tableCache.invalidate(cachedTable);
```
okumin: I guess it's a little more robust to use `CachingCatalog#invalidateTable`. Please correct me if the current implementation is intentional.
https://github.com/apache/iceberg/blob/apache-iceberg-1.9.2/core/src/main/java/org/apache/iceberg/CachingCatalog.java#L186-L191

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: Hive is case-insensitive for metadata including table names ; not sure if catalog caller normalizes it so taking the safe choice (belt & suspender).

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: Updated to a more robust and more efficient version.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
okumin: I could not agree more.

```
0: jdbc:hive2://localhost:10000/> create table tEsT (nAmE STRING);
...
0: jdbc:hive2://localhost:10000/> show create table test;
...
+----------------------------------------------------+
|                   createtab_stmt                   |
+----------------------------------------------------+
| CREATE EXTERNAL TABLE `test`(                      |
|   `name` string)                                   |
```

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
```diff
@@ -33,21 +33,23 @@
 
 /**
  * Catalog &amp; servlet factory.
+ * <p>This class is derivable on purpose; the factory class name is a configuration property, this class
+ * can serve as a base for specialization.</p>
  */
 public class HMSCatalogFactory {
-  private static final String SERVLET_ID_KEY = "metastore.in.test.iceberg.catalog.servlet.id";
+ private static final String SERVLET_ID_KEY = "metastore.in.test.iceberg.catalog.servlet.id";
```
okumin:
```suggestion
  private static final String SERVLET_ID_KEY = "metastore.in.test.iceberg.catalog.servlet.id";
```

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
```diff
@@ -33,21 +33,23 @@
 
 /**
  * Catalog &amp; servlet factory.
+ * <p>This class is derivable on purpose; the factory class name is a configuration property, this class
+ * can serve as a base for specialization.</p>
```
okumin, line 37: What extends this class? If nothing does it, I'd like you to revert the visibility of methods and variables from `protected` to `private`.

henrib: It is meant to allow easy extension (different versions of Iceberg REST API) and ease implementation of added features (like authorization handling); it does not harm Hive to keep this convenient way to add features with no disruption to the code line.

okumin: I got your point, and I respectfully have a small challenge 😃 . I have two points about why we can avoid `protected` here. First, I suppose most Java projects recommend using as restrictive accessibility[1] as possible, and Hive's source code follows the principle[2]. We need use cases when we use generous modifiers. Second, we can use not an inheritance but a composition when we really want to add some traits to `HMSCatalogFactory`; We will lose flexibility with `protected`.

- [1] [Controlling Access to Members of a Class](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html), Effective Java
- [2] I understand we have some implementation that exceptionally abuses non-private modifiers

henrib: I hear and welcome your challenge 😃 ; let me try and further explain why I believe this 'pattern of extensible inheritable factory' is an acceptable compromise in this context.
I actually embrace Mr Bloch rule phrased as 'favor composition over inheritance'; we are in a delineated case where we only want to allow real subtype of our factory.
If this factory class were to become composable, this would imply we can delegate to an instance and call its methods from a user defined class instance; then all its methods should become public including its constructor. And this could lead towards exposing an interface, adding even more visibility and more complexity.
If we switch back to private, no user defined class can use Hive's instance which will force anyone to duplicate (almost) all of its code to add specialization. Certainly a better alternative than the previous one at the expense of code duplication.
The protected version allows surgical precision in extension. Since the catalog cache already delegates to the Catalog instance, it is a very attractive and convenient extension point. We could even make that more obvious by exposing a cache wrapping method.
Let me know what you think.

okumin: Thanks. Can we separate it from this PR? The title of this PR is "Fixing cache handling for REST catalog" and your point does not belong to the original interest of this PR.
I suppose we need to clarify the three points to justify the new ticket. One, we need clear use cases to create a subclass. I don't rule out the potential use cases, but "potential" does not justify the change; Every class is likely to have a potential use case to be inherited. Two, we will need to be confident that `extends HMSCatalogFactory` is the best method to satisfy the use case. We are recently removing inheritance[1] because we failed to manage inflated complexity using inheritance. I suppose we would likely be more successful with YAGNI in this case; We can add the best abstraction, which will have real users.
I can approve and merge this PR if we can separate the problem from this PR.

- [1] https://docs.google.com/document/d/1lppPj35fBOOCjGNIouoJCsniMPpyzWx3zcE7kvu0Be4/edit?tab=t.0#heading=h.1upmmt6vhhfa

okumin: Just for your information, I'm not confident about what our way should be. It may turn out that you may be right. Regardless of the decision, this PR looks awesome already 🏆 ; The caching feature is probably more needed than the convenience of the potential user-side extensibility. As our dev or review capacity is limited[1], I prefer to merge the important part first.

- [1] [I review many PRs in parallel](https://okumin.com/wip/), but I can work with my laptop for only 1-2 hours per day because of a medical reason

henrib: Extensibility removed from factory in last iteration.

okumin (approved):
LGTM!

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -63,6 +69,48 @@ public List<Namespace> listNamespaces(Namespace nmspc) throws NoSuchNamespaceExc
...
   }
 
+  @Override
+  public Table loadTable(TableIdentifier identifier) {
+    final Cache<TableIdentifier, Table> cache = this.tableCache;
+    final HiveCatalog catalog = this.hiveCatalog;
+    final TableIdentifier canonicalized = identifier.toLowerCase();
+    Table cachedTable = cache.getIfPresent(canonicalized);
+    if (cachedTable != null) {
+      String location = catalog.getTableLocation(canonicalized);
```
deniskuzZ, line 79: @henrib, did you even test the code ??? I don't see even a single test that validates the caching functionality

`catalog.getTableLocation(canonicalized)`
file:/var/folders/qy/whx78wjd1w72_z8_wgqfwd_c0000gn/T/RESTCatalogServer_b5e4f53d-d4c9-4b1a-8897-c9b9e82ef29e/external/newdb.db/newtable/metadata/00000-fd422e1e-bf30-4529-9baf-606662a2253a.metadata.json

`cachedTable.location()`
file:/var/folders/qy/whx78wjd1w72_z8_wgqfwd_c0000gn/T/RESTCatalogServer_b5e4f53d-d4c9-4b1a-8897-c9b9e82ef29e/external/newdb.db/newtable

one returns metadata file location and another just the root table location. I am reverting this

henrib: To ensure the cached table is not stale, we check against the stored reference that the location it points to is still current. This situation (staleness) can occur if users are updating the table from HMS/Hive concurrently with through Catalog and/or HA. @okumin  do you concur ?

okumin: The current server-side cache assumes the following race conditions. Say we have two HAed HMS instances X and Y with RANDOM load-balancing.

1. User A caches a table at T1 in HMS X
2. User B updates the same table to T2 and caches it in HMS Y
3. User C fetches the same table from HMS X, and HMS X returns the table at T1

I suppose the server-side eventual consistency requires us to design and advertise the consistency semantics carefully. I'd say the client side should do such a control.

deniskuzZ: location objects won't ever match !!!
one returns metadata file location and another just the root table location

deniskuzZ: > I suppose the server-side eventual consistency requires us to design and advertise the consistency semantics carefully. I'd say the client side should do such a control.

@okumin, I think similar thing was discussed in https://lists.apache.org/thread/zb5ldjfxvjz4fjwll1lc9cm8rl015zx2

henrib: > location objects won't ever match !!! one returns metadata file location and another just the root table location

You are absolutely right, I must missed that during debug, thanks for catching it. It could have been fixed with a call to:
```
  private static String getMetadataLocation(final Table table) {
    if (table instanceof HasTableOperations tableOps) {
      final TableOperations ops = tableOps.operations();
      final TableMetadata meta;
      if (ops != null && (meta = ops.current()) != null) {
        return meta.metadataFileLocation();
      }
    }
    return null;
  }
```

deniskuzZ: 1 remark: `hive-catalog` is a clone of iceberg's `hive-metastore`. So whenever we need to apply changes there - we should raise a PR for iceberg as well.

maybe instead of adding `getTableLocation` we could use 
````
getMetadataLocation(catalog.loadTable(tableidentifier))
````

File: standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java [hunk 10]
```diff
@@ -63,6 +69,48 @@ public List<Namespace> listNamespaces(Namespace nmspc) throws NoSuchNamespaceExc
...
+                 canonicalized, cachedTable.location(), location);
+        // Invalidate the cached table if the location is different
+        invalidateTable(canonicalized);
+      } else {
+        LOG.debug("Returning cached table: {}", canonicalized);
+        return cachedTable;
+      }
+    }
+    Table table = cache.get(canonicalized, catalog::loadTable);
+    if (table instanceof BaseMetadataTable) {
```
deniskuzZ, line 93: are we only supposed to cache metadata tables?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
okumin: > location objects won't ever match !!!

You're right. My bad. Let's add an addendum patch.

> 1 remark: hive-catalog is a clone of iceberg's hive-metastore. So whenever we need to apply changes there - we should raise a PR for iceberg as well.

I'm not finding the policy about how to manage cloned files, but it is also right anyway. I'm wondering if we can push the entire caching feature into HiveCatalog in apache/iceberg. I can try it in my spare time if no one tries it. Probably off topic

> @okumin, I think similar thing was discussed in https://lists.apache.org/thread/zb5ldjfxvjz4fjwll1lc9cm8rl015zx2

Thanks for identifying the related discussion. We might want to implement the finally-decided solution, probably the ETag-based one.
- https://github.com/apache/iceberg/issues/11766
- https://docs.google.com/document/d/1rnVSP_iv2I47giwfAe-Z3DYhKkKwWCVvCkC9rEvtaLA/edit?tab=t.0#heading=h.5voryi2hj0go

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: Line 92 does cache non-metadata table. Per the caffeine doc, cache 'get' obeys "if cached, return; otherwise create, cache and return" pattern ; line 93 and after are meant to deal with the specifics of metadata table.
What am I missing ?

henrib:
Newer version in https://github.com/apache/hive/pull/6022 with an actual test that verifies the correct caching behavior.

PR #5883, opened 2025-06-21
HIVE-29034: ALTER TABLE CHANGE COLUMN fails with Oracle backed Hive M…

dengzhhu653:
…etastore

Change-Id: Ia0043312e62b0c2773a41d2422005a2131487dbe

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
@@ -897,8 +897,7 @@ class DirectSqlUpdatePart {
           @Override
           public List<Long> run(List<Long> input) throws Exception {
             String idLists = MetaStoreDirectSql.getIdListForIn(input);
-            String queryText = "select \"CD_ID\" from \"SDS\" where \"CD_ID\" in ( "
-                + idLists + ") group by \"CD_ID\"";
+            String queryText = "select DISTINCT \"CD_ID\" from \"SDS\" where \"CD_ID\" in ( " + idLists + ")";
             List<Long> cdIds = new ArrayList<>();
             try (QueryWrapper query = new QueryWrapper(pm.newQuery("javax.jdo.query.SQL", queryText))) {
               List<Object> sqlResult = executeWithArray(query.getInnerQuery(), null, queryText);
@@ -907,8 +906,8 @@ class DirectSqlUpdatePart {
                   cdIds.add(MetastoreDirectSqlUtils.extractSqlLong(cdId));
                 }
               }
-              return cdIds;
             }
+            return cdIds;
           }
     }));
     List<Long> unusedCdIds = cdIdsMayDelete.stream().filter(id -> !usedIds.contains(id)).collect(Collectors.toList());
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
@@ -892,18 +892,25 @@ class DirectSqlUpdatePart {
         sdIdToNewCdId.containsKey(sdId) ? sdIdToNewCdId.get(sdId) : cdId);
     updateSDInBatch(validSdIds, idToSd, sdIdToCdId);
 
-    List<Long> usedIds = Batchable.runBatched(maxBatchSize, cdIdsMayDelete,
+    Set<Long> usedIds = new HashSet<>(Batchable.runBatched(maxBatchSize, cdIdsMayDelete,
         new Batchable<Long, Long>() {
           @Override
           public List<Long> run(List<Long> input) throws Exception {
             String idLists = MetaStoreDirectSql.getIdListForIn(input);
-            String queryText = "select \"CD_ID\" from \"SDS\" where \"CD_ID\" in ( " + idLists + ")";
+            String queryText = "select \"CD_ID\" from \"SDS\" where \"CD_ID\" in ( "
+                + idLists + ") group by \"CD_ID\"";
+            List<Long> cdIds = new ArrayList<>();
             try (QueryWrapper query = new QueryWrapper(pm.newQuery("javax.jdo.query.SQL", queryText))) {
-              List<Long> sqlResult = executeWithArray(query.getInnerQuery(), null, queryText);
-              return new ArrayList<>(sqlResult);
+              List<Object> sqlResult = executeWithArray(query.getInnerQuery(), null, queryText);
+              if (sqlResult != null) {
+                for (Object cdId : sqlResult) {
+                  cdIds.add(MetastoreDirectSqlUtils.extractSqlLong(cdId));
+                }
+              }
+              return cdIds;
             }
           }
-    });
+    }));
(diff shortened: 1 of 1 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -892,18 +892,25 @@ public List<Void> run(List<Long> input) throws Exception {
...
     updateSDInBatch(validSdIds, idToSd, sdIdToCdId);
 
-    List<Long> usedIds = Batchable.runBatched(maxBatchSize, cdIdsMayDelete,
+    Set<Long> usedIds = new HashSet<>(Batchable.runBatched(maxBatchSize, cdIdsMayDelete,
         new Batchable<Long, Long>() {
           @Override
           public List<Long> run(List<Long> input) throws Exception {
             String idLists = MetaStoreDirectSql.getIdListForIn(input);
-            String queryText = "select \"CD_ID\" from \"SDS\" where \"CD_ID\" in ( " + idLists + ")";
+            String queryText = "select \"CD_ID\" from \"SDS\" where \"CD_ID\" in ( "
```
deniskuzZ: why not `select distinct`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -892,18 +892,25 @@ public List<Void> run(List<Long> input) throws Exception {
...
+                + idLists + ") group by \"CD_ID\"";
+            List<Long> cdIds = new ArrayList<>();
             try (QueryWrapper query = new QueryWrapper(pm.newQuery("javax.jdo.query.SQL", queryText))) {
-              List<Long> sqlResult = executeWithArray(query.getInnerQuery(), null, queryText);
-              return new ArrayList<>(sqlResult);
+              List<Object> sqlResult = executeWithArray(query.getInnerQuery(), null, queryText);
+              if (sqlResult != null) {
+                for (Object cdId : sqlResult) {
```
deniskuzZ, line 906: why not use stream map API?
````
return sqlResult
  .map(MetastoreDirectSqlUtils::extractSqlLong)
  .toList()
````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java [hunk 3]
```diff
@@ -892,18 +892,25 @@ public List<Void> run(List<Long> input) throws Exception {
...
             String idLists = MetaStoreDirectSql.getIdListForIn(input);
-            String queryText = "select \"CD_ID\" from \"SDS\" where \"CD_ID\" in ( " + idLists + ")";
+            String queryText = "select \"CD_ID\" from \"SDS\" where \"CD_ID\" in ( "
+                + idLists + ") group by \"CD_ID\"";
+            List<Long> cdIds = new ArrayList<>();
             try (QueryWrapper query = new QueryWrapper(pm.newQuery("javax.jdo.query.SQL", queryText))) {
-              List<Long> sqlResult = executeWithArray(query.getInnerQuery(), null, queryText);
-              return new ArrayList<>(sqlResult);
+              List<Object> sqlResult = executeWithArray(query.getInnerQuery(), null, queryText);
+              if (sqlResult != null) {
```
deniskuzZ, line 904: Can it be null? Can we move `cdIds` initialization here?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
dengzhhu653: I think there is no difference. In MetastoreDirectSql, we also use `group by` to get the unique `CD_ID`:
https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L3087

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
dengzhhu653: Add the null check just in case, as this `sqlResult` is from calling the dependent lib.
> Can we move cdIds initialization here?

I don't want the `Batchable.runBatched` return a null list, which might be pushed to create a new `HashSet`, nor two `return` lines, one for `cdIds`, the other `Collections.emptyList()`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
okumin: I prefer `SELECT DISTINCT` for readability. But performance might rely on RDBMS.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -892,18 +892,25 @@ public List<Void> run(List<Long> input) throws Exception {
...
+                for (Object cdId : sqlResult) {
+                  cdIds.add(MetastoreDirectSqlUtils.extractSqlLong(cdId));
```
okumin, line 906: Is it necessary to use `extractSqlLong`? If not, we may have a shorthand(with a bit more efficient ArrayList initialization, which is very tiny).

```sql
List<Long> sqlResult = executeWithArray(query.getInnerQuery(), null, queryText);
return sqlResult != null ? new ArrayList<>(sqlResult) : Collections.emptyList();
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
deniskuzZ: The return is inside of try-with-resources block, should we move it out for better redability

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
deniskuzZ: https://www.linkedin.com/pulse/optimizing-sql-performance-distinct-vs-group-fact-analysis-rahman-z5sec/
For relational DBs, `SELECT DISTINCT`  is typically faster or at least simpler than `GROUP BY`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
dengzhhu653: yes, the result returning from Oracle is the BigDecimal

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
dengzhhu653: Done

PR #5884, opened 2025-06-21
HIVE-28983: Log HS2 and HMS PID and update hive-env.sh template

Aggarwal-Raghav:
Log the HS2 and HMS PID in their respective logs and give sample hms and hs2 HADOOP_OPTS in hive-env.sh.template.

During debugging prod issue, I saw the need for this.
1. If HMS/HS2 is crashing periodically then the heapdump will have the name in default format i.e. `java_pid14784.hprof`, it becomes difficult to identify whether it belongs to HMS or HS2 without loading it in any analyzer tool.
2. To get the gc logs for a particular HS2 or HMS crash requires extraction of timestamp from logs using text manipulation commands etc.

Yes, in HS2 and HMS logs PID will be logged

On local setup

### DIFF
--- a/conf/hive-env.sh.template
+++ b/conf/hive-env.sh.template
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
@@ -18,31 +20,6 @@
 # to control the execution of Hive. It should be used by admins to configure
 # the Hive installation (so that users do not have to set environment variables
 # or set command line parameters to get correct behavior).
-#
-# The hive service being invoked (CLI etc.) is available via the environment
-# variable SERVICE
-
-
-# Hive Client memory usage can be an issue if a large number of clients
-# are running at the same time. The flags below have been useful in 
-# reducing memory usage:
-#
-# if [ "$SERVICE" = "cli" ]; then
-#   if [ -z "$DEBUG" ]; then
-#     export HADOOP_OPTS="$HADOOP_OPTS -XX:NewRatio=12 -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:+UseParNewGC -XX:-UseGCOverheadLimit"
-#   else
-#     export HADOOP_OPTS="$HADOOP_OPTS -XX:NewRatio=12 -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:-UseGCOverheadLimit"
-#   fi
-# fi
-
-# The heap size of the jvm stared by hive shell script can be controlled via:
-#
-# export HADOOP_HEAPSIZE=1024
-#
-# Larger heap size may be required when running queries over large number of files or partitions. 
-# By default hive shell scripts use a heap size of 256 (MB).  Larger heap size would also be 
-# appropriate for hive server.
-
 
 # Set HADOOP_HOME to point to a specific hadoop install directory
 # HADOOP_HOME=${bin}/../../hadoop
@@ -52,3 +29,45 @@
 
 # Folder containing extra libraries required for hive compilation/execution can be controlled by:
 # export HIVE_AUX_JARS_PATH=
+
+# HIVE Log Directory can be controlled by and ensure it exists on the filesystem:
+# export HIVE_LOG_DIR=
+
+# The hive service being invoked (beeline etc.) is available via the environment
+# variable SERVICE
+
+# Hive Client memory usage can be an issue if a large number of clients
+# are running at the same time. The flags below have been useful in
+# reducing memory usage:
+
+# if [ "$SERVICE" = "beeline" ]; then
+#   if [ -z "$DEBUG" ]; then
+#     export HADOOP_OPTS="$HADOOP_OPTS -XX:NewRatio=12 -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:+G1GC -XX:-UseGCOverheadLimit"
+#   else
+#     export HADOOP_OPTS="$HADOOP_OPTS -XX:NewRatio=12 -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:-UseGCOverheadLimit"
+#   fi
+# fi
+
+# The heap size of the jvm stared by hive shell script can be controlled via:
+#
+# export HADOOP_HEAPSIZE=1024
+#
+# Larger heap size may be required when running queries over large number of files or partitions.
+# By default hive shell scripts use a heap size of 256 (MB).  Larger heap size would also be
+# appropriate for hive server.
+
+# if [ "$SERVICE" = "metastore" ]; then
+#   export HADOOP_HEAPSIZE=1024
+#   export HADOOP_OPTS="$HADOOP_OPTS -Dhive.log.dir=$HIVE_LOG_DIR -Dhive.log.file=hive$SERVICE.log \
+#     -Dlog4j.configurationFile=$HIVE_CONF_DIR/hive-log4j2.properties -XX:+UseG1GC \
+#     -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=$HIVE_LOG_DIR \
+#     -Xlog:gc*=info:file=$HIVE_LOG_DIR/gc-${SERVICE}-%p-%t.log:time,uptime,level,tags:filecount=5,filesize=16M"
+# fi
+#
+# if [ "$SERVICE" = "hiveserver2" ]; then
+#   export HADOOP_HEAPSIZE=1024
+#   export HADOOP_OPTS="$HADOOP_OPTS -Dhive.log.dir=$HIVE_LOG_DIR \
+#     -Dhive.log.file=$SERVICE.log -Dlog4j.configurationFile=$HIVE_CONF_DIR/hive-log4j2.properties \
+#     -XX:+UseG1GC -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=$HIVE_LOG_DIR \
+#     -Xlog:gc*=info:file=$HIVE_LOG_DIR/gc-${SERVICE}-%p-%t.log:time,uptime,level,tags:filecount=5,filesize=16M"
+# fi
--- a/conf/hive-env.sh.template
+++ b/conf/hive-env.sh.template
@@ -30,7 +30,7 @@
 # Folder containing extra libraries required for hive compilation/execution can be controlled by:
 # export HIVE_AUX_JARS_PATH=
 
-# HIVE Log Directory can be controlled by and ensure it exists on the filesystem:
+# HIVE Log Directory can be controlled by the following and ensure it exists on the filesystem:
 # export HIVE_LOG_DIR=
 
 # The hive service being invoked (beeline etc.) is available via the environment
@@ -42,9 +42,9 @@
 
 # if [ "$SERVICE" = "beeline" ]; then
 #   if [ -z "$DEBUG" ]; then
-#     export HADOOP_OPTS="$HADOOP_OPTS -XX:NewRatio=12 -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:+G1GC -XX:-UseGCOverheadLimit"
+#     export HADOOP_OPTS="$HADOOP_OPTS -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:+UseG1GC -XX:-UseGCOverheadLimit"
 #   else
-#     export HADOOP_OPTS="$HADOOP_OPTS -XX:NewRatio=12 -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:-UseGCOverheadLimit"
+#     export HADOOP_OPTS="$HADOOP_OPTS -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:-UseGCOverheadLimit"
 #   fi
 # fi
 
--- a/conf/hive-env.sh.template
+++ b/conf/hive-env.sh.template
@@ -42,9 +42,9 @@
 
 # if [ "$SERVICE" = "beeline" ]; then
 #   if [ -z "$DEBUG" ]; then
-#     export HADOOP_OPTS="$HADOOP_OPTS -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:+UseG1GC -XX:-UseGCOverheadLimit"
+#     export HADOOP_CLIENT_OPTS="$HADOOP_CLIENT_OPTS -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:+UseG1GC -XX:-UseGCOverheadLimit"
 #   else
-#     export HADOOP_OPTS="$HADOOP_OPTS -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:-UseGCOverheadLimit"
+#     export HADOOP_CLIENT_OPTS="$HADOOP_CLIENT_OPTS -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:-UseGCOverheadLimit"
 #   fi
 # fi
 
@@ -58,7 +58,7 @@
 
 # if [ "$SERVICE" = "metastore" ]; then
 #   export HADOOP_HEAPSIZE=1024
-#   export HADOOP_OPTS="$HADOOP_OPTS -Dhive.log.dir=$HIVE_LOG_DIR -Dhive.log.file=hive$SERVICE.log \
+#   export HADOOP_CLIENT_OPTS="$HADOOP_CLIENT_OPTS -Dhive.log.dir=$HIVE_LOG_DIR -Dhive.log.file=hive$SERVICE.log \
 #     -Dlog4j.configurationFile=$HIVE_CONF_DIR/hive-log4j2.properties -XX:+UseG1GC \
 #     -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=$HIVE_LOG_DIR \
 #     -Xlog:gc*=info:file=$HIVE_LOG_DIR/gc-${SERVICE}-%p-%t.log:time,uptime,level,tags:filecount=5,filesize=16M"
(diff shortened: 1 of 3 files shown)


Aggarwal-Raghav:
For example:
Screenshot 2025-06-21 at 1 45 54 PM
I can easily tell that PID 14784 belongs to HS2 heap dump and can easily fetch gc logs for that PID easily.

Aggarwal-Raghav:
@ayushtkn @deniskuzZ , I would like to get you opinion on this small improvement patch. I have done this in our internal hive distribution and has helped a lot.

Aggarwal-Raghav:
HMS log will look like
```
2025-06-21 13:00:24,689 INFO metastore.HiveMetaStore: Starting hive metastore on port 9083. PID is 7951
```

HS2 log will look like
```
2025-06-21 13:01:04,646 INFO server.HiveServer2: Starting HiveServer2. PID is: 8083

```

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -1206,7 +1206,8 @@ public void startPrivilegeSynchronizer(HiveConf hiveConf) throws Exception {
   private static void startHiveServer2() throws Throwable {
     long attempts = 0, maxAttempts = 1;
     while (true) {
-      LOG.info("Starting HiveServer2");
+      long pid = ProcessHandle.current().pid();
```
deniskuzZ: Do we need a separate var for pid extraction?

Aggarwal-Raghav: We can inline it in LOG message in both HS2 and HMS, if you think that's best.

conf/hive-env.sh.template
```diff
@@ -18,18 +20,29 @@
...
-# are running at the same time. The flags below have been useful in 
+# are running at the same time. The flags below have been useful in
 # reducing memory usage:
-#
-# if [ "$SERVICE" = "cli" ]; then
+
+# if [ "$SERVICE" = "beeline" ]; then
 #   if [ -z "$DEBUG" ]; then
-#     export HADOOP_OPTS="$HADOOP_OPTS -XX:NewRatio=12 -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:+UseParNewGC -XX:-UseGCOverheadLimit"
+#     export HADOOP_OPTS="$HADOOP_OPTS -XX:NewRatio=12 -Xms10m -XX:MaxHeapFreeRatio=40 -XX:MinHeapFreeRatio=15 -XX:+G1GC -XX:-UseGCOverheadLimit"
```
okumin: Can we remove `-XX:+G1GC` since it's the default with JDK 17+? I also see [-XX:NewRatio is not recommended](https://docs.oracle.com/en/java/javase/21/gctuning/garbage-first-garbage-collector-tuning.html#GUID-0BB3B742-A985-4D5E-A9C5-433A127FE0F6). However, as I don't know the intention of the DEBUG option, we may keep it.

okumin: `-XX:+UseG1GC` is the correct one?

Aggarwal-Raghav: Thanks for the review @okumin. Few points:
1. Yes -`XX:+UseG1GC` is the correct one. I doubled confirmed using:
```java -XX:+UnlockExperimentalVMOptions -XX:+PrintFlagsFinal -version | grep -i "G1GC"```
2. Regarding removing the GC flags and NewRatio, I am inclined towards removing every flag other than Xmx and Xms. Reason being, G1GC and ZGC doesn't require much tuning and tuning should be system specific + workload specifc which varies cluster to cluster.
3. I "want to believe" that the DEBUG option is there to showcase user the possibilities / kind of tuning users can do.

Let me know your thoughts based on that I will update the PR.

File: service/src/java/org/apache/hive/service/server/HiveServer2.java (hunk 1 above)
Comment (deniskuzZ) [reply service/src/java/org/apache/hive/service/server/HiveServer2.java]: not sure, I am just used to inline when the var is not used in multiple places

File: conf/hive-env.sh.template (hunk 2 above)
Comment (deniskuzZ) [reply conf/hive-env.sh.template]: sorry, just to confirm, `-XX:+G1GC` was replaced with `XX:+UseG1GC` ? if yes, we still have `-XX:+G1GC` in  beeline HADOOP_OPTS and need to replace

Comment (Aggarwal-Raghav) [reply conf/hive-env.sh.template]: Addressed in ba2c9caf135d1449ff11d061a6dbe7a5ead1fb47

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java [hunk 3]
```diff
@@ -262,7 +262,11 @@ public static void main(String[] args) throws Throwable {
     startupShutdownMessage(HiveMetaStore.class, args, LOG);
 
     try {
-      String msg = "Starting hive metastore on port " + cli.getPort();
+      String msg =
```
deniskuzZ, line 265: i think it's better to use logger template "{}"
why do we need to add this into `System.err`?

Aggarwal-Raghav: > why do we need to add this into System.err?

when HMS starts with verbose mode i.e. `bin/hive --service metastore --verbose`, the verbose info/logs are printed via `System.err`. It was done in [HIVE-2139](https://issues.apache.org/jira/browse/HIVE-2139). After doing some research (googling) found this:

`
In Java, System.err is typically used for error messages and verbose output because it's generally not buffered, unlike System.out, which is often buffered. This means System.err sends output directly to the console or error stream, ensuring that error messages are displayed immediately, even if System.out's buffer is not yet full. This behavior is designed to prioritize the display of error information, even if it means potentially interrupting or appearing before the buffered output from System.out. 
`

Also, "Starting hive metastore" is printed 3 times (attaching screenshot for the same)
Screenshot 2025-06-25 at 11 31 54 PM
Screenshot 2025-06-25 at 11 37 17 PM

> i think it's better to use logger template "{}"

As its not a LOG line but something like this can be done:
```
 String msg =
          "Starting hive metastore on port %d. PID is %d"
              .formatted(cli.getPort(), ProcessHandle.current().pid());
```

deniskuzZ: There is no verbose mode for HS2?

deniskuzZ (approved):
+1

conf/hive-env.sh.template
```diff
@@ -18,37 +20,54 @@
...
-# HADOOP_HOME=${bin}/../../hadoop
-
-# Hive Configuration Directory can be controlled by:
-# export HIVE_CONF_DIR=
-
-# Folder containing extra libraries required for hive compilation/execution can be controlled by:
-# export HIVE_AUX_JARS_PATH=
+# if [ "$SERVICE" = "metastore" ]; then
+#   export HADOOP_HEAPSIZE=1024
+#   export HADOOP_OPTS="$HADOOP_OPTS -Dhive.log.dir=$HIVE_LOG_DIR -Dhive.log.file=hive$SERVICE.log \
```
zhangbutao: @Aggarwal-Raghav Have you did the test using the latest master code? 
I found that `HADOOP_OPTS` cannot make the Hive log directory take effect. It was only after I changed to `HADOOP_CLIENT_OPTS` that it worked. However, it was indeed possible to take effect by using `HADOOP_OPTS` in the past.

image
The screenshot shows that I deployed Hive 4.1.0 locally. I had to modify `HADOOP_CLIENT_OPTS` to make the log directory effective.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStore.java
Aggarwal-Raghav: @deniskuzZ, there is no verbose mode for HS2.

conf/hive-env.sh.template
Aggarwal-Raghav: @zhangbutao , thanks for looking into this and **yes, you are correct. Even in my setup I have `HADOOP_CLIENT_OPTS`  for quite some time:**
https://github.com/Aggarwal-Raghav/hive-mac-setup/blob/main/hive/hive-env.sh

But reason why I went ahead with `HADOOP_OPTS`:
1. In my prod clusters, we have HADOOP_OPTS set in hive-env.sh in Amabri UI and it's working as expected.
2. Even in Ambari opensource, `HADOOP_OPTS` is used:
https://github.com/apache/ambari/blob/trunk/ambari-server/src/main/resources/stacks/BIGTOP/3.2.0/services/HIVE/configuration/hive-env.xml#L376
3. Because of point [1] and [2], I stick to HADOOP_OPTS which are **also present in older hive-env.sh.template** in hive,

Let me know what are your thoughts on this? Should I change to `HADOOP_CLIENT_OPTS` in `hive-env.sh.template`?

zhangbutao: I just want to confirm whether HADOOP_OPTS will make the Hive directory configuration take effect. Because it didn't seem to work in my test environment using latest branch, then I switched to using HADOOP_CLIENT_OPTS.

Aggarwal-Raghav: No, I also can't make HADOOP_OPTS work.

Aggarwal-Raghav:
Moving from `HADOOP_OPTS` => `HADOOP_CLIENT_OPTS` in hive-env.sh.template.
Reason:
1. JDK17 --add-opens changes were made in HADOOP_CLIENT_OPTS and same is used in Jenkinsfile as well.

@deniskuzZ @zhangbutao , can you please review once again?

conf/hive-env.sh.template
zhangbutao: Sorry, I'd like to confirm again to avoid any misunderstanding due to my unclear description. 

When I was testing the latest&branch-4.1 code, I found that the **hive log** parameters configured with `HADOOP_OPTS` (**Dhive.log.dir=$HIVE_LOG_DIR -Dhive.log.file=hive$SERVICE.log**) were not taking effect, and I didn't see any log file in the specified `hive.log.dir`. Only when I replaced `HADOOP_OPTS `with `HADOOP_CLIENT_OPTS` did this log directory take effect. 

But I'm not sure where the problem lies that causes this situation, **because before this, I was able to make the hive.log.dir parameter take effect by using HADOOP_OPTS.** 

So, if your current test results are the same as mine, it might be a problem caused by changes in the Hive code. If this issue is real, it would be best to fix it on the Hive side. Otherwise, similar downstream components like **Ambari** might have to modify their code to adapt to the new `HADOOP_CLIENT_OPTS` parameter, which could be quite troublesome.

zhangbutao:
> Moving from `HADOOP_OPTS` => `HADOOP_CLIENT_OPTS` in hive-env.sh.template. Reason:
> 
> 1. JDK17 --add-opens changes were made in HADOOP_CLIENT_OPTS and same is used in Jenkinsfile as well.
> 
> @deniskuzZ @zhangbutao , can you please review once again?

If this is the issue, could we add `--add-opens` to the HADOOP_OPTS setting to fix this problem?
https://github.com/apache/hive/blob/925df927248c1be11b84df3e8a0c3e43955e133e/bin/ext/hiveserver2.sh#L53-L54

cc @tanishq-chugh @ayushtkn

zhangbutao:
> Moving from `HADOOP_OPTS` => `HADOOP_CLIENT_OPTS` in hive-env.sh.template. Reason:
> 
> 1. JDK17 --add-opens changes were made in HADOOP_CLIENT_OPTS and same is used in Jenkinsfile as well.
> 
> @deniskuzZ @zhangbutao , can you please review once again?

```
diff --git a/bin/ext/hiveserver2.sh b/bin/ext/hiveserver2.sh
index 3431f7de17..c45031b982 100644
--- a/bin/ext/hiveserver2.sh
+++ b/bin/ext/hiveserver2.sh
@@ -51,7 +51,7 @@ hiveserver2() {
     killAndWait $pid $timeout
   else
     export HADOOP_CLIENT_OPTS=" -Dproc_hiveserver2 --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED  --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.util.concurrent=ALL-UNNAMED --add-opens java.base/java.util.concurrent.atomic=ALL-UNNAMED --add-opens java.base/java.util.regex=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED $HADOOP_CLIENT_OPTS "
-    export HADOOP_OPTS="$HIVESERVER2_HADOOP_OPTS $HADOOP_OPTS"
+    export HADOOP_OPTS="$HIVESERVER2_HADOOP_OPTS --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED  --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.util.concurrent=ALL-UNNAMED --add-opens java.base/java.util.concurrent.atomic=ALL-UNNAMED --add-opens java.base/java.util.regex=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED $HADOOP_OPTS"
     commands=$(exec $HADOOP jar $JAR $CLASS -H | grep -v '-hiveconf' | awk '{print $1}')
     start_hiveserver2='Y'
     for i in "$@"; do
diff --git a/bin/ext/metastore.sh b/bin/ext/metastore.sh
index 251013b2bb..8b646d29e8 100644
--- a/bin/ext/metastore.sh
+++ b/bin/ext/metastore.sh
@@ -28,7 +28,8 @@ metastore() {
   # Append --add-opens args that is required for JDK-17
   export HADOOP_CLIENT_OPTS=" -Dproc_metastore --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED  --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.util.concurrent=ALL-UNNAMED --add-opens java.base/java.util.concurrent.atomic=ALL-UNNAMED --add-opens=java.base/java.util.regex=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED $HADOOP_CLIENT_OPTS "

-  export HADOOP_OPTS="$HIVE_METASTORE_HADOOP_OPTS $HADOOP_OPTS"
+  export HADOOP_OPTS="$HIVE_METASTORE_HADOOP_OPTS --add-opens java.base/java.nio=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED  --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.util.concurrent=ALL-UNNAMED --add-opens java.base/java.util.concurrent.atomic=ALL-UNNAMED --add-opens=java.base/java.util.regex=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED $HADOOP_OPTS"
   exec $HADOOP jar $JAR $CLASS "$@"
 }
```

I  did this change, and using `HADOOP_OPTS` in hive-env.sh to specify `hive.log.dir`, but it still didn't work.

Aggarwal-Raghav:
@zhangbutao , the patch that you have sent above is not what I meant regarding add-opens. Let me rephrase, It seems HADOOP_OPTS are not picked up for starting hive services no matter where you set them i.e. `bin/hive.sh`, `bin/hive-config.sh`,`bin/ext/hiveserver2.sh`, etc or atleast I couldn't make them work. Confirmed from `ps -ef | grep -i  hiveserver2` output.

In HIVE-26473, all the add-opens were appended in HADOOP_CLIENT_OPTS, that's why I changed from HADOOP_OPTS => HADOOP_CLIENT_OPTS in `hive-env.sh.template` file.

zhangbutao:
> @zhangbutao , the patch that you have sent above is not what I meant regarding add-opens. Let me rephrase, It seems HADOOP_OPTS are not picked up for starting hive services no matter where you set them i.e. `bin/hive.sh`, `bin/hive-config.sh`,`bin/ext/hiveserver2.sh`, etc or atleast I couldn't make them work. Confirmed from `ps -ef | grep -i hiveserver2` output.
> 
> In [HIVE-26473](https://issues.apache.org/jira/browse/HIVE-26473), all the add-opens were appended in HADOOP_CLIENT_OPTS, that's why I changed from HADOOP_OPTS => HADOOP_CLIENT_OPTS in `hive-env.sh.template` file.

Yes, I understand what you mean. 
I just don't understand why `HADOOP_OPTS` no longer takes effect after HIVE-26473 was merged. 
For me, using `HADOOP_CLIENT_OPTS` at present is acceptable, but I think, in the future, there might be users who are confused about the issue of `HADOOP_OPTS` being ineffective. 

Maybe we can address this issue later.

Aggarwal-Raghav:
> I just don't understand why HADOOP_OPTS no longer takes effect after [HIVE-26473](https://issues.apache.org/jira/browse/HIVE-26473) was merged.

Just to clarify, i think it was not working way before HIVE-26473.

zhangbutao:
> > I just don't understand why HADOOP_OPTS no longer takes effect after [HIVE-26473](https://issues.apache.org/jira/browse/HIVE-26473) was merged.
> 
> Just to clarify, i think it was not working way before [HIVE-26473](https://issues.apache.org/jira/browse/HIVE-26473).

OK. It might also be a issue caused by other PR. Because previously, when I conducted local tests, using `HAODOOP_OPTS` was always effective.

zhangbutao:
@Aggarwal-Raghav I have just figured out the reason why the `HADOOP_OPTS` in my **hive-env.sh** file is ineffective.

Because in order for `Hadoop 3.4.1` to run on JDK 17, I configured the `HADOOP_OPTS` parameter in the **hadoop-env.sh** file:
`export HADOOP_OPTS="--add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/sun.net.util=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED"`

So when Hive starts, the value of the `HADOOP_OPTS` parameter in **hive-env.sh** is always overwritten by the parameter value in **hadoop-env.sh**.

Therefore, in order to prevent the parameters in **hive-env.sh** from being overridden by those in **hadoop-env.sh**, we can specify a clean **hadoop-env.sh** file for hive (a separate hadoop client for hive)


Back to this current PR, regarding whether we should change the `HADOOP_OPTS `in the **hive-env.sh template** to `HADOOP_CLIENT_OPTS`. My answer is yes.
See HIVE-19886
https://github.com/apache/hive/blob/925df927248c1be11b84df3e8a0c3e43955e133e/service/src/java/org/apache/hive/service/server/HiveServer2.java#L1469-L1480

But I also want to hear the opinions of the other folks. :)

Thanks.

zhangbutao (approved):
+1 LGTM

zhangbutao:
@okumin @deniskuzZ Thanks for the reivew!

PR #5887, opened 2025-06-21
[WIP] HIVE-29019: Test Hive Iceberg REST Catalog with the compatibility kit

okumin:
### What changes were proposed in this pull request?

Test the Iceberg REST API with CatalogTests with RESTCatalog.

https://issues.apache.org/jira/browse/HIVE-29019

### Why are the changes needed?

For better test coverage.

### Does this PR introduce _any_ user-facing change?

No. This feature has not been shipped.

### How was this patch tested?

```
mvn clean test -pl metastore-rest-catalog
```

PR #5888, opened 2025-06-23
HIVE-29036: Support ViewCatalog through Iceberg REST API

okumin:

Add the view support with integration tests.
https://issues.apache.org/jira/browse/HIVE-29036

With the current configuration, the Iceberg client ignores view-related endpoints.

No. REST Catalog has not been released yet.

```
mvn clean test -pl metastore-rest-catalog
```


standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
```diff
@@ -110,19 +112,18 @@ public class HMSCatalogAdapter implements RESTClient {
...
 
   public HMSCatalogAdapter(Catalog catalog) {
+    Preconditions.checkArgument(catalog instanceof SupportsNamespaces);
+    Preconditions.checkArgument(catalog instanceof ViewCatalog);
     this.catalog = catalog;
-    this.asNamespaceCatalog =
-        catalog instanceof SupportsNamespaces ? (SupportsNamespaces) catalog : null;
-    this.asViewCatalog = catalog instanceof ViewCatalog ? (ViewCatalog) catalog : null;
+    this.asNamespaceCatalog = (SupportsNamespaces) catalog;
+    this.asViewCatalog = (ViewCatalog) catalog;
```
okumin, line 126: I believe our adapter does not need to be so generic

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
```diff
@@ -133,82 +134,54 @@ enum HTTPMethod {
...
-    LOAD_VIEW(HTTPMethod.GET, VIEWS_PATH,
-            null, LoadViewResponse.class),
-    CREATE_VIEW(HTTPMethod.POST, "v1/namespaces/{namespace}/views",
-            CreateViewRequest.class, LoadViewResponse.class),
-    UPDATE_VIEW(HTTPMethod.POST, VIEWS_PATH,
-            UpdateTableRequest.class, LoadViewResponse.class),
-    RENAME_VIEW(HTTPMethod.POST, "v1/views/rename",
-            RenameTableRequest.class, null),
-    DROP_VIEW(HTTPMethod.DELETE, VIEWS_PATH);
+    TOKENS(HTTPMethod.POST, "v1/oauth/tokens", null),
```
okumin, line 130: I copied & pasted 1.9.1 of the upstream and removed unused fields. When our `/v1/config` returns the list of endpoints, the original form, such as `v1/namespaces`, doesn't work. We need to be adaptable to the new form such as `/v1/{prefix}/namespaces`.
https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/ResourcePaths.java

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
```diff
@@ -271,29 +241,10 @@ static String hmsCatalogMetricCount(String route) {
...
-      metricNames.add(hmsCatalogMetricCount(route.name()));
-    }
-    return metricNames;
-  }
-
   private ConfigResponse config() {
-    return castResponse(ConfigResponse.class, ConfigResponse.builder().build());
+    final List<Endpoint> endpoints = Arrays.stream(Route.values())
+        .map(r -> Endpoint.create(r.method.name(), r.resourcePath)).collect(Collectors.toList());
+    return castResponse(ConfigResponse.class, ConfigResponse.builder().withEndpoints(endpoints).build());
```
okumin, line 240: This change is one of the key points of this PR. The Iceberg client doesn't support views unless we explicitly declare that we support view endpoints.
https://github.com/apache/hive/pull/5887#issuecomment-2994195117

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
deniskuzZ: what is the role of {`prefix}` - catalogName aka `iceberg`?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
```diff
@@ -110,19 +111,18 @@ public class HMSCatalogAdapter implements RESTClient {
...
+    Preconditions.checkArgument(catalog instanceof ViewCatalog);
     this.catalog = catalog;
-    this.asNamespaceCatalog =
-        catalog instanceof SupportsNamespaces ? (SupportsNamespaces) catalog : null;
-    this.asViewCatalog = catalog instanceof ViewCatalog ? (ViewCatalog) catalog : null;
+    this.asNamespaceCatalog = (SupportsNamespaces) catalog;
+    this.asViewCatalog = (ViewCatalog) catalog;
   }
 
   enum HTTPMethod {
```
deniskuzZ: @okumin, could you please drop this enum and import `org.apache.iceberg.rest.HTTPRequest.HTTPMethod`?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
okumin: It is an optional path parameter, which we can freely use.
https://github.com/apache/iceberg/pull/5233

S3 Tables use it to put a bucket identifier.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-integrating-open-source.html#endpoint-parameter

Glue.
https://docs.aws.amazon.com/glue/latest/dg/connect-glu-iceberg-rest.html#prefix-catalog-path-parameters

I guess multi-tenant services utilize it, but I've not used it so far.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
okumin: nice advice

PR #5889, opened 2025-06-23
HIVE-28607: Iceberg: Syntax sugar for Iceberg Branching.

ayushtkn:

Add some syntax sugar for Iceberg Branching & Tag feature

Better Usability

Yes, the User can use the new commands introduced for Branching & Tagging

UT

### DIFF
--- a/iceberg/iceberg-handler/src/test/queries/positive/iceberg_branches_and_tags.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/iceberg_branches_and_tags.q
@@ -11,8 +11,11 @@ insert into default.ice01.branch_branch1 values (5), (6);
 select * from default.ice01.branch_branch1;
 
 -- replace branch
-create or replace branch test_branch_1 FROM ice01;;
+create or replace branch test_branch_1 FROM ice01;
 select * from default.ice01.branch_branch1;
+
+-- create branch if not exist
+create branch if not exists branch1 FROM ice01;
 
 -- create tag
 create tag tag1 FROM ice01;
@@ -22,6 +25,9 @@ select * from default.ice01.tag_tag1;
 delete from ice01 where id=2;
 create or replace tag tag1 FROM ice01;
 select * from default.ice01.tag_tag1;
+
+-- create tag if not exists
+create tag if not exists tag1 FROM ice01;
 
 -- drop branch
 drop branch branch1 FROM ice01;
--- a/parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
+++ b/parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
@@ -649,7 +649,7 @@ alterStatementSuffixDropTag
 createTagStatement
 @init { gParent.pushMsg("alter table drop tag (if exists) tagName", state); }
 @after { gParent.popMsg(state); }
-    : KW_CREATE KW_TAG ifNotExists? tagName=identifier snapshotIdOfRef? refRetain? KW_FROM tableName
+    : KW_CREATE KW_TAG ifNotExists? tagName=identifier KW_FROM tableName snapshotIdOfRef? refRetain?
     -> ^(TOK_ALTERTABLE tableName
      ^(TOK_ALTERTABLE_CREATE_TAG $tagName ifNotExists? snapshotIdOfRef? refRetain?)
      )
--- a/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergBranchOperation.java
+++ b/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/TestHiveIcebergBranchOperation.java
@@ -72,15 +72,20 @@ public class TestHiveIcebergBranchOperation extends HiveIcebergStorageHandlerWit
             fileFormat, HiveIcebergStorageHandlerTestUtils.CUSTOMER_RECORDS, 2);
 
     String branchName = "test_branch_1";
+    String branchName2 = "test_branch_2";
     Long snapshotId = table.history().get(0).snapshotId();
     shell.executeStatement(String.format("ALTER TABLE customers CREATE BRANCH %s FOR SYSTEM_VERSION AS OF %d",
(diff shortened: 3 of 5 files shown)


parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
```diff
@@ -558,13 +558,42 @@ alterStatementSuffixReplaceTag
     -> ^(TOK_ALTERTABLE_REPLACE_SNAPSHOTREF KW_TAG $sourceBranch $snapshotId refRetain?)
     ;
 
+dropBranchStatement
+@init { gParent.pushMsg("alter table drop branch (if exists) branchName", state); }
+@after { gParent.popMsg(state); }
+    : KW_DROP KW_BRANCH ifExists? branchName=identifier KW_FROM tableName
+    -> ^(TOK_ALTERTABLE tableName
+     ^(TOK_ALTERTABLE_DROP_BRANCH ifExists? $branchName)
+     )
```
okumin, line 567: I heard some catalogs support DB or catalog-level branching and that's why Trino uses `IN TABLE|SCHEMA|CATALOG identifier` instead of `FROM identifier`. Just a tip, though Hive might not support it.

ayushtkn: Sounds cool but I don't think Hive is anyway near supporting that. But just casually thinking it might not be a naive task to implement. need to see how Trino did it.

Do you think having From isn't good, or are you ok with the current syntax proposed

okumin: IMO, FROM is the best name to refer to a table name. Trino probably reserves it to add `CREATE BRANCH x IN TABLE tbl FROM other_branch_name`. For me, it is more straightforward and consistent to support it as `CREATE BRANCH x FROM cat.db.tbl.branch_x`. No objection.

https://docs.google.com/document/d/1jEF4IkWu-2Gzk5ii2Nb0exuEnAUeo98UbiM3i0xtgWQ/edit?tab=t.0#heading=h.dglxb51zw9m2

iceberg/iceberg-handler/src/test/queries/positive/iceberg_branches_and_tags.q
```diff
@@ -0,0 +1,30 @@
...
+
+--- create branch
+create branch branch1 FROM ice01;
+
+-- insert some values in branch1
+insert into default.ice01.branch_branch1 values (5), (6);
+select * from default.ice01.branch_branch1;
+
+-- replace branch
+create or replace branch test_branch_1 FROM ice01;;
```
difin: NIT: extra semicolon

parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
```diff
@@ -558,13 +558,42 @@ alterStatementSuffixReplaceTag
...
     -> ^(TOK_ALTERTABLE_DROP_BRANCH ifExists? $branchName)
     ;
 
+createBranchStatement
+@init { gParent.pushMsg("create branch statement", state); }
+@after { gParent.popMsg(state); }
+    : KW_CREATE KW_BRANCH ifNotExists? branchName=identifier KW_FROM tableName
+      snapshotIdOfRef? refRetain? retentionOfSnapshots?
+    -> ^(TOK_ALTERTABLE tableName
+          ^(TOK_ALTERTABLE_CREATE_BRANCH $branchName ifNotExists? snapshotIdOfRef? refRetain? retentionOfSnapshots?)
```
difin, line 583: This part of the syntax seem to be not covered by the new q-test that you introduced: `ifNotExists? snapshotIdOfRef? refRetain? retentionOfSnapshots?`, should they be tested? This can be tested by junit tests like `TestParseOptimizeTable` not necessary q-test.

okumin (approved):
I personally agree. I would like one more +1, as it relies on the preference.

zhangbutao:
It looks great. The "branch/tag" syntax is a bit more concise compared to the previous one. 

But I'm wondering how we should provide friendly SQL reference documentation for users. We have added a lot of semantics related to "iceberg", but our conf/doc doesn't seem to have a good record of these additions. Many of the new SQL semantics currently exist only in the code and SQL test scripts, , which is not very user-friendly for Hive users.

I hope that in the future we can continue to improve the Hive documentation. Let's work hard! :)

BTW, iceberg has some doc about hive-iceberg branch, https://iceberg.apache.org/docs/nightly/hive/#branches-and-tags. But from a long-term perspective, our hive also should have better documentation for this.

ayushtkn:
Thanx everyone for the reviews, I will try to address them in few days.

@zhangbutao I know Hive documentation isn't very cool & I was trying to improve last time via wiki:
https://cwiki.apache.org/confluence/display/Hive/Apache+Hive+4.0.X

I did something along with few friends in buisness, but It wasn't complete, it has Iceberg section it does have a mention of branches & tags.

I do have a blog where I mentioned some syntax around branches & tags (maybe I will revisit that after this), but nothing official

https://medium.com/@ayushtkn/apache-hive-4-x-with-iceberg-branches-tags-3d52293ac0bf

Will require volunteers to make the documentation great :-)

parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
ayushtkn: I have added more tests as you suggested

difin: LGTM

slfan1989:
@ayushtkn I attended the Asia 2025 conference yesterday and joined the session on Data Lake & Data Warehouse. During the talk titled The Future of ETL with Branching & Tagging in Apache Hive, I noticed a reference to a PR you had submitted.

It was great to see your contribution highlighted! It would be even better if you could personally present in a future session.

Here's the session link for your reference:
https://asia.communityovercode.org/sessions/datalake-914197.html

PR #5890, opened 2025-06-24
HIVE-29030: Alter partition change column cannot use the direct sql

dengzhhu653:

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
@@ -47,7 +47,6 @@ import org.slf4j.LoggerFactory;
 import org.apache.hadoop.conf.Configuration;
 
 import javax.jdo.PersistenceManager;
-import javax.jdo.Transaction;
 import javax.jdo.datastore.JDOConnection;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -473,10 +472,8 @@ class DirectSqlUpdatePart {
                                                       List<TransactionalMetaStoreEventListener> transactionalListeners)
           throws MetaException {
 
-    Transaction tx = pm.currentTransaction();
     try {
       dbType.lockInternal();
-      tx.begin();
       JDOConnection jdoConn = null;
       Map<String, Map<String, String>> result;
       try {
@@ -519,16 +516,12 @@ class DirectSqlUpdatePart {
       } finally {
         closeDbConn(jdoConn);
       }
-      tx.commit();
       return result;
     } catch (Exception e) {
       LOG.error("Unable to update Column stats for  " + tbl.getTableName(), e);
       throw new MetaException("Unable to update Column stats for  " + tbl.getTableName()
               + " due to: "  + e.getMessage());
     } finally {
-      if (tx.isActive()) {
-        tx.rollback();
-      }
       dbType.unlockInternal();
     }
   }
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
@@ -42,10 +42,6 @@ import org.apache.hadoop.hive.metastore.model.MPartitionColumnStatistics;
 import org.apache.hadoop.hive.metastore.model.MStringList;
 import org.apache.hadoop.hive.metastore.tools.SQLGenerator;
 import org.apache.hadoop.hive.metastore.txn.TxnUtils;
-import org.datanucleus.ExecutionContext;
-import org.datanucleus.api.jdo.JDOPersistenceManager;
-import org.datanucleus.metadata.AbstractClassMetaData;
-import org.datanucleus.metadata.IdentityType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.conf.Configuration;
@@ -76,6 +72,7 @@ import static org.apache.hadoop.hive.metastore.MetastoreDirectSqlUtils.executeWi
 import static org.apache.hadoop.hive.metastore.MetastoreDirectSqlUtils.extractSqlClob;
 import static org.apache.hadoop.hive.metastore.MetastoreDirectSqlUtils.extractSqlInt;
 import static org.apache.hadoop.hive.metastore.MetastoreDirectSqlUtils.extractSqlLong;
+import static org.apache.hadoop.hive.metastore.MetastoreDirectSqlUtils.getModelIdentity;
 
 /**
  * This class contains the optimizations for MetaStore that rely on direct SQL access to
@@ -213,7 +210,6 @@ class DirectSqlUpdatePart {
   }
 
   private void insertIntoPartColStatTable(Map<PartColNameInfo, MPartitionColumnStatistics> insertMap,
-                                          long maxCsId,
                                           Connection dbConn) throws SQLException, MetaException, NoSuchObjectException {
     int numRows = 0;
     String insert = "INSERT INTO \"PART_COL_STATS\" (\"CS_ID\", \"COLUMN_NAME\", \"COLUMN_TYPE\", \"PART_ID\","
@@ -228,7 +224,7 @@ class DirectSqlUpdatePart {
         Long partId = partColNameInfo.partitionId;
         MPartitionColumnStatistics mPartitionColumnStatistics = (MPartitionColumnStatistics) entry.getValue();
 
-        preparedStatement.setLong(1, maxCsId);
+        preparedStatement.setLong(1, getModelIdentity(pm, MPartitionColumnStatistics.class));
         preparedStatement.setString(2, mPartitionColumnStatistics.getColName());
         preparedStatement.setString(3, mPartitionColumnStatistics.getColType());
         preparedStatement.setLong(4, partId);
@@ -249,7 +245,6 @@ class DirectSqlUpdatePart {
         preparedStatement.setLong(19, mPartitionColumnStatistics.getLastAnalyzed());
         preparedStatement.setString(20, mPartitionColumnStatistics.getEngine());
 
-        maxCsId++;
         numRows++;
         preparedStatement.addBatch();
         if (numRows == maxBatchSize) {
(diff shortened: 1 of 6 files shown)


Aggarwal-Raghav:
@dengzhhu653, I will be marking [HIVE-29039](https://issues.apache.org/jira/browse/HIVE-29039) as duplicate of HIVE-29030.
As I have a repro case on my setup, let me test this PR :-)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -473,15 +467,13 @@ private void setAnsiQuotes(Connection dbConn) throws SQLException {
    * @return map of partition key to column stats if successful, null otherwise.
    */
   public Map<String, Map<String, String>> updatePartitionColumnStatistics(Map<String, ColumnStatistics> partColStatsMap,
-                                                      Table tbl, long csId,
+                                                      Table tbl,
                                                       String validWriteIds, long writeId,
                                                       List<TransactionalMetaStoreEventListener> transactionalListeners)
           throws MetaException {
 
-    Transaction tx = pm.currentTransaction();
```
deniskuzZ, line 481: is it safe to remove?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
```diff
@@ -1254,19 +1159,17 @@ public List<Void> run(List<Long> input) throws Exception {
...
-          newCdIdToCols.put(newCdId, newCols);
-          oldCdIdToNewCdId.put(cdId, newCdId);
-          sdIdToNewCdId.put(sdId, newCdId);
-          for (int i = 0; i < oldCols.size(); i++) {
-            FieldSchema oldCol = oldCols.get(i);
-            int newIdx = newCols.indexOf(oldCol);
-            if (newIdx != -1) {
-              oldCdIdToColIdxPairs.computeIfAbsent(cdId, k -> new ArrayList<>()).add(Pair.of(i, newIdx));
-            }
+      if (!oldCols.equals(newCols) && newCols != null) {
```
deniskuzZ, line 1162: is that safe, won't we get NullPointer when oldCols == null?

Aggarwal-Raghav:
@dengzhhu653 , i ran with this patch. Few points:
1. Direct SQL is working fine, No "falling back to ORM" log is observed 👍🏻.
2. There is performance degradation for the same scenario that we ran in HIVE-28972 (alter table add column cascade), instead of taking 70-80 sec, it is taking  approx. 550 sec. I suspect the issue is moving from alterPartition => alterPartitions in HIVE-28956. Because without HIVE-28956, it is taking 80 sec approx. 

Can you also validate please, in order to rule out setup issues?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
dengzhhu653: It's safe. I move the transaction open to ObjectStore, this method will be executed under the transaction:
https://github.com/apache/hive/pull/5890/files#diff-ddc1cac0afb55312b1d426288dbb7c32611ea996d2224f1692f3a53134fd9a80R9699

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java
dengzhhu653: no, the `oldCols` is created here: 
https://github.com/apache/hive/blob/83d5fdd5a196754d38e5c0d92d1833a28af05608/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/DirectSqlUpdatePart.java#L1157

dengzhhu653:
> @dengzhhu653 , i ran with this patch. Few points:
> 
> 1. Direct SQL is working fine, No "falling back to ORM" log is observed 👍🏻.
> 2. There is performance degradation for the same scenario that we ran in [HIVE-28972](https://issues.apache.org/jira/browse/HIVE-28972) (alter table add column cascade), instead of taking 70-80 sec, it is taking  approx. 550 sec. I suspect the issue is moving from alterPartition => alterPartitions in [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956). Because without [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956), it is taking 80 sec approx.
> 
> Can you also validate please, in order to rule out setup issues?

Thank you. Let me try

deniskuzZ:
@Aggarwal-Raghav, do we know what change in [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) introduced a perf degradation: batching or `alterPartitions` API?
cc [vikramahuja1001](https://github.com/vikramahuja1001)

dengzhhu653:
@Aggarwal-Raghav same to me, it takes me 402.786s. I guess it due to insert the new columns per partition, instead of sharing the same CD_ID the table holds.

@deniskuzZ  @Aggarwal-Raghav  should we file a new jira to address this issue?

dengzhhu653:
> @Aggarwal-Raghav, do we know what change in [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) introduced a perf degradation: batching or `alterPartitions` API? cc [vikramahuja1001](https://github.com/vikramahuja1001)

The alterPartitions API doesn't share the CD_ID the table holds, for each partition, it creates a new CD_ID with 800+ columns in COLUMNS_V2

Aggarwal-Raghav:
> @Aggarwal-Raghav same to me, it takes me 402.786s. I guess it due to insert the new columns per partition, instead of sharing the same CD_ID the table holds.
> 
> @deniskuzZ @Aggarwal-Raghav should we file a new jira to address this issue?

@dengzhhu653 , as code freeze is planned for tomorrow i.e. June 25th, Can we make it in time? I am inclined towards reverting [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) and chase it later in master branch? What do you think @deniskuzZ ?

deniskuzZ:
> > @Aggarwal-Raghav same to me, it takes me 402.786s. I guess it due to insert the new columns per partition, instead of sharing the same CD_ID the table holds.
> > @deniskuzZ @Aggarwal-Raghav should we file a new jira to address this issue?
> 
> @dengzhhu653 , as code freeze is planned for tomorrow i.e. June 25th, Can we make it in time? I am inclined towards reverting [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) and chase it later in master branch? What do you think @deniskuzZ ?

+1

deniskuzZ:
> > @Aggarwal-Raghav, do we know what change in [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) introduced a perf degradation: batching or `alterPartitions` API? cc [vikramahuja1001](https://github.com/vikramahuja1001)
> 
> The alterPartitions API doesn't share the CD_ID the table holds, for each partition, it creates a new CD_ID with 800+ columns in COLUMNS_V2

@dengzhhu653, but performance-wise that shouldn't be any different from using loop over partitions and calling `alterPartition` API, isn't it? if we manage to add `CD_ID` reuse, that should improve the performance.

dengzhhu653:
> > > @Aggarwal-Raghav, do we know what change in [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) introduced a perf degradation: batching or `alterPartitions` API? cc [vikramahuja1001](https://github.com/vikramahuja1001)
> > 
> > 
> > The alterPartitions API doesn't share the CD_ID the table holds, for each partition, it creates a new CD_ID with 800+ columns in COLUMNS_V2
> 
> @dengzhhu653, but performance-wise that shouldn't be any different from using loop over partitions and calling `alterPartition` API, isn't it? if we manage to add `CD_ID` reuse, that should improve the performance.

In theory yes, the performance should be similar(or better than) as before we upgrade the DataNucleus.

deniskuzZ:
@Aggarwal-Raghav, @dengzhhu653, do we have an agreement on [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) revert? If yes, I can do that.

Also, do we have a benchmark in metastore-tools for `alter table add column cascade`?  Or how do you find the regression post DataNucleus upgrade, private setup ?

dengzhhu653:
Personally I'm -1 on reverting the already merged PR without strong reason, it's upset for the community. Filed HIVE-29042, I can prepare a fix next day if allow.

Aggarwal-Raghav:
> @Aggarwal-Raghav, @dengzhhu653, do we have an agreement on [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) revert? If yes, I can do that.
> 
> Also, do we have a benchmark in metastore-tools for `alter table add column cascade`? Or how do you find the regression post DataNucleus upgrade, private setup ?

======
> how do you find the regression post DataNucleus upgrade, private setup ?

@deniskuzZ, I found it because of some prod use case and checked the behaviour in master branch and found the performace issue.

vikramahuja1001:
@Aggarwal-Raghav , thanks for testing out the patch HIVE-28956 with along with HIVE-28909 and HIVE-28972. When i was working on the fix and tested the same, the 2 patches were not merged.
From the above tests, post HIVE-28972 it seems that the JDO performance is much better than the current DirectSql Implementation. 



> @Aggarwal-Raghav, do we know what change in [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) introduced a perf degradation: batching or `alterPartitions` API? cc [vikramahuja1001](https://github.com/vikramahuja1001)

@deniskuzZ , in HIVE-28956, i had moved from alterPartition API to alterPartitions API with batching. As is the case currently, alterPartition does not support DirectSql, the degradation is caused by the API change itself and is not related to batching.





> @Aggarwal-Raghav, @dengzhhu653, do we have an agreement on [HIVE-28956](https://issues.apache.org/jira/browse/HIVE-28956) revert? If yes, I can do that.
> 
> Also, do we have a benchmark in metastore-tools for `alter table add column cascade`? Or how do you find the regression post DataNucleus upgrade, private setup ?

Though reverting the patch HIVE-28956 would definitely improve the performance of alter table add column cascade operation, but it would still not fix the issue of DirectSql being slower than JDO in the alterPartitions API which currently is being used in other flows as well, one of the instance being: alter table rename operation. 
This use case of rename table should also potentially suffer from the same issue, based on the analysis in the above comments. HIVE-17969 is doing the similar change of moving from alterPartition API to alterPartitions API in the alter table rename flow. We should also perhaps look and compare the performance of the rename operation as well in JDO vs DirectSql and then take a call.

deniskuzZ:
fixed the directSql perf in https://github.com/apache/hive/pull/5895

PR #5891, opened 2025-06-24
HIVE-29040: Add UUIDV7 built in function

BsoBird:

support uuid_v7 function
https://issues.apache.org/jira/browse/HIVE-29040

No.

See  TestUDFUUID.class

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDv7.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDv7.java
@@ -34,12 +34,12 @@ import org.apache.hadoop.io.Text;
 @Description(name = "uuid_v7",
         value = "_FUNC_() - Returns a universally unique identifier (UUID_V7) string.",
         extended = """
-                The value is returned as a canonical UUID 36-character string.
+                The value is returned as a canonical UUIDv7 36-character string.
                 Example:
                   > SELECT _FUNC_();
-                  '0baf1f52-53df-487f-8292-99a03716b688'
+                  '0197b082-4dc8-770e-9a75-bd601dff26a1'
                   > SELECT _FUNC_();
-                  '36718a53-84f5-45d6-8796-4f79983ad49d'""")
+                  '0197b082-bc45-745f-adb7-43df40797b0b'""")
 @UDFType(deterministic = false)
 public class UDFUUIDv7 extends GenericUDF {
   private final Text result = new Text();
--- a/ql/src/test/queries/clientpositive/udf_uuid_v7.q
+++ b/ql/src/test/queries/clientpositive/udf_uuid_v7.q
@@ -1,0 +1,4 @@
+DESCRIBE FUNCTION uuid_v7;
+DESCRIBE FUNCTION EXTENDED uuid_v7;
+
+select uuid_v7();


BsoBird:
@zhangbutao hello,Sir.
Can you help check this?tks.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java
```diff
@@ -0,0 +1,71 @@
...
+        value = "_FUNC_() - Returns a universally unique identifier (UUID_V7) string.",
+        extended = """
+                The value is returned as a canonical UUID 36-character string.
+                Example:
+                  > SELECT _FUNC_();
+                  '0baf1f52-53df-487f-8292-99a03716b688'
+                  > SELECT _FUNC_();
+                  '36718a53-84f5-45d6-8796-4f79983ad49d'""")
+@UDFType(deterministic = false)
+public class UDFUUIDV7 extends UDF {
```
deniskuzZ: could we lowercase the version, `UDFUUIDv7`?

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * @return Text
+   */
+  public Text evaluate() {
+    result.set(randomUUIDV7().toString());
+    return result;
+  }
+
+
+  private UUID randomUUIDV7() {
```
deniskuzZ: @BsoBird, where did you get the UUIDv7 impl? 

- https://bugs.openjdk.org/browse/JDK-8357251, https://github.com/openjdk/jdk/pull/25303
- https://antonz.org/uuidv7/#java 
- import https://github.com/f4b6a3/uuid-creator mvn dependency

````
UUID uuid = UuidCreator.getTimeOrderedEpoch();
````

File: ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java (hunk 1 above)
Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: sure

File: ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java (hunk 2 above)
Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @deniskuzZ https://github.com/robsonkades/uuidv7
This is a fast method for calculating UUIDV7.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: https://github.com/f4b6a3/uuid-creator seems to have higher rating score

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @deniskuzZ Okay,let me see.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @deniskuzZ I reviewed the code and conducted some simple tests. Currently, I have selected a UUIDv7 generation algorithm with relatively the best performance (approximately generating 18 million UUIDs per second in a single thread). Of course, if we need to switch to a different implementation library, I have no objections. What do you think? After all, the UUIDv7 algorithm isn't very complex, so perhaps there's no need to introduce a complicated library?After all, the performance of using SecureRandom + ByteBuffer to generate UUIDV7 is quite poor.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @BsoBird how about the java spec proposal: https://bugs.openjdk.org/browse/JDK-8357251? could you please check?

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: > @BsoBird how about the java spec proposal: https://bugs.openjdk.org/browse/JDK-8357251? could you please check?

sure.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @deniskuzZ JDK-8357251 also uses SecureRandom, so the efficiency of generating UUIDV7 should not be high.

tips: Perhaps we should suggest to the JDK official to refer to the algorithm provided at https://github.com/robsonkades/uuidv7.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @liach, why the thumbs down? Do you recommend the [JDK-8357251](https://bugs.openjdk.org/browse/JDK-8357251) implementation?

Comment (liach) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: Oh, I left the reaction based on the JDK perspective. Java aims to ensure security of its APIs, so I don't think Java will ever use pseudorandom generators that can be easily cracked. (I had wondered if SecureRandom could be modified to pass the random numbers as long/int on the stack instead of creating heap byte arrays, but guess that might not be the main overhead of SecureRandom)

I don't know what the functional requirements Apache Hive has; if Hive doesn't have concern about attackers reverse engineering the generated random numbers, I think Hive can go ahead.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @liach 
I'm sorry, sir, I apologize for underestimating the JVM's considerations for security. I now understand that there are valid reasons for using SecureRandom in the JVM. I just found that the efficiency of generating UUIDs with SecureRandom is really poor. UUIDv4 can still maintain a relatively decent production rate, while the current efficiency of UUIDv7 is far lower than that of UUIDv4. For a large data warehouse, using a UUIDv7 function with excessively low production efficiency may not meet the requirements at all. Therefore, I have chosen a relatively high-performance algorithm (although it does not fully comply with cryptographic security). If the performance of the JVM's built-in implementation is significantly improved in the future, I would prefer to use the JVM's built-in UUIDv7 algorithm.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]:
```
-- Meters ----------------------------------------------------------------------
use-security-random
             count = 58635
         mean rate = 5329.31 events/second
     1-minute rate = 5168.78 events/second
     5-minute rate = 5145.16 events/second
    15-minute rate = 5141.06 events/second
```

```
-- Meters ----------------------------------------------------------------------
use-threadlocal-random
             count = 146583184
         mean rate = 18320875.98 events/second
     1-minute rate = 18035808.20 events/second
     5-minute rate = 18035808.20 events/second
    15-minute rate = 18035808.20 events/second
```

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @liach @deniskuzZ 
Sir, this is the efficiency of UUIDV7 generation I tested in my local environment. The reason I don't use SecureRandom is simply because it's too slow. There's about a 3400-fold efficiency gap between the two algorithms. How should we make the trade-off? 
For example, we provide both fast algorithms and secure algorithms?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: Since Hive is primarily designed for ETL rather than BI, I think we should prioritize security over efficiency
cc @okumin, @ayushtkn, @zabetak WDYT?

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @deniskuzZ But sir, a production rate of five thousand per second is really too slow. Additionally, the UUIDs generated by pseudo-randomness can at most be predicted, but what impact could that possibly have? HIVE's UDFs themselves extensively use pseudo-random functions. For example, the rand() function.

In the vast majority of database/data warehouse cases, uuid_v7 is only used as a unique key for records. In this scenario, even if the UUID is predicted, it doesn't seem to have much impact.

Assuming we generate 20 billion uuid_v7 ids, with 300 running in parallel, it would take a full 3.7 hours to complete. However, using the original uuid (uuidv4) function only takes 0.009 hours. This gap is simply too large. I can hardly imagine how the inefficient UUID_V7 could be effectively used in a production environment.

Assuming the existing secure version of uuid_v7 can achieve a production rate of around 100,000 per second, I would definitely choose the secure version. Because at this point, the impact on efficiency is acceptable.

If we absolutely need a secure uuid function, how about we add parameters to the uuid_v7 function, similar to pg, to control the generation behavior of uuid_v7 by passing parameters?

Comment (zabetak) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: We just have to make sure that the implementation matches the specification. If the UUIDv7 specification does not impose any requirements about security then I don't have a strong preference of which one to use.

Apart from that, if the implementation is based on another open-source project then I would rather add the respective dependency in Hive rather than copy-pasting code. That way licensing implications are clear and we don't maintain the same code in multiple places.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java]: @zabetak Sir, the main reason I didn't add the Maven dependency is that the UUID_V7 calculation code is only about ten lines long. Isn't it too heavy to introduce a Maven dependency just for these ten lines of code?
If we must introduce a Maven dependency, then I would prefer to add the Maven dependency from https://github.com/f4b6a3/uuid-creator.
However, JDK is currently adding built-in support for UUIDV7. The dependency we add now might need to be removed later. So wouldn't it be better to not add any dependency for now and wait until the time is ripe to replace it with JDK's built-in UUIDV7 implementation?

File: ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDv7.java [hunk 3]
```diff
@@ -0,0 +1,71 @@
...
+        value = "_FUNC_() - Returns a universally unique identifier (UUID_V7) string.",
+        extended = """
+                The value is returned as a canonical UUID 36-character string.
+                Example:
+                  > SELECT _FUNC_();
+                  '0baf1f52-53df-487f-8292-99a03716b688'
+                  > SELECT _FUNC_();
+                  '36718a53-84f5-45d6-8796-4f79983ad49d'""")
+@UDFType(deterministic = false)
+public class UDFUUIDv7 extends UDF {
```
okumin: I'd like to use GenericUDF for new one because UDF is deprecated.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java
okumin: UUID users expect every generated ID to be unique. The random generator needs to be as secure as we can guarantee the uniqueness in my opinion.
As for the dependency, using mvn would make it easier to maintain licenses.

zabetak: It really depends if we are talking about adding a new dependency to a huge library or if we are adding a dependency with a single Java class. Copying code is not a good practice and has various disadvantages. 

If for instance you copy the code in Hive then we cannot take advantage of improvements/bug fixes/security patches that will land in the dependent library. Nobody, will care to bring these changes to Hive and it may be too late when we realize that somethings is needed.

Moreover, when you copy code from open-source projects and you have to follow the ASF policies (i.e., https://www.apache.org/legal/src-headers.html#3party) that I don't think you did in this case. This has legal implications that make our life more difficult.

Dependencies come and go so its fine to add/remove things when it makes sense to do so.

BsoBird: @zabetak @deniskuzZ @okumin 
I replaced the UUIDv7 generation code with the uuid-creator dependency introduced via Maven.
Currently, I am still using the high-performance code to generate UUIDv7: `Guid.v7().toString().`
If we need to enforce the use of cryptographically secure UUIDv7, please let me know.


```
-- Meters ----------------------------------------------------------------------
uuid-creator:use-threadlocal-random
count = 165457936
mean rate = 18360701.04 events/second
1-minute rate = 17878586.40 events/second
5-minute rate = 17878586.40 events/second
15-minute rate = 17878586.40 events/second
```

BsoBird: > UUID users expect every generated ID to be unique. The random generator needs to be as secure as we can guarantee the uniqueness in my opinion. As for the dependency, using mvn would make it easier to maintain licenses.

@okumin If the generation speed of UUIDV7 using security-random could be a bit faster (for example, 100,000 pieces of data per second), I wouldn't mind using it. The biggest problem now is that its generation speed is too slow.

>In the vast majority of database/data warehouse cases, uuid_v7 is only used as a unique key for records. In this scenario, even if the UUID is predicted, it doesn't seem to have much impact.

>Assuming we generate 20 billion uuid_v7 ids, with 300 running in parallel, it would take a full 3.7 hours to complete. However, using the original uuid (uuidv4) function only takes 0.009 hours. This gap is simply too large. I can hardly imagine how the inefficient UUID_V7 could be effectively used in a production environment.

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
```diff
@@ -23,26 +23,41 @@
...
-    
     assertEquals(id1.length(), 36);
     assertEquals(id2.length(), 36);
-
     GenericUDFBridge bridge = new GenericUDFBridge("uuid", false, UDFUUID.class.getName());
     assertFalse(FunctionRegistry.isDeterministic(bridge));
   }
+
+  @Test
+  public void testUUIDV7() throws Exception {
```
deniskuzZ: minor, lowercase `v7`

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDv7.java
```diff
@@ -0,0 +1,65 @@
...
+                  '0baf1f52-53df-487f-8292-99a03716b688'
+                  > SELECT _FUNC_();
+                  '36718a53-84f5-45d6-8796-4f79983ad49d'""")
+@UDFType(deterministic = false)
+public class UDFUUIDv7 extends GenericUDF {
+  private final Text result = new Text();
+
+  @Override
+  public ObjectInspector initialize(ObjectInspector[] arguments) throws UDFArgumentException {
+    if(arguments!=null){
```
deniskuzZ: please format

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDv7.java
deniskuzZ: +1

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
BsoBird: done

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java
deniskuzZ: `uuid-creator` dependency should be good and project has also multiple stars

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDv7.java
BsoBird: Tks! Done.

deniskuzZ (approved):
+1

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
```diff
@@ -23,26 +23,41 @@
...
-
     GenericUDFBridge bridge = new GenericUDFBridge("uuid", false, UDFUUID.class.getName());
     assertFalse(FunctionRegistry.isDeterministic(bridge));
   }
+
+  @Test
+  public void testUUIDv7() throws Exception {
+    UDFUUIDv7 udf = new UDFUUIDv7();
+    String id1 = udf.evaluate(null).toString();
+    TimeUnit.MILLISECONDS.sleep(1);
```
deniskuzZ: @BsoBird why do you need sleep? please see https://sonarcloud.io/project/issues?impactSeverities=MEDIUM&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=5891&id=apache_hive

BsoBird: @deniskuzZ Because UUIDV7 is incrementing over time, I want to prove that the UUID generated after 1 millisecond must have a larger lexicographical order than the previous UUID.

BsoBird: @deniskuzZ There is no such thing as asynchronously waiting for the previous task to complete before executing the next task here. I just want to delay the time by 1 millisecond before generating the second UUIDV7 and ensure that the latter UUID is definitely greater than the former one. If this condition is not met, then the UUID we generated must not comply with the UUIDV7 specification.
```
String v7_uuid_id1 = udf.evaluate(null).toString();
TimeUnit.MILLISECONDS.sleep(1);
String v7_uuid_id2 = udf.evaluate(null).toString();
assertTrue(v7_uuid_id2.compareTo(v7_uuid_id1)>0);  //id2 must be greater than ID1
```
If we don't verify the time-sequential characteristic of UUIDV7 in unit tests, that seems too strange.

zabetak: I am not even sure we need such kind of tests in this repo. This seems to be testing the correctness of the underlying implementation in `com.github.f4b6a3` so such tests should be contributed there if they don't exist already.

deniskuzZ: +1 on the above point

BsoBird: Removed the test code for verifying UUIDV7 increment

BsoBird: @deniskuzZ @zabetak 
Sir, I have replaced the test cases and removed Thread.sleep(). However, I still personally believe that some basic test cases should be added. Not adding any tests feels very strange. Based on consistent experience, we can assume that code without test cases simply does not work properly. Therefore, not adding test cases makes me feel very uncomfortable. (Maybe I have OCD.)😝

deniskuzZ: @BsoBird, I think we need a functional test that validates the actual UDF invocation end-to-end. 
I've only found 1 q tests (ppd2.q) that uses the `uuid()`. 
Can we extend one of the existing `udf*.q` tests or create a new one?

zabetak: An end-to-end test in a .q file is the way to go. 

Given that we depend on an external library I would also suggest to run a smoke test by starting a Hive docker container (with these changes), set `tez.local.mode=false`, and execute a simple query making use of the new function. It's quite common to see `ClassNotFoundException` when new libraries are added in the ql module.

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
```diff
@@ -23,26 +23,44 @@
...
+    String id1 = udf.evaluate(null).toString();
+    long timestampAfter = System.currentTimeMillis();
+    UUID v7id1 = UUID.fromString(id1);
+    assertEquals(7, v7id1.version());
+    long v7IdTimeStamp = v7id1.getMostSignificantBits() >>> 16;
+    assertTrue(v7IdTimeStamp <= timestampAfter && timestampBefore <= v7IdTimeStamp);
+    String id2 = udf.evaluate(null).toString();
+    assertNotEquals(id1, id2);
+    assertEquals(36, id1.length());
+    assertEquals(36, id2.length());
```
zabetak: This tests the underlying UUID implementation so we don't really need it in the Hive repo. Please remove it.

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
```diff
@@ -23,26 +23,44 @@
...
+    GenericUDFBridge bridge = new GenericUDFBridge("uuid_v7", false, UDFUUIDv7.class.getName());
+    assertFalse(FunctionRegistry.isDeterministic(bridge));
```
zabetak: The test for the determinism of the function should be moved to `TestFunctionRegistry`.

ql/src/test/queries/clientpositive/ppd2.q
```diff
@@ -104,7 +104,7 @@ CREATE
  UNION ALL
      SELECT
         dt_from as DT_FROM,
-        uuid() as DT_TO
+        uuid_v7() as DT_TO
```
zabetak: The purpose of the `ppd2.q` test is to test predicate pushdown not the addition of new UDFs. Please restore the changes here and create a new file (i.e., udf_uuidv7.q) and put your test there. Note that we want to test the actual execution of the function not the plan (EXPLAIN).

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
BsoBird: Okay

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
BsoBird: Okay

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
```diff
@@ -33,12 +33,12 @@ public class TestUDFUUID {
   @Test
   public void testUUID() throws Exception {
     UDFUUID udf = new UDFUUID();
-
```
zabetak, line 36: Please remove all unnecessary white-space changes. Such changes pollute the git history and complicate merges and backports.

ql/src/test/queries/clientpositive/ppd2.q
```diff
@@ -34,9 +34,9 @@ from (
 where b.cc>1;
 
 EXPLAIN
-SELECT user_id
```
zabetak, line 37: Idem for whitespace changes in this file.

ql/src/test/queries/clientpositive/udf_uuid_v7.q
```diff
@@ -0,0 +1,4 @@
+DESCRIBE FUNCTION uuid_v7;
+DESCRIBE FUNCTION EXTENDED uuid_v7;
+
+select uuid_v7();
```
zabetak, line 4: There is no corresponding file .q.out for this change which means two things:
* you didn't run the test at all locally before pushing your changes;
* the whole CI run will run for nothing and will come back red.

ql/src/test/org/apache/hadoop/hive/ql/udf/TestUDFUUID.java
zabetak: @BsoBird Did you try the smoke test mentioned above? The PR should not be merged unless this test is done.

ql/src/test/queries/clientpositive/udf_uuid_v7.q
BsoBird: @zabetak Thank you for your correction, sir. I haven't fully processed this PR yet, I just temporarily submitted the code. Because I have multiple personal computers.(I need to get the commit change history on another computer) Maybe I should first change this PR to Draft status.

zabetak: It's better to avoid submitting changes when the PR has known issues for two reasons:
* They trigger notifications that goes to multiple places; reviewers believe that your work is ready for review and thus jump in to finish up the work.
* CI starts running and takes precedence over PRs that are in a better shape

Changing it to draft is better but unfortunately it will still trigger a CI run and notifications. If you want to keep changes temporarily private push them in separate branch.

BsoBird: @zabetak True. Tks.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDv7.java
```diff
@@ -0,0 +1,65 @@
...
+                  '0197b082-bc45-745f-adb7-43df40797b0b'""")
+@UDFType(deterministic = false)
+public class UDFUUIDv7 extends GenericUDF {
+  private final Text result = new Text();
+
+  @Override
+  public ObjectInspector initialize(ObjectInspector[] arguments) throws UDFArgumentException {
+    if(arguments != null) {
+      checkArgsSize(arguments, 0, 0);
+    }
```
okumin, line 51: We can remove the null check because `arguments` is non-nullable.

ql/src/java/org/apache/hadoop/hive/ql/udf/UDFUUIDV7.java
okumin: I understand that UUID v7 is too slow and has very limited use cases. I also wonder why UUID v4 with SecureRandom is so fast but UUID v7 is not. Just curious

sergeyprokhorenko:
I recommend to use the PostgreSQL 18 [implementation](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=78c5e141e9c13) of uuidv7() function as a reference

deniskuzZ:
@BsoBird, would you be able to try the recommendation above?

BsoBird:
> @BsoBird, would you be able to try the recommendation above?

@deniskuzZ Sorry, sir, I have been too busy with work recently and don't have time to handle this PR for now. Once I finish what's at hand, I will come back to deal with this issue.

PR #5894, opened 2025-06-24
HIVE-29043: Ban the Jupiter dependency from hive-unit to prevent unit tests from being skipped

abstractdog:
Ban jupiter dependency.

To prevent skipping hive-unit unit tests silently again.

No.

Tested with adding the problematic dependency, and instead of silently skipping unit test, it failed.
```
mvn install -Pitests -Dtest=TestHttpServices -pl ./itests/hive-unit -nsu

...

[WARNING] Rule 0: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
Found Banned Dependency: org.junit.jupiter:junit-jupiter:jar:5.11.2
Use 'mvn dependency:tree' to locate the source of the banned dependencies.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  11.237 s
[INFO] Finished at: 2025-06-24T14:56:04+02:00
[INFO] ------------------------------------------------------------------------
[INFO] 5 goals, 5 executed
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce (ban-jupiter) on project hive-it-unit: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed. -> [Help 1]
[ERROR]
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1664,7 +1664,7 @@
             </configuration>
           </execution>
           <execution>
-            <id>enforce-banned-dependencies-licenses</id>
+            <id>enforce-banned-dependencies</id>
             <goals>
               <goal>enforce</goal>
             </goals>
@@ -1677,17 +1677,6 @@
                   </excludes>
                   <message>A banned license dependency was found!</message>
                 </bannedDependencies>
-              </rules>
-              <fail>true</fail>
-            </configuration>
-          </execution>
-          <execution>
-            <id>enforce-banned-dependencies-logging</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <configuration>
-              <rules>
                 <bannedDependencies>
                   <excludes>
                     <!-- Move to SLF4J -->
--- a/itests/hive-unit/pom.xml
+++ b/itests/hive-unit/pom.xml
@@ -549,6 +549,33 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>enforce-banned-dependencies</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <excludes>
+                    <exclude>org.junit.jupiter:junit-jupiter</exclude>
+                  </excludes>
+                  <searchTransitive>true</searchTransitive>
+                  <message>
+                    Banned jupiter dependency was found!
+                    Only remove this rule if you confirmed that adding jupiter dependency
+                    to the project won't make unit tests silently skipped.
+                  </message>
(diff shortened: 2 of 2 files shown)


aturoczy (approved):

InvisibleProgrammer (approved):

pom.xml
```diff
@@ -1747,6 +1747,24 @@
               </rules>
             </configuration>
           </execution>
+          <!-- Only remove this rule if you confirmed that adding jupiter dependency -->
+          <!-- to the project won't make unit tests silently skipped -->
```
zabetak: This comment would be more helpful if it was part of the failure. How about putting it in `<message>...</message>`

standalone-metastore/pom.xml
```diff
@@ -610,6 +610,47 @@
...
+                    <!--LGPL licenced library-->
+                    <exclude>com.google.code.findbugs:annotations</exclude>
+                  </excludes>
+                </bannedDependencies>
+              </rules>
+              <fail>true</fail>
+            </configuration>
+          </execution>
+          <!-- Only remove this rule if you confirmed that adding jupiter dependency -->
+          <!-- to the project won't make unit tests silently skipped -->
```
zabetak: Same comment regarding the message.

standalone-metastore/pom.xml
```diff
@@ -610,6 +610,47 @@
...
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <excludes>
+                    <!--LGPL licenced library-->
+                    <exclude>com.google.code.findbugs:annotations</exclude>
+                  </excludes>
+                </bannedDependencies>
+              </rules>
+              <fail>true</fail>
```
zabetak: I think the default is true so possible we could skip this.

standalone-metastore/pom.xml
```diff
@@ -610,6 +610,47 @@
...
+            </configuration>
+          </execution>
+          <!-- Only remove this rule if you confirmed that adding jupiter dependency -->
+          <!-- to the project won't make unit tests silently skipped -->
+          <execution>
+            <id>ban-jupiter</id>
```
zabetak: Why do we want to put the rule in a separate execution? We could merge `enforce-banned-dependencies` and `ban-jupiter` and it would lead to less boilerplate code.

abstractdog:
> I like the idea but based on the test failures it seems that we already have jupiter dependencies in the project so we have to revisit those.
> 
> Other than that left some minor suggestions for potential improvements.

thanks for the comments, finally, even if I wanted to ban the dependency globally so bad, I had to face that we do use the jupiter api extensively in some modules, so I ended banning it from itests/hive-unit, where it caused the problem all the time

there how I ended up with: https://github.com/apache/hive/pull/5894/commits/8e57d3b9f75bbfd8f59396ccb3d25090600370e7

zabetak (approved):
LGTM! Thanks Laszlo.

PR #5814, opened 2025-05-14
HIVE-28956 : Use msdb.alterPartitions() API and implement batching for alter table add column cascade command

vikramahuja1001:
### What changes were proposed in this pull request?
Currently Alter table add column cascade uses msdb.alterPartition API. This API only JDO based implementation. Instead of this use msdb.alterPartitions which supports both DirectSql and JDO Implementation and use batching as well.

### Why are the changes needed?
DirectSql bein more optimized than JDO can improve performance for tables with very high number of partitions and very high number of columns.

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
Tested this on a 3 node cluster on a table with 5000+ partitions and 900+ columns.

PR #5895, opened 2025-06-24
HIVE-29042: Performance degradation for add columns cascade

deniskuzZ:


Aggarwal-Raghav:
Thanks for the quick resolution @deniskuzZ. I'll also run the test scenario and update. Hope that's ok 😅

Aggarwal-Raghav:
It is taking approx **60 seconds** to run the scenario 🚀

deniskuzZ:
> It is taking approx **60 seconds** to run the scenario 🚀

Thank you for confirming, @Aggarwal-Raghav!

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java
```diff
@@ -420,35 +420,30 @@ public List<Void> run(List<Partition> input) throws Exception {
...
                     }
                   }
                   Deadline.checkTimeout();
-                  if (cascade) {
-                    msdb.alterPartitions(catalogName, databaseName, tableName, partVals, input, newt.getWriteId(),
-                            writeIdList);
-                  } else {
-                    msdb.alterPartitions(catalogName, newDbName, newTblName, partVals, oldParts, newt.getWriteId(),
-                            writeIdList);
-                  }
```
okumin, line 451: Can we assume `databaseName = newDbName` and `tableName = newTblName`?

deniskuzZ: that is how it was before the recent change: https://github.com/apache/hive/pull/5814/files#r2163346022

okumin: Definitely. I will approve the original one had been correct

okumin (approved):

PR #5897, opened 2025-06-25
HIVE-29046: Avoid unecessary perf overhead from handling deprecated properties in MetastoreConf

zabetak:
1. Move deprecated property registation in the existing static block
2. Drop constructors and variables related to deprecated keys in ConfVars
3. Drop DEPRECATED_TEST_ENTRY that is only useful for testing purposes and adapt existing tests

Improve performance & code maintenance.

Looping through ~400 ConfVars enumeration entries on every creation of a new Configuration entry is resource wasteful and unecessary. The deprecated properties are static and known at compile time so there is no reason to search and register them on every creation of a new Configuration instance. In addition, the presence of deprecation information inside the ConfVars enumeration leads to boilerplate code that is not really necessary.

No

Existing tests

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -128,6 +128,13 @@ public class MetastoreConf {
       keyToVars.put(var.varname, var);
       keyToVars.put(var.hiveName, var);
     }
+    Configuration.addDeprecations(new Configuration.DeprecationDelta[] {
+        new Configuration.DeprecationDelta("metastore.compactor.history.retention.attempted",
+            ConfVars.COMPACTOR_HISTORY_RETENTION_DID_NOT_INITIATE.varname),
+        new Configuration.DeprecationDelta("hive.compactor.history.retention.attempted",
+            ConfVars.COMPACTOR_HISTORY_RETENTION_DID_NOT_INITIATE.varname),
+        new Configuration.DeprecationDelta("datanucleus.plugin.pluginRegistryBundleCheck",
+            ConfVars.DATANUCLEUS_PLUGIN_REGISTRY_BUNDLE_CHECK.varname)});
   }
 
   @VisibleForTesting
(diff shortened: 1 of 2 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -128,6 +128,13 @@ public class MetastoreConf {
       keyToVars.put(var.varname, var);
       keyToVars.put(var.hiveName, var);
     }
+    Configuration.addDeprecations(new Configuration.DeprecationDelta[] {
```
deniskuzZ, line 131: can we just drop them in Hive-4.1?

zabetak: We don't have a policy in place about when and how to remove deprecated props so its our call. Personally, my goal here is to avoid the unnecessary perf hit so I don't mind removing them completely if you prefer this path.

Most often, when projects follow semantic versioning breaking changes only happen during major version bumps so ideally the removal should happen in 5.x. In Hive, we are not so dogmatic about it so I think we can afford such small breaking changes even in minor version bumps.

okumin: I didn't know this feature. Looks so nice.
https://github.com/apache/hadoop/blob/6eae1589aeea9bd9c6885e405bd9be5ef6199df7/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java#L555-L572

Personally, +1 with the current approach.

deniskuzZ: Ideally, we should have dropped those configs in 4.0 release. I don't think we should keep them in the next release.

zabetak: Since the work/effort is done and the tests are green there is no downside in merging this PR as is. I will create a follow-up to remove these props completely.

zabetak: Created follow-up: https://github.com/apache/hive/pull/5931

PR #5900, opened 2025-06-25
[TEST] Respect timezone passed to the reader

deniskuzZ:
TEST


ql/src/java/org/apache/hadoop/hive/ql/io/orc/RecordReaderImpl.java
```diff
@@ -458,14 +460,15 @@ static TimestampWritableV2 nextTimestamp(ColumnVector vector,
         result = (TimestampWritableV2) previous;
       }
       TimestampColumnVector tcv = (TimestampColumnVector) vector;
-      result.setInternal(tcv.time[row], tcv.nanos[row]);
+      tcv.setIsUTC(readerOptions.getUseUTCTimestamp());
```
deniskuzZ: temp workaround until https://github.com/apache/orc/pull/2300

PR #5901, opened 2025-06-26
HIVE-28790: ACID deletes are failing with ArrayIndexOutOfBoundsExcept…

kokila-19:
…ion when direct insert is enabled

If the length of outPathsCommitted array is shorter than the writeroffset, then extend the outPathsCommitted array to the length of writeroffset + 1

The problem is in the FileSinkOperator.createDynamicBucket method:
Acid deletes will fail with ArrayIndexOutOfBoundException when closing the writes, because the size of the array is not compared with the writerOffsetbecause and there is a possibility that the outPathsCommitted array is shorter than the updaters array.

No

qtest

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
@@ -501,7 +501,9 @@ public class FileSinkOperator extends TerminalOperator<FileSinkDesc> implements
 
       if (this.finalPaths[writerOffset] == null) {
         if (conf.isDirectInsert()) {
-          this.outPathsCommitted = Arrays.copyOf(outPathsCommitted, writerOffset + 1);
+          if (outPathsCommitted.length <= writerOffset) {
+            this.outPathsCommitted = Arrays.copyOf(outPathsCommitted, writerOffset + 1);
+          }
           this.finalPaths[writerOffset] = buildTmpPath();
           this.outPaths[writerOffset] = buildTmpPath();
         } else {
--- a/ql/src/test/queries/clientpositive/acid_direct_update_delete.q
+++ b/ql/src/test/queries/clientpositive/acid_direct_update_delete.q
@@ -18,4 +18,13 @@ DELETE FROM test_update_bucketed WHERE id IN ('2', '11', '10');
 UPDATE test_update_bucketed SET value='New value2' WHERE id IN ('2','18', '19');
 SELECT * FROM test_update_bucketed;
 
+CREATE TABLE test_delete(id int) STORED AS ORC TBLPROPERTIES('transactional'='true');
+
+INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id != 3 and id <=9;
+INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id > 9;
+
+DELETE FROM test_delete WHERE id in (5, 13);
+SELECT * FROM test_delete;
+
 DROP TABLE IF EXISTS test_update_bucketed;
+DROP TABLE IF EXISTS test_delete;
--- a/ql/src/test/results/clientpositive/llap/acid_direct_update_delete.q.out
+++ b/ql/src/test/results/clientpositive/llap/acid_direct_update_delete.q.out
@@ -136,6 +136,59 @@ POSTHOOK: Output: hdfs://### HDFS PATH ###
 7	seven
 8	eight
 9	nine
+PREHOOK: query: CREATE TABLE test_delete(id int) STORED AS ORC TBLPROPERTIES('transactional'='true')
+PREHOOK: type: CREATETABLE
+PREHOOK: Output: database:default
+PREHOOK: Output: default@test_delete
+POSTHOOK: query: CREATE TABLE test_delete(id int) STORED AS ORC TBLPROPERTIES('transactional'='true')
+POSTHOOK: type: CREATETABLE
+POSTHOOK: Output: database:default
+POSTHOOK: Output: default@test_delete
+PREHOOK: query: INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id != 3 and id <=9
+PREHOOK: type: QUERY
+PREHOOK: Input: default@test_update_bucketed
+PREHOOK: Output: default@test_delete
+POSTHOOK: query: INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id != 3 and id <=9
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@test_update_bucketed
+POSTHOOK: Output: default@test_delete
+POSTHOOK: Lineage: test_delete.id EXPRESSION [(test_update_bucketed)test_update_bucketed.FieldSchema(name:id, type:string, comment:null), ]
+PREHOOK: query: INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id > 9
+PREHOOK: type: QUERY
+PREHOOK: Input: default@test_update_bucketed
+PREHOOK: Output: default@test_delete
+POSTHOOK: query: INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id > 9
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@test_update_bucketed
+POSTHOOK: Output: default@test_delete
+POSTHOOK: Lineage: test_delete.id EXPRESSION [(test_update_bucketed)test_update_bucketed.FieldSchema(name:id, type:string, comment:null), ]
+PREHOOK: query: DELETE FROM test_delete WHERE id in (5, 13)
+PREHOOK: type: QUERY
+PREHOOK: Input: default@test_delete
+PREHOOK: Output: default@test_delete
+POSTHOOK: query: DELETE FROM test_delete WHERE id in (5, 13)
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@test_delete
+POSTHOOK: Output: default@test_delete
+PREHOOK: query: SELECT * FROM test_delete
+PREHOOK: type: QUERY
+PREHOOK: Input: default@test_delete
+PREHOOK: Output: hdfs://### HDFS PATH ###
+POSTHOOK: query: SELECT * FROM test_delete
+POSTHOOK: type: QUERY
+POSTHOOK: Input: default@test_delete
+POSTHOOK: Output: hdfs://### HDFS PATH ###
+1
+14
+16
+17
+18
+19
+20
+6
+7
+8
+9
 PREHOOK: query: DROP TABLE IF EXISTS test_update_bucketed
 PREHOOK: type: DROPTABLE
 PREHOOK: Input: default@test_update_bucketed
@@ -146,3 +199,13 @@ POSTHOOK: type: DROPTABLE
 POSTHOOK: Input: default@test_update_bucketed
 POSTHOOK: Output: database:default
 POSTHOOK: Output: default@test_update_bucketed
+PREHOOK: query: DROP TABLE IF EXISTS test_delete
+PREHOOK: type: DROPTABLE
+PREHOOK: Input: default@test_delete
+PREHOOK: Output: database:default
+PREHOOK: Output: default@test_delete
+POSTHOOK: query: DROP TABLE IF EXISTS test_delete
+POSTHOOK: type: DROPTABLE
+POSTHOOK: Input: default@test_delete
+POSTHOOK: Output: database:default
+POSTHOOK: Output: default@test_delete


deniskuzZ:
cc @zhangbutao

zhangbutao:
I'm following this PR.  :)

ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
```diff
@@ -497,11 +497,13 @@ public int createDynamicBucket(int bucketNum) {
         this.updaters = Arrays.copyOf(updaters, writerOffset + 1);
         this.outPaths = Arrays.copyOf(outPaths, writerOffset + 1);
         this.finalPaths = Arrays.copyOf(finalPaths, writerOffset + 1);
+        if (conf.isDirectInsert()) {
+          this.outPathsCommitted = Arrays.copyOf(outPathsCommitted, writerOffset + 1);
```
deniskuzZ: the current branch compares `updaters` length with `writerOffset`. is it ok to apply range copy to `outPathsCommitted` here?

deniskuzZ: based on Marta's comment in a JIRA:
````
But in the second part when the outPathsCommitted array is copied, the size of the array is not compared with the writerOffset. So it can happen that the outPathsCommitted array is reduced. If this situation happens it leads to the ArrayIndexOutOfBoundsException when closing the writes, because the outPathsCommitted array is shorter than the updaters array.
````
it seems you just need to add the `outPathsCommitted` size check:
````
if (this.finalPaths[writerOffset] == null) {
  if (conf.isDirectInsert()) {
----------------------------------------------------------------------------------
    if (outPathsCommitted.length <= writerOffset) {
      this.outPathsCommitted = Arrays.copyOf(outPathsCommitted, writerOffset + 1);
    }
----------------------------------------------------------------------------------
    this.finalPaths[writerOffset] = buildTmpPath();
    this.outPaths[writerOffset] = buildTmpPath();
  }
````
the new test passed with that changes

Comment (kokila-19) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java]: I did check that. Based on how the outPathsCommitted array is utilized , comparing updaters with writeroffset should also work fine.

```
for (int i = 0; i < updaters.length; i++) {
        if (updaters[i] != null) {
          SerDeStats stats = updaters[i].getStats();
          // Ignore 0 row files except in case of insert overwrite or delete or update
          if (isDirectInsert
              && (stats.getRowCount() > 0 || isInsertOverwrite || AcidUtils.Operation.DELETE.equals(acidOperation)
                  || AcidUtils.Operation.UPDATE.equals(acidOperation))) {

            OrcRecordUpdater recordUpdater = (OrcRecordUpdater) updaters[i];
            switch (acidOperation) {
            case INSERT:
              outPathsCommitted[i] = recordUpdater.getUpdatedFilePath();
              break;
            case UPDATE:
              // In case of update operation, we need both the deleteFilePath and the updatedFilePath.
              // The updateFilePath will be added to the outPathsCommitted array and the deleteFilePath
              // will be collected in a separate list.
              outPathsCommitted[i] = recordUpdater.getUpdatedFilePath();
              if (deleteDeltas != null) {
                deleteDeltas.add(recordUpdater.getDeleteFilePath());
                LOG.debug("The following path has been added to the deleteDeltas list: "
                    + recordUpdater.getDeleteFilePath().toString());
              }
              break;
            case DELETE:
              // In case of delete operation, the deleteFilePath has to be used, not the updatedFilePath
              outPathsCommitted[i] = recordUpdater.getDeleteFilePath();
              break;
            default:
              break;
            }
......
        }
```

outPathsCommitted is written based on updaters length , so i considered we can keep copying the array in common block.

Comment (kokila-19) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java]: The other solution also looks good to me and no need to have two conditional check for direct insert. Let me also test.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java]: It's not apparent from the code snippet above that outPathsCommitted.length is the same as updaters.length

File: ql/src/test/queries/clientpositive/acid_direct_delete.q [hunk 2]
```diff
@@ -0,0 +1,18 @@
...
+
+create external table ext(a int) stored as textfile;
+insert into table ext values(1),(2),(3),(4),(5),(6),(7), (8), (9), (12);
+create table full_acid(a int) stored as orc tblproperties("transactional"="true");
+
+insert into table full_acid select * from ext where a != 3 and a <=7 group by a;
+insert into table full_acid select * from ext where a>7 group by a;
+
+set mapreduce.job.reduces=1;
+delete from full_acid where a in (2, 12);
```
deniskuzZ: Why aren't you validating the table content post deletion ?

ql/src/test/queries/clientpositive/acid_direct_delete.q
```diff
@@ -0,0 +1,18 @@
...
+
+drop table if exists full_acid;
+
+set mapreduce.job.reduces=7;
+
+create external table ext(a int) stored as textfile;
+insert into table ext values(1),(2),(3),(4),(5),(6),(7), (8), (9), (12);
+create table full_acid(a int) stored as orc tblproperties("transactional"="true");
+
+insert into table full_acid select * from ext where a != 3 and a <=7 group by a;
```
deniskuzZ: why do you need `group by`?

ql/src/test/queries/clientpositive/acid_direct_delete.q
```diff
@@ -0,0 +1,18 @@
+set hive.mapred.mode=nonstrict;
```
deniskuzZ: why introduce a new q.test instead of extending the existing `acid_direct_update_delete.q`?

````
-- SORT_QUERY_RESULTS

set hive.acid.direct.insert.enabled=true;

DROP TABLE IF EXISTS test_update_bucketed;

CREATE TABLE test_update_bucketed(id string, value string) CLUSTERED BY(id) INTO 10 BUCKETS STORED AS ORC TBLPROPERTIES('transactional'='true');

INSERT INTO test_update_bucketed values ('1','one'),('2','two'),('3','three'),('4','four'),('5','five'),('6','six'),('7','seven'),('8','eight'),('9','nine'),('10','ten'),('11','eleven'),('12','twelve'),('13','thirteen'),('14','fourteen'),('15','fifteen'),('16','sixteen'),('17','seventeen'),('18','eighteen'),('19','nineteen'),('20','twenty');

SELECT * FROM test_update_bucketed;
DELETE FROM test_update_bucketed WHERE id IN ('2', '4', '12', '15');
UPDATE test_update_bucketed SET value='New value' WHERE id IN ('6', '11', '18', '20');
SELECT * FROM test_update_bucketed;
DELETE FROM test_update_bucketed WHERE id IN ('2', '11', '10');
UPDATE test_update_bucketed SET value='New value2' WHERE id IN ('2', '18', '19');
SELECT * FROM test_update_bucketed;

CREATE TABLE test_delete(id int) STORED AS ORC TBLPROPERTIES('transactional'='true');

INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id <= 9;
INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id > 9;

DELETE FROM test_delete WHERE id in (5, 13);
SELECT * FROM test_delete;

DROP TABLE test_update_bucketed;
DROP TABLE test_delete;
````

Comment (deniskuzZ) [inline ql/src/test/queries/clientpositive/acid_direct_delete.q]: this config is deprecated

File: ql/src/test/queries/clientpositive/acid_direct_delete.q (hunk 3 above)
Comment (kokila-19) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: Without `group by `, what we have is the following with the same bucket id. Deleting the rows with same bucketid will not trigger this issue .

```
{"writeid":1,"bucketid":536870912,"rowid":0}	1
{"writeid":1,"bucketid":536870912,"rowid":1}	2
{"writeid":1,"bucketid":536870912,"rowid":2}	4
{"writeid":1,"bucketid":536870912,"rowid":3}	5
{"writeid":1,"bucketid":536870912,"rowid":4}	6
{"writeid":1,"bucketid":536870912,"rowid":5}	7
{"writeid":2,"bucketid":536870912,"rowid":0}	8
{"writeid":2,"bucketid":536870912,"rowid":1}	9
{"writeid":2,"bucketid":536870912,"rowid":2}	12
```
To repro this issue, we need to delete two rows with different bucket id's where bucket num of 1st row  > 2nd row
With group by , we get 
```
{"writeid":1,"bucketid":536936448,"rowid":0}	1
{"writeid":1,"bucketid":537001984,"rowid":0}	7
{"writeid":1,"bucketid":537067520,"rowid":0}	4
{"writeid":1,"bucketid":537067520,"rowid":1}	6
{"writeid":1,"bucketid":537198592,"rowid":0}	5
{"writeid":1,"bucketid":537264128,"rowid":0}	2
{"writeid":2,"bucketid":536936448,"rowid":0}	9
{"writeid":2,"bucketid":537133056,"rowid":0}	12
{"writeid":2,"bucketid":537264128,"rowid":0}	8
```

where in the qtest , there is a deletion of 2(bucket id : 537264128) and 12 (bucketid : 537133056) where their bucket number is 6 and 4 respectively which will trigger this particular scenario where we get ArrayIndexOutOfBounds Exception for index 6 as outPathCommitted Array is shortened based on last bucket number which will be 4.

File: ql/src/test/queries/clientpositive/acid_direct_delete.q (hunk 4 above)
Comment (kokila-19) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: As this is a very specific use case, I wrote it in different q test.

Comment (deniskuzZ) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: it doesn't need a separate test, it's an overkill

File: ql/src/test/queries/clientpositive/acid_direct_delete.q (hunk 3 above)
Comment (deniskuzZ) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q:14]: there is nothing to group, `a` is already distinct. please check the complete test snippet I've shared here.

File: ql/src/test/queries/clientpositive/acid_direct_delete.q (hunk 4 above)
Comment (kokila-19) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: Agree on that. I will rewrite it.

File: ql/src/test/queries/clientpositive/acid_direct_delete.q (hunk 3 above)
Comment (kokila-19) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: I can repro the issue using the test snippet `acid_direct_update_delete.q ` .  I have rewritten and updated.
In the test steps provided in JIRA , this is not reproducible without group by though.

File: ql/src/test/queries/clientpositive/acid_direct_delete.q (hunk 4 above)
Comment (deniskuzZ) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: please check the complete `acid_direct_update_delete.q` test snippet I've shared here.

File: ql/src/test/queries/clientpositive/acid_direct_delete.q (hunk 3 above)
Comment (deniskuzZ) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: no need for group by when selecting from bucketed table. attached snippet is reproducing the issue, no need to change anything
````
CREATE TABLE test_delete(id int) STORED AS ORC TBLPROPERTIES('transactional'='true');

INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id <= 9;
INSERT INTO test_delete SELECT id FROM test_update_bucketed WHERE id > 9;

DELETE FROM test_delete WHERE id in (5, 13);
````

Comment (kokila-19) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: I was just wondering why without group by the new test case was not reproducing the issue. We are selecting it from a text file creating an unbucketed table and we needed to group by to distribute. 

And Yes, we don't need `group by ` in the `acid_direct_update_delete.q ` , so I have not added it. 

`I have rewritten and updated.`
I mentioned I have rewritten in the sense that I have updated the test case in the existing file and deleted the new one. I have not added group by as its not required here.

Comment (deniskuzZ) [reply ql/src/test/queries/clientpositive/acid_direct_delete.q]: test_update_bucketed is bucketed, so it's already distributed

File: ql/src/test/queries/clientpositive/acid_direct_update_delete.q [hunk 5]
```diff
@@ -18,4 +18,16 @@ DELETE FROM test_update_bucketed WHERE id IN ('2', '11', '10');
 UPDATE test_update_bucketed SET value='New value2' WHERE id IN ('2','18', '19');
 SELECT * FROM test_update_bucketed;
 
-DROP TABLE IF EXISTS test_update_bucketed;
\ No newline at end of file
+CREATE TABLE test_delete(id int) STORED AS ORC TBLPROPERTIES('transactional'='true');
+set mapreduce.job.reduces=7;
```
deniskuzZ: no need to set # of reducers! just use the provided snippet as is

kokila-19: Updated it.
I was testing few things and missed removing those lines. Thanks for finding it.

deniskuzZ (approved):
+1, pending tests

@kokila-19 please cherry-pick the fix into branch-4.1, once this PR is merged

PR #5902, opened 2025-06-26
HIVE-28975: [HiveAcidReplication] Remove dangling txns from Target side post incremental replication

harshal-16:
Details:
	* Currently, if for some reason on the source side if there is any entry missing in notification_log / txn_write_notification_log, then it may lead to dangling open transaction on Target side post replication
	* Repl created Open transaction can lead to repl_incompatible error once ACIDHouseKeeper thread kills the dangling transaction
	* Instead of that we can proactively clean the dangling transactions on the Target side preventing replication failure
	* It is proactive measure to protect replication against future bugs similar to HIVE-27797

Testing:
	* Tested on Cluster
	* Added test case


deniskuzZ:
@harshal-16, why do we need gen classes for cpp, php, rb, py? multi-language HMS thrift clients?

harshal-16:
> @harshal-16, why do we need gen classes for cpp, php, rb, py? multi-language HMS thrift clients?

I am not very sure about its usage, but as it was already present, I have kept the generated code. It might be used for connecting some other services other than HS2, which I am not aware of.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -10944,6 +10944,21 @@ private void abortReplCreatedOpenTxnsForDatabase(String query) throws TException
...
+  @Override
+  public ReplayedTxnsForPolicyResult get_replayed_txns_for_policy(String policyName)  {
+    Exception ex = null;
+    try{
+      startFunction("get_replayed_txns_for_policy");
+      return getTxnHandler().getReplayedTxnsForPolicy(policyName);
+    } catch (Exception e) {
+      LOG.error("Caught exception", e);
+      ex = e;
+      throw e;
```
deniskuzZ: - no need to log
- throw newMetaException(e);

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -10944,6 +10944,21 @@ private void abortReplCreatedOpenTxnsForDatabase(String query) throws TException
...
+    } finally {
+      endFunction("get_replayed_txns_for_policy", ex == null, ex);
```
deniskuzZ:
````
ret = getTxnHandler().getReplayedTxnsForPolicy(policyName)
} finally {
endFunction("get_replayed_txns_for_policy", ret != null, ex);
}
return ret;
````

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java [hunk 3]
```diff
@@ -657,6 +658,21 @@ public void replTableWriteIdState(ReplTblWriteIdStateRequest rqst) throws MetaEx
...
 
+  /**
+   *
+    * @param replPolicy replication policy for which we want to get the replayed transactions
+   * @return Map of source and target transaction ids for the given replication policy.
+   */
+  @Override
+  public ReplayedTxnsForPolicyResult getReplayedTxnsForPolicy(String replPolicy) {
+    try {
+      return jdbcResource.execute(new GetReplayedTxnsForPolicyHandler(replPolicy));
```
deniskuzZ: should it be retriable?
````
sqlRetryHandler.executeWithRetry(
````

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
harshal-16: Done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
harshal-16: Done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java
harshal-16: Done

harshal-16:
incorporated sonarcube suggestion

deniskuzZ:
rebased with master

harshal-16:
SonarCube shows an issue from someone else's old code https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=5902&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

PR #5904, opened 2025-06-26
HIVE-29048: Increase default pool and queue size of HS2 Background Async Threadpool.

mdayakar:
HIVE-29048: Increase default pool and queue size of HS2 Background Async Threadpool.

To increase the default value for the [hive.server2.async.exec.threads](https://github.com/apache/hive/blob/68f8734b3b7db2fdc8bc0c11ad3c37eb2e2a808c/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java#L4233-L4234) and [hive.server2.async.exec.wait.queue.size](https://github.com/apache/hive/blob/68f8734b3b7db2fdc8bc0c11ad3c37eb2e2a808c/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java#L4238-L4240) configuration from **100** to **250**.

Recently most of the customers reported below error on beeline
```
0: jdbc:hive2://localhost> show databases;
Error: The background threadpool cannot accept new task for execution, please retry the operation (state=,code=0)
```

HS2 server logs shows below exception
```
Caused by: java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.FutureTask@24363395 rejected from java.util.concurrent.ThreadPoolExecutor@54ef7045[Running, pool size = 100, active threads = 100, queued tasks = 100, completed tasks = 682504]
at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) ~[?:1.8.0_342]
at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) ~[?:1.8.0_342]
at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) ~[?:1.8.0_342]
at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:112) ~[?:1.8.0_342]
at org.apache.hive.service.cli.session.SessionManager.submitBackgroundOperation(SessionManager.java:760) ~[hive-service-3.1.3]
at org.apache.hive.service.cli.session.HiveSessionImpl.submitBackgroundOperation(HiveSessionImpl.java:598) ~[hive-service-3.1.3]
at org.apache.hive.service.cli.operation.SQLOperation.runInternal(SQLOperation.java:279) ~[hive-service-3.1.3]
... 50 more
```
So here for most of the customers load, the default value 100 is not sufficient so their queries are getting rejected. 

The changes required to handle most of the customer's cluster basic load.

No

NA

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -4230,12 +4230,12 @@ public class HiveConf extends Configuration {
         "excessive threads are killed after this time interval."),
 
     // Configuration for async thread pool in SessionManager
-    HIVE_SERVER2_ASYNC_EXEC_THREADS("hive.server2.async.exec.threads", 100,
+    HIVE_SERVER2_ASYNC_EXEC_THREADS("hive.server2.async.exec.threads", 250,
         "Number of threads in the async thread pool for HiveServer2"),
     HIVE_SERVER2_ASYNC_EXEC_SHUTDOWN_TIMEOUT("hive.server2.async.exec.shutdown.timeout", "10s",
         new TimeValidator(TimeUnit.SECONDS),
         "How long HiveServer2 shutdown will wait for async threads to terminate."),
-    HIVE_SERVER2_ASYNC_EXEC_WAIT_QUEUE_SIZE("hive.server2.async.exec.wait.queue.size", 100,
+    HIVE_SERVER2_ASYNC_EXEC_WAIT_QUEUE_SIZE("hive.server2.async.exec.wait.queue.size", 250,
         "Size of the wait queue for async thread pool in HiveServer2.\n" +
         "After hitting this limit, the async thread pool will reject new requests."),
     HIVE_SERVER2_ASYNC_EXEC_KEEPALIVE_TIME("hive.server2.async.exec.keepalive.time", "10s",


deniskuzZ:
@mdayakar, these configs remain unchanged since Hive 2.3. Typically, customers can override them by modifying hive-site.xml
downstream we use hive-site.xml template with the following values, but it’s not a one-size-fits-all solution
````
hive.server2.async.exec.threads=200
hive.server2.async.exec.wait.queue.size=1000
````

mdayakar:
> @mdayakar, these configs remain unchanged since Hive 2.3. Typically, customers can override them by modifying hive-site.xml downstream we use hive-site.xml template with the following values, but it’s not a one-size-fits-all solution
> 
> ```
> hive.server2.async.exec.threads=200
> hive.server2.async.exec.wait.queue.size=1000
> ```
Hi @deniskuzZ ,
I agree that users can modify the value in hive-site.xml or they can set through via beeline also but if they are not configuring it in hive-site.xml then the default value 100 will be considered and for most of the customer's workload this value is not sufficient. So I proposed to increase this value to 250. Anyway if there is no load on the cluster these threads will be terminated based on keepalive timeout(10s) so even if the default value is 250 there will not be any issue. This is mainly for those customers who uses the default value configured in HiveConf.java file and faces the RejectedException due to 100 default value.

deniskuzZ:
> Hi @deniskuzZ , I agree that users can modify the value in hive-site.xml or they can set through via beeline also but if they are not configuring it in hive-site.xml then the default value 100 will be considered and for most of the customer's workload this value is not sufficient. So I proposed to increase this value to 250. Anyway if there is no load on the cluster these threads will be terminated based on keepalive timeout(10s) so even if the default value is 250 there will not be any issue. This is mainly for those customers who uses the default value configured in HiveConf.java file and faces the RejectedException due to 100 default value.

Cluster administrator should manage the configs. None of production setups should rely on default values. Note, I don't have strong opinions about this change

+0

PR #5907, opened 2025-06-27
HIVE-29047: Adjust httpclient5 httpcore dependencies in TestHttpServices

richardantal:
Adjust httpclient5 httpcore dependencies in TestHttpServices
https://github.com/richardantal/hive/pull/new/HIVE-29047

In TestHttpServices we use transitive httpclient5 and httpcore5 dependencies because of
https://github.com/apache/hive/commit/174ac6215a2b70c655d3dcfed3dc0001ecb97b9b
Later It will break in case of 1.26 avatica version

No.

I ran TestHttpServices unit test

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hive/service/TestHttpServices.java
+++ b/itests/hive-unit/src/test/java/org/apache/hive/service/TestHttpServices.java
@@ -20,11 +20,11 @@ package org.apache.hive.service;
 
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
-import org.apache.hc.client5.http.classic.methods.HttpGet;
-import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
-import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
-import org.apache.hc.client5.http.impl.classic.HttpClients;
-import org.apache.hc.core5.http.Header;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.impl.client.HttpClients;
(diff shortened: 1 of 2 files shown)


iceberg/pom.xml
```diff
@@ -74,6 +74,16 @@
...
+        <exclusions>
+          <exclusion>
+            <groupId>org.apache.httpcomponents.client5</groupId>
+            <artifactId>httpclient5</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>org.apache.httpcomponents.core5</groupId>
+            <artifactId>httpcore5</artifactId>
+          </exclusion>
+        </exclusions>
```
okumin: What happens if we don't exclude it? We eventually need this dependency when we adapt Hive QL to Iceberg REST API. It sounds fine to exclude it at this point.

okumin (approved):
+1

iceberg/pom.xml
richardantal: It works without the exclusions too.

richardantal:
I rebased the change on top of master

richardantal:
Do we need an other clean test run or how can we move forward with this?

okumin:
@richardantal
I periodically check the CI status and rerun it if it fails. Therefore, there are no remaining tasks on your end. The GitHub Actions error is likely caused by the following issue, which will be resolved soon. I will rerun the task in a few hours. Thank you!

https://lists.apache.org/thread/qbfh1fq6rj629y5xrt45v0qw64828zf2

abstractdog:
@richardantal : thanks for this patch! please ban http components 5 import to prevent future unintentional usage
if that's taken care of, I'm fine with the patch!

abstractdog (approved):
LGTM, pending tests

okumin:
Merged. @richardantal Thanks for submitting the patch, @abstractdog thanks for the additional review!

richardantal:
Thank you for merging, and for the reviews.

PR #5908, opened 2025-06-27
HIVE-29051: TestTezTPCDS30TBPerfCliDriver leaks Postgres container causing intermittent failures elsewhere

zabetak:
1. Move metastore/container destroy logic in buildClassRule of CliAdapter. All database/container lifecyle handling must be in the same place to avoid leaks and inconsistent behavior.
2. Rename QTestMetastoreHandler#afterTest to QTestMetastoreHandler#truncateDatabase to better reflect what is actually doing
3. Perform database truncation (Derby only) only inside buildTestRule. No need to do it also as part of the class rule
4. Handle metastore configuration exclusively in buildClassRule. We need to set the configuration only once before the first test runs and this is exactly what happens if we put it in the class rule.
5. Drop firstTestNotYetRun flag along with related code since there is no longer overlapping logic between test and class rule.
6. Drop now redundant QTestMetastoreHandler#beforeTest method

No

Run the following and after each verified (using `docker ps`) that no container is running.
```
mvn test -Dtest=TestTezTPCDS30TBPerfCliDriver -Dqfile=cbo_query1.q,cbo_query2.q 
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dtest.metastore.db=postgres -Dqfile="acid_insert_overwrite_update.q,acid_stats3.q"
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile="acid_insert_overwrite_update.q,acid_stats3.q"
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile_regex=cbo_.* -Dtest.output.overwrite 
```

### DIFF
--- a/itests/qtest/src/test/java/org/apache/hadoop/hive/cli/TestTezTPCDS30TBPerfCliDriver.java
+++ b/itests/qtest/src/test/java/org/apache/hadoop/hive/cli/TestTezTPCDS30TBPerfCliDriver.java
@@ -44,27 +44,8 @@ public class TestTezTPCDS30TBPerfCliDriver {
   @ClassRule
   public static TestRule cliClassRule = adapter.buildClassRule();
 
-  /**
-   * Rule for calling only {@link CliAdapter#setUp()} and {@link CliAdapter#tearDown()} before/after running each test.
-   *
-   * At the moment of writing this class the rule is mostly necessary for calling {@link CliAdapter#tearDown()} to avoid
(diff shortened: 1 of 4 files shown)


zhangbutao:
I guess this PR can fix the flaky test https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5884/3/tests

zabetak:
@zhangbutao That's exactly what is fixing :D

zabetak:
Many thanks for the review @zhangbutao !

PR #5909, opened 2025-06-28
[WIP] HIVE-29053: _REPLACE_WITH directory are created when running qtest using sysdb

Aggarwal-Raghav:
[HIVE-29053](https://issues.apache.org/jira/browse/HIVE-29053), Skip the LOCATION field for proto tables when running qtests.

Running the qtest containing `--! qt:sysdb`, causes creation of following dir in itests/qtest directory.
```
_REPLACE_WITH_APP_DATA_LOCATION_
_REPLACE_WITH_DAG_DATA_LOCATION_
_REPLACE_WITH_DAG_META_LOCATION_
_REPLACE_WITH_QUERY_DATA_LOCATION_
```

It better to have the proto tables share same location as other sys tables for testing/debugging purposes. 

No

Ran the following command with and without the PR and observed the directory creation in `itests/qtest` directory
```
mvn clean test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=query_history.q -Drat.skip -pl itests/qtest -Pitests
mvn clean test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=sysdb.q -Drat.skip -pl itests/qtest -Pitests
```

### DIFF
--- a/cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
+++ b/cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
@@ -500,6 +500,9 @@ public class CliDriver {
     StringBuilder qsb = new StringBuilder();
 
     while ((line = r.readLine()) != null) {
+      // Skipping the proto tables location.
+      if (line.startsWith("LOCATION '_REPLACE_WITH_")) continue;
+
       // Skipping through comments
       if (! line.startsWith("--")) {
         qsb.append(line + "\n");


Aggarwal-Raghav:
@okumin @deniskuzZ @KiranVelumuri , can you please help with the review?

okumin:
I verified the issue exists on the master branch, and this patch will resolve it

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
```diff
@@ -500,6 +500,9 @@ public CommandProcessorResponse processReader(BufferedReader r) throws IOExcepti
     StringBuilder qsb = new StringBuilder();
 
     while ((line = r.readLine()) != null) {
+      // Skipping the proto tables location.
+      if (line.startsWith("LOCATION '_REPLACE_WITH_")) continue;
```
okumin, line 504: The issue might not occur with this change, but this workaround sounds a bit tricky. Although I haven't checked the feasibility, is it impossible to update this handler?
https://github.com/apache/hive/blob/master/itests/util/src/main/java/org/apache/hadoop/hive/ql/qoption/QTestSysDbHandler.java

Aggarwal-Raghav: Hi @okumin, thanks for the review. Here are few things that I checked before arriving at this PR:

1. `HiveSchemaTool#replaceLocationForProtoLogTables()` is the one which is doing the replacement for LOCATION during the schematool initialization at runtime (cluster setup etc.). But calling that method in qtest (CliDriver) is "little" troublesome and even after achieving that might not yield expected results.
2. "Why calling `replaceLocationForProtoLogTables` is troublesome" => because there are factors like conf object and adding dependency on tez-api and hive-beeline modules.
3. "Even after achieving that might not yield expected results" => in qtests, sysdb tables like `CDS`, `COLUMNS_V2` will have table location `.../itests/qtest/target/localfs/warehouse/sys.db/<tbl_name>/` but for proto tables the location will either be picked from config or default location i.e. `/tmp/` or `/tmp/query_data`.
4. IMO, having a uniform location for all tables for testing purpose is better.

KiranVelumuri:
LGTM
nit: you could combine this into the existing logic to have like - skipping through comments and proto tables location

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
okumin: How about simply replacing the variables with the full path on the test filesystem?
I would say the current one has no side effect since CliDriver is used in only tests. When we drop it, we have to implement the same logic on Beeline. It is likely hacky or troublesome in the worst case to interrupt with the parser's behavior

okumin: To state my intention clearly, I'm saying we can replace this line

```
String stsdbPath =
          HiveTestEnvSetup.HIVE_ROOT + "/standalone-metastore/metastore-server/src/main/sql/hive/hive-schema-" + schemaVersion + ".hive.sql";
qt.getCliDriver().processLine("source " + stsdbPath)
```

with like this. Pseudo code.

```
String templatePath =
          HiveTestEnvSetup.HIVE_ROOT + "/standalone-metastore/metastore-server/src/main/sql/hive/hive-schema-" + schemaVersion + ".hive.sql";
String newQueries = read(templatePath).replace("_REPLACE_WITH_APP_DATA_LOCATION_", ".../itests/qtest/target/localfs/warehouse/sys.db/app_data").replace("_REPLACE_WITH_(omitted)", "(omitted)").replace("(omitted)", "(omitted)").replace("(omitted)", "(omitted)");
String sourcePath = write(templatePath, "somewhere");
qt.getCliDriver().processLine("source " + sourcePath)
```

Then, we don't need to hack a parser or drivers.

Aggarwal-Raghav: @okumin, with the above pseudocode, it will end up reading the schema file + writing modified schema file (QTestSysDbHandler.java) and then again reading a modified schema file (CliDriver.java) **for each q file using sysdb**. I'm not inline with this approach.

okumin: So, how about using `processLine` directly or a suitable method? `cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java` is not test code, but production code, although it's deprecated. It's better to avoid adding contextual branches.

Aggarwal-Raghav: @okumin, just thinking out loud and feel free to correct me if I'm wrong, writing the modified schema can lead to concurrency problems when UT (splits) are running parallely. 

I understand why you want changes in QTestSysDbHandler instead of CliDriver. I'll see if its possible and update accordingly.

okumin: If we find a contention, we can add a UUID to the path name. We should minimize the code that modifies an SQL statement outside the ANTLR parser. It could result in the most severe vulnerability, such as an SQL injection.

Aggarwal-Raghav:
Just an update, i just found that the issue exists outside of qtest as well. Basically any test sourcing/using 4.1.0 schema file will face this.

https://github.com/apache/hive/blob/master/itests/hive-unit/src/test/java/org/apache/hive/service/server/InformationSchemaWithPrivilegeTestBase.java#L298-L304

running the following test also create _REPLACE dir.
```
cd itests/hive-unit; mvn clean test -Dtest=TestInformationSchemaWithPrivilegeZookeeperPlain -Drat.skip
```

deniskuzZ:
hi @okumin,  is this good to be merged or it still requires some changes?

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
deniskuzZ, line 504: @Aggarwal-Raghav could we move the string literal to constants (i.e. `PROTO_TABLE_LOCATION_PREFIX`)?

Aggarwal-Raghav:
@deniskuzZ , thanks for taking a look. 

The presence of `LOCATION` field for the proto tables in the hive-schema-4.*.0.hive.sql is the problem, the way I see it.
No other tables (TBLS etc.) has the location field. If schema is instantiated via schemtool functionality then its fine as `HiveSchemaTool#replaceLocationForProtoLogTables()` will help. 
But in UT, the schematool is not used. Hence the following will create the problem.

1. In QTestSysDbHandler.java
2. InformationSchemaWithPrivilegeTestBase.java 

The current PR will help with [1] but for [2] it will not work.

okumin:
@deniskuzZ @Aggarwal-Raghav Sorry for the late response. I currently prefer to rewrite an SQL outside the compiler rather than to rewrite a part of an SQL inside the processor. That's because the latter approach is very ad-hoc and can potentially bring security problems.

PR #5911, opened 2025-07-01
HIVE-22810: Add drop scheduled query IF EXISTS support

Neer393:
The changes proposed adds support for IF EXISTS while dropping a Scheduled Query

User wants a safety net while dropping a scheduled query and not straightaway get a NoSuchQueryObjectException

Now, user can add IF EXISTS while dropping Scheduled Query which would not throw an Exception when the Scheduled Query does not exist

I tested it on my local machine


parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g
```diff
@@ -1745,10 +1745,8 @@ createScheduledQueryStatement
 dropScheduledQueryStatement
 @init { pushMsg("drop scheduled query statement", state); }
 @after { popMsg(state); }
-    : KW_DROP KW_SCHEDULED KW_QUERY name=identifier
-    -> ^(TOK_DROP_SCHEDULED_QUERY
-            $name
-        )
+    : KW_DROP KW_SCHEDULED KW_QUERY ifExists? name=identifier
+    -> ^(TOK_DROP_SCHEDULED_QUERY $name ifExists?)
```
deniskuzZ: could we please keep the old formatting for readability
````
: KW_DROP KW_SCHEDULED KW_QUERY ifExists? name=identifier
-> ^(TOK_DROP_SCHEDULED_QUERY
        $name ifExists?
    )
````

Comment (Neer393) [reply parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g]: Sure.
Other than that, does the PR look good ?

File: ql/src/java/org/apache/hadoop/hive/ql/parse/ScheduledQueryAnalyzer.java [hunk 2]
```diff
@@ -74,7 +74,11 @@ public void analyzeInternal(ASTNode ast) throws SemanticException {
     ScheduledQueryMaintenanceWork work;
     ScheduledQueryMaintenanceRequestType type = translateAstType(ast.getToken().getType());
     ScheduledQuery parsedSchq = interpretAstNode(ast);
-    ScheduledQuery schq = fillScheduledQuery(type, parsedSchq);
+    boolean dropIfExists = ast.getToken().getType() == HiveParser.TOK_DROP_SCHEDULED_QUERY && ast.getChildCount() > 1 && ast.getChild(1).getType() == HiveParser.TOK_IFEXISTS;
```
deniskuzZ: I think, we should create a separate analyzer here, similar to `DropFunctionAnalyzer`
````
@DDLType(types = HiveParser.TOK_DROP_SCHEDULED_QUERY)
public class DropScheduledQueryAnalyzer extends AbstractFunctionAnalyzer {
````

or at least handle it properly
````
boolean throwException = true;
if (ScheduledQueryMaintenanceRequestType.DROP == type) {
  boolean ifExists = ast.getFirstChildWithType(HiveParser.TOK_IFEXISTS) != null;
  throwException = !ifExists && !HiveConf.getBoolVar(conf, ConfVars.DROP_IGNORES_NON_EXISTENT);
}
````

File: ql/src/java/org/apache/hadoop/hive/ql/parse/ScheduledQueryAnalyzer.java [hunk 3]
```diff
@@ -74,7 +74,11 @@ public void analyzeInternal(ASTNode ast) throws SemanticException {
...
+    ScheduledQuery schq = fillScheduledQuery(type, parsedSchq, !dropIfExists);
+    if (schq == null) {
```
deniskuzZ, line 84:
````
if (schq == null) {
  if (throwException) {
    throw new SemanticException(ErrorMsg.INVALID_SCHEDULED_QUERY.getMsg(parsedSchq));
  } else {
    return; // Fail silently
  }
````

File: ql/src/java/org/apache/hadoop/hive/ql/parse/ScheduledQueryAnalyzer.java [hunk 4]
```diff
@@ -102,7 +106,11 @@ private ScheduledQuery  fillScheduledQuery(ScheduledQueryMaintenanceRequestType
         schqStored.setNextExecutionIsSet(false);
         return composeOverlayObject(schqChanges, schqStored);
       } catch (TException e) {
-        throw new SemanticException("unable to get Scheduled query" + e);
+        if (throwException) {
```
deniskuzZ: log the exception and return null

deniskuzZ:
It would be nice to have at least 1 test for the new functionality

PR #5913, opened 2025-07-01
HIVE-28667: Error initializing field trimmer instance when starting HS2 (#5879)

zhangbutao:

A cherry-pick for HIVE-28667.

Although this patch is a relatively minor change. However, this change can eliminate unnecessary error logs in Hive, which will help improve the user experience when viewing Hive logs.
So it is necessary to cherry-pick this patch to branch-4.1.


zhangbutao:
Since this is a small cherry-pick, I will directly merge it if there is no further review,

PR #5916, opened 2025-07-01
HIVE-29055: TestHplSqlViaBeeLine#testHplSqlExitConditionHandler is broken and passes silently

mdayakar:
HIVE-29055: TestHplSqlViaBeeLine#testHplSqlExitConditionHandler is broken and passes silently

Its just a test code change to correct a testcase related to exit handler feature testcase.

To properly test the exit handler feature.

No

_mvn test -Dtest=TestHplSqlViaBeeLine#testHplSqlExitConditionHandler -pl itests/hive-unit -Pitests_

### DIFF
--- a/itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
+++ b/itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
@@ -1052,17 +1052,22 @@ public class TestHplSqlViaBeeLine {
   @Test
   public void testHplSqlExitConditionHandler() throws Throwable {
     String scriptText =
-        "CREATE PROCEDURE p1()\n" +
+        "DROP TABLE IF EXISTS result;\n" +
+            "CREATE TABLE result (s string);\n" +
+            "CREATE PROCEDURE p1()\n" +
             "BEGIN\n" +
-            " PRINT('Exit CONDITION Handler invoked.');\n" +
+            " INSERT INTO result VALUES('Exit CONDITION Handler invoked.');\n" +
             "END;\n" +
-            "DECLARE cnt_condition CONDITION;\n" +
-            "DECLARE EXIT HANDLER FOR cnt_condition\n" +
-            " p1();\n" +
-            "IF 1 <> 2 THEN\n" +
-            " SIGNAL cnt_condition;\n" +
-            "END IF;";
-    testScriptFile(scriptText, args(), "Exit CONDITION Handler invoked.", OutStream.ERR);
+            "CREATE PROCEDURE p2()" +
+            "BEGIN\n" +
+            " DECLARE exit_condition CONDITION;\n" +
+            " DECLARE EXIT HANDLER FOR exit_condition\n" +
+            "   p1();\n" +
+            " SIGNAL exit_condition;\n" +
+            "END;\n" +
+            "p2();" +
+            "SELECT * FROM result;";
+    testScriptFile(scriptText, args(), "Exit CONDITION Handler invoked.");
   }
 
   @Test


itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
```diff
@@ -1052,17 +1052,22 @@ public void testHplSqlContinueConditionHandler() throws Throwable {
...
+            "CREATE PROCEDURE p2()" +
+            "BEGIN\n" +
+            " DECLARE exit_condition CONDITION;\n" +
+            " DECLARE EXIT HANDLER FOR exit_condition\n" +
+            "   p1();\n" +
+            " SIGNAL exit_condition;\n" +
+            "END;\n" +
+            "p2();" +
+            "SELECT * FROM result;";
+    testScriptFile(scriptText, args(), "Exit CONDITION Handler invoked.");
```
abstractdog, line 1070: this makes sense, as it asserts on STDOUT similarly to the continue handler, thanks!

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
```diff
@@ -1052,17 +1052,22 @@ public void testHplSqlContinueConditionHandler() throws Throwable {
...
-            "END IF;";
-    testScriptFile(scriptText, args(), "Exit CONDITION Handler invoked.", OutStream.ERR);
+            "CREATE PROCEDURE p2()" +
+            "BEGIN\n" +
+            " DECLARE exit_condition CONDITION;\n" +
+            " DECLARE EXIT HANDLER FOR exit_condition\n" +
+            "   p1();\n" +
+            " SIGNAL exit_condition;\n" +
+            "END;\n" +
+            "p2();" +
```
abstractdog, line 1068: can you please eaplain here for future clarity why the original testcase was invalid?
I'm not familiar with hplsql, so this makes me think that these statements somehow didn't trigger the exit condition:
```
IF 1 <> 2 THEN
 SIGNAL cnt_condition;
```

also PRINT has been changed to INSERT INTO: isn't a PRINT sufficient to demonstrate this hplsql feature and also demonstrate that p1() has been called as an EXIT HANDLER?

abstractdog:
TestTriggersNoTezSessionPool failure is not related, but I'm afraid we'll need a new test run

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
mdayakar: Here we are inserting string data into the table as a part of exit condition handler invocation and the same is getting asserted by seleting the data from that table.

itests/hive-unit/src/test/java/org/apache/hive/beeline/TestHplSqlViaBeeLine.java
abstractdog: I'm fine with "SELECT * FROM result", I was just wondering

mdayakar: Here in old testcase, the exit handler was getting invoked but somehow the print is not getting captured. So I changed it to use as a part of a procedure(which is normal use of HPLSQL) now its getting triggered and I could see in the Outstream.ERR but asserting there again will be same issue as its logs the triggered HPLSQL statements also. So now I changed to use inserting the string data into a table when exit handler is invoked and asserting the same by quering the data from that table.

abstractdog (approved):
LGTM, pending tests

PR #5917, opened 2025-07-02
HIVE-22810: Add drop scheduled query IF EXISTS support

Neer393:
The changes proposed adds support for IF EXISTS while dropping a Scheduled Query

User wants a safety net while dropping a scheduled query and not straightaway get a NoSuchQueryObjectException

Now, user can add IF EXISTS while dropping Scheduled Query which would not throw an Exception when the Scheduled Query does not exist

I tested it on my local machine

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/ScheduledQueryAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/ScheduledQueryAnalyzer.java
@@ -34,6 +34,7 @@ import org.apache.hadoop.hive.metastore.api.ScheduledQuery._Fields;
 import org.apache.hadoop.hive.metastore.api.ScheduledQueryKey;
 import org.apache.hadoop.hive.metastore.api.ScheduledQueryMaintenanceRequestType;
 import org.apache.hadoop.hive.ql.Context;
+import org.apache.hadoop.hive.ql.ErrorMsg;
 import org.apache.hadoop.hive.ql.QueryState;
 import org.apache.hadoop.hive.ql.exec.TaskFactory;
 import org.apache.hadoop.hive.ql.plan.HiveOperation;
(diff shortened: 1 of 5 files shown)


Neer393:
@deniskuzZ I made all the changes as per requested by you.
The only concern I have is here
`throwException = !ifExists && !HiveConf.getBoolVar(conf, ConfVars.DROP_IGNORES_NON_EXISTENT);`

I checked and found that the default value of this configuration is true. So a user running hive with default configuration would not get an error when he tries to drop a scheduled query which does not exist.
So if this sounds okay to you, we can move forward

deniskuzZ:
> @deniskuzZ I made all the changes as per requested by you. The only concern I have is here `throwException = !ifExists && !HiveConf.getBoolVar(conf, ConfVars.DROP_IGNORES_NON_EXISTENT);`
> 
> I checked and found that the default value of this configuration is true. So a user running hive with default configuration would not get an error when he tries to drop a scheduled query which does not exist. So if this sounds okay to you, we can move forward

isn't it the same for `drop function` or `drop table`?

Neer393:
> > @deniskuzZ I made all the changes as per requested by you. The only concern I have is here `throwException = !ifExists && !HiveConf.getBoolVar(conf, ConfVars.DROP_IGNORES_NON_EXISTENT);`
> > I checked and found that the default value of this configuration is true. So a user running hive with default configuration would not get an error when he tries to drop a scheduled query which does not exist. So if this sounds okay to you, we can move forward
> 
> isn't it the same for `drop function` or `drop table`?

Yes it's the same for them as well.

ql/src/java/org/apache/hadoop/hive/ql/parse/ScheduledQueryAnalyzer.java
```diff
@@ -102,7 +115,8 @@ private ScheduledQuery  fillScheduledQuery(ScheduledQueryMaintenanceRequestType
         schqStored.setNextExecutionIsSet(false);
         return composeOverlayObject(schqChanges, schqStored);
       } catch (TException e) {
-        throw new SemanticException("unable to get Scheduled query" + e);
+        LOG.error("Unable to find Scheduled query: " + schqChanges.getScheduleKey().getScheduleName(), e);
```
deniskuzZ, line 118: @Neer393, sorry, my bad. Can we pass `throwException` as it was before, similar to a `DropTableAnalyzer`   
````
ScheduledQuery schq = fillScheduledQuery(type, parsedSchq, throwException);
````
and then
````
} catch (NoSuchObjectException e) {
  if (throwException) {
    throw new SemanticException(ErrorMsg.INVALID_SCHEDULED_QUERY.getMsg(schqChanges.getScheduleKey().getScheduleName()), e);
  }
} catch (TException e) {
  throw new SemanticException(e.getMessage(), e);
}
````

Comment (Neer393) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/ScheduledQueryAnalyzer.java:118]: Sure, will do the changes

Comment (Neer393) [reply ql/src/java/org/apache/hadoop/hive/ql/parse/ScheduledQueryAnalyzer.java:118]:
```
} catch (NoSuchObjectException e) {
        LOG.error("Unable to find Scheduled query: " + schqChanges.getScheduleKey().getScheduleName(), e);
        if (throwException) {
          throw new SemanticException(ErrorMsg.INVALID_SCHEDULED_QUERY.getMsg(schqChanges.getScheduleKey().getScheduleName()), e);
        }
        return null;
      } catch (TException e) {
      }
```

You mean something like this ?

deniskuzZ: yes, just drop the logger since we return an exception to the client

Neer393: Okay. Let me just test the changes once again just to make sure I haven't broken anything making changes and I will raise the changes

deniskuzZ:
````
} catch (TException e) { 
  throw new SemanticException(e.getMessage(), e); 
}
````

PR #5919, opened 2025-07-02
HIVE-29032: Enhance qt:database option to expose connection properties in qfiles

zabetak:
1. Allow to define name/id for each database used via the qt:database directive
2. Expose database connection properties via system properties and manage them during the life-cycle of the database

1. Facilitate writing of qtests for the JDBC storage handler
3. Avoid harcoding connection properties in qfiles
4. Opens up the stage for supporting multiple databases of the same kind in qtests

No, it only affects the test infrastructure.

```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=cbo_jdbc_joincost.q,dataconnector_mysql.q,jdbc_filter_expand_row_operator.q,jdbc_partition_table_pruned_pcolumn.q,jdbc_project_pushdown.q,jdbc_table_dml_postgres.q,jdbc_table_limit_postgres.q,jdbc_table_with_schema_mariadb.q,jdbc_table_with_schema_mssql.q,jdbc_table_with_schema_oracle.q,jdbc_table_with_schema_postgres.q,qt_database_mssql.q,qt_database_mysql.q,qt_database_postgres.q -Dtest.output.overwrite
```

### DIFF
--- a/itests/util/src/main/java/org/apache/hadoop/hive/ql/qoption/QTestDatabaseHandler.java
+++ b/itests/util/src/main/java/org/apache/hadoop/hive/ql/qoption/QTestDatabaseHandler.java
@@ -27,20 +27,18 @@ import org.apache.hadoop.hive.ql.externalDB.PostgresExternalDB;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.EnumMap;
-import java.util.Map;
+import java.util.List;
 
 /**
  * An option handler for spinning (resp. stopping) databases before (resp. after) running a test.
  *
- * Syntax: qt:database:DatabaseType:path_to_init_script
+ * Syntax: qt:database:DatabaseType:databaseName:path_to_init_script
  *
- * The database type ({@link DatabaseType}) is obligatory argument. The initialization script can be omitted.
- *
+ * The database type ({@link DatabaseType}) and the database name are obligatory arguments. The initialization script can be omitted.
+ * The database name is mainly used to distinguish between different databases when they are accessed via a system {@link AbstractExternalDB.ConnectionProperty}.
  * Current limitations:
  * <ol>
  *   <li>Only one test/file per run</li>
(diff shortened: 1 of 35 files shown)


itests/util/src/main/java/org/apache/hadoop/hive/ql/qoption/QTestDatabaseHandler.java
```diff
@@ -27,20 +27,18 @@
...
-import java.util.Map;
+import java.util.List;
 
 /**
  * An option handler for spinning (resp. stopping) databases before (resp. after) running a test.
  *
- * Syntax: qt:database:DatabaseType:path_to_init_script
- *
- * The database type ({@link DatabaseType}) is obligatory argument. The initialization script can be omitted.
+ * Syntax: qt:database:DatabaseType:databaseName:path_to_init_script
```
deniskuzZ, line 38: Can we make it a bit more user-friendly?
````
qt:database:DatabaseType:schema:DatabaseName:path_to_init_script
````
--!qt:database:mysql:schema:qdb:q_test_author_book_tables.sql

Comment (zabetak) [reply itests/util/src/main/java/org/apache/hadoop/hive/ql/qoption/QTestDatabaseHandler.java:38]: This will make the option parsing trickier and maybe it's not worth it given that most devs will copy-paste from one file to another.

For the record, I also considered using commons-cli and [Options](https://github.com/apache/commons-cli/blob/e32c382354a54f3a719848180a480c3569cdd5b7/src/main/java/org/apache/commons/cli/Options.java) to allow the following syntax:
```
qt:database:--type mysql --name qdb --script q_test_author_book_tables.sql
qt:database:-t mysql -n qdb -s q_test_author_book_tables.sql
```
But it felt a bit of an overkill especially since `databaseType` and `databaseName` are mandatory args.

If the rest of the changes in this PR look good how about merging this (since tests are green) and following up with another ticket if you feel that its worth it?

Comment (deniskuzZ) [reply itests/util/src/main/java/org/apache/hadoop/hive/ql/qoption/QTestDatabaseHandler.java:38]: ok, though I like this syntax more:
```` 
qt:database:--type mysql --name qdb --script q_test_author_book_tables.sql
````

File: itests/util/src/main/java/org/apache/hadoop/hive/ql/qoption/QTestDatabaseHandler.java [hunk 2]
```diff
@@ -82,52 +80,48 @@ AbstractExternalDB create() {
     abstract AbstractExternalDB create();
   }
 
-  private final Map<DatabaseType, String> databaseToScript = new EnumMap<>(DatabaseType.class);
+  private final String scriptsDir;
+  private final List<AbstractExternalDB> databases = new ArrayList<>();
```
deniskuzZ, line 84: @zabetak, should it be a Set? What if the used defines init scripts for multiple schemas?

zabetak: Hey @deniskuzZ , I am not sure I understand what do you mean when you say init scripts for multiple schemas. Can you elaborate a bit about the use-case that you have in mind?

Note that as part of this change the initialization script is now part of the `AbstractExternalDB` object, which was not the case before.

deniskuzZ: @zabetak, in the below example, we could try to init multiple schemas in the same database
````
--!qt:database:mysql:qdb1:q_test_country_state_city_tables.sql
--!qt:database:mysql:qdb2:q_test_country_state_city_tables.sql
...
--!qt:database:mysql:qdb5:q_test_country_state_city_tables.sql
````

won't we try to init mysql db container 5 times
````
for (AbstractExternalDB db : databases) {
  db.start();
} 
````

Comment (zabetak) [reply itests/util/src/main/java/org/apache/hadoop/hive/ql/qoption/QTestDatabaseHandler.java:84]: Allowing the creation of multiple databases of the same type (e.g., mysql) is fine and would allow to simulate real customer use-cases. 

Defining the same database mysql multiple times won't work directly after this patch gets in but my intention is to make this happen by migrating `AbstractExternalDB` hierarchy to use [testcontainers](https://testcontainers.com/) (similar to HIVE-24419).

In addition, we already have some tests that use multiple derby databases and I would like to migrate them to use the qt:database framework (HIVE-29031) but I am waiting for this and few other tickets to go in first.

What is definitely not possible is to have the same database name defined multiple times. The following is invalid since it would be impossible to use the `System` property convention adopted here.
```
--!qt:database:mysql:qdb1:q_test_country_state_city_tables.sql
--!qt:database:postgres:qdb1:q_test_country_state_city_tables.sql
```

All in all if we really want to make the DB definition mechanism fully bullet-proof we would have to do two checks:
* Ensure DB name is unique
* Ensure DB type is unique (this limitation will be removed once we migrate to testcontainers library)

Assuming that we want to enforce both then the `List` data structure stills makes sense.

Do you want that I push a new commit to implement the aforementioned checks?

deniskuzZ: > Assuming that we want to enforce both then the List data structure stills makes sense.

how so, list doesn’t ensure uniqueness

zabetak: I mean using a `List` and doing a check before/after adding an element. If we use a `Set` or `Map` we would still need to iterate through all entries if we want to enforce the two checks mentioned above. The reason that I said that `List` makes sense is because iteration through all elements is usually faster.

deniskuzZ: ok, let's merge this

PR #5923, opened 2025-07-03
HIVE-29062: Add standalone module for packaging the Metastore

dengzhhu653:
- mvn clean install -DskipTests -Pdist,docker from root/standalone-metastore directory 
- Run the generated Metastore Docker image against Derby, MySQL, Postgres.


standalone-metastore/packaging/src/assembly/src.xml
```diff
@@ -28,11 +28,11 @@
     <format>tar.gz</format>
   </formats>
 
-  <baseDirectory>apache-${project.artifactId}-${project.version}-src</baseDirectory>
+  <baseDirectory>apache-${project.parent.artifactId}-server-${project.version}-src</baseDirectory>
```
deniskuzZ: src tarball should include all modules, please see https://issues.apache.org/jira/browse/HIVE-20451?focusedCommentId=16594201&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16594201

deniskuzZ:
I wanted to do the same, but wasn't sure on the naming. Should we still build the `apache-hive-standalone-metastore-[server]`-bin.tar.gz?

Also, we are lacking the HMS client. There is already a patch with +1, but for some reason it wasn't merged.
- Separate metastore client code into its own module
https://issues.apache.org/jira/browse/HIVE-20189

standalone-metastore/packaging/pom.xml
```diff
@@ -0,0 +1,228 @@
...
+            <artifactId>maven-assembly-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>assemble</id>
+                <phase>package</phase>
+                <goals>
+                  <goal>single</goal>
+                </goals>
+                <configuration>
+                  <finalName>apache-${project.parent.artifactId}-server-${project.version}</finalName>
```
deniskuzZ: Should we keep `server` part? i didn't remove it since it was in the server module before. Note, I don't have strong opinion on that, but in Hive-3.x it's missing
https://downloads.apache.org/hive/hive-standalone-metastore-3.0.0/

standalone-metastore/packaging/src/assembly/bin.xml
```diff
@@ -29,7 +29,7 @@
     <format>tar.gz</format>
   </formats>
 
-  <baseDirectory>apache-hive-metastore-${project.version}-bin</baseDirectory>
+  <baseDirectory>apache-${project.parent.artifactId}-server-${project.version}-bin</baseDirectory>
```
deniskuzZ: why change?

standalone-metastore/packaging/pom.xml
```diff
@@ -0,0 +1,228 @@
...
+    <profile>
+      <id>docker</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>exec-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>build-docker-image</id>
```
deniskuzZ: `id` could be removed and phase changed to `package`

standalone-metastore/packaging/src/docker/README.md
```diff
@@ -51,7 +51,7 @@ The `build.sh` provides ways to build the image against specified version of the
 
 ##### Build from source
 ```shell
-mvn clean install -DskipTests -DallModules -Pdocker
+mvn clean install -DskipTests -Pdocker
```
deniskuzZ: mvn clean `package`. Previously, there was a plugin conflict, so I had to use `install`

deniskuzZ:
@dengzhhu653, do you plan to include this in branch-4.1? 
cc @zhangbutao

standalone-metastore/packaging/src/docker/README.md
dengzhhu653: done

standalone-metastore/packaging/src/assembly/src.xml
dengzhhu653: yes, it includes all the modules of standalone metastore, I can build the project from the src tarball

standalone-metastore/packaging/pom.xml
dengzhhu653: I rename it to `hive-standalone-metastore-4.2.0-SNAPSHOT-bin.tar.gz` to align with the old releases.

standalone-metastore/packaging/src/assembly/bin.xml
dengzhhu653: renamed it to `${project.parent.artifactId}-${project.version}-bin` to keep the same pattern with the old releases, i.e, the hive-standalone-metastore-4.2.0-SNAPSHOT-bin.tar.gz

dengzhhu653:
> I wanted to do the same, but wasn't sure on the naming. Should we still build the `apache-hive-standalone-metastore-[server]`-bin.tar.gz?
> 
> Also, we are lacking the HMS client. There is already a patch with +1, but for some reason it wasn't merged.
> 
> * Separate metastore client code into its own module
>   https://issues.apache.org/jira/browse/HIVE-20189

Actually the Metastore needs the client, one example is the `PartitionManagementTask`.

> do you plan to include this in branch-4.1?

yes, it's applicable

dengzhhu653:
Upon the latest master, there is an error on the `Initiator`:
```
2025-07-03 15:48:50 Caused by: java.lang.NoClassDefFoundError: org/apache/tez/dag/app/TezTaskCommunicatorImpl
2025-07-03 15:48:50     at java.base/java.lang.ClassLoader.defineClass1(Native Method) ~[?:?]
2025-07-03 15:48:50     at java.base/java.lang.ClassLoader.defineClass(Unknown Source) ~[?:?]
2025-07-03 15:48:50     at java.base/java.security.SecureClassLoader.defineClass(Unknown Source) ~[?:?]
2025-07-03 15:48:50     at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(Unknown Source) ~[?:?]
2025-07-03 15:48:50     at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(Unknown Source) ~[?:?]
2025-07-03 15:48:50     at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(Unknown Source) ~[?:?]
2025-07-03 15:48:50     at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(Unknown Source) ~[?:?]
2025-07-03 15:48:50     at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(Unknown Source) ~[?:?]
2025-07-03 15:48:50     at java.base/java.lang.ClassLoader.loadClass(Unknown Source) ~[?:?]
2025-07-03 15:48:50     at org.apache.hadoop.hive.ql.exec.tez.TezSessionState.<clinit>(TezSessionState.java:115) ~[hive-exec-4.2.0-SNAPSHOT-core.jar:4.2.0-SNAPSHOT]
2025-07-03 15:48:50     at org.apache.hadoop.hive.ql.session.SessionState.start(SessionState.java:766) ~[hive-exec-4.2.0-SNAPSHOT-core.jar:4.2.0-SNAPSHOT]
2025-07-03 15:48:50     at org.apache.hadoop.hive.ql.session.SessionState.start(SessionState.java:683) ~[hive-exec-4.2.0-SNAPSHOT-core.jar:4.2.0-SNAPSHOT]
2025-07-03 15:48:50     at org.apache.iceberg.mr.hive.compaction.IcebergTableOptimizer.findPotentialCompactions(IcebergTableOptimizer.java:85) ~[hive-iceberg-handler-4.2.0-SNAPSHOT.jar:4.2.0-SNAPSHOT]

```
I think it might due to HIVE-29028, I haven't seen such error before `HIVE-29045: Preparing for 4.2.0 development`

zhangbutao:
`2025-07-03 15:48:50 Caused by: java.lang.NoClassDefFoundError: org/apache/tez/dag/app/TezTaskCommunicatorImpl`

It seems that the HS2 process requires the loading of **tez-dag.jar**, but currently in our HS2, only tez-api.jar is loaded.

standalone-metastore/packaging/src/assembly/bin.xml
deniskuzZ: @dengzhhu653 fyi, that is not an archive name, but the folder inside

standalone-metastore/packaging/pom.xml
deniskuzZ: @dengzhhu653 , fyi that is not an archive name, but a folder inside. old release had `apache-hive-metastore-3.0.0-bin`

deniskuzZ: should we keep the `apache-` prefix? maybe not, to align with the old releases
````
apache-hive-standalone-metastore-4.2.0-SNAPSHOT-bin.tar.gz
````

File: standalone-metastore/packaging/src/assembly/bin.xml (hunk 3 above)
Comment (deniskuzZ) [reply standalone-metastore/packaging/src/assembly/bin.xml]: old release package dir was `apache-hive-metastore-3.0.0-bin`

**Comment by deniskuzZ** [comment]
@dengzhhu653,  @zhangbutao, IcebergTableOptimizer is called from the HMS Initiator process, it shouldn't require the tez. I think we should revert
cc @difin

**Comment by deniskuzZ** [comment]
> Actually the Metastore needs the client, one example is the PartitionManagementTask.

true, still i've raised a PR for the [HIVE-20189: Separate metastore client code into its own module](https://github.com/apache/hive/pull/5924#top) to have a clear separation:
https://github.com/apache/hive/pull/5924

`hive-standalone-metastore-common` is used in multiple places including the `hive-common`

WDYT?

**Comment by deniskuzZ** [comment]
@dengzhhu653 you can also now drop the scope from
````
<dependency>
  <groupId>org.apache.hive</groupId>
  <artifactId>hive-standalone-metastore-server</artifactId>
  <version>${hive.version}</version>
  <scope>provided</scope>
</dependency>
````

File: standalone-metastore/packaging/pom.xml (hunk 2 above)
Comment (dengzhhu653) [reply standalone-metastore/packaging/pom.xml]: This PR keeps the old name pattern without the `apache-`

File: standalone-metastore/packaging/src/assembly/bin.xml (hunk 3 above)
Comment (dengzhhu653) [reply standalone-metastore/packaging/src/assembly/bin.xml]: now the dir is `apache-hive-metastore-4.2.0-SNAPSHOT-bin` after unzipping the tarball

File: standalone-metastore/packaging/src/docker/Dockerfile [hunk 6]
```diff
@@ -18,23 +18,23 @@ ARG BUILD_ENV
...
-ONBUILD COPY apache-hive-standalone-metastore-server-*-bin.tar.gz /opt
+ONBUILD COPY hive-standalone-metastore-*-bin.tar.gz /opt
 
 FROM ubuntu as archive
 ARG HADOOP_VERSION
 ARG HIVE_VERSION
 ONBUILD RUN apt-get update && apt-get -y install wget
 ONBUILD RUN wget https://archive.apache.org/dist/hadoop/core/hadoop-$HADOOP_VERSION/hadoop-$HADOOP_VERSION.tar.gz && \
- wget https://archive.apache.org/dist/hive/hive-standalone-metastore-$HIVE_VERSION/apache-hive-standalone-metastore-server-$HIVE_VERSION-bin.tar.gz
+ wget https://archive.apache.org/dist/hive/hive-standalone-metastore-$HIVE_VERSION/hive-standalone-metastore-$HIVE_VERSION-bin.tar.gz
```
deniskuzZ: please replace with `https://downloads.apache.org/hive/....`

packaging/src/docker/Dockerfile
```diff
@@ -26,9 +26,9 @@ ARG HADOOP_VERSION
 ARG HIVE_VERSION
 ARG TEZ_VERSION
 ONBUILD RUN apt-get update && apt-get -y install wget
-ONBUILD RUN wget https://archive.apache.org/dist/tez/$TEZ_VERSION/apache-tez-$TEZ_VERSION-bin.tar.gz && \
+ONBUILD RUN wget https://downloads.apache.org/tez/$TEZ_VERSION/apache-tez-$TEZ_VERSION-bin.tar.gz && \
  wget https://archive.apache.org/dist/hadoop/core/hadoop-$HADOOP_VERSION/hadoop-$HADOOP_VERSION.tar.gz && \
```
deniskuzZ, line 30: hadoop is still using archive, is that intentional?
````
https://downloads.apache.org/hadoop/core/hadoop-$HADOOP_VERSION/hadoop-$HADOOP_VERSION.tar.gz
````

File: standalone-metastore/packaging/pom.xml [hunk 8]
```diff
@@ -0,0 +1,220 @@
...
+        <exclusion>
+          <groupId>org.apache.hive</groupId>
+          <artifactId>hive-iceberg-catalog</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-exec</artifactId>
+      <version>${project.version}</version>
```
deniskuzZ: should we use `${hive.version}` instead of ${project.version} ? TBH, I am not sure why we needed the  `${hive.version}` property

packaging/src/docker/Dockerfile
dengzhhu653: yes, in case the hadoop being built cannot find in the `https://downloads.apache.org`

standalone-metastore/packaging/src/assembly/src.xml
```diff
@@ -28,28 +28,38 @@
...
       </includes>
-      <useDefaultExcludes>true</useDefaultExcludes>
-    </fileSet>
-    <fileSet>
-      <directory>${project.basedir}</directory>
-      <useDefaultExcludes>true</useDefaultExcludes>
       <excludes>
         <exclude>**/*.log</exclude>
         <exclude>**/*.iml</exclude>
+        <exclude>packaging/cache/**</exclude>
```
deniskuzZ, line 52: should we exclude
````
  <exclude>metastore-common/src/gen/version/**</exclude>
  <exclude>metastore-server/src/gen/**</exclude>
```

standalone-metastore/packaging/src/assembly/src.xml
```diff
@@ -28,28 +28,38 @@
...
 
-  <baseDirectory>apache-${project.artifactId}-${project.version}-src</baseDirectory>
+  <baseDirectory>apache-hive-metastore-${project.version}-src</baseDirectory>
 
   <fileSets>
     <fileSet>
-      <directory>${project.basedir}</directory>
+      <directory>${project.parent.basedir}</directory>
       <includes>
         <include>DEV-README</include>
```
deniskuzZ, line 37: main packaging for some reason includes
````
  <include>.checkstyle</include>
  <include>.gitattributes</include>
  <include>.gitignore</include>
````

File: standalone-metastore/packaging/pom.xml (hunk 8 above)
Comment (dengzhhu653) [reply standalone-metastore/packaging/pom.xml]: will change the dependencies to `hive.version` outside of the standalone-metastore

File: standalone-metastore/packaging/src/assembly/src.xml (hunk 10 above)
Comment (dengzhhu653) [inline standalone-metastore/packaging/src/assembly/src.xml:37]: these files are not present in the `standalone-metastore`, should I include them?

File: standalone-metastore/packaging/src/assembly/src.xml [hunk 11]
```diff
@@ -28,28 +28,38 @@
...
         <include>README*</include>
         <include>LICENSE*</include>
```
deniskuzZ: should we include these as well? note, was able to find `CHANGELOG`
```` 
<include>LICENSE</include>
<include>NOTICE</include>
<include>CHANGELOG</include>
<include>RELEASE_NOTES.txt</include>
````

File: standalone-metastore/packaging/src/assembly/src.xml [hunk 12]
```diff
@@ -28,28 +28,38 @@
...
         <include>NOTICE*</include>
         <include>pom.xml</include>
+        <include>metastore-common/**/*</include>
+        <include>metastore-server/**/*</include>
+        <include>metastore-tools/**/*</include>
+        <include>metastore-rest-catalog/**/*</include>
+        <include>packaging/**/*</include>
```
deniskuzZ, line 46:
````    
<include>packaging/pom.xml</include>
<include>packaging/src/**/*</include>
<include>pom.xml</include>
````

File: standalone-metastore/packaging/src/docker/Dockerfile [hunk 13]
```diff
@@ -18,23 +18,23 @@ ARG BUILD_ENV
...
 FROM ubuntu as unarchive
 ONBUILD COPY hadoop-*.tar.gz /opt
-ONBUILD COPY apache-hive-standalone-metastore-server-*-bin.tar.gz /opt
+ONBUILD COPY hive-standalone-metastore-*-bin.tar.gz /opt
 
 FROM ubuntu as archive
 ARG HADOOP_VERSION
 ARG HIVE_VERSION
 ONBUILD RUN apt-get update && apt-get -y install wget
 ONBUILD RUN wget https://archive.apache.org/dist/hadoop/core/hadoop-$HADOOP_VERSION/hadoop-$HADOOP_VERSION.tar.gz && \
```
deniskuzZ, line 27: same as https://github.com/apache/hive/pull/5923/files#r2184664815

packaging/src/docker/Dockerfile
deniskuzZ: version doesn't exist? are we using archive version?
cc @ayushtkn

standalone-metastore/packaging/src/assembly/src.xml
deniskuzZ: 1. `LICENSE` and `NOTICE` do not need a pattern
2. i think it would be good to have `RELEASE_NOTES.txt` for HMS, but i don't know how to generate it, see https://issues.apache.org/jira/secure/ConfigureReleaseNote.jspa?projectId=12310843
3. `CHANGELOG` could be dropped from the main assembly

packaging/src/docker/Dockerfile
deniskuzZ: @dengzhhu653 https://github.com/apache/hive/blob/937d10069dc11143c42a521bb2fe0896a0b2d9d8/standalone-metastore/metastore-server/src/docker/build.sh#L69 was working for me. 
But if I missed something, we should change back hadoop tarball to archive

standalone-metastore/packaging/src/assembly/src.xml
deniskuzZ: https://github.com/apache/hive/pull/5923/files#r2184736512

packaging/src/docker/Dockerfile
dengzhhu653: Imaging if the version is 3.1.3, the hadoop binary couldn't find in `https://downloads.apache.org`, the `archive` contains all the releases of the hadoop

ayushtkn: In current hive the hadoop version is 3.4.1 & that is there in downloads.apache.org
https://downloads.apache.org/hadoop/core/hadoop-3.4.1/

Most probably @dengzhhu653 didn't drop /dist from the path, downloads.apache.org doesn't have /dist in the path

But using archive may be better, say today 3.4.1 isn't archived but later if it does or someone wants to build docker image with an older version of hadoop, that won't be found in downloads.apache.org where as archive.apache has all versions including the current ones
https://archive.apache.org/dist/hadoop/common/

deniskuzZ: The archive also includes EOL versions, which are not recommended for use

https://archive.apache.org/dist/hadoop/
````
THEY MAY BE UNSUPPORTED AND UNSAFE TO USE
Current releases can be found on our [download server](http://downloads.apache.org/).
````

In that case, @dengzhhu653 we need to revert `hadoop` to archive in https://github.com/apache/hive/blob/6c5e715005c3953de42b584d13b0169e48628e3f/standalone-metastore/packaging/src/docker/build.sh#L69

dengzhhu653: sure

standalone-metastore/packaging/src/assembly/src.xml
dengzhhu653: Any pointers how to generate the `CHANGELOG`? checked the hive tarball, this file is absent from here.

deniskuzZ: that is what i mentioned: `RELEASE_NOTES.txt` are basically a `CHANGELOG`, we can just drop it.
As for HMS `RELEASE_NOTES.txt` I don't have rights to configure new project or filter (https://issues.apache.org/jira/secure/ReleaseNote.jspa), @ayushtkn do you?

deniskuzZ: We can do https://issues.apache.org/jira/sr/jira.issueviews:searchrequest-html-current-fields/temp/SearchRequest.html?jqlQuery=project+%3D+HIVE+AND+fixVersion+%3D+4.1.0+and+component+in+%28%22Metastore%22%2C+%22Standalone+Metastore%22%2C+metastore%29+and+resolution+%3D+Fixed+

and then ask gemini to format it using Hive's template

dengzhhu653: I've just updated some's component to `Standalone Metastore`, there might be some more need to be updated

deniskuzZ: At the very least, a brief changelog is better than nothing :)

PR #5924, opened 2025-07-03
HIVE-20189: Separate metastore client code into its own module

deniskuzZ:

move client classes into it's own module

improve the structure of standalone-metastore classes

No

jenkins


standalone-metastore/metastore-client/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+    <groupId>org.apache.hive</groupId>
+    <version>4.2.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>hive-standalone-metastore-client</artifactId>
+  <name>Hive Metastore Client</name>
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-standalone-metastore-common</artifactId>
```
dengzhhu653, line 25: The new module depends on the `metastore-common`, which doesn't reduce the size or dependencies from `metastore-common`, not sure if it's ok, for the user what will benefit from the new module?

deniskuzZ: It's expected for a client module to depend on a common module. The client module delivers core client functionality and can be enhanced later with features like caching. 
Offering a 'client' JAR aligns with common conventions and is expected to be user-friendly.
ATM we have 2 distinct cache wrappers in different Hive modules, just because there was no structure:

- org.apache.hadoop.hive.metastore.HiveClientCache
- org.apache.iceberg.hive.CachedClientPool

deniskuzZ: another point I don't understand, why `ql` and `beeline` modules depend on `metastore-server`? I think we should drop the dependency on server and move the classes into the metastore common or client.

okumin: I basically agree that we can potentially improve the structure in the future. We will have obvious guidelines about how to organize files.
- metastore-client: Client-specific files, which Server doesn't need
- metastore-server: Server-specific files, which Client doesn't need
- metastore-common: Common modules

> another point I don't understand, why ql and beeline modules depend on metastore-server? I think we should drop the dependency on server and move the classes into the metastore common or client.

I guess `ql` requires metastore-server to use an embedded HMS.

deniskuzZ: isn't embedded HMS only used in tests?

okumin: No, it isn't. It can be used when `metastore.thrift.uris` is empty. For example, our Hive docker image(not HMS docker image) can set up a HiveServer2 without HMS. It probably uses the embedded mode that runs HMS-equivalent threads in HS2.

https://github.com/apache/hive/blob/d7b3a5ba4a6d3c4ea894fd36d0fc7e1f87a53e64/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java#L2654-L2663

okumin: Alternatively, we can say it is for testing purposes and separate the classes from hive-exec.

deniskuzZ:
I've kept the package unchanged for `IMetaStoreClient`, `HiveMetaStoreClient`, `RetryingMetaStoreClient`. Ideally, they should be moved under the `client` package. 
@dengzhhu653, @ngsg, @okumin,  WDYT should we refactor here (not a small PR already) or in a follow-up?

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveMetaStoreClientWithLocalCache.java
```diff
@@ -41,7 +41,7 @@
 import org.apache.hadoop.hive.metastore.api.PartitionsByExprRequest;
 import org.apache.hadoop.hive.metastore.api.PartitionsStatsRequest;
 import org.apache.hadoop.hive.metastore.api.Table;
-import org.apache.hadoop.hive.metastore.client.HiveMetaStoreClientUtils;
+import org.apache.hadoop.hive.metastore.client.utils.HiveMetaStoreClientUtils;
```
ngsg: I think the order of import might be swapped as follows:
```
import org.apache.hadoop.hive.metastore.client.MetaStoreClientWrapper;
import org.apache.hadoop.hive.metastore.client.utils.HiveMetaStoreClientUtils;
```

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
```diff
@@ -92,7 +92,7 @@
 import org.apache.hadoop.hive.metastore.api.TableValidWriteIds;
 import org.apache.hadoop.hive.metastore.api.UniqueConstraintsRequest;
 import org.apache.hadoop.hive.metastore.api.UniqueConstraintsResponse;
-import org.apache.hadoop.hive.metastore.client.HiveMetaStoreClientUtils;
+import org.apache.hadoop.hive.metastore.client.utils.HiveMetaStoreClientUtils;
```
ngsg: The same import order issue here.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/RetryingMetaStoreClient.java
```diff
@@ -65,15 +67,21 @@ public class RetryingMetaStoreClient implements InvocationHandler {
   private final UserGroupInformation ugi;
   private final int retryLimit;
   private final long retryDelaySeconds;
-  private final ConcurrentHashMap<String, Long> metaCallTimeMap;
+  private final Map<String, Long> metaCallTimeMap;
```
ngsg, line 69: What do you think about removing all uses of `ConcurrentHashMap` from this file? It seems that only one method uses `ConcurrentHashMap`, and the actual value passed to that method is always null.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/TableIterable.java
ngsg, line 1: It seems that the tab size is set to 4 inside `Iterator { }`. Could we restore it to 2?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/builder/HiveMetaStoreClientBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package org.apache.hadoop.hive.metastore.client.builder;
+
+import com.google.common.base.Preconditions;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.metastore.HiveMetaHookLoader;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.apache.hadoop.hive.metastore.api.MetaException;
+
```
ngsg: Could we remove the empty line and sort the import list?

ngsg (commented):
The patch looks reasonable to me. I quickly checked whether there are any remaining candidates that could be moved to standalone-metastore-client, and it seems that all client-only-related classes have been gathered into the client package. I'm also in favor of the new `HiveMetaStoreClientBuilder` class.

I left a few minor comments where I had questions. Also, it seems that the CI failed due to the unification of HiveClientCache. Could you update the `TestPassProperties#testSequenceTableWriteReadMR` accordingly?

---

Regarding `IMetaStoreClient`, `HiveMetaStoreClient`, and `RetryingMetaStoreClient`, I'm leaning towards refactoring them in a follow-up PR, as this large refactoring might obscure other changes.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/builder/HiveMetaStoreClientBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.function.Function;
+
+public class HiveMetaStoreClientBuilder {
+
+    private final Configuration conf;
+    private IMetaStoreClient client;
+
+    public HiveMetaStoreClientBuilder(Configuration conf) {
+        Preconditions.checkNotNull(conf);
+        this.conf = conf;
```
okumin: You can simplify it with one-line like `this.conf = Preconditions.checkNotNull(conf)`. I prefer `Objects.requireNonNull` for new code, though.

metastore/src/java/org/apache/hadoop/hive/metastore/HiveClientCache.java
```diff
@@ -107,7 +107,7 @@ public HiveClientCache(final int timeout) {
   /**
    * @param timeout the length of time in seconds after a client is created that it should be automatically removed
    */
-  private HiveClientCache(final int timeout, final int initialCapacity, final int maxCapacity, final boolean enableStats) {
+  protected HiveClientCache(final int timeout, final int initialCapacity, final int maxCapacity, final boolean enableStats) {
```
okumin: Is it simpler to make this public and directly instantiate HiveClientCache in HCatUtil?

deniskuzZ: WDYM, it is instantiated directly in HCatUtil, see https://github.com/apache/hive/pull/5924/files#diff-70dc922fa5f53c56617a8ac9684b41630cd9aa1b673504c7cf29e2f5798a81acR573-R582

okumin: Since a protected constructor is directly invisible from HCatUtil, we create it as an anonymous class that inherits `HiveClientCache`.

```
hiveClientCache = new HiveClientCache(
              hiveConf.getInt(
                  HCAT_HIVE_CLIENT_EXPIRY_TIME, DEFAULT_HIVE_CACHE_EXPIRY_TIME_SECONDS),
              hiveConf.getInt(
                  HCAT_HIVE_CLIENT_CACHE_INITIAL_CAPACITY, DEFAULT_HIVE_CACHE_INITIAL_CAPACITY),
              hiveConf.getInt(
                  HCAT_HIVE_CLIENT_CACHE_MAX_CAPACITY, DEFAULT_HIVE_CACHE_MAX_CAPACITY),
              hiveConf.getBoolean(
                  HCAT_HIVE_CLIENT_CACHE_STATS_ENABLED, DEFAULT_HIVE_CLIENT_CACHE_STATS_ENABLED))
{}; // ← THIS PART
```

deniskuzZ: omg, i missed that and was wondering how did it work with protected. 
changed to public, thanks!

hcatalog/core/src/main/java/org/apache/hive/hcatalog/common/HCatUtil.java
```diff
@@ -335,33 +323,24 @@ public static boolean validateMorePermissive(FsAction first, FsAction second) {
...
-    Configuration conf, String userName) throws Exception {
-    // LOG.info("getJobTrackerDelegationToken("+conf+","+userName+")");
-    JobClient jcl = new JobClient(new JobConf(conf, HCatOutputFormat.class));
-    Token<org.apache.hadoop.mapreduce.security.token.delegation.DelegationTokenIdentifier> t = jcl
-      .getDelegationToken(new Text(userName));
-    // LOG.info("got "+t);
+        Configuration conf, String userName) throws Exception {
+    Token<org.apache.hadoop.mapreduce.security.token.delegation.DelegationTokenIdentifier> t;
+    try (JobClient jcl = new JobClient(new JobConf(conf, HCatOutputFormat.class))) {
+      t = jcl.getDelegationToken(new Text(userName));
```
okumin: We may directly return it

deniskuzZ: changed

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -19,9 +19,9 @@
 
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.hadoop.hive.common.TableName;
-import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.TableType;
 import org.apache.hadoop.hive.metastore.Warehouse;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
```
ngsg: nit: I think the original import is sorted correctly, so we might not need this change.

deniskuzZ: reverted

PR #5925, opened 2025-07-03
HIVE-29064: Drop usage of repository.apache.org

zabetak:
The repository.apache.org is not a general-purpose maven repository so we shouldn't use it to fetch artifacts from there. Unnecessary usage puts stress on the Apache infrastructure and can lead into IP bans and CI failures. See for more info: https://infra.apache.org/abc/#BL018

In general, the less we rely on non-standard remote repositories the better in many aspects for devs, CI, and consumers that depend on us.

No

Built the project successfully using a clean local maven repository.
```
mvn clean install -DskipTests -Pitests -Dmaven.repo.local=/tmp/m2/hive
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -248,16 +248,6 @@
       </releases>
     </repository>
     <repository>
-      <id>repository-release</id>
-      <url>https://repository.apache.org/content/repositories/releases/</url>
-      <releases>
-        <enabled>true</enabled>
-      </releases>
-      <snapshots>
-        <enabled>true</enabled>
-      </snapshots>
-    </repository>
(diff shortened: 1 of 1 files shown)


ngsg:
+1 (non-binding). I experienced a similar compile issue with `repository.apache.org` last time and resolved it in exactly the same way.

zhangbutao (approved):
+1 LGTM

aturoczy (approved):

zabetak:
Thanks for the reviews @ngsg @zhangbutao @aturoczy !

PR #5926, opened 2025-07-04
HIVE-29067 - insert-select-union-all query fails while publishing InsertEvent if source query does not return output

nareshpr:
Bug fix to avoid insert-select-union-all query failure while publishing InsertEvent if source produce empty output

Query fails without this fix

No

Included testcase in q file

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -3961,7 +3961,7 @@ private void constructOneLBLocationMap(FileStatus fSta,
     }
   }
 
-  private void fireInsertEvent(Table tbl, Map<String, String> partitionSpec, boolean replace, List<FileStatus> newFiles)
+  void fireInsertEvent(Table tbl, Map<String, String> partitionSpec, boolean replace, List<FileStatus> newFiles)
       throws HiveException {
     if (conf.getBoolVar(ConfVars.FIRE_EVENTS_FOR_DML)) {
       LOG.debug("Firing dml insert event");
@@ -3974,11 +3974,10 @@ private void constructOneLBLocationMap(FileStatus fSta,
         FireEventRequestData data = new FireEventRequestData();
         InsertEventRequestData insertData = new InsertEventRequestData();
         insertData.setReplace(replace);
+        insertData.setFilesAdded(new ArrayList());
         data.setInsertData(insertData);
         if (newFiles != null && !newFiles.isEmpty()) {
           addInsertFileInformation(newFiles, fileSystem, insertData);
-        } else {
-          insertData.setFilesAdded(new ArrayList<String>());
         }
         FireEventRequest rqst = new FireEventRequest(true, data);
         rqst.setDbName(tbl.getDbName());
--- a/ql/src/test/org/apache/hadoop/hive/ql/metadata/TestHive.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/metadata/TestHive.java
@@ -1055,6 +1055,17 @@ public class TestHive {
       Path insertedPath = new Path(insertEvent.getFiles().get(i));
       Assert.assertEquals(expectedCheckSums.get(insertedPath.getName()), checkSums.get(i));
     }
+
+    // Fire the InsertEvent with empty folder
+    hiveDb.fireInsertEvent(table, null, false,
+      Lists.newArrayList(new FileStatus(5, true, 1, 64, 100, tablePath)));
+    // Get the last Metastore event
+    InsertEvent insertEvent1 = DummyFireInsertListener.getLastEvent();
+    // Check the event
+    Assert.assertNotNull(insertEvent1);
+    // getFiles should be empty and not null
+    Assert.assertNotNull(insertEvent1.getFiles());
+    Assert.assertTrue(insertEvent1.getFiles().isEmpty());
   }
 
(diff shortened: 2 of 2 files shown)


ngsg:
The patch looks reasonable to me, but I couldn't reproduce the reported issue in my local environment using the included query. (I used `TestMiniLlapLocalCliDriver` in the master branch.) Could you please check the qfile again?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -3977,10 +3977,9 @@ private void fireInsertEvent(Table tbl, Map<String, String> partitionSpec, boole
         InsertEventRequestData insertData = new InsertEventRequestData();
         insertData.setReplace(replace);
         data.setInsertData(insertData);
+        insertData.setFilesAdded(new ArrayList<String>());
```
deniskuzZ: no need for an explicit type. 
also could you please move it 1 line higher, so that initialization of insertData isn't scattered

nareshpr: Updated

nareshpr:
@ngsg Thrift validation before publishing events to remote HMS is failing, testcases run with embedded metastore.

deniskuzZ:
> @ngsg Thrift validation before publishing events to remove HMS is failing, testcases run with embedded metastore.

@nareshpr, I’m unable to reproduce the query failure as well. If the test case doesn’t trigger the issue, is it necessary to include it?
````
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=union2.q -Drat.skip=true
````
maybe add the test in `TestRemoteHiveMetaStore` or create iceberg module test using `TestHiveMetastore` & `HiveMetaStoreClient`

nareshpr:
@deniskuzZ testcase fails in cluster with remote metastore. I couldn't find straight forward test framework to test insert with remote metastore. I can remove the testcase if required, however i wanted to document it for future reference.

deniskuzZ:
> @deniskuzZ testcase fails in cluster with remote metastore. I couldn't find straight forward test framework to test insert with remote metastore. I can remove the testcase if required, however i wanted to document it for future reference.

@nareshpr there are number of tests that perform against the remote HMS. please check the `TestHiveRemote`

here is what Cursor generated for me:
````
Subject: [PATCH] UnitTest
---
Index: ql/src/test/org/apache/hadoop/hive/ql/metadata/TestHive.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/test/org/apache/hadoop/hive/ql/metadata/TestHive.java b/ql/src/test/org/apache/hadoop/hive/ql/metadata/TestHive.java
--- a/ql/src/test/org/apache/hadoop/hive/ql/metadata/TestHive.java	(revision 74e6fa6b9bfaa0f7691ca3eb6d3a4fe54f085fc5)
+++ b/ql/src/test/org/apache/hadoop/hive/ql/metadata/TestHive.java	(date 1751715623481)
@@ -1004,6 +1004,7 @@
     assertTrue(prevHiveObj != newHiveObj);
   }
 
+  @Test
   public void testFireInsertEvent() throws Throwable {
     Hive hiveDb = Hive.getWithFastCheck(hiveConf, false);
     String tableName = "test_fire_insert_event";
@@ -1012,6 +1013,13 @@
         HiveIgnoreKeyTextOutputFormat.class);
     Table table = hiveDb.getTable(tableName);
     Path tablePath = table.getDataLocation();
+
+    // Fire the InsertEvent with empty files list
+    hiveDb.fireInsertEvent(table, null, false,
+        Lists.newArrayList(
+            new FileStatus(5, true, 1, 64, 100, tablePath)
+        ));
+
     // Create some files that "inserted"
     FileSystem fileSystem = tablePath.getFileSystem(hiveConf);
     fileSystem.deleteOnExit(tablePath);
Index: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java	(revision 74e6fa6b9bfaa0f7691ca3eb6d3a4fe54f085fc5)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java	(date 1751715008738)
@@ -3963,7 +3963,7 @@
     }
   }
 
-  private void fireInsertEvent(Table tbl, Map<String, String> partitionSpec, boolean replace, List<FileStatus> newFiles)
+  void fireInsertEvent(Table tbl, Map<String, String> partitionSpec, boolean replace, List<FileStatus> newFiles)
       throws HiveException {
     if (conf.getBoolVar(ConfVars.FIRE_EVENTS_FOR_DML)) {
       LOG.debug("Firing dml insert event");

````

please apply and drop the qfile change

deniskuzZ (changes requested):
https://github.com/apache/hive/pull/5926#issuecomment-3038567414

nareshpr:
Thanks for the help @deniskuzZ
I will incorporate the changes and update the commit.

Earlier i tried to use org.apache.hadoop.hive.ql.metadata.Hive#fireInsertEvent(java.lang.String, java.lang.String, java.util.Map<java.lang.String,java.lang.String>, boolean, java.util.List<java.lang.String>) in testcase which accepts string path & got converted into file instead of directory. But your suggestion helps to pass empty folder.

PR #5927, opened 2025-07-04
HIVE-29065: Fix Dockerfile warnings - FromAsCasing and InvalidDefault…

kokila-19:
…ArgInFrom

In the Dockerfile
- Match the case for all keywords (UPPER CASE). 
- Default value for BUILD_ENV argument.  "FROM" determines the base image for the build and arg BUILD_ENV is used here. If it is empty/invalid , the image will not be created. So docker lint checks for this issue. To avoid it, a default value needs to set. 

To remove the warnings thrown during docker build.

No

Create the docker image using build script for both hive and hms image.
./build.sh

### DIFF
--- a/packaging/src/docker/Dockerfile
+++ b/packaging/src/docker/Dockerfile
@@ -14,14 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-ARG BUILD_ENV
+ARG BUILD_ENV=local
 
-FROM ubuntu as unarchive
+FROM ubuntu AS local
 ONBUILD COPY hadoop-*.tar.gz /opt
 ONBUILD COPY apache-hive-*-bin.tar.gz /opt
 ONBUILD COPY apache-tez-*-bin.tar.gz /opt
 
-FROM ubuntu as archive
+FROM ubuntu AS release
 ARG HADOOP_VERSION
 ARG HIVE_VERSION
 ARG TEZ_VERSION
@@ -33,7 +33,7 @@ ONBUILD RUN mv /apache-tez-$TEZ_VERSION-bin.tar.gz /opt && \
  mv hadoop-$HADOOP_VERSION.tar.gz /opt && \
  mv apache-hive-$HIVE_VERSION-bin.tar.gz /opt
 
-FROM ubuntu as buildarchive
+FROM ubuntu AS hybrid
 ARG HADOOP_VERSION
 ARG HIVE_VERSION
 ARG TEZ_VERSION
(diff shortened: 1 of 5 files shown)


packaging/src/docker/Dockerfile
```diff
@@ -14,14 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-ARG BUILD_ENV
+ARG BUILD_ENV=unarchive
```
aturoczy: unarchive means temp here? Maybe the staging is more expressive

packaging/src/docker/Dockerfile
```diff
@@ -14,14 +14,14 @@
...
 
-FROM ubuntu as unarchive
+FROM ubuntu AS unarchive
 ONBUILD COPY hadoop-*.tar.gz /opt
 ONBUILD COPY apache-hive-*-bin.tar.gz /opt
 ONBUILD COPY apache-tez-*-bin.tar.gz /opt
 
-FROM ubuntu as archive
+FROM ubuntu AS archive
```
aturoczy: Prod or maybe release is more expressive to me

aturoczy (approved):

packaging/src/docker/Dockerfile
kokila-19: Sorry for the delay in responding.
You're absolutely right that "unarchive" is confusing. Both "staging" and "local" would be much more expressive names:

1. "staging" clearly indicates this is for development/testing workflows
2. Since this stage specifically uses jars from local builds rather than remote repositories, "local" might be the most accurate and intuitive choice. It tells developers they need to have the binaries locally.

packaging/src/docker/Dockerfile
kokila-19: Proposed naming convention
1. local (instead of unarchive) :  Uses local build for both Hadoop and Hive
2. release (instead of archive) : Downloads official Apache releases for both components
3. hybrid (instead of buildarchive) : Uses locally built Hive with official Hadoop release from Apache repository

Do these terms provide clarity?

kokila-19:
Hi @deniskuzZ @difin, this PR was automatically closed due to staleness. The changes are still relevant. Would it be possible to reopen it or should I open a new PR instead? 

```
 => WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 19)                                                                                                                           
 => WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 24)                                                                                                                            
 => WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 36)                                                                                                                            
 => WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 47)                                                                                                                            
 => WARN: InvalidDefaultArgInFrom: Default value for ARG ${BUILD_ENV} results in empty or invalid base image name (line 47) 
```

PR #5928, opened 2025-07-04
HIVE-24419: Refactor junit database rules to exploit testcontainers

zabetak:
1. Use testcontainers library to manage the lifecycle of dockerized databases
2. Use the newest version of the library (1.21.3) everywhere

1. Reduce maintenance burden since we no longer have to manually craft and fix docker handling logic
2. Stabilize CI and by avoiding common problems as port conflicts, container leaks, etc., which are handled automatically by the library
3.  Enhance OS independence since we no longer call Runtime.exec methods. The underlying docker-java library takes care of using the appropriate APIs (REST, etc.)

No

```
mvn test -Dtest=TestMssql,TestDerby,TestMysql,TestOracle,TestPostgres -Dtest.groups= -Dverbose.schematool
```

### DIFF
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/dbinstall/rules/MetastoreRuleFactory.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/dbinstall/rules/MetastoreRuleFactory.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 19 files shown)


standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/dbinstall/rules/MetastoreRuleFactory.java
```diff
@@ -0,0 +1,48 @@
...
+   * @throws IllegalArgumentException if the database type is unsupported
+   */
+  public static DatabaseRule create(String dbType) {
+    return switch (dbType.toLowerCase()) {
+      case "mysql" -> new Mysql();
+      case "postgres" -> new Postgres();
+      case "postgres.tpcds" -> new PostgresTPCDS();
+      case "mariadb" -> new Mariadb();
+      case "derby" -> new Derby();
+      case "derby.clean" -> new Derby(true);
```
deniskuzZ, line 42: why do we need  `purgeInAfter` only for Derby?

zabetak: In some places (e.g., in [QTestMetaStoreHandler](https://github.com/apache/hive/blob/93c765df0272523d5d70614b8ff422f1f2f2168a/itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestMetaStoreHandler.java#L100)) there are some Derby specific optimizations to only truncate some tables instead of deleting the entire database for performance reasons. 

Implementing similar functionality for other DBMS may be useful but out-of-scope of the current PR.

zabetak:
@soumyakanti3578 Yes, we don't use containers for Derby because it runs in the same JVM with the tests for performance reasons. I am not sure if it makes sense to isolate Derby in its own container but this is a much bigger discussion that probably deserves its own ticket/PR.

zabetak:
Thanks for the reviews @soumyakanti3578 @deniskuzZ !

PR #5929, opened 2025-07-04
HIVE-28578: Concurrency issue in updateTableColumnStatistics

dengzhhu653:
No

Tested using the Repro.java against MySQL and Postgres, I didn't see the issue any more.


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -11448,36 +11454,34 @@ public List<WriteEventInfo> getAllWriteEventInfo(long txnId, String dbName, Stri
...
-      try (Statement statement = ((Connection) jdoConn.getNativeConnection()).createStatement()) {
-        statement.execute(s);
-      } finally {
-        jdoConn.close();
-      }
-    }
-  }
-
-  private void lockNotificationSequenceForUpdate() throws MetaException {
+  private void lockForUpdate(String tableName, String column, String rowFilter)
```
deniskuzZ: Wouldn't it be better with `Optional`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -9517,6 +9517,9 @@ public Map<String, String> updateTableColumnStatistics(ColumnStatistics colStats
       // So let's not use them anywhere unless absolutely necessary.
       String catName = statsDesc.isSetCatName() ? statsDesc.getCatName() : getDefaultCatalog(conf);
       MTable mTable = ensureGetMTable(catName, statsDesc.getDbName(), statsDesc.getTableName());
+      lockForUpdate("TBLS", "TBL_ID", "\"TBL_ID\" = " + mTable.getId());
```
deniskuzZ: Why do we need a lock on "TBLS" when updating stats? Why not lock on `TAB_COL_STATS`?

deniskuzZ:
There was another PR for the same: https://github.com/apache/hive/pull/5567
cc @InvisibleProgrammer 
I lean toward using retries rather than introducing locking. Non-blocking, better throughput under normal conditions.

````
SqlRetryHandler sqlRetryHandler = new SqlRetryHandler(conf, dbType);
...
return sqlRetryHandler.executeWithRetry(
    new SqlRetryCallProperties()
        .withCallerId("updateTableColumnStatistics")
        .withRetryOnDuplicateKey(true),
    () -> updateTableColumnStatisticsInternal(
            colStats, validWriteIds, writeId, catName, dbName, tableName));
````
And since HMS provides `RetryingMetaStoreClient`, this might be redundant.

-1 on the locking approach

**Comment by dengzhhu653** [comment]
> Why locking approach isn't good?
> Retries might bring another problem: it might cause the param key `COLUMN_STATS_ACCURATE` inaccurate or missing the marker for some columns, for example:
>   step 1: client1 has col1 updated, and set `COLUMN_STATS_ACCURATE` to `col1: true`
>   step 2: client2 for col2 and client3 for col3 both throw the duplicate key exception, then retry...
>   step 3: client2 and client 3 both read `COLUMN_STATS_ACCURATE` as `col1: true`, then update the param key separately.
>  step 4:  After step 3 either `col2 : true` or `col3: true` is missing in `COLUMN_STATS_ACCURATE`.

We can use Optimistic Locking instead - higher throughput and concurrency
````
UPDATE TBS
SET last_updated = NOW()
WHERE TBL_ID = <tblId> AND last_updated = '<timestamp>'; -- The exact timestamp you originally read
````
and retry if update affects 0 rows

The same strategy is used in `NextCompactionFunction`

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java (hunk 2 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java]: In this issue we have concurrency updates on the table params, locking on the `TAB_COL_STATS` (`TBL_ID`, `COLUMN_NAME`) is of no use, we still have the same issue when updating multi columns stats in parallel.
And the `TAB_COL_STATS` (`TBL_ID`) might not exist, so every client can take over the lock and proceed in this case, resulting to the same issue.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java]: In that scenario, retrying seems like a better approach than placing a global lock on the commonly accessed `TBLS`.

**Comment by dengzhhu653** [comment]
> UPDATE TBS
SET last_updated = NOW()
WHERE TBL_ID = <tblId> AND last_updated = '<timestamp>'; -- The exact timestamp you originally read

As far as I know in MySQL this will block others for the same row(the same for update), if the 0 row returns, means we have acquired the X `tblId` row lock, why do we need to retry again?

But this inspires me we can alter the table during the updating the column statistics, so we don't ask for the row lock 
explicitly, I will try and test this approach. Thank you!

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java (hunk 2 above)
Comment (dengzhhu653) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java]: As far as I know it won't prevent others from reading the whole `TBLS`, the cases affected in HMS are dropping/altering this `TBL_ID` table, or dropping the database where this table sits.

**Comment by dengzhhu653** [comment]
> > UPDATE TBS
> > SET last_updated = NOW()
> > WHERE TBL_ID =  AND last_updated = ''; -- The exact timestamp you originally read
> 
> As far as I know in MySQL this will block others for the same row(the same for update), if the 0 row returns, means we have acquired the X `tblId` row lock, why do we need to retry again?
> 
> But this inspires me we can alter the table during the updating the column statistics, so we don't ask for the row lock explicitly, I will try and test this approach. Thank you!

After some research, I find it's hard to use DataNucleus without the pure "UPDATE" query,
```java
      mTable.setLastAccessTime((int) (System.currentTimeMillis()/1000));
      pm.flush(); // here might flush the old MTable into data store
      pm.refresh(mTable);
```
In `pm.flush()` the old state of the table can get overwritten the one in data store, resulting to some columns missing in COLUMN_STATS_ACCURATE, for example:
`{"COLUMN_STATS":{"col_0":"true","col_1":"true","col_2":"true","col_3":"true","col_5":"true","col_6":"true","col_7":"true","col_8":"true"}}`
`col_4` and `col_9` are missing.

deniskuzZ:
please give me some time to review and reply

dengzhhu653:
@nrg4878 @saihemanth-cloudera @zhangbutao @wecharyu could you check this PR as well please?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
saihemanth-cloudera: directSql.lockDbTable("TBLS") => MetaStoreDirectSql#lockDbTable() resolves to lock TBLS in exclusive mode, which prevents other users from reading or writing to the TBLS table. We'll have to pass a boolean argument to MetaStoreDirectSql#lockDbTable() to decide whether to lock the table exclusively or not.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
saihemanth-cloudera: +1

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -9621,6 +9624,9 @@ public Map<String, String> updatePartitionColumnStatistics(Table table, MTable m
       if (partition == null) {
         throw new NoSuchObjectException("Partition for which stats is gathered doesn't exist.");
       }
+      lockForUpdate("PARTITIONS", "PART_ID", "\"TBL_ID\" = " + mTable.getId()
```
saihemanth-cloudera: I thin, Only "NOTIFICATION_SEQUENCE" table should be exclusively locked and "TBLS", "PARTITIONS" tables can have Share/Row exclusive

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
dengzhhu653: no, this `lockForUpdate`, effectively `select "TBL_ID" from "TBLS" where  \"TBL_ID\" = " + mTable.getId() for update`, only places an exclusive lock on the matching row: `\"TBL_ID\" = " + mTable.getId()`, it won't prevent any read or the update on other rows.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
dengzhhu653: This places an exclusive lock on the selected row: `select "PART_ID" from "PARTITIONS" where "TBL_ID" = " + mTable.getId() for update`

saihemanth-cloudera: lockForUpdate() is calling directSql.lockDbTable(tableName)#L11464; and MetaStoreDirectSql#lockDbTable(String tableName) has this code
`String lockCommand = "lock table \"" + tableName + "\" in exclusive mode";
    try {
      executeNoResult(lockCommand);
    } catch (SQLException sqle) {
      throw new MetaException("Error while locking table " + tableName + ": " + sqle.getMessage());
    }`
So we are indeed locking the table in exclusive mode, right?

dengzhhu653: this is for derby only `sqlGenerator.getDbProduct().isDERBY()`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -9605,23 +9608,28 @@ public Map<String, String> updatePartitionColumnStatistics(Table table, MTable m
...
+      List<Long> partitionIds = directSql.getPartitionFieldsViaSqlFilter(catName, statsDesc.getDbName(), statsDesc.getTableName(),
+          Arrays.asList("\"PART_ID\""), "\"PARTITIONS\".\"PART_NAME\" = ?",
+          Arrays.asList(Warehouse.makePartName(table.getPartitionKeys(), partVals)), Collections.emptyList(), -1);
+      if (partition == null || partitionIds.isEmpty()) {
         throw new NoSuchObjectException("Partition for which stats is gathered doesn't exist.");
       }
 
+      Map<String, MPartitionColumnStatistics> oldStats = getPartitionColStats(table, statsDesc
+          .getPartName(), colNames, colStats.getEngine());
+      lockForUpdate("PARTITIONS", "PART_ID", Optional.of("\"PART_ID\" = " + partitionIds.getFirst()));
```
dengzhhu653: change it to look up the partition by primary key

deniskuzZ (changes requested):
-1

> As far as I know in MySQL this will block others for the same row(the same for update), if the 0 row returns, means we have acquired the X tblId row lock, why do we need to retry again?

Using retries instead of blocking aligns better with optimistic concurrency control, allowing transactions to proceed without waiting and only retry if a conflict actually occurs.

dengzhhu653:
> -1
> 
> > As far as I know in MySQL this will block others for the same row(the same for update), if the 0 row returns, means we have acquired the X tblId row lock, why do we need to retry again?
> 
> Using retries instead of blocking aligns better with optimistic concurrency control, allowing transactions to proceed without waiting and only retry if a conflict actually occurs.
> 
> The same approach is applied in several areas, including Iceberg’s conflict resolution ([HIVE-26882](https://issues.apache.org/jira/browse/HIVE-26882)) and compaction.

I don't think retry is the good way, it introduces another issue, the `COLUMN_STATS` in `TABLE_PARAMS` would be got overwritten, and lead to some column markers missing, so it might make the stats import useless.
For updating the stats, we have already had the in-process lock, it doesn't solve the parallel imports across the HMS instances, somehow we need the distributed lock to achieve it, the DB lock is one way.

dengzhhu653:
For optimistic concurrency control, IMO this is more suitable for read skew scenario, otherwise it's a waste of the resource to retry on conflict

deniskuzZ:
> I don't think retry is the good way, it introduces another issue, the `COLUMN_STATS` in `TABLE_PARAMS` would be got overwritten, and lead to some column markers missing, so it might make the stats import useless. For updating the stats, we have already had the in-process lock, it doesn't solve the parallel imports across the HMS instances, somehow we need the distributed lock to achieve it, the DB lock is one way.

Locking TBLS and PARTITIONS isn’t necessarily a better solution. In large-scale, highly concurrent systems,  locking often becomes a scalability bottleneck. Instead, most systems adopt optimistic locking strategies. Each operation assumes no conflict and only validates at commit, reducing contention and avoiding blocking.

The idea is to balance consistency with throughput. Extending this principle to stats import (which btw is not a core functionality) we could consider a mechanism based on versioning rather than traditional locking. Such an approach generally scales far better in multi-instance HMS environments than depending solely on table or partition locks.

dengzhhu653:
> > I don't think retry is the good way, it introduces another issue, the `COLUMN_STATS` in `TABLE_PARAMS` would be got overwritten, and lead to some column markers missing, so it might make the stats import useless. For updating the stats, we have already had the in-process lock, it doesn't solve the parallel imports across the HMS instances, somehow we need the distributed lock to achieve it, the DB lock is one way.
> 
> Locking TBLS and PARTITIONS isn’t necessarily a better solution. In large-scale, highly concurrent systems, locking often becomes a scalability bottleneck. Instead, most systems adopt optimistic locking strategies. Each operation assumes no conflict and only validates at commit, reducing contention and avoiding blocking.
> 
> The idea is to balance consistency with throughput. Extending this principle to stats import (which btw is not a core functionality) we could consider a mechanism based on versioning rather than traditional locking. Such an approach generally scales far better in multi-instance HMS environments than depending solely on table or partition locks.

Currently the HMS is lack of versioning, the RDMS nowadays provides MVCC for writes not block any reads, so the write is the main concern here.
The lock provides here is the row-level lock, it only blocks the write operation against the same table, such as the alter/drop the same table/partition, drop the db, or update the stats of this table/partition.

Even without the explicit lock, we still have some exclusive row lock under the scenes, the RDMS in background ensures strong consistency and reliability the HMS benefits from and relies on, and provides a milliseconds to seconds query execution time, that can also have a satisfied throughput.

deniskuzZ:
> Currently the HMS is lack of versioning, the RDMS nowadays provides MVCC for writes not block any reads, so the write is the main concern here. 

MVCC (Multi-Version Concurrency Control) and SELECT ... FOR UPDATE (S4U) operate at completely different conceptual levels — they’re not comparable mechanisms even though both relate to concurrency.

-- MVCC style (optimistic)
BEGIN;
SELECT balance FROM accounts WHERE id = 1;  -- snapshot read
-- some processing
UPDATE accounts SET balance = balance - 100 WHERE id = 1;
COMMIT;
-- If another transaction modified same row → commit fails (conflict)

-- Pessimistic locking
BEGIN;
SELECT balance FROM accounts WHERE id = 1 FOR UPDATE;
-- locks row immediately
UPDATE accounts SET balance = balance - 100 WHERE id = 1;
COMMIT;


> The lock provides here is the row-level lock, it only blocks the write operation against the same table, such as the alter/drop the same table/partition, drop the db, or update the stats of this table/partition.

Every iceberg table commits involves alter table operation and it's non-blocking ATM.

> Even without the explicit lock, we still have some exclusive row lock under the scenes, the RDMS in background ensures strong consistency and reliability the HMS benefits from and relies on, and provides a milliseconds to seconds query execution time, that can also have a satisfied throughput.

That’s the opposite of what CU experienced on a highly loaded MySQL cluster with S4U on the NEXT_TXN_ID table.

dengzhhu653:
> -- MVCC style (optimistic)
BEGIN;
SELECT balance FROM accounts WHERE id = 1; -- snapshot read
-- some processing
UPDATE accounts SET balance = balance - 100 WHERE id = 1;
COMMIT;
-- If another transaction modified same row → commit fails (conflict)

Here we still obtained a row-level lock on `accounts(id=1)` in MySQL/Postgres until the transaction is ended.

As I explained, this would cause lost-update, 
After some research, I find it's hard to use DataNucleus without the pure "UPDATE" query,

      mTable.setLastAccessTime((int) (System.currentTimeMillis()/1000));
      pm.flush(); // here might flush the old MTable into data store
      pm.refresh(mTable);
In pm.flush() the old state of the table can get overwritten the one in data store, resulting to some columns missing in COLUMN_STATS_ACCURATE, for example:
{"COLUMN_STATS":{"col_0":"true","col_1":"true","col_2":"true","col_3":"true","col_5":"true","col_6":"true","col_7":"true","col_8":"true"}}
col_4 and col_9 are missing.

> Every iceberg table commits involves alter table operation and it's non-blocking ATM.

The Iceberg commit relies on the DB transaction atomicity, it should involve the row lock behind the scenes, though the lock is quite small(TBL_ID, PARAM_KEY), if the table has multiple commits at the same time, only one is allowed to alter the `TABLE_PARAMS`.

> That’s the opposite of what CU experienced on a highly loaded MySQL cluster with S4U on the NEXT_TXN_ID table.

This is due to every HMS request needs to lock the only one `NEXT_TXN_ID`,  compared the same level lock distributed among different tables and affects the alter/drop request of the target table/partition

dengzhhu653:
I guess I'm a bit get your optimistic concurrency control, let me try

dengzhhu653:
Filed another PR: https://github.com/apache/hive/pull/6159 to catch the nowait for update

PR #5930, opened 2025-07-04
HIVE-29060: HivePreparedStatements failed with regex column specification

Neer393:
Currently, regex column specification which includes ( ? ) fails with Hive Prepared Statements. The changes proposed fixes the issue
For eg:- Consider the query  ```select `(id)?.` from tb1 where a=?``` , when the user sets the 1st and the only parameter in the Hive Prepared Query to a value say 'asd', the expected behaviour was the query formed should be ```select `(id)?.` from tb1 where a='asd'``` and using the regex, return matching columns but the query fails saying ```parameter #2 is unset``` . The reason this happens is because Hive Prepared Statement injects the provided parameter into  ( ? ) of the column regex considering it as a parameter which is ideally not.

So that column regex specification can work with Hive Prepared Statements

Now the user will be able to query using Hive Prepared Statements along with regex column specification.

Tested it on my local machine and added a test which was run pre and post changes to verify the fix.

### DIFF
--- a/jdbc/src/java/org/apache/hive/jdbc/HivePreparedStatement.java
+++ b/jdbc/src/java/org/apache/hive/jdbc/HivePreparedStatement.java
@@ -141,6 +141,7 @@ public class HivePreparedStatement extends HiveStatement implements PreparedStat
   private List<String> splitSqlStatement(String sql) {
     List<String> parts = new ArrayList<>();
     int apCount = 0;
+    int backTicksCount = 0;
     int off = 0;
     boolean skip = false;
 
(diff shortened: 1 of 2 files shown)


Neer393:
@deniskuzZ review required.

Aggarwal-Raghav:
LGTM +1 (non-binding)

deniskuzZ (approved):
+1

Neer393:
@deniskuzZ good to merge 👍

jdbc/src/java/org/apache/hive/jdbc/HivePreparedStatement.java
```diff
@@ -157,8 +158,11 @@ private List<String> splitSqlStatement(String sql) {
       case '\\':
         skip = true;
         break;
+      case '`':
+        backTicksCount++;
```
deniskuzZ, line 162: can't we combine "`" and "\\'"?

Neer393: It's better if we keep them separate because backticks are used for regex column searches mostly or when column names are reserved sql keywords.

deniskuzZ: ok

PR #5932, opened 2025-07-04
HIVE-29069: Fix Query Failure while computing Range Partition Column stats on Numeric partition column types

Indhumathi27:
The code was modified to check if a partition value is numeric using NumberUtils.isCreatable(partVal) before parsing it, instead of only filtering out the configured default partition name. This ensures that only valid numeric partition values are processed, and any non-numeric values are skipped.

If the value of hive.exec.default.partition.name is set to one value during data insertion (for example, 'somevalue') and then changed to another value (for example, 'othervalue') during querying, Hive can encounter partition directories named with non-numeric strings. The old logic would try to parse these unexpected strings as numbers, causing a NumberFormatException. The new logic prevents this error by ignoring any partition value that isn't numeric, making the code more robust to configuration changes and avoiding runtime failures.

No

New test case added

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
@@ -39,6 +39,7 @@ import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.common.StatsSetupConst;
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/stats/StatsUtils.java
```diff
@@ -647,7 +647,7 @@ private static Range getRangePartitionColumn(PartitionIterable partitions, Strin
       double max = Double.MIN_VALUE;
       for (Partition partition : partitions) {
         partVal = partition.getSpec().get(partColName);
-        if (!partVal.equals(defaultPartName)) {
+        if (NumberUtils.isCreatable(partVal)) {
```
deniskuzZ: can we merge `DECIMAL_TYPE` with `DOUBLE_TYPE` handling ? The use of BigDecimal followed by a cast to double seems unnecessary unless preserving precision is required

Indhumathi27: In this case, Directly converting to double should work. made changes

deniskuzZ: thanks!

deniskuzZ (approved):
+1, pending tests

zhangbutao:
@deniskuzZ CI is green. Can we merge this change?

PR #5934, opened 2025-07-06
HIVE-27944: When HIVE-LLAP reads the ICEBERG table, a deadlock may occur.

BsoBird:

Modified the modifier of the cache property in the SplitGroup class to make it non-static.

JIRA: https://issues.apache.org/jira/browse/HIVE-27944

Since the Properties object implement HashTable interface, all the methods of the HashTable interface are synchronised.
In a multi-threaded environment, a deadlock will occur when propA.equals(propB) and propB.equals(propA) occur at the same time.
The problem can be solved with minimal modification by changing SplitGroup.cache to a non-static property.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
@@ -29,11 +29,10 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.ConcurrentHashMap;
 
+import com.google.common.collect.Maps;
 import org.apache.hadoop.hive.ql.io.AcidUtils;
 import org.apache.hadoop.hive.ql.io.BucketizedHiveInputFormat;
-import org.apache.hadoop.mapred.InputFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.conf.Configuration;
@@ -66,15 +65,10 @@ import com.google.common.collect.Multimap;
  * also enforces restrictions around schema, file format and bucketing.
  */
 public class SplitGrouper {
-
   private static final Logger LOG = LoggerFactory.getLogger(SplitGrouper.class);
 
-  // TODO This needs to be looked at. Map of Map to Map... Made concurrent for now since split generation
-  // can happen in parallel.
-  private static final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
-      new ConcurrentHashMap<>();
-
   private final TezMapredSplitsGrouper tezGrouper = new TezMapredSplitsGrouper();
+  private final Map<Path, Path> cache = Maps.newHashMap();
 
   /**
    * group splits for each bucket separately - while evenly filling all the
@@ -91,7 +85,7 @@ public class SplitGrouper {
 
     // allocate map bucket id to grouped splits
     Multimap<Integer, InputSplit> bucketGroupedSplitMultimap =
-        ArrayListMultimap.<Integer, InputSplit> create();
+        ArrayListMultimap.create();
 
     // use the tez grouper to combine splits once per bucket
     for (int bucketId : bucketSplitMultimap.keySet()) {
@@ -137,9 +131,8 @@ public class SplitGrouper {
         String [] locations = split.getLocations();
         if (locations != null && locations.length > 0) {
           // Worthwhile only if more than 1 split, consistentGroupingEnabled and is a FileSplit
-          if (consistentLocations && locations.length > 1 && split instanceof FileSplit) {
+          if (consistentLocations && locations.length > 1 && split instanceof FileSplit fileSplit) {
             Arrays.sort(locations);
-            FileSplit fileSplit = (FileSplit) split;
             Path path = fileSplit.getPath();
             long startLocation = fileSplit.getStart();
             int hashCode = Objects.hash(path, startLocation);
@@ -153,8 +146,8 @@ public class SplitGrouper {
             locationHints.add(TaskLocationHint.createTaskLocationHint(locationSet, null));
           } else {
             locationHints.add(TaskLocationHint
-                .createTaskLocationHint(new LinkedHashSet<String>(Arrays.asList(split
-                    .getLocations())), null));
+                .createTaskLocationHint(new LinkedHashSet<>(Arrays.asList(split
+                  .getLocations())), null));
           }
         } else {
           locationHints.add(TaskLocationHint.createTaskLocationHint(null, null));
@@ -189,16 +182,15 @@ public class SplitGrouper {
     boolean isMinorCompaction = true;
     MapWork mapWork = populateMapWork(jobConf, inputName);
     // ArrayListMultimap is important here to retain the ordering for the splits.
-    Multimap<Integer, InputSplit> schemaGroupedSplitMultiMap = ArrayListMultimap.<Integer, InputSplit> create();
+    Multimap<Integer, InputSplit> schemaGroupedSplitMultiMap = ArrayListMultimap.create();
     if (HiveConf.getVar(jobConf, HiveConf.ConfVars.SPLIT_GROUPING_MODE).equalsIgnoreCase("compactor")) {
       List<Path> paths = Utilities.getInputPathsTez(jobConf, mapWork);
       for (Path path : paths) {
         List<String> aliases = mapWork.getPathToAliases().get(path);
         if ((aliases != null) && (aliases.size() == 1)) {
-          Operator<? extends OperatorDesc> op = mapWork.getAliasToWork().get(aliases.get(0));
-          if ((op != null) && (op instanceof TableScanOperator)) {
-            TableScanOperator tableScan = (TableScanOperator) op;
-            PartitionDesc partitionDesc = mapWork.getAliasToPartnInfo().get(aliases.get(0));
+          Operator<? extends OperatorDesc> op = mapWork.getAliasToWork().get(aliases.getFirst());
+          if (op instanceof TableScanOperator tableScan) {
+            PartitionDesc partitionDesc = mapWork.getAliasToPartnInfo().get(aliases.getFirst());
             isMinorCompaction &= AcidUtils.isCompactionTable(partitionDesc.getTableDesc().getProperties());
             if (!tableScan.getConf().isTranscationalTable() && !isMinorCompaction) {
               String splitPath = getFirstSplitPath(splits);
@@ -260,7 +252,7 @@ public class SplitGrouper {
   Multimap<Integer, InputSplit> getCompactorSplitGroups(InputSplit[] rawSplits, Configuration conf,
       boolean isMinorCompaction) {
     // Note: For our case, this multimap will essentially contain one value (one TezGroupedSplit) per key 
-    Multimap<Integer, InputSplit> bucketSplitMultiMap = ArrayListMultimap.<Integer, InputSplit> create();
+    Multimap<Integer, InputSplit> bucketSplitMultiMap = ArrayListMultimap.create();
     HiveInputFormat.HiveInputSplit[] splits = new HiveInputFormat.HiveInputSplit[rawSplits.length];
     int i = 0;
     for (InputSplit is : rawSplits) {
@@ -345,10 +337,10 @@ public class SplitGrouper {
                                                     Map<Integer, Collection<InputSplit>> bucketSplitMap) {
 
     // mapping of bucket id to size of all splits in bucket in bytes
-    Map<Integer, Long> bucketSizeMap = new HashMap<Integer, Long>();
+    Map<Integer, Long> bucketSizeMap = new HashMap<>();
 
     // mapping of bucket id to number of required tasks to run
-    Map<Integer, Integer> bucketTaskMap = new HashMap<Integer, Integer>();
+    Map<Integer, Integer> bucketTaskMap = new HashMap<>();
 
     // TODO HIVE-12255. Make use of SplitSizeEstimator.
     // The actual task computation needs to be looked at as well.
@@ -362,12 +354,11 @@ public class SplitGrouper {
         // the case of SMB join. So in this case, we can do an early exit by not doing the
         // calculation for bucketSizeMap. Each bucket will assume it can fill availableSlots * waves
         // (preset to 0.5) for SMB join.
-        if (!(s instanceof FileSplit)) {
+        if (!(s instanceof FileSplit fsplit)) {
           bucketTaskMap.put(bucketId, (int) (availableSlots * waves));
           earlyExit = true;
           continue;
         }
-        FileSplit fsplit = (FileSplit) s;
         size += fsplit.getLength();
         totalSize += fsplit.getLength();
       }
@@ -413,7 +404,7 @@ public class SplitGrouper {
   }
 
   private boolean schemaEvolved(InputSplit s, InputSplit prevSplit, boolean groupAcrossFiles,
-                                       MapWork work) throws IOException {
+                                     MapWork work) throws IOException {
     boolean retval = false;
     Path path = ((FileSplit) s).getPath();
     PartitionDesc pd = HiveFileFormatUtils.getFromPathRecursively(
@@ -434,7 +425,6 @@ public class SplitGrouper {
       previousDeserializerClass = prevPD.getDeserializerClassName();
       previousInputFormatClass = prevPD.getInputFileFormatClass();
     }
-
     if ((currentInputFormatClass != previousInputFormatClass)
         || (!currentDeserializerClass.equals(previousDeserializerClass))) {
       retval = true;
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/CombineHiveRecordReader.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/CombineHiveRecordReader.java
@@ -42,6 +42,8 @@ import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.mapred.Reporter;
 import org.apache.hadoop.mapred.RecordReader;
 import org.apache.hadoop.mapred.lib.CombineFileSplit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * CombineHiveRecordReader.
@@ -51,7 +53,7 @@ import org.apache.hadoop.mapred.lib.CombineFileSplit;
  */
 public class CombineHiveRecordReader<K extends WritableComparable, V extends Writable>
     extends HiveContextAwareRecordReader<K, V> {
-  private org.slf4j.Logger LOG = org.slf4j.LoggerFactory.getLogger(CombineHiveRecordReader.class);
+  private static final Logger LOG = LoggerFactory.getLogger(CombineHiveRecordReader.class);
 
   private Map<Path, PartitionDesc> pathToPartInfo;
 
@@ -104,17 +106,16 @@ public class CombineHiveRecordReader<K extends WritableComparable, V extends Wri
 
     //If current split is from the same file as preceding split and the preceding split has footerbuffer,
     //the current split should use the preceding split's footerbuffer in order to skip footer correctly.
-    if (preReader != null && preReader instanceof CombineHiveRecordReader
-        && ((CombineHiveRecordReader)preReader).getFooterBuffer() != null) {
+    if (preReader instanceof CombineHiveRecordReader
+        && ((CombineHiveRecordReader) preReader).getFooterBuffer() != null) {
       if (partition != 0 && hsplit.getPaths()[partition -1].equals(hsplit.getPaths()[partition]))
         this.setFooterBuffer(((CombineHiveRecordReader)preReader).getFooterBuffer());
     }
-
   }
 
   private PartitionDesc extractSinglePartSpec(CombineHiveInputSplit hsplit) throws IOException {
     PartitionDesc part = null;
-    Map<Map<Path,PartitionDesc>, Map<Path,PartitionDesc>> cache = new HashMap<>();
+    Map<Path, Path> cache = new HashMap<>();
     for (Path path : hsplit.getPaths()) {
       PartitionDesc otherPart = HiveFileFormatUtils.getFromPathRecursively(
           pathToPartInfo, path, cache);
@@ -147,12 +148,12 @@ public class CombineHiveRecordReader<K extends WritableComparable, V extends Wri
 
   @Override
   public K createKey() {
-    return (K) recordReader.createKey();
+    return recordReader.createKey();
   }
 
   @Override
   public V createValue() {
-    return (V) recordReader.createValue();
+    return recordReader.createValue();
   }
 
   @Override
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java
@@ -130,10 +130,6 @@ public final class HiveFileFormatUtils {
           .build();
     }
 
-    public Set<Class<? extends InputFormat>> registeredClasses() {
-      return inputFormatCheckerMap.keySet();
-    }
-
     public Set<Class<? extends InputFormat>> registeredTextClasses() {
       return textInputFormatCheckerMap.keySet();
     }
@@ -179,7 +175,6 @@ public final class HiveFileFormatUtils {
    *
    * Note: an empty set of files is considered compliant.
    */
-  @SuppressWarnings("unchecked")
   public static boolean checkInputFormat(FileSystem fs, HiveConf conf,
       Class<? extends InputFormat> inputFormatCls, List<FileStatus> files)
       throws HiveException {
@@ -202,7 +197,7 @@ public final class HiveFileFormatUtils {
           .getInputFormatCheckerInstance(checkerCls);
       try {
         if (checkerInstance == null) {
-          checkerInstance = checkerCls.newInstance();
+          checkerInstance = checkerCls.getDeclaredConstructor().newInstance();
           FileChecker.getInstance().putInputFormatCheckerInstance(checkerCls, checkerInstance);
         }
         return checkerInstance.validateInput(fs, conf, files);
@@ -213,7 +208,6 @@ public final class HiveFileFormatUtils {
     return true;
   }
 
-  @SuppressWarnings("unchecked")
   private static boolean checkTextInputFormat(FileSystem fs, HiveConf conf,
       List<FileStatus> files) throws HiveException {
     List<FileStatus> files2 = new LinkedList<>(files);
@@ -270,13 +264,13 @@ public final class HiveFileFormatUtils {
       if (isCompressed) {
         jc_output = new JobConf(jc);
         String codecStr = conf.getCompressCodec();
-        if (codecStr != null && !codecStr.trim().equals("")) {
+        if (codecStr != null && !codecStr.trim().isEmpty()) {
           Class<? extends CompressionCodec> codec = 
               JavaUtils.loadClass(codecStr);
           FileOutputFormat.setOutputCompressorClass(jc_output, codec);
         }
         String type = conf.getCompressType();
-        if (type != null && !type.trim().equals("")) {
+        if (type != null && !type.trim().isEmpty()) {
           CompressionType style = CompressionType.valueOf(type);
           SequenceFileOutputFormat.setOutputCompressionType(jc, style);
         }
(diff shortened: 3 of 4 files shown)


deniskuzZ (dismissed):
+1, pending tests

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
```diff
@@ -71,8 +70,8 @@ public class SplitGrouper {
 
   // TODO This needs to be looked at. Map of Map to Map... Made concurrent for now since split generation
   // can happen in parallel.
-  private static final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
-      new ConcurrentHashMap<>();
+  private final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
```
deniskuzZ, line 73: please replace with Multimap
````
private final Multimap<Map<Path, PartitionDesc>, PartitionDesc> cache =
    ArrayListMultimap.create();
````

Comment (deniskuzZ) [inline ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @BsoBird, can we refactor with MultiMap?
````
Subject: [PATCH] refactor
---
Index: ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java b/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java	(revision ade51d810b383dd68874ecaa275c3a6cc9157884)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java	(date 1751805392461)
@@ -23,6 +23,7 @@
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -32,6 +33,7 @@
 import java.util.Properties;
 import java.util.Set;
 
+import org.apache.commons.collections.MapUtils;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
@@ -72,6 +74,7 @@
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Multimap;
 
 /**
  * An util class for various Hive file format tasks.
@@ -360,17 +363,17 @@
   }
 
   public static <T> T getFromPathRecursively(Map<Path, T> pathToPartitionInfo, Path dir,
-      Map<Map<Path, T>, Map<Path, T>> cacheMap) throws IOException {
+      Multimap<Map<Path, T>, Path> cacheMap) throws IOException {
     return getFromPathRecursively(pathToPartitionInfo, dir, cacheMap, false);
   }
 
   public static <T> T getFromPathRecursively(Map<Path, T> pathToPartitionInfo, Path dir,
-      Map<Map<Path, T>, Map<Path, T>> cacheMap, boolean ignoreSchema) throws IOException {
+      Multimap<Map<Path, T>, Path> cacheMap, boolean ignoreSchema) throws IOException {
     return getFromPathRecursively(pathToPartitionInfo, dir, cacheMap, ignoreSchema, false);
   }
 
   public static <T> T getFromPathRecursively(Map<Path, T> pathToPartitionInfo, Path dir,
-      Map<Map<Path, T>, Map<Path, T>> cacheMap, boolean ignoreSchema, boolean ifPresent)
+      Multimap<Map<Path, T>, Path> cacheMap, boolean ignoreSchema, boolean ifPresent)
           throws IOException {
     T part = getFromPath(pathToPartitionInfo, dir);
 
@@ -381,14 +384,19 @@
 
       Map<Path, T> newPathToPartitionInfo = null;
       if (cacheMap != null) {
-        newPathToPartitionInfo = cacheMap.get(pathToPartitionInfo);
+        Collection<Path> keys = cacheMap.get(pathToPartitionInfo);
+        if (keys != null) {
+          newPathToPartitionInfo =
+              populateNewT(keys, pathToPartitionInfo.values());
+        }
       }
 
-      if (newPathToPartitionInfo == null) { // still null
-        newPathToPartitionInfo = populateNewT(pathToPartitionInfo);
+      if (MapUtils.isEmpty(newPathToPartitionInfo)) { // still null
+        newPathToPartitionInfo =
+            populateNewT(pathToPartitionInfo.keySet(), pathToPartitionInfo.values());
 
         if (cacheMap != null) {
-          cacheMap.put(pathToPartitionInfo, newPathToPartitionInfo);
+          cacheMap.putAll(pathToPartitionInfo, newPathToPartitionInfo.keySet());
         }
       }
       part = getFromPath(newPathToPartitionInfo, dir);
@@ -401,12 +409,18 @@
     }
   }
 
-  private static <T> Map<Path, T> populateNewT(Map<Path, T> pathToPartitionInfo) {
+  private static <T> Map<Path, T> populateNewT(Collection<Path> keys, Collection<T> values) {
+    if (keys.size() != values.size()) {
+      throw new IllegalArgumentException("Key and value collections must have the same size.");
+    }
     Map<Path, T> newPathToPartitionInfo = new HashMap<>();
-    for (Map.Entry<Path, T> entry: pathToPartitionInfo.entrySet()) {
-      T partDesc = entry.getValue();
-      Path pathOnly = Path.getPathWithoutSchemeAndAuthority(entry.getKey());
-      newPathToPartitionInfo.put(pathOnly, partDesc);
+    Iterator<Path> keyIterator = keys.iterator();
+    Iterator<T> valueIterator = values.iterator();
+
+    while (keyIterator.hasNext() && valueIterator.hasNext()) {
+      Path key = Path.getPathWithoutSchemeAndAuthority(
+          keyIterator.next());
+      newPathToPartitionInfo.put(key, valueIterator.next());
     }
     return newPathToPartitionInfo;
   }
Index: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java	(revision ade51d810b383dd68874ecaa275c3a6cc9157884)
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java	(date 1751800901858)
@@ -71,8 +71,8 @@
 
   // TODO This needs to be looked at. Map of Map to Map... Made concurrent for now since split generation
   // can happen in parallel.
-  private static final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
-      new ConcurrentHashMap<>();
+  private final Multimap<Map<Path, PartitionDesc>, Path> cache =
+      ArrayListMultimap.create();
 
   private final TezMapredSplitsGrouper tezGrouper = new TezMapredSplitsGrouper();

````

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @deniskuzZ 
At present, only one code is involved:
```
// IOPrepareCache.get().allocatePartitionDescMap()
public Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> allocatePartitionDescMap() {
    if (partitionDescMap == null) {
      partitionDescMap = new HashMap<>();
    }
    return partitionDescMap;
  }
```

usage demo:
```
        // extract all the inputFormatClass names for each chunk in the
        // CombinedSplit.
        PartitionDesc part = HiveFileFormatUtils.getFromPathRecursively(pathToPartitionInfo,
            inputSplitShim.getPath(0), IOPrepareCache.get().getPartitionDescMap());
```

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @deniskuzZ It would be even better if you could explain why we need to use multiMap.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: Maybe we don’t need it. Are you suggesting that Map<Path, Path> would be sufficient?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: As I understand it, the goal here is simply to save on calls to `Path.getPathWithoutSchemeAndAuthority`

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @deniskuzZ 
`Multimap<Map<Path, PartitionDesc>, Path> cache`
In what scenarios would we use PartitionDesc in cache?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: my guess was that we wanted to keep the cache based on PartitionDesc in case it changes

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: updated
````
From c62bbba350e9168d5bab4e727f1f5e18ff8fc26d Mon Sep 17 00:00:00 2001
From: Denys Kuzmenko <deniskuzz@gmail.com>
Date: Sun, 6 Jul 2025 15:40:18 +0300
Subject: [PATCH 1/1] test

---
 .../hadoop/hive/ql/exec/tez/SplitGrouper.java |  5 +--
 .../hive/ql/io/HiveFileFormatUtils.java       | 39 +++++++++----------
 2 files changed, 19 insertions(+), 25 deletions(-)

diff --git a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
index 2ce7ed88c8..f867c251a8 100644
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
@@ -69,10 +69,7 @@ public class SplitGrouper {
 
   private static final Logger LOG = LoggerFactory.getLogger(SplitGrouper.class);
 
-  // TODO This needs to be looked at. Map of Map to Map... Made concurrent for now since split generation
-  // can happen in parallel.
-  private static final Map<Map<Path, PartitionDesc>, Map<Path, PartitionDesc>> cache =
-      new ConcurrentHashMap<>();
+  private final Map<Path, Path> cache = new ConcurrentHashMap<>();
 
   private final TezMapredSplitsGrouper tezGrouper = new TezMapredSplitsGrouper();
 
diff --git a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java b/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java
index 20fde98c26..644ce9a3fd 100644
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/HiveFileFormatUtils.java
@@ -360,17 +360,17 @@ private static RecordUpdater getRecordUpdater(JobConf jc,
   }
 
   public static <T> T getFromPathRecursively(Map<Path, T> pathToPartitionInfo, Path dir,
-      Map<Map<Path, T>, Map<Path, T>> cacheMap) throws IOException {
+      Map<Path, Path> cacheMap) throws IOException {
     return getFromPathRecursively(pathToPartitionInfo, dir, cacheMap, false);
   }
 
   public static <T> T getFromPathRecursively(Map<Path, T> pathToPartitionInfo, Path dir,
-      Map<Map<Path, T>, Map<Path, T>> cacheMap, boolean ignoreSchema) throws IOException {
+      Map<Path, Path> cacheMap, boolean ignoreSchema) throws IOException {
     return getFromPathRecursively(pathToPartitionInfo, dir, cacheMap, ignoreSchema, false);
   }
 
   public static <T> T getFromPathRecursively(Map<Path, T> pathToPartitionInfo, Path dir,
-      Map<Map<Path, T>, Map<Path, T>> cacheMap, boolean ignoreSchema, boolean ifPresent)
+      Map<Path, Path> cacheMap, boolean ignoreSchema, boolean ifPresent)
           throws IOException {
     T part = getFromPath(pathToPartitionInfo, dir);
 
@@ -379,18 +379,7 @@ public static <T> T getFromPathRecursively(Map<Path, T> pathToPartitionInfo, Pat
             || (dir.toUri().getScheme() == null || dir.toUri().getScheme().trim().equals(""))
             || FileUtils.pathsContainNoScheme(pathToPartitionInfo.keySet()))) {
 
-      Map<Path, T> newPathToPartitionInfo = null;
-      if (cacheMap != null) {
-        newPathToPartitionInfo = cacheMap.get(pathToPartitionInfo);
-      }
-
-      if (newPathToPartitionInfo == null) { // still null
-        newPathToPartitionInfo = populateNewT(pathToPartitionInfo);
-
-        if (cacheMap != null) {
-          cacheMap.put(pathToPartitionInfo, newPathToPartitionInfo);
-        }
-      }
+      Map<Path, T> newPathToPartitionInfo = populateNewT(pathToPartitionInfo, cacheMap);
       part = getFromPath(newPathToPartitionInfo, dir);
     }
     if (part != null || ifPresent) {
@@ -401,13 +390,21 @@ public static <T> T getFromPathRecursively(Map<Path, T> pathToPartitionInfo, Pat
     }
   }
 
-  private static <T> Map<Path, T> populateNewT(Map<Path, T> pathToPartitionInfo) {
-    Map<Path, T> newPathToPartitionInfo = new HashMap<>();
-    for (Map.Entry<Path, T> entry: pathToPartitionInfo.entrySet()) {
-      T partDesc = entry.getValue();
-      Path pathOnly = Path.getPathWithoutSchemeAndAuthority(entry.getKey());
+  private static <T> Map<Path, T> populateNewT(Map<Path, T> pathToPartitionInfo,
+      Map<Path, Path> cacheMap) {
+    Map<Path, T> newPathToPartitionInfo = new HashMap<>(pathToPartitionInfo.size());
+
+    pathToPartitionInfo.forEach((originalPath, partDesc) -> {
+      Path pathOnly = cacheMap != null ?
+        cacheMap.get(originalPath) : null;
+      if (pathOnly == null) {
+        pathOnly = Path.getPathWithoutSchemeAndAuthority(originalPath);
+        if (cacheMap != null) {
+          cacheMap.put(originalPath, pathOnly);
+        }
+      }
       newPathToPartitionInfo.put(pathOnly, partDesc);
-    }
+    });
     return newPathToPartitionInfo;
   }
 
-- 
2.48.0
````

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @BsoBird, please take a look at https://github.com/apache/hive/pull/5935. If it's ok, please update the PR

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @deniskuzZ
Sir,How about WeakHashMap?

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @deniskuzZ 
Also, I don't quite understand. The SplitGrouper object is released immediately after use. Is it necessary for us to use GuavaCache for caching?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: replaced with `IOPrepareCache.get().allocatePartitionDescMap()`

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @deniskuzZ 

> i reverted to static as it was a global cache before

So, if I do this, is it possible that it will cause splitGrouper to produce incorrect calculation results? Because no one updates it.
```
alter table xxx drop partition([pt=11);
alter table xxx add partition([pt=11) location '/p1';
alter table xxx drop partition([pt=11);
alter table xxx add partition([pt=11) location '/p2';
alter table xxx drop partition([pt=11);
alter table xxx add partition([pt=11) location '/p1';
```

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: I've been trying to verify the cache is being used through SplitGrouper, but so far without success.
`dir.toUri().getScheme()` is not null or empty

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @BsoBird, how did you reproduce this in first place? Could you please share the query and tables definitions.
I'm not trying to test for concurrency issues, I just want to trigger SplitGrouper in a way that uses the cache.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: please take a look at https://github.com/apache/hive/pull/5935/files, it has some extra checkstyle fixes. If that looks OK, please re-apply here. 
TBH, I have no idea why in a world someone came up with this `Map<Map<>,Map<>>` data structure for a normalized path map.

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: > @deniskuzZ
> 
> > i reverted to static as it was a global cache before
> 
> So, if I do this, is it possible that it will cause splitGrouper to produce incorrect calculation results? Because no one updates it.
> 
> ```
> alter table xxx drop partition([pt=11);
> alter table xxx add partition([pt=11) location '/p1';
> alter table xxx drop partition([pt=11);
> alter table xxx add partition([pt=11) location '/p2';
> alter table xxx drop partition([pt=11);
> alter table xxx add partition([pt=11) location '/p1';
> ```

i removed the global cache as i don't see how it's even being used. However, it wouldn't cause an issue since it just contains normalized path mapping and doesn't cache the `PartitionDesc`

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: > how did you reproduce this in first place? Could you please share the query and tables definitions.
> I'm not trying to test for concurrency issues, I just want to trigger SplitGrouper in a way that uses the cache.

@deniskuzZ 
Currently, my memory is a bit hazy. The necessary conditions to trigger this issue are as follows:

1.Use LLAP or a persistent Tez DAG(reuse on yarn) to read an Iceberg partition table.
2.The Iceberg table is just a regular Hadoop-catalog partitioned table.(But we may need to test the behavior when we set bucketing_version=1 or bucketing_version=2 for the Iceberg location based table..)

However, this issue is not consistently reproducible. We had been using IcebergStorageHandler for a long time without encountering this problem before.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: > please take a look at https://github.com/apache/hive/pull/5935/files, it has some extra checkstyle fixes. If that looks OK, please re-apply here. TBH, I have no idea why in a world someone came up with this `Map<Map<>,Map<>>` data structure for a normalized path map.

I will update it later.
I guess the previous author added PartitionDesc to the map-key to maintain the partition information after the user modifies the partition.
For example, after modifying the partition, the user might have changed the path information of the partition. In the original structure, we could capture this change.
However, using a map to store the data is quite strange. Maybe he just wanted to use a Tuple2<Path, PartitionDesc>?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @BsoBird, thanks for the details, it helped a lot.  i managed to create a test that initializes the cache from `SplitGroper`. 
Note, I struggled to create an iceberg table backed by HadoopCatalog

````
set iceberg.catalog.ice01.type=hadoop;
set iceberg.catalog.ice01.warehouse=/tmp;

CREATE EXTERNAL TABLE orders (orderid INT, quantity INT, itemid INT, tradets TIMESTAMP)
    PARTITIONED BY (p1 STRING, p2 STRING)
STORED BY ICEBERG STORED AS ORC
LOCATION '/tmp/ice01/orders'
TBLPROPERTIES('format-version'='2', 'iceberg.catalog'='ice01');
````
Caused by: java.lang.IllegalArgumentException: Cannot set a custom location for a path-based table. Expected /tmp/default/orders but got file:/tmp/ice01/orders

````
CREATE EXTERNAL TABLE orders (orderid INT, quantity INT, itemid INT, tradets TIMESTAMP)
    PARTITIONED BY (p1 STRING, p2 STRING)
STORED BY ICEBERG STORED AS ORC
LOCATION '/tmp/ice01/orders'
TBLPROPERTIES('format-version'='2', 'iceberg.catalog'='ice01');
````
Caused by: java.lang.IllegalArgumentException: Table location not set

Wonder how were you able to create one?

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: > I guess the previous author added PartitionDesc to the map-key to maintain the partition information after the user modifies the partition. For example, after modifying the partition, the user might have changed the path information of the partition. In the original structure, we could capture this change. However, using a map to store the data is quite strange. Maybe he just wanted to use a Tuple2<Path, PartitionDesc>?

I don't follow. If the user changes the path - he won't be able to do cache lookup and need to create new path mapping. I still don't get why PartitionDesc is needed here at all

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: !!! Currently, this cache isn't useful for Iceberg tables because SplitGrouper doesn't receive any partition information. The dir path always points to the table directory, and the same applies to the keys in pathToPartitionInfo

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @BsoBird, please ping me once you sync the PR with https://github.com/apache/hive/pull/5935 and there is a green build.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: > @BsoBird, thanks for the details, it helped a lot. i managed to create a test that initializes the cache from `SplitGroper`. Note, I struggled to create an iceberg table backed by HadoopCatalog
> 
> ```
> set iceberg.catalog.ice01.type=hadoop;
> set iceberg.catalog.ice01.warehouse=/tmp;
> 
> CREATE EXTERNAL TABLE orders (orderid INT, quantity INT, itemid INT, tradets TIMESTAMP)
>     PARTITIONED BY (p1 STRING, p2 STRING)
> STORED BY ICEBERG STORED AS ORC
> LOCATION '/tmp/ice01/orders'
> TBLPROPERTIES('format-version'='2', 'iceberg.catalog'='ice01');
> ```
> 
> Caused by: java.lang.IllegalArgumentException: Cannot set a custom location for a path-based table. Expected /tmp/default/orders but got file:/tmp/ice01/orders
> 
> ```
> CREATE EXTERNAL TABLE orders (orderid INT, quantity INT, itemid INT, tradets TIMESTAMP)
>     PARTITIONED BY (p1 STRING, p2 STRING)
> STORED BY ICEBERG STORED AS ORC
> LOCATION '/tmp/ice01/orders'
> TBLPROPERTIES('format-version'='2', 'iceberg.catalog'='ice01');
> ```
> 
> Caused by: java.lang.IllegalArgumentException: Table location not set
> 
> Wonder how were you able to create one?

@deniskuzZ 
Example:
```
----spark3.4.1+iceberg 1.4.2
CREATE TABLE IF NOT EXISTS datacenter.test.test_data_02 (
id string,name string
)
PARTITIONED BY (name,bucket(16,id))
TBLPROPERTIES ('read.orc.vectorization.enabled'='true','write.format.default'='orc','write.orc.bloom.filter.columns'='id','write.orc.compression-codec'='zstd','write.metadata.previous-versions-max'='3','write.metadata.delete-after-commit.enabled'='true')
STORED AS iceberg;

insert into datacenter.test.test_data_02(id,name) values('1','a'),('2','b'); 


--hive
 CREATE EXTERNAL TABLE iceberg_dwd.test_data_02 
 STORED BY 'org.apache.iceberg.mr.hive.HiveIcebergStorageHandler' 
LOCATION 'hdfs://xxxxxx/iceberg-catalog/warehouse/test/test_data_02'
TBLPROPERTIES ('iceberg.catalog'='location_based_table','engine.hive.enabled'='true');
```

But in reality, using Iceberg tables from other catalogs in HIVE is not always smooth. For example, if the table's location is set only using Iceberg properties without setting the HIVE table's LOCATION property, there's a high chance the table won't function properly.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: > !!! Currently, this cache isn't useful for Iceberg tables because SplitGrouper doesn't receive any partition information. The dir path always points to the table directory, and the same applies to the keys in pathToPartitionInfo

@deniskuzZ 
You mean, even if we keep PartitionDesc, we shouldn't encounter a deadlock?
That's very strange. We did experience a deadlock when reading Iceberg tables before. We never encountered this issue when reading regular ORC tables.
However, at that time we were using HIVE 3.1.3 to read ICEBERG tables. Has this issue disappeared in HIVE4?

However, no matter what, I also agree that SplitGrouper.cache should not use a strange data structure like Map<Map<>,Map<>>.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @deniskuzZ 
Sir, I have synchronized your changes, but I did not synchronize the modifications to HadoopCatalog. It may need to be split into another ISSUE.

Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @deniskuzZ 
BTW, since we're talking about HadoopCatalog, I'd like to recommend myself to show you a prototype project I previously used for research. It currently implements a reliable commit mechanism based on any file system. As HMS is trending towards being separated as a standalone metadata management tool, if it could have a built-in catalog that doesn't rely on DB or other services, but only a lightweight implementation dependent on the file system, I think many people would be willing to try it.

https://github.com/BsoBird/filesystem-catalog-original

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: @BsoBird, perhaps you'd be interested in sharing your project on the Hive dev mailing list to attract contributors who might be interested in collaborating.

**Review by deniskuzZ — APPROVED** [review]
+1, pending tests

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java (hunk 1 above)
Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: could you please share configs you've used for 'iceberg.catalog'='location_based_table'?
example:
```
set iceberg.catalog.ice01.type=hadoop;
set iceberg.catalog.ice01.warehouse=/tmp;
```

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: > You mean, even if we keep PartitionDesc, we shouldn't encounter a deadlock?

we might, that what you reproduced. I was just saying that PartitionDesc doesn't have much value for iceberg, at least right now.

**Comment by deniskuzZ** [comment]
@BsoBird, you need to rebase with master, it's on jdk21 now

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java (hunk 1 above)
Comment (BsoBird) [reply ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java]: > could you please share configs you've used for 'iceberg.catalog'='location_based_table'? example:
> 
> ```
> set iceberg.catalog.ice01.type=hadoop;
> set iceberg.catalog.ice01.warehouse=/tmp;
> ```
@deniskuzZ 
no need config,just use this:
```
CREATE EXTERNAL TABLE iceberg_dwd.test_data_02 
 STORED BY 'org.apache.iceberg.mr.hive.HiveIcebergStorageHandler' 
LOCATION 'hdfs://xxxxxx/iceberg-catalog/warehouse/test/test_data_02'
TBLPROPERTIES ('iceberg.catalog'='location_based_table','engine.hive.enabled'='true');
```

deniskuzZ: @BsoBird, that setup uses `HadoopTables` rather than `HadoopCatalog` !

- **HadoopTables**: Implementation of Iceberg tables that uses the Hadoop FileSystem to store metadata and manifests
- **HadoopCatalog**: provides a way to use table names like db.table to work with path-based tables under a common location. It uses a specified directory under a specified filesystem as the warehouse directory, and organizes multiple levels directories that mapped to the database, namespace and the table respectively.

zhangbutao:
Faild tests unrelated this PR:
`Unable to create live FilePath for hive-precommit-pr-5934-5-mhfgv-v07p8-ftfcz; hive-precommit-pr-5934-5-mhfgv-v07p8-ftfcz was marked offline: Pod failed (Reason: Evicted, Message: Pod ephemeral local storage usage exceeds the total limit of containers 20Gi. )
`

BTW, have we reached an agreement on this PR?

BsoBird:
> Faild tests unrelated this PR: `Unable to create live FilePath for hive-precommit-pr-5934-5-mhfgv-v07p8-ftfcz; hive-precommit-pr-5934-5-mhfgv-v07p8-ftfcz was marked offline: Pod failed (Reason: Evicted, Message: Pod ephemeral local storage usage exceeds the total limit of containers 20Gi. ) `
> 
> BTW, have we reached an agreement on this PR?

@zhangbutao 
i think,yes.
Because in any case, this strange caching structure in SplitGrouper should no longer be used.

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/SplitGrouper.java
BsoBird: > @BsoBird, that setup uses `HadoopTables` rather than `HadoopCatalog` !
> 
> * **HadoopTables**: Implementation of Iceberg tables that uses the Hadoop FileSystem to store metadata and manifests
> * **HadoopCatalog**: provides a way to use table names like db.table to work with path-based tables under a common location. It uses a specified directory under a specified filesystem as the warehouse directory, and organizes multiple levels directories that mapped to the database, namespace and the table respectively.

Sorry,Sir.
Because I previously reviewed the code and found that the submission and management logic of location_based_table is basically the same as that of hadoopCatalog. Therefore, I previously defaulted to treating location_based_table as HadoopCatalogTable. I confused the concepts, and this is my mistake.
Additionally, in the production environment, we currently only use HadoopTables (location_based_table) in HIVE.
(In Spark, I use hadoopCatalog to create a table. After Spark maintains this table, in HIVE I use HadoopTables to query the hadoopCatalog table maintained by Spark.)

Therefore, I only set the properties of IcebergHiveCatalog in HIVE (But I did not use HMS Catalog in the production environment, it was only used for some testing) , and did not set the related properties of other types of iceberg (including HadoopCatalog,Because HadoopTables can be used without any setup,so no need setup/use hadoopCatalog).

deniskuzZ: HadoopCatalog is generally a better choice for production environments, as it offers namespace management, table discovery, and safe concurrent access.

deniskuzZ:
re-triggered the build

deniskuzZ:
again some cluster issue

zabetak:
I just stopped the precommits since they failed due to resource limits. Make sure to rebase the PR before relaunching.

deniskuzZ:
rebased

PR #5936, opened 2025-07-07
HIVE-29062: Add standalone module for packaging the Metastore (#5923)

dengzhhu653:
Add a new module to package the standalone Metastore

On the thread https://lists.apache.org/thread/tvozv88sjjmnx8n5rktjg1c6x1y0mlpv, we agree to release the standalone Metastore and the Docker image in 4.1.0. In order to fix the circular dependency issue we build the Metastore binary tarball and the image from the module metastore-server, we introduced a profile `allModule` to include all the necessities, and the source tarball is built from the standalone metastore root. this is quite not straightforward and don't follow the way we build the Hive tarball. This new module also normalizes the Metastore binary/source tarball we are going to release, which keeps the old behavior of Hive Metastore 3.0.0.

No

Tested locally using the command `mvn clean package -Pdist,docker`


aturoczy:
Could you please fill out the description? This ticket means the Metastore will be separate entity and it could deploy separate?

dengzhhu653:
> Could you please fill out the description? This ticket means the Metastore will be separate entity and it could deploy separate?

Done, we have a standard way to provide the official standalone Metastore binary/source tarball and the Docker image.

PR #5942, opened 2025-07-07
Add Hive Metastore RELEASE_NOTES

dengzhhu653:


dengzhhu653:
I collected 93 standalone Metastore issues from https://issues.apache.org/jira/sr/jira.issueviews:searchrequest-html-current-fields/temp/SearchRequest.html?jqlQuery=project+%3D+HIVE+AND+fixVersion+%3D+4.1.0+and+component+in+%28%22Metastore%22%2C+%22Standalone+Metastore%22%2C+metastore%29+and+resolution+%3D+Fixed+. Let me know if we need to drop or add more issues to the notes, cc @zhangbutao @deniskuzZ

standalone-metastore/RELEASE_NOTES.txt
```diff
@@ -0,0 +1,111 @@
+Release Notes - Hive Metastore - Version 4.1.0
```
zhangbutao, line 1: Should we exclude **standalone Metastore related issues** from the root directory [RELEASE_NOTES.txt](https://github.com/apache/hive/blob/branch-4.1/RELEASE_NOTES.txt)

How should we link the two RELEASE_NOTES? I'm worried that users might miss the release notes under the standalone Metastore section.

dengzhhu653: This release note will be included in the standalone Metastore bin/src tarball only, for the root it will shipped with the original hive bin/src tarball, we can stay as it is.

zhangbutao (approved):
+1 LGTM
Thanks!

standalone-metastore/RELEASE_NOTES.txt
deniskuzZ: @dengzhhu653, should we add a template and packaging to the master, so we won't forget this in the next releases? 
Also would be nice to add instructions in the release guide.

deniskuzZ (approved):
+1

standalone-metastore/RELEASE_NOTES.txt
dengzhhu653: sure will add it

dengzhhu653:
Given no code changes in this PR, I will merge it, the only test failure is testPartitionDiscoveryTransactionalTable – org.apache.hadoop.hive.metastore.TestPartitionManagement. @zhangbutao could you please check the test if have secs? thank you!

zhangbutao:
> Given no code changes in this PR, I will merge it, the only test failure is testPartitionDiscoveryTransactionalTable – org.apache.hadoop.hive.metastore.TestPartitionManagement. @zhangbutao could you please check the test if have secs? thank you!

@dengzhhu653 I just ran this test several times locally and the results were fine. Therefore, this failed test should be considered flaky. There is no need for us to worry about it.

PR #5943, opened 2025-07-08
HIVE-29066: PARTITION_NAME_WHITELIST_PATTERN is not honouring session level configuration

Aggarwal-Raghav:
Check [HIVE-29066](https://issues.apache.org/jira/browse/HIVE-29066), Ensure that the partitionValidationPattern is picked from the metaconf instead of HMSHandler Object.

To honour, modified session level metaconf.

NO

q file is attached.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
@@ -24,6 +24,7 @@ import static org.apache.hadoop.hive.ql.security.authorization.HiveCustomStorage
 import static org.apache.hadoop.hive.ql.security.authorization.HiveCustomStorageHandlerUtils.setWriteOperation;
 import static org.apache.hadoop.hive.ql.security.authorization.HiveCustomStorageHandlerUtils.setWriteOperationIsSorted;
 
+import com.google.common.collect.Lists;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
@@ -38,20 +39,17 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.function.BiFunction;
-
-import com.google.common.collect.Lists;
-
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-
 import org.apache.hadoop.hive.common.FileUtils;
 import org.apache.hadoop.hive.common.StatsSetupConst;
 import org.apache.hadoop.hive.conf.HiveConf;
-import org.apache.hadoop.hive.conf.HiveConfUtil;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
+import org.apache.hadoop.hive.conf.HiveConfUtil;
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
+import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.ql.CompilationOpContext;
 import org.apache.hadoop.hive.ql.ErrorMsg;
 import org.apache.hadoop.hive.ql.exec.Utilities.MissingBucketsContext;
@@ -81,7 +79,11 @@ import org.apache.hadoop.hive.ql.plan.TableDesc;
 import org.apache.hadoop.hive.ql.plan.api.OperatorType;
 import org.apache.hadoop.hive.ql.stats.StatsCollectionContext;
 import org.apache.hadoop.hive.ql.stats.StatsPublisher;
-import org.apache.hadoop.hive.serde2.*;
+import org.apache.hadoop.hive.serde2.AbstractSerDe;
+import org.apache.hadoop.hive.serde2.ColumnProjectionUtils;
+import org.apache.hadoop.hive.serde2.SerDeException;
+import org.apache.hadoop.hive.serde2.SerDeStats;
+import org.apache.hadoop.hive.serde2.Serializer;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorUtils;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorUtils.ObjectInspectorCopyOption;
@@ -92,13 +94,11 @@ import org.apache.hadoop.hive.serde2.objectinspector.primitive.IntObjectInspecto
 import org.apache.hadoop.hive.shims.HadoopShims.StoragePolicyShim;
 import org.apache.hadoop.hive.shims.HadoopShims.StoragePolicyValue;
 import org.apache.hadoop.hive.shims.ShimLoader;
-
 import org.apache.hadoop.io.IntWritable;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Writable;
 import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.util.ReflectionUtils;
-
 import org.apache.hive.common.util.HiveStringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -1125,12 +1125,19 @@ public class FileSinkOperator extends TerminalOperator<FileSinkDesc> implements
           }
         }
 
-        String invalidPartitionVal;
-        if((invalidPartitionVal = HiveStringUtils.getPartitionValWithInvalidCharacter(dpVals, dpCtx.getWhiteListPattern()))!=null) {
-          throw new HiveFatalException("Partition value '" + invalidPartitionVal +
-              "' contains a character not matched by whitelist pattern '" +
-              dpCtx.getWhiteListPattern().toString() + "'.  " + "(configure with " +
-              HiveConf.ConfVars.METASTORE_PARTITION_NAME_WHITELIST_PATTERN.varname + ")");
+        String invalidPartitionVal =
+            HiveStringUtils.getPartitionValWithInvalidCharacter(
+                dpVals, dpCtx.getWhiteListPattern());
+
+        if (invalidPartitionVal != null) {
+          String errorMsg =
+              ("Partition value '%s' contains a character not matched by whitelist pattern '%s'. Configure with %s")
+                  .formatted(
+                      invalidPartitionVal,
+                      dpCtx.getWhiteListPattern().toString(),
+                      MetastoreConf.ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getVarname());
+
+          throw new HiveFatalException(errorMsg);
         }
         fpaths = getDynOutPaths(dpVals, lbDirName);
         dynamicPartitionSpecs.add(fpaths.dpDirForCounters);
--- a/ql/src/java/org/apache/hadoop/hive/ql/plan/DynamicPartitionCtx.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/plan/DynamicPartitionCtx.java
@@ -91,7 +91,7 @@ public class DynamicPartitionCtx implements Serializable {
     this.spPath = null;
     String confVal;
     try {
-      confVal = Hive.get().getMetaConf(ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getHiveName());
+      confVal = Hive.get().getMetaConf(ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getVarname());
     } catch (HiveException e) {
       throw new SemanticException(e);
     }
@@ -126,7 +126,7 @@ public class DynamicPartitionCtx implements Serializable {
     }
     String confVal;
     try {
-      confVal = Hive.get().getMetaConf(ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getHiveName());
+      confVal = Hive.get().getMetaConf(ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getVarname());
     } catch (HiveException e) {
       throw new SemanticException(e);
     }
--- a/ql/src/test/queries/clientnegative/dynamic_partitions_with_whitelist.q
+++ b/ql/src/test/queries/clientnegative/dynamic_partitions_with_whitelist.q
@@ -2,7 +2,6 @@
 set hive.strict.checks.bucketing=false;
 
 set hive.mapred.mode=nonstrict;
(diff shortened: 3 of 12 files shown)


Aggarwal-Raghav:
The test failure is flaky as reported in https://issues.apache.org/jira/browse/HIVE-29061, will re-trigger if there are any review commets.

Aggarwal-Raghav:
Can you please help with the review @deniskuzZ @okumin ?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -525,6 +520,15 @@ public String getMetaConf(String key) throws MetaException {
     return getConf().get(key, confVar.getDefaultVal().toString());
   }
 
+  private Pattern getPartitionValidationRegex() {
```
deniskuzZ: move it to MetaStoreServerUtils#validatePartitionNameCharacters
````
MetaStoreServerUtils.validatePartitionNameCharacters(part.getValues(), conf)
````

**Comment by Aggarwal-Raghav** [comment]
Just an Update, Issue exist in dynamic partition insertion as well on HS2 as well. Changes are required in `DynamicPartitionCtx` as well.

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java (hunk 1 above)
Comment (Aggarwal-Raghav) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: @deniskuzZ , can you please check if the latest commit is what you meant by moving it to MetaStoreServerUtils?

**Comment by Aggarwal-Raghav** [comment]
Also, in HS2 DynamicPartitionCtx#whiteListPattern field exist per object. I am seeing the issue in cluster as well.
**Testing (when setting the conf in beeline session)**
Screenshot 2025-07-12 at 8 26 10 PM

**(when i set the config in hive-site.xml)**
Screenshot 2025-07-12 at 8 27 52 PM

Unfortunately, its not happening in q file. Can you suggest should I make similar changes in DynamicPartitionCtx and [org/apache/hadoop/hive/ql/exec/FileSinkOperator.java:1129](https://github.com/apache/hive/blob/1411a401da57cbb3b047c7f25d04774b0204ad04/ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java#L1130)?

**Comment by Aggarwal-Raghav** [comment]
The reason why session level `hive.metastore.partition.name.whitelist.pattern` is because of this line: https://github.com/apache/hive/blob/f2fdea058f0124153514d40fa827e91c730c0e51/ql/src/java/org/apache/hadoop/hive/ql/plan/DynamicPartitionCtx.java#L129

If we are using `hive.metastore.partition.name.whitelist.pattern` then it should look in:
```
confVal = Hive.get().getConf().get(ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getHiveName());
```
instead of **getMetaConf**

In q files, as HMS runs is embedded the conf object "might be" shared.

File: ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java [hunk 2]
```diff
@@ -1241,6 +1227,8 @@ public void process(Object row, int tag) throws HiveException {
     } catch (SerDeException e) {
       closeWriters(true);
       throw new HiveException(e);
+    } catch (MetaException ex) {
+      throw new RuntimeException(ex);
```
deniskuzZ: should we throw HiveException? maybe squash with above `SerDeException `

ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
```diff
@@ -66,10 +65,7 @@
 import org.apache.hadoop.hive.ql.io.StatsProvidingRecordWriter;
 import org.apache.hadoop.hive.ql.io.StreamingOutputFormat;
 import org.apache.hadoop.hive.ql.io.orc.OrcRecordUpdater;
-import org.apache.hadoop.hive.ql.metadata.HiveException;
-import org.apache.hadoop.hive.ql.metadata.HiveFatalException;
-import org.apache.hadoop.hive.ql.metadata.HiveStorageHandler;
-import org.apache.hadoop.hive.ql.metadata.HiveUtils;
+import org.apache.hadoop.hive.ql.metadata.*;
```
deniskuzZ: please avoid wildcard imports

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -255,31 +253,44 @@ public static double decimalToDouble(Decimal decimal) {
...
-  public static void validatePartitionNameCharacters(List<String> partVals,
-                                                     Pattern partitionValidationPattern) throws MetaException {
-
-    String invalidPartitionVal = getPartitionValWithInvalidCharacter(partVals, partitionValidationPattern);
-    if (invalidPartitionVal != null) {
-      throw new MetaException("Partition value '" + invalidPartitionVal +
-          "' contains a character " + "not matched by whitelist pattern '" +
-          partitionValidationPattern.toString() + "'.  " + "(configure with " +
-          MetastoreConf.ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getVarname() + ")");
+  public static Pattern getPartitionValidationRegex(Configuration conf) {
```
deniskuzZ: can it be private?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -775,9 +786,8 @@ public static String validateSkewedColNamesSubsetCol(List<String> skewedColNames
     return copySkewedColNames.toString();
   }
 
-  public static boolean partitionNameHasValidCharacters(List<String> partVals,
-                                                        Pattern partitionValidationPattern) {
-    return getPartitionValWithInvalidCharacter(partVals, partitionValidationPattern) == null;
+  public static boolean partitionNameHasValidCharacters(List<String> partVals, Configuration conf) {
```
deniskuzZ, line 788: same, can it be private?

ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
Aggarwal-Raghav: Apologies, Editor's doing.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -255,31 +253,44 @@ public static double decimalToDouble(Decimal decimal) {
...
+              .formatted(
+                  invalidPartitionVal,
+                  partitionValidationPattern.toString(),
+                  MetastoreConf.ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getHiveName(),
+                  MetastoreConf.ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getVarname()));
     }
+  }
 
-    return null;
+  public static String getPartitionValWithInvalidCharacter(
```
deniskuzZ: we should move this code to `metastore-common` module (MetaStoreUtils), otherwise, we are coupling hive-exec with `metastore-server`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
Aggarwal-Raghav: It is referenced in `HMSHandler#partition_name_has_valid_characters()` but we can make `getPartitionValWithInvalidCharacter` private.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
Aggarwal-Raghav: Ack. Let me check.

ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
Aggarwal-Raghav: Ack.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
deniskuzZ: thanks!

ql/src/java/org/apache/hadoop/hive/ql/exec/FileSinkOperator.java
Aggarwal-Raghav:
```
} catch (SerDeException | MetaException e) {
      closeWriters(true);
      throw new HiveException(e);
    }
```

Doing so can result in error in q file in dynamic_partitions_with_whitelist.q
```
< Caused by: java.lang.NullPointerException: Cannot invoke "org.apache.hadoop.hive.ql.exec.FileSinkOperator$FSPaths.closeWriters(boolean, java.util.List)" because "this.fpaths" is null
---
> Caused by: java.lang.RuntimeException: MetaException(message:Partition value 'val_129' contains a character not matched by whitelist pattern '[^9]*'. Configure with hive.metastore.partition.name.whitelist.pattern for dynamic partitioning otherwise use metaconf:metastore.partition.name.whitelist.pattern)
37a38,39
```

Earlier it was `HiveFatalException`, let's contiue that what do you think?
```
 } catch (SerDeException e) {
      closeWriters(true);
      throw new HiveException(e);
    } catch (MetaException e) {
      throw new HiveFatalException(e);
    }
```
CC @deniskuzZ

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
Aggarwal-Raghav: @deniskuzZ, I have a question, can you please help me clear a doubt for me?
1. FileSinkOperator#process() calling MetaStoreServerUtils.validatePartitionNameCharacters (Line 1120) is also coupling right? Yes/No 😅
2. I believe its Yes, then I'm inclined toward having duplicate methods for partition regex validation in HS2 as earlier it was. 

Another question:
As FileSinkOperator runs inside TezTask for dynamic partitioning (insert overwrite query) is that why it is working? as necessary java classes are serialized via kryo to tez task because without thrift, the communication shouldn't be possible b/w `ql` module and `metastore-server`? Am I missing something?

ql/src/java/org/apache/hadoop/hive/ql/plan/DynamicPartitionCtx.java
```diff
@@ -91,7 +91,7 @@ public DynamicPartitionCtx(List<String> partColNames, String defaultPartName,
     this.spPath = null;
     String confVal;
     try {
-      confVal = Hive.get().getMetaConf(ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getHiveName());
+      confVal = Hive.get().getMetaConf(ConfVars.PARTITION_NAME_WHITELIST_PATTERN.getVarname());
```
Aggarwal-Raghav, line 94: This change is enough for fixing dynamic partition. Reasoning:

hivename: hive.metastore.partition.name.whitelist.pattern
varname: metastore.partition.name.whitelist.pattern

As users do `set metaconf:metastore.partition.name.whitelist.pattern=[^9]*`, then varname is used not the hivename. I believe it was never working earlier also for session level.

NOTE: This also means that in hive-site.xml as well this is now going to be picked instead of hivename. 
```
  <property>
    <name>metastore.partition.name.whitelist.pattern</name>
    <value>[^9]*</value>
  </property>
```
This shouldn't be an issue because:
1. in HiveConf this config is deprecated.
2. If users still uses hive.metastore.partition.name.whitelist.pattern, then also MoveTask will fail  (add_partitions_req) which is indireclty the expected behaviour 
```
Error: Error while compiling statement: FAILED: Execution Error, return code 40000 from org.apache.hadoop.hive.ql.exec.MoveTask. MetaException(message:Partition value '09' contains a character not matched by whitelist pattern '[^9]*'. Configure with metastore.partition.name.whitelist.pattern); Query ID: raghav_20250729215001_82d33e22-04db-4be3-8b01-189504089bf6 (state=08S01,code=40000)
```
3. If users uses hive.metastore.partition.name.whitelist.pattern for HMS Side operation like alter then it will also work becuase of `MetastoreConf#getVar()` which check for hivename as failsafe. That's why q file are not required to change.

ql/src/test/queries/clientnegative/dynamic_partitions_with_whitelist.q
```diff
@@ -17,4 +16,5 @@ load data local inpath '../../data/files/bmj/000000_0' INTO TABLE source_table p
 -- If the directory is not empty the hook will throw an error, instead the error should come from the metastore
 -- This shows that no dynamic partitions were created and left behind or had directories created
 
+SET metaconf:metastore.partition.name.whitelist.pattern=[^9]*;
```
Aggarwal-Raghav, line 19: The reason for moving this line below is, the metaconf will be resetted after any meta-operation i.e create etc. My assumption is, metconf should always be a line before the config is consumed in the query. the question can be tracked [here](https://issues.apache.org/jira/browse/HIVE-29066?focusedCommentId=17995799&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17995799).

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
deniskuzZ: Use of `standalone-metastore-server` module classes in any of Hive modules like `hive-exec` should be avoided.

> As FileSinkOperator runs inside TezTask for dynamic partitioning (insert overwrite query) is that why it is working? as necessary java classes are serialized via kryo to tez task because without thrift, the communication shouldn't be possible b/w ql module and metastore-server? Am I missing something?

Tez uses `hive-exec` jar that is a shaded fat jar that includes the `metastore-server` dependency and it's classes. Once we fix the coupling - we should drop it from `hive-exec`

PR #5947, opened 2025-07-09
HIVE-29072:Cleanup IMetaStoreClient#getTables APIs

ramitg254:

cleanup of unnecessary wrapper methods in IMetastoreClient

There are different variations of same methods which are not required as per [HIVE-23971](https://issues.apache.org/jira/browse/HIVE-23971)

No

via local build and unit tests

### DIFF
--- a/hcatalog/core/src/test/java/org/apache/hive/hcatalog/cli/TestPermsGrp.java
+++ b/hcatalog/core/src/test/java/org/apache/hive/hcatalog/cli/TestPermsGrp.java
@@ -245,7 +245,7 @@ public class TestPermsGrp {
 
   private void silentDropDatabase(String dbName) throws MetaException, TException {
     try {
-      for (String tableName : msc.getTables(dbName, "*")) {
+      for (String tableName : msc.getTables(dbName, "*", null)) {
         msc.dropTable(dbName, tableName);
       }
 
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetastoreTransformer.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetastoreTransformer.java
@@ -101,7 +101,7 @@ public class TestHiveMetastoreTransformer {
 
   private static void silentDropDatabase(String dbName) throws TException {
     try {
-      for (String tableName : client.getTables(dbName, "*")) {
+      for (String tableName : client.getTables(dbName, "*", null)) {
         client.dropTable(dbName, tableName);
       }
       client.dropDatabase(dbName);
--- a/itests/hive-unit/src/test/java/org/apache/hadoop/hive/metastore/TestTenantBasedStorageHierarchy.java
+++ b/itests/hive-unit/src/test/java/org/apache/hadoop/hive/metastore/TestTenantBasedStorageHierarchy.java
@@ -102,7 +102,7 @@ public class TestTenantBasedStorageHierarchy {
 
   private static void silentDropDatabase(String dbName) throws TException {
     try {
-      for (String tableName : client.getTables(dbName, "*")) {
+      for (String tableName : client.getTables(dbName, "*", null)) {
         client.dropTable(dbName, tableName);
       }
       client.dropDatabase(dbName);
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -1989,9 +1989,9 @@ public class Hive implements AutoCloseable {
         }
       } else {
         if (pattern != null) {
-          result = getMSC().getTables(dbName, pattern);
+          result = getMSC().getTables(dbName, pattern, null);
         } else {
-          result = getMSC().getTables(dbName, ".*");
+          result = getMSC().getTables(dbName, ".*", null);
         }
       }
       return result;
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
@@ -191,34 +191,6 @@ public interface IMetaStoreClient extends AutoCloseable {
 
   /**
    * Get the names of all tables in the specified database that satisfy the supplied
-   * table name pattern.
-   * @param dbName database name.
-   * @param tablePattern pattern for table name to conform to
-   * @return List of table names.
-   * @throws MetaException error fetching information from the RDBMS
-   * @throws TException thrift transport error
-   * @throws UnknownDBException indicated database to search in does not exist.
-   */
-  List<String> getTables(String dbName, String tablePattern)
-      throws MetaException, TException, UnknownDBException;
-
-  /**
-   * Get the names of all tables in the specified database that satisfy the supplied
-   * table name pattern.
-   * @param catName catalog name.
-   * @param dbName database name.
-   * @param tablePattern pattern for table name to conform to
-   * @return List of table names.
-   * @throws MetaException error fetching information from the RDBMS
-   * @throws TException general thrift error
-   * @throws UnknownDBException indicated database to search in does not exist.
-   */
-  List<String> getTables(String catName, String dbName, String tablePattern)
-      throws MetaException, TException, UnknownDBException;
-
-
-  /**
-   * Get the names of all tables in the specified database that satisfy the supplied
    * table name pattern and table type (MANAGED_TABLE || EXTERNAL_TABLE || VIRTUAL_VIEW)
    * @param dbName Name of the database to fetch tables in.
    * @param tablePattern pattern to match for table names.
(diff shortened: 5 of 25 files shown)


itests/hive-unit/src/test/java/org/apache/hadoop/hive/metastore/TestHiveMetastoreTransformer.java
```diff
@@ -101,7 +101,7 @@ public void setUp() throws Exception {
 
   private static void silentDropDatabase(String dbName) throws TException {
     try {
-      for (String tableName : client.getTables(dbName, "*")) {
+      for (String tableName : client.getTables(dbName, "*", null)) {
```
deniskuzZ, line 104: same

itests/hive-unit/src/test/java/org/apache/hadoop/hive/metastore/TestTenantBasedStorageHierarchy.java
```diff
@@ -102,7 +102,7 @@ public void setUp() throws Exception {
 
   private static void silentDropDatabase(String dbName) throws TException {
     try {
-      for (String tableName : client.getTables(dbName, "*")) {
+      for (String tableName : client.getTables(dbName, "*", null)) {
```
deniskuzZ, line 105: same

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -1989,9 +1989,9 @@ public List<String> getTablesByType(String dbName, String pattern, TableType typ
         }
       } else {
         if (pattern != null) {
-          result = getMSC().getTables(dbName, pattern);
+          result = getMSC().getTables(dbName, pattern, null);
         } else {
-          result = getMSC().getTables(dbName, ".*");
+          result = getMSC().getTables(dbName, ".*", null);
```
deniskuzZ, line 1994: same

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
```diff
@@ -147,7 +147,7 @@ public CheckResult checkMetastore(String catName, String dbName, String tableNam
         // no table specified, check all tables and all partitions.
         List<String> tables = new ArrayList<>();
         try{
-          tables = getMsc().getTables(catName, dbName, ".*");
+          tables = getMsc().getTables(catName, dbName, ".*", null);
```
deniskuzZ, line 150: same

hcatalog/core/src/test/java/org/apache/hive/hcatalog/cli/TestPermsGrp.java
```diff
@@ -245,7 +245,7 @@ private int callHCatCli(String[] args) throws Exception {
 
   private void silentDropDatabase(String dbName) throws MetaException, TException {
     try {
-      for (String tableName : msc.getTables(dbName, "*")) {
+      for (String tableName : msc.getTables(dbName, "*", null)) {
```
deniskuzZ, line 248: why not keep simplified API that doesn't require tableType?

ramitg254: due to this part of code 

```
if (tableType == TableType.MANAGED_TABLE || tableType == TableType.EXTERNAL_TABLE) {
        for (Map.Entry<String, org.apache.hadoop.hive.ql.metadata.Table> tableData : tables.entrySet()) {
          matcher.reset(tableData.getKey());
          if (matcher.matches()) {
            if (tableData.getValue().getTableType() == tableType) {
              // If tableType is the same that we are requesting,
              // add table the the list
              combinedTableNames.add(tableData.getKey());
            } else {
              // If tableType is not the same that we are requesting,
              // remove it in case it was added before, as temp table
              // overrides original table
              combinedTableNames.remove(tableData.getKey());
            }
          }
        }
      }

```
 in `SessionHiveMetaStoreClient.java`
 
 and occurences like 
 ```
List<String> viewNames =
              clients.run(client -> client.getTables(database, "*", TableType.VIRTUAL_VIEW));
```

```
List<String> materializedViews =
          getTables(req.getCatalogName(), req.getName(), ".*", TableType.MATERIALIZED_VIEW);
```
which require tableType

Comment (ramitg254) [reply hcatalog/core/src/test/java/org/apache/hive/hcatalog/cli/TestPermsGrp.java:248]: and these simplified api are just wrappers around the current methods which was asked to be removed in (https://issues.apache.org/jira/browse/HIVE-23971#:~:text=Also%20affects-,getTables%20API,-.%20Bumping%20this%20up) 
and all of these parameters can be dealt by  https://github.com/apache/hive/pull/5947/files#:~:text=static%20GetTablesRequest-,convertToGetTablesRequest,-(String%20catName  other than `tableType `cause it is not part of `GetTablesRequest` object

Comment (deniskuzZ) [reply hcatalog/core/src/test/java/org/apache/hive/hcatalog/cli/TestPermsGrp.java:248]: isn't it more convenient to use `msc.getTables(dbName, "*") -> msc.getTables(dbName, "*", null)` than hardcoding nulls?
btw, you can just drop/modify the public API, you have to deprecate it first

Comment (ramitg254) [reply hcatalog/core/src/test/java/org/apache/hive/hcatalog/cli/TestPermsGrp.java:248]: yeah your point is also correct but I referenced this earlier pr regarding cleanup: https://github.com/apache/hive/pull/3072/files
and their also:
```
client.getPartitionsByNames(catName, dbName, tableName,
                       Collections.singletonList(partName), true, ENGINE);
```
 was replaced with:
 ```
GetPartitionsByNamesRequest req = convertToGetPartitionsByNamesRequest(database, tableName,
           Collections.singletonList(partName), true, ENGINE, null, null);
       Partition partition = client.getPartitionsByNames(req).getPartitions().get(0);
```
 with two extra null values 
 so what I understood from that is despite having a separate method it is better to have an extra param which can deal with both null as well as non null args

ramitg254: If this seems right from maintainability perspective then I'll deprecate the other methods otherwise I'll close this pr if current state of getTables is more convenient

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -189,34 +189,6 @@ List<String> getDatabases(String catName, String databasePattern)
...
-   * Get the names of all tables in the specified database that satisfy the supplied
-   * table name pattern.
-   * @param dbName database name.
-   * @param tablePattern pattern for table name to conform to
-   * @return List of table names.
-   * @throws MetaException error fetching information from the RDBMS
-   * @throws TException thrift transport error
-   * @throws UnknownDBException indicated database to search in does not exist.
-   */
-  List<String> getTables(String dbName, String tablePattern)
```
deniskuzZ, line 202: @ramitg254, we can't just go ahead and change the client API. If we have an agreement that this could be dropped , we should first deprecate it
cc @dengzhhu653

deniskuzZ:
@ramitg254, there are number of deprecated `getTable` methods, maybe it's time to drop them. It seems that Hive-26537 skipped the client interface cleanup.
````
  @Deprecated
  Table getTable(String dbName, String tableName);

  @Deprecated
  Table getTable(String catName, String dbName, String tableName);

  @Deprecated
  Table getTable(String dbName, String tableName, 
      boolean getColumnStats, String engine);
  
  @Deprecated
  Table getTable(String catName, String dbName, String tableName, 
      String validWriteIdList);

  @Deprecated
  Table getTable(String catName, String dbName, String tableName,
      String validWriteIdList, boolean getColumnStats, String engine) 
````
cc @saihemanth-cloudera, @dengzhhu653

saihemanth-cloudera:
@deniskuzZ - I have deprecated those APIs in a major release Hive 4.0. So I thought it may be better to remove the deprecated APIs in the next major release Hive 5.0. Do you believe that removing these deprecated APIs in a minor release like 4.2.0 is a good idea?

deniskuzZ:
> @deniskuzZ - I have deprecated those APIs in a major release Hive 4.0. So I thought it may be better to remove the deprecated APIs in the next major release Hive 5.0. Do you believe that removing these deprecated APIs in a minor release like 4.2.0 is a good idea?

@saihemanth-cloudera, but you dropped it from the thrift API, see https://github.com/apache/iceberg/issues/12878

saihemanth-cloudera:
> @saihemanth-cloudera, but you dropped it from the thrift API, see [apache/iceberg#12878](https://github.com/apache/iceberg/issues/12878)
@deniskuzZ - I have removed the get_table() APIs that were already deprecated in HIVE 3.x release. I have modified the data connector related APIs because they were not in the Hive 3.x release.

deniskuzZ:
> > @saihemanth-cloudera, but you dropped it from the thrift API, see [apache/iceberg#12878](https://github.com/apache/iceberg/issues/12878)
> > @deniskuzZ - I have removed the get_table() APIs that were already deprecated in HIVE 3.x release. I have modified the data connector related APIs because they were not in the Hive 3.x release.

that's fine, but why keep the client implementing that API (also marked as deprecated) ? 
I think it make sense to expose the HMS API using Swagger or a similar tool to improve documentation

PR #5950, opened 2025-07-09
HIVE-28813: MSCK/Analyze commands can show a warning in console for Small files.

DanielZhu58:
Add warning when MSCK/Analyze commands can show a warning in console for table/partition if (totalSize / numFiles) is less than avgFileSize.

So the users can know if there are small files.

No.

Manual tests.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
@@ -21,6 +21,9 @@ package org.apache.hadoop.hive.ql.ddl.misc.msck;
 import static org.apache.hadoop.hive.metastore.Msck.getProxyClass;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 
 import org.apache.hadoop.hive.common.TableName;
 import org.apache.hadoop.hive.metastore.HiveMetaStoreClient;
@@ -75,7 +78,18 @@ public class MsckOperation extends DDLOperation<MsckDesc> {
       MsckInfo msckInfo = new MsckInfo(SessionState.get().getCurrentCatalog(), tableName.getDb(), tableName.getTable(),
           desc.getFilterExp(), desc.getResFile(), desc.isRepairPartitions(),
           desc.isAddPartitions(), desc.isDropPartitions(), partitionExpirySeconds);
-      return msck.repair(msckInfo);
+      int result = msck.repair(msckInfo);
+      if (msckInfo.getSmallFilesStats() != null && !msckInfo.getSmallFilesStats().isEmpty()){
+        List<String> logInfo = new ArrayList<>(msckInfo.getSmallFilesStats().size());
+        for (Map.Entry<String, String> me : msckInfo.getSmallFilesStats().entrySet()){
+          String partitionName = me.getKey();
+          String smallFilesStats = me.getValue();
+          logInfo.add("This table/partition average file size is less than Hive average file size.\n " +
+                  "The partition name is " + partitionName + ". " + smallFilesStats);
+        }
+        LOG.info("There are small files exist.\n{}", String.join("\n", logInfo));
+      }
+      return result;
     } catch (MetaException | MetastoreException e) {
       LOG.error("Unable to create msck instance.", e);
       throw e;
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
@@ -38,6 +38,7 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
@@ -59,6 +60,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathFilter;
 import org.apache.hadoop.fs.RemoteIterator;
+import org.apache.hadoop.hive.common.StatsSetupConst;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.GetPartitionsFilterSpec;
 import org.apache.hadoop.hive.metastore.api.GetPartitionsRequest;
@@ -348,6 +350,7 @@ public class HiveMetaStoreChecker {
       allPartDirs = partDirs;
     }
 
+    Map<String, String> smallFilesStats = new HashMap<>();
     // check that the partition folders exist on disk
     for (Partition partition : parts) {
       if (partition == null) {
@@ -358,6 +361,28 @@ public class HiveMetaStoreChecker {
       if (partPath == null) {
         continue;
       }
+
+      // check the table/partition if (totalSize / numFiles) is less than avgFileSize
+      Map<String, String> partitionParameters = partition.getParameters();
+      if (partitionParameters != null) {
+        long totalSize = Long.parseLong(partitionParameters.getOrDefault(StatsSetupConst.TOTAL_SIZE, "0"));
+        long numFiles = Long.parseLong(partitionParameters.getOrDefault(StatsSetupConst.NUM_FILES, "0"));
+        if (numFiles != 0) {
+          long avgFileSize = totalSize / numFiles;
+          long HIVE_FILES_AVG_SIZE = conf.getLong("hive.merge.smallfiles.avgsize", 0);
+          if (avgFileSize <= HIVE_FILES_AVG_SIZE) {
+            StringBuilder tmpStatsSb = new StringBuilder();
+            tmpStatsSb.append("totalSize = ").append(totalSize).append(", numFiles = ")
+                    .append(numFiles).append(". ");
+            smallFilesStats.putIfAbsent(Warehouse.makePartName(table.getPartitionKeys(), partition.getValues()), tmpStatsSb.toString());
+            result.setSmallFilesStats(smallFilesStats);
+          }
+        } else {
+          LOG.warn("Partition total number of files is 0.");
+          LOG.debug("Partition total number of files is 0. Table name : {}, Partition values : {}. ", partition.getTableName(), partition.getValues().toString());
+        }
+      }
+
       fs = partPath.getFileSystem(conf);
 
       CheckResult.PartitionResult prFromMetastore = new CheckResult.PartitionResult();
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
@@ -143,6 +143,7 @@ public class Msck {
       // accessed through getPartitionNotOnFS
       result = checker.checkMetastore(msckInfo.getCatalogName(), msckInfo.getDbName(), msckInfo.getTableName(),
         msckInfo.getFilterExp(), table);
+      msckInfo.setSmallFilesStats(result.getSmallFilesStats());
       Set<CheckResult.PartitionResult> partsNotInMs = result.getPartitionsNotInMs();
       Set<CheckResult.PartitionResult> partsNotInFs = result.getPartitionsNotOnFs();
       Set<CheckResult.PartitionResult> expiredPartitions = result.getExpiredPartitions();
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MsckInfo.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MsckInfo.java
@@ -18,6 +18,9 @@
 package org.apache.hadoop.hive.metastore;
 
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Metadata related to Msck.
  */
@@ -32,6 +35,8 @@ public class MsckInfo {
   private final boolean addPartitions;
   private final boolean dropPartitions;
   private final long partitionExpirySeconds;
+
+  private Map<String, String> smallFilesStats = new HashMap<>();
 
   public MsckInfo(String catalogName, String dbName, String tableName, byte[] filterExp, String resFile,
                   boolean repairPartitions, boolean addPartitions,
@@ -81,5 +86,11 @@ public class MsckInfo {
 
   public long getPartitionExpirySeconds() {
     return partitionExpirySeconds;
+  }
+
+  public Map<String, String> getSmallFilesStats() { return smallFilesStats; }
+
+  public void setSmallFilesStats(Map<String, String> smallFilesStats){
+    this.smallFilesStats = smallFilesStats;
   }
 }
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/CheckResult.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/CheckResult.java
@@ -20,6 +20,7 @@ package org.apache.hadoop.hive.metastore;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
@@ -36,6 +37,9 @@ public class CheckResult {
   private Set<PartitionResult> partitionsNotInMs = new TreeSet<>();
   private Set<PartitionResult> expiredPartitions = new TreeSet<>();
   private Set<PartitionResult> correctPartitions = new TreeSet<>();
+
+  private Map<String, String> smallFilesStats = new HashMap<>();
+
   private long maxWriteId;
   private long maxTxnId;
 
@@ -113,6 +117,14 @@ public class CheckResult {
 
   public void setCorrectPartitions(final Set<PartitionResult> correctPartitions) {
     this.correctPartitions = correctPartitions;
+  }
+
+  public Map<String, String> getSmallFilesStats() {
+    return this.smallFilesStats;
+  }
+
+  public void setSmallFilesStats(Map<String, String> smallFilesStats) {
+    this.smallFilesStats = smallFilesStats;
   }
 
   public long getMaxWriteId() {


ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
```diff
@@ -75,7 +78,18 @@ public int execute() throws HiveException, IOException, TException, MetastoreExc
       MsckInfo msckInfo = new MsckInfo(SessionState.get().getCurrentCatalog(), tableName.getDb(), tableName.getTable(),
           desc.getFilterExp(), desc.getResFile(), desc.isRepairPartitions(),
           desc.isAddPartitions(), desc.isDropPartitions(), partitionExpirySeconds);
-      return msck.repair(msckInfo);
+      int result = msck.repair(msckInfo);
+      if (msckInfo.getSmallFilesStats() != null && !msckInfo.getSmallFilesStats().isEmpty()){
```
deniskuzZ: can we move the log warning inside of `msck.repair` function?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MsckInfo.java
```diff
@@ -33,6 +36,8 @@ public class MsckInfo {
   private final boolean dropPartitions;
   private final long partitionExpirySeconds;
 
+  private Map<String, String> smallFilesStats = new HashMap<>();
```
deniskuzZ, line 39: it can be costly in terms of memory. isn't simple counter enough here?

DanielZhu58: We need to show the end user that in this specific partition, there is a total size, and there is a total number of files, which means there are many small files existing in this partition.
And that can help the end user when they do the "msck" operation.

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
```diff
@@ -75,7 +78,25 @@ public int execute() throws HiveException, IOException, TException, MetastoreExc
...
+        List<String> logInfo = new ArrayList<>(msckInfo.getSmallFilesStats().size());
+        for (Map.Entry<String, String> me : msckInfo.getSmallFilesStats().entrySet()){
+          String partitionName = me.getKey();
+          String smallFilesStats = me.getValue();
+          logInfo.add("This table/partition average file size is less than Hive average file size.\n " +
```
deniskuzZ, line 87: should this be under warn?

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
```diff
@@ -75,7 +78,25 @@ public int execute() throws HiveException, IOException, TException, MetastoreExc
...
+                  "The partition name is " + partitionName + ". " + smallFilesStats);
+        }
+        SessionState ss = SessionState.get();
+        if (ss != null && ss.getConsole() != null) {
+          ss.getConsole().printInfo("[MSCK] Small files detected.");
+          for (String line : logInfo) {
+            ss.getConsole().printInfo("[MSCK] " + line);
+          }
+        }
+        LOG.info("There are small files exist.\n{}", String.join("\n", logInfo));
```
deniskuzZ: should this be under warn?

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
```diff
@@ -75,7 +78,25 @@ public int execute() throws HiveException, IOException, TException, MetastoreExc
...
-      return msck.repair(msckInfo);
+      int result = msck.repair(msckInfo);
+      if (msckInfo.getSmallFilesStats() != null && !msckInfo.getSmallFilesStats().isEmpty()){
+        List<String> logInfo = new ArrayList<>(msckInfo.getSmallFilesStats().size());
+        for (Map.Entry<String, String> me : msckInfo.getSmallFilesStats().entrySet()){
+          String partitionName = me.getKey();
+          String smallFilesStats = me.getValue();
+          logInfo.add("This table/partition average file size is less than Hive average file size.\n " +
+                  "The partition name is " + partitionName + ". " + smallFilesStats);
+        }
```
difin: It would be more modern to re-implement it in a functional style:
`List<String> logInfo = msckInfo.getSmallFilesStats().stream().forEach(...).toList();`

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
```diff
@@ -75,7 +78,25 @@ public int execute() throws HiveException, IOException, TException, MetastoreExc
...
+        SessionState ss = SessionState.get();
+        if (ss != null && ss.getConsole() != null) {
+          ss.getConsole().printInfo("[MSCK] Small files detected.");
+          for (String line : logInfo) {
+            ss.getConsole().printInfo("[MSCK] " + line);
+          }
```
difin: Should it be in functional style `logInfo.stream().forEach(...);`?

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
DanielZhu58: I would prefer to not add the "small files warning" in `msck.repair()` function.
The purpose of repair() will be only plan and apply metadata fixes, user messaging can be belong to the check/plan stage and the CLI layer (MsckOperation).

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
DanielZhu58: The text warnings are printed to the end user in the console.
So this doesn't have to be warnings in the LOG, and small files are common to see.

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
DanielZhu58: Acknowledged. I will use` logInfo.forEach(line -> ss.getConsole().printInfo("[MSCK] " + line));`.

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/exec/StatsTask.java
```diff
@@ -124,6 +133,125 @@ public int execute() {
...
+  /**
+   * Check for small files after ANALYZE completes
+   */
+  private void checkSmallFilesAfterAnalyze(Hive db, Table tbl) {
+    // Use the existing merge small files threshold configuration
+    long smallFilesThreshold = conf.getLongVar(HiveConf.ConfVars.HIVE_MERGE_MAP_FILES_AVG_SIZE);
+    List<String> smallFilesWarnings = new ArrayList<>();
+
+    // Get the console from SessionState for output
+    PrintStream console = SessionState.get().out;
```
dengzhhu653: `SessionState.get().out` for printing the message might not be viewable in the Beeline? use `SessionState.get().getConsole()`?

dengzhhu653: we might need to check whether `SessionState.get()` is null or not

DanielZhu58: This whole code part is removed already. Thanks.

standalone-metastore/metastore-server/pom.xml
```diff
@@ -470,6 +470,12 @@
       <artifactId>mssqlserver</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
```
dengzhhu653: why need this dependency?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
```diff
@@ -358,6 +361,28 @@ void checkTable(Table table, PartitionIterable parts, byte[] filterExp, CheckRes
       if (partPath == null) {
         continue;
       }
+
+      // check the table/partition if (totalSize / numFiles) is less than avgFileSize
+      Map<String, String> partitionParameters = partition.getParameters();
+      if (partitionParameters != null) {
+        long totalSize = Long.parseLong(partitionParameters.getOrDefault(StatsSetupConst.TOTAL_SIZE, "0"));
```
dengzhhu653: if we can take care of the newly added partition, then can this be removed?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -239,6 +244,57 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+            Partition p;
+            try {
+              // the newly added partition should exist in metastore
+              p = getMsc().getPartition(table.getCatName(), table.getDbName(), table.getTableName(), statsVals);
+            } catch (NoSuchObjectException nsoe) {
+              // the partition is not actually added in metastore
+              continue;
+            }
+
+            Path loc = new Path(p.getSd().getLocation());
```
dengzhhu653: we have the partition location in `pr`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -239,6 +244,57 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+            try {
+              FileSystem fs = loc.getFileSystem(getConf());
```
dengzhhu653: we might need a new property to disable the stats gather if can

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
```diff
@@ -182,6 +184,48 @@ public Object process(StatsAggregator statsAggregator) throws HiveException, Met
...
+
+          long avg = totalSize / numFiles;
+          if (avg <= threshold) {
+            // both work for non-partitioned and partitioned tables
+            String who = (p.getPartition() == null)
+                    ? ("table " + p.getTable().getFullyQualifiedName())
+                    : ("partition " + p.getPartition().getName());
+
+            // add the small files warnings in the log
+            LOG.info("[ANALYZE] Small files detected: {} avgBytes={}, files={}, totalBytes={}",
```
dengzhhu653: this message will be logged as well in case ss != null && ss.getConsole() != null, which may lead to duplication in the log file

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
```diff
@@ -75,7 +78,25 @@ public int execute() throws HiveException, IOException, TException, MetastoreExc
...
           desc.getFilterExp(), desc.getResFile(), desc.isRepairPartitions(),
           desc.isAddPartitions(), desc.isDropPartitions(), partitionExpirySeconds);
-      return msck.repair(msckInfo);
+      int result = msck.repair(msckInfo);
+      Map<String, String> smallFilesStats = msckInfo.getSmallFilesStats();
+      if (smallFilesStats != null && !smallFilesStats.isEmpty()) {
+        // keep the small files information in the logs
+        List<String> logInfo = smallFilesStats.entrySet().stream()
+                .map(entry -> String.format(
+                        "This table/partition average file size is less than Hive average file size.%n The partition name is %s. %s",
```
dengzhhu653: lots of the same message might flush the client, we need to shorten the message per line

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
```diff
@@ -75,7 +78,25 @@ public int execute() throws HiveException, IOException, TException, MetastoreExc
...
+                        entry.getKey(), entry.getValue()))
+                .collect(Collectors.toList());
+        LOG.info("There are small files exist.\n{}", String.join("\n", logInfo));
```
dengzhhu653: we just need to keep one log by `ss.getConsole().printInfo`

ql/src/java/org/apache/hadoop/hive/ql/exec/StatsTask.java
```diff
@@ -124,7 +124,6 @@ public int execute() {
     return 0;
   }
 
-
```
dengzhhu653, line 127: nit: remove this change

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -181,6 +184,35 @@ public void run() {
         }
         parameters.putAll(basicStatistics);
         StatsSetupConst.setBasicStatsState(parameters, StatsSetupConst.TRUE);
+        try {
+          String ts = parameters.get(StatsSetupConst.TOTAL_SIZE);
```
dengzhhu653: the code is almost the same as `BasicStatsTask`, can we able to make it a static util method

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -293,6 +325,28 @@ public void run() {
         parameters.put(StatsSetupConst.NUM_FILES, String.valueOf(numFiles));
         parameters.put(StatsSetupConst.NUM_ERASURE_CODED_FILES, String.valueOf(numErasureCodedFiles));
 
+        try {
+          if (numFiles > 1 && fileSize > 0) {
```
dengzhhu653: why again in the same class

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
DanielZhu58: OK, changed to a case condition. If there is console, print to both console and log. If there is no console, only print in the logs.

ql/src/java/org/apache/hadoop/hive/ql/exec/StatsTask.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
DanielZhu58: Because ANALYZE NOSCAN has two different code paths:
Storage-handler path (HiveStorageHandlerStatCollector): stats come from the storage handler.
Footer path (FooterStatCollector): stats are computed from ORC/Parquet footers.

standalone-metastore/metastore-server/pom.xml
DanielZhu58: To retrieve the HiveConf.ConfVars.HIVE_MERGE_MAP_FILES_AVG_SIZE.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
DanielZhu58: Acknowledged. The case will be handled in the newly added partitions instead of here.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: After discussed with Naresh, the msck command can compute the basic stats, including the total size and total amount of files. 
So we don't need the file system params anymore currently.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: I think we don't need this anymore if we changed the implementation approach to using the partition parameters.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
DanielZhu58: Let me think about it.
If I decide to refactor it, I will upload it in my next commit.

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/ddl/misc/msck/MsckOperation.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1821,6 +1821,10 @@ public enum ConfVars {
             "A ZooKeeper instance must be up and running when using zookeeper Hive lock manager "),
     HIVE_TXN_STATS_ENABLED("hive.txn.stats.enabled", "hive.txn.stats.enabled", true,
         "Whether Hive supports transactional stats (accurate stats for transactional tables)"),
+    MSCK_SMALLFILES_AVG_SIZE("metastore.msck.smallfiles.avgsize", "metastore.msck.smallfiles.avgsize", (long) (16 * 1000 * 1000),
```
dengzhhu653, line 1814: if we are not able to access HiveConf HIVE_MERGE_MAP_FILES_AVG_SIZE in Metastore, we can use `conf.get("hive.merge.smallfiles.avgsize", "16000000")` compared a new property

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -181,6 +185,21 @@ public void run() {
         }
         parameters.putAll(basicStatistics);
         StatsSetupConst.setBasicStatsState(parameters, StatsSetupConst.TRUE);
+        try {
```
dengzhhu653: can this be simplified as 
```
  String who = (partish.getPartition() == null) ? ("table " + partish.getTable().getFullyQualifiedName())
                  : ("partition " + partish.getPartition().getName());        
  SmallFilesWarningUtil.showSmallFileWarnings(SessionState.get() != null ? SessionState.get().getConsole() : LOG, parameters, conf, who, "[ANALYZE][NOSCAN]");
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +244,61 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+              // parse the partition name into a few partition values
+              statsVals = Warehouse.makeValsFromName(pr.getPartitionName(), null);
+            } catch (MetaException me) {
+              // skip this partition if this somehow failed
+              continue;
+            }
+            Partition p;
+            try {
+              // the newly added partition should exist in metastore
+              p = getMsc().getPartition(table.getCatName(), table.getDbName(), table.getTableName(), statsVals);
```
dengzhhu653: instead of retrieving the partition one by one, can we use `msc.getPartitionsByNames`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +244,61 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+            } catch (NoSuchObjectException nsoe) {
+              // the partition is not actually added in metastore
+              continue;
+            }
+            Map<String, String> params = p.getParameters();
```
dengzhhu653: can we add a new method `static boolean smallAverageFilesDetected(getConf(), params)` in `SmallFilesWarningUtil` to include the small average file validation?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
DanielZhu58: This is part of the design of this PR. We can add this conf here, it can also be used in any future use.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: Acknowledged. I changed the code about get partitions.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: Acknowledged. The code is changed.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +246,76 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+          final List<String> names = new ArrayList<>(partsNotInMs.size());
+          for (CheckResult.PartitionResult pr : partsNotInMs) {
+            final String name = pr.getPartitionName();
+            if (name != null && !name.isEmpty()) {
+              names.add(name);
+            }
+          }
+
+          // Batch over the names to limit single-RPC payload.
+          final int BATCH = 1000;
```
dengzhhu653: why not use `Batchable`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +246,76 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
 
+        // Generate small files warnings only for partsNotInMs
+        try {
+          // Threshold in bytes for average file size considered "small"
+          final long threshold =
+                  MetastoreConf.getLongVar(getConf(), MetastoreConf.ConfVars.MSCK_SMALLFILES_AVG_SIZE);
+
+          // Collect partition names
+          final List<String> names = new ArrayList<>(partsNotInMs.size());
+          for (CheckResult.PartitionResult pr : partsNotInMs) {
```
dengzhhu653: final List<String> names = partsNotInMs.stream()...

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +246,76 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+            for (Partition p : plist) {
+              final String pName = Warehouse.makePartName(table.getPartitionKeys(), p.getValues());
+              byName.put(pName, p);
+            }
+          }
+
+          // Build small-files stats for partitions that have quick HMS stats.
+          final Map<String, String> smallFilesStats = new TreeMap<>();
+
+          for (String pName : names) {
```
dengzhhu653, line 293: why not `for (Partition p : plist)`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +246,76 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+            if (SmallFilesWarningUtil.smallAverageFilesDetected(SmallFilesWarningUtil.DEFAULT_MIN_FILES, threshold, params)) {
+              // We know stats exist and are well-formed; compute numbers for the value string
+              final long totalSize = Long.parseLong(params.get(StatsSetupConst.TOTAL_SIZE));
+              final long numFiles  = Long.parseLong(params.get(StatsSetupConst.NUM_FILES));
+              final long avg       = Math.floorDiv(totalSize, numFiles);
+              smallFilesStats.put(pName, "avgBytes=" + avg + ", partition total files=" + numFiles + ", totalBytes=" + totalSize);
+            }
+          }
+
+          msckInfo.setSmallFilesStats(smallFilesStats.isEmpty() ? Collections.emptyMap() : smallFilesStats);
```
dengzhhu653: why not `msckInfo.setSmallFilesStat(smallFilesStats)`

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -181,6 +185,21 @@ public void run() {
...
+          long threshold = conf.getLongVar(HiveConf.ConfVars.HIVE_MERGE_MAP_FILES_AVG_SIZE);
+          SmallFilesWarningUtil.smallFilesWarnings(parameters, 100L, threshold, who, "[ANALYZE][NOSCAN]")
+                                .ifPresent(msg -> {
+                                  LOG.info(msg);
+                                  SessionState ss = SessionState.get();
+                                  if (ss != null && ss.getConsole() != null) {
+                                    ss.getConsole().printInfo(msg);
+                                  }
+                                });
+        } catch (Throwable t) {
```
dengzhhu653: Is there any chance `SmallFilesWarningUtil.smallFilesWarnings` might throw the exception,  if so why not catch `Throwable` in `SmallFilesWarningUtil.smallFilesWarnings`?

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
```diff
@@ -182,6 +186,59 @@ public Object process(StatsAggregator statsAggregator) throws HiveException, Met
...
+
+        // Only warn when there are multiple files and total size > 0
+        if (numFiles > 100 && totalSize > 0) {
+          long threshold = (conf != null)
+                  ? conf.getLongVar(HiveConf.ConfVars.HIVE_MERGE_MAP_FILES_AVG_SIZE)
+                  : HiveConf.ConfVars.HIVE_MERGE_MAP_FILES_AVG_SIZE.defaultLongVal;
+
+          long avg = totalSize / numFiles;
+
+          if (avg <= threshold) {
```
dengzhhu653: here we can also use 
```
    String who = (partish.getPartition() == null) ? ("table " + partish.getTable().getFullyQualifiedName())
                                : ("partition " + partish.getPartition().getName());
          long threshold = conf.getLongVar(HiveConf.ConfVars.HIVE_MERGE_MAP_FILES_AVG_SIZE);
          SmallFilesWarningUtil.smallFilesWarnings(parameters, 100L, threshold, who, "[ANALYZE]")
                                .ifPresent(msg -> {
                                  LOG.warn(msg);
                                  SessionState ss = SessionState.get();
                                  if (ss != null && ss.getConsole() != null) {
                                    ss.getConsole().printInfo(msg);
                                  }
                                });
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: Acknowledged. The Batchable way is added.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: Acknowledged. Now using this:
final List<String> names = partsNotInMs.stream()
                  .map(CheckResult.PartitionResult::getPartitionName)
                  .filter(Objects::nonNull)
                  .filter(s -> !s.isEmpty())

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: They are pretty much the same. 
We need to do `smallFilesStats.put(pName, "avgBytes=" + avg + ", partition total files=" + numFiles + ", totalBytes=" + totalSize);` later, so it's easier to use pName to iterate.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
DanielZhu58: Actually there is no chance that `SmallFilesWarningUtil.smallFilesWarnings` will throw exception. 
The util does no I/O and no RPC—it just reads a Map, parses two longs, and formats a string. And all the possible failures haven been considered. 
params == null → returns Optional.empty()
missing keys → returns Optional.empty()
bad numbers → caught as NumberFormatException and returns Optional.empty()

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
DanielZhu58: Acknowledged. A new static method about printing info is added in BasicStatsNoJobTask.java.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -181,6 +186,15 @@ public void run() {
         }
         parameters.putAll(basicStatistics);
         StatsSetupConst.setBasicStatsState(parameters, StatsSetupConst.TRUE);
+        try {
+          String who = (partish.getPartition() == null) ? ("table " + partish.getTable().getFullyQualifiedName())
+                  : ("partition " + partish.getPartition().getName());
+          long threshold = conf.getLongVar(HiveConf.ConfVars.HIVE_MERGE_MAP_FILES_AVG_SIZE);
+          SmallFilesWarningUtil.smallFilesWarnings(parameters, 100L, threshold, who, "[ANALYZE][NOSCAN]")
+                                .ifPresent(smallFilesConsoleAndWarn(LOG));
+        } catch (Exception e) {
```
dengzhhu653: if there is no exception assumed from the block, why not move the `try{ } catch{}`?

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -293,6 +307,15 @@ public void run() {
         parameters.put(StatsSetupConst.NUM_FILES, String.valueOf(numFiles));
         parameters.put(StatsSetupConst.NUM_ERASURE_CODED_FILES, String.valueOf(numErasureCodedFiles));
 
+        try {
+          String who = (partish.getPartition() == null) ? ("table " + partish.getTable().getFullyQualifiedName())
+                                : ("partition " + partish.getPartition().getName());
+          long threshold = conf.getLongVar(HiveConf.ConfVars.HIVE_MERGE_MAP_FILES_AVG_SIZE);
+          SmallFilesWarningUtil.smallFilesWarnings(parameters, 100L, threshold, who, "[ANALYZE][NOSCAN]")
+                                .ifPresent(smallFilesConsoleAndWarn(LOG));
+        } catch (Throwable t) {
```
dengzhhu653: remove `catch (Throwable t)`?

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -307,7 +330,16 @@ public void run() {
         console.printInfo("[Warning] could not update stats for " + partish.getSimpleName() + ".", "Failed with exception " + e.getMessage() + "\n" + StringUtils.stringifyException(e));
       }
     }
+  }
 
+  public static Consumer<String> smallFilesConsoleAndWarn(Logger log) {
```
dengzhhu653: if this method is just for a simple message, we can just move it to the caller, e.g,
```
.ifPresent(s -> {
      SessionState ss = SessionState.get();
      if (ss != null && ss.getConsole() != null) {
        ss.getConsole().printInfo(msg);
      } else {
         LOG.warn(msg);
      }
 }
```

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/SmallFilesWarningUtil.java
```diff
@@ -0,0 +1,104 @@
...
+        return avg <= avgSizeThreshold;
+    }
+
+    /**
+     * Convenience overload: reads the threshold from {@code conf} and uses {@link #DEFAULT_MIN_FILES}.
+     * Returns false if {@code conf} is null.
+     */
+    public static boolean smallAverageFilesDetected(Configuration conf,
+                                                    Map<String, String> parameters) {
+        if (conf == null) return false;
```
dengzhhu653: why `conf` would be null?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +248,74 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+                  .map(CheckResult.PartitionResult::getPartitionName)
+                  .filter(Objects::nonNull)
+                  .filter(s -> !s.isEmpty())
+                  .collect(Collectors.toList());
+
+          // Batch over the names to limit single-RPC payload.
+          final int PART_FETCH_BATCH = 1000;
+          final Map<String, Partition> byName = new HashMap<>(names.size() * 2);
+          List<Partition> allParts = Batchable.runBatched(
+            PART_FETCH_BATCH,
```
dengzhhu653: PART_FETCH_BATCH -> `MetastoreConf.getIntVar(conf, MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX)`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +248,74 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+          for (String pName : names) {
+            final Partition p = byName.get(pName);
+            if (p == null) {
+              // not found / not visible yet
+              continue;
+            }
+
+            final Map<String, String> params = p.getParameters();
+            // Use util to decide if this partition should trigger the warning
+            if (SmallFilesWarningUtil.smallAverageFilesDetected(SmallFilesWarningUtil.DEFAULT_MIN_FILES, threshold, params)) {
```
dengzhhu653: nit: why not `SmallFilesWarningUtil. smallAverageFilesDetected(conf, params)`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +248,74 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
...
+            final String pName = Warehouse.makePartName(table.getPartitionKeys(), p.getValues());
+            byName.put(pName, p);
+          }
+
+          // Build small-files stats for partitions that have quick HMS stats.
+          final Map<String, String> smallFilesStats = new TreeMap<>();
+
+          for (String pName : names) {
+            final Partition p = byName.get(pName);
+            if (p == null) {
```
dengzhhu653: remove the `p == null` check, this shouldn't be null

DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/SmallFilesWarningUtil.java
DanielZhu58: Normally it would not be null.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
```diff
@@ -182,6 +187,27 @@ public Object process(StatsAggregator statsAggregator) throws HiveException, Met
...
+
+        SmallFilesWarningUtil.smallFilesWarnings(parameters, 100L, threshold, who, "[ANALYZE]")
+                .ifPresent(msg -> {
+                  LOG.warn(msg);
+                  SessionState ss = SessionState.get();
+                  if (ss != null && ss.getConsole() != null) {
+                    ss.getConsole().printInfo(msg);
+                  }
+                });
+      } catch (Exception e) {
```
dengzhhu653: nit: remove this `catch`?

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
```diff
@@ -30,6 +30,7 @@
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
```
dengzhhu653: nit: import is unnecessary

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -240,6 +248,68 @@ public int repair(MsckInfo msckInfo) throws TException, MetastoreException, IOEx
           }
         }
 
+        // Generate small files warnings only for partsNotInMs
+        try {
+          // Threshold in bytes for average file size considered "small"
+          final long threshold =
```
dengzhhu653: remove `threshold`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
```diff
@@ -26,24 +26,31 @@
...
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.TreeMap;
```
dengzhhu653, line 36: nit: unused import

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsNoJobTask.java
DanielZhu58: Acknowledged.

ql/src/java/org/apache/hadoop/hive/ql/stats/BasicStatsTask.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: Acknowledged.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/Msck.java
DanielZhu58: They are used around line 256, line 287.

dengzhhu653 (approved):

standalone-metastore/metastore-server/pom.xml
DanielZhu58: This dependency is now removed.

PR #5951, opened 2025-07-10
HIVE-29078: RuntimeException: null when starting standalone HMS

dengzhhu653:
In standalone Metastore, `new HiveConf(...)` would ignore the properties in `metastore-site.xml`

no

Run the standalone HMS with the same properties


zhangbutao:
If we must choose one between `hivemetastore-site.xml `and `metastore-site.xml`, would it be better to choose `hivemetastore-site.xml`? 
Because in my memory, the `hivemetastore-site.xml` was used more frequently.

dengzhhu653:
> If we must choose one between `hivemetastore-site.xml `and `metastore-site.xml`, would it be better to choose `hivemetastore-site.xml`? Because in my memory, the `hivemetastore-site.xml` was used more frequently.

I'd prefer the `metastore-site.xml` for the long term purpose, so far I haven't seen any user using the `hivemetastore-site.xml`. I can restore the HiveConf/MetastoreConf to take the `hivemetastore-site.xml` as the legacy property source if we are wondering some users might use this file.

zhangbutao:
> > If we must choose one between `hivemetastore-site.xml `and `metastore-site.xml`, would it be better to choose `hivemetastore-site.xml`? Because in my memory, the `hivemetastore-site.xml` was used more frequently.
> 
> I'd prefer the `metastore-site.xml` for the long term purpose, so far I haven't seen any user using the `hivemetastore-site.xml`. I can restore the HiveConf/MetastoreConf to take the `hivemetastore-site.xml` as the legacy property source if we are wondering some users might use this file.

Ambari has been consistently using `hivemetastore-site.xml` instead of `metastore-site.xml`, and there are many users of Ambari.
https://github.com/apache/ambari/blob/trunk/ambari-server/src/main/resources/stacks/BIGTOP/3.2.0/services/HIVE/configuration/hivemetastore-site.xml 

For now, having both `hivemetastore-site.xml` and `metastore-site.xml` take effect simultaneously seems like a better compromise. 

However, I agree that in the long term, we should retain only one configuration file. Perhaps we need more folks to discuss which configuration file to keep. :)

zhangbutao:
@dengzhhu653 
Since this PR just makes the metastore-site.xml to take effect, without removing the hivemetastore-site.xml. 

We can make some modifications to the description, removing the user-facing change.
```
yes, now hivemetastore-site.xml is no longer used for configuring the HMS
```

zhangbutao:
BTW, let's include this PR into branch-4.1.

cc @deniskuzZ ,please also take a look if this change is ok. Thanks.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PersistenceManagerProvider.java
```diff
@@ -653,7 +653,6 @@ private static <T> T retry(Supplier<T> s) {
             LOG.warn("Exception retry limit reached, not retrying any longer.", e);
           } else {
             LOG.debug("Non-retriable exception.", e);
-            break;
```
zhangbutao, line 656: Why remove the `break`?  When encountering an un-retriable exception, shouldn't we immediately exit the loop?

zhangbutao: BTW, Should we place `ex = e` before the` if else` block to ensure it executes even if a break is encountered?
```
@@ -649,13 +649,13 @@ private static <T> T retry(Supplier<T> s) {
           // If we're here, we'll proceed down the next while loop iteration.
         } else {
           // we've reached our limit, throw the last one.
+          ex = e;
           if (retriable) {
             LOG.warn("Exception retry limit reached, not retrying any longer.", e);
           } else {
             LOG.debug("Non-retriable exception.", e);
             break;
           }
-          ex = e;
         }
       }
     }
```

dengzhhu653: yeah, that's one reason, the break makes the real exception disappear, with `null` in this case.
believe it or not, in our tests we need the retry even on the fatal exception, the driver class might not be found due to the class loader the first time, a retry works afterwards

zhangbutao: I haven't studied which exceptions should be retried, but since we currently can't enumerate all retryable exceptions, it seems safer to retry for all exceptions.

I am ok to remove the break.

zhangbutao:
Please rebase this PR to get a green CI.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -93,6 +93,7 @@ public class HiveConf extends Configuration {
   private static URL hiveDefaultURL = null;
   private static URL hiveSiteURL = null;
   private static URL hivemetastoreSiteUrl = null;
+  private static URL newMetastoreSiteUrl = null;
```
deniskuzZ: why not simply `metastoreSiteUrl`?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -93,6 +93,7 @@ public class HiveConf extends Configuration {
   private static URL hiveDefaultURL = null;
   private static URL hiveSiteURL = null;
   private static URL hivemetastoreSiteUrl = null;
```
deniskuzZ: could we please change to camel-case? `hiveMetastoreSiteUrl` or maybe rename to `legacyMetastoreSiteUrl`

dengzhhu653: done

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -7026,15 +7033,15 @@ public static void setHiveSiteLocation(URL location) {
   }
 
   public static void setHivemetastoreSiteUrl(URL location) {
```
deniskuzZ, line 7040: Why `setHivemetastoreSiteUrl`, but `getMetastoreSiteLocation`? Do you think we could refactor here?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -7026,15 +7033,15 @@ public static void setHiveSiteLocation(URL location) {
...
   public static URL getHiveSiteLocation() {
     return hiveSiteURL;
   }
 
   public static URL getMetastoreSiteLocation() {
-    return hivemetastoreSiteUrl;
+    return metastoreSiteUrl;
   }
 
   public static URL getHiveServer2SiteLocation() {
```
deniskuzZ, line 7047: is it the same story as with `hivemetastore-site.xml`? is it even used? i don't see we use it in the `initialize`

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
dengzhhu653: This is the `public static` method that has been here for some time, I dont want to change it without strong reason

deniskuzZ: it's not used anywhere, just 3 tests. Try searching through Apache projects on GitHub. I only came across an older version of the Flink connector, and it doesn’t appear to be in recent one.
if you want to play safe, we can deprecate and add a new method with valid name.

dengzhhu653: done

deniskuzZ:
@zhangbutao, is this a duplicate of https://github.com/apache/hive/pull/5956?

zhangbutao:
> @zhangbutao, is this a duplicate of #5956?

Yes.

zhangbutao:
Closing this PR as it was superseded by https://github.com/apache/hive/pull/5956

PR #5953, opened 2025-07-10
HIVE-29080: Fix Incorrect results for queries on Decimal column Partitioned table with Vectorization enabled

Indhumathi27:
Updated the way decimal partition column values are set in vectorized execution. It ensures that the full decimal value, including its precision and scale, is correctly preserved when loading partition data into the vectorized engine.

The previous approach lost decimal precision and scale, causing incorrect results for joins and group by operations on partitioned decimal columns when vectorization is enabled. The fix ensures correct query results by accurately setting the decimal value.

No

Testcase added

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizedRowBatchCtx.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizedRowBatchCtx.java
@@ -576,7 +576,7 @@ public class VectorizedRowBatchCtx {
           dv.isNull[0] = true;
           dv.isRepeating = true;
         } else {
-          dv.fill(((HiveDecimal) value).longValue());
+          dv.fill((HiveDecimal) value);
         }
       } else {
         DecimalColumnVector dv = (DecimalColumnVector) col;
--- a/ql/src/test/queries/clientpositive/vector_decimal_partition.q
+++ b/ql/src/test/queries/clientpositive/vector_decimal_partition.q
@@ -14,3 +14,16 @@ INSERT INTO decimal_part PARTITION (nr_bank = 8801) VALUES (1, '8801');
 
 EXPLAIN VECTORIZATION EXPRESSION SELECT count(*), nr_bank FROM decimal_part GROUP BY nr_bank;
 SELECT count(*), nr_bank FROM decimal_part GROUP BY nr_bank;
+
+DROP TABLE IF EXISTS decimal_part1;
+
+CREATE EXTERNAL TABLE decimal_part1 (quantity INT) PARTITIONED BY (sales_percent DECIMAL(10,2)) STORED AS ORC;
+INSERT INTO decimal_part1 VALUES (1, 24518.01);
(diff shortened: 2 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/exec/vector/VectorizedRowBatchCtx.java
```diff
@@ -574,10 +574,12 @@ public void addPartitionColsToBatch(ColumnVector col, Object value, int colIndex
         if (value == null) {
           dv.noNulls = false;
           dv.isNull[0] = true;
-          dv.isRepeating = true;
         } else {
-          dv.fill(((HiveDecimal) value).longValue());
+          // fill serialized decimal value
+          dv.set(0, (HiveDecimal) value);
```
deniskuzZ: why do we hardcode with 0?

Indhumathi27: here, we are just filling partition vector value to 0th index. Before, dv.fill() method also sets the value to 0th index.

deniskuzZ: My bad, I missed that. Why not just `dv.fill((HiveDecimal) value)`
Should we change [here](https://github.com/apache/hive/pull/5953/files#diff-2773a0f7ae920d36ecdaa185be6df7007e45ce9fb184f2b8754dbfc3038c00b1R590) as well?

Indhumathi27: For DecimalColumnVector it is using SET method internally. so the issue is only with Decimal64ColumnVector

Indhumathi27: there is no dv.fill() implementation in Decimal64ColumnVector. Either i can implement this in Decimal64ColumnVector or keep current changes. what do you suggest @deniskuzZ  ?

deniskuzZ: ideally, we should replace the `fill(long value)` with
````
public void fill(HiveDecimal value) {
  isRepeating = true;
  isNull[0] = false;
  if (vector[0] == null) {
    vector[0] = new HiveDecimalWritable(value);
  }
  set(0, value);
}
````

Indhumathi27: Decimal64ColumnVector will convert the value along with precision and scale to Long value and set it to vector.
Modified the code to fit this case.

below suggested code is not required.
`  if (vector[0] == null) {
    vector[0] = new HiveDecimalWritable(value);
  }`

deniskuzZ: @Indhumathi27 could you please drop the `fill(long value)` from  Decimal64ColumnVector?

PR #4444, opened 2023-06-21
HIVE-12679: Allow users to be able to specify an implementation of IMetaStoreClient via HiveConf

okumin:
### What changes were proposed in this pull request?

Make it possible to replace the default IMetaStoreClient with a custom one.

It is the third time to try to send this patch. Austin originally opened [HIVE-12679](https://issues.apache.org/jira/browse/HIVE-12679) in 2016, and @moomindani tried to send his patch in #1402, and then I took over it again.

### Why are the changes needed?

In some environments, we want to connect to another data catalog rather than the native Hive Metastore. Looks like, [AWS Glue Data Catalog](https://github.com/awslabs/aws-glue-data-catalog-client-for-apache-hive-metastore/blob/branch-3.4.0/aws-glue-datacatalog-hive3-client/src/main/java/com/amazonaws/glue/catalog/metastore/AWSCatalogMetastoreClient.java) is one of the cases. We also have a similar case for some historical reasons.

As we can see, more than 40 people are watching [HIVE-12679](https://issues.apache.org/jira/browse/HIVE-12679), and some of them have asked us about the progress of this ticket. I'm sure it is worth maintaining this feature.

### Does this PR introduce _any_ user-facing change?

This doesn't change the original behavior at all.

### Is the change a dependency upgrade?
No

### How was this patch tested?
I tested HiveServer2 still works with this patch.

PR #1402, opened 2020-08-13
HIVE-12679: Allow users to be able to specify an implementation of IMetaStoreClient via HiveConf

moomindani:
### What changes were proposed in this pull request?

This change makes it possible for users to choose an implementation of IMetaStoreClient via HiveConf, i.e. hive-site.xml.

This patch is to retry merging the patch of HIVE-12679. 
I talked with Austin, the original contributor for this issue, and he said it is okay to submit this patch again. 
https://issues.apache.org/jira/browse/HIVE-12679

### Why are the changes needed?

Currently, in Hive the choice is hard coded to be SessionHiveMetaStoreClient in org.apache.hadoop.hive.ql.metadata.Hive. 
There is no other direct reference to SessionHiveMetaStoreClient other than the hard coded class name in Hive.java and the QL component operates only on the IMetaStoreClient interface so the change would be minimal and it would be quite similar to how an implementation of RawStore is specified and loaded in hive-metastore. One use case this change would serve would be one where a user wishes to use an implementation of this interface without the dependency on the Thrift server.

### Does this PR introduce _any_ user-facing change?

Yes. User will be able to specify own implementation of IMetaStoreClient. (e.g. IMetaStoreClient for AWS Glue Data Catalog).

### How was this patch tested?

Unit test and integration test with Glue Data Catalog client.

PR #5628, opened 2025-01-31
HIVE-28658 Add Iceberg REST Catalog client support

zratkai:
Change-Id: I5bb2559f7ca602b71f8ca03c852e2deff1a1bc52

### What changes were proposed in this pull request?
Iceberg REST client implementation added to support Iceberg REST server connection.

### Why are the changes needed?
To support Iceberg REST server connection.

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
Unit test.

PR #5955, opened 2025-07-11
HIVE-12679: Allow users to be able to specify an implementation of IMetaStoreClient via HiveConf

ngsg:

This patch aims to support configurable IMetaStoreClient.

This patch has been submitted three times previously. Austin originally proposed HIVE-12679 and submitted the initial patch in 2016. @moomindani later resubmitted it in https://github.com/apache/hive/pull/1402, and then @okumin took it over in https://github.com/apache/hive/pull/4444.

In some scenarios, users may want to run Hive as a general-purpose execution engine on top of a third-party data catalog. For instance, [AWS Glue Data Catalog](https://github.com/awslabs/aws-glue-data-catalog-client-for-apache-hive-metastore) is an example of such usage, and there is also an open ticket to implement an Iceberg REST catalog (HIVE-28658). As noted in the previous PR (https://github.com/apache/hive/pull/4444), more than 40 people are watching HIVE-12679, which I believe reflects strong demand for using Hive in this more flexible architecture. Therefore, I believe it is worth completing this patch to support external data catalogs in Hive.

No, users can use the existing HMS client without any changes in their configuration files. 
For the documentation fix, I will update hive-site once this patch is accepted and merged.

This patch includes two simple tests loading MetaStoreClient via `HiveMetaStoreClientFactory`.

---

Note for those familiar with the earlier patch:

- Regarding the design decision about MetaStoreClientFactory:
This patch does not support a customizable MSC factory as the original patch did; instead, it supports a customizable MSC. In other words, this patch introduces `metastore.client.class`, not `metastore.client.factory.class`.
I made this change not because I preferred this approach, but because I followed the majority opinion in the discussion [here]( https://github.com/apache/hive/pull/4444#pullrequestreview-1492711587). Personally, I think both approaches are valid and have their merits, and I do not have a strong opinion among them. I would appreciate it if you could share your thoughts on this decision.

- Regarding RetryingMetaStoreClient and other classes that instantiate MetaStoreClient ([link](https://github.com/apache/hive/pull/4444#pullrequestreview-1687155221)):
After resolving HIVE-27473, Hive now creates ThriftHiveMetaStoreClient only in `Hive.java` and `HiveMetaStoreClient`. Therefore, I believe it is sufficient to replace these two call sites with the new `HiveMetaStoreClientFactory`.


deniskuzZ:
Hi @ngsg, let's get https://github.com/apache/hive/pull/5924 in first. Please review it when time allows. 
Thank you!

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/client/HiveMetaStoreClientFactory.java
```diff
@@ -0,0 +1,58 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * A factory class creating a MetaStoreClient specified in a given configuration.
+ */
+public class HiveMetaStoreClientFactory {
+  private static final Logger LOG = LoggerFactory.getLogger(HiveMetaStoreClientFactory.class);
+
+  public static IMetaStoreClient newClient(Configuration conf, boolean allowEmbedded) throws MetaException {
+    String mscClassName = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.METASTORE_CLIENT_CLASS);
```
okumin: Can we use MetastoreConf.getClass?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1728,6 +1728,10 @@ public enum ConfVars {
                     " and password. Any other value is ignored right now but may be used later."
                 + "If JWT- Supported only in HTTP transport mode. If set, HMS Client will pick the value of JWT from "
                 + "environment variable HMS_JWT and set it in Authorization header in http request"),
+    METASTORE_CLIENT_CLASS("metastore.client.class",
+        "hive.metastore.client.class",
+        "org.apache.hadoop.hive.metastore.client.ThriftHiveMetaStoreClient",
+        "The name of MetaStoreClient class that implements the IMetaStoreClient interface."),
```
okumin, line 1734: @moomindani Please feel free to give us a suggestion if you have any thoughts.

> Regarding the design decision about MetaStoreClientFactory:
This patch does not support a customizable MSC factory as the original patch did; instead, it supports a customizable MSC. In other words, this patch introduces metastore.client.class, not metastore.client.factory.class.
I made this change not because I preferred this approach, but because I followed the majority opinion in the discussion https://github.com/apache/hive/pull/4444#pullrequestreview-1492711587. Personally, I think both approaches are valid and have their merits, and I do not have a strong opinion among them. I would appreciate it if you could share your thoughts on this decision.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/client/HiveMetaStoreClientFactory.java
ngsg: Sure, I'll modify it to use `getClass` once HIVE-20189 is merged.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
moomindani: Thanks, LGTM, I have no concerns around this.

deniskuzZ:
[HIVE-20189](https://issues.apache.org/jira/browse/HIVE-20189) is merged, @ngsg could you please rebase

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
dengzhhu653: I have a concern here: if we specify the client via the configuration, it means each time we are allowed to query one type of metadata only. For example, we set the client to Glue for the tables stored in Glue, if we need the Hive tables later on in the same session, we should reset the client to Hive, and just think loud, if the SQL contains the Hive and Glue table?

From my point of view, I would suggest the pattern(the database or table) or the catalog routed way to choose which the client will use to obtain the metadata.

ngsg: > From my point of view, I would suggest the pattern(the database or table) or the catalog routed way to choose which the client will use to obtain the metadata.

I believe this is somewhat similar to the design proposed in HIVE-28879 (Federated Catalog).
From my perspective, I see HIVE-12679 as a milestone for both HIVE-28658 (Iceberg REST Catalog, #5628) and HIVE-28879. So, I'm fine with moving directly to HIVE-28879 or pursuing any other approach, as long as we aim to support third-party data catalogs. However, since this ticket predates my contributions and has many watchers, I want to ask for and respect others' opinions as well.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/builder/HiveMetaStoreClientBuilder.java
```diff
@@ -18,65 +18,89 @@
...
-        this.client = SynchronizedMetaStoreClient.newClient(conf, client);
-        return this;
-    }
+  public IMetaStoreClient build() {
+    return Objects.requireNonNull(client);
+  }
 
-    public IMetaStoreClient build() {
-        return Objects.requireNonNull(client);
+  public static IMetaStoreClient createClient(Configuration conf, boolean allowEmbedded) throws MetaException {
```
deniskuzZ: should be private

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -61,19 +61,24 @@ public HiveMetaStoreClient(Configuration conf, HiveMetaHookLoader hookLoader) th
 
   public HiveMetaStoreClient(Configuration conf, HiveMetaHookLoader hookLoader, Boolean allowEmbedded)
     throws MetaException {
-    this(conf, hookLoader, new ThriftHiveMetaStoreClient(conf, allowEmbedded));
+    this(conf, hookLoader, HiveMetaStoreClientBuilder.createClient(conf, allowEmbedded));
```
deniskuzZ: `new HiveMetaStoreClientBuilder(conf).newClient(allowEmbedded).build()`

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -61,19 +61,24 @@ public HiveMetaStoreClient(Configuration conf, HiveMetaHookLoader hookLoader) th
...
   }
 
   private HiveMetaStoreClient(Configuration conf, HiveMetaHookLoader hookLoader,
-      ThriftHiveMetaStoreClient thriftClient) {
-    super(createUnderlyingClient(conf, hookLoader, thriftClient), conf);
-    this.thriftClient = thriftClient;
+      IMetaStoreClient baseMetaStoreClient) {
+    super(createUnderlyingClient(conf, hookLoader, baseMetaStoreClient), conf);
+
+    if (baseMetaStoreClient instanceof ThriftHiveMetaStoreClient) {
```
deniskuzZ:
````
if (baseMetaStoreClient instanceof ThriftHiveMetaStoreClient metaStoreClient) {
  this.thriftClient = metaStoreClient;
}
````
and drop the final

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java [hunk 6]
```diff
@@ -82,19 +87,35 @@ private static IMetaStoreClient createUnderlyingClient(Configuration conf, HiveM
   // methods for test
 
   public boolean createType(Type type) throws TException {
-    return thriftClient.createType(type);
+    if (thriftClient != null) {
```
deniskuzZ, line 93: how about using generic method?
````
@FunctionalInterface
interface ThriftCallable<T> {
  T call() throws TException;
}

private <T> T doCall(ThriftCallable<T> callable) throws TException {
  if (thriftClient != null) {
    return callable.call();
  } else {
    throw new UnsupportedOperationException();
  }
}

public boolean createType(Type type) throws TException {
  return doCall(() -> thriftClient.createType(type));
}
````

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: @ngsg maybe `METASTORE_CLIENT_IMPL`, similar to `RAW_STORE_IMPL`?

deniskuzZ: IMO, [HIVE-28879](https://issues.apache.org/jira/browse/HIVE-28879) will likely require significantly more effort, so we can consider this PR as a intermediate step forward.
however, maybe we can introduce a static [DbCatalog -> Client] mapping here, WDYT?

dengzhhu653: how about a new property to configure those mappings? e.g, `metastore.client.meta.class.mappings`=`catalog:client_classname,catalog.db(or db_pattern):client_classname,catalog.db.table:client_classname`

deniskuzZ: I don't think we need multiple client impl for the same catalog (aka Snowflake, Glue, HMS, Rest, etc) 
Instead, I’d suggest using a simple comma-separated list. Later, that would be persisted to the backend database during the Catalog registration (HIVE-26227)
````
METASTORE_CLIENT_IMPL("metastore.client.impl", catalog:client_classname)
````

deniskuzZ: Maybe we should address this in [HIVE-28879](https://issues.apache.org/jira/browse/HIVE-28879), since we also need to provide connection details per catalog

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/builder/HiveMetaStoreClientBuilder.java
ngsg: fixed

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
ngsg: fixed

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
ngsg: Fixed as suggested. I wasn't aware of that syntax before. Thank you for the helpful suggestion.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
ngsg: Fixed to use generic method.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
ngsg: > @ngsg maybe `METASTORE_CLIENT_IMPL`, similar to `RAW_STORE_IMPL`?

I changed `METASTORE_CLINT_CLASS` to `METASTORE_CLIENT_IMPL` and `metastore.client.class` to `metastore.client.impl`.

---

> how about a new property to configure those mappings? e.g, `metastore.client.meta.class.mappings`=`catalog:client_classname,catalog.db(or db_pattern):client_classname,catalog.db.table:client_classname`

Regarding the mapping, if there were no ThriftHiveMetaStoreClient-related logic in HiveMetaStoreClient, I think I could implement a mapping from a catalog to the actual client. However, due to that logic, I don't currently have a clear design to support it. Let me think more about a better design and share my thoughts once I’ve summarized them.

---

Another thought I had is that if we eventually implement HIVE-28879, we may not need this configuration key to specify either a single class or a mapping, since the relevant information would be stored in the MetaStore backend database. Given that we likely have enough time before the next release, I'm unsure about the feasibility of introducing a temporary configuration key that won't appear in any release. What do you think about keeping this patch as a fallback option in case HIVE-28879 isn't ready in time for the next release, and proceeding directly with HIVE-28879 for now?

deniskuzZ: > Another thought I had is that if we eventually implement [HIVE-28879](https://issues.apache.org/jira/browse/HIVE-28879), we may not need this configuration key to specify either a single class or a mapping, since the relevant information would be stored in the MetaStore backend database.

Totally agree + the proposed mapping might require connection details as well. I think we should go ahead and merge this PR, and start thinking on a more generic solution like [HIVE-28879](https://issues.apache.org/jira/browse/HIVE-28879)

deniskuzZ (approved):
+1

ayushtkn:
This seems good to go? if yes, @ngsg mind hitting the merge button?

ngsg:
@ayushtkn,  thanks for the reminder. I'll merge it shortly.

ngsg:
Thanks all for the review!

---

I have read the [guide](https://hive.apache.org/docs/latest/howtocommit_27362108/) and tried to follow the documented process for merging. Since this is my first time merging, I would appreciate it if someone could check whether I've made any mistakes I might have overlooked.

ayushtkn:
> I have read the [guide](https://hive.apache.org/docs/latest/howtocommit_27362108/) and tried to follow the documented process for merging. Since this is my first time merging, I would appreciate it if someone could check whether I've made any mistakes I might have overlooked.

@ngsg  It is all good

PR #5959, opened 2025-07-11
HIVE-29083: Iceberg: Unable to create table using HadoopCatalog

deniskuzZ:

Set the table location based on iceberg.catalog.\<name\>.warehouse config

BugFix: `Caused by: java.lang.IllegalArgumentException: Table location not set`

No

mvn test -Dtest=TestIcebergLlapLocalCliDriver -Dqfile=hadoop_catalog_create_table.q

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
@@ -242,16 +242,18 @@ public class HiveIcebergMetaHook implements HiveMetaHook {
 
       // If not using HiveCatalog check for existing table
       try {
-
         this.icebergTable = IcebergTableUtil.getTable(conf, catalogProperties, true);
 
+        if (Catalogs.hadoopCatalog(conf, catalogProperties) && hmsTable.getSd() != null &&
+                hmsTable.getSd().getLocation() == null) {
+          hmsTable.getSd().setLocation(icebergTable.location());
+        }
         Preconditions.checkArgument(catalogProperties.getProperty(InputFormatConfig.TABLE_SCHEMA) == null,
             "Iceberg table already created - can not use provided schema");
         Preconditions.checkArgument(catalogProperties.getProperty(InputFormatConfig.PARTITION_SPEC) == null,
             "Iceberg table already created - can not use provided partition specification");
 
         LOG.info("Iceberg table already exists {}", icebergTable);
-
         return;
       } catch (NoSuchTableException nte) {
         // If the table does not exist we will create it below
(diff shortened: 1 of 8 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -815,8 +815,8 @@ private void assertFileFormat(String format) {
   }
 
   private void setCommonHmsTablePropertiesForIceberg(org.apache.hadoop.hive.metastore.api.Table hmsTable) {
-    // If the table is not managed by Hive catalog then the location should be set
-    if (!Catalogs.hiveCatalog(conf, catalogProperties)) {
+    // If the table is not managed by Hive or Hadoop catalog, then the location should be set
+    if (!Catalogs.hiveCatalog(conf, catalogProperties) && !Catalogs.hadoopCatalog(conf, catalogProperties)) {
```
zhangbutao: I don't get this change. I still can create Hadoop catalog tables without setting the location.
```
set iceberg.catalog.ice01.type=hadoop;
set iceberg.catalog.ice01.warehouse=hdfs://127.0.0.1:8020/tmp;

CREATE EXTERNAL TABLE orders (orderid INT, quantity INT, itemid INT, tradets TIMESTAMP) PARTITIONED BY (p1 STRING, p2 STRING) STORED BY ICEBERG STORED AS ORC;
```

deniskuzZ:
````
 org.apache.hadoop.hive.ql.exec.tez.TezRuntimeException: Vertex failed, vertexName=Map 1, vertexId=vertex_1752223623655_0001_3_00, diagnostics=[Vertex vertex_1752223623655_0001_3_00 [Map 1] killed/failed due to:ROOT_INPUT_INIT_FAILURE, Vertex Input: orders initializer failed, vertex=vertex_1752223623655_0001_3_00 [Map 1], java.io.IOException: cannot find dir = /tmp/default/orders in pathToPartitionInfo: [file:/Users/dkuzmenko/projects/hive/upstream/hive/itests/qtest-iceberg/target/localfs/warehouse/orders]
````
table dir should be pre-created

Comment (zhangbutao) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java]: My bad, i just forget to add TBLPROPERTIES when creating table:

```
set iceberg.catalog.ice01.type=hadoop;
set iceberg.catalog.ice01.warehouse=hdfs://127.0.0.1:8020/tmp;

CREATE EXTERNAL TABLE orders (orderid INT, quantity INT, itemid INT, tradets TIMESTAMP) PARTITIONED BY (p1 STRING, p2 STRING) STORED BY ICEBERG STORED AS ORC TBLPROPERTIES('format-version'='2', 'iceberg.catalog'='ice01');
```

Can not create Hadoop catalog tables without setting the location. Will throw exception:
`Caused by: java.lang.IllegalArgumentException: Table location not set`

deniskuzZ:
@zhangbutao should we include this into 4.1?

zhangbutao:
> @zhangbutao should we include this into 4.1?

@deniskuzZ Sure.

zhangbutao:
Some related failed were related this change.

Since the Hadoop Catalog is relatively less used, and users can also create Hadoop Catalog tables by specifying location. So let's just incorporate this fix in version master branch, no need inclued into branch-4.1.

deniskuzZ:
> Some related failed were related this change.
> 
> Since the Hadoop Catalog is relatively less used, and users can also create Hadoop Catalog tables by specifying location. So let's just incorporate this fix in version master branch, no need inclued into branch-4.1.

it was a faulty test, fixed it. 
@zhangbutao, you can't create Hadoop Catalog tables by specifying a custom location, see
````
@Override
    public TableBuilder withLocation(String location) {
      Preconditions.checkArgument(
          location == null || location.equals(defaultLocation),
          "Cannot set a custom location for a path-based table. Expected "
              + defaultLocation
              + " but got "
              + location);
      return this;
    }
````

PR #5961, opened 2025-07-11
HIVE-29006: Iceberg: V3 support deletion vectors

deniskuzZ:

Added base for Iceberg V3 deletion vectors support

Iceberg V3 deletion vectors support

No

mvn test -Dtest=TestIcebergLlapLocalCliDriver -Dqfile=iceberg_v3_deletion_vectors.q -Drat.skip=true


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -392,27 +396,47 @@ public static void cherryPick(Table table, long snapshotId) {
   }
 
   public static boolean isV2Table(Map<String, String> props) {
```
SourabhBadhya: Maybe you can rename the function as `isAtleastV2Table()` or something on these lines.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -1641,6 +1645,12 @@ static void overlayTableProperties(Configuration configuration, TableDesc tableD
 
       // serialize table object into config
       Table serializableTable = SerializableTable.copyOf(table);
+      // set table format-version and write-mode information from tableDesc
+      List<String> writeConfigList = ImmutableList.of(
+          FORMAT_VERSION, DELETE_MODE, UPDATE_MODE, MERGE_MODE);
+      if (IcebergTableUtil.isV2Table(props::getProperty)) {
```
SourabhBadhya: Is it necessary to create this function - Cant we simply pass as map which is already a defined util function instead of passing a binary operator. Even if we want to use a BinaryOperator, can we remove the isV2Table() which takes in a map? Or is this publicly defined function which can't be removed?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -392,27 +396,47 @@ public static void cherryPick(Table table, long snapshotId) {
...
-        mode = props.apply(TableProperties.MERGE_MODE,
-            TableProperties.MERGE_MODE_DEFAULT);
-        break;
+  public static boolean isV2Table(BinaryOperator<String> props) {
+    return IcebergTableUtil.formatVersion(props) >= 2;
+  }
+
+  public static Integer formatVersion(Map<String, String> props) {
+    if (props == null) {
+      return 2; // default to V2
```
SourabhBadhya: nit: Maybe you can add a TODO stating that we have to move the default to 3 whenever applicable?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
```diff
@@ -392,27 +396,47 @@ public static void cherryPick(Table table, long snapshotId) {
...
     }
-    return RowLevelOperationMode.COPY_ON_WRITE.modeName().equalsIgnoreCase(mode);
+    return IcebergTableUtil.formatVersion(props::getOrDefault);
+  }
+
+  private static Integer formatVersion(BinaryOperator<String> props) {
```
SourabhBadhya, line 422: Same argument here - should we really create another function which takes in a BinaryOperator?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergWriterBase.java
```diff
@@ -108,19 +115,53 @@ static PartitioningWriter<Record, DataWriteResult> newDataWriter(
...
+        PreviousDeleteLoader.create(table, rewritableDeletes);
     FileIO io = table.io();
     boolean inputOrdered = context.inputOrdered();
     long targetFileSize = context.targetDeleteFileSize();
     DeleteGranularity deleteGranularity = context.deleteGranularity();
 
-    if (inputOrdered) {
+    if (context.useDVs()) {
+      return new PartitioningDVWriter<>(files, previousDeleteLoader);
+    } else if (inputOrdered && rewritableDeletes == null) {
```
SourabhBadhya, line 131: Is it because ClusteredDeleteWriter doesn't support passing deletion vectors?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/WriterBuilder.java
```diff
@@ -120,27 +132,69 @@ public HiveIcebergWriter build() {
...
+                operation.name());
+      };
     }
 
     WriterRegistry.registerWriter(attemptID, tableName, writer);
     return writer;
   }
 
+  private Map<String, DeleteFileSet> rewritableDeletes() {
+    TableScan scan = table.newScan().caseSensitive(false).ignoreResiduals();
```
SourabhBadhya: Is it right that we are ignoring residual filters? Is the idea here to just use all the rows applicable for deletion?

iceberg/iceberg-handler/src/test/queries/positive/iceberg_v3_deletion_vectors.q
```diff
@@ -0,0 +1,75 @@
...
+ insert into ice01 values (5),(6),(7),(8);
+
+-- check the values, only the delete value shouldn't be there
+select * from ice01 order by id;
+
+-- delete one value
+delete from ice01 where id=7;
+
+-- check the entries, the deleted entries shouldn't be there.
+select * from ice01 order by id;
```
SourabhBadhya, line 45: Can we also display content offset for each of this delete tests using Position delete metadata tables, so that we are sure that deletion vectors are written in action.

Reference on deletion vectors metadata columns:
https://github.com/apache/iceberg/commit/026a9b00459ae458eb4c6c620c78742dc43a00e8

SourabhBadhya (commented):
This PR is in good shape. Is there any changes that needs to be made in the read side. Are the existing readers already using the information of deletion vectors and the implementation is within Iceberg?

Also something to use - I see a new interface with a functionality to load equalityDeletes. 
https://github.com/apache/iceberg/blob/77f1f5ba47f0e5250b8af4e16cbe6febc8244ee5/data/src/main/java/org/apache/iceberg/data/DeleteLoader.java
I am not sure, if that is required as part of this. But it is worth giving an investigation.

iceberg/iceberg-handler/src/test/queries/positive/iceberg_v3_deletion_vectors.q
deniskuzZ: @SourabhBadhya, wdym? select doesn't return the deleted rows

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergWriterBase.java
deniskuzZ: it's done the same way in SparkWriter

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
deniskuzZ: we don't always have a Map, sometimes it's a Properties object

iceberg/iceberg-handler/src/test/queries/positive/iceberg_v3_deletion_vectors.q
SourabhBadhya: Maybe add tests which does a select query on the position delete metadata tables which contains information like content offset etc.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
deniskuzZ: props is of Properties type, not Map

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java
deniskuzZ: renamed to `isV2TableOrAbove`

iceberg/iceberg-handler/src/test/queries/positive/iceberg_v3_deletion_vectors.q
deniskuzZ: added

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/WriterBuilder.java
deniskuzZ: refactored and replaced with BatchScan

deniskuzZ:
> Also something to use - I see a new interface with a functionality to load equalityDeletes. https://github.com/apache/iceberg/blob/77f1f5ba47f0e5250b8af4e16cbe6febc8244ee5/data/src/main/java/org/apache/iceberg/data/DeleteLoader.java I am not sure, if that is required as part of this. But it is worth giving an investigation.

`DeleteLoader` is used for caching and we already implement it in Hive, please see [CachingDeleteLoader](https://github.com/apache/hive/blob/master/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/data/CachingDeleteLoader.java)

SourabhBadhya (approved):
LGTM +1.

deniskuzZ:
Thanks for the review, @SourabhBadhya ! I'll merge it shortly.

PR #5966, opened 2025-07-12
HIVE-29082: Enhance HiveAuthzContext with a Client Configuration Map

jjiang037:

Enhance HiveAuthzContext with a Client Configuration Map

Adding a map to hold client additional information in HiveAuthzContext will improve auditing and authorization flexibility by making the system aware of the client application making a request

No

Regression Tests

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
@@ -79,6 +79,24 @@ import java.util.stream.Collectors;
 public class HiveMetaStoreAuthorizer extends MetaStorePreEventListener implements MetaStoreFilterHook {
   private static final Logger LOG = LoggerFactory.getLogger(HiveMetaStoreAuthorizer.class);
 
+  /**
+   * The client configuration.
+   */
+  private static final ThreadLocal<Map<String, Object>> cConfig = new ThreadLocal<Map<String, Object>>() {
+    @Override
+    protected Map<String, Object> initialValue() {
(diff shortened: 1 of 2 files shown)


ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizer.java
```diff
@@ -457,6 +486,17 @@ private List<String> filterTableNames(HiveMetaStoreAuthzInfo hiveMetaStoreAuthzI
...
+      if (clientConfig != null) {
+        HiveAuthzContext.Builder builder = new HiveAuthzContext.Builder();
+        builder.setClientConfig(clientConfig);
+        if (hiveAuthzContext != null) {
+          builder.setCommandString(hiveAuthzContext.getCommandString());
+          builder.setUserIpAddress(hiveAuthzContext.getIpAddress());
+          builder.setForwardedAddresses(hiveAuthzContext.getForwardedAddresses());
+        }
+        hiveAuthzContext = builder.build();
+      }
```
saihemanth-cloudera, line 499: why not directly do this?:
`Map<String, Object> clientConfig = cConfig.get();
if (clientConfig != null) {
hiveAuthzContext. setClientConfig(clientConfig);
}`

jjiang037: I didn't do hiveAuthzContext.setClientConfig(clientConfig) is because HiveAuthzContext is designed as an immutable class, the setter methods are only on the builder class, not on the main HiveAuthzContext class

saihemanth-cloudera (approved):
LGTM +1

PR #5968, opened 2025-07-13
HIVE-29086: Fix ClassNotFoundException: org.apache.tez.dag.DAG when engine is mr

Abyss-lord:

Fix ClassNotFoundException: org.apache.tez.dag.DAG when engine is mr

fix: [HIVE-29086](https://issues.apache.org/jira/projects/HIVE/issues/HIVE-29086)

no 

local test.

### DIFF
--- a/cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
+++ b/cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
@@ -396,8 +396,8 @@ public class CliDriver {
 
           // First, kill any running MR jobs
           HadoopJobExecHelper.killRunningJobs();
-          //when engine is not tez,it will throw ClassNotFoundException: org.apache.tez.dag.DAG
-          if ("tez".equalsIgnoreCase(conf.get("hive.execution.engine"))) {
+          if ("tez".equalsIgnoreCase(HiveConf.getVar(conf,
+                  HiveConf.ConfVars.HIVE_EXECUTION_ENGINE))) {
             TezJobExecHelper.killRunningJobs();
           }
           HiveInterruptUtils.interrupt();


cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
```diff
@@ -396,7 +396,10 @@ public void handle(Signal signal) {
 
           // First, kill any running MR jobs
           HadoopJobExecHelper.killRunningJobs();
-          TezJobExecHelper.killRunningJobs();
+          //when engine is not tez,it will throw ClassNotFoundException: org.apache.tez.dag.DAG
```
ayushtkn: this comment isn't required

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
```diff
@@ -396,7 +396,10 @@ public void handle(Signal signal) {
...
+          if ("tez".equalsIgnoreCase(conf.get("hive.execution.engine"))) {
```
ayushtkn: Change it to
```
"tez".equalsIgnoreCase(HiveConf.getVar(conf, HiveConf.ConfVars.HIVE_EXECUTION_ENGINE))
```

ayushtkn (commented):
Can you put the trace in the ticket & a sample query to reproduce this

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
Abyss-lord: Fix

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
Abyss-lord: Fix

Abyss-lord:
> Can you put the trace in the ticket & a sample query to reproduce this

I currently don’t have access to the trace — it’s on my work laptop. I’ll retrieve it first thing Monday and share both the trace and a sample query to help reproduce the issue.

Thanks for your patience!

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
```diff
@@ -396,7 +396,10 @@ public void handle(Signal signal) {
 
           // First, kill any running MR jobs
           HadoopJobExecHelper.killRunningJobs();
-          TezJobExecHelper.killRunningJobs();
+          if ("tez".equalsIgnoreCase(HiveConf.getVar(conf,
```
deniskuzZ, line 399: please avoid hardcoded literals

zhangbutao:
I think if the Hadoop cluster's classpath does not include the Tez packages and in case of MR engine, this exception will occur.

cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java
zhangbutao: It seems that many places in the Hive code use hardcoded `mr `or `tez `constants. Perhaps we can create a separate PR to fix the all the `mr `& `tez `hardcoding issue. :)

zhangbutao (approved):
+1 LGTM

PR #5969, opened 2025-07-13
HIVE-29087: Add MAX_LIFE_TIME configuration for hikaricp connection pool

Abyss-lord:

Add `MAX_LIFE_TIME` configuration for hikaricp connection pool.

fix: [HIVE-29087](https://issues.apache.org/jira/projects/HIVE/issues/HIVE-29087)

no

local test.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
@@ -22,6 +22,7 @@ import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.metastore.DatabaseProduct;
+import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.metastore.metrics.Metrics;
 import org.apache.hadoop.hive.metastore.utils.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/datasource/HikariCPDataSourceProvider.java
```diff
@@ -49,6 +50,8 @@ public DataSource create(Configuration hdpConfig, int maxPoolSize) throws SQLExc
     String driverUrl = DataSourceProvider.getMetastoreJdbcDriverUrl(hdpConfig);
     String user = DataSourceProvider.getMetastoreJdbcUser(hdpConfig);
     String passwd = DataSourceProvider.getMetastoreJdbcPasswd(hdpConfig);
+    long maxLifeTime = MetastoreConf.getLongVar(hdpConfig,
+            MetastoreConf.ConfVars.METASTORE_HIKARICP_MAX_LIFE_TIME);
```
ayushtkn, line 54: Do we need a seprate config, there is a logic below
```
    Properties properties = replacePrefix(DataSourceProvider.getPrefixedProperties(hdpConfig, HIKARI));

    HikariConfig config;
    try {
      config = new HikariConfig(properties);
```

I believe that should take care of setting maxLifetime

okumin: No, we don't, need to have it, probably. Please see the discussion in this PR and the test case I added.
https://github.com/apache/hive/pull/5639
https://github.com/apache/hive/pull/5639/files#diff-21013298e701863201669c0a08dce28bff302bca654dda0287b76cb344ae2c7a

okumin:
@Abyss-lord
Do you agree with the solution using `hikaricp..maxLifetime`? If the answer is yes, we may close this pull request 👍

Abyss-lord:
> @Abyss-lord Do you agree with the solution using `hikaricp..maxLifetime`? If the answer is yes, we may close this pull request 👍

Yes, I agree, plz help to close this pr and jira.

PR #5973, opened 2025-07-14
HIVE-28724:CREATE TABLE is not handled by DDLSemanticAnalyzerFactory

ramitg254:

CREATE TABLE is handled by DDLSemanticAnalyzerFactory as it is a ddl statement

Earlier handles function of  DDLSemanticAnalyzerFactory returns false for TOK_CREATETABLE

No

local build and unit tests

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyzer.java
@@ -1,0 +1,818 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.hive.ql.ddl.table.create;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.common.TableName;
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.metastore.TableType;
+import org.apache.hadoop.hive.metastore.TransactionalValidationListener;
+import org.apache.hadoop.hive.metastore.Warehouse;
+import org.apache.hadoop.hive.metastore.api.Database;
+import org.apache.hadoop.hive.metastore.api.FieldSchema;
+import org.apache.hadoop.hive.metastore.api.MetaException;
+import org.apache.hadoop.hive.metastore.api.Order;
+import org.apache.hadoop.hive.metastore.api.SQLCheckConstraint;
+import org.apache.hadoop.hive.metastore.api.SQLDefaultConstraint;
+import org.apache.hadoop.hive.metastore.api.SQLForeignKey;
+import org.apache.hadoop.hive.metastore.api.SQLNotNullConstraint;
+import org.apache.hadoop.hive.metastore.api.SQLPrimaryKey;
+import org.apache.hadoop.hive.metastore.api.SQLUniqueConstraint;
+import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
+import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
+import org.apache.hadoop.hive.ql.ErrorMsg;
+import org.apache.hadoop.hive.ql.QueryState;
+import org.apache.hadoop.hive.ql.ddl.DDLSemanticAnalyzerFactory.DDLType;
+import org.apache.hadoop.hive.ql.ddl.DDLWork;
+import org.apache.hadoop.hive.ql.ddl.misc.sortoder.SortFieldDesc;
+import org.apache.hadoop.hive.ql.ddl.misc.sortoder.SortFields;
+import org.apache.hadoop.hive.ql.ddl.table.constraint.ConstraintsUtils;
+import org.apache.hadoop.hive.ql.ddl.table.convert.AlterTableConvertOperation;
+import org.apache.hadoop.hive.ql.ddl.table.create.like.CreateTableLikeDesc;
+import org.apache.hadoop.hive.ql.ddl.table.storage.skewed.SkewedTableUtils;
+import org.apache.hadoop.hive.ql.exec.TaskFactory;
+import org.apache.hadoop.hive.ql.io.SchemaInferenceUtils;
+import org.apache.hadoop.hive.ql.metadata.HiveException;
+import org.apache.hadoop.hive.ql.metadata.HiveStorageHandler;
+import org.apache.hadoop.hive.ql.metadata.HiveUtils;
+import org.apache.hadoop.hive.ql.metadata.SessionHiveMetaStoreClient;
+import org.apache.hadoop.hive.ql.metadata.Table;
+import org.apache.hadoop.hive.ql.parse.ASTNode;
+import org.apache.hadoop.hive.ql.parse.CalcitePlanner;
+import org.apache.hadoop.hive.ql.parse.EximUtil;
+import org.apache.hadoop.hive.ql.parse.HiveParser;
+import org.apache.hadoop.hive.ql.parse.PartitionTransform;
+import org.apache.hadoop.hive.ql.parse.QB;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+import org.apache.hadoop.hive.ql.parse.StorageFormat;
+import org.apache.hadoop.hive.ql.parse.TableMask;
+import org.apache.hadoop.hive.ql.plan.HiveOperation;
+import org.apache.hadoop.hive.ql.session.SessionState;
+import org.apache.hadoop.hive.ql.session.SessionStateUtil;
+import org.apache.hadoop.hive.ql.util.NullOrdering;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.base.Strings;
+
+import static org.apache.hadoop.hive.conf.HiveConf.ConfVars.HIVE_DEFAULT_STORAGE_HANDLER;
+import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.DEFAULT_TABLE_TYPE;
+import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.META_TABLE_LOCATION;
+import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.META_TABLE_STORAGE;
+import static org.apache.hadoop.hive.metastore.api.hive_metastoreConstants.TABLE_IS_CTAS;
+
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyser.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.hadoop.hive.ql.QueryState;
+import org.apache.hadoop.hive.ql.ddl.DDLSemanticAnalyzerFactory.DDLType;
+import org.apache.hadoop.hive.ql.parse.ASTNode;
+import org.apache.hadoop.hive.ql.parse.CalcitePlanner;
+import org.apache.hadoop.hive.ql.parse.HiveParser;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+
+@DDLType(types = HiveParser.TOK_CREATETABLE)
+public class CreateTableAnalyser extends CalcitePlanner {
```
abstractdog: nit: we tend to use the Analyzer instead of Analyser

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -14077,9 +14077,7 @@ ASTNode analyzeCreateTable(
       storageFormat.setStorageHandler(defaultStorageHandler);
       isUserStorageFormat = true;
     }
-
-    // CREATE TABLE is a DDL and yet it's not handled by DDLSemanticAnalyzerFactory
-    // FIXME: HIVE-28724
+    
     queryProperties.setQueryType(QueryProperties.QueryType.DDL);
```
abstractdog: if the this patch works properly, this line is not needed anymore:
```
    queryProperties.setQueryType(QueryProperties.QueryType.DDL);
```

and the correctness will be tested automatically here:
https://github.com/apache/hive/blob/f4ab76bf51859f8a2f4e929b2163bb891d17008f/ql/src/test/org/apache/hadoop/hive/ql/parse/TestSemanticAnalyzer.java#L391-L393

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyser.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public CreateTableAnalyser(QueryState queryState)
```
abstractdog: It does look a bit odd that the new class doesn't extend anything.

I believe one of the key goals of decoupling the DDL analyzers is to offload logic from the massive SemanticAnalyzer, e.g.

```
  ASTNode analyzeCreateTable(
```

abstractdog (changes requested):

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyser.java
ramitg254: I'll wait for the pending tests to check if patch is working after that I'll move the logic from semanticAnalyzer to this CreateTableAnalyzer class

ramitg254:
I don't know why SonarQubeCloud throwing new issues in files which I haven't modified and most of the code I have offloaded from SemanticAnalyzer to CreateTableAnalyzer so I don't think it shouldn't be any issue.
correct me if I'm wrong?

ramitg254:
changed to draft cause need to fix some test failures

ramitg254:
@abstractdog  can you please review the changes

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -449,9 +430,9 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
    */
   private boolean rootTasksResolved;
 
-  private TableMask tableMask;
+  protected TableMask tableMask;
 
-  CreateTableDesc tableDesc;
+  protected CreateTableDesc tableDesc;
```
deniskuzZ, line 435: why do we need CreateTableDesc here since we created a dedicated SA?

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -13000,24 +12979,15 @@ void gatherUserSuppliedFunctions(ASTNode ast) throws SemanticException {
...
-    if (ast.getToken().getType() == HiveParser.TOK_CREATETABLE) {
-      // if it is not CTAS, we don't need to go further and just return
-      if ((child = analyzeCreateTable(ast, qb, plannerCtx)) == null) {
-        return false;
-      }
-    } else {
-      // TODO: reiterate on this in HIVE-28750
-      queryState.setCommandType(HiveOperation.QUERY);
-    }
+    queryState.setCommandType(HiveOperation.QUERY);
```
deniskuzZ, line 12987: please update the numbering (step 2 is gone)

deniskuzZ:
Nice to see progress on the SA refactoring! I've added a few comments. Note, I think we should request a review from @kasakrisz.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
ramitg254: tableDesc is required in methods like `genFileSinkPlan ,materializeCTE, handleLineage` etc. in SA which I assumed are out of scope for CreateTableAnalyzer that's why CreateTableDesc is needed here

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
ramitg254: done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyzer.java
```diff
@@ -0,0 +1,829 @@
...
+  protected ASTNode handlePostCboRewriteContext(PreCboCtx cboCtx, ASTNode newAST)
+      throws SemanticException {
+    if (cboCtx.getType() == PreCboCtx.Type.CTAS) {
+      // CTAS
+      init(false);
+      setAST(newAST);
+      newAST = reAnalyzeCTASAfterCbo(newAST);
+    } else {
+      // All others
+      init(false);
```
kasakrisz: If the "All others" branch needs to be improved in the future it must be done in two places in the code.
How about calling `super.handlePostCboRewriteContext`?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyzer.java
```diff
@@ -0,0 +1,829 @@
...
+    }
+    tableMask = new TableMask(this, conf, ctx.isSkipTableMasking());
+    gatherUserSuppliedFunctions(child);
+    Phase1Ctx ctx_1 = initPhase1Ctx();
+    if (!doPhase1(child, getQB(), ctx_1, plannerCtx)) {
+      return false;
+    }
+    LOG.info("Completed phase 1 of Semantic Analysis");
+    getMetaData(getQB(), true);
+    LOG.info("Completed getting MetaData in Semantic Analysis");
```
kasakrisz: This part was copied from SemanticAnalyzer and now 2 copies exists. Could you please extract it to a method.

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
```diff
@@ -233,15 +233,39 @@ public boolean isPrepareQuery() {
...
   }
 
-  static final class RowFormatParams {
+  public static final class RowFormatParams {
     String fieldDelim = null;
     String fieldEscape = null;
     String collItemDelim = null;
     String mapKeyDelim = null;
     String lineDelim = null;
     String nullFormat = null;
```
kasakrisz: Can these be private?
Maybe final too if the parse logic is extracted to a factory method
```
public static final class RowFormatParams {
    
  public static RowFormatParams analyzeRowFormat(ASTNode child) {
    String fieldDelim = null;
    case HiveParser.TOK_TABLEROWFORMATFIELD:
            fieldDelim = unescapeSQLString(rowChild.getChild(0)
    ...
    return new RowFormatParams(fieldDelim ...)
  }

  private final String fieldDelim; 
  ...
  private RowFormatParams(String fieldDelim, ...) {
    this.fieldDelim = fieldDelim;
    ...
  }

  public String getFieldDelim() {
      return fieldDelim;
  }
}
```

I don't have a string opinion but since you have added getters to this class which i think is a good refactor we can go further and make it immutable.

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -479,15 +460,13 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
...
-      hive_metastoreConstants.TABLE_BUCKETING_VERSION
-  };
-
   private int subQueryExpressionAliasCounter = 0;
   private static final ObjectMapper JSON_OBJECT_MAPPER = new ObjectMapper();
-  static class Phase1Ctx {
+
+  protected static ObjectMapper getJsonObjectMapper(){
+    return JSON_OBJECT_MAPPER;
+  }
```
kasakrisz: This might be unnecessary because JSON_OBJECT_MAPPER acts as a constant just make it protected.

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
ramitg254: @kasakrisz making these private sounds good but in case of making these fields immutable via final may lead to extra handling of behaviour for RowFormatParams object as there are many occurences where `analyzeRowFormat()` method is called only when this condition holds: `case HiveParser.TOK_TABLEROWFORMAT` and if this condition is false then default values of these fields which is null is passed to descriptor like :
```
new CreateTableDesc(qualifiedTabName, isExt, isTemporary, cols, partColNames, bucketCols, sortCols,
            numBuckets, rowFormatParams.getFieldDelim(), rowFormatParams.getFieldEscape(),
            rowFormatParams.getCollItemDelim(), rowFormatParams.getMapKeyDelim(), rowFormatParams.getLineDelim(),
            comment, storageFormat.getInputFormat(), storageFormat.getOutputFormat(), location,
            storageFormat.getSerde(), storageFormat.getStorageHandler(), storageFormat.getSerdeProps(), tblProps,
            ifNotExists, skewedColNames, skewedValues, true, primaryKeys, foreignKeys, uniqueConstraints,
            notNullConstraints, defaultConstraints, checkConstraints);
```
so by use of factory method via `analyzeRowFormat` due to immutable fields will give arise to more use of if statements or ternary operators in situtation like these

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
ramitg254: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyzer.java
ramitg254: Done extracted to analyzeAndResolveChildTree method

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/create/CreateTableAnalyzer.java
ramitg254: Done

ramitg254: Done

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
kasakrisz: Maybe it would be better to pass the reference to the immutable `RowFormatParams` instance instead of passing 6 parameters and add null check to the getters of `CreateTableDesc`.

Anyway I understand that this would increase the size of this patch.

kasakrisz:
@ramitg254 
Please do not squash the commits before pushing. It helps reviewers to identify what was addressed and how.
Commiters always squash all the commits when merging the patch.

ramitg254:
> @ramitg254 Please do not squash the commits before pushing. It helps reviewers to identify what was addressed and how. Commiters always squash all the commits when merging the patch.

Okay, I'll keep this in mind next time and Thanks for the review

kasakrisz:
@deniskuzZ, @abstractdog 
Can I merge this patch ?

PR #5974, opened 2025-07-14
HIVE-29085: CREATE EXTERNAL TABLE fails for JDBC tables when both hiv…

soumyakanti3578:
…e.sql.table and hive.sql.query are set

Added a check in SemanticAnalyzer to not allow table creation for external JDBC tables when both `hive.sql.query` and `hive.sql.table` are set.

Setting both these properties together is ambiguous. Also we essentially get an NPE when both of them are set, as shown in https://issues.apache.org/jira/browse/HIVE-29085

No

```
mvn test -pl itests/qtest -Pitests -Dtest=TestNegativeLlapCliDriver -Dtest.output.overwrite=true -Dqfile="create_external_table_ambiguous_jdbc_table_properties.q"
mvn test -Dtest=org.apache.hadoop.hive.ql.parse.TestSemanticAnalyzer#testValidateNonAmbiguousJDBCTableProperties
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
@@ -13875,6 +13875,8 @@ public class SemanticAnalyzer extends BaseSemanticAnalyzer {
       boolean isTemporaryTable, boolean isTransactional, boolean isManaged, String[] qualifiedTabName, boolean isTableTypeChanged) throws SemanticException {
     Map<String, String> retValue = Optional.ofNullable(tblProp).orElseGet(HashMap::new);
 
+    validateJDBCTablePropertiesAreNonAmbiguous(isExt, storageFormat, retValue);
+
     String paraString = HiveConf.getVar(conf, ConfVars.NEW_TABLE_DEFAULT_PARA);
     if (paraString != null && !paraString.isEmpty()) {
       for (String keyValuePair : paraString.split(",")) {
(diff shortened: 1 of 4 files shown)


ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -13875,6 +13875,8 @@ private Map<String, String> validateAndAddDefaultProperties(
       boolean isTemporaryTable, boolean isTransactional, boolean isManaged, String[] qualifiedTabName, boolean isTableTypeChanged) throws SemanticException {
     Map<String, String> retValue = Optional.ofNullable(tblProp).orElseGet(HashMap::new);
 
+    validateJDBCTablePropertiesAreNonAmbiguous(isExt, storageFormat, retValue);
```
zabetak, line 13878: This verification check seems to be specific to the JDBC storage handler so it would fit better if the checks were delegated there (or to some other JDBC related class). If we start adding storage handler specific behavior here the class will start getting tightly coupled with storage that we probably don't want.

ql/src/test/org/apache/hadoop/hive/ql/parse/TestSemanticAnalyzer.java
```diff
@@ -493,4 +494,44 @@ private void checkTablesUsed(String query, Set<String> tables) throws Exception
 
     Assert.assertEquals(new TreeSet<>(tables), new TreeSet<>(result));
   }
+
+  @Test
+  public void testValidateJDBCTablePropertiesAreNonAmbiguous() {
+    StorageFormat storageFormat = mock(StorageFormat.class);
```
zabetak, line 500: Is it complicated to create a new instance of `StorageFormat/JdbcStorageHandler`? There are various drawback when we use mocks (debugging complexity, library dependencies, reduce test coverage, etc.) so we should strive to avoid it unless it is really complicated to create "real" objects.

ql/src/test/org/apache/hadoop/hive/ql/parse/TestSemanticAnalyzer.java
```diff
@@ -493,4 +494,44 @@ private void checkTablesUsed(String query, Set<String> tables) throws Exception
 
     Assert.assertEquals(new TreeSet<>(tables), new TreeSet<>(result));
   }
+
+  @Test
+  public void testValidateJDBCTablePropertiesAreNonAmbiguous() {
```
zabetak, line 499: The method seems to contain three tests and not just one so it would be better to split and name them accordingly so that when something fails we know exactly where the problem is.

zabetak: Additionally, I am not sure if this is the best class to put these new tests but let's revisit this once we agree on the prod changes.

PR #5975, opened 2025-07-15
HIVE-28804: The user does not have the permission for the table hdfs,…

zxl-333:
… but can delete the metadata

The user does not have the permission for the table hdfs, but can delete the metadata

When I create a table using the hdfs user and write data into it, and then use the hive user to delete this table, the engine side shows that the deletion is successful. However, the metastore log indicates that the deletion failed due to insufficient permissions when deleting the HDFS directory. Nevertheless, the metadata has been deleted. This situation may result in the data of this table becoming junk data.
2025-03-04 16:44:27,617 | WARN | org.apache.hadoop.hive.metastore.utils.FileUtils | Failed to move to trash: hdfs://myns/warehouse/tablespace/managed/hive/test_drop; Force to delete it. 2025-03-04 16:44:27,621 | ERROR | org.apache.hadoop.hive.metastore.utils.MetaStoreUtils | Got exception: org.apache.hadoop.security.AccessControlException Permission denied: user=hive, access=ALL, inode="/warehouse/tablespace/managed/hive/test_drop":hdfs:hadoop:drwxr-xr-x at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkSubAccess(FSPermissionChecker.java:455) at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:356) at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermissionWithContext(FSPermissionChecker.java:370) at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:240) at org.apache.hadoop.hdfs.server.namenode.FSDirectory.checkPermission(FSDirectory.java:1943) at org.apache.hadoop.hdfs.server.namenode.FSDirDeleteOp.delete(FSDirDeleteOp.java:105) at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.delete(FSNamesystem.java:3300) at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.delete(NameNodeRpcServer.java:1153) at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.delete(ClientNamenodeProtocolServerSideTranslatorPB.java:725) at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java) at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:614) at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:582) at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:566) at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1116) at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1060) at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:983) at java.security.AccessController.doPrivileged(Native Method) at javax.security.auth.Subject.doAs(Subject.java:422) at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1890) at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2997)org.apache.hadoop.security.AccessControlException: Permission denied: user=hive, access=ALL, inode="/warehouse/tablespace/managed/hive/test_drop":hdfs:hadoop:drwxr-xr-x at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkSubAccess(FSPermissionChecker.java:455) at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:356) at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermissionWithContext(FSPermissionChecker.java:370) at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:240) at org.apache.hadoop.hdfs.server.namenode.FSDirectory.checkPermission(FSDirectory.java:1943) at org.apache.hadoop.hdfs.server.namenode.FSDirDeleteOp.delete(FSDirDeleteOp.java:105) at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.delete(FSNamesystem.java:3300) at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.delete(NameNodeRpcServer.java:1153) at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.delete(ClientNamenodeProtocolServerSideTranslatorPB.java:725) at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java) at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:614) at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:582) at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:566) at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1116) at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1060) at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:983) at java.security.AccessController.doPrivileged(Native Method) at javax.security.auth.Subject.doAs(Subject.java:422) at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1890) at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2997)        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:1.8.0_352] at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[?:1.8.0_352] at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[?:1.8.0_352] at java.lang.reflect.Constructor.newInstance(Constructor.java:423) ~[?:1.8.0_352] at org.apache.hadoop.ipc.RemoteException.instantiateException(RemoteException.java:121) ~[hadoop-common-3.3.3.jar:?] at org.apache.hadoop.ipc.RemoteException.unwrapRemoteException(RemoteException.java:88) ~[hadoop-common-3.3.3.jar:?] at org.apache.hadoop.hdfs.DFSClient.delete(DFSClient.java:1664) ~[hadoop-hdfs-client-3.3.3.jar:?] at org.apache.hadoop.hdfs.DistributedFileSystem$19.doCall(DistributedFileSystem.java:992) ~[hadoop-hdfs-client-3.3.3.jar:?] at org.apache.hadoop.hdfs.DistributedFileSystem$19.doCall(DistributedFileSystem.java:989) ~[hadoop-hdfs-client-3.3.3.jar:?] at org.apache.hadoop.fs.FileSystemLinkResolver.resolve(FileSystemLinkResolver.java:81) ~[hadoop-common-3.3.3.jar:?] at org.apache.hadoop.hdfs.DistributedFileSystem.delete(DistributedFileSystem.java:999) ~[hadoop-hdfs-client-3.3.3.jar:?] at org.apache.hadoop.hive.metastore.utils.FileUtils.moveToTrash(FileUtils.java:97) ~[hive-exec-3.1.2.jar:3.1.2] at org.apache.hadoop.hive.metastore.HiveMetaStoreFsImpl.deleteDir(HiveMetaStoreFsImpl.java:41) [hive-exec-3.1.2.jar:3.1.2] at org.apache.hadoop.hive.metastore.Warehouse.deleteDir(Warehouse.java:363) [hive-exec-3.1.2.jar:3.1.2] at org.apache.hadoop.hive.metastore.Warehouse.deleteDir(Warehouse.java:351) [hive-exec-3.1.2.jar:3.1.2] at org.apache.hadoop.hive.metastore.HiveMetaStore$HMSHandler.deleteTableData(HiveMetaStore.java:2586) [hive-exec-3.1.2.jar:3.1.2] at org.apache.hadoop.hive.metastore.HiveMetaStore$HMSHandler.drop_table_core(HiveMetaStore.java:2559) [hive-exec-3.1.2.jar:3.1.2] at org.apache.hadoop.hive.metastore.HiveMetaStore$HMSHandler.drop_table_with_environment_context(HiveMetaStore.java:2708) [hive-exec-3.1.2.jar:3.1.2] at sun.reflect.GeneratedMethodAccessor238.invoke(Unknown Source) ~[?:?] at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_352] at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_352] at org.apache.hadoop.hive.metastore.RetryingHMSHandler.invokeInternal(RetryingHMSHandler.java:147) [hive-exec-3.1.2.jar:3.1.2] at org.apache.hadoop.hive.metastore.RetryingHMSHandler.invoke(RetryingHMSHandler.java:108) [hive-exec-3.1.2.jar:3.1.2] at com.sun.proxy.$Proxy27.drop_table_with_environment_context(Unknown Source) [?:?] at org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore$Processor$drop_table_with_environment_context.getResult(ThriftHiveMetastore.java:15068) [hive-exec-3.1.2.jar:3.1.2]

No

Use the existing unit tests.

### DIFF
--- a/ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java
@@ -132,7 +132,7 @@ public class TestHiveMetaStoreAuthorizer {
       dropDcReq.setIfNotExists(true);
       dropDcReq.setCheckReferences(true);
       hmsHandler.drop_dataconnector_req(dropDcReq);
-      hmsHandler.drop_table(dbName, tblName, true);
+      hmsHandler.drop_table("default", tblName, true);
       hmsHandler.drop_database(dbName, true, false);
       hmsHandler.drop_catalog(new DropCatalogRequest(catalogName));
       FileUtils.deleteDirectory(new File(TEST_DATA_DIR));
@@ -390,7 +390,7 @@ public class TestHiveMetaStoreAuthorizer {
     UserGroupInformation.setLoginUser(UserGroupInformation.createRemoteUser(authorizedUser));
     try {
       Table table = new TableBuilder()
-          .setTableName("test_drop")
+          .setTableName(tblName)
           .addCol("name", ColumnType.STRING_TYPE_NAME)
           .setOwner(authorizedUser)
           .build(conf);
@@ -399,7 +399,7 @@ public class TestHiveMetaStoreAuthorizer {
       Table alteredTable = new TableBuilder()
           .addCol("dep", ColumnType.STRING_TYPE_NAME)
           .build(conf);
-      hmsHandler.alter_table("default", "test_drop", alteredTable);
+      hmsHandler.alter_table("default", tblName, alteredTable);
     } catch (Exception e) {
       // No Exception for create table for authorized user
     }


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -3047,7 +3047,8 @@ private boolean drop_table_core(final RawStore ms, final String catName, final S
       tableDataShouldBeDeleted = checkTableDataShouldBeDeleted(tbl, deleteData);
       if (tableDataShouldBeDeleted && tbl.getSd().getLocation() != null) {
         tblPath = new Path(tbl.getSd().getLocation());
-        if (!wh.isWritable(tblPath.getParent())) {
+	// HIVE-28804 drop table user should have table path and parent path permission
```
deniskuzZ: Tables in Hive are owned by the hive user. To implement fine-grained access control, consider using Apache Ranger.

deniskuzZ: Also, I don’t understand why it’s necessary to check the parent directory, when the description mentions that the table directory is owned by the HDFS user, not hive

deniskuzZ: Another point, since this is a MANAGED table `/warehouse/tablespace/managed/hive/test_drop`, it should be owned by Hive, and external attempts to create or load data into it aren't supported.

deniskuzZ (dismissed):

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
zxl-333: When there is no ranger management permission, the above-mentioned exception will occur.The parent directory does not have the permission and cannot be deleted. This logic has always existed.Here, the hive user is merely a test user. It could also be other users such as spark.

deniskuzZ: @zxl-333 this is not an EXTERNAL table, but a MANAGED! Only hive user is supposed to have access, no-one else!!! Please check the docs for MANAGED table access patterns.

zxl-333: This table grants read access to other users through HDFS, but does not grant write permission. However, it is possible to delete the table. When deleting, the metastore throws the above exception message.

deniskuzZ: @zxl-333, is it possible to delete the table under non-hive user? Everything under the managed directory should be owned exclusively by the Hive user

- https://cwiki.apache.org/confluence/display/Hive/Managed+vs.+External+Tables
- https://medium.com/@sujathamudadla1213/explain-the-purpose-of-the-hive-external-tables-and-how-do-they-differ-from-managed-tables-bd9aa6f179bc

Use managed tables when Hive should manage the lifecycle of the table

zxl-333: > @zxl-333, is it possible to delete the table under non-hive user? Everything under the managed directory should be owned exclusively by the Hive user
> 
> * https://cwiki.apache.org/confluence/display/Hive/Managed+vs.+External+Tables
> * https://medium.com/@sujathamudadla1213/explain-the-purpose-of-the-hive-external-tables-and-how-do-they-differ-from-managed-tables-bd9aa6f179bc
> 
> Use managed tables when Hive should manage the lifecycle of the table

Yes, not a Hive user.For example, when the hdfs user created the "test_drop" table, the spark user was authorized by hdfs to read this table. However, the spark user deleted the "test_drop" table. On the engine side, a successful deletion message was received, but the metastore log showed the following abnormal information. The hdfs file did not actually get deleted.

deniskuzZ: @zxl-333, managed table concept only allows hive users to create/modify managed tables. Any other user shouldn't be granted write access to a managed location.

deniskuzZ: oh, ok. so you need to propagate the exception back to client? why add check for parent dir?
Btw, since it's a managed table, the 'test_drop' table should have been created by the Hive user, not the HDFS user.

why do you even create a MANAGED table? just use EXTERNAL

zxl-333: > oh, ok. so you need to propagate the exception back to client? why add check for parent dir? Btw, since it's a managed table, the 'test_drop' table should have been created by the Hive user, not the HDFS user.
> 
> why do you even create a MANAGED table? just use EXTERNAL

Not having the permission should be considered a failure, rather than deleting the metadata without deleting the HDFS file directory.

The "manage" table can not only be for Hive users, but also for other users on the cluster. If there is no permission to write in the parent directory, then even if the current directory has write permission, the deletion operation cannot be completed successfully. Below is what I tested through the HDFS command.
**parent direct:**
drwxr-xr-x   - hdfs   hadoop          0 2025-07-16 17:48 /tmp/test_drop
**child direct:**
drwxrwxrwx   - hdfs hadoop          0 2025-07-16 17:48 /tmp/test_drop/t1
use hive user delete direct
hdfs dfs -rm -r -f /tmp/test_drop/t1
rm: **Failed** to move to trash: hdfs://ns1/tmp/test_drop/t1: Permission denied: user=hive, access=WRITE, inode="/tmp/test_drop":hdfs:hadoop:drwxr-xr-x

deniskuzZ: The "manage" table  is a table governed by Hive.  All modifications must be performed through Hive. Any external changes, whether via CLI or other means are not supported !!!
Managed tables can only be created by the `hive` user.

zxl-333: > The "manage" table is a table governed by Hive. All modifications must be performed through Hive. Any external changes, whether via CLI or other means are not supported !!!
> Managed tables can only be created by the `hive` user.

The "manage" table has different users for each database. A cluster cannot have only one user.

deniskuzZ: An end-user cannot directly modify or create/delete the table directory for a managed table. Their interaction is with the Hive service, not the file system. They ask Hive to perform the action, and Hive does it on their behalf after checking permissions.
Managed table dir should have HIVE user as an owner, not HDFS

zxl-333: It is quite normal for other users to create the "manage" table as well. Otherwise, every time a table is created or deleted, the superuser "hive" would have to be used. A cluster can have hundreds of thousands of tables or even more. Every time a table is created or deleted, one has to contact the Hive superuser. The offline SQL scripts all contain operations for creating and deleting tables. On the YARN queue, it is mapped through the user name and the queue. This ensures that when running the offline SQL, there won't be a situation where the creation and deletion of tables are switched to be executed by the Hive superuser, which is not a reasonable approach. Especially, when creating tables, there will be expressions like "create table tbl_name as select". If other users are not allowed to create tables, then the entire cluster will have to rely solely on the "hive" superuser.

zxl-333:
@ayushtkn Please help review the code.Thank you.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
ayushtkn: @deniskuzZ In case `hive.server2.enable.doAs` is `true`, the managed table hdfs location would still be created by `hive` user? It would be the actual user right? only in case of impersonation turned off, it is hive user everywhere or am I missing something here?

deniskuzZ: yes
````
set metastore.strict.managed.tables=true;

set hive.server2.enable.doAs=false;
create table tab1 (c1 string, c2 string) stored as orc tblproperties ('transactional'='true', 'transactional_properties'='insert_only');

set hive.server2.enable.doAs=true;
create table tab2 (c1 string, c2 string) stored as orc tblproperties ('transactional'='true', 'transactional_properties'='insert_only');
````
describe output
````
describe formatted tab1;
| Owner:                        | systest

describe formatted tab2;
| Owner:                        | systest
````

hdfs dfs -ls hdfs://ns1/warehouse/tablespace/managed/hive/
````
drwxrwx---+  - hive   hive          0 2025-07-17 04:26 hdfs://ns1/warehouse/tablespace/managed/hive/tab1
drwxrwx---+  - hive   hive          0 2025-07-17 04:26 hdfs://ns1/warehouse/tablespace/managed/hive/tab2
````

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: @zxl-333 how did you create a table with `hdfs` user, using HiveQL? Woud be great if you could add repro steps to the PR description.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: @ayushtkn, when `hive.server2.enable.doAs=true` is set, the owner of the filesystem folder should be the user who executed the CREATE TABLE command.
I might be wrong, however, we should add at least some tests.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: > kinit -kt hdfs.keytab hdfs/*****
> ./beeline -u "jdbc:hive2://localhost:10001"

> 0: jdbc:hive2://localhost:10000> set hive.server2.enable.doAs;

````
+--------------------------------+
|              set               |
+--------------------------------+
| hive.server2.enable.doAs=true  |
+--------------------------------+
````

> create table tab1(i int) stored as orc TBLPROPERTIES ('transactional'='true', 'transactional_properties'='insert_only');

> describe formatted tab1;

````
| OwnerType:                    | USER                                               | NULL                                               |
| Owner:                        | hdfs                                               | NULL                                               |
| CreateTime:                   | Thu Jul 17 08:19:31 PDT 2025                       | NULL                                               |
| LastAccessTime:               | UNKNOWN                                            | NULL                                               |
| Retention:                    | 0                                                  | NULL                                               |
| Location:                     | hdfs://ns1/warehouse/tablespace/managed/hive/tab1  | NULL                                               |
| Table Type:                   | MANAGED_TABLE                                      | NULL
````

> !sh hdfs dfs -ls hdfs://ns1/warehouse/tablespace/managed/hive/

````
drwxrwx---+  - hdfs   hive          0 2025-07-17 08:19 hdfs://ns1/warehouse/tablespace/managed/hive/tab1
````

> kdestroy
> kinit -kt hive.keytab hive/*****
> ./beeline -u "jdbc:hive2://localhost:10001"

> drop table tab1

Table `tab1` is gone since `hive` is part of the owner group.

Comment (zxl-333) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: 1. The directory permissions for the /warehouse/tablespace/managed/hive/ are set to **777**
2. The **test_20250718** database was created using **user_1**, and the directory permissions and HDFS path permissions for the library are both set to **777**.
/warehouse/tablespace/managed/hive/test_20250718.db
3. Use the **user_1** user to create a test table, and set the HDFS directory permission of the table to **755**. create table test_1(k string);
4. Switch to the **user_2** account to delete the table **test_1**. At this point, you will notice that the metastore database of the **test_1** table has been deleted, but the HDFS directory has not been deleted. By checking the metastore log, you will find the abnormal stack information I provided.
Under normal circumstances, user_2 should not be able to delete test_1.

Comment (zhangbutao) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: @zxl-333 Your test case appears to be able to reproduce this issue. However, in the actual production environment, user clusters usually have Ranger or other authentication plugins enabled to ensure metadata permissions and storage permissions. 

So I’d like to ask, have you encountered this issue in your real production environment with Ranger or similar authentication plugins enabled?  Or, in actual production environments, how do you ensure Hive table permission control?

Thanks.

Comment (zxl-333) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: > @zxl-333 Your test case appears to be able to reproduce this issue. However, in the actual production environment, user clusters usually have Ranger or other authentication plugins enabled to ensure metadata permissions and storage permissions.
> 
> So I’d like to ask, have you encountered this issue in your real production environment with Ranger or similar authentication plugins enabled? Or, in actual production environments, how do you ensure Hive table permission control?
> 
> Thanks.

One of our clusters does not use ranger. Therefore, the users raised this issue as feedback. Using ranger is completely fine. This cluster also does not employ Hive's own authorization model; instead, it solely relies on HDFS permission control.

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: thanks @zxl-333 for the details! Now I get it and fix makes sense. Could you please add a test for the described scenario? Thanks!

Comment (zxl-333) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: > thanks @zxl-333 for the details! Now I get it and fix makes sense. Could you please add a test for the described scenario? Thanks!

I have add test

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java [hunk 2]
```diff
@@ -3047,11 +3047,17 @@ private boolean drop_table_core(final RawStore ms, final String catName, final S
...
       if (tableDataShouldBeDeleted && tbl.getSd().getLocation() != null) {
         tblPath = new Path(tbl.getSd().getLocation());
+	// HIVE-28804 drop table user should have table path and parent path permission
         if (!wh.isWritable(tblPath.getParent())) {
           String target = indexName == null ? "Table" : "Index table";
           throw new MetaException(target + " metadata not deleted since " +
               tblPath.getParent() + " is not writable by " +
               SecurityUtils.getUser());
+        } else if (!wh.isWritable(tblPath)) {
+          String target = indexName == null ? "Table" : "Index table";
```
deniskuzZ: @zxl-333, could you please move  the `target` outside the if block and use String formatter
````
throw new MetaException("%s metadata not deleted since %s is not writable by %s"
    .formatted(target, tblPath, SecurityUtils.getUser()));
````

Comment (zxl-333) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: ok

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]: @zxl-333, please do not change the `target` initialization, it doesn't need string.format. 
Please use formatter for the exception message construction. I've added the code snippet as an example in a comment above

Comment (deniskuzZ) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java]:
````
String target = indexName == null ? "Table" : "Index table";
if (!wh.isWritable(tblPath.getParent())) {
  throw new MetaException("%s metadata not deleted since %s is not writable by %s"
      .formatted(target, tblPath, SecurityUtils.getUser()));
} else if (!wh.isWritable(tblPath)) {
  throw new MetaException("%s metadata not deleted since %s is not writable by %s"
      .formatted(target, tblPath, SecurityUtils.getUser()));
}
````

File: ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java [hunk 3]
```diff
@@ -120,6 +121,8 @@ public void setUp() throws Exception {
     hmsHandler = new HMSHandler("test", conf);
     hmsHandler.init();
     rawStore = new ObjectStore();
+    wh = new Warehouse(conf);
+    handler.wh = wh;
```
deniskuzZ: @zxl-333 that doesn't even compile! use this
````
Subject: [PATCH] test
---
Index: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
IDEA additional info:
<+>UTF-8
diff --git a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java	(revision 67cb35e75297a9e6887a905ea92fb8c6a858b014)
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java	(date 1753175396594)
@@ -311,6 +311,11 @@
 
   @Override
   public void init() throws MetaException {
+    init(new Warehouse(conf));
+  }
+
+  @VisibleForTesting
+  public void init(Warehouse wh) throws MetaException {
     Metrics.initialize(conf);
     initListeners = MetaStoreServerUtils.getMetaStoreListeners(
         MetaStoreInitListener.class, conf, MetastoreConf.getVar(conf, ConfVars.INIT_HOOKS));
@@ -322,7 +327,7 @@
     String alterHandlerName = MetastoreConf.getVar(conf, ConfVars.ALTER_HANDLER);
     alterHandler = ReflectionUtils.newInstance(JavaUtils.getClass(
         alterHandlerName, AlterHandler.class), conf);
-    wh = new Warehouse(conf);
+    this.wh = wh;
 
     synchronized (HMSHandler.class) {
       if (currentUrl == null || !currentUrl.equals(MetaStoreInit.getConnectionURL(conf))) {
@@ -3047,11 +3052,13 @@
       tableDataShouldBeDeleted = checkTableDataShouldBeDeleted(tbl, deleteData);
       if (tableDataShouldBeDeleted && tbl.getSd().getLocation() != null) {
         tblPath = new Path(tbl.getSd().getLocation());
-        if (!wh.isWritable(tblPath.getParent())) {
-          String target = indexName == null ? "Table" : "Index table";
-          throw new MetaException(target + " metadata not deleted since " +
-              tblPath.getParent() + " is not writable by " +
-              SecurityUtils.getUser());
+        String target = indexName == null ? "Table" : "Index table";
+        if (!wh.isWritable(tblPath.getParent())) {
+          throw new MetaException("%s metadata not deleted since %s is not writable by %s"
+              .formatted(target, tblPath, SecurityUtils.getUser()));
+        } else if (!wh.isWritable(tblPath)) {
+          throw new MetaException("%s metadata not deleted since %s is not writable by %s"
+              .formatted(target, tblPath, SecurityUtils.getUser()));
         }
       }
 
Index: ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java
IDEA additional info:
<+>UTF-8
diff --git a/ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java b/ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java
--- a/ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java	(revision 67cb35e75297a9e6887a905ea92fb8c6a858b014)
+++ b/ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java	(date 1753178967863)
@@ -23,6 +23,7 @@
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.common.FileUtils;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.ColumnType;
@@ -70,6 +71,7 @@
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.eq;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 /*
 Test whether HiveAuthorizer for MetaStore operation is trigger and HiveMetaStoreAuthzInfo is created by HiveMetaStoreAuthorizer
@@ -94,6 +96,7 @@
   private RawStore rawStore;
   private Configuration conf;
   private HMSHandler hmsHandler;
+  private Warehouse wh;
 
   static HiveAuthorizer mockHiveAuthorizer;
   static final List<String> allowedUsers = Arrays.asList("sam", "rob");
@@ -116,9 +119,10 @@
     conf.set("hadoop.proxyuser.hive.users", "*");
 
     MetaStoreTestUtils.setConfForStandloneMode(conf);
+    wh = Mockito.spy(new Warehouse(conf));
 
     hmsHandler = new HMSHandler("test", conf);
-    hmsHandler.init();
+    hmsHandler.init(wh);
     rawStore = new ObjectStore();
     rawStore.setConf(hmsHandler.getConf());
     // Create the 'hive' catalog with new warehouse directory
@@ -386,7 +390,7 @@
     UserGroupInformation.setLoginUser(UserGroupInformation.createRemoteUser(authorizedUser));
     try {
       Table table = new TableBuilder()
-          .setTableName(tblName)
+          .setTableName("test_drop")
           .addCol("name", ColumnType.STRING_TYPE_NAME)
           .setOwner(authorizedUser)
           .build(conf);
@@ -395,7 +399,7 @@
       Table alteredTable = new TableBuilder()
           .addCol("dep", ColumnType.STRING_TYPE_NAME)
           .build(conf);
-      hmsHandler.alter_table("default", tblName, alteredTable);
+      hmsHandler.alter_table("default", "test_drop", alteredTable);
     } catch (Exception e) {
       // No Exception for create table for authorized user
     }
@@ -829,4 +833,29 @@
       }
     }
   }
+
+  @Test
+  public void testDropTableNoTablePathWritePermissionShouldFail() throws Exception {
+    UserGroupInformation.setLoginUser(
+        UserGroupInformation.createRemoteUser(authorizedUser));
+
+    Table table = new TableBuilder()
+        .setTableName(tblName)
+        .addCol("name", ColumnType.STRING_TYPE_NAME)
+        .setOwner(authorizedUser)
+        .build(conf);
+    hmsHandler.create_table(table);
+
+    Path tablePath = new Path(table.getSd().getLocation());
+    when(wh.isWritable(Mockito.eq(tablePath.getParent()))).thenReturn(true);
+    when(wh.isWritable(Mockito.eq(tablePath))).thenReturn(false);
+
+    try {
+      hmsHandler.drop_table("default", tblName, true);
+    } catch (MetaException e) {
+      String expected = "%s metadata not deleted since %s is not writable by %s"
+          .formatted("Table", tablePath.toString(), authorizedUser);
+      assertEquals(expected, e.getMessage());
+    }
+  }
 }
````

Comment (zxl-333) [reply ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java]: thank you

File: ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java [hunk 4]
```diff
@@ -386,7 +390,7 @@ public void testJ_AlterTable_AuthorizedUser() throws Exception {
     UserGroupInformation.setLoginUser(UserGroupInformation.createRemoteUser(authorizedUser));
     try {
       Table table = new TableBuilder()
-          .setTableName(tblName)
+          .setTableName("test_drop")
```
deniskuzZ: @zxl-333 please revert changes in that test method. that's my bad
submitted the fix. 

I'll merge the PR with the green build

PR #5976, opened 2025-07-15
Fix MSCK Repair adding invalid partitions

vikramahuja1001:


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,41 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
...
-      return Long.toString(Long.parseLong(partitionValue));
-    } else if (type.equalsIgnoreCase("float")){
-      return Float.toString(Float.parseFloat(partitionValue));
-    } else if (type.equalsIgnoreCase("double")){
-      return Double.toString(Double.parseDouble(partitionValue));
-    } else if (type.startsWith("decimal")){
-      // Decimal datatypes are stored like decimal(10,10)
-      return new BigDecimal(partitionValue).stripTrailingZeros().toPlainString();
+    try {
+      if (type.equalsIgnoreCase("tinyint")
```
deniskuzZ: could this be replaced with switch?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,41 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
     return result;
   }
 
-  public static String getNormalisedPartitionValue(String partitionValue, String type) {
-
-    if (!NumberUtils.isParsable(partitionValue)) {
+  public static String getNormalisedPartitionValue(String partitionValue, String type,
+      Configuration conf) {
+    if (!NumberUtils.isParsable(partitionValue) && !type.equalsIgnoreCase("string")
```
deniskuzZ: don't we have constants or enum for type like string?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
vikramahuja1001: Done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
vikramahuja1001: Changed to used an enum

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,77 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
...
+    public static PartitionDataType fromTypeString(String type) {
+      if (type == null || type.isEmpty()){
+        return UNKNOWN;
+      }
+      String base = type.toLowerCase();
+      if (base.startsWith("decimal")){
+        return DECIMAL;
+      }
+      return switch (base) {
+        case "tinyint" -> TINYINT;
```
deniskuzZ: can you simply use `base.toUpperCase`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,77 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
     return result;
   }
 
-  public static String getNormalisedPartitionValue(String partitionValue, String type) {
-
-    if (!NumberUtils.isParsable(partitionValue)) {
+  public enum PartitionDataType {
```
deniskuzZ: why not use `org.apache.hadoop.hive.metastore.ColumnType.getTypeName()`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,42 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
...
-      return Long.toString(Long.parseLong(partitionValue));
-    } else if (type.equalsIgnoreCase("float")){
-      return Float.toString(Float.parseFloat(partitionValue));
-    } else if (type.equalsIgnoreCase("double")){
-      return Double.toString(Double.parseDouble(partitionValue));
-    } else if (type.startsWith("decimal")){
-      // Decimal datatypes are stored like decimal(10,10)
-      return new BigDecimal(partitionValue).stripTrailingZeros().toPlainString();
+    try {
+      switch (type.toLowerCase()) {
```
deniskuzZ: maybe `ColumnType.getTypeName(type)`? this would handle the `type.startsWith("decimal")`

ql/src/test/results/clientpositive/llap/msck_repair_9.q.out
```diff
@@ -90,7 +90,7 @@ PREHOOK: Output: default@tbl_y
 POSTHOOK: query: MSCK REPAIR TABLE tbl_y
 POSTHOOK: type: MSCK
 POSTHOOK: Output: default@tbl_y
-Partitions not in metastore:	tbl_y:month=12/day=2	tbl_y:month=12/day=3	tbl_y:month=12/day=__HIVE_DEFAULT_PARTITION__	tbl_y:month=ANOTHER_PARTITION/day=3
+Partitions not in metastore:	tbl_y:month=12/day=2	tbl_y:month=12/day=3	tbl_y:month=ANOTHER_PARTITION/day=3
```
deniskuzZ, line 93: why `tbl_y:month=12/day=__HIVE_DEFAULT_PARTITION__` is gone?

ql/src/test/results/clientpositive/llap/msck_repair_9.q.out
```diff
@@ -100,8 +100,40 @@ POSTHOOK: type: SHOWPARTITIONS
 POSTHOOK: Input: default@tbl_y
 month=12/day=2
 month=12/day=3
-month=12/day=__HIVE_DEFAULT_PARTITION__
```
deniskuzZ, line 103: same here

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java
```diff
@@ -444,7 +444,12 @@ void findUnknownPartitions(Table table, Set<Path> missingPartDirs,
...
-          partPath, partColNames, partitionColToTypeMap);
+          partPath, partColNames, partitionColToTypeMap, conf);
+      if (partitionName == null) {
+        // Skip this partition if there is some issue in the partition validation
+        LOG.warn("Skipping partition : " + partPath.getName());
+        continue;
+      }
       LOG.debug("PartitionName: " + partitionName);
 
       if (partitionName != null) {
```
deniskuzZ: why check for null, if we skip the iteration when `partitionName == null`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,42 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
     return result;
   }
 
-  public static String getNormalisedPartitionValue(String partitionValue, String type) {
-
-    if (!NumberUtils.isParsable(partitionValue)) {
+  public static String getNormalisedPartitionValue(String partitionValue, String type,
+      Configuration conf) {
+    if ((!NumberUtils.isParsable(partitionValue) && !Objects.equals(type, ColumnType.STRING_TYPE_NAME)
```
deniskuzZ, line 1643: check for string type seems redundant

deniskuzZ: i don't understand this change:

if the partitionValue is not a number but type is String, we'll attempt to parse the number in partitionValue - doesn't make any sense

DEFAULTPARTITIONNAME only applies to String type

deniskuzZ: I see it like
````
public static String getNormalisedPartitionValue(String partitionValue, String type, Configuration conf) {
  // 1. Handle simple exit cases first.
  if (type == null) {
    return partitionValue;
  }
  if (Objects.equals(partitionValue, MetastoreConf.getVar(conf,
        MetastoreConf.ConfVars.DEFAULTPARTITIONNAME))) {
    // This is the special partition name for NULL values. It should never be parsed.
    return partitionValue;
  }
  // 2. If the type is not numeric, no normalization is needed.
  String colType = ColumnType.getTypeName(type);
  if (!ColumnType.NumericTypes.contains(colType)) {
    return partitionValue;
  }
  // 3. At this point, we have a numeric type that needs normalization.
  LOG.debug("Normalizing partition value '{}' for type '{}'.", partitionValue, type);
  try {
    switch (colType) {
      case ColumnType.TINYINT_TYPE_NAME:
      case ColumnType.SMALLINT_TYPE_NAME:
      case ColumnType.INT_TYPE_NAME:
        return Integer.toString(Integer.parseInt(partitionValue));
      case ColumnType.BIGINT_TYPE_NAME:
        return Long.toString(Long.parseLong(partitionValue));
      case ColumnType.FLOAT_TYPE_NAME:
        return Float.toString(Float.parseFloat(partitionValue));
      case ColumnType.DOUBLE_TYPE_NAME:
        return Double.toString(Double.parseDouble(partitionValue));
      case ColumnType.DECIMAL_TYPE_NAME:
        return new BigDecimal(partitionValue).stripTrailingZeros().toPlainString();
    }
  } catch (NumberFormatException e) {
    // 4. Handle cases where the value cannot be parsed as the expected number type.
    String validationMode = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.MSCK_PATH_VALIDATION);
    if ("throw".equals(validationMode)) {
      LOG.error("Invalid partition value: Cannot parse '{}' as type '{}'. Failing MSCK. "
          + "Set hive.msck.path.validation=skip to ignore invalid partitions.", partitionValue, type);
      throw e;
    } else if ("skip".equals(validationMode)) {
      LOG.warn("Skipping invalid partition value '{}' for type '{}' due to parsing error.", partitionValue, type);
      // Signals the caller to skip this partition.
      return null;
    }
  }
  return partitionValue;
}
````

**Comment by deniskuzZ** [comment]
@vikramahuja1001, please check the comments

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java (hunk 5 above)
Comment (vikramahuja1001) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java]: Yes, it should work. Let me try it out. Thanks for the suggestion.

File: ql/src/test/results/clientpositive/llap/msck_repair_9.q.out (hunk 6 above)
Comment (vikramahuja1001) [reply ql/src/test/results/clientpositive/llap/msck_repair_9.q.out:93]: Okay, so this test was buggy to begin with. Since default partition name is already changed to ANOTHER_PARTITION and since these are int partition columns, the only partition other than int type MSCK repair should add is the default partition value which in this case is ANOTHER_PARTITION. 
So partition value __HIVE_DEFAULT_PARTITION__ shouldn't be added in this case.

This bug is actually induced in HIVE-28654.

File: ql/src/test/results/clientpositive/llap/msck_repair_9.q.out (hunk 7 above)
Comment (vikramahuja1001) [reply ql/src/test/results/clientpositive/llap/msck_repair_9.q.out:103]: Same as above

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java (hunk 8 above)
Comment (vikramahuja1001) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java]: Good idea. I will remove the later null check. It's redundant in our case now

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java (hunk 9 above)
Comment (vikramahuja1001) [reply standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java]: Sure, i will update the same

File: ql/src/test/results/clientpositive/llap/msck_repair_9.q.out (hunk 6 above)
Comment (deniskuzZ) [reply ql/src/test/results/clientpositive/llap/msck_repair_9.q.out:93]: > INSERT INTO tbl_x (id, name, month) values(3, 'ccc', 12);

PARTITION(month=12,day=__HIVE_DEFAULT_PARTITION__)

> SET hive.exec.default.partition.name=ANOTHER_PARTITION;
> INSERT INTO tbl_x (id, name, day) values(4, 'ddd', 3);

PARTITION(month=ANOTHER_PARTITION,day=3)

What happens to `PARTITION(month=12,day=__HIVE_DEFAULT_PARTITION__)` data after the MSCK REPAIR? 
Should it be moved under `PARTITION(month=12,day=ANOTHER_PARTITION)` ?

Comment (vikramahuja1001) [reply ql/src/test/results/clientpositive/llap/msck_repair_9.q.out:93]: I think for now the user should be moving the data files in case they need it. 
I think the only reason they will change the default partition property is if they do not need the previous data or want to try something with the new data. In both the cases the old data is not needed. 
We technically shouldn't be allowing any string partition value other than default partition values for non string partition columns. It can lead to other issues down the line during querying. 

Let me know what you think about it

Comment (deniskuzZ) [reply ql/src/test/results/clientpositive/llap/msck_repair_9.q.out:93]: isn't it a regression, since it was working before?

Comment (vikramahuja1001) [reply ql/src/test/results/clientpositive/llap/msck_repair_9.q.out:93]: No regression. It was never working before HIVE-28654. Post HIVE-28654 we are able to add any string partition value to a non string partition column as of now which doesn't make sense. This is a bug induced by HIVE-28654.

Comment (deniskuzZ) [reply ql/src/test/results/clientpositive/llap/msck_repair_9.q.out:93]: got it, thanks @vikramahuja1001! 
is https://github.com/apache/hive/pull/5976/files#diff-6465db75ef87970b9c4ccc96661af7ff4aa691c1624eb87688657da70f27d3a6R450 called for PARTITION(month=12,day=__HIVE_DEFAULT_PARTITION__)?

@okumin, @Aggarwal-Raghav, could you please review as well since you were involved in the original fix

Comment (Aggarwal-Raghav) [reply ql/src/test/results/clientpositive/llap/msck_repair_9.q.out:93]: @vikramahuja1001 , i would recomment you to check the disucssion in this thread https://github.com/apache/hive/pull/5568#discussion_r1871258402 . Based on this discussion only I went ahead allowing any string partition name.

**My understanding is this**: If user is adding a string partition name in non-string column then its user's responsibility and he should be consistent with the default partition name for that table. As DEFAULT PARTITION is configurable at session level, we can't stop user from changing it / playing around.

Imposing checks done in your current PR will be a breaking change b/w 4.0.1 and next hive release version and forcing users to set this on cluster level `hive.msck.path.validation=skip` from what I can infer from modified msck_repair_9.q

File: standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreChecker.java [hunk 10]
```diff
@@ -444,34 +444,37 @@ void findUnknownPartitions(Table table, Set<Path> missingPartDirs,
     for (Path partPath : missingPartDirs) {
       FileSystem fs = partPath.getFileSystem(conf);
       String partitionName = getPartitionName(fs.makeQualified(tablePath),
-          partPath, partColNames, partitionColToTypeMap);
+          partPath, partColNames, partitionColToTypeMap, conf);
+      if (partitionName == null) {
```
Aggarwal-Raghav, line 448: In which case partitionName is null?

vikramahuja1001: When ` hive.msck.path.validation=skip`

ql/src/test/results/clientpositive/llap/msck_repair_9.q.out
vikramahuja1001: Hey, I have gone through the discussion in the thread [PR5568](https://github.com/apache/hive/pull/5568#discussion_r1871258402) . 

Have gone through both the scenarios that you have mentioned in that discussion.  I believe under no circumstance the user should be allowed to add arbitrary strings for non string partition values by using MSCK repair it can lead to unsafe comparisons down the line. Only DEFAULT PARTITION VALUE should be considered at a time. If not, this leads to various issues like:
1. **Type Integrity Violation:**
The core expectation in Hive is that partition values must adhere to the column's defined type. Allowing arbitrary strings in a BIGINT/INT or any other non string column undermines this contract.
2. Downstream Failures:
Operations downstream (like queries, filters, and metadata tools) may fail or behave unpredictably when encountering such invalid partition values opening a pandora box of further issues. For instance: [HIVE-29100](https://issues.apache.org/jira/browse/HIVE-29100). We will need to specifically handle such cases now.
3. Silent Corruption Risk:
Since MSCK is a metadata recovery tool, silently accepting bad values gives a false sense of correctness, potentially leading to hard-to-debug issues in the future.
4. Inconsistent Behavior Across Interfaces:
Hive does not allow such partitions to be added via standard DDL operations (e.g., ALTER TABLE ADD PARTITION) due to strict type checking. This inconsistency between interfaces (DDL vs MSCK) can confuse users and lead to misconfiguration.

We should allow partitions under hive.exec.default.partition.name property and thus a basic validation like is must and the property` hive.msck.path.validation` can decide whether to throw an error for such cases or just log and skip those arbitrary partitions.

The config ` hive.msck.path.validation=skip` can be set as session level. By default I have kept this value as `throw`. No need to set anything at cluster level.

deniskuzZ: It makes sense to me, however, I might be missing some edge-cases. 

@vikramahuja1001, could you please clarify again what happens in the scenario below 
>INSERT INTO tbl_x (id, name, month) values(3, 'ccc', 12);

PARTITION(month=12,day=HIVE_DEFAULT_PARTITION)

>SET hive.exec.default.partition.name=ANOTHER_PARTITION;
>INSERT INTO tbl_x (id, name, day) values(4, 'ddd', 3);

PARTITION(month=ANOTHER_PARTITION,day=3)

>MSCK REPAIR TABLE tbl_x;

Do we log a warning for `PARTITION(month=12,day=__HIVE_DEFAULT_PARTITION__)` or fail ?

vikramahuja1001: @deniskuzZ, in the above scenario that you described:
`INSERT INTO tbl_x (id, name, month) values(3, 'ccc', 12);` -> PARTITION(month=12,day=HIVE_DEFAULT_PARTITION)
`SET hive.exec.default.partition.name=ANOTHER_PARTITION;`
`INSERT INTO tbl_x (id, name, day) values(4, 'ddd', 3);` -> PARTITION(month=ANOTHER_PARTITION,day=3)
`MSCK REPAIR TABLE tbl_x;`

hive.exec.default.partition.name is an exactly one value. So at one point of time only one one value is valid, rest all the other values are invalid. So when MSCK repair is ran , `hive.exec.default.partition.name` has been already set to `ANOTHER_PARTITION` which makes the  HIVE_DEFAULT_PARTITION not a valid partition anymore.

Since by default `hive.msck.path.validation` is set to `throw`,  msck repair will throw an exception as it encounters a invalid partition (month=12,day=HIVE_DEFAULT_PARTITION) and will fail the whole operation without adding any new partition to the HMS with the exception: `Invalid partition value: Cannot parse HIVE_DEFAULT_PARTITION as type int. Failing MSCK. Set hive.msck.path.validation=skip to ignore invalid partitions` 

But when the user sets `hive.msck.path.validation=skip` at  the session level, msck repair operation will succeed but it will skip all the invalid partitions by adding a warning log (month=12,day=HIVE_DEFAULT_PARTITION in this case) and will only add the valid partitions in that case((month=ANOTHER_PARTITION,day=3).

deniskuzZ: LGTM, @Aggarwal-Raghav, do you see anything major that we should be concerned about?

Aggarwal-Raghav: @deniskuzZ @vikramahuja1001 , I'll review this PR by Monday. Currently Out of station.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,50 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
     return result;
   }
 
-  public static String getNormalisedPartitionValue(String partitionValue, String type) {
-
-    if (!NumberUtils.isParsable(partitionValue)) {
```
Aggarwal-Raghav, line 1641: nit: remove unused import statement as well.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,50 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
...
+    if (type == null) {
       return partitionValue;
     }
-
-    LOG.debug("Converting '" + partitionValue + "' to type: '" + type + "'.");
-
-    if (type.equalsIgnoreCase("tinyint")
-    || type.equalsIgnoreCase("smallint")
-    || type.equalsIgnoreCase("int")){
-      return Integer.toString(Integer.parseInt(partitionValue));
```
Aggarwal-Raghav, line 1650: nit: can we use JDK 21 switch expression here? It looks more clean!

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
```diff
@@ -1636,27 +1638,50 @@ public static String getPartitionName(Path tablePath, Path partitionPath, Set<St
...
-  public static String getNormalisedPartitionValue(String partitionValue, String type) {
-
-    if (!NumberUtils.isParsable(partitionValue)) {
+  public static String getNormalisedPartitionValue(String partitionValue, String type, Configuration conf) {
+    // 1. Handle simple exit cases first.
+    if (type == null) {
       return partitionValue;
     }
-
-    LOG.debug("Converting '" + partitionValue + "' to type: '" + type + "'.");
```
Aggarwal-Raghav, line 1645: Can we use string formatter here as sona type is also flagging it?

Aggarwal-Raghav:
@vikramahuja1001, Can you add a new q file for hive.msck.path.validation = ignore. as there is none at the moment. It will help in more coverage.

Aggarwal-Raghav:
**NOTE**: Just want to re-iterate and highlight the breaking change here. Anyway as +1 is there on the PR which means it's ok from dev perspective.

Have give some nit comments for improving code readability which are purely optional. 

**Before this PR:**
scenarion2_3_without_patch

**After this PR:**
Screenshot 2025-07-27 at 3 43 39 PM

Aggarwal-Raghav:
+0 (non-binding)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
vikramahuja1001: Sure

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/utils/MetaStoreServerUtils.java
vikramahuja1001: Sure

vikramahuja1001:
Thanks for the review @Aggarwal-Raghav 

@deniskuzZ , kindly check again.

deniskuzZ:
hi @Aggarwal-Raghav, were your comments resolved?

Aggarwal-Raghav:
> hi @Aggarwal-Raghav, were your comments resolved?

the changes I requested were mostly syntactic sugar i.e string formatting and removing unused imports and using JDK21 style switch expression. They were not functionality specific.

vikramahuja1001:
@deniskuzZ , everything has been handled. Let me know if something else needs to be done for this.

PR #5977, opened 2025-07-15
HIVE-29093 : Fix alter table add partition if DEFAULT_PARTITION_NAME propery is a substring of partitition value to be added

vikramahuja1001:
…g of partition value.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
@@ -56,7 +56,7 @@ public final class PartitionUtils {
   }
 
   /**
-   * Certain partition values are are used by hive. e.g. the default partition in dynamic partitioning and the
+   * Certain partition values are used by hive. e.g. the default partition in dynamic partitioning and the
    * intermediate partition values used in the archiving process. Naturally, prohibit the user from creating partitions
    * with these reserved values. The check that this function is more restrictive than the actual limitation, but it's
    * simpler. Should be okay since the reserved names are fairly long and uncommon.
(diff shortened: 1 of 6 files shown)


ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
```diff
@@ -63,20 +63,28 @@ private PartitionUtils() {
...
     reservedPartitionValues.add(HiveConf.getVar(conf, ConfVars.DEFAULT_ZOOKEEPER_PARTITION_NAME));
     // Partition value can't end in this suffix
-    reservedPartitionValues.add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_ORIGINAL));
-    reservedPartitionValues.add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_ARCHIVED));
-    reservedPartitionValues.add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_EXTRACTED));
+    reservedSuffixPartitionValues.add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_ORIGINAL));
+    reservedSuffixPartitionValues.add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_ARCHIVED));
+    reservedSuffixPartitionValues.add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_EXTRACTED));
 
     for (Entry<String, String> e : partitionSpec.entrySet()) {
```
deniskuzZ: could we please refactor with 
````
partitionSpec.foreach((key, value) -> {
  if (value != null && reservedPartitionValue.containsKey(value)) {
  ...
  }
}
````
similar for `reservedPartitionSuffixes`

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java [hunk 2]
```diff
@@ -63,20 +63,28 @@ private PartitionUtils() {
    */
   public static void validatePartitions(HiveConf conf, Map<String, String> partitionSpec) throws SemanticException {
     Set<String> reservedPartitionValues = new HashSet<>();
+    Set<String> reservedSuffixPartitionValues = new HashSet<>();
```
deniskuzZ: maybe `reservedPartitionSuffixes` ? also why not do anonymous initialization?
````
Set<String> reservedPartitionSuffixes = 
  new HashSet<String>() {{
    add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_ORIGINAL));
    add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_ARCHIVED));
    add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_EXTRACTED));
  }};
````

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java [hunk 3]
```diff
@@ -62,24 +62,33 @@ private PartitionUtils() {
...
-        }
+    Set<String> reservedPartitionSuffixes =
+            new HashSet<String>() {{
+              add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_ORIGINAL));
+              add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_ARCHIVED));
+              add(HiveConf.getVar(conf, ConfVars.METASTORE_INT_EXTRACTED));
+            }};
+
+    partitionSpec.forEach((key, value) -> {
+      if (value != null && reservedPartitionValues.contains(value)) {
```
deniskuzZ: sorry, my bad
````
partitionSpec.forEach((key, value) -> {
  if (value == null) {
    return;
  }
  reservedPartitionValues.stream().filter(value::contains).findAny()
    .ifPresent(s -> {
      throw new RuntimeException(ErrorMsg.RESERVED_PART_VAL.getMsg(
              "(User value: " + value + " Reserved string: " + s + ")"));
  });
  reservedPartitionSuffixes.stream().filter(value::endsWith)
    .ifPresent(s -> {
      throw new RuntimeException(ErrorMsg.RESERVED_PART_VAL.getMsg(
              "(User value: " + value + " Partition value cannot end with Reserved substring: " + s + ")"));
  });
});
````

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java]: @vikramahuja1001, please check the above comment. Current logic should be broken, but looks like test don't cover all cases. Please add the missing one.
Exception should be thrown if partValue contains one of the reserved keys, not other way around.

Comment (vikramahuja1001) [reply ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java:81]: Thanks for the clarification @deniskuzZ , i will make the necessary changes and push. Don't you think it should be 
`  reservedPartitionValues.stream().filter(value::equals)` and not `  reservedPartitionValues.stream().filter(value::contains)`?

For instance let's say my partition column is of type string and i have set Default partition name as 2012. Then even partitions like 2012-03-03 will not get added because of this. Let me know what you think?

Currently i am pushing the code with `  reservedPartitionValues.stream().filter(value::equals)`

Comment (vikramahuja1001) [reply ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java]: Thanks for the clarification @deniskuzZ , i have made the changes and added the test cases as well

Comment (deniskuzZ) [reply ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java]: thanks @vikramahuja1001! could you also please use 4 spaces instead of 8

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java [hunk 4]
```diff
@@ -61,25 +61,37 @@ private PartitionUtils() {
    * with these reserved values. The check that this function is more restrictive than the actual limitation, but it's
    * simpler. Should be okay since the reserved names are fairly long and uncommon.
    */
-  public static void validatePartitions(HiveConf conf, Map<String, String> partitionSpec) throws SemanticException {
-    Set<String> reservedPartitionValues = new HashSet<>();
+  public static void validatePartitions(HiveConf conf, Map<String, String> partitionSpec) {
```
deniskuzZ, line 64: please remove 2 `are` from the java doc in "Certain partition values are are used by hive"

vikramahuja1001: Sure

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
vikramahuja1001: Sure, done

deniskuzZ (approved):
+1, pending tests

vikramahuja1001:
Thanks for the review @deniskuzZ

PR #5978, opened 2025-07-15
HIVE-29029: When stack UDTF is used with UNION ALL throwing ClassCastException in CBO mode.

mdayakar:
HIVE-29029: When stack UDTF is used with UNION ALL throwing ClassCastException in CBO mode.

Ignore HiveUnionSimpleSelectsToInlineTableRule when stack UDTF is used with UNION ALL.

To fix the ClassCastException when stack UDTF is used with UNION ALL.

No

_mvn -Dtest=TestMiniLlapLocalCliDriver -Dqfile=udtf_with_unionall.q -pl itests/qtest -Pitests_

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
@@ -218,7 +218,16 @@ public class HiveUnionSimpleSelectsToInlineTableRule extends RelOptRule {
     if (input.getInputs().size() == 0) {
       return true;
     }
-    if (!(((RexCall)((HiveTableFunctionScan) input).getCall()).getOperands().get(0) instanceof RexCall)) {
+    RexNode call = ((HiveTableFunctionScan) input).getCall();
+    if (!(call instanceof RexCall)) {
+      return false;
+    }
+    // there should be operands present, if not return false
+    if (((RexCall) call).getOperands().size() == 0) {
+      return false;
+    }
+    // the operands should be of type RexCall, if not return false
+    if (!(((RexCall) call).getOperands().get(0) instanceof RexCall)) {
       return false;
     }
     return isDummyTableScan(input.getInput(0));
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
@@ -218,6 +218,9 @@ public class HiveUnionSimpleSelectsToInlineTableRule extends RelOptRule {
     if (input.getInputs().size() == 0) {
       return true;
     }
+    if (!(((RexCall)((HiveTableFunctionScan) input).getCall()).getOperands().get(0) instanceof RexCall)) {
+      return false;
+    }
     return isDummyTableScan(input.getInput(0));
   }
 
--- a/ql/src/test/queries/clientpositive/udtf_with_unionall.q
+++ b/ql/src/test/queries/clientpositive/udtf_with_unionall.q
@@ -1,0 +1,55 @@
+SELECT STACK(3,'A',10,date '2015-01-01','z','B',20,date '2016-01-01','y','C',30,date '2017-08-09','x') AS (col0,col1,col2,col3)
+  UNION ALL
+ SELECT STACK(3,'A',10,date '2015-01-01','n','B',20,date '2016-01-01','m','C',30,date '2017-08-09','l') AS (col0,col1,col2,col3);
+
+EXPLAIN SELECT stack(3,'A',10,date '2015-01-01','z','B',20,date '2016-01-01','y','C',30,date '2017-08-09','x') AS (col0,col1,col2,col3)
+  UNION ALL
+ SELECT STACK(3,'A',10,date '2015-01-01','n','B',20,date '2016-01-01','m','C',30,date '2017-08-09','l') AS (col0,col1,col2,col3);
+
+SELECT * FROM (VALUES(1, '1'), (2, 'orange'), (5, 'yellow')) AS Colors1
+  UNION ALL
+ SELECT * FROM (VALUES(10, 'green'), (11, 'blue'), (12, 'indigo'), (20, 'violet')) AS Colors2
+  UNION ALL
+ SELECT STACK(2,10,'X',20,'Y');
+
+EXPLAIN SELECT * FROM (VALUES(1, '1'), (2, 'orange'), (5, 'yellow')) AS Colors1
+  UNION ALL
+ SELECT * FROM (VALUES(10, 'green'), (11, 'blue'), (12, 'indigo'), (20, 'violet')) AS Colors2
+  UNION ALL
+ SELECT STACK(2,10,'X',20,'Y');
+
+SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2016-02-02')))
+  UNION ALL
+ SELECT STACK(2,'X',10,date '2017-01-01','Y',20,date '2018-01-01');
+
+EXPLAIN SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2016-02-02')))
+  UNION ALL
+ SELECT STACK(2,'X',30,date '2017-01-01','Y',40,date '2018-01-01');
+
+SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2015-02-02')))
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')));
+
+EXPLAIN SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2015-02-02')))
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')));
+
+SELECT EXPLODE(map('A',10,'B',20,'C',30))
+  UNION ALL
+ SELECT EXPLODE(map('X',70,'Y',80,'Z',90));
+
+EXPLAIN SELECT EXPLODE(map('A',10,'B',20,'C',30))
+  UNION ALL
+ SELECT EXPLODE(map('X',70,'Y',80,'Z',90));
+
+SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2015-02-02')))
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')))
+  UNION ALL
+ SELECT STACK(2,'X',50,date '2017-01-01','Y',60,date '2017-01-01');
+
+EXPLAIN SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2015-02-02')))
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')))
+  UNION ALL
+ SELECT STACK(2,'X',50,date '2017-01-01','Y',60,date '2017-01-01');
--- a/ql/src/test/queries/clientpositive/udtf_with_unionall.q
+++ b/ql/src/test/queries/clientpositive/udtf_with_unionall.q
@@ -2,7 +2,7 @@ SELECT STACK(3,'A',10,date '2015-01-01','z','B',20,date '2016-01-01','y','C',30,
   UNION ALL
  SELECT STACK(3,'A',10,date '2015-01-01','n','B',20,date '2016-01-01','m','C',30,date '2017-08-09','l') AS (col0,col1,col2,col3);
 
-EXPLAIN SELECT stack(3,'A',10,date '2015-01-01','z','B',20,date '2016-01-01','y','C',30,date '2017-08-09','x') AS (col0,col1,col2,col3)
+EXPLAIN CBO SELECT stack(3,'A',10,date '2015-01-01','z','B',20,date '2016-01-01','y','C',30,date '2017-08-09','x') AS (col0,col1,col2,col3)
   UNION ALL
  SELECT STACK(3,'A',10,date '2015-01-01','n','B',20,date '2016-01-01','m','C',30,date '2017-08-09','l') AS (col0,col1,col2,col3);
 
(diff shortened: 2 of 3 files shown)


ql/src/test/queries/clientpositive/udtf_with_unionall.q
```diff
@@ -0,0 +1,55 @@
+SELECT STACK(3,'A',10,date '2015-01-01','z','B',20,date '2016-01-01','y','C',30,date '2017-08-09','x') AS (col0,col1,col2,col3)
+  UNION ALL
+ SELECT STACK(3,'A',10,date '2015-01-01','n','B',20,date '2016-01-01','m','C',30,date '2017-08-09','l') AS (col0,col1,col2,col3);
```
zabetak, line 3: The `ClassCastException` occurs during compilation so testing result correctness is slightly out of scope for the current PR. It is probably fine to add one end-to-end test but not sure if we need all of them. Does each test in this file contribute to additional test coverage?

ql/src/test/queries/clientpositive/udtf_with_unionall.q
```diff
@@ -0,0 +1,55 @@
...
+
+EXPLAIN SELECT stack(3,'A',10,date '2015-01-01','z','B',20,date '2016-01-01','y','C',30,date '2017-08-09','x') AS (col0,col1,col2,col3)
+  UNION ALL
+ SELECT STACK(3,'A',10,date '2015-01-01','n','B',20,date '2016-01-01','m','C',30,date '2017-08-09','l') AS (col0,col1,col2,col3);
+
+SELECT * FROM (VALUES(1, '1'), (2, 'orange'), (5, 'yellow')) AS Colors1
+  UNION ALL
+ SELECT * FROM (VALUES(10, 'green'), (11, 'blue'), (12, 'indigo'), (20, 'violet')) AS Colors2
```
zabetak, line 11: Do we need two `VALUES` branches to trigger the problem or one suffices? Do we need three rows on each `VALUES`?

Whatever is not needed to repro the problem should be dropped. Keeping tests minimal speeds-up reviews, improves readability, and minimizes flakiness in the longterm.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
```diff
@@ -0,0 +1,55 @@
...
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')));
+
+EXPLAIN SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2015-02-02')))
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')));
+
+SELECT EXPLODE(map('A',10,'B',20,'C',30))
+  UNION ALL
+ SELECT EXPLODE(map('X',70,'Y',80,'Z',90));
```
zabetak: I like the fact that we are testing various UDTFs with UNION ALL but not sure under what criteria we picked the combinations to test. Is it exhaustive? Is it on most commonly used? Is it something else?

The fix is localized in a single rule so if we want to add test cases for this purpose it feels more appropriate to do it via unit tests (see Test*Rule.java classes) if that's feasible.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
```diff
@@ -0,0 +1,55 @@
...
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')))
+  UNION ALL
+ SELECT STACK(2,'X',50,date '2017-01-01','Y',60,date '2017-01-01');
+
+EXPLAIN SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2015-02-02')))
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')))
+  UNION ALL
+ SELECT STACK(2,'X',50,date '2017-01-01','Y',60,date '2017-01-01');
```
zabetak: nit: No new line at the end of file.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
```diff
@@ -0,0 +1,55 @@
...
+  UNION ALL
+ SELECT STACK(3,'A',10,date '2015-01-01','n','B',20,date '2016-01-01','m','C',30,date '2017-08-09','l') AS (col0,col1,col2,col3);
+
+SELECT * FROM (VALUES(1, '1'), (2, 'orange'), (5, 'yellow')) AS Colors1
+  UNION ALL
+ SELECT * FROM (VALUES(10, 'green'), (11, 'blue'), (12, 'indigo'), (20, 'violet')) AS Colors2
+  UNION ALL
+ SELECT STACK(2,10,'X',20,'Y');
+
+EXPLAIN SELECT * FROM (VALUES(1, '1'), (2, 'orange'), (5, 'yellow')) AS Colors1
```
zabetak: Since the problem is in CBO we don't care about the final (physical plan). `EXPLAIN CBO` is sufficient, easier to read, less subject to changes due to modifications in the compiler.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
```diff
@@ -218,6 +218,9 @@ private boolean isInlineTableOperand(RelNode input) {
     if (input.getInputs().size() == 0) {
       return true;
     }
+    if (!(((RexCall)((HiveTableFunctionScan) input).getCall()).getOperands().get(0) instanceof RexCall)) {
+      return false;
+    }
```
zabetak: This check is not very readable and it also makes various assumptions about the type and operands of `TableFunctionScan#getCall` that may lead to other `ClassCastException` or other problems. Maybe we should consider adding a helper function and putting all the necessary checks there. For instance:
```java
private static boolean isInlineArrayCall(RexNode expr) {
    if (!(expr instanceof RexCall)) {
      return false;
    }
    RexCall call = (RexCall) expr;
    SqlOperator op = call.getOperator();
    if (!"inline".equalsIgnoreCase(op.getName())) {
      return false;
    }
    if (call.getOperands().size() != 1) {
      return false;
    }
    RexNode operand = call.getOperands().get(0);
    if (!(operand instanceof RexCall)) {
      return false;
    }
    call = (RexCall) operand;
    return "array".equalsIgnoreCase(call.getOperator().getName());
  }
```
It is a bit verbose but not sure if there is a better way to put it :/

Other than that I am not sure if we should make the check here or rather in Line 145-146 as you initially proposed.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
mdayakar: Added new line at the end of the file.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
mdayakar: I just want to make sure the results are also as expected, if not required I can remove.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
mdayakar: Here I want to test some scenario like UNION ALL is applied for VALUES, for which rule will come into picture and converts two values into one INLINE udtf and STACK, for which rule should not be applied.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
mdayakar: Used `EXPLAIN CBO` now.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
mdayakar: Added required checks now with separate if conditions. We cannot use "inline" text here as for VALUES also it creates dummy table.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
mdayakar: Removed other UDTFs which are not required to be tested as a part of this PR.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
```diff
@@ -218,6 +218,18 @@ private boolean isInlineTableOperand(RelNode input) {
...
+    // there should be operands present, if not return false
+    if (((RexCall) call).getOperands().size() == 0) {
+      return false;
+    }
+    // the operands should be of type RexCall, if not return false
+    if (!(((RexCall) call).getOperands().get(0) instanceof RexCall)) {
+      return false;
+    }
```
zabetak, line 232: The check implemented here could be translated in natural language as:

"The rule must accept any expression that starts with a function call (RexCall), the call has some operands (not zero), and the first operand is also a function call (RexCall)"

The check is a bit general and although it avoids the `ClassCastException` it doesn't seem sufficient to guarantee that the rule will behave correctly for arbitrary combinations of a `GenericUDTF` with other functions.

Having said that solving the `ClassCastException` is a progress compared to the previous state so I think we can leave with this solution for the time being.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
zabetak: Validating that the results are correct is a good idea so let's keep the tests.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
zabetak: With your explanation I understand but the intention of the test is not visible somewhere. As I mentioned earlier for validating how certain rules work unit tests (`Test*Rule.java`) are usually faster and more explicit. I guess we can keep these tests now but consider adding some comments to clarify the expected outcome.

ql/src/test/results/clientpositive/llap/udtf_with_unionall.q.out
```diff
@@ -0,0 +1,168 @@
...
+      HiveTableScan(table=[[_dummy_database, _dummy_table]], table:alias=[_dummy_table])
+  HiveProject(col0=[$0], col1=[$1], col2=[$2], col3=[$3])
+    HiveTableFunctionScan(invocation=[stack(3, _UTF-16LE'A':VARCHAR(2147483647) CHARACTER SET "UTF-16LE", 10, 2015-01-01:DATE, _UTF-16LE'n':VARCHAR(2147483647) CHARACTER SET "UTF-16LE", _UTF-16LE'B':VARCHAR(2147483647) CHARACTER SET "UTF-16LE", 20, 2016-01-01:DATE, _UTF-16LE'm':VARCHAR(2147483647) CHARACTER SET "UTF-16LE", _UTF-16LE'C':VARCHAR(2147483647) CHARACTER SET "UTF-16LE", 30, 2017-08-09:DATE, _UTF-16LE'l':VARCHAR(2147483647) CHARACTER SET "UTF-16LE")], rowType=[RecordType(VARCHAR(2147483647) col0, INTEGER col1, DATE col2, VARCHAR(2147483647) col3)])
```
zabetak, line 37: The use of `STRING` literals lead into plans with lots of boilerplate info `_UTF-16LE'l':VARCHAR(2147483647) CHARACTER SET "UTF-16LE"`. If you were using `INTEGER` literals the plans would be more readable. Anyways that's a minor point so you can ignore for now but maybe worth considering for future PRs.

ql/src/test/results/clientpositive/llap/udtf_with_unionall.q.out
```diff
@@ -0,0 +1,168 @@
...
+  UNION ALL
+ SELECT * FROM (VALUES(10, 'green'), (11, 'blue'), (12, 'indigo'), (20, 'violet')) AS Colors2
+  UNION ALL
+ SELECT STACK(2,10,'X',20,'Y')
+POSTHOOK: type: QUERY
+POSTHOOK: Input: _dummy_database@_dummy_table
+#### A masked pattern was here ####
+10	X
+20	Y
+1	1
```
zabetak, line 58: The choice of `1 1` seems inconsistent while all the rest are combinations of integers and colours. Minor but it does appear strange.

ql/src/test/queries/clientpositive/udtf_with_unionall.q
```diff
@@ -0,0 +1,39 @@
...
+  UNION ALL
+ SELECT STACK(2,10,'X',20,'Y');
+
+SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2015-02-02')))
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')));
+
+EXPLAIN CBO SELECT INLINE(array(struct('A',10,date '2015-01-01'),struct('B',20,date '2015-02-02')))
+  UNION ALL
+ SELECT INLINE(array(struct('C',30,date '2016-01-01'),struct('D',40,date '2016-02-02')));
```
zabetak, line 27: This test is a bit redundant since the query below is more complex and validates the same point that two inline branches  are merged by the rule. It doesn't seem to add to test coverage.

zabetak (approved):
LGTM. Spotted some small stuff but we can leave with those.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
zabetak: Can you elaborate a bit on how the VALUES is related to this method? I don't understand the connection.

mdayakar: For VALUES also finally it converts to inline UDTF when it comes to this rule and they will be combined. I thought for VALUES it will be converted differently so I said we can not use "inline" string comparison but we don't know exactly the other scenarios so I feel current checks might be enough to prevent ClassCastException and no impact to the existing behaviour..

zabetak: They are enough to prevent the ClassCastException observed here but not sufficient to cover problems with other combinations UDTF as pointed out in another comment. Anyways, this is out-of-scope of this PR so marking this discussion as resolved.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveUnionSimpleSelectsToInlineTableRule.java
zabetak: We will keep this for now and revisit the rule if/when it fails again.

zabetak:
Thanks for the PR @mdayakar !

ql/src/test/results/clientpositive/llap/udtf_with_unionall.q.out
mdayakar: Changed accordingly.

PR #5638, opened 2025-02-09
HIVE-28748: Docker: Redirect daemon's logs to STDOUT

okumin:
### What changes were proposed in this pull request?

Put the default log4j properties in the official Docker image.

### Why are the changes needed?

File-based logging is inconvenient, and it doesn't follow the common practice of containers.

### Does this PR introduce _any_ user-facing change?

Yes.

### Is the change a dependency upgrade?

No.

### How was this patch tested?

I tested the output in my local machine.

PR #5980, opened 2025-07-16
HIVE-29094: Unifying the metastore-site.xml File for HMS Docker and HMS tar

zhangbutao:

Unify the metastore-site.xml files for HMS Docker and HMS tar to ensure consistent default configurations when starting Standalone HMS.

No

Local build Standalone HMS tar and Standalone HMS docker images


standalone-metastore/packaging/src/docker/Dockerfile
```diff
@@ -67,15 +67,13 @@ ENV HADOOP_HOME=/opt/hadoop \
 ENV PATH=$HIVE_HOME/bin:$HADOOP_HOME/bin:$PATH
 
 COPY entrypoint.sh /
-COPY conf $HIVE_HOME/conf
```
deniskuzZ, line 70: @zhangbutao, please check the logger output with docker conf log4j2 properties and the metastore one. 
If you tail the container logs, you'll notice that most of the logging is gone.

zhangbutao: Yes, i know what you mean.

The `docker logs -f` command is quite convenient for simple testing and viewing logs. However, in a real production environment, users typically persist the metastore logs locally for better log management. 

After this PR, log4j defaults to storing HMS logs in /tmp/hive/metastore.log. If users want to view the logs on the local host machine, they can mount the HMS logs to a local directory when starting the HMS Docker container. For example, using the command:

`docker run -itd --name='hivehms' -v /data/hmslog:/tmp/hive apache/hive:standalone-metastore-4.1.0`

Then, users can check the HMS logs in the local directory at `/data/hmslog/metastore.log`.

zhangbutao: Additionally, if users want to modify the `metastore-log4j2.properties` file before starting the HMS Docker, they can also use the local mount method. That is, they can use a local `metastore-log4j2.properties` file when starting the HMS Docker. This allows users to flexibly modify the `metastore-log4j2.properties`, such as changing the HMS log level or outputting logs to `stdout`.

```
ls /data/hmsdocker/conf

metastore-log4j2.properties
metastore-site.xml
```

`docker run -itd --name='hive420hms_stdout' -v /data/hmsdocker/conf:/opt/hive/conf apache/hive:standalone-metastore-4.1.0`

deniskuzZ: ok, then we should probably make it consistent with Hive docker. Same thing there, output is redirected to System.out

- Could we include details about logging in the README?
- Maybe consider changing the log dir to `/var/log/hive/ `

zhangbutao: Hive docker log was changed by https://github.com/apache/hive/pull/5638.
@okumin Any thought if we change it back?


> maybe change the log dir to /var/hive/log/metastore.log ?

Yes, The log dir can be changed by when `export HADOOP_CLIENT_OPTS`
https://github.com/apache/hive/blob/5512ffda602bef605ad144f9040676149cc2c960/standalone-metastore/packaging/src/docker/entrypoint.sh#L48

```
-export HADOOP_CLIENT_OPTS="$HADOOP_CLIENT_OPTS -Xmx1G $SERVICE_OPTS"
+export HADOOP_CLIENT_OPTS="$HADOOP_CLIENT_OPTS -Xmx1G -Dmetastore.log.dir=/var/hive/log -Dmetastore.log.file=metastore.log $SERVICE_OPTS"
```

deniskuzZ: okumin might be on vacation, let's wait a bit

deniskuzZ: hi @okumin, do you agree with the above suggestion?

deniskuzZ (dismissed):
LGTM +1

standalone-metastore/packaging/src/docker/Dockerfile
okumin: Sorry for the late response(I've been on medical leave and I missed notifications during a crisis). I suppose Kubernetes is almost the only production-grade container orchestration tool. In the Kubernetes world, a main container typically outputs logs to stdout/stderr, a Kubernetes administrator sets up log collectors such as a Fluentd daemonset, the log collector transfers all containers' logs to a central system such as Elasticsearch, and some chores such as log rotations are handled by Kublet or something configured by a k8s administrator.

zhangbutao: IMO, whether logs are output to a local directory or to stdout/stderr does not affect log collection in Kubernetes. We have many services (such as StarRocks/Doris) running on Kubernetes, and these services output logs to local directories, which are eventually collected by components like Elasticsearch and persisted.

okumin: For instance, Google Kubernetes Engine can send stdout/stderr to Google Cloud Logging without additional configurations. I suppose it is [a typical logging practice with k8s](https://kubernetes.io/docs/concepts/cluster-administration/logging/) and can be the default choice of HMS.

PR #5981, opened 2025-07-16
HIVE-29092: NoClassDefFoundError: io/netty/util/NetUtil when using hive jdbc standalone jar

zhangbutao:

Fix the missing jars when using hive-jdbc-standalone.jar. See details from HIVE-29092 to check how to test this case.

No

Local java api test using hive-jdbc-standalone jar

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -467,6 +467,10 @@
                       <pattern>org.codehaus.stax2</pattern>
                       <shadedPattern>org.apache.hive.org.codehaus.stax2</shadedPattern>
                     </relocation>
+                    <relocation>
+                      <pattern>io.netty</pattern>
+                      <shadedPattern>org.apache.hive.io.netty</shadedPattern>
+                    </relocation>
                   </relocations>
                 </configuration>
               </execution>
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -266,6 +266,13 @@
                       </includes>
                     </filter>
                     <filter>
+                      <artifact>io.netty:netty-common</artifact>
+                      <includes>
+                        <include>io/netty/util/*</include>
+                        <include>io/netty/util/internal/**</include>
+                      </includes>
+                    </filter>
+                    <filter>
                       <artifact>*:*</artifact>
                       <excludes>
                         <!--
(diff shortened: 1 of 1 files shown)


jdbc/pom.xml
```diff
@@ -328,7 +335,14 @@
                       <exclude>dnsjava:*</exclude>
                       <exclude>io.airlift:*</exclude>
                       <exclude>io.dropwizard.metrics:*</exclude>
-                      <exclude>io.netty:*</exclude>
+                      <exclude>io.netty:netty-handler</exclude>
```
zhangbutao, line 338: @ayushtkn @abstractdog Could you help to check if this change would cause regression for HIVE-24138? 

I just included `io.netty:netty-common` to fix the `NoClassDefFoundError: io/netty/util/NetUtil when using hive jdbc standalone jar`.

Thanks.

jdbc/pom.xml
```diff
@@ -265,6 +265,13 @@
                         <include>org/apache/hadoop/yarn/api/records/ApplicationId.class</include>
                       </includes>
                     </filter>
+                    <filter>
+                      <artifact>io.netty:netty-common</artifact>
+                      <includes>
+                        <include>io/netty/util/*</include>
```
dengzhhu653, line 271: The issue is solved after the PR, should we shade these classes as well?

zhangbutao: HIVE-24138 reverted the `io.netty` shade classes. 

If we were to shade it again, I'm not sure if there will be a regression problem.

zhangbutao: Shade the `io.netty:netty-common  `https://github.com/apache/hive/pull/5981/commits/455b45215355b244bcd32e71640dde2c10db5bd7

I checked HIVE-24138, and it seems that HIVE-24138 is related [netty-buffer](https://mvnrepository.com/artifact/io.netty/netty-buffer) `io/netty/buffer/ArrowBuf`.

We just want to include `netty-common ` in this PR, so i think this change maybe safe.
@ayushtkn Could help to double check?

zhangbutao: BTW, HIVE-24138 is caused by the shading issue in hive-exec(used by Llap external client), not in hive-jdbc.

dengzhhu653 (approved):

zhangbutao:
@dengzhhu653 @ayushtkn Thanks for the review!

PR #5983, opened 2025-07-17
HIVE-29088: Add a configuration option to skip corrupted files

Lucas61000:

This PR introduces a new Hive configuration parameter hive.exec.skip.protobuf.corruptfile to control whether Hive skips parsing of protocol buffer files that are corrupted (e.g., due to data truncation or incomplete writes) during processing, allowing tasks to continue with valid files instead of failing.

https://issues.apache.org/jira/browse/HIVE-29088

Users can now set hive.exec.skip.protobuf.corruptfile=true to skip corrupted protocol buffer files during parsing, whereas previously such errors would cause task failures.

Local

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -814,6 +814,10 @@ public class HiveConf extends Configuration {
         "Whether to use a native APIs for load queries to non-native table(like iceberg), if false uses a Tez job for" +
             " load queries"),
 
+    HIVE_SKIP_PROTOBUF_CORRUPTFILE("hive.exec.skip.protobuf.corruptfile", false,
+            "Whether Hive skips Protocol Buffer files corrupted, \n" +
+                    "during parsing (e.g., due to data truncation or file incompletion) to allow task continuation."),
+
     HIVE_IN_TEST("hive.in.test", false, "internal usage only, true in test mode", true),
     HIVE_IN_TEST_SSL("hive.in.ssl.test", false, "internal usage only, true in SSL test mode", true),
     // TODO: this needs to be removed; see TestReplicationScenarios* comments.
--- a/shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
+++ b/shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
@@ -23,6 +23,7 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.security.AccessControlException;
 import java.util.ArrayList;
@@ -132,6 +133,7 @@ public abstract class HadoopShimsSecure implements HadoopShims {
     protected RecordReader<K, V> curReader;
     protected boolean isShrinked;
     protected long shrinkedLength;
+    protected boolean skipCorruptfile;
 
     @Override
     public boolean next(K key, V value) throws IOException {
@@ -148,13 +150,32 @@ public abstract class HadoopShimsSecure implements HadoopShims {
 
     @Override
     public K createKey() {
-      K newKey = curReader.createKey();
-      return (K)(new CombineHiveKey(newKey));
+      skipCorruptfile = jc.getBoolean("hive.exec.skip.protobuf.corruptfile", false);
+      K newKey = null;
+      if (skipCorruptfile) {
+        try {
+          newKey = curReader.createKey();
+        } catch (NullPointerException e) {
+          LOG.info("=================Caught exception: " + e.getMessage() + ", skipping file======================");
+        }
+      } else {
+        newKey = curReader.createKey();
+      }
+      return (K) (new CombineHiveKey(newKey));
     }
 
     @Override
     public V createValue() {
-      return curReader.createValue();
+      skipCorruptfile = jc.getBoolean("hive.exec.skip.protobuf.corruptfile", false);
+      if (skipCorruptfile) {
+        try {
+          return curReader.createValue();
+        } catch (NullPointerException e) {
+          return null;
+        }
+      } else {
+        return curReader.createValue();
+      }
     }
 
     /**
@@ -234,7 +255,7 @@ public abstract class HadoopShimsSecure implements HadoopShims {
      * Get the record reader for the next chunk in this CombineFileSplit.
      */
     protected boolean initNextRecordReader(K key) throws IOException {
-
+      skipCorruptfile = jc.getBoolean("hive.exec.skip.protobuf.corruptfile", false);
       RecordReader preReader = curReader; //it is OK, curReader is closed, for we only need footer buffer info from preReader.
       if (curReader != null) {
         curReader.close();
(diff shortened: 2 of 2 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -814,6 +814,10 @@ public static enum ConfVars {
         "Whether to use a native APIs for load queries to non-native table(like iceberg), if false uses a Tez job for" +
             " load queries"),
 
+    HIVE_SKIP_PROTOBUF_CORRUPTFILE("hive.exec.skip.protobuf.corruptfile", false,
```
ngsg, line 817: - Could we start the config key with `hive.io` or something other than `hive.exec`? Since this does not affect the hive-exec module, I think the key should not start with `hive.exec`.

- We probably don't need the '\n' at the end of the line. Also, could we change it to use 4-space indentation, like this?
```suggestion
    HIVE_SKIP_PROTOBUF_CORRUPTFILE("hive.exec.skip.protobuf.corruptfile", false,
        "Whether Hive skips Protocol Buffer files corrupted, " +
        "during parsing (e.g., due to data truncation or file incompletion) to allow task continuation."),
```

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
```diff
@@ -132,6 +133,7 @@ public static class CombineFileRecordReader<K, V> implements RecordReader<K, V>
     protected RecordReader<K, V> curReader;
     protected boolean isShrinked;
     protected long shrinkedLength;
+    protected boolean skipCorruptfile;
```
ngsg, line 136: Could we change this to `private final skipCorruptFile`?

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
```diff
@@ -148,13 +150,32 @@ public boolean next(K key, V value) throws IOException {
 
     @Override
     public K createKey() {
-      K newKey = curReader.createKey();
-      return (K)(new CombineHiveKey(newKey));
+      skipCorruptfile = jc.getBoolean("hive.exec.skip.protobuf.corruptfile", false);
+      K newKey = null;
+      if (skipCorruptfile) {
```
ngsg, line 155: Could we check `skipCorruptFile` inside the catch clause like this?
https://github.com/apache/hive/blob/5512ffda602bef605ad144f9040676149cc2c960/ql/src/java/org/apache/hadoop/hive/ql/optimizer/physical/Vectorizer.java#L2077-L2083

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
```diff
@@ -148,13 +150,32 @@ public boolean next(K key, V value) throws IOException {
...
+        try {
+          newKey = curReader.createKey();
+        } catch (NullPointerException e) {
+          LOG.info("=================Caught exception: " + e.getMessage() + ", skipping file======================");
```
ngsg, line 159: I don't think we need the extra '=' characters in the log message. Could you simplify the message? Also, there's no corresponding log message in `createValue()`. I think these methods should either both emit logs or neither, for consistency.

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
```diff
@@ -249,26 +270,49 @@ protected boolean initNextRecordReader(K key) throws IOException {
...
+
+          // setup some helper config variables.
+          jc.set("map.input.file", split.getPath(idx).toString());
+          jc.setLong("map.input.start", split.getOffset(idx));
+          jc.setLong("map.input.length", split.getLength(idx));
+        } catch (Exception e) {
+          curReader = HiveIOExceptionHandlerUtil.handleRecordReaderCreationException(
+                  e, jc);
+        }
+        idx++;
```
ngsg, line 314: It seems that the skipCorruptFile-enabled path does not properly update `idx`. Could you check this?

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
```diff
@@ -249,26 +270,49 @@ protected boolean initNextRecordReader(K key) throws IOException {
...
+          if (key != null) {
+            K newKey = curReader.createKey();
+            ((CombineHiveKey) key).setKey(newKey);
+          }
+
+          // setup some helper config variables.
+          jc.set("map.input.file", split.getPath(idx).toString());
+          jc.setLong("map.input.start", split.getOffset(idx));
+          jc.setLong("map.input.length", split.getLength(idx));
+        } catch (InvocationTargetException ITe) {
```
ngsg, line 289: - Could we use lowercase for `ITe`?
- It looks like we skip all remaining splits once we encounter an `InvocationTargetException`. Shouldn't we proceed with them instead of just returning false?

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
```diff
@@ -249,26 +270,49 @@ protected boolean initNextRecordReader(K key) throws IOException {
...
+          return false;
+        } catch (Exception e) {
+          curReader = HiveIOExceptionHandlerUtil.handleRecordReaderCreationException(e, jc);
```
ngsg, line 292: It appears that Hive already provides a way to recover from exceptions thrown by RR via `HiveIOExceptionHandlerUtil#handleRecordReaderCreationException`. Could this be a solution to your issue?

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
```diff
@@ -148,13 +150,32 @@ public boolean next(K key, V value) throws IOException {
...
     @Override
     public K createKey() {
-      K newKey = curReader.createKey();
-      return (K)(new CombineHiveKey(newKey));
+      skipCorruptfile = jc.getBoolean("hive.exec.skip.protobuf.corruptfile", false);
+      K newKey = null;
+      if (skipCorruptfile) {
+        try {
+          newKey = curReader.createKey();
+        } catch (NullPointerException e) {
```
ngsg, line 158: I'm not sure about handling `NullPointerException` at this point. Can we say with certainty that a `NullPointerException` always results from a corrupted file? If not, I don't think we could swallow it.

ngsg (commented):
Thank you for submitting a PR. I've left comments on every place where I have questions.

Could you share the stacktrace and the reproduction steps? They would be very helpful for the community to review your patch. I would also appreciate it if you could provide more information about the following questions:
- Why is this patch related to protobuf?
- When do we need to skip a corrupted file? I see that the patch modifies a general KV reader, and currently I don’t have a strong reason to support this kind of recovery.
- Why is this recovery logic implemented only in `HadoopShimsSecure$CombineFileRecordReader`? I think there are other classes implementing combined RecordReaders that might also be controlled by the new configuration.

deniskuzZ:
@Lucas61000, could you please elaborate a bit more why this change is needed, where do you see an exception, and what protobuf files are you mentioning here?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
Lucas61000: Thanks for the review. Sure, I'll modify it.

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
Lucas61000: Thanks for the review. Sure, I'll modify it.

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
Lucas61000: Thanks for the review. Sure, I'll modify it.

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
Lucas61000: Thanks for the review. Sure, I'll modify it.

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
Lucas61000: Thanks for the review. The idx++ is not within the skipCorruptFile-enabled path.

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
Lucas61000: Thanks for the review. Sure, I'll modify it.

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
Lucas61000: Thanks for the review. The issue I’m facing is that some users don’t care about data integrity—they just want their jobs to run smoothly and hope to skip files corrupted during transmission.

shims/common/src/main/java/org/apache/hadoop/hive/shims/HadoopShimsSecure.java
Lucas61000: Thanks for the review. A NullPointerException always results from a corrupted file.

Lucas61000:
The stack trace involved have been included in the description. Thank you all.

Aggarwal-Raghav:
@Lucas61000 , please check [HIVE-28026](https://issues.apache.org/jira/browse/HIVE-28026) which I fixed for the similar stacktrace in tez https://github.com/apache/tez/pull/334. Can you confirm?

deniskuzZ:
> @Lucas61000 , please check [HIVE-28026](https://issues.apache.org/jira/browse/HIVE-28026) which I fixed for the similar stacktrace in tez [apache/tez#334](https://github.com/apache/tez/pull/334). Can you confirm?

@Aggarwal-Raghav, does it mean it's already fixed?

Aggarwal-Raghav:
> > @Lucas61000 , please check [HIVE-28026](https://issues.apache.org/jira/browse/HIVE-28026) which I fixed for the similar stacktrace in tez [apache/tez#334](https://github.com/apache/tez/pull/334). Can you confirm?
> 
> @Aggarwal-Raghav, does it mean it's already fixed?

Yes, I think so because the stacktrace is similar and error message is the same.

PR #5984, opened 2025-07-18
HIVE-29107: Drop avatica version management and use of shaded jars

zabetak:
Avatica is tightly connected with Calcite so it doesn't make sense to manage it separately Moreover, there is no need to use the shaded avatica artifact since the latter pulls in tons of things that Hive doesn't need.

No product facing changes for Hive users.

However, due to the removal of the shaded avatica jar some classes will not be part hive-exec so maven projects that depend on hive-exec may run into `ClassNotFoundException` issues if they depend implicitly on code bundled in avatica.

1. Existing tests
2. Build the hive binaries using mvn clean install -DskipTests -Pdist
3. Deployed to local cluster in pseudo-distributed mode and run a bunch of queries

### DIFF
--- a/itests/hive-blobstore/pom.xml
+++ b/itests/hive-blobstore/pom.xml
@@ -289,10 +289,6 @@
       <artifactId>jackson-databind</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.apache.calcite.avatica</groupId>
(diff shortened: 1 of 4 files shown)


deniskuzZ (approved):
+1

zabetak:
Many thanks for the review @deniskuzZ !

PR #5987, opened 2025-07-19
HIVE-28941: Do not create empty commit files in HiveIcebergOutputCommitter

Neer393:
Enhanced the HiveIcebergOutputCommitter to skip it from creating empty commit files

The changes are needed so as to reduce file clutter on the storage that gets created by HiveIcebergOutputCommitter

No, the underlying functionality remains same

Executed qtests before and after the changes and found out that empty commit files are no more being created.

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
@@ -786,14 +786,19 @@ public class HiveIcebergOutputCommitter extends OutputCommitter {
         .retry(3)
         .run(taskId -> {
           final String taskFileName = generateFileForCommitLocation(location, conf, jobContext.getJobID(), taskId);
-          final FilesForCommit files = readFileForCommit(taskFileName, io);
-          LOG.debug("Found Iceberg commitTask manifest file: {}\n{}", taskFileName, files);
+          try {
+            final FilesForCommit files;
+            files = readFileForCommit(taskFileName, io);
+            LOG.debug("Found Iceberg commitTask manifest file: {}\n{}", taskFileName, files);
 
-          dataFiles.addAll(files.dataFiles());
-          deleteFiles.addAll(files.deleteFiles());
-          replacedDataFiles.addAll(files.replacedDataFiles());
-          referencedDataFiles.addAll(files.referencedDataFiles());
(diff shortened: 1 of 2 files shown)


Neer393:
@deniskuzZ Can you review this for me and tell me if it's okay?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergOutputCommitter.java
```diff
@@ -786,14 +786,19 @@ private static FilesForCommit collectResults(int numTasks, ExecutorService execu
...
+            final FilesForCommit files;
+            files = readFileForCommit(taskFileName, io);
+            LOG.debug("Found Iceberg commitTask manifest file: {}\n{}", taskFileName, files);
+
+            dataFiles.addAll(files.dataFiles());
+            deleteFiles.addAll(files.deleteFiles());
+            replacedDataFiles.addAll(files.replacedDataFiles());
+            referencedDataFiles.addAll(files.referencedDataFiles());
+            mergedAndDeletedFiles.addAll(files.mergedAndDeletedFiles());
+          } catch (NotFoundException e) {
```
SourabhBadhya, line 799: This seems to me like a quick fix but ideally we should rethink the strategy of how we are writing the FilesForCommit object into the outputFile in such scenarios / possibly preserve some state indicating that this file is not written and not just throw NotFoundException.

Neer393: Yes even I felt that this is just a workaround where we are just ignoring the non existent commit files but this is not a good method.
The issue is currently we just find the number of tasks based on number of mapper/reducer jobs and assume that it is equal to the number of files and iterate over it. This is the reason why it still tries to find the not created commit files

Neer393: I have a possible solution

In the process where we are committing task files, we maintain a list of task file names that have been committed and at last when we are done, we create a new file in the same directory maybe called commitTasksInfo and store this list there. 

Then when we are collecting results, we read this commitTasksInfo file and determine which commit files were created and only try retrieving them. 

How does this sound? 
@SourabhBadhya @deniskuzZ @abstractdog

deniskuzZ: currently we have commit manifest file per task, are you proposing to squash them in 1 manifest file?

PR #5988, opened 2025-07-21
HIVE-28844: Clean up property hive.metastore.server.tcp.keepalive from Metastore

ramitg254:

Clean up of hive.metastore.server.tcp.keepalive property

since keepalive property is always true according to https://github.com/apache/thrift/blob/0.16.0/lib/java/src/org/apache/thrift/transport/TSocket.java#L78
and so cleanup of this property is required

 Yes, no custom value can be set for hive.metastore.server.tcp.keepalive

local build

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -283,7 +283,6 @@ public class HiveConf extends Configuration {
       HiveConf.ConfVars.METASTORE_CONNECT_URL_KEY,
       HiveConf.ConfVars.METASTORE_SERVER_MIN_THREADS,
       HiveConf.ConfVars.METASTORE_SERVER_MAX_THREADS,
-      HiveConf.ConfVars.METASTORE_TCP_KEEP_ALIVE,
       HiveConf.ConfVars.METASTORE_INT_ORIGINAL,
       HiveConf.ConfVars.METASTORE_INT_ARCHIVED,
       HiveConf.ConfVars.METASTORE_INT_EXTRACTED,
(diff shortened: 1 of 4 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -1096,13 +1095,6 @@ public static enum ConfVars {
     @Deprecated
     METASTORE_SERVER_MAX_THREADS("hive.metastore.server.max.threads", 1000,
         "Maximum number of worker threads in the Thrift server's pool."),
-    /**
-     * @deprecated Use MetastoreConf.TCP_KEEP_ALIVE
-     */
-    @Deprecated
-    METASTORE_TCP_KEEP_ALIVE("hive.metastore.server.tcp.keepalive", true,
-        "Whether to enable TCP keepalive for the metastore server. Keepalive will prevent accumulation of half-open connections."),
-
```
dengzhhu653, line 1105: cloud you also drop it from `conf/hive-default.xml.template` please?

ramitg254: @dengzhhu653  Sure,
but this property is not present in conf/hive-default.xml.template but similar one `hive.server.tcp.keepalive` was there related to
```
SERVER_TCP_KEEP_ALIVE("hive.server.tcp.keepalive", true,
        "Whether to enable TCP keepalive for the Hive Server. Keepalive will prevent accumulation of half-open connections."),
```
which is different from this property so i didn't drop.
So should I drop this property as well as it seems it will also remain true all the time?

dengzhhu653: I might refer to an old generated file. I think we can also drop this `hive.server.tcp.keepalive` as well

ramitg254: Sure I'll do that

ramitg254:
conf/hive-default.xml.template was in gitignore that's why force adding its changes are looking as new file in diff

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
ramitg254: Done

dengzhhu653:
> conf/hive-default.xml.template was in gitignore that's why force adding its changes are looking as new file in diff

could you drop this file? sounds like this file is generated automatically based on `HiveConf.ConfVars`. others look good to me

ramitg254:
> > conf/hive-default.xml.template was in gitignore that's why force adding its changes are looking as new file in diff
> 
> could you drop this file? sounds like this file is generated automatically based on `HiveConf.ConfVars`. others look good to me

done

dengzhhu653 (approved):

ramitg254:
Thanks @dengzhhu653 for the review

PR #5989, opened 2025-07-21
HIVE-29096: Enhance the metatool updateLocation for multiple table or…

dengzhhu653:
… serde properties

Enhance the metatool updateLocation to accept multiple properties

Sometimes we need to update multiple properties in case of NameNode switch. For example for Iceberg tables, both of the metadata_location and previous_metadata_location should update

Yes

The new added UT,
Running the example in the jira.

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreDirectSqlUtils.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetastoreDirectSqlUtils.java
@@ -56,7 +56,7 @@ import java.util.TreeMap;
 /**
  * Helper utilities used by DirectSQL code in HiveMetastore.
  */
-class MetastoreDirectSqlUtils {
+public class MetastoreDirectSqlUtils {
   private static final Logger LOG = LoggerFactory.getLogger(MetastoreDirectSqlUtils.class);
   private MetastoreDirectSqlUtils() {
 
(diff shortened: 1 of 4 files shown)


Aggarwal-Raghav:
@dengzhhu653, for the iceberg usecase mentioned in description, I believe updating location will not fully resolve the problem as metadata (snapshot.avro, metadata.json etc) also needs to be updated (unless there is some internal solution to update the metadata) explicitly. 

Reference: https://github.com/apache/iceberg/issues/1617 , I think we need to implement this proceduce in hive
https://iceberg.apache.org/docs/latest/spark-procedures/#rewrite_table_path
CC @deniskuzZ 

**But this PR will be helpful post updating the metadata in iceberg.**

dengzhhu653:
> @dengzhhu653, for the iceberg usecase mentioned in description, I believe updating location will not fully resolve the problem as metadata (snapshot.avro, metadata.json etc) also needs to be updated (unless there is some internal solution to update the metadata) explicitly.
> 
> Reference: [apache/iceberg#1617](https://github.com/apache/iceberg/issues/1617) , I think we need to implement this proceduce in hive https://iceberg.apache.org/docs/latest/spark-procedures/#rewrite_table_path CC @deniskuzZ
> 
> **But this PR will be helpful post updating the metadata in iceberg.**

Thank you @Aggarwal-Raghav for pointing it out. Not sure if the `rewrite_table_path` will update the `metadata_location` in HMS table, looks it's not, as it uses the `register_table` to load the new copied table in the last step.

deniskuzZ:
@Aggarwal-Raghav yes, you are right, that won't work for iceberg, at least not until https://github.com/apache/iceberg/issues/13141 is resolved.

PR #5990, opened 2025-07-23
HIVE-28819: [HiveAcidReplication] copy gets stuck for hive.repl.retry.total.duration for FileNotFoundException

harshal-16:
Problem Statement:
  * If some files are deleted from the source side while the copy operation is going on, then the replication job gets stuck for 24 hours
  * This was happening because FileNotFoundException was getting wrapped into IOException, and that is undera  retryable function with retry on IOException and exit on FileNotFoundException

Testing:
  * Added test case


common/src/java/org/apache/hadoop/hive/common/FileUtils.java
```diff
@@ -830,7 +830,14 @@ public static boolean copy(FileSystem srcFS, Path[] srcs, FileSystem dstFS, Path
           throw new IOException("copying multiple files, but last argument `" + dst + "' is not a directory");
         }
       } catch (FileNotFoundException var16) {
-        throw new IOException("`" + dst + "': specified destination directory does not exist", var16);
+        // Create a new FileNotFoundException with the custom message and the original message
+        FileNotFoundException e = new FileNotFoundException("'" + dst + "': specified destination directory does not exist");
```
deniskuzZ: that's a common code, I don't think it's safe to change it just to handle replication edge-case

ayushtkn: Mostly these copy functions are used by replication for data copy, but I think they got used somewhere else as well. But this method I remember was written for copying with Preserving Xattrs during Replication. Ideally we should not catch FNF just let it propagate back to the client rather than wrapping it in IOE. It shouldn't by incompatible FNF is a child class of IOE.

But if think there can be issues in that case @harshal-16 maybe the CopyUtils.retryableFxn() can be improved as well. 

Maybe something like
```
  private void isNonRetryableException(Exception e) throws IOException {
    Throwable current = e;
    while (current != null) {
      final Throwable curr = current;
      if (failOnParentExceptionList.stream().anyMatch(k -> k.isAssignableFrom(curr.getClass()))) {
        throw new IOException(e);
      }
      current = current.getCause();
    }
  }
```

or maybe use ExceptionUtils.getRootCause(e) or any other in the same utils class

harshal-16: Yes @ayushtkn , this copy function is used only by CopyUtils.copyFilesBetweenFS() which is specific to replication.

For your suggestion about improving retryableFxn(), the problem is below is the current implementation. and if you see 
It retries on IOException, and because FileNotFoundException was wrapped inside IOException, it will not go to the catch block before getting stuck in the retry loop.

```java
private <T> T retryableFxn(Callable<T> callable) throws IOException {
    Retryable retryable = Retryable.builder()
            .withHiveConf(hiveConf)
            .withRetryOnException(IOException.class).withFailOnParentExceptionList(failOnParentExceptionList).build();
    try {
      return retryable.executeCallable(() -> callable.call());
    } catch (Exception e) {
      if (failOnParentExceptionList.stream().anyMatch(k -> k.isAssignableFrom(e.getClass()))) {
        throw new IOException(e);
      }
      throw new IOException(ErrorMsg.REPL_FILE_SYSTEM_OPERATION_RETRY.getMsg(), e);
    }
  }

Comment (ayushtkn) [reply common/src/java/org/apache/hadoop/hive/common/FileUtils.java]: So, can't we fix that in ``executeCallable``, It has a catch block
```
} catch (Exception e) {
        if (this.failOnParentExceptions.stream().noneMatch(k -> k.isAssignableFrom(e.getClass()))
          && this.failOn.stream().noneMatch(k -> e.getClass().equals(k))
          && this.retryOn.stream().anyMatch(k -> e.getClass().isAssignableFrom(k))) {
          if (elapsedTimeInSeconds(startTime) + delay > this.totalDurationInSeconds) {
            // case where waiting would go beyond max duration. So throw exception and return
            throw e;
          }
```

in that play with ``k -> k.isAssignableFrom(e.getClass())`` & check if the exception doesn't have anyone of the exceptions  assignable mentioned in `failOnParentExceptions` wrapped inside it?

Comment (harshal-16) [reply common/src/java/org/apache/hadoop/hive/common/FileUtils.java]: Done, please have a look

File: common/src/java/org/apache/hadoop/hive/common/FileUtils.java [hunk 2]
```diff
@@ -843,6 +843,9 @@ public static boolean copy(FileSystem srcFS, Path[] srcs, FileSystem dstFS, Path
           if (!doIOUtilsCopyBytes(srcFS, srcFS.getFileStatus(src), dstFS, dst, deleteSource, overwrite, preserveXAttr, conf, copyStatistics)) {
             returnVal = false;
           }
+        } catch (FileNotFoundException var15) {
+          // Wrapping the FileNotFoundException in a new IOException and re-throw immediately.
+          throw new IOException("Copy operation failed", var15);
```
ayushtkn, line 848: why do you want to change exception in the scope of this PR. FNF is already sorted & not retried right?

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
```diff
@@ -244,4 +261,91 @@ public void testParallelCopySuccess() throws Exception {
...
+  @Test
+  public void testCopyFilesBetweenFSWithDestDirNotExistFailure() throws IOException {
+    Path srcPath1 = new Path(basePath, "file1.txt");
+    Path srcPath2 = new Path(basePath, "file2.txt");
+    Path dstPath = new Path(basePath, "copyDst");
+
+    try {
+      // Create source files
+      fileSystem.create(srcPath1).write("Content of file1".getBytes());
+      fileSystem.create(srcPath2).write("Content of file2".getBytes());
```
ayushtkn: who closes the file handle here? you just opened a FS stream & wrote something on it, didn't flush nor closed it

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
```diff
@@ -244,4 +261,91 @@ public void testParallelCopySuccess() throws Exception {
...
+
+
+      CopyUtils copyUtils = new CopyUtils("hive", hiveConf, fileSystem);
+      DataCopyStatistics copyStatistics = new DataCopyStatistics();
+      IOException thrown =
+              Assert.assertThrows(IOException.class, () -> {
+                copyUtils.copyFilesBetweenFS(fileSystem, srcPaths, fileSystem, dstPath, false, true, copyStatistics);
+              });
+      // this is supposed to come out of retryable function immediately without waiting as FileNotFound is not auto-recoverable error
+      Assert.assertEquals(IOException.class, thrown.getCause().getClass());
```
ayushtkn: If the above assertion passes, can this ever fail?

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
```diff
@@ -244,4 +261,91 @@ public void testParallelCopySuccess() throws Exception {
...
+      Assert.assertEquals(FileNotFoundException.class, thrown.getCause().getCause().getClass());
```
ayushtkn: does this work
```
      Assert.assertEquals(FileNotFoundException.class, ExceptionUtils.getRootCause(thrown).getClass());
```

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
```diff
@@ -244,4 +261,91 @@ public void testParallelCopySuccess() throws Exception {
...
+      Assert.assertEquals("'/tmp/copyDst': specified destination directory does not exist", thrown.getCause().getMessage());
```
ayushtkn: assertThrows throws as an option to assert the exception message as well, check if you can assert this as part of that only

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
```diff
@@ -244,4 +261,91 @@ public void testParallelCopySuccess() throws Exception {
...
+    } finally {
+      // Clean up
+      fileSystem.delete(srcPath1, false);
+      fileSystem.delete(srcPath2, false);
+      fileSystem.delete(dstPath, true);
+    }
+  }
+
+  @Test
+  public void testCopyFilesBetweenFSWithSourceFileGettingDeletedFailure() throws IOException {
```
ayushtkn: I don't think this test is required, it ain't testing the change which we did, the above test is enough & moreover the catch block below shows like this might not works sometime, so better to drop it

common/src/java/org/apache/hadoop/hive/common/FileUtils.java
harshal-16: Without this, I can not check the getCause() in https://github.com/apache/hive/pull/5990/files#diff-2c8fade5555e7ddbbb285946c7916f432a7b1764ac3115d6272a95c0749003faR105
As the existing implementation below at line no#851 just adds the message, and it loses the underlying cause object

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
harshal-16: Yes, this works, as I am throwing error in FileUtils.java, which preserves the cause
```java 
  } catch (FileNotFoundException var15) {
          // Wrapping the FileNotFoundException in a new IOException and re-throw immediately.
          throw new IOException("Copy operation failed", var15);
```

common/src/java/org/apache/hadoop/hive/common/FileUtils.java
ayushtkn: Why do we need to have it in Cause, if the exception itself is FNF. If the original exception is non retriable, it should bail out there itself, it shouldn't go for checking the cause itself, if it isn't a match then only it should progress on checking the cause

harshal-16: I didn't quite get your point here, Can please explain a bit more. While copy is going on, if files get cleaned from the source location (the most common scenario is the CM root getting cleaned), then it will throw FileNotFoundException, which is fatal, and there is no way the system will recover on its own like other IOException. 
Moreover, we may not want to throw all different exceptions; that's why we are wrapping all kinds of exceptions in IOException. 
That's what i am checking as part of 
```java
private boolean containsNonRetryableCause(Exception e) {
    Throwable current = e;
    while (current != null) {
      final Throwable curr = current;
      if (this.failOnParentExceptions.stream().anyMatch(k -> k.isAssignableFrom(curr.getClass()))) {
        return true;
      }
      current = current.getCause();
    }
    return false;
  }
```

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
harshal-16: ack, will update

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
harshal-16: ack, will check

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
harshal-16: ack

ql/src/test/org/apache/hadoop/hive/ql/parse/repl/TestCopyUtils.java
harshal-16: removed this

deniskuzZ:
hi @ayushtkn, looks like this PR is waiting on your input. could you please check

PR #5991, opened 2025-07-23
HIVE-25324: provide a config to disable PartitionManagementTask

saihemanth-cloudera:
Provided a config to disable partition management task

Cloud environments incur unnecessary costs as this background task runs frequently. This task may not be necessary as the partitions may not be added externally.

No.

Local env. Added unit test

### DIFF
--- a/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestPartitionManagement.java
+++ b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestPartitionManagement.java
@@ -49,7 +49,6 @@ import org.apache.hadoop.hive.metastore.client.builder.PartitionBuilder;
 import org.apache.hadoop.hive.metastore.client.builder.TableBuilder;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf.ConfVars;
-import org.apache.hadoop.hive.metastore.leader.AuditLeaderListener;
 import org.apache.hadoop.hive.metastore.leader.HouseKeepingTasks;
 import org.apache.hadoop.hive.metastore.leader.StaticLeaderElection;
 import org.apache.hadoop.hive.metastore.security.HadoopThriftAuthBridge;


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1257,6 +1257,10 @@ public enum ConfVars {
 
 
     // Partition management task params
+    PARTITION_MANAGEMENT_ENABLE("metastore.partition.management.enable", "hive.metastore.partition.management.enable",
```
dengzhhu653: how about setting the `metastore.partition.management.task.frequency` to 0 to disable the task?

deniskuzZ: +1, but maybe `-1` as default to be more explicit
see `metastore.iceberg.catalog.cache.expiry`, default is `-1`

saihemanth-cloudera: If we set it 0 or -1 then we didn't handle else condition properly here: https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/leader/HouseKeepingTasks.java#L116-L118
`long freq = task.runFrequency(TimeUnit.MILLISECONDS);
  runningTasks.add(task);
  metastoreTaskThreadPool.getPool().scheduleAtFixedRate(task, freq, freq, TimeUnit.MILLISECONDS);`
  and scheduleAtFixedRate() would throw illegalArgumentException.
  
  Also, I thought a separate config would make it clear for end-users. If you think there are too many configs and we should only work with `metastore.partition.management.task.frequency`, please let me know. Thanks.

dengzhhu653: I would prefer the `metastore.partition.management.task.frequency`, simply making it as 0 or -1, can achieve the same purpose with a new property setting as false explicitly, without adding extra document to train the user.

We should add the `freq > 0` validation in `HouseKeepingTasks`.

saihemanth-cloudera: Ack

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/PartitionManagementTask.java
```diff
@@ -91,43 +91,46 @@ public void run() {
...
-            new TableFetcher.Builder(msc, catalogName, dbPattern, tablePattern).tableTypes(tableTypes)
-                .tableCondition(
-                    hive_metastoreConstants.HIVE_FILTER_FIELD_PARAMS + "discover__partitions like \"true\" ")
-                .build()
-                .getTables();
-
-        if (candidates.isEmpty()) {
-          LOG.info("Got empty table list in catalog: {}, dbPattern: {}", catalogName, dbPattern);
-          return;
+        if (MetastoreConf.getTimeVar(conf, MetastoreConf.ConfVars.PARTITION_MANAGEMENT_TASK_FREQUENCY,
```
deniskuzZ: This seems redundant when we already have https://github.com/apache/hive/pull/5991/files#diff-3ff73ebef2cdbb1f6cd76a03cdc7683ac0aa52285185ebf82dc9f66ccc241997R117-R120, isn't it?

saihemanth-cloudera: The test directly runs partitionManagementTask.run() which directly comes here, so this condition is required for the test to run successfully. The other condition is used by housekeeping tasks that are run periodically. I can go ahead and remove the additional condition in HouseKeepingTasks class as we already have it here. Please let me know. Thanks.

deniskuzZ: Please update/fix the test and drop the redundant check

saihemanth-cloudera: I have updated the test to verify that the PartitionManagementTask is disabled when freq is set to zero, and the task is run via housekeeping tasks.

PR #5992, opened 2025-07-24
HIVE-29101: Add test coverage for the standalone jdbc

dengzhhu653:

### DIFF
--- a/itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItAbstractContainer.java
+++ b/itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItAbstractContainer.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hive.jdbc;
+
+public abstract class ItAbstractContainer {
+
+  protected String getZkConnectionUrl() {
+    throw new UnsupportedOperationException("not implemented");
+  }
+
+  protected String getHttpJdbcUrl(){
+    throw new UnsupportedOperationException("not implemented");
+  }
+
+  protected String getBaseJdbcUrl() {
+    throw new UnsupportedOperationException("not implemented");
+  }
+
+  public abstract void start() throws Exception;
+
+  public abstract void stop() throws Exception;
+
+}
--- a/itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItHiveServer2.java
+++ b/itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItHiveServer2.java
@@ -1,0 +1,270 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hive.jdbc;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.minikdc.MiniKdc;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.testcontainers.containers.BindMode;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.AbstractWaitStrategy;
+import org.testcontainers.utility.DockerImageName;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class ItHiveServer2 extends ItAbstractContainer {
+  protected File workDir;
+  private final MiniJdbcKdc miniKdc;
+  protected GenericContainer<?> container;
+  protected final String imageName = "apache/hive:4.0.1";
+  private int[] realmPorts;
+
+  public ItHiveServer2() {
+    this.miniKdc = null;
+  }
+
+  public ItHiveServer2(File workDir) throws Exception {
+    this.workDir = new File(workDir, "test-standalone-jdbc-file");
+    this.miniKdc = new MiniJdbcKdc(this.workDir);
+
+    String principal = MiniJdbcKdc.HIVE_TEST_USER_1;
+    Configuration configuration = new Configuration();
+    configuration.set("hadoop.security.authentication", "kerberos");
+    UserGroupInformation.setConfiguration(configuration);
+    UserGroupInformation.loginUserFromKeytab(principal, miniKdc.getKeyTabFile(principal));
+    assertEquals(principal, UserGroupInformation.getLoginUser().getShortUserName());
+  }
+
+  protected Map<String, String> prepareEnvArgs() {
+    Map<String, String> envArgs = new HashMap<>();
+    envArgs.put("SERVICE_NAME", "hiveserver2");
+    envArgs.put("HIVE_SERVER2_TRANSPORT_MODE", "all");
+    List<String> properties = new ArrayList<>();
+    properties.add("hive.server2.authentication=KERBEROS");
+    properties.add("hive.server2.authentication.kerberos.principal=hive/_HOST@" + miniKdc.getRealm());
+    properties.add("hive.server2.authentication.kerberos.keytab=/opt/hive/conf/hive.keytab");
+    StringBuilder builder = new StringBuilder();
+    for (String prop : properties) {
+      builder.append("-D").append(prop).append(" ");
+    }
+    envArgs.put("SERVICE_OPTS", builder.toString());
+    return envArgs;
+  }
+
+  private Map<String, String> prepareFiles() {
+    Map<String, String> boundFiles = new HashMap<>();
+
+    boundFiles.put(createNewKrbConf().getPath(), "/etc/krb5.conf");
+    boundFiles.put(miniKdc.getKeyTabFile(miniKdc.getServicePrincipalForUser("hive")), "/opt/hive/conf/hive.keytab");
+    StringBuilder core_site = new StringBuilder("<configuration>").append(System.lineSeparator());
+    core_site.append("<property>\n" + "    <name>hadoop.security.authentication</name>\n" + "    <value>kerberos</value>\n" + "  </property>");
+    core_site.append("<property>\n" + "    <name>hadoop.proxyuser.hive.groups</name>\n" + "    <value>*</value>\n" + "  </property>");
+    core_site.append("<property>\n" + "    <name>hadoop.proxyuser.hive.hosts</name>\n" + "    <value>*</value>\n" + "  </property>");
+    core_site.append("</configuration>");
+    boundFiles.put(genConfFile("core-site.xml", core_site.toString()), "/opt/hive/conf/core-site.xml");
+
+    // java.io.IOException: Can't get Master Kerberos principal for use as renewer if missing the yarn-site
+    StringBuilder yarn_site = new StringBuilder("<configuration>").append(System.lineSeparator());
+    yarn_site.append("<property>\n" + "    <name>yarn.resourcemanager.principal</name>\n" +
+        "    <value>yarn/_HOST@" + miniKdc.getRealm() + "</value>\n" + "  </property>");
+    yarn_site.append("</configuration>");
+    boundFiles.put(genConfFile("yarn-site.xml", yarn_site.toString()), "/opt/hive/conf/yarn-site.xml");
+
+    return boundFiles;
+  }
+
+  @Override
+  public void start() {
+    container = new GenericContainer<>(DockerImageName.parse(imageName))
+        .withEnv(prepareEnvArgs())
+        .waitingFor(new AbstractWaitStrategy() {
+          @Override
+          protected void waitUntilReady() {
+            long timeout = TimeUnit.MINUTES.toMillis(15);
+            long start;
+            do {
+              start = System.currentTimeMillis();
+              try (Connection conn = DriverManager.getConnection(getBaseJdbcUrl())) {
+                break;
+              } catch (Exception e) {
+                try {
+                  Thread.sleep(10 * 1000);
+                } catch (InterruptedException ex) {
+                  break;
+                }
+              }
+            } while ((timeout += start - System.currentTimeMillis()) > 0);
+          }
+        });
+    beforeStart(container);
+    container.start();
+  }
+
+  protected void beforeStart(GenericContainer<?> container) {
+    Map<String, String> boundFiles = prepareFiles();
+    for (Map.Entry<String, String> entry : boundFiles.entrySet()) {
+      container.withFileSystemBind(entry.getKey(), entry.getValue(), BindMode.READ_ONLY);
+    }
+    container
+        .withCreateContainerCmdModifier(it -> it.withHostName(MiniJdbcKdc.HOST))
+        .withExposedPorts(10000, 10001);
+    if (realmPorts != null) {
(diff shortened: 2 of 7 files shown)


dengzhhu653:
cc @zhangbutao

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItAbstractContainer.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hive.jdbc;
+
+public abstract class ItAbstractContainer {
```
zhangbutao: How about `ITAbstractContainer ` & `ITHiveServer2 `&  `ITZKHiveServer2`?

We already similay naming like `ITestDbTxnManager`.

pom.xml
```diff
@@ -61,6 +61,7 @@
     <module>standalone-metastore</module>
     <module>kafka-handler</module>
     <module>iceberg</module>
+    <module>itests/test-jdbc</module>
```
zhangbutao: Why we add this module in root pom?

Shouldn't this module be placed here https://github.com/apache/hive/blob/master/itests/pom.xml?

And then build this module through `-Pitests`
`mvn clean package -Pitests`

itests/test-jdbc/src/test/resources/custom_hosts_file
```diff
@@ -0,0 +1,3 @@
+127.0.0.1       localhost
+127.0.0.1       test-standalone-jdbc-plain
+127.0.0.1       test-standalone-jdbc-kerberos
```
zhangbutao, line 3: Does the JDBC JAR relate to the authentication mode? If there is no relationship, can we keep only the simplest plain mode to simplify the code?

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItHiveServer2.java
```diff
@@ -0,0 +1,270 @@
...
+import org.testcontainers.utility.DockerImageName;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class ItHiveServer2 extends ItAbstractContainer {
+  protected File workDir;
+  private final MiniJdbcKdc miniKdc;
+  protected GenericContainer<?> container;
+  protected final String imageName = "apache/hive:4.0.1";
```
zhangbutao: Here I have two concerns:  

1) Can we avoid using the Hive Docker approach and instead start a local Hive process like other integration tests, then perform standalone JDBC testing? Because the Hive CI frequently encounters issues like download timeouts or failures for the docker image. If we can start a Hive service locally, we can avoid such problems.  

2) If the Docker approach is mandatory to meet testing requirements, how should we control the Hive image version? Could there be compatibility issues between Hive 4.0.1 server and Hive 4.2.0 & 5.0.0 jdbc jars?

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,226 @@
...
+    ZK_HS2 = new ItZKHiveServer2(workDir);
+
+    HS2.start();
+    ZK_HS2.start();
+  }
+
+  @Test
+  public void testBinaryJdbc() throws Exception {
+    testMetaOp(HS2.getBaseJdbcUrl(), true);
+    //testMetaOp(ZK_HS2.getJdbcUrl(), false);
```
zhangbutao: Remove this line if we no need.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItHiveServer2.java
zhangbutao: BTW, i noticed the execution time of CI is 3h. But the execution time of others PR's CI is about 2.5h.

https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5993/2/pipeline
https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5991/1/pipeline
https://ci.hive.apache.org/blue/organizations/jenkins/hive-precommit/detail/PR-5987/1/pipeline


Maybe we can run multipe times to double check if this PR can increase the execution time of CI.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItAbstractContainer.java
dengzhhu653: ack

pom.xml
dengzhhu653: ack, should go into itests/pom.xml

itests/test-jdbc/src/test/resources/custom_hosts_file
dengzhhu653: This is needed for testing the zookeeper jdbc, the HS2 will register the hostname to zookeeper, and jdbc client reads this hostname from zookeeper and uses it to connect to HS2. if we miss the host file, the jdbc cannot resolve the actual HS2 address, leading to unknown host exception

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ItHiveServer2.java
dengzhhu653: Starting the HS2 locally means we have to add the classes needed to the classpath, then we have no way to tell whether a class should be added or removed to/from the standalone jdbc or not, the Docker provides a way to isolate the HS2 from the standalone jdbc completely.
Detecting the compatibility issue, then fixing it early, is also a purpose of this PR.  I think we need to avoid the compatibility issue in the future releases, especially for the jdbc, we cannot always make sure the upgrade of client will keep pace with the HS2.
For the CI, last run of the same PR took about 2.5h,  the image size is 1.6gb, which is near to the size of Oracle(1.9gb) in https://github.com/apache/hive/blob/master/itests/util/src/main/java/org/apache/hadoop/hive/ql/externalDB/Oracle.java.  I think it's ok from the pipeline history regarding the timeout or failure on installing the Oracle db.

itests/test-jdbc/src/test/resources/custom_hosts_file
zhangbutao: Do we need to run the test in the kerberos mode?  Is the plain model not sufficient?

dengzhhu653: Not enough I think, For example the Delegation token is based on Kerberos, we need some hadoop classes to support it

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ITHiveServer2.java
```diff
@@ -0,0 +1,270 @@
...
+    envArgs.put("SERVICE_OPTS", builder.toString());
+    return envArgs;
+  }
+
+  private Map<String, String> prepareFiles() {
+    Map<String, String> boundFiles = new HashMap<>();
+
+    boundFiles.put(createNewKrbConf().getPath(), "/etc/krb5.conf");
+    boundFiles.put(miniKdc.getKeyTabFile(miniKdc.getServicePrincipalForUser("hive")), "/opt/hive/conf/hive.keytab");
+    StringBuilder core_site = new StringBuilder("<configuration>").append(System.lineSeparator());
```
zhangbutao: Should we consider putting the content of core_site and yarn_site into local files core-site.xml and yarn-site.xml? This would make it easier to extend and add related configurations later without modifying the code here to add new configurations.

dengzhhu653: Good catch, will do

InvisibleProgrammer:
Hi, I wonder, why it is in itests, not in the jdbc directory?

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ITAbstractContainer.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hive.jdbc;
+
+public abstract class ITAbstractContainer {
```
InvisibleProgrammer, line 20: What does the IT prefix mean in those classes? ITAbstractContainer, ITHiveServer2, ...

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+    workDir.mkdirs();
+    HS2 = new ITHiveServer2(workDir);
+    ZK_HS2 = new ITZKHiveServer2(workDir);
+
+    HS2.start();
+    ZK_HS2.start();
+  }
+
+  @Test
+  public void testBinaryJdbc() throws Exception {
```
InvisibleProgrammer, line 62: It looks like 3 different tests, not one.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+        // describe table
+        sql = "describe " + tableName;
+        System.out.println("Running: " + sql);
+        res = stmt.executeQuery(sql);
+        while (res.next()) {
+          System.out.println(res.getString(1) + "\t" + res.getString(2));
+        }
+
+        // "Client cannot authenticate via:[TOKEN, KERBEROS]" in running the Kerberized tez local mode
+        if (!kerberos) {
```
InvisibleProgrammer, line 95: It looks like two different test cases: one with Kerberos, and one without Kerberos.
Also it only has assertion if it is a not Kerberos test case. 
In case of kerberos == true, it tests nothing.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+    HS2 = new ITHiveServer2(workDir);
+    ZK_HS2 = new ITZKHiveServer2(workDir);
+
+    HS2.start();
+    ZK_HS2.start();
+  }
+
+  @Test
+  public void testBinaryJdbc() throws Exception {
+    testMetaOp(HS2.getBaseJdbcUrl(), true);
```
InvisibleProgrammer, line 63: I don't think if HS2 as a parameter is necessary: it is a private static variable of this test class.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+      }
+      ResultSet resultSet = metaData.getColumns("hive", ".*", ".*", ".*");
+      assertTrue(resultSet.next());
+      while (resultSet.next()) {
+        System.out.println("DatabaseMetaData.getColumns: " + rs.getString(1));
+      }
+      resultSet = metaData.getTypeInfo();
+      assertTrue(resultSet.next());
+      while (resultSet.next()) {
+        System.out.println("DatabaseMetaData.getTypeInfo: " + rs.getString(1));
```
InvisibleProgrammer: What is the purpose of the System.out... calls? Especially after the last assertion.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+  private void testMetaOp(String url, boolean kerberos) throws Exception {
+    try (Connection con = DriverManager.getConnection(url)) {
+      try (Statement stmt = con.createStatement()) {
+        String tableName = "testHiveDriverTable1";
+        stmt.execute("drop table if exists " + tableName);
+        stmt.execute("create table " + tableName + " (key string, value string)");
+        // show tables
+        String sql = "show tables '" + tableName + "'";
+        System.out.println("Running: " + sql);
+        ResultSet res = stmt.executeQuery(sql);
```
InvisibleProgrammer: Do we need to close the ResultSet objects?

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+      assertTrue(resultSet.next());
+      while (resultSet.next()) {
+        System.out.println("DatabaseMetaData.getTypeInfo: " + rs.getString(1));
+      }
+    }
+  }
+
+  @Test
+  public void testHttpJdbc() throws Exception {
+    testMetaOp(HS2.getHttpJdbcUrl(), true);
```
InvisibleProgrammer, line 144: It looks like 4 different tests, not one.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+    } catch (SQLException e) {
+      e.printStackTrace();
+    } finally {
+      con.close();
+    }
+  }
+
+  @Test
+  public void testZookeeperJdbc() throws Exception {
+    testMetaOp(ZK_HS2.getZkConnectionUrl(), false);
```
InvisibleProgrammer, line 162: It looks like 3 different test cases, not one.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+      os.write(dt.getBytes());
+    }
+    SecurityUtils.setTokenStr(UserGroupInformation.getCurrentUser(), dt, "hiveserver2ClientToken");
+    //  HADOOP_TOKEN_FILE_LOCATION
+    updateEnv("HADOOP_TOKEN_FILE_LOCATION", dtFile.getPath());
+    url += ";auth=delegationToken";
+    testMetaOp(url, true);
+  }
+
+  public void updateEnv(String name, String val) throws ReflectiveOperationException {
```
InvisibleProgrammer: Can it be private?

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
InvisibleProgrammer: https://github.com/junit-team/junit4/wiki/parameterized-tests

dengzhhu653:
> Hi, I wonder, why it is in itests, not in the jdbc directory?

Thanks @InvisibleProgrammer. This is for verifying we have packaged all necessary classes into the standalone jdbc jar, it's testing whether this standalone jar is self-contained or not. If we put it in the jdbc directory, we have no way to tell whether a class should be added or removed to/from the standalone jdbc or not.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/ITAbstractContainer.java
dengzhhu653: The first character of Integration Test

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
dengzhhu653: yeah, this is different from the jdbc unit test, we are verifying:
1. packaged all necessary classes into the standalone jdbc jar;
2. if have the compatible issue

If there is no problem running these queries, e.g, no class def found or unexpected exception, then we can say this standalone jar can work as expected

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
dengzhhu653: All of them are verifying the binary connection, a user could use this connection for these kinds of operations.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
dengzhhu653: We are testing zk discovery, http, binary and kerberos/token based auth in this class, making the method accept the url can reuse the same `testMetaOp` in the following tests.

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
dengzhhu653: To check  no unexpected exception thrown from  from `rs.getString(1)`

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
dengzhhu653: check the http mode connection

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
dengzhhu653: check the zk discovery

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
dengzhhu653: nice catch

itests/test-jdbc/src/test/java/org/apache/hive/jdbc/TestStandaloneJdbc.java
dengzhhu653: The non-Kerberos case is Tez job needed, otherwise we would see the "Client cannot authenticate via:[TOKEN, KERBEROS]" while submitting the Tez job, I'm not sure how to solve it.

dengzhhu653:
Thank you for the comments and review, cc @zhangbutao @InvisibleProgrammer

PR #5993, opened 2025-07-24
HIVE-29098 : Fix dataloss post migrating a partitioned table with default partition to iceberg

vikramahuja1001:

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java
@@ -32,7 +32,6 @@ import java.util.Properties;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.stream.IntStream;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.LocatedFileStatus;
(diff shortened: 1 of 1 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java
```diff
@@ -169,7 +169,10 @@ private static List<DataFile> getDataFiles(RemoteIterator<LocatedFileStatus> fil
           dataFile -> {
             StructLike structLike = dataFile.partition();
             for (int pos : stringFields) {
-              structLike.set(pos, FileUtils.unescapePathName(structLike.get(pos, String.class)));
+              String val = structLike.get(pos, String.class);
```
deniskuzZ: @vikramahuja1001 i think we can simply revert HIVE-27367,  https://github.com/apache/iceberg/pull/7744 fixed the issue. Could you please try it out.

vikramahuja1001: Thanks for the review @deniskuzZ . I will check the same and revert back.

vikramahuja1001: @deniskuzZ , simple reverting the code part of HIVE-27367 works in this case. I have reverted the code but have kept the test case.
Also added new tests as well as a part of this PR.

vikramahuja1001:
Simply reverting [HIVE-27367](https://issues.apache.org/jira/browse/HIVE-27367), https://github.com/apache/iceberg/pull/7744 fixes this issue. Reverted the code part of HIVE-27367 and have kept the test case to prevent any future regressions wrt HIVE-27367. Also added more tests as a part of this PR so that no regression happens in the future.
cc @deniskuzZ

PR #5997, opened 2025-07-28
HIVE-29106: When drop a table, there is no need to verify the directo…

zxl-333:
…ry permission

When drop a table, there is no need to verify the directory permission. An exception should be thrown when the table path deletion fails.

When HiveServer2 integrates Ranger for authentication, there will be a bug in the permission verification. Ranger has already granted write permissions for the database tables and even partitions. However, when attempting to delete a table, it still prompts that there is no permission for the database directory or the table directory, resulting in the inability to delete the table.

The correct solution is to remove the table without verifying the directory write permission, not to catch the exception when deleting the data directory, but to throw the exception instead, and at the same time roll back the metadata information.

No

### DIFF
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
@@ -3050,18 +3050,6 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
       firePreEvent(new PreDropTableEvent(tbl, deleteData, this));
 
       tableDataShouldBeDeleted = checkTableDataShouldBeDeleted(tbl, deleteData);
-      if (tableDataShouldBeDeleted && tbl.getSd().getLocation() != null) {
-        tblPath = new Path(tbl.getSd().getLocation());
-        String target = indexName == null ? "Table" : "Index table";
-	 // HIVE-28804 drop table user should have table path and parent path permission
-        if (!wh.isWritable(tblPath.getParent())) {
-          throw new MetaException("%s metadata not deleted since %s is not writable by %s"
-              .formatted(target, tblPath.getParent(), SecurityUtils.getUser()));
-        } else if (!wh.isWritable(tblPath)) {
-          throw new MetaException("%s metadata not deleted since %s is not writable by %s"
-              .formatted(target, tblPath, SecurityUtils.getUser()));
-        }
-      }
 
       // Drop the partitions and get a list of locations which need to be deleted
       // In case of drop database cascade we need not to drop the partitions, they are already dropped.
@@ -3090,11 +3078,16 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
       if (!success) {
         ms.rollbackTransaction();
       } else if (tableDataShouldBeDeleted) {
-        // Data needs deletion. Check if trash may be skipped.
-        // Delete the data in the partitions which have other locations
-        deletePartitionData(partPaths, ifPurge, ReplChangeManager.shouldEnableCm(db, tbl));
-        // Delete the data in the table
-        deleteTableData(tblPath, ifPurge, ReplChangeManager.shouldEnableCm(db, tbl));
+        try {
+          // Data needs deletion. Check if trash may be skipped.
+          // Delete the data in the partitions which have other locations
+          deletePartitionData(partPaths, ifPurge, ReplChangeManager.shouldEnableCm(db, tbl));
+          // Delete the data in the table
+          deleteTableData(tblPath, ifPurge, ReplChangeManager.shouldEnableCm(db, tbl));
+        } catch (Exception e) {
+          ms.rollbackTransaction();
+          throw new MetaException(org.apache.hadoop.util.StringUtils.stringifyException(e));
+        }
       }
 
       if (!listeners.isEmpty()) {
@@ -3216,7 +3209,8 @@ public class HMSHandler extends FacebookBase implements IHMSHandler {
         wh.deleteDir(path, true, ifPurge, shouldEnableCm);
       }
     } catch (Exception e) {
-      LOG.error("Failed to delete directory: {}", path, e);
+      throw new MetaException("Failed to delete directory: %s . Exception detail : %s")
+          .formatted(path, org.apache.hadoop.util.StringUtils.stringifyException(e));
     }
   }
 
--- a/ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/TestHiveMetaStoreAuthorizer.java
@@ -23,6 +23,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.common.FileUtils;
 import org.apache.hadoop.hive.conf.HiveConf;
@@ -69,6 +70,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.eq;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -846,16 +848,20 @@ public class TestHiveMetaStoreAuthorizer {
         .build(conf);
     hmsHandler.create_table(table);
 
-    Path tablePath = new Path(table.getSd().getLocation());
-    when(wh.isWritable(Mockito.eq(tablePath.getParent()))).thenReturn(true);
-    when(wh.isWritable(Mockito.eq(tablePath))).thenReturn(false);
+    FileStatus fileStatus = Mockito.mock(FileStatus.class);
+    when(fileStatus.isDirectory()).thenReturn(true);
+    when(fileStatus.getPath()).thenReturn(new Path(tblName));
+    when(wh.getFs().listStatus(new Path(tblName))).thenReturn(new FileStatus[] { fileStatus });
+
+    doThrow(new FileNotFoundException("Failed to delete director:"))
+       .when(wh.getFs()).delete(any(Path.class), anyBoolean());
 
     try {
-      hmsHandler.drop_table("default", tblName, true);
+      hmsHandler.drop_table(default_db, tblName, true);
+      fail("Expected exception to be thrown due to lack of write permission");
     } catch (MetaException e) {
-      String expected = "%s metadata not deleted since %s is not writable by %s"
-          .formatted("Table", tablePath.toString(), authorizedUser);
-      assertEquals(expected, e.getMessage());
+      String expected = "Failed to delete director:";
+      assertTrue(e.getMessage().contains(expected));
     }
   }
 }


zxl-333:
@deniskuzZ  Please review the code.Thanks

zxl-333:
@ayushtkn Please help review the code.Thank you.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -3050,18 +3050,6 @@ private boolean drop_table_core(final RawStore ms, final String catName, final S
       firePreEvent(new PreDropTableEvent(tbl, deleteData, this));
 
       tableDataShouldBeDeleted = checkTableDataShouldBeDeleted(tbl, deleteData);
-      if (tableDataShouldBeDeleted && tbl.getSd().getLocation() != null) {
```
deniskuzZ, line 3053: @zxl-333 initially, you added the permission check in HIVE-28804, mentioning that some Hive clusters don't use Ranger and now you're removing it? 
I'm not sure I understand the reasoning. 

I see you've added `ms.rollbackTransaction()`. So the flow is: make a DB call to drop the metadata, and then depending on whether the directory on the filesystem was successfully removed, you either commit or roll back, right?

zxl-333: @deniskuzZ  When Ranger is not enabled, there is no issue with HIVE-28804. However, when Ranger is enabled, the HMS code fails to retrieve authorization information from Ranger during verification, leading to the following situation: even though write permissions for the table have been granted through Ranger, if the HDFS ACL permissions do not include write access, the user will be prompted with a "no write permission" error, resulting in the failure to drop the table.

Now, we need to resolve the issue where table deletion fails when permissions are available and succeeds when permissions are unavailable, regardless of whether Ranger is enabled. The solution is to skip the verification of HDFS ACL permissions for database and table directories. Instead, the permission verification during deletion should be entrusted to the NameNode. If the deletion fails due to permission issues, the metadata must also be rolled back, and the engine side will receive the exception information about the deletion failure.

deniskuzZ: @zxl-333, please share detailed step-by-step reproduce. I don't recall any customer using Ranger raising escalations related to that functionality.

deniskuzZ: cc @saihemanth-cloudera, @dengzhhu653

saihemanth-cloudera: @zxl-333 - Based on the scenario you mentioned, I believe this is a Hive or Ranger config issue.
Regarding Hive config, what is hive.server2.enable.doAs set to? It should be set to true. By doing so, assuming that you are running queries from jdbc client like beeline, when an end-user runs some query, user 'hive' does all the operations in the HDFS. Now if an end-user drops a table, user 'hive' will try to drop the data for the table path and its partitions.  Now user 'hive' is having issues with the permissions with the path that leads to the Ranger config issue.
Regarding ranger config, user 'hive' should be added in hdfs service in the ranger policy for hive warehouse table path, and I believe the issue here because of this missing policy in hdfs service. Also, when ranger policies are missing, the Hadoop ACLs permissions takes into effect.

So I feel, this is not a product issue, and you might have to abandon this patch.

zxl-333: @saihemanth-cloudera @deniskuzZ 
Do not use ranger. Refer to the steps in HIVE-18888.
When using ranger, even if ranger has been authorized, it may still result in the failure of the ACL write permission verification in HDFS, preventing the deletion of the table.

specific steps are as follows:
1、User_a has only **read** permissions for the test db and the following tables in **HDFS**.
hdfs://nn1/usr/warhouse/hive/test.db 
2、However, now the **read and write** permissions for the 'test' db and its tables have been granted through the '**ranger**' authorization.
3. If 'user_a' attempts to delete the 'test_1' table within the 'test' library, the following exception will be thrown
**test_1 metadata not deleted since hdfs://nn//usr/warhouse/hive/test.db is not writable by user_a**
The exception thrown above will result in the user being unable to delete even if the ranger authorization is granted.

Therefore, regardless of whether the ranger permission check is enabled or not, in order to ensure that users do not have read or write permissions for the database tables, the operation must fail and the metadata database transaction must be rolled back.

saihemanth-cloudera: From your comments, you seem to use Ranger only for metadata permissions and not file permissions (HDFS service is not being used in Ranger for auth OR there are no policies for user 'hive'). So you would rely on Hadoop ACLs for file authorization. So (2) you mentioned doesn't apply. And (3) is expected to happen and it should happen in the absence of Ranger related Hadoop policies.

zxl-333: Yes. After enabling the ranger control of HDFS file permissions, it is usually not necessary to use HDFS's ACL permissions for re-authorization. If this modification is not made, then after using ranger to control HDFS permissions, it is still necessary to grant HDFS ACL permissions to the user. This is unacceptable because the user needs to be authorized twice for a table path.

saihemanth-cloudera: @zxl-333 -- Can you please give me a detailed repro of this issue (probably a doc with screenshots with ranger policies, core-site.xml (in Hadoop conf) etc.), as previously asked by @deniskuzZ as well.
Because I cannot get this whatsoever

> then after using ranger to control HDFS permissions, it is still necessary to grant HDFS ACL permissions to the user.

If you are using to control HDFS permissions, then you don't need to grant the end user the required permissions. 
I'm guessing that you are missing this config in your core-site.xml of Hadoop service:
`  <property>
        <name>hadoop.proxyuser.hive.hosts</name>
        <value>*</value>
    </property>`
 `   <property>
        <name>hadoop.proxyuser.hive.groups</name>
        <value>*</value>
    </property>`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -3084,17 +3076,26 @@ private boolean drop_table_core(final RawStore ms, final String catName, final S
...
+          deletePartitionData(partPaths, ifPurge, ReplChangeManager.shouldEnableCm(db, tbl));
+          // Delete the data in the table
+          deleteTableData(tblPath, ifPurge, ReplChangeManager.shouldEnableCm(db, tbl));
+        }
         success = ms.commitTransaction();
       }
+    } catch (Exception e) {
+      ms.rollbackTransaction();
+      LOG.error("Failed to drop table " + name + ", rolling back transaction", e);
+      if (e instanceof MetaException) {
```
deniskuzZ, line 3091: throw ExceptionHandler.newMetaException(e)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -3216,7 +3217,8 @@ private void deleteDataExcludeCmroot(Path path, boolean ifPurge, boolean shouldE
         wh.deleteDir(path, true, ifPurge, shouldEnableCm);
       }
     } catch (Exception e) {
-      LOG.error("Failed to delete directory: {}", path, e);
+      throw new MetaException("Failed to delete directory: " + path.toString() + " . Exception detail : "
```
deniskuzZ, line 3220: throw ExceptionHandler.newMetaException(e)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
deniskuzZ: why do we need this?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
deniskuzZ: keep the log.error and maybe throw IOException. that is not a MetaException

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
zxl-333: The core-site has already set these parameters.
**steps:**
1、create database and table
24bf816baa5448ed92ef7269d72c83f1_compress
2、Check the HDFS permissions for the database and table
0a25709a32c2ed837f03c68cbf32d851_compress
3、Grant all permissions to the dmp_query user through the ranger authorization.
1dfedc7f796e09fc67dcdc5be1e2be59_compress
9cca9ae0906ab9a47fcb7e18623d4d4f_compress
4、Verify whether the dmp_query user has write permissions
73e4316733764c64bde4c67767b6054f_compress
5、Use the dmp_query user to drop this table
b286af7dd47afffdffbe0d65c0f9781e_compress
We observed that although ranger was authorized, the inability to delete was caused by the permission verification stored in HDFS.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
zxl-333: When the deletion of HDFS data fails, we ensure that the metadata data of HMS is rolled back.

deniskuzZ: I meant, why do we need to remap the exception to `MetaException`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
zxl-333: I have changed

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
zxl-333: I have changed

zxl-333:
@deniskuzZ please check again

zxl-333:
@saihemanth-cloudera please check again

zxl-333:
@deniskuzZ  I can't see the error message indicating the compilation failure.

zxl-333:
> @deniskuzZ I can't see the error message indicating the compilation failure.

@saihemanth-cloudera

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
saihemanth-cloudera: The user permissions in step 3 are incorrect, 
with hive.server2.enable.doAs set to false, user 'hive' will do all the operations in HDFS on behalf of the end user. Step 2 emphasizes the fact that the directory is owned by user 'hive', so in the HDFS ranger policy, 'hive' user should be granted with read/write permissions.
The end-user permissions on the table would be managed in Hadoop-SQL policies anyway.

deniskuzZ:
hi @zxl-333, sorry for late reply. 
I don't see the compilation failure, but qtest outputs have changed, try running with `-Dtest.output.overwrite=true` and see the diff
https://ci.hive.apache.org/job/hive-precommit/job/PR-5997/10/
````
MetaException(message:java.io.IOException: Failed to delete directory: file:/tmp . Exception detail : MetaException(message:Unable to delete directory: file:/tmp)
 at org.apache.hadoop.hive.metastore.HiveMetaStoreFsImpl.deleteDir(HiveMetaStoreFsImpl.java:45)
 at org.apache.hadoop.hive.metastore.Warehouse.deleteDir(Warehouse.java:475)
 at org.apache.hadoop.hive.metastore.HMSHandler.deleteDataExcludeCmroot(HMSHandler.java:3208)
 at org.apache.hadoop.hive.metastore.HMSHandler.deleteTableData(HMSHandler.java:3121)
 at org.apache.hadoop.hive.metastore.HMSHandler.drop_table_core(HMSHandler.java:3079)
 at org.apache.hadoop.hive.metastore.HMSHandler.drop_table_req(HMSHandler.java:3366)
````

saihemanth-cloudera:
@zxl-333 - Can you please address this comment: https://github.com/apache/hive/pull/5997#discussion_r2427342639

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
zxl-333: What's wrong with Step 3? Step 3 now authorizes a client user dmp_query, thereby granting this user read/write permissions to the table in the Ranger policy. 
When hive.server2.enable.doAs is false, the HDFS Ranger policy should grant read/write permissions to the "hive" user.

aturoczy:
Can we help to @zxl-333 to make it happen?

PR #5998, opened 2025-07-29
HIVE-28222: Ambiguous table alias exception for queries with self joins

zabetak:
1. Collect all table aliases under joins
2. Introduce derived table whenever the same alias appears more than once

Some queries that contain joins of the same table more than twice fail during compilation while trying to transform the optimized AST (obtained from CBO) to an Operator tree. The "Ambiguous table alias" exception is raised when a HiveJoin contains the same table scan multiple times (without an interleaving project) cause the generated AST is ambiguous.

No.

```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile_regex=cbo_self_join.*
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/PlanModifierForASTConv.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/PlanModifierForASTConv.java
@@ -33,7 +33,6 @@ import org.apache.calcite.adapter.jdbc.JdbcRules;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.plan.hep.HepRelVertex;
 import org.apache.calcite.plan.volcano.RelSubset;
-import org.apache.calcite.rel.RelHomogeneousShuttle;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.SingleRel;
 import org.apache.calcite.rel.core.Aggregate;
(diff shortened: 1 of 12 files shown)


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/PlanModifierForASTConv.java
```diff
@@ -123,29 +127,35 @@ public static RelNode convertOpTree(RelNode rel, List<FieldSchema> resultSchema,
     return newTopNode;
   }
 
-  private static String getTblAlias(RelNode rel) {
+  private static class AliasCollector extends RelHomogeneousShuttle {
+    List<String> aliases = new ArrayList<>();
```
kasakrisz: What happens when `accept` is called more than once on the same instance ? Is this list reseted before each?

zabetak: I discovered some issues with the previous code so I reworked the approach and force-pushed a new version. Apologies for rebasing and squashing, I haven't seen your comments before pushing the new version.

kasakrisz: No worries. Please let me know if you need another review.

zabetak: @kasakrisz The PR is now ready for review, please take a look when you get the chance.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/PlanModifierForASTConv.java
```diff
@@ -230,6 +225,47 @@ private static void convertOpTree(RelNode rel, RelNode parent) {
...
+   * The handler traverses the plan and collects aliases of tables it encounters for each join branch.
+   * When the same alias occurs in both branches of a join, it introduces a derive table (Project)
+   * over the left branch to break the ambiguity.
+   */
+  private static class SelfJoinHandler extends RelHomogeneousShuttle {
+    private final Set<String> aliases = new HashSet<>();
+
+    @Override
+    public RelNode visit(final RelNode rel) {
+      if (rel instanceof Join join) {
```
kasakrisz: Can `HiveRelShuttleImpl` be used instead of `RelHomogeneousShuttle` to get rid of `rel instanceof <Something>`?

zabetak: @kasakrisz I applied your suggestion in https://github.com/apache/hive/pull/5998/commits/b12cff8cfbff6453e41fda561e9241c0b3a0cb67 and tests are again green. Please take another look when you find some bandwitdth.

kasakrisz (approved):

zabetak:
Many thanks for the review @kasakrisz !

PR #6002, opened 2025-07-30
HIVE-29112: HMS embedded servlets https support 

henrib:
- added configuration variable hive.metastore.httpserver.use.https that drives if catalog/property servlets are using https;
- requires the same keystore/truststore configuration than useSSL;
- added tests & self signed certificates (good for 10 years);

Configuration embedded servlets to be served through an https connector

Security

No

Unit tests


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java
```diff
@@ -310,7 +310,18 @@ static void loginServerPrincipal(Configuration conf) throws IOException {
...
+  }
+
+  /**
+   * Creates an SSL context factory if configuration states so.
+   * @param conf the configuration
+   * @param condition the condition variable to check for SSL
+   * @return null if no ssl in config, an instance otherwise
+   * @throws IOException if getting password fails
+   */
+  static SslContextFactory createSslContextFactory(Configuration conf, MetastoreConf.ConfVars condition) throws IOException {
```
nrg4878: nit: the method name suggest that this method creates a SSL Context Factory. But it is conditional based on whether or not a config is enabled. If you do not have a better name, its fine as well.

henrib: Renamed to createSslContextFactoryIf().

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletServerBuilder.java
```diff
@@ -50,6 +53,10 @@
  * different ports.
  */
 public class ServletServerBuilder {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ServletServerBuilder.class);
+  private static final String HTTP11 = "http/1.1";
```
ayushtkn: this is unused

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java
```diff
@@ -304,13 +304,24 @@ static void loginServerPrincipal(Configuration conf) throws IOException {
...
+
+  /**
+   * Creates an SSL context factory if a configuration variable states so.
+   * @param conf the configuration
+   * @param condition the condition variable to check for SSL
+   * @return null if no ssl in config, an instance otherwise
+   * @throws IOException if getting password fails
+   */
+  static SslContextFactory createSslContextFactoryIf(Configuration conf, MetastoreConf.ConfVars condition) throws IOException {
+    final boolean useSsl = MetastoreConf.getBoolVar(conf, condition);
```
ayushtkn: this is crossing the line limit
https://github.com/apache/hive/blob/master/checkstyle/checkstyle.xml#L159-L160

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletServerBuilder.java
```diff
@@ -159,24 +166,55 @@ private Server createServer() {
...
+   * @param conf The configuration to use
+   * @return The created SslContextFactory or null if creation failed
+   */
+  private static SslContextFactory createSslContextFactory(Configuration conf) {
+    try {
+      return ServletSecurity.createSslContextFactoryIf(conf, MetastoreConf.ConfVars.HTTPSERVER_USE_HTTPS);
+    } catch (IOException e) {
+      LOGGER.error("Failed to create SSL context factory", e);
+      return null;
+    }
```
ayushtkn: Why do we need this logic. IOE is thrown when getting password from the credential provider fails. That should be fatal I believe. Here I think we are treating it same as the conf was set off

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletServerBuilder.java
```diff
@@ -159,24 +166,55 @@ private Server createServer() {
...
+   * @param port The port to bind the connector to
+   * @return The created ServerConnector
    */
   private ServerConnector createConnector(Server server, SslContextFactory sslContextFactory, int port) {
-    final ServerConnector connector = new ServerConnector(server, sslContextFactory);
+    final ServerConnector connector;
+    if (sslContextFactory == null) {
+      connector = new ServerConnector(server);
+      connector.setName(HTTP);
+      connector.setReuseAddress(true);
```
ayushtkn: this is getting called twice. It would be called below at L217 as well, so we can drop it here

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest.java
```diff
@@ -92,30 +97,38 @@ protected void stopServer(int port) throws Exception {
...
+      : HttpClients.createDefault();
+  }
 
   @Override
   protected PropertyClient createClient(Configuration conf, int sport) throws Exception {
     String path = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.PROPERTIES_SERVLET_PATH);
-    URL url = new URL("http://hive@localhost:" + sport + "/" + path + "/" + NS);
+    String scheme = MetastoreConf.getBoolVar(conf, MetastoreConf.ConfVars.HTTPSERVER_USE_HTTPS)
+            ? "https" : "http";
+    URI uri = new URI(scheme + "://hive@localhost:" + sport + "/" + path + "/" + NS);
```
ayushtkn: curious why the variable name is "sport", Sport in english means something totally different :-)

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest.java
```diff
@@ -156,16 +169,19 @@ public Map<String, String> getProperties(List<String> selection) {
 
   @Test
   public void testServletEchoA() throws Exception {
-    URL url = new URL("http://hive@localhost:" + servletPort + "/" + path + "/" + NS);
+    String path = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.PROPERTIES_SERVLET_PATH);
+    String scheme = MetastoreConf.getBoolVar(conf, MetastoreConf.ConfVars.HTTPSERVER_USE_HTTPS)
+            ? "https" : "http";
```
ayushtkn: this is done like 3 times in this class itself, can we have a util like getScheme()

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest1.java
```diff
@@ -167,4 +172,54 @@ public Map<String, String> getProperties(List<String> selection) throws IOExcept
     }
   }
 
+  @Test
+  public void testServletEchoB() throws Exception {
```
ayushtkn: can you choose a better name, unable to decode what does B mean here

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSTestBase.java
```diff
@@ -145,6 +156,78 @@ public synchronized void tearDown() throws Exception {
...
+   */
+  public static void setHttpsConf(Configuration conf) {
+    MetastoreConf.setBoolVar(conf, MetastoreConf.ConfVars.USE_SSL, true);
+    MetastoreConf.setBoolVar(conf, MetastoreConf.ConfVars.HTTPSERVER_USE_HTTPS, true);
+    MetastoreConf.setVar(conf, MetastoreConf.ConfVars.SSL_KEYSTORE_PATH,
+            STORES_DIR + File.separator + LOCALHOST_KEY_STORE_NAME);
+    MetastoreConf.setVar(conf, MetastoreConf.ConfVars.SSL_KEYSTORE_PASSWORD,
+            KEY_STORE_TRUST_STORE_PASSWORD);
+    MetastoreConf.setVar(conf, MetastoreConf.ConfVars.SSL_KEYSTORE_TYPE, "PKCS12");
+    MetastoreConf.setVar(conf, MetastoreConf.ConfVars.SSL_KEYMANAGERFACTORY_ALGORITHM, KeyManagerFactory.getDefaultAlgorithm());
```
ayushtkn: crosses the line limit I believe, can you format the code

ayushtkn (approved):
LGTM

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1895,6 +1895,10 @@ public enum ConfVars {
             "hive.metastore.httpserver.threadpool.idle", 60_000L,
             "HMS embedded HTTP server thread idle time."
     ),
+    HTTPSERVER_USE_HTTPS("hive.metastore.httpserver.use.https",
```
deniskuzZ: why do we need new config? can't we reuse `USE_SSL`?

henrib: I did not want to presume that USE_SSL would imply https for Iceberg Catalog and Properties; routing to these endpoints using https may imply more configuration on the client side (truststore).
We could use USE_SSL as default if ...use.https is not explicitly set ?

deniskuzZ: Is there a real production setup where these configs would have different values? If so, that's fine, I just thought we should either use SSL everywhere or not at all.

henrib: I honestly don't know and will remove it since you deem it unnecessary.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java
```diff
@@ -314,13 +314,25 @@ static void loginServerPrincipal(Configuration conf) throws IOException {
...
+  /**
+   * Creates an SSL context factory if a configuration variable states so.
+   * @param conf the configuration
+   * @param condition the condition variable to check for SSL
+   * @return null if no ssl in config, an instance otherwise
+   * @throws IOException if getting password fails
+   */
+  static SslContextFactory createSslContextFactoryIf(Configuration conf, MetastoreConf.ConfVars condition)
+          throws IOException {
+    final boolean useSsl = MetastoreConf.getBoolVar(conf, condition);
```
deniskuzZ: could we drop this method and use MetastoreConf.ConfVars.USE_SSL directly?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletServerBuilder.java
```diff
@@ -159,24 +165,49 @@ private Server createServer() {
...
+   * @param server The server to create the connector for
+   * @param sslContextFactory The ssl context factory to use;
+   *                          if null, the connector will be HTTP; if not null, the connector will be HTTPS
+   * @param port The port to bind the connector to
+   * @return The created ServerConnector
    */
   private ServerConnector createConnector(Server server, SslContextFactory sslContextFactory, int port) {
-    final ServerConnector connector = new ServerConnector(server, sslContextFactory);
+    final ServerConnector connector;
+    if (sslContextFactory == null) {
```
deniskuzZ: can we refactor this to
````
HttpConfiguration httpConf = new HttpConfiguration();

// Do not leak information - this logic is now in one place.
httpConf.setSendServerVersion(false);
httpConf.setSendXPoweredBy(false);

final ServerConnector connector;
if (sslContextFactory != null) {
    httpConf.setSecureScheme(HTTPS);
    httpConf.setSecurePort(port);
    httpConf.addCustomizer(new SecureRequestCustomizer());

    connector = new ServerConnector(server, sslContextFactory, new HttpConnectionFactory(httpConf));
    connector.setName(HTTPS);
} else {
    connector = new ServerConnector(server, new HttpConnectionFactory(httpConf));
    connector.setName(HTTP);
}

connector.setPort(port);
connector.setReuseAddress(true);
return connector;
````

File: standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest1.java [hunk 13]
```diff
@@ -22,25 +22,34 @@
...
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.entity.ContentType;
 import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.client.HttpClients;
+import org.apache.http.message.BasicNameValuePair;
+import org.junit.Assert;
+import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
 @Category(MetastoreUnitTest.class)
```
deniskuzZ, line 55: Not related to the PR, but if you could come up with a better name than `HMSServletTest1` - would be nice

deniskuzZ (approved):
+1, pending tests

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletServerBuilder.java
```diff
@@ -46,10 +48,13 @@
...
 
 /**
- * Helper class to ease creation of embedded Jetty serving servlets on
+ * Helper class to ease the creation of embedded Jetty serving servlets on
  * different ports.
  */
 public class ServletServerBuilder {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ServletServerBuilder.class);
+  private static final String HTTP = "http";
+  private static final String HTTPS = "https";
```
deniskuzZ: why not use `org.eclipse.jetty.util.URIUtil` ?

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest.java
```diff
@@ -26,41 +25,35 @@
...
 
+@TestMethodOrder(MethodOrderer.MethodName.class)
 @Category(MetastoreUnitTest.class)
 public class HMSServletTest extends HMSTestBase {
   String path = null;
   Server servletServer = null;
   int servletPort = -1;
-  
+
+  @Override
```
deniskuzZ, line 55: note, you can drop the @Override and just remove `super.setUp()`

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest.java
```diff
@@ -92,30 +86,34 @@ protected void stopServer(int port) throws Exception {
     }
   }
 
+  protected static String getScheme(Configuration conf) {
+    return MetastoreConf.getBoolVar(conf, MetastoreConf.ConfVars.USE_SSL) ? "https" : "http";
```
deniskuzZ: can we use constants from `org.eclipse.jetty.util.URIUtil`? 
does it have to be protected?

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest1.java
```diff
@@ -167,4 +185,90 @@ public Map<String, String> getProperties(List<String> selection) throws IOExcept
...
+        Gson gson = new GsonBuilder().create();
+        ContentType contentType = ContentType.getOrDefault(entity);
+        Charset charset = contentType.getCharset();
+        Reader reader = new InputStreamReader(entity.getContent(), charset);
+        result = (Map<String, String>) gson.fromJson(reader, Object.class);
+      }
+      Assert.assertNotNull(result);
+      Assert.assertEquals(2, result.size());
+    } finally {
+      if (response instanceof AutoCloseable) {
```
deniskuzZ, line 225: could be simplified
````
if (response instanceof AutoCloseable closeable) {
  closeable.close()
}
````
maybe create helper method
```` 
static void closeResource(Object resource) {
  if (resource instanceof AutoCloseable closeable) {
    closeable.close();
  }
}
````

File: standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest1.java [hunk 18]
```diff
@@ -57,23 +66,32 @@ public void tearDown() throws Exception {
...
+            ? HttpClients.custom().setSSLContext(sslCtxt).setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE).build()
+            : HttpClients.createDefault();
   }
 
   /**
    * A property client that uses Apache HttpClient as base.
    */
-  public static class JSonHttpClient implements HttpPropertyClient, AutoCloseable {
-    private final String uri;
+  public static class JsonHttpClient implements HttpPropertyClient, AutoCloseable {
```
deniskuzZ: inner classes are usually declared as private static

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest.java
henrib: Will do with constants; method is reused in a derived test class.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest.java
henrib: I don't follow; the super.setUp() is necessary and this setUp() extracts the servlet path.

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSServletTest1.java
deniskuzZ: not sure why Sonar suggests these: https://sonarcloud.io/project/issues?impactSeverities=HIGH&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=6002&id=apache_hive. 
As I understand it might not always be Closeable.
@henrib, could you please double-check and let's merge

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/properties/HMSJsonClientSslTest.java
```diff
@@ -92,30 +89,34 @@ protected void stopServer(int port) throws Exception {
...
+  protected static String getScheme(Configuration conf) {
+    return MetastoreConf.getBoolVar(conf, MetastoreConf.ConfVars.USE_SSL) ? HTTPS : HTTP;
+  }
 
   @Override
-  protected PropertyClient createClient(Configuration conf, int sport) throws Exception {
-    String path = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.PROPERTIES_SERVLET_PATH);
-    URL url = new URL("http://hive@localhost:" + sport + "/" + path + "/" + NS);
+  protected PropertyClient createClient(Configuration conf, int port) throws Exception {
+    String servletPath = MetastoreConf.getVar(conf, MetastoreConf.ConfVars.PROPERTIES_SERVLET_PATH);
```
deniskuzZ, line 98: minor: The local variable declaration could be replaced with a `path` field in several places, including in inherited test classes such as HMSHttpClientSslTest

PR #6006, opened 2025-07-31
HIVE-29117: Refactor deleteDir and moveToTrash

wecharyu:
We can refactor `MetaStoreFS.deleteDir()` and `FileUtils.moveToTrash()` to get a cleaner code:
1. remove `MetaStoreFS` interface, we can use `FileUtils` in replace
2. remove unused parameter `recursive`
3. remove `purge` parameter in `moveToTrash` because we should call `moveToTrash` only when `purge` is false
4. move `fs.delete()` calls from `moveToTrash()` to `deleteDir()`
5. log the root cause when `moveToTrash()` failed

1. remove redundant api implementation
2. distinguish `deleteDir` and `moveToTrash` functions, the latter only used to move path to trash.

No.

Passing existing tests.


wecharyu:
@ayushtkn @deniskuzZ could you pls take a look?

okumin:
It looks like it's been introduced by HIVE-1881, for a Facebook-specific problem.

I guess we can void the following tickets if this is the right direction.
- https://issues.apache.org/jira/browse/HIVE-12321
- https://issues.apache.org/jira/browse/HIVE-1894
- https://issues.apache.org/jira/browse/HIVE-1895

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
```diff
@@ -114,12 +92,30 @@ public static boolean moveToTrash(FileSystem fs, Path f, Configuration conf, boo
...
+  }
+
+  /**
+   * Move a particular file or directory to the trash.
+   * @param fs FileSystem to use
+   * @param f path of file or directory to move to trash.
+   * @param conf configuration object
+   * @return true if move successful
+   */
+  public static boolean moveToTrash(FileSystem fs, Path f, Configuration conf) {
```
okumin: This can be private

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
```diff
@@ -71,41 +71,19 @@ public boolean accept(Path p) {
...
-      } else {
-        result = Trash.moveToAppropriateTrash(fs, f, conf);
-        if (result) {
-          LOG.trace("Moved to trash: " + f);
-          return true;
-        }
-      }
-    } catch (IOException ioe) {
       // for whatever failure reason including that trash has lower encryption zone
       // retry with force delete
```
okumin, line 86: I guess the current logic can't reproduce this behavior, I'm not sure that it is really necessary

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
```diff
@@ -114,12 +92,30 @@ public static boolean moveToTrash(FileSystem fs, Path f, Configuration conf, boo
           .contains("Snapshot"))
         deleteReplRelatedSnapshots(fs, f);
       // retry delete after attempting to delete replication related snapshots
-      result = fs.delete(f, true);
+      fs.delete(f, true);
     }
-    if (!result) {
-      LOG.error("Failed to delete " + f);
+    return true;
```
okumin: This method is likely to always return true. Is it expected?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
```diff
@@ -114,12 +92,30 @@ public static boolean moveToTrash(FileSystem fs, Path f, Configuration conf, boo
...
+
+  /**
+   * Move a particular file or directory to the trash.
+   * @param fs FileSystem to use
+   * @param f path of file or directory to move to trash.
+   * @param conf configuration object
+   * @return true if move successful
+   */
+  public static boolean moveToTrash(FileSystem fs, Path f, Configuration conf) {
+    LOG.debug("moving  " + f + " to trash");
```
okumin: Probably, you will see warning messages from SonarQube about how to use slf4j once you rebase this branch

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
wecharyu: I think this logic is reasonable, for example if the hdfs does not enable Trash, the `moveToTrash` would return false and we could try to delete the path here.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
wecharyu: Nice catch! This method will return true or throw IOException in current implement.
It's reasonable to remove the return value.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
wecharyu: Addressed.

okumin:
I see some lint errors
https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=6006&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
okumin: I guess moveToTrasn does not return false but throws an IOException in that case. This is really a guess, based on the original code, and I've not confirmed it

wecharyu: Oh, we catch the `IOException` in `moveToTrash`, it would return false.

wecharyu:
@okumin Please take a look again, the remaining 2 sonar issues seem inaccurate, we return null in the lambda expressions instead of the finally blocks.

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
```diff
@@ -2215,9 +2215,7 @@ private boolean deletePartitionLocation(Partition partition, boolean purgeData)
       Path path = getWh().getDnsPath(new Path(location));
       try {
         do {
-          if (!getWh().deleteDir(path, true, purgeData, false)) {
-            throw new MetaException("Unable to delete partition at " + location);
```
okumin, line 2219: This changes the behavior. We may restore the behavior or validate the new behavior

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -5948,13 +5948,14 @@ public static boolean trashFiles(final FileSystem fs, final FileStatus[] statuse
...
-        result &= FileUtils.moveToTrash(fs, status.getPath(), conf, purge);
+        org.apache.hadoop.hive.metastore.utils.FileUtils.deleteDir(fs, status.getPath(), purge, conf);
       } else {
         futures.add(pool.submit(new Callable<Boolean>() {
           @Override
           public Boolean call() throws Exception {
             SessionState.setCurrentSessionState(parentSession);
-            return FileUtils.moveToTrash(fs, status.getPath(), conf, purge);
+            org.apache.hadoop.hive.metastore.utils.FileUtils.deleteDir(fs, status.getPath(), purge, conf);
+            return true;
```
okumin: This probably changes the behavior; the following part checks the list of booleans.

ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/FSRemover.java
```diff
@@ -122,9 +122,8 @@ private List<Path> removeFiles(CleanupRequest cr)
       if (needCmRecycle) {
         replChangeManager.recycle(dead, ReplChangeManager.RecycleType.MOVE, cr.isPurge());
       }
-      if (FileUtils.moveToTrash(fs, dead, conf, cr.isPurge())) {
-        deleted.add(dead);
-      }
+      FileUtils.deleteDir(fs, dead, cr.isPurge(), conf);
+      deleted.add(dead);
```
okumin: This change might affect the result of compactions

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
wecharyu: Revert to original behavior.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HMSHandler.java
```diff
@@ -1139,7 +1139,7 @@ private void dropCatalogCore(String catName, boolean ifExists)
       }
     } finally {
       if (success) {
-        wh.deleteDir(wh.getDnsPath(new Path(cat.getLocationUri())), false, false, false);
+        wh.deleteDir(wh.getDnsPath(new Path(cat.getLocationUri())), false, false);
```
okumin, line 1142: Just a note for myself and reviewers: As mentioned in the description, `HiveMetaStoreFsImpl#deleteDir` ignores the `recursive` argument, and that's why we can accept this change.

okumin:
Merged. @wecharyu Thanks for the cleanup!

PR #6009, opened 2025-08-06
HIVE-29122: Vectorization - Support IGNORE NULLS for FIRST_VALUE and …

soumyakanti3578:
…LAST_VALUE

Pass `respectNulls` to the `first_value` and `last_value` VectorPTFEvaluator classes and update `evaluateGroupBatch` methods to support `IGNORE NULLS`

Explained in detail: https://issues.apache.org/jira/browse/HIVE-29122

No

```
mvn test -pl itests/qtest -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile="vectorized_first_last_value_ignore_nulls_double.q,vectorized_first_last_value_ignore_nulls_int.q,vectorized_first_last_value_ignore_nulls_decimal.q"
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/plan/VectorPTFDesc.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/plan/VectorPTFDesc.java
@@ -137,6 +137,7 @@ public class VectorPTFDesc extends AbstractVectorDesc  {
 
   private String[] evaluatorFunctionNames;
   private boolean[] evaluatorsAreDistinct;
+  private boolean[] evaluatorsRespectNulls;
   private WindowFrameDef[] evaluatorWindowFrameDefs;
   private List<ExprNodeDesc>[] evaluatorInputExprNodeDescLists;
 
@@ -156,6 +157,7 @@ public class VectorPTFDesc extends AbstractVectorDesc  {
 
     evaluatorFunctionNames = null;
     evaluatorsAreDistinct = null;
+    evaluatorsRespectNulls = null;
     evaluatorInputExprNodeDescLists = null;
 
     orderExprNodeDescs = null;
@@ -170,7 +172,7 @@ public class VectorPTFDesc extends AbstractVectorDesc  {
 
   // We provide this public method to help EXPLAIN VECTORIZATION show the evaluator classes.
   public static VectorPTFEvaluatorBase getEvaluator(SupportedFunctionType functionType,
-      boolean isDistinct, WindowFrameDef windowFrameDef, Type[] columnVectorTypes,
+      boolean isDistinct, boolean respectNulls, WindowFrameDef windowFrameDef, Type[] columnVectorTypes,
       VectorExpression[] inputVectorExpressions, int outputColumnNum) {
 
     final boolean isRowEndCurrent = (windowFrameDef.getWindowType() == WindowType.ROWS
@@ -321,6 +323,7 @@ public class VectorPTFDesc extends AbstractVectorDesc  {
       default:
         throw new RuntimeException("Unexpected column vector type " + columnVectorType + " for " + functionType);
       }
+      evaluator.setRespectNulls(respectNulls);
       break;
     case LAST_VALUE:
       switch (columnVectorType) {
@@ -336,6 +339,7 @@ public class VectorPTFDesc extends AbstractVectorDesc  {
       default:
         throw new RuntimeException("Unexpected column vector type " + columnVectorType + " for " + functionType);
       }
+      evaluator.setRespectNulls(respectNulls);
       break;
     case COUNT:
       if (inputVectorExpression == null) {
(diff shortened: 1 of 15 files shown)


kasakrisz:
@soumyakanti3578 
How cases handled when the rows in the window does not fit into one batch?
* Case of `first_value respect nulls` is easy because we take the first row in the first batch.
* Case of `last_value respect nulls` is more challenging because we need the value from the last batch. However the last row in the window is not always the last row in the batch. I mean it is not `batch.size - 1`. Is this true?
* Case of `first_value ignore nulls`. What if all the values are null in the first batch? We need to check the follow up batches.
* Case of `last_value ignore nulls`. As you already implemented in this patch iterate back from `batch.size - 1` until we find a non-null value. This is good when we have one batch but it is more challenging when we have follow-up batches.

soumyakanti3578:
@kasakrisz 
I debugged a bit to understand the workflow. To simulate the scenario where rows with the same key doesn't fit into 1 single batch, we can run the tests with `SET hive.vectorized.testing.reducer.batch.size=2;` With this I have noticed that `VectorPTFGroupBatches.bufferedBatches` contains batches whose size is at most 2. 

For the data in `vectorized_first_last_value_ignore_nulls_int.q`, for key 1 and 2, I have seen batches like:
```
Batch 1:   [[1, null], [1, null]]
Batch 2:   [1, null]
Batch 3:   [[2, null], [2, 20]]
Batch 4:   [2, null]
```
Let's look at `last_value ignore nulls` computation.
First we will compute last value for Range with key = 1, which consists of Batch 1 and 2. A new evaluator will be created for `VectorPTFEvaluatorLongLastValue`, which we will keep reusing.

First, we will call `VectorPTFEvaluatorLongLastValue.evaluateGroupBatch(batch 1)` will be called, and we will have to check both vector elements for non null values. Since they are null, we will have `isGroupResultNull = true` and `lastValue = null`.

Then, we will call `VectorPTFEvaluatorLongLastValue.evaluateGroupBatch(batch 2)` on the same evaluator object, but the  value of `isGroupResultNull` and `lastValue` doesn't change.

Next we will call `VectorPTFEvaluatorLongLastValue.evaluateGroupBatch(batch 3)` on the same evaluator, and this time, the values will be `isGroupResultNull = false` and `lastValue = 20`.

And then, we will call `VectorPTFEvaluatorLongLastValue.evaluateGroupBatch(batch 4)` on the same evaluator, and the values will be `isGroupResultNull = true` and `lastValue = 20`. Note that `lastValue` remains unchanged.

Now if we didn't have other rows in the table, we would get out of the `while` loop in `VectorPTFGroupBatches.BlockIterator.run` and call `getEvaluatorResult(evaluator)` as seen below:
```
    Object run(VectorPTFEvaluatorBase evaluator) throws HiveException {
      runEvaluator(evaluator, range);

      while (hasNext()) {
        nextBlock();
        runEvaluator(evaluator, range);
      }

      return getEvaluatorResult(evaluator);
    }

    Object getEvaluatorResult(VectorPTFEvaluatorBase evaluator) {
      return evaluator.isGroupResultNull() ? null : evaluator.getGroupResult();
    }
```

`VectorPTFEvaluatorLongLastValue.isGroupResultNull()` was updated in this PR as:
```
  public boolean isGroupResultNull() {
    return isGroupResultNull && doesRespectNulls();
  }
```
This returns `false` even though `isGroupResultNull = true` at the end because `doesRespectNulls()` returns false. So finally `evaluator.getGroupResult()` is called, which returns `lastValue = 20`.

If we were running `last value respect nulls`, it would have returned `null` because `isGroupResultNull()` would have returned `true`.

Similarly `first value ignore nulls` can be explained too. One important difference is that `VectorPTFEvaluatorLongFirstValue` has the boolean variable `haveFirstValue`. Once we find a non null value, this variable is set to true and for all batches after that we directly write the value into the output vector.

Hopefully this makes sense, but please let me know if I am missing something. If needed, I can add `SET hive.vectorized.testing.reducer.batch.size=2;` to all three test files too - they are all passing on my local machine.

deniskuzZ:
> And then, we will call VectorPTFEvaluatorLongLastValue.evaluateGroupBatch(batch 4) on the same evaluator, and the values will be isGroupResultNull = true and lastValue = 20. Note that lastValue remains unchanged.

shouldn't the `isGroupResultNull` stay false?

kasakrisz:
@soumyakanti3578 
Thanks for the detailed explanation. Overall the patch seems to be ok to me.

One note for the testing part: IMHO it worth adding test cases for both:
* batch size is big enough to load all the records into one batch from the table
* or batch size is too small and the windows are loaded into more than one batch.

ql/src/java/org/apache/hadoop/hive/ql/plan/VectorPTFDesc.java
```diff
@@ -321,6 +323,7 @@ public static VectorPTFEvaluatorBase getEvaluator(SupportedFunctionType function
       default:
         throw new RuntimeException("Unexpected column vector type " + columnVectorType + " for " + functionType);
       }
+      evaluator.setRespectNulls(respectNulls);
```
kasakrisz: Could you please move this call to the end of the method before returning the `evaluator` instance? In that case this would be called only once int the entire method.

soumyakanti3578: Done!

soumyakanti3578:
@deniskuzZ 
Yes, but it didn't affect the results. But in the latest commit I have updated the initial value to `isGroupResultNull = lastValue == null;` which should take care of the inconsistency.

I also changed 
```
  public boolean isGroupResultNull() {
    return isGroupResultNull && doesRespectNulls();
  }
```
back to just `return isGroupResultNull`, as we don't need to do this if we are setting the value of `isGroupResultNull` correctly, as described above.

@kasakrisz 
I have added more tests in all three files with `set hive.vectorized.testing.reducer.batch.size=2;`

Please review again! Thanks!

kasakrisz (approved):
+1 pending tests

PR #6010, opened 2025-08-06
HIVE-29123: Extend ProtobufInputFormat to handle EOFException for par…

rtrivedi12:
…tially written proto files

This PR introduces a more lenient version of ProtobufInputFormat that ignores EOFException when reading partially written proto files.
ProtobufMessageInputFormat skips zero-byte files by creating <SafeRecordReader>, which ignores EOFException. 

Abrupt AM termination or OOM failures can result in partially written protobuf files. When reading data from the corresponding external table, these incomplete files may trigger an EOFException, causing the entire query to fail.

To improve resilience, it would be better to gracefully skip unreadable or corrupted proto files and continue reading the remaining valid data, instead of failing the entire query.

Yes, this would silently skip reading partially written proto files instead of failing with EOFException.

Manually tested with a sample partially written dag data file.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/io/protobuf/LenientProtobufMessageInputFormat.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/io/protobuf/LenientProtobufMessageInputFormat.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.hive.ql.io.protobuf;
+
+import com.google.protobuf.MessageLite;
+import org.apache.hadoop.mapred.InputSplit;
+import org.apache.hadoop.mapred.JobConf;
+import org.apache.hadoop.mapred.RecordReader;
+import org.apache.hadoop.mapred.Reporter;
+import org.apache.tez.dag.history.logging.proto.ProtoMessageWritable;
+
+import java.io.EOFException;
+import java.io.IOException;
+
+/**
+ * Lenient variant of ProtobufMessageInputFormat that tolerates read failures
+ * in SequenceFileRecordReader by using a forgiving subclass.
+ */
+public class LenientProtobufMessageInputFormat<K, V extends MessageLite>
+    extends ProtobufMessageInputFormat<K, V> {
+
+  private static final String PROTO_CLASS = "proto.class";
+
+  @Override
+  public RecordReader<K, ProtoMessageWritable<V>> getRecordReader(InputSplit split, JobConf job, Reporter reporter)
+      throws IOException {
+    final RecordReader<K, ProtoMessageWritable<V>> reader = super.getRecordReader(split, job, reporter);
+    return new SafeRecordReader<>(reader);
+  }
+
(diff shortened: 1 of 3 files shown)


deniskuzZ:
is that trying to solve same issue as https://github.com/apache/hive/pull/5983?
cc @Aggarwal-Raghav

Aggarwal-Raghav:
> is that trying to solve same issue as #5983? cc @Aggarwal-Raghav

Based on the description provided in the PR, the issue appears to be different compared to #5983. @rtrivedi12, can you please provide steps for how you created "partially written dag data file." which you used for testing?

rtrivedi12:
> Based on the description provided in the PR, the issue appears to be different compared to #5983. @rtrivedi12, can you please provide steps for how you created "partially written dag data file." which you used for testing?

This `EOFException` happens much before , in low-level I/O when SequenceFileReader finds end-of-file before completing the read. And I think [#5983] is about `InvalidProtocolBufferException` that happens when hive successfully reads a record, but the record itself has bad protobuf encoding.

This partially written dag_data file was created by executing a long- running query and killing the yarn application in between  which leaves a partially written dag_data file by Tez `ProtoHistoryLoggingService`.

ql/src/java/org/apache/hadoop/hive/ql/io/protobuf/LenientProtobufMessageInputFormat.java
```diff
@@ -0,0 +1,91 @@
...
+    public float getProgress() throws IOException {
+      return mainReader != null ? mainReader.getProgress() : 1.0f;
+    }
+
+    @Override
+    public boolean next(K arg0, ProtoMessageWritable<V> arg1) throws IOException {
+      try {
+        return mainReader != null ? mainReader.next(arg0, arg1) : false;
+      } catch (EOFException e) {
+        return false;
```
abstractdog, line 99: Does this happen on the unhappy path only? If so, there should be a LOG.warn before returning. The log message might show e.g. getPos() or anything useful about the current reading context.

ql/src/java/org/apache/hadoop/hive/ql/io/protobuf/LenientProtobufMessageInputFormat.java
```diff
@@ -0,0 +1,91 @@
...
+import java.io.IOException;
+
+/**
+ * Lenient variant of ProtobufMessageInputFormat that tolerates read failures
+ * in SequenceFileRecordReader by using a forgiving subclass.
+ */
+public class LenientProtobufMessageInputFormat<K, V extends MessageLite>
+    extends ProtobufMessageInputFormat<K, V> {
+
+  private static final String PROTO_CLASS = "proto.class";
```
abstractdog: PROTO_CLASS looks unused

abstractdog:
I believe the relationship between this PR and this codepath below should be clarified:
https://github.com/apache/hive/blob/df98c2e2f4e2370c603f8873f006ee8728afa8a6/ql/src/java/org/apache/hadoop/hive/ql/io/protobuf/ProtobufMessageInputFormat.java#L67-L75

If this PR calls the new RecordReader a SafeRecordReader, then maybe we need to change it to distinguish between the two "safe" things:
1. here we can call it IgnoreEOFProtoMessageRecordReader
2. create class javadoc comment on IgnoreEOFProtoMessageRecordReader to explicitly say that this record reader simply returns false from the next method when it reaches EOF during read (which is not the same what the parent does)

ql/src/java/org/apache/hadoop/hive/ql/io/protobuf/LenientProtobufMessageInputFormat.java
rtrivedi12: Thanks for the review comments @abstractdog! This happens only when something goes wrong (partially written/corrupt file). Will add a meaningful LOG.warn message.

rtrivedi12:
> I believe the relationship between this PR and this codepath below should be clarified:
> 
> https://github.com/apache/hive/blob/df98c2e2f4e2370c603f8873f006ee8728afa8a6/ql/src/java/org/apache/hadoop/hive/ql/io/protobuf/ProtobufMessageInputFormat.java#L67-L75
> 
> If this PR calls the new RecordReader a SafeRecordReader, then maybe we need to change it to distinguish between the two "safe" things:
> 
> 1. here we can call it IgnoreEOFProtoMessageRecordReader
> 2. create class javadoc comment on IgnoreEOFProtoMessageRecordReader to explicitly say that this record reader simply returns false from the next method when it reaches EOF during read (which is not the same what the parent does)

Good point, missed it! It may be confusing with Parent SafeRecordReader; I will rename and clarify in the Java doc.

abstractdog (approved):
+1 LGTM

PR #6011, opened 2025-08-07
HIVE-29124: Avoid committing files when a task is aborted even though some source has completed.

zhengchenyu:

Two change:

* Does not allow abort to be set from true to false.
* dummyOps also aborts.  

I found that when the task is almost completed (more precisely, when the source has been processed), but not closed, if an exception is thrown at this time, it may cause the file to be committed incorrectly.

Look at the below code. abort may be set from true to false. It is not reasonable. The correct logic is that as long as abort is set to true at one place, abort should always be true, then do not commit.
https://github.com/apache/hive/blob/9b07c5c7136863ae1eb469e7a3c11357299d2ea1/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java#L349

When I tried to reproduce this bug, I found that only [dummyOp.close(abort)](https://github.com/apache/hive/blob/9b07c5c7136863ae1eb469e7a3c11357299d2ea1/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java#L369) caused the problem. I initially thought that the problem would occur at [reducer.close(abort)](https://github.com/apache/hive/blob/9b07c5c7136863ae1eb469e7a3c11357299d2ea1/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java#L356) because the reduce op's abort was set to true in the abort. However, dummyOps was not properly aborted. Here, dummyOps should also be aborted. Therefore, the issue only occurs when dummyOps is used, such as in mapjoin.

No.

Use real tasks. 

> Note: This is a low-probability issue, and I added sleep code at a specific location to increase the probability of this bug.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
@@ -307,16 +307,11 @@ public class ReduceRecordProcessor extends RecordProcessor {
     }
     if (mergeWorkList != null) {
       for (BaseWork redWork : mergeWorkList) {
-        ((ReduceWork) redWork).getReducer().abort();
+        redWork.abort();
       }
     }
     if (reduceWork != null) {
-      List<HashTableDummyOperator> dummyOps = reduceWork.getDummyOps();
-      if (dummyOps != null) {
-        for (Operator<?> dummyOp : dummyOps) {
-          dummyOp.abort();
-        }
-      }
+      reduceWork.abort();
     }
   }
 
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordSource.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordSource.java
@@ -520,6 +520,14 @@ public class ReduceRecordSource implements RecordSource {
     }
   }
 
+
+  /**
+   * Closes resources and returns whether the records were successfully processed.
+   * @return boolean indicating the success status:
+   * - true: All data has been processed successfully without exceptions.
+   * - false: Exceptions were encountered during data processing.
+   * @throws Exception unexpected errors occur during closing.
+   */
   boolean close() throws Exception {
     try {
       if (handleGroupKey && groupKey != null) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/plan/BaseWork.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/plan/BaseWork.java
@@ -545,4 +545,12 @@ public abstract class BaseWork extends AbstractOperatorDesc {
           String workName, RuntimeValuesInfo runtimeValuesInfo) {
     inputSourceToRuntimeValuesInfo.put(workName, runtimeValuesInfo);
   }
+
+  public void abort() {
+    if (dummyOps != null) {
+      for (Operator<?> dummyOp : dummyOps) {
+        dummyOp.abort();
+      }
+    }
+  }
 }
--- a/ql/src/java/org/apache/hadoop/hive/ql/plan/ReduceWork.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/plan/ReduceWork.java
@@ -381,4 +381,12 @@ public class ReduceWork extends BaseWork {
   public TezEdgeProperty getEdgePropRef() {
     return edgeProp;
   }
+
+  @Override
+  public void abort() {
+    super.abort();
+    if (reducer != null) {
+      reducer.abort();
+    }
+  }
 }


zhengchenyu:
@kgyrtkirk @abstractdog @zabetak Can you please review this PR?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
```diff
@@ -345,9 +353,9 @@ void close() {
     try {
       if (isAborted()) {
         for (ReduceRecordSource rs : sources) {
-          if (!rs.close()) {
-            setAborted(false); // Preserving the old logic. Hmm...
-            break;
+          boolean abort = rs.close();
+          if (!isAborted() && abort) {
```
deniskuzZ: isAborted() is always true here. i don't understand what are you trying to accomplish here: set true -> true?

zhengchenyu: @deniskuzZ Thanks for your review! I have fix it. Sorry for migrating code incorrectly.

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
```diff
@@ -343,12 +351,10 @@ void close() {
     }
 
     try {
-      if (isAborted()) {
-        for (ReduceRecordSource rs : sources) {
-          if (!rs.close()) {
-            setAborted(false); // Preserving the old logic. Hmm...
-            break;
-          }
+      for (ReduceRecordSource rs : sources) {
```
deniskuzZ, line 355: in my opinion it's not logical to return true in `rs.close()` when you need to abort. true usually means success
````
boolean close() throws Exception {
 ...
  return !abort;
}
````
and then
````
boolean abort = isAborted();
for (ReduceRecordSource rs: sources) {
  abort |= !rs.close();
}
if (abort) {
  setAborted(true);
}
````

File: ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java [hunk 3]
```diff
@@ -305,6 +305,14 @@ public void abort() {
     } else {
       LOG.info("reducer not setup yet. abort not being forwarded");
     }
+    if (reduceWork != null) {
```
deniskuzZ: do we need to abort `mergeWorkList` as well? 
i am not sure we need this here, won't we release the resources in close()?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
zhengchenyu: I agree with you! `return !abort;` means success, it is more reasonable. I will fix it!

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
zhengchenyu: @deniskuzZ
Here just to set dummyop's abort to true.
In my bug scenario, abort was incorrectly set from true to false. Therefore, when dummyOp.close(abort) was called, it would cause unexpected commit file. If dummyOp's abort was set to true here, it would not commit.

> Note: If `for (ReduceRecordSource rs: sources)` is handled correctly, there is no problem if it is not modified here. I just think that when abort is set to true in one place, it means that this task attempt is useless, so all ops should be set to true.

zhengchenyu: @deniskuzZ Sorry I misread it. You were referring to mergeWorkList. Personally, I think abort would make sense. Because I think as long as abort is set to true, any logic in op.close where abort is false should not be executed. What do you think?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
abstractdog: returning a boolean from close is confusing anyway without a proper method javadoc on ReduceRecordSource.close(), please create one
as this return value is used in ReduceRecordProcessor only, I'm OK with changing the behavior, but let's clarify it with a javadoc comment

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
```diff
@@ -305,6 +305,19 @@ public void abort() {
...
+      }
+    }
+    if (reduceWork != null) {
+      List<HashTableDummyOperator> dummyOps = reduceWork.getDummyOps();
+      if (dummyOps != null) {
+        for (Operator<?> dummyOp : dummyOps) {
+          dummyOp.abort();
+        }
+      }
+    }
```
abstractdog: this implementation doesn't belong here, ReduceRecordProcessor is already overloaded, what about introducing reduceWork.abort() and do the rest there

abstractdog:
overall looks good to me, as it's in line with the description of [HIVE-14042](https://issues.apache.org/jira/browse/HIVE-14042) (just linked in jira), more specifically `ReduceRecordProcessor: dummyOp handling` so this issue has been suspected already
just left a few comments @zhengchenyu

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
```diff
@@ -305,6 +305,19 @@ public void abort() {
     } else {
       LOG.info("reducer not setup yet. abort not being forwarded");
     }
+    if (mergeWorkList != null) {
+      for (BaseWork redWork : mergeWorkList) {
+        ((ReduceWork) redWork).getReducer().abort();
```
abstractdog: I mentioned introducing a reduceWork.abort(), what about reusing it here too?
so instead of 
```
((ReduceWork) redWork).getReducer().abort();
```
maybe you can do:
```
((ReduceWork) redWork).abort();
```

so ReduceWork.abort() could take care of:
1. aborting the Operator (which is returned by getReducer)
2. aborting dummy ops

this might cover an uncovered case (not sure if it's an issue at all), which is dummyOps field of ReduceWork instances in this mergeWorkList

zhengchenyu: This code is indeed better, I have fix it.

abstractdog (approved):
LGTM pending tests

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/ReduceRecordProcessor.java
deniskuzZ: yep

PR #6012, opened 2025-08-07

Neer393:
Added new HMS API for get table objects instead of name by filter so that in IcebergHouseKeeperService.java, we can reduce the number of msc calls to one and get direct table objects

No it just adds a new method to get tables as object by filter

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/HiveSplitGenerator.java
@@ -241,7 +241,7 @@ public class HiveSplitGenerator extends InputInitializer {
         mrSplit.writeTo(out);
       }
       LOG.debug("Split #{} event to output path: {} written in {} ms", count, filePath,
-          fileWriteStarted - Time.monotonicNow());
+          Time.monotonicNow() - fileWriteStarted);
     }
 
     Path getSerializedFilePath(int index) {


standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -117,6 +118,33 @@ public List<TableName> getTables() throws Exception {
...
+      Database database = client.getDatabase(catalogName, db);
+      if (MetaStoreUtils.checkIfDbNeedsToBeSkipped(database)) {
+        LOG.debug("Skipping table under database: {}", db);
+        continue;
+      }
+      if (MetaStoreUtils.isDbBeingPlannedFailedOver(database)) {
+        LOG.info("Skipping table that belongs to database {} being failed over.", db);
+        continue;
+      }
+      List<Table> tables = client.listTablesByFilter(catalogName, db, tableFilter, -1);
```
vikramahuja1001: Can you implement Batching in it? 
client.listTablesByFilter(catalogName, db, tableFilter, -1) can potentially return a very large object if the db has large number of tables.
Please refer to TableIterable, PartitionIterable classes and their uses to get an idea of how hive uses batched calls to fetch from the metastore.

Neer393: Sure. I will try to implement it. Thanks for the reference @vikramahuja1001 👍

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
```diff
@@ -128,7 +128,7 @@ public Set<CompactionInfo> findPotentialCompactions(long lastChecked, ShowCompac
 
   private List<org.apache.hadoop.hive.common.TableName> getTables() {
     try {
-      return IcebergTableUtil.getTableFetcher(client, null, "*", null).getTables();
+      return IcebergTableUtil.getTableFetcher(client, null, "*", null).getTableNames();
```
deniskuzZ: why don't you reuse the table object and avoid extra call `IcebergTableUtil.getTable` here?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -50,7 +47,7 @@ public class IcebergHouseKeeperService implements MetastoreTaskThread {
   private ExecutorService deleteExecutorService = null;
 
   // table cache to avoid making repeated requests for the same Iceberg tables more than once per day
-  private final Cache<TableName, Table> tableCache = Caffeine.newBuilder()
+  private final Cache<org.apache.hadoop.hive.metastore.api.Table, Table> tableCache = Caffeine.newBuilder()
```
deniskuzZ: why are you changing the iceberg table cache?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
Neer393: I don't have access to the table object here.
Plus this was already implemented this way I just renamed ```getTables()``` to ```getTableNames()``` because I had changed the definition for it

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
Neer393: This change is required as the main purpose of this JIRA was to reduce the number of msc calls. If you look at the earlier implementation, you will see the following definition for ```getIcebergTable()```

```
private Table getIcebergTable(TableName tableName, IMetaStoreClient msc) {
    return tableCache.get(tableName, key -> {
      LOG.debug("Getting iceberg table from metastore as it's not present in table cache: {}", tableName);
      GetTableRequest request = new GetTableRequest(tableName.getDb(), tableName.getTable());
      try {
        return IcebergTableUtil.getTable(conf, msc.getTable(request));
      } catch (TException e) {
      }
    });
  }
```

Here we looked in the cache for the table name and if not found, we created the table object using ```msc.getTable(request)``` .
Hence keeping a cache mapping of tablename and IcebergTableObject was not feasible as this would not reduce the msc calls.

The newer implementation makes the cache mapping to HMSTableAPIObject and IcebergTableObject and hence even when the IcebergTableObject is not in cache, we can directly create it using the HMSTableAPIObject and there is no need for msc calls again.

```
private Table getIcebergTable(org.apache.hadoop.hive.metastore.api.Table table) {
    return tableCache.get(table, key -> {
      LOG.debug("Getting iceberg table from metastore as it's not present in table cache: {}", table.getTableName());
      return IcebergTableUtil.getTable(conf, table);
    });
  }
```

**Hence this cache mapping change is necessary. Plus this is only consumed by this file only and there are no other consumers to this so this won't break anything**

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/ThriftHiveMetaStoreClient.java
```diff
@@ -2162,6 +2162,16 @@ public List<String> listTableNamesByFilter(String catName, String dbName, String
         isClientFilterEnabled, filterHook, catName, dbName, tableNames);
   }
 
+  @Override
+  public List<Table> listTablesByFilter(String catName, String dbName, String filter,
+      int maxTables) throws TException {
+    List<Table> tables =
+        client.get_tables_by_filter(prependCatalogToDbName(catName, dbName, conf), filter,
```
vikramahuja1001: Instead of adding a new Get_Tables_by_filter Api along with all this thrift code, we can reuse the existing get_table_names_by_filter api and then can use the batched implementation of TableIterable to then fetch the required table objects by names to achieve the same result. No need to expose a new API here

Neer393: But that would completely defy the point of this JIRA to reduce the number of msc calls, won't it ?

Neer393: Also I don't think batching can be applied here as batching requires a list of table names already for which you just make batched calls for ```getTableObjectsByName()``` .But here in we don't have table names so that would also not work

vikramahuja1001: By reducing it to just one HMS call we risk thrift memory issues in case the number of tables are in 1000s. Instead of 1 call we can break it down into batches so that overall calls to HMS are reduced from 1000s to say 40 or 50. That is what batching is meant to achieve.

Names can be achieved through already implemented listTableNamesByFilter API and can be sent to TableIterable object

Neer393: Yeah makes sense. 
Let me implement this

PR #6013, opened 2025-08-07
HIVE-29116 : Create a DDL for setting hive default partition name at the table level

vikramahuja1001:
1. Expose a DDL to set default partition name at table level
2. Restrict set hive.exec.default.partition at runtime
3. use table property to check the default partition along with hive.exec.default.partition at cluster level

For more details refer: https://issues.apache.org/jira/browse/HIVE-29116

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/common/FileUtils.java
+++ b/common/src/java/org/apache/hadoop/hive/common/FileUtils.java
@@ -160,9 +160,8 @@ public final class FileUtils {
     // prevent instantiation
   }
 
-  public static String makePartName(List<String> partCols, List<String> vals, Map<String, String> tableParams,
-      Configuration conf) {
-    return makePartName(partCols, vals, null, tableParams, conf);
+  public static String makePartName(List<String> partCols, List<String> vals, String defaultPartitionName) {
+    return makePartName(partCols, vals, null, defaultPartitionName);
   }
 
   /**
@@ -174,15 +173,15 @@ public final class FileUtils {
    * @return An escaped, valid partition name.
    */
   public static String makePartName(List<String> partCols, List<String> vals,
-      String defaultStr, Map<String, String> tableParams, Configuration conf) {
+      String defaultStr, String defaultPartitionName) {
     StringBuilder name = new StringBuilder();
     for (int i = 0; i < partCols.size(); i++) {
       if (i > 0) {
         name.append(Path.SEPARATOR);
       }
-      name.append(escapePathName((partCols.get(i)).toLowerCase(), defaultStr, tableParams, conf));
+      name.append(escapePathName((partCols.get(i)).toLowerCase(), defaultStr, defaultPartitionName));
       name.append('=');
-      name.append(escapePathName(vals.get(i), defaultStr, tableParams, conf));
+      name.append(escapePathName(vals.get(i), defaultStr, defaultPartitionName));
     }
     return name.toString();
   }
@@ -195,9 +194,9 @@ public final class FileUtils {
    * @return
    */
   public static String makeDefaultListBucketingDirName(List<String> skewedCols,
-      String name, Map<String, String> tableParams, Configuration conf) {
+      String name, String defaultPartitionName) {
     String lbDirName;
-    String defaultDir = FileUtils.escapePathName(name, tableParams, conf);
+    String defaultDir = FileUtils.escapePathName(name, defaultPartitionName);
     StringBuilder defaultDirPath = new StringBuilder();
     for (int i = 0; i < skewedCols.size(); i++) {
       if (i > 0) {
@@ -215,16 +214,15 @@ public final class FileUtils {
    * @param vals The skewed values
    * @return An escaped, valid list bucketing directory name.
    */
-  public static String makeListBucketingDirName(List<String> lbCols, List<String> vals, Map<String, String> tableParams,
-      Configuration conf) {
+  public static String makeListBucketingDirName(List<String> lbCols, List<String> vals, String defaultPartitionName) {
     StringBuilder name = new StringBuilder();
     for (int i = 0; i < lbCols.size(); i++) {
       if (i > 0) {
         name.append(Path.SEPARATOR);
       }
-      name.append(escapePathName((lbCols.get(i)).toLowerCase(), tableParams, conf));
+      name.append(escapePathName((lbCols.get(i)).toLowerCase(), defaultPartitionName));
       name.append('=');
-      name.append(escapePathName(vals.get(i), tableParams, conf));
+      name.append(escapePathName(vals.get(i), defaultPartitionName));
     }
     return name.toString();
   }
@@ -276,8 +274,8 @@ public final class FileUtils {
     return c < charToEscape.size() && charToEscape.get(c);
   }
 
-  public static String escapePathName(String path, Map<String, String> tableParams, Configuration conf) {
-    return escapePathName(path, null, tableParams, conf);
+  public static String escapePathName(String path,  String defaultPartitionName) {
+    return escapePathName(path, null, defaultPartitionName);
   }
 
   /**
@@ -287,8 +285,7 @@ public final class FileUtils {
    *          The default name for the path, if the given path is empty or null.
    * @return An escaped path name.
    */
-  public static String escapePathName(String path, String defaultPath, Map<String, String> tableParams,
-      Configuration conf) {
+  public static String escapePathName(String path, String defaultPath, String defaultPartitionName) {
 
     // __HIVE_DEFAULT_NULL__ is the system default value for null and empty string.
     // TODO: we should allow user to specify default partition or HDFS file location.
@@ -296,7 +293,7 @@ public final class FileUtils {
       if (defaultPath == null) {
         //previously, when path is empty or null and no default path is specified,
         // __HIVE_DEFAULT_PARTITION__ was the return value for escapePathName
-        return MetaStoreUtils.getDefaultPartitionName(tableParams, conf);
+        return defaultPartitionName;
       } else {
         return defaultPath;
       }
--- a/hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/FileOutputCommitterContainer.java
+++ b/hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/FileOutputCommitterContainer.java
@@ -56,6 +56,7 @@ import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
+import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
 import org.apache.hadoop.hive.ql.ddl.table.partition.PartitionUtils;
 import org.apache.hadoop.hive.ql.metadata.HiveStorageHandler;
 import org.apache.hadoop.hive.ql.metadata.Table;
@@ -396,8 +397,8 @@ class FileOutputCommitterContainer extends OutputCommitterContainer {
           HdfsUtils.setFullFileStatus(conf, status, status.getFileStatus().getGroup(), fs,
               partPath, false);
         }
-        partPath = constructPartialPartPath(partPath, partKey.getName().toLowerCase(), partKVs, table.getParameters(),
-            conf);
+        partPath = constructPartialPartPath(partPath, partKey.getName().toLowerCase(), partKVs,
+            MetaStoreUtils.getDefaultPartitionName(table.getParameters(), conf));
       }
     }
 
@@ -408,7 +409,8 @@ class FileOutputCommitterContainer extends OutputCommitterContainer {
 
     // Set the location in the StorageDescriptor
     if (dynamicPartitioningUsed) {
-      String dynamicPartitionDestination = getFinalDynamicPartitionDestination(table, partKVs, jobInfo, conf);
+      String dynamicPartitionDestination = getFinalDynamicPartitionDestination(table, partKVs, jobInfo,
+          MetaStoreUtils.getDefaultPartitionName(table.getParameters(), conf));
       if (harProcessor.isEnabled()) {
         harProcessor.exec(context, partition, partPath);
         partition.getSd().setLocation(
@@ -423,14 +425,13 @@ class FileOutputCommitterContainer extends OutputCommitterContainer {
   }
 
   private String getFinalDynamicPartitionDestination(Table table, Map<String, String> partKVs,
-      OutputJobInfo jobInfo, Configuration conf) {
+      OutputJobInfo jobInfo, String defaultPartitionName) {
     Path partPath = new Path(table.getTTable().getSd().getLocation());
     if (!customDynamicLocationUsed) {
       // file:///tmp/hcat_junit_warehouse/employee/_DYN0.7770480401313761/emp_country=IN/emp_state=KA  ->
       // file:///tmp/hcat_junit_warehouse/employee/emp_country=IN/emp_state=KA
       for (FieldSchema partKey : table.getPartitionKeys()) {
-        partPath = constructPartialPartPath(partPath, partKey.getName().toLowerCase(), partKVs, table.getParameters(),
-            conf);
+        partPath = constructPartialPartPath(partPath, partKey.getName().toLowerCase(), partKVs, defaultPartitionName);
       }
 
       return partPath.toString();
@@ -441,7 +442,7 @@ class FileOutputCommitterContainer extends OutputCommitterContainer {
         partPath = new Path(partPath, jobInfo.getCustomDynamicRoot());
       }
       return new Path(partPath, HCatFileUtil.resolveCustomPath(jobInfo, partKVs, false,
-          table.getParameters(), conf)).toString();
+          defaultPartitionName)).toString();
     }
   }
 
@@ -458,11 +459,11 @@ class FileOutputCommitterContainer extends OutputCommitterContainer {
   }
 
   private Path constructPartialPartPath(Path partialPath, String partKey, Map<String, String> partKVs,
-      Map<String, String> tableParams, Configuration conf) {
+      String defaultPartitionName) {
 
-    StringBuilder sb = new StringBuilder(FileUtils.escapePathName(partKey, tableParams, conf));
+    StringBuilder sb = new StringBuilder(FileUtils.escapePathName(partKey, defaultPartitionName));
     sb.append("=");
-    sb.append(FileUtils.escapePathName(partKVs.get(partKey), tableParams, conf));
+    sb.append(FileUtils.escapePathName(partKVs.get(partKey), defaultPartitionName));
     return new Path(partialPath, sb.toString());
   }
 
@@ -996,7 +997,8 @@ class FileOutputCommitterContainer extends OutputCommitterContainer {
     // final destination of each partition and move its output.
     for (Entry<String, Map<String, String>> entry : partitionsDiscoveredByPath.entrySet()) {
       Path src = new Path(entry.getKey());
-      Path destPath = new Path(getFinalDynamicPartitionDestination(table, entry.getValue(), jobInfo, conf));
+      Path destPath = new Path(getFinalDynamicPartitionDestination(table, entry.getValue(), jobInfo,
+          MetaStoreUtils.getDefaultPartitionName(table.getParameters(), conf)));
       moveTaskOutputs(conf, src, src, destPath, true);
     }
     // delete the parent temp directory of all custom dynamic partitions
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
@@ -893,7 +893,8 @@ public class HiveIcebergStorageHandler extends DefaultStorageHandler implements 
     Table table = IcebergTableUtil.getTable(conf, tableDesc.getProperties());
 
     DynamicPartitionCtx dpCtx = new DynamicPartitionCtx(Maps.newLinkedHashMap(),
-        hiveConf.getIntVar(ConfVars.DYNAMIC_PARTITION_MAX_PARTS_PER_NODE), hmsTable.getParameters(), hiveConf);
+        hiveConf.getIntVar(ConfVars.DYNAMIC_PARTITION_MAX_PARTS_PER_NODE),
+        MetaStoreUtils.getDefaultPartitionName(hmsTable.getParameters(), hiveConf));
 
     if (table.spec().isPartitioned() &&
           hiveConf.getIntVar(ConfVars.HIVE_OPT_SORT_DYNAMIC_PARTITION_THRESHOLD) >= 0) {
@@ -2012,7 +2013,8 @@ public class HiveIcebergStorageHandler extends DefaultStorageHandler implements 
       Map<String, String> partitionSpec, RewritePolicy policy) throws SemanticException {
     validatePartSpec(table, partitionSpec, policy);
     try {
-      String partName = Warehouse.makePartName(partitionSpec, false, table.getParameters(), conf);
+      String partName = Warehouse.makePartName(partitionSpec, false,
+          MetaStoreUtils.getDefaultPartitionName(table.getParameters(), conf));
       return new DummyPartition(table, partName, partitionSpec);
     } catch (MetaException e) {
       throw new SemanticException("Unable to construct name for dummy partition due to: ", e);
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
@@ -503,8 +503,8 @@ public class SessionHiveMetaStoreClient extends MetaStoreClientWrapper {
         List<String> result = new ArrayList<>();
         int lastIndex = (maxParts < 0 || maxParts > partitions.size()) ? partitions.size() : maxParts;
         for (int i = 0; i < lastIndex; i++) {
-          result.add(makePartName(tmpTable.getPartitionKeys(), partitions.get(i).getValues(), tmpTable.getParameters(),
-              conf));
+          result.add(makePartName(tmpTable.getPartitionKeys(), partitions.get(i).getValues(),
+              MetaStoreUtils.getDefaultPartitionName(tmpTable.getParameters(), conf)));
         }
         Collections.sort(result);
         return result;
@@ -542,8 +542,8 @@ public class SessionHiveMetaStoreClient extends MetaStoreClientWrapper {
 
         List<String> result = new ArrayList<>();
         for (int i = 0; i < lastIndex; i++) {
-          result.add(makePartName(tmpTable.getPartitionKeys(), partitions.get(i).getValues(),tmpTable.getParameters(),
-              conf));
+          result.add(makePartName(tmpTable.getPartitionKeys(), partitions.get(i).getValues(),
+              MetaStoreUtils.getDefaultPartitionName(tmpTable.getParameters(), conf)));
         }
         Collections.sort(result);
         GetPartitionNamesPsResponse response = new GetPartitionNamesPsResponse();
@@ -1413,7 +1413,7 @@ public class SessionHiveMetaStoreClient extends MetaStoreClientWrapper {
         int numPartitions = maxParts < 0 || maxParts > partitionList.size() ? partitionList.size() : maxParts;
         for(int i = 0; i < numPartitions; i++) {
           results.add(Warehouse.makePartName(table.getPartitionKeys(), partitionList.get(i).getValues(),
-              table.getParameters(), conf));
+              MetaStoreUtils.getDefaultPartitionName(table.getParameters(), conf)));
         }
         return results;
       }
@@ -1453,9 +1453,10 @@ public class SessionHiveMetaStoreClient extends MetaStoreClientWrapper {
         }
       }
 
+      String defaultPartitionName = MetaStoreUtils.getDefaultPartitionName(table.getParameters(), conf);
       try {
-        return Warehouse.makePartName(table.getPartitionKeys(), o1.getValues(), table.getParameters(), conf).compareTo(
-            Warehouse.makePartName(table.getPartitionKeys(), o2.getValues(), table.getParameters(), conf));
+        return Warehouse.makePartName(table.getPartitionKeys(), o1.getValues(), defaultPartitionName).compareTo(
+            Warehouse.makePartName(table.getPartitionKeys(), o2.getValues(), defaultPartitionName));
       } catch (MetaException e) {
         throw new RuntimeException(e);
       }
@@ -1815,7 +1816,8 @@ public class SessionHiveMetaStoreClient extends MetaStoreClientWrapper {
             tt.listPartitions();
         List<String> partitionNames = new ArrayList<>();
         for (Partition p : partitions) {
-          partitionNames.add(makePartName(table.getPartitionKeys(), p.getValues(), table.getParameters(), conf));
+          partitionNames.add(makePartName(table.getPartitionKeys(), p.getValues(),
+              MetaStoreUtils.getDefaultPartitionName(table.getParameters(), conf)));
         }
         if (partitionNames.isEmpty() && partitions.isEmpty()) {
           throw new MetaException("Cannot obtain list of partition by filter:\"" + request.getFilter() +
@@ -1890,8 +1892,8 @@ public class SessionHiveMetaStoreClient extends MetaStoreClientWrapper {
     assert table != null;
     ExpressionTree.FilterBuilder filterBuilder = new ExpressionTree.FilterBuilder(true);
     Map<String, Object> params = new HashMap<>();
-    exprTree.accept(new ExpressionTree.JDOFilterGenerator(conf,
-        table.getPartitionKeys(), filterBuilder, params, table.getParameters()));
+    exprTree.accept(new ExpressionTree.JDOFilterGenerator(conf, table.getPartitionKeys(), filterBuilder, params,
+        MetaStoreUtils.getDefaultPartitionName(table.getParameters(), conf)));
     StringBuilder stringBuilder = new StringBuilder(filterBuilder.getFilter());
     params.entrySet().stream().forEach(e -> {
       int index = stringBuilder.indexOf(e.getKey());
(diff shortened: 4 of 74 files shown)


vikramahuja1001:
@deniskuzZ , @chinnaraolalam could you please take a look at this PR?

hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/DynamicPartitionFileRecordWriterContainer.java
```diff
@@ -72,7 +73,8 @@ class DynamicPartitionFileRecordWriterContainer extends FileRecordWriterContaine
    */
   public DynamicPartitionFileRecordWriterContainer(
       RecordWriter<? super WritableComparable<?>, ? super Writable> baseWriter,
-      TaskAttemptContext context) throws IOException, InterruptedException {
+      TaskAttemptContext context, org.apache.hadoop.hive.metastore.api.Table tbl)
```
ngsg: Could we import `Table` and replace the qualified type name with it?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -876,7 +877,7 @@ public DynamicPartitionCtx createDPContext(
     Table table = IcebergTableUtil.getTable(conf, tableDesc.getProperties());
 
     DynamicPartitionCtx dpCtx = new DynamicPartitionCtx(Maps.newLinkedHashMap(),
-        hiveConf.getVar(ConfVars.DEFAULT_PARTITION_NAME),
+        PartitionUtils.getDefaultPartitionName(hmsTable.getParameters(), (HiveConf) conf),
```
ngsg: I think we can use `hiveConf` instead of `conf` here.

parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
```diff
@@ -104,6 +104,7 @@ alterTblPartitionStatementSuffix[boolean partition]
   | alterStatementSuffixRenameCol
   | alterStatementSuffixAddCol
   | alterStatementSuffixDropCol
+  | alterStatementSuffixSetDefaultPartition
```
ngsg: Do we need to support queries like `ALTER TABLE tbl PARTITION (part_spec) SET DEFAULT PARTITION TO "NA"`?
If not, I think it would be better to place `alterStatementSuffixSetDefaultPartition` under `alterTableStatementSuffix`.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
```diff
@@ -191,4 +192,17 @@ public static void addTablePartsOutputs(Hive db, Set<WriteEntity> outputs, Table
       }
     }
   }
+
+  public static String getDefaultPartitionName(Map<String, String> tableParams, HiveConf conf) {
+    return getDefaultPartitionName(tableParams, HiveConf.getVar(conf, HiveConf.ConfVars.DEFAULT_PARTITION_NAME));
+  }
+
+  public static String getDefaultPartitionName(Map<String, String> tableParams, String defaultPartitionName) {
```
ngsg: Can we unify the two `getDefaultPartitionName()` into one that takes a `Configuration`? It seems that we always have either `HiveConf` or `Configuration` when calling it.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/setdefaultpartition/AlterTableSetDefaultPartitionAnalyser.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.hadoop.hive.ql.parse.HiveParser;
+import org.apache.hadoop.hive.ql.parse.SemanticException;
+
+import java.util.Map;
+
+/**
+ * Analyzer for set default partition command.
+ */
+@DDLType(types = HiveParser.TOK_ALTERTABLE_SETDEFAULTPARTITION)
+public class AlterTableSetDefaultPartitionAnalyser extends AbstractAlterTableAnalyzer {
```
ngsg: Could we rename `Analyser` to use `Analyzer` as other subclasses?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/setdefaultpartition/AlterTableSetDefaultPartitionAnalyser.java
```diff
@@ -0,0 +1,52 @@
...
+  public AlterTableSetDefaultPartitionAnalyser(QueryState queryState) throws SemanticException {
+    super(queryState);
+  }
+
+  @Override
+  protected void analyzeCommand(TableName tableName, Map<String, String> partitionSpec, ASTNode command)
+      throws SemanticException {
+    String tableLevelDefaultPartitionName = unescapeSQLString(command.getChild(0).getText());
+    AlterTableSetDefaultPartitionDesc desc = new AlterTableSetDefaultPartitionDesc(tableName,
+            tableLevelDefaultPartitionName);
```
ngsg: Could we use an indent of 4 spaces(2 tabs) instead of 8 spaces?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/setdefaultpartition/AlterTableSetDefaultPartitionDesc.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final long serialVersionUID = 1L;
+  private final String tableLevelDefaultPartitionName;
+
+  public AlterTableSetDefaultPartitionDesc(TableName tableName, String tableLevelDefaultPartitionName) throws SemanticException {
+    super(AlterTableType.SETDEFAULTPARTITION, tableName, null, null, false, false, null);
+    this.tableLevelDefaultPartitionName = tableLevelDefaultPartitionName;
+  }
+
+  @Explain(displayName = "default partition name", explainLevels = { Level.USER, Level.DEFAULT, Level.EXTENDED })
+  public String tableLevelDefaultPartitionName() {
```
ngsg: Could we rename the method name to `getTableLevelDefaultPartitionName()`?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/setdefaultpartition/AlterTableSetDefaultPartitionDesc.java
```diff
@@ -0,0 +1,50 @@
...
+    return tableLevelDefaultPartitionName;
+  }
+
+  @Override
+  public boolean mayNeedWriteId() {
```
ngsg: The default bahaviour of this method is always returning `true`, so we don't need to override it with identical logic here.

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
```diff
@@ -1569,8 +1570,8 @@ public List<Partition> dropPartitions(String catName, String dbName, String tblN
         List<Partition> result = new ArrayList<>();
         for (Pair<Integer, byte[]> pair : partExprs) {
           byte[] expr = pair.getRight();
-          String filter = generateJDOFilter(table, expr,
-              conf.get(HiveConf.ConfVars.DEFAULT_PARTITION_NAME.varname));
+          String filter = generateJDOFilter(table, expr, PartitionUtils.getDefaultPartitionName(table.getParameters(),
+                  conf.get(HiveConf.ConfVars.DEFAULT_PARTITION_NAME.varname)));
```
ngsg: Could you restore the indentation level as before?

ql/src/java/org/apache/hadoop/hive/ql/plan/HiveOperation.java
```diff
@@ -49,6 +49,8 @@ public enum HiveOperation {
       null),
   ALTERTABLE_DROPCOL("ALTERTABLE_DROPCOL", HiveParser.TOK_ALTERTABLE_DROPCOL, new Privilege[]{Privilege.ALTER_METADATA},
           null),
+  ALTERTABLE_SETDEFAULTPARTITION("ALTERTABLE_SETDEFAULTPARTITION", HiveParser.TOK_ALTERTABLE_SETDEFAULTPARTITION, new Privilege[]{Privilege.ALTER_METADATA},
```
ngsg: I have a minor question here: if I unerstand correctly, the new operation can rename the default partition directory. Then would it be more appropriate to use ALTER_DATA? What do you think about it?

ql/src/test/queries/clientpositive/alter_table_set_default_partition.q
```diff
@@ -0,0 +1,65 @@
+drop table if exists tbl_orc;
```
ngsg: We don't need "DROP IF EXISTS"; CliDiver cleans up all the table generated during the tests.

ql/src/test/queries/clientpositive/alter_table_set_default_partition.q
```diff
@@ -0,0 +1,65 @@
...
+select count(*) from tbl_orc;
+INSERT INTO tbl_orc (a, b, c) values(4, 'ddd', 23);
+show partitions tbl_orc;
+select * from tbl_orc;
+select count(*) from tbl_orc;
+dfs -ls ${hiveconf:hive.metastore.warehouse.dir}/tbl_orc/;
+drop table if exists tbl_orc;
+
+
+create external table tbl_orc(a int, b string, c int) partitioned by (d int, e string, f bigint) stored as orc;
```
ngsg: Could we rename this table so that each table in the qfile has a unique name?
e.g. `tbl_orc` -> `tbl_orc_multipart`

ql/src/test/queries/clientpositive/msck_repair_9.q
```diff
@@ -7,28 +7,31 @@ INSERT INTO tbl_x values(1, 'aaa', 12, 2);
...
 
-SET hive.exec.default.partition.name=OTHER_PARTITION;
+alter table tbl_y set default partition to 'OTHER_PARTITION';
 INSERT INTO tbl_y (id, name, day) values(4, 'ddd', 3);
-
 SHOW PARTITIONS tbl_y;
 
 set hive.msck.path.validation=ignore;
+
+dfs ${system:test.dfs.mkdir} -p ${system:test.tmp.dir}/apps/hive/warehouse/test.db//tbl_x/month=DEFAULT/day=DEFAULT;
```
ngsg: nit `//` -> `/`
***
What do you think about generating the dummy directory before the first msck call? I think the first call examines the bahaviour of `hive.msck.path.valudation=skip`, but after the patch applied, IIUC, there is no invalid directory at that point.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -888,7 +887,7 @@ public boolean generateSqlFilterForPushdown(String catName, String dbName, Strin
     result.catName = catName;
     result.filter = PartitionFilterGenerator.generateSqlFilter(catName, dbName, tableName,
         partitionKeys, tree, result.params, result.joins, dbHasJoinCastBug,
-        ((defaultPartitionName == null) ? defaultPartName : defaultPartitionName),
+        ((defaultPartitionName == null) ? getDefaultPartitionName(tableParams, conf ) : defaultPartitionName),
```
ngsg: nit: `conf )` -> `conf)`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -3352,12 +3352,13 @@ public List<String> listPartitionNames(String catName, String dbName, String tab
...
       final String order, final int maxParts) throws MetaException, NoSuchObjectException {
-    final String defaultPartitionName = getDefaultPartitionName(defaultPartName);
+    final String defaultPartitionName = getDefaultPartitionName(defaultPartName,
+            ensureGetMTable(catName, dbName, tblName).getParameters());
     final boolean isEmptyFilter = exprBytes.length == 1 && exprBytes[0] == -1;
     ExpressionTree tmp = null;
     if (!isEmptyFilter) {
       tmp = PartFilterExprUtil.makeExpressionTree(expressionProxy, exprBytes,
-          getDefaultPartitionName(defaultPartName), conf);
+              defaultPartitionName, conf);
```
ngsg: Could you restore the indentation level as before?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
```diff
@@ -4761,7 +4771,7 @@ protected boolean canUseDirectSql(GetHelper<List<Partition>> ctx) throws MetaExc
           // if there are more than one filter string we AND them together
           initExpressionTree();
           return directSql.generateSqlFilterForPushdown(table.getCatName(), table.getDbName(), table.getTableName(),
-                  table.getPartitionKeys(), tree, null, filter);
+                  table.getPartitionKeys(), tree, null, filter, ctx.table.getParameters());
```
ngsg: nit: `ctx.table.getParameters()` -> `table.getParameters()`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/cache/CachedStore.java
```diff
@@ -3441,4 +3440,12 @@ private boolean shouldGetConstraintFromRawStore(String catName, String dbName, S
     return !shouldCacheTable(catName, dbName, tblName) || (canUseEvents && rawStore.isActiveTransaction())
         || !sharedCache.isTableConstraintValid(catName, dbName, tblName);
   }
+
+  public static String getDefaultPartitionName(Map<String, String> tableParams, Configuration conf) {
```
ngsg: Could we unify `CachedStore#getDefaultPartitionName` and `HiveMetaStoreChecker#getDefaultPartitionName`?

ngsg (commented):
Hi @vikramahuja1001, I have the same concern about setting the default partition name, and I'm glad to see the patch addressing it. I have reviewed the changes, except for `HiveAlterHandler.java` (I need more time to study the changes and the relevant context), and left comments on the places where I have questions. Could you please review my comments on the changes?

hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/DynamicPartitionFileRecordWriterContainer.java
```diff
@@ -88,7 +90,8 @@ public DynamicPartitionFileRecordWriterContainer(
     this.dynamicContexts = new HashMap<String, org.apache.hadoop.mapred.TaskAttemptContext>();
     this.dynamicObjectInspectors = new HashMap<String, ObjectInspector>();
     this.dynamicOutputJobInfo = new HashMap<String, OutputJobInfo>();
-    this.HIVE_DEFAULT_PARTITION_VALUE = HiveConf.getVar(context.getConfiguration(), HiveConf.ConfVars.DEFAULT_PARTITION_NAME);
+    this.HIVE_DEFAULT_PARTITION_VALUE = PartitionUtils.getDefaultPartitionName(tbl.getParameters(),
+            HiveConf.getVar(context.getConfiguration(), HiveConf.ConfVars.DEFAULT_PARTITION_NAME));
```
Indhumathi27: can you move `HiveConf.getVar(context.getConfiguration(), HiveConf.ConfVars.DEFAULT_PARTITION_NAME)` inside getDefaultPartitionName method

ql/src/test/queries/clientpositive/msck_repair_9.q
```diff
@@ -7,28 +7,31 @@ INSERT INTO tbl_x values(1, 'aaa', 12, 2);
...
 CREATE EXTERNAL TABLE tbl_y (id INT, name STRING) PARTITIONED BY (month INT, day INT) stored as ORC location '${system:test.tmp.dir}/apps/hive/warehouse/test.db/tbl_x/';
+alter table tbl_y set default partition to 'ANOTHER_PARTITION';
 
 set hive.msck.path.validation=skip;
 
 MSCK REPAIR TABLE tbl_y;
 
 SHOW PARTITIONS tbl_y;
 
-SET hive.exec.default.partition.name=SECOND_PARTITION;
```
Indhumathi27, line 23: Can you add a test to validate `hive.exec.default.partition.name` is no more session configurable

vikramahuja1001:
Thanks for reviewing this PR @ngsg @Indhumathi27 
I will handle the review comments, push the changes soon.

hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/DynamicPartitionFileRecordWriterContainer.java
vikramahuja1001: Done

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
vikramahuja1001: Changed to hiveconf

parser/src/java/org/apache/hadoop/hive/ql/parse/AlterClauseParser.g
vikramahuja1001: Yes, makes sense. Changed the same.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/partition/PartitionUtils.java
vikramahuja1001: Yes, unified it.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/setdefaultpartition/AlterTableSetDefaultPartitionAnalyser.java
vikramahuja1001: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/setdefaultpartition/AlterTableSetDefaultPartitionAnalyser.java
vikramahuja1001: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/setdefaultpartition/AlterTableSetDefaultPartitionDesc.java
vikramahuja1001: Done

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/setdefaultpartition/AlterTableSetDefaultPartitionDesc.java
vikramahuja1001: removed it, thanks for pointing it out.

ql/src/java/org/apache/hadoop/hive/ql/metadata/SessionHiveMetaStoreClient.java
vikramahuja1001: done

ql/src/test/queries/clientpositive/alter_table_set_default_partition.q
vikramahuja1001: Changed the name to make all table names unique in this qfile

ql/src/test/queries/clientpositive/msck_repair_9.q
vikramahuja1001: Sure, made the change

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
vikramahuja1001: done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
vikramahuja1001: Sure, done

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ObjectStore.java
vikramahuja1001: Changed

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/cache/CachedStore.java
vikramahuja1001: Done. HiveMetaStoreChecker has the code for this now.

hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/DynamicPartitionFileRecordWriterContainer.java
vikramahuja1001: Done

ql/src/test/queries/clientpositive/alter_table_set_default_partition.q
vikramahuja1001: Removed it from the .q file

ql/src/java/org/apache/hadoop/hive/ql/plan/HiveOperation.java
vikramahuja1001: Yes, makes sense. I have changed it to ALTER_DATA. Even compaction is handled similarly. Thanks for pointing it out.

Aggarwal-Raghav:
@vikramahuja1001 , i think change is required in `org.apache.hadoop.hive.ql.exec.DDLPlanUtils#checkIfDefaultPartition()`
https://github.com/apache/hive/blob/73503291cbdd0f95bc33add0769234fed6683840/ql/src/java/org/apache/hadoop/hive/ql/exec/DDLPlanUtils.java#L307

Aggarwal-Raghav:
Remove hardcoding from here as well + update comments as well accordingly
org.apache.hadoop.hive.common.FileUtils#escapePathName(java.lang.String, java.lang.String)
https://github.com/apache/hive/blob/413069ea0faeb893fe78faa215b08a70ece80595/common/src/java/org/apache/hadoop/hive/common/FileUtils.java#L298

https://github.com/apache/hive/blob/90820ac0a241468482073beb269bcaa3928fbc6f/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java#L310

Aggarwal-Raghav:
Remove hardcoding from following place:

https://github.com/apache/hive/blob/733d4f09011f27bea17a5301537a9c6174f28874/hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/FileOutputCommitterContainer.java#L715

https://github.com/apache/hive/blob/ba0217ff17501fb849d8999e808d37579db7b4f1/hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/HCatFileUtil.java#L71

vikramahuja1001:
@Aggarwal-Raghav ,very good point, makes sense. Thanks for pointing this out. There are indeed a lot of places where HIVE_DEFAULT_PARTITION is hard coded. Have put a few commits to fix that.

vikramahuja1001:
@ngsg , @Aggarwal-Raghav , @Indhumathi27 i have reworked all the comments and pushed all the changes. In the build pipeline all the tests are passing but there seems to be javadoc issue with the thrift gen code, no idea what's causing that. Looking at that atm.

Could you guys please re-review it again?

@chinnaraolalam  and @deniskuzZ , could you also please take a look at this PR and maybe review it?

ngsg:
Thanks a lot for the hard work on this PR. I haven't fully reviewed all the changes yet (the diff is quite large), but let me leave a few comments for now:

- In some cases, handling default partition name isn't necessary. I would suggest introducing new methods with Preconditions for those cases to avoid passing {tableParam, conf} or a default partition name. For example, when calling `Warehouse#escapePathName` in `Warehouse#makePartNameUtil`: `e.getKey()` doesn't require a default partition name, and `e.getValue()` is guaranteed not to be empty or null.

- Extending the above point, I would suggest passing just a default partition name instead of both tableParam and configuration whenever possible. For example, since `FileUtils#escapePathNames` already takes `defaultPath`, it would be better to require callers to always pass a proper default partition name. This way, we propagate only the minimal required information and lower the risk of human error when dealing with a potentially mutable map instance.

- Regarding `InsertHiveLocksCommand`, I would suggest considering alternatives. Do we really need to ship the entire content of tableParams over `LockComponent`? Could we instead just send the default partition name, or ensure that the partition name is not null or empty? (We might even assume there is no empty string or null based on the implementation of `com.google.common.base.Splitter$MapSplitter#split`, though this code is outside our control.)

- I haven't fully thought this through, but since the changes now affect the Thrift interface, I would like to hear others' opinions. What do you think about extending the definition of `Table` and `Partition` in the Thrift interface instead of using tableParams? In this way, we might be able to simplify default partition determining steps.

Also, regarding the error in Javadoc, it seems the issue comes from the javadoc comment in `LockComponentBuilder#setTableParams`: the parameter name should be `tableParams`, not `tableName`.

vikramahuja1001:
@ngsg , thanks for spending time to review this PR and providing your feedback. I will rework as per your above comment and push the changes in a few days. Thanks

vikramahuja1001:
@ngsg Apologies for the delayed response. Couldn't revert back/rework due to limited connectivity. I have reworked the complete PR wrt most of the points mentioned in your previous comment. 

Please take a look at the PR once. Let me know if something else also needs to be handled.

vikramahuja1001:
Cc @chinnaraolalam , could you also take a look at this PR

common/src/java/org/apache/hadoop/hive/common/FileUtils.java
```diff
@@ -173,15 +173,15 @@ public static String makePartName(List<String> partCols, List<String> vals) {
...
   public static String makePartName(List<String> partCols, List<String> vals,
-      String defaultStr) {
+      String defaultStr, String defaultPartitionName) {
     StringBuilder name = new StringBuilder();
     for (int i = 0; i < partCols.size(); i++) {
       if (i > 0) {
         name.append(Path.SEPARATOR);
       }
-      name.append(escapePathName((partCols.get(i)).toLowerCase(), defaultStr));
+      name.append(escapePathName((partCols.get(i)).toLowerCase(), defaultStr, defaultPartitionName));
```
ngsg, line 182: Since it is a column name, I think we don't need to pass default partition name, which cannot be a column name.

common/src/java/org/apache/hadoop/hive/common/FileUtils.java
```diff
@@ -214,15 +214,15 @@ public static String makeDefaultListBucketingDirName(List<String> skewedCols,
...
    */
-  public static String makeListBucketingDirName(List<String> lbCols, List<String> vals) {
+  public static String makeListBucketingDirName(List<String> lbCols, List<String> vals, String defaultPartitionName) {
     StringBuilder name = new StringBuilder();
     for (int i = 0; i < lbCols.size(); i++) {
       if (i > 0) {
         name.append(Path.SEPARATOR);
       }
-      name.append(escapePathName((lbCols.get(i)).toLowerCase()));
+      name.append(escapePathName((lbCols.get(i)).toLowerCase(), defaultPartitionName));
```
ngsg, line 223: The same as above, we don't need default partition name for column name.

common/src/java/org/apache/hadoop/hive/common/FileUtils.java
```diff
@@ -274,8 +274,8 @@ static boolean needsEscaping(char c) {
     return c < charToEscape.size() && charToEscape.get(c);
   }
 
-  public static String escapePathName(String path) {
-    return escapePathName(path, null);
+  public static String escapePathName(String path,  String defaultPartitionName) {
```
ngsg, line 277: Could we change the two `escapePathName` as follows:

```suggestion
  public static String escapePathName(String path, String defaultPartitionName) {
    if (path == null || path.isEmpty()) {
      Preconditions.checkArgument(defaultPartitionName != null && !defaultPartitionName.isEmpty());
      return escapePathName(defaultPartitionName);
    } else {
      return escapePathName(path);
    }
  }

  /**
   * Escapes a path name.
   * @param path The path to escape.
   * @return An escaped path name.
   */
  public static String escapePathName(String path) {
    Preconditions.checkArgument(path != null && !path.isEmpty());

    //  Fast-path detection, no escaping and therefore no copying necessary
    int firstEscapeIndex = -1;
```

hcatalog/core/src/main/java/org/apache/hive/hcatalog/mapreduce/FileOutputCommitterContainer.java
```diff
@@ -453,11 +458,12 @@ private Map<String, String> getStorerParameterMap(StorerInfo storer) {
     return params;
   }
 
-  private Path constructPartialPartPath(Path partialPath, String partKey, Map<String, String> partKVs) {
+  private Path constructPartialPartPath(Path partialPath, String partKey, Map<String, String> partKVs,
+      String defaultPartitionName) {
 
-    StringBuilder sb = new StringBuilder(FileUtils.escapePathName(partKey));
+    StringBuilder sb = new StringBuilder(FileUtils.escapePathName(partKey, defaultPartitionName));
```
ngsg, line 464: This is a column name, so it would be better to use `escapePathName` without `defaultPartitionName`.

ql/src/java/org/apache/hadoop/hive/ql/plan/BucketMapJoinContext.java
```diff
@@ -207,31 +209,32 @@ public String getMappingBigFile(String alias, String smallFile) {
...
   // prepends partition spec of input path to candidate file name
-  private String prependPartSpec(String inputPath, String fileName) {
+  private String prependPartSpec(String inputPath, String fileName, Configuration conf) {
     Map<String, String> mapping = inputToPartSpecMapping == null ?
         inputToPartSpecMapping = revert(bigTablePartSpecToFileMapping) : inputToPartSpecMapping;
     String partSpec = mapping.get(URI.create(inputPath).getPath());
+    // if partSpec is not null or not empty, tableParams and conf can be null
     return partSpec == null || partSpec.isEmpty() ? fileName :
-      "(" + FileUtils.escapePathName(partSpec) + ")" + fileName;
+      "(" + FileUtils.escapePathName(partSpec, MetaStoreUtils.getDefaultPartitionName( null, conf)) + ")" + fileName;
```
ngsg, line 237: Could you verify the logic here? It looks like we are prepending an escaped `partSpec` if `partSpec` is not null and nonempty. Then can we safely use `FileUtils.escapePathName` without default partition name?

ql/src/java/org/apache/hadoop/hive/ql/plan/DynamicPartitionCtx.java
```diff
@@ -101,14 +103,14 @@ public DynamicPartitionCtx(List<String> partColNames, String defaultPartName,
     this.customSortNullOrder = new LinkedList<>();
   }
 
-  public DynamicPartitionCtx(Map<String, String> partSpec, String defaultPartName,
```
ngsg, line 104: Could you revert the changes here? It seems that the constructor can be used as-is.

ql/src/test/org/apache/hadoop/hive/ql/exec/TestFileSinkOperator.java
```diff
@@ -401,7 +402,9 @@ private FileSinkOperator getFileSink(AcidUtils.Operation writeType,
       partCols.add(new ExprNodeColumnDesc(TypeInfoFactory.stringTypeInfo, PARTCOL_NAME, "a", true));
       Map<String, String> partColMap= new LinkedHashMap<String, String>(1);
       partColMap.put(PARTCOL_NAME, null);
-      DynamicPartitionCtx dpCtx = new DynamicPartitionCtx(partColMap, "Sunday", 100);
+      HiveConf conf = new HiveConf();
+      conf.set(HiveConf.ConfVars.DEFAULT_PARTITION_NAME.varname, "Sunday");
+      DynamicPartitionCtx dpCtx = new DynamicPartitionCtx(partColMap, 100, MetaStoreUtils.getDefaultPartitionName(null, conf));
```
ngsg, line 407: Could we use this test as-is? Since the test is not focused on default partition name, I think it is OK to directly pass "Sunday".

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
```diff
@@ -556,8 +556,8 @@ public static String makePartPath(Map<String, String> spec)
    * @return partition name
    * @throws MetaException
    */
-  public static String makePartNameUtil(Map<String, String> spec, boolean addTrailingSeperator, boolean dynamic)
-          throws MetaException {
+  public static String makePartNameUtil(Map<String, String> spec, boolean addTrailingSeperator, boolean dynamic,
+      String defaultPartitionName) throws MetaException {
```
ngsg, line 560: Could we keep the previous method definition? Since the method body stops or throws an error in case `e.getValue()` is null or empty, I think we don't need `defaultPartitionName` in both `escapePathName` calls.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/FileUtils.java
```diff
@@ -299,15 +300,15 @@ public static String escapePathName(String path) {
    *          The default name for the path, if the given path is empty or null.
    * @return An escaped path name.
    */
-  public static String escapePathName(String path, String defaultPath) {
+  public static String escapePathName(String path, String defaultPath, String defaultPartitionName) {
```
ngsg, line 303: Could we update the methods as I commented in `hive.common.FileUtils`?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
```diff
@@ -882,7 +880,7 @@ public List<FileStatus> getFileStatusesForUnpartitionedTable(Database db, Table
    * @throws MetaException
    */
   public static String makePartName(List<FieldSchema> partCols,
-      List<String> vals, String defaultStr) throws MetaException {
+      List<String> vals, String defaultStr, String defaultPartitionName) throws MetaException {
```
ngsg, line 883: Could we unify `defaultStr` and `defaultPartitionName` as suggested in `FileUtils#escapePathName`?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
```diff
@@ -823,8 +821,8 @@ public boolean isDir(Path f) throws MetaException {
   }
 
   public static String makePartName(List<FieldSchema> partCols,
-      List<String> vals) throws MetaException {
-    return makePartName(partCols, vals, null);
+      List<String> vals, String defaultPartitionName) throws MetaException {
```
ngsg, line 824: Could we remove this method and force the callers to use the below method?

PR #6014, opened 2025-08-07
HIVE-29084: use nextAlias for the output schema of LV columns after AST Conversion

konstantinb:

HIVE-29084: Proposing changes to ASTConverter's logic of tableAlias assignment for Lateral View Queries

Before these changes, ASTConverter used to assign the base table alias as the tableAlias of all columns of the query tree. Technically, LV columns are "separate" tables participating in an implicit join. Therefore, PPD processing considered filters with conditions between table columns and LV columns as conditions on the columns of the same table. 
The following condition: https://github.com/apache/hive/blob/5dddb6ebf46ac40b91ab3273e3bb8910d6163dc2/ql/src/java/org/apache/hadoop/hive/ql/ppd/ExprWalkerProcFactory.java#L262
made these expressions considered "pushable candidates", while the subsequent processing logic has no knowledge on how to optimize/convert/process such expressions, so they are ultimately discarded during the LateralViewJoinerPPD.removeAllCandidates() call

A very simple query to confirm the bug is
```
SELECT t.key, t.value, lv.col
FROM (SELECT '238' AS key, 'val_238' AS value) t
LATERAL VIEW explode(array('238', '86', '311')) lv AS col
WHERE t.key = '333' OR lv.col = '86'
ORDER BY t.key, lv.col;
```

No

- Tested locally primarily with TestMiniLlapLocalCliDriver
- Applied the same patch to a custom Hive implementation based on Hive 4.0.1, confirmed the accuracy of the results of impacted queries after the tix

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
@@ -1229,10 +1229,6 @@ public class ASTConverter {
         add(new ColumnInfo(tabAlias, field.getName()));
       }
     }
-
-    Schema(java.util.Collection<? extends ColumnInfo> c) {
-      super(c);
-    }
   }
 
   /*


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
```diff
@@ -602,12 +604,26 @@ private QueryBlockInfo convertSource(RelNode r) throws CalciteSemanticException
...
       QueryBlockInfo tableFunctionSource = convertSource(tfs.getInput(0));
-      String sqAlias = tableFunctionSource.schema.get(0).table;
-      // the schema will contain the base table source fields
-      s = new Schema(tfs, sqAlias);
 
-      ast = createASTLateralView(tfs, s, tableFunctionSource, sqAlias);
+      // Create schema that preserves base table columns with original alias,
+      // but gives new UDTF columns the unique lateral view alias
+      int baseFieldCount = tableFunctionSource.schema.size();
+      List<RelDataTypeField> allOutputFields = tfs.getRowType().getFieldList();
```
zabetak: Is it necessary to preserve the alias from the base table? Basically, I would like to know if we can get rid of the additional complexity to maintain aliases by just generating a fresh alias (`nextAlias()`) for each lateral view.

For instance, when we handle the branches of a `Union`, we don't care to maintain aliases from below and just generate a fresh one. I am wondering if the same approach is applicable here.

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
+-- HIVE-29084: LATERAL VIEW cartesian product schema construction
+DROP TABLE IF EXISTS src_lv_cart;
```
zabetak: You can skip the test cleanup logic since it is taken care of from test infrastructure. 
```
DROP TABLE IF EXISTS src_lv_cart;
DROP TABLE src_lv_cart;
DROP TABLE tmp_join;
```

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+CREATE TABLE src_lv_cart (val_array array<string>);
+INSERT INTO src_lv_cart VALUES (array('a', 'b'));
+
+SELECT first_val, second_val
+FROM src_lv_cart
+LATERAL VIEW explode(val_array) lv1 AS first_val
+LATERAL VIEW explode(val_array) lv2 AS second_val
+WHERE first_val != second_val
+ORDER BY first_val, second_val;
```
zabetak: The `ORDER BY` clause is not relevant for the problem in question so in order to simplify the tests please drop it from everywhere.

In order to keep the query results sorted/reproducible you can add the `-- SORT_QUERY_RESULTS` directive once at the beginning of the file.

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+ORDER BY lv1.val1, lv2.val2, lv3.val3;
+
+SELECT first_val, second_val
+FROM (SELECT array('p', 'q', 'r') AS val_array) src
+LATERAL VIEW explode(val_array) lv1 AS first_val
+LATERAL VIEW explode(val_array) lv2 AS second_val
+WHERE first_val = 'p' OR second_val = 'r'
+ORDER BY first_val, second_val;
+
+SET hive.cbo.enable=false;
```
zabetak: Disabling CBO is highly discouraged in production and the respective codepath/property is on the road to deprecation. We are not really gonna fix issues in this area so adding regression tests is somewhat redundant. Please drop all tests that make use of `hive.cbo.enable=false;`

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+SELECT first_val, second_val
+FROM (SELECT array('m', 'n') AS val_array) src
+LATERAL VIEW explode(val_array) lv1 AS first_val
+LATERAL VIEW explode(val_array) lv2 AS second_val
+WHERE first_val != second_val
+ORDER BY first_val, second_val;
+
+SET hive.cbo.enable=true;
```
zabetak: Consider doing `hive.cbo.enable=true;` only once at the beginning of the file or not at all given that it is a widely accepted default that is very unlikely to change in the future.

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+SET hive.optimize.ppd=false;
+SELECT first_val, second_val
+FROM (SELECT array('s', 't') AS val_array) src
+LATERAL VIEW explode(val_array) lv1 AS first_val
+LATERAL VIEW explode(val_array) lv2 AS second_val
+WHERE first_val != second_val
+ORDER BY first_val, second_val;
+
+SET hive.cbo.enable=false;
+SET hive.optimize.ppd=false;
```
zabetak: The problem reported here is not reproducible when `hive.optimize.ppd` is `false` so I am not sure if we need extra tests for this purpose.

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+LATERAL VIEW explode(arr_with_nulls) lv2 AS val2
+WHERE val1 IS NOT NULL AND val2 IS NOT NULL AND val1 != val2
+ORDER BY val1, val2;
+
+SELECT num1, num2
+FROM (SELECT array(1, 2, 3) AS num_array) src
+LATERAL VIEW explode(num_array) lv1 AS num1
+LATERAL VIEW explode(num_array) lv2 AS num2
+WHERE num1 < num2
+ORDER BY num1, num2;
```
zabetak: Given that the problem and the fix are targeting the `ASTConverter` and in particular the aliases testing with different datatypes and predicates feels a bit out of context. In addition, I am not sure if we gain something in test coverage with the various combinations so I would prefer to leave out any tests that are not directly relevant to the problem.

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+
+SELECT outer_result
+FROM (
+  SELECT CONCAT(inner_val, '_processed') AS outer_result
+  FROM (SELECT array('sub1', 'sub2') AS sub_array) inner_t
+  LATERAL VIEW explode(sub_array) inner_lv AS inner_val
+) src
+ORDER BY outer_result;
+
+CREATE TEMPORARY TABLE tmp_join (id int, data string, join_array array<string>);
```
zabetak: What's the purpose of the tests involving the temporary table?

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+SET hive.optimize.ppd=true;
+SELECT outer_val, inner_val1, inner_val2
+FROM (
+  SELECT array('alpha', 'beta') AS outer_array, array('1', '2') AS inner_array
+) src
+LATERAL VIEW explode(outer_array) lv_outer AS outer_val
+LATERAL VIEW explode(inner_array) lv_inner1 AS inner_val1
+LATERAL VIEW explode(inner_array) lv_inner2 AS inner_val2
+WHERE outer_val != 'alpha' OR (inner_val1 != inner_val2)
+ORDER BY outer_val, inner_val1, inner_val2;
```
zabetak: Can you please add a brief comment before each test to clarify its purpose/need?

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+
+SELECT id1, name1, id2, name2
+FROM (SELECT array(struct(1, 'Alice'), struct(2, 'Bob'), struct(3, 'Charlie')) AS struct_array) src
+LATERAL VIEW inline(struct_array) lv1 AS id1, name1
+LATERAL VIEW inline(struct_array) lv2 AS id2, name2
+WHERE id1 != id2
+ORDER BY id1, id2;
+
+SELECT val1, val2
+FROM (SELECT array('a1', 'a2', 'a3', 'a4', 'a5', 'a6', 'a7', 'a8', 'a9', 'a10') AS large_array) src
```
zabetak: How does increasing/decreasing the values inside the array relates to the problem?

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
```diff
@@ -0,0 +1,172 @@
...
+SELECT val, COUNT(*) AS cnt
+FROM (SELECT array('group1', 'group2', 'group1', 'group3') AS grp_array) src
+LATERAL VIEW explode(grp_array) lv AS val
+GROUP BY val
+ORDER BY val;
+
+SELECT val, ROW_NUMBER() OVER (ORDER BY val) AS rn
+FROM (SELECT array('w1', 'w2', 'w3', 'w1') AS win_array) src
+LATERAL VIEW explode(win_array) lv AS val
+ORDER BY rn;
```
zabetak: Why is it important to test window functions with lateral views? 

We can probably generate an arbitrary number of SQL queries that are making use of lateral views in combinations with other SQL features so I am trying to understand what's the reasoning behind those variants and how they relate to the problem/fix.

ql/src/test/results/clientpositive/llap/lineage2.q.out
```diff
@@ -724,7 +724,7 @@ PREHOOK: query: select identity, ep1_id from relations
...
+{"version":"1.0","engine":"tez","database":"default","hash":"436a649a0d9540e8f093f8353d86813a","queryText":"select identity, ep1_id from relations\n  lateral view explode(ep1_ids) nav_rel as ep1_id","edges":[{"sources":[2],"targets":[0],"edgeType":"PROJECTION"},{"sources":[3],"targets":[1],"expression":"nav_rel.ep1_id","edgeType":"PROJECTION"}],"vertices":[{"id":0,"vertexType":"COLUMN","vertexId":"identity"},{"id":1,"vertexType":"COLUMN","vertexId":"ep1_id"},{"id":2,"vertexType":"COLUMN","vertexId":"default.relations.identity"},{"id":3,"vertexType":"COLUMN","vertexId":"default.relations.ep1_ids"}]}
```
zabetak, line 727: The lineage information changed but not sure what information is correct/better. The new or the old one?

zabetak: https://issues.apache.org/jira/browse/HIVE-12287 seems relevant

zabetak: Hmm, I see that the same line was modified in HIVE-27399.

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
konstantinb: @zabetak

The original bug originated from query results differing, being accurate with hive.cbo.enable=false and inaccurate with hive.cbo.enable=true

This was the main reason for including these tests, but, technically, after my fix, there should be no difference with either true or false. If there is no need to ensure the matching results regardless of whether CBO is enabled or disabled, I will be happy to drop these redundancies

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
konstantinb: @zabetak 
removing those; followed
`
When you do need to use a SELECT statement, make sure you use the ORDER BY clause to minimize the chances of spurious diffs due to output order differences leading to test failures.`

from https://hive.apache.org/development/qtest/#tutorial-how-to-add-a-new-test-case

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
konstantinb: It looks like
`final String baseTableAlias = nextAlias();`
works just as well, thank you for the suggestion!

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
konstantinb: sorry, went overboard with too elaborate tests, removing

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
konstantinb: Doh, made my comment with not fully recompiled code. It is very essential to preserve the initial tableAlias for all base table columns; before the change, the whole new schema was constructed with it, see 

`String sqAlias = tableFunctionSource.schema.get(0).table;
`
in the old code. The fix is to have _fewer_ columns to be assigned to the original tableAlias. I have changed the variable name to better match with pre-existing code

konstantinb:
> Many thanks for the PR @konstantinb !
> 
> I left a lot of comments in the test case but let me highlight that I really appreciate the effort that you put in crafting all those queries. Feel free to push-back on anything that you believe that needs to be retained and we can iterate more based on your feedback.

@zabetak thank you very much for your thorough review. I have provided answers about code changes and built a new, fairly small set of tests. Waiting on the pipeline now. Will get back on the lineage test result changes, and to the PR description soon.

Please see the following screenshot on how the old code changes test output:
HIVE-29084-extra-results-before-fix

ql/src/test/results/clientpositive/llap/lineage2.q.out
konstantinb: I believe the data with the fix is a bit more accurate, as it reflects the actual table alias and column name instead of generic values, see the part of the query:
`lateral view explode(ep1_ids) nav_rel as ep1_id`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
zabetak: From the syntax definition, a [LATERAL VIEW](https://hive.apache.org/docs/latest/language/languagemanual-lateralview/) is a virtual table with a user-defined table alias. Conceptually, every column that is in the output of the lateral view has the same table alias so I would expect that all columns in the same schema should have the same alias.

For all conversions, inside the `ASTConverter` we should distinguish the input schema(s) from the output schema. Both are very important for correctly and unambiguously constructing the AST/SQL query. For the lateral view case, input and output schema are somewhat mixed together and maybe they shouldn't. Some code inside the `createASTLateralView` method operates on the input schema and some other on the output schema. In other words, up to a certain point in the code, I think we could use the schema as is from the input/source and once we are done we could simply generate the output (new) schema using a new (generated) table alias. The idea is outlined on the comment below.

konstantinb:
@zabetak I'd greatly appreciate taking a second peek at this PR

zabetak:
@konstantinb I will check tomorrow. Apologies for the delay but I was off for some time.

ql/src/test/queries/clientpositive/lateral_view_cartesian_test.q
zabetak: I guess the choice between `SORT_QUERY_RESULTS` and explicit `ORDER BY` in the query is somewhat subjective.
Both can avoid test flakiness and each has its own advantages & disadvantages.

Putting an `ORDER BY` in every query makes the tests more verbose and expands its scope. The plans will have more operators, `EXPLAIN` outputs will contain more info than strictly necessary, and potentially more rules will match/apply and affect the output plan. On the positive side, it is a native way to enforce sorted output and avoid potential test flakiness.

The `SORT_QUERY_RESULTS` applies to all queries inside the file and it is a post-processing step completely independent of the query execution. Test inputs/outputs are less verbose and flakiness does not interfere with the query execution and the actual testing scope.

Personally, for this case I feel that `SORT_QUERY_RESULTS` is a better choice but don't feel that strongly about it. I am OK to accept the `ORDER BY` approach if you prefer that. However, currently the test file contains both `SORT_QUERY_RESULTS` and `ORDER BY` clauses so we should remove one of them. I leave the final choice to you.

ql/src/test/queries/clientpositive/lateral_view_cbo_ppd_filter_loss.q
```diff
@@ -0,0 +1,33 @@
+-- SORT_QUERY_RESULTS
+-- HIVE-29084: LATERAL VIEW cartesian product schema construction
+
+-- Verifies PPD doesn't eliminate OR filter comparing base table vs lateral view columns
+SELECT t.key, t.value, lv.col
+FROM (SELECT '238' AS key, 'val_238' AS value) t
+LATERAL VIEW explode(array('238', '86', '311')) lv AS col
+WHERE t.key = '333' OR lv.col = '86'
```
zabetak: The first part of the disjunction (`t.key = '333'`) is not verified. Given that the base table (`t`) does not have the value `333` we cannot verify that the respective part of the filter was not removed just by looking at the query output. Currently we are only testing `WHERE lv.col = '86'` not sure if that's the intention.

ql/src/test/queries/clientpositive/lateral_view_cbo_ppd_filter_loss.q
```diff
@@ -0,0 +1,33 @@
+-- SORT_QUERY_RESULTS
+-- HIVE-29084: LATERAL VIEW cartesian product schema construction
+
+-- Verifies PPD doesn't eliminate OR filter comparing base table vs lateral view columns
+SELECT t.key, t.value, lv.col
+FROM (SELECT '238' AS key, 'val_238' AS value) t
```
zabetak: nit: It seems that the value column (`'val_238' AS value`) is somewhat redundant for this and all subsequent queries so possibly the test cases can be simplified a bit further.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
```diff
@@ -602,12 +603,23 @@ private QueryBlockInfo convertSource(RelNode r) throws CalciteSemanticException
...
+      Stream<ColumnInfo> baseColumnsStream = allOutputFields.subList(0, baseFieldCount).stream()
+          .map(field -> new ColumnInfo(sqAlias, field.getName()));
+
+      final String lateralViewAlias = nextAlias();
+      Stream<ColumnInfo> udtfColumnsStream =
+          allOutputFields.subList(baseFieldCount, allOutputFields.size()).stream()
+              .map(field -> new ColumnInfo(lateralViewAlias, field.getName()));
+
+      s = new Schema(Stream.concat(baseColumnsStream, udtfColumnsStream).toList());
+      ast = createASTLateralView(tfs, s, tableFunctionSource, lateralViewAlias);
```
zabetak: Concretely, I was thinking something like the following:
```suggestion
      final String lateralViewAlias = nextAlias();
      ast = createASTLateralView(tfs, tableFunctionSource, lateralViewAlias);
      s = new Schema(tfs, lateralViewAlias);
```
assuming that the code inside `createASTLateralView` that needs a schema can use directly `tableFunctionSource.schema`.

In this case, the "input" schema is that inside `tableFunctionSource` and the "output" schema is the one constructed in `s`.

ql/src/test/queries/clientpositive/lateral_view_cbo_ppd_filter_loss.q
```diff
@@ -0,0 +1,33 @@
+-- SORT_QUERY_RESULTS
+-- HIVE-29084: LATERAL VIEW cartesian product schema construction
```
zabetak: nit: Remove completely or use the same summary with the JIRA ticket.

ql/src/test/queries/clientpositive/lateral_view_cbo_ppd_filter_loss.q
konstantinb: this is indeed a redundant column in all test queries, eliminating

ql/src/test/queries/clientpositive/lateral_view_cbo_ppd_filter_loss.q
konstantinb: I deliberately used a value that did not match the key to demonstrate that the filter was **not** eliminating values with incorrect key column values. I have simplified the test query, but left this situation in, as I believe it very clearly demonstrates the nature of the bug:
HIVE-2904 extra rows using pre-fix code

zabetak: Due to the bug the `t.key = '333' OR lv.col = '86'` predicate is removed completely. If we don't look at the plan (and just look at the results) how we can be sure that `t.key = '333'` is still taking effect after the fix if we don't have such values as part of the data.

konstantinb: I am sorry @zabetak , it has taken me some time to fully understand your point. The query was ok to show the problem with the old code, but it was not confirming accuracy of filter retention with the new code; I believe that my latest commit https://github.com/apache/hive/pull/6014/commits/ee0bff669b6393819a194ab9d3f5b26e794e620f addresses that

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/ASTConverter.java
konstantinb: > Concretely, I was thinking something like the following:
> 
> assuming that the code inside `createASTLateralView` that needs a schema can use directly `tableFunctionSource.schema`.
> 
> In this case, the "input" schema is that inside `tableFunctionSource` and the "output" schema is the one constructed in `s`.

@zabetak Thank you very much for this suggestion, it definitely makes perfect sense to better distinguish input & output schemas. I believe that the commit https://github.com/apache/hive/pull/6014/commits/f413e345205aab278f7d61a920444237bcb095c6 is in a very close match with your proposal. I would appreciate it if you can give it another look when you get a chance.

zabetak: @konstantinb I took the liberty to some further refactoring and cleanup in https://github.com/apache/hive/pull/6014/commits/2eaabefadb29dd24aa2a405262af6a91d55732b4 and  https://github.com/apache/hive/pull/6014/commits/829ecf9693f66b30b2e2ac64de04ca3bdb3bd304. If everything looks good to you then I think we are ready to merge it.

konstantinb:
> @konstantinb I am waiting a final confirmation from your side regarding the changes that I pushed but from my side everything looks good and this PR is ready to go in!

@zabetak thank you very much for the modifications. I was sure I had already removed the no longer used schema constructor; my apologies.

Your refactoring is almost a 1:1 match with an intermediate working variant I had; I was concerned that the signature change of createASTLateralView() plus an early return in convertSource() might be harder to follow and could be frowned upon. I am fully comfortable with your refactoring, thank you!

zabetak:
@konstantinb Many thanks for the PR and your thorough analysis and explanations here and under the JIRA ticket.

PR #6016, opened 2025-08-08

Neer393:

No it just adds a new method to get tables as object


Neer393:
As per suggestions from @vikramahuja1001 I have made changes to not add a new HMS API. In this way I am reducing the msc calls from one request per table to a few depending upon the batch size.
Please review it @deniskuzZ @abstractdog

PR #6017, opened 2025-08-08
HIVE-29126: Add missing equal sign when using JVM option

Chaho12:

Add missing `=` after --add-opens.
- https://issues.apache.org/jira/browse/HIVE-29126

Equals sign helps some tooling handle the option as a single argument, even if the JVM itself does not strictly require it for command-line usage. 

Keep the format same with other JVM options.

No

Not a critical issue right as of 4.1.0


zhangbutao (approved):
+1 LGTM

Thanks for the fix.

deniskuzZ:
restarted the build

ayushtkn (approved):
LGTM

PR #6020, opened 2025-08-08

Neer393:

No it just adds a new method to get tables as object

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
@@ -24,12 +24,10 @@ import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.hive.common.TableName;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.HiveMetaStoreClient;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.MetastoreTaskThread;
-import org.apache.hadoop.hive.metastore.api.GetTableRequest;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.metastore.txn.NoMutex;
 import org.apache.hadoop.hive.metastore.txn.TxnStore;
@@ -37,7 +35,6 @@ import org.apache.hadoop.hive.metastore.txn.TxnUtils;
 import org.apache.iceberg.ExpireSnapshots;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.mr.hive.IcebergTableUtil;
-import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -50,7 +47,7 @@ public class IcebergHouseKeeperService implements MetastoreTaskThread {
   private ExecutorService deleteExecutorService = null;
 
   // table cache to avoid making repeated requests for the same Iceberg tables more than once per day
-  private final Cache<TableName, Table> tableCache = Caffeine.newBuilder()
+  private final Cache<org.apache.hadoop.hive.metastore.api.Table, Table> tableCache = Caffeine.newBuilder()
       .maximumSize(1000)
       .expireAfterWrite(1, TimeUnit.DAYS)
       .build();
@@ -79,35 +76,19 @@ public class IcebergHouseKeeperService implements MetastoreTaskThread {
 
   private void expireTables(String catalogName, String dbPattern, String tablePattern) {
     try (IMetaStoreClient msc = new HiveMetaStoreClient(conf)) {
-      // TODO: HIVE-28952 – modify TableFetcher to return HMS Table API objects directly,
-      // avoiding the need for subsequent msc.getTable calls to fetch each matched table individually
-      List<TableName> tables = IcebergTableUtil.getTableFetcher(msc, catalogName, dbPattern, tablePattern).getTables();
-
+      List<org.apache.hadoop.hive.metastore.api.Table> tables =
+          IcebergTableUtil.getTableFetcher(msc, catalogName, dbPattern, tablePattern).getTables();
       LOG.debug("{} candidate tables found", tables.size());
-
-      for (TableName table : tables) {
-        try {
-          expireSnapshotsForTable(getIcebergTable(table, msc));
-        } catch (Exception e) {
-          LOG.error("Exception while running iceberg expiry service on catalog/db/table: {}/{}/{}",
-              catalogName, dbPattern, tablePattern, e);
-        }
+      for (org.apache.hadoop.hive.metastore.api.Table table : tables) {
+        expireSnapshotsForTable(getIcebergTable(table));
       }
     } catch (Exception e) {
       throw new RuntimeException("Error while getting tables from metastore", e);
     }
   }
 
-  private Table getIcebergTable(TableName tableName, IMetaStoreClient msc) {
-    return tableCache.get(tableName, key -> {
-      LOG.debug("Getting iceberg table from metastore as it's not present in table cache: {}", tableName);
-      GetTableRequest request = new GetTableRequest(tableName.getDb(), tableName.getTable());
-      try {
-        return IcebergTableUtil.getTable(conf, msc.getTable(request));
-      } catch (TException e) {
-        throw new RuntimeException(e);
-      }
-    });
+  private Table getIcebergTable(org.apache.hadoop.hive.metastore.api.Table table) {
+    return tableCache.get(table, key -> IcebergTableUtil.getTable(conf, table));
   }
 
   /**
--- a/standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
+++ b/standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
@@ -20,9 +20,11 @@ package org.apache.hadoop.hive.metastore.utils;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.hadoop.hive.common.TableName;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.apache.hadoop.hive.metastore.TableIterable;
 import org.apache.hadoop.hive.metastore.TableType;
 import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.Database;
+import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -90,7 +92,7 @@ public class TableFetcher {
     this.tableFilter = String.join(" and ", conditions);
   }
 
-  public List<TableName> getTables() throws Exception {
+  public List<TableName> getTableNames() throws Exception {
     List<TableName> candidates = new ArrayList<>();
 
     // if tableTypes is empty, then a list with single empty string has to specified to scan no tables.
@@ -113,6 +115,41 @@ public class TableFetcher {
       }
       List<String> tablesNames = client.listTableNamesByFilter(catalogName, db, tableFilter, -1);
       tablesNames.forEach(tablesName -> candidates.add(TableName.fromString(tablesName, catalogName, db)));
+    }
+    return candidates;
+  }
+
+  public List<Table> getTables() throws Exception {
+    List<Table> candidates = new ArrayList<>();
+
+    // if tableTypes is empty, then a list with single empty string has to specified to scan no tables.
+    if (tableTypes.isEmpty()) {
+      LOG.info("Table fetcher returns empty list as no table types specified");
+      return candidates;
+    }
+
+    List<String> databases = client.getDatabases(catalogName, dbPattern);
+
+    for (String db : databases) {
+      Database database = client.getDatabase(catalogName, db);
+      if (MetaStoreUtils.checkIfDbNeedsToBeSkipped(database)) {
+        LOG.debug("Skipping table under database: {}", db);
+        continue;
+      }
+      if (MetaStoreUtils.isDbBeingPlannedFailedOver(database)) {
+        LOG.info("Skipping table that belongs to database {} being failed over.", db);
+        continue;
+      }
+      List<String> tablesNames = client.listTableNamesByFilter(catalogName, db, tableFilter, -1);
+      int maxBatchSize = 300;
+      try {
+        maxBatchSize = Integer.parseInt(client.getConfigValue("hive.metastore.batch.retrieve.max", "300"));
+      } catch (Exception e) {
+        LOG.debug("Unable to read max batch size config, falling to default value");
+      }
+      for (Table table : new TableIterable(client, db, tablesNames, maxBatchSize)) {
+        candidates.add(table);
+      }
     }
     return candidates;
   }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
@@ -88,7 +88,7 @@ public class IcebergTableOptimizer extends TableOptimizer {
       Set<String> skipDBs, Set<String> skipTables) {
     Set<CompactionInfo> compactionTargets = Sets.newHashSet();
 
-    getTables().stream()
+    getTableNames().stream()
         .filter(table -> !skipDBs.contains(table.getDb()))
         .filter(table -> !skipTables.contains(table.getNotEmptyDbTable()))
         .map(table -> {
@@ -126,9 +126,9 @@ public class IcebergTableOptimizer extends TableOptimizer {
     return compactionTargets;
   }
 
-  private List<org.apache.hadoop.hive.common.TableName> getTables() {
+  private List<org.apache.hadoop.hive.common.TableName> getTableNames() {
     try {
-      return IcebergTableUtil.getTableFetcher(client, null, "*", null).getTables();
+      return IcebergTableUtil.getTableFetcher(client, null, "*", null).getTableNames();
     } catch (Exception e) {
       throw new RuntimeMetaException(e, "Error getting table names");
     }
--- a/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/metastore/task/TestIcebergHouseKeeperService.java
+++ b/iceberg/iceberg-handler/src/test/java/org/apache/iceberg/mr/hive/metastore/task/TestIcebergHouseKeeperService.java
@@ -23,7 +23,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
-import org.apache.hadoop.hive.common.TableName;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.TableType;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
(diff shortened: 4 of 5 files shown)


standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -117,6 +119,41 @@ public List<TableName> getTables() throws Exception {
...
+        continue;
+      }
+      if (MetaStoreUtils.isDbBeingPlannedFailedOver(database)) {
+        LOG.info("Skipping table that belongs to database {} being failed over.", db);
+        continue;
+      }
+      List<String> tablesNames = client.listTableNamesByFilter(catalogName, db, tableFilter, -1);
+      int maxBatchSize = 300;
+      try {
+        maxBatchSize = Integer.parseInt(client.getConfigValue("hive.metastore.batch.retrieve.max", "300"));
```
vikramahuja1001: Can you get it from a HiveConf object rather than doing a HMS call?

Neer393: Done. Grabbing it from HiveConf now.

Neer393:
Hi @vikramahuja1001 @deniskuzZ @abstractdog can we have this reviewed ?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -117,6 +119,35 @@ public List<TableName> getTables() throws Exception {
     return candidates;
   }
 
+  public List<Table> getTables(int maxBatchSize) throws Exception {
+    List<Table> candidates = new ArrayList<>();
```
vikramahuja1001, line 114: I think you can take the common code from this method and the getTableNames method into a common private method to avoid duplication or perhaps just try calling getTableNames method from getTable.

Neer393: The issue with this is in ```getTables()``` within the for loop I am iterating over all databases and for each database I am getting the names of all tables within that and for each db then I am getting those tables batch wise so there is some processing happening within getTables whereas that is not the case with ```getTableNames()``` .

Coming to your second point of calling getTableNames withing getTables will also not reduce redundancy as if you take a look at this line, you can see that I need to iterate over databases which would then require me to make a double msc call to ```client.getDatabases(catalogName, dbPattern)``` (once called by getTableNames() and once I will call within getTables()). So that would still increase the number of msc calls by one.
```
for (Table table : new TableIterable(client, db, tablesNames, maxBatchSize)) {
        candidates.add(table);
}
```

So I think we cannot extract them to a common method. Do let me know if you think otherwise :)

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -79,35 +76,20 @@ public void run() {
...
-      LOG.debug("Getting iceberg table from metastore as it's not present in table cache: {}", tableName);
-      GetTableRequest request = new GetTableRequest(tableName.getDb(), tableName.getTable());
-      try {
-        return IcebergTableUtil.getTable(conf, msc.getTable(request));
-      } catch (TException e) {
-        throw new RuntimeException(e);
-      }
-    });
+  private Table getIcebergTable(org.apache.hadoop.hive.metastore.api.Table table) {
+    return tableCache.get(table, key -> IcebergTableUtil.getTable(conf, table));
```
okumin: The overall table object might be overkill as a cache key. I'd also suspect the original `TableName` might not be sufficient because the table object might have been updated, or it might not be an Iceberg table already. Is there a suitable candidate as a key? What I proposed in #5882 is using a metadata location, i.e., the location of the `metadata.json`.

```
 CREATE EXTERNAL TABLE `test`(                      |
|   `id` int)                                        |
| ROW FORMAT SERDE                                   |
|   'org.apache.iceberg.mr.hive.HiveIcebergSerDe'    |
| STORED BY                                          |
|   'org.apache.iceberg.mr.hive.HiveIcebergStorageHandler'  |
|                                                    |
| LOCATION                                           |
|   'file:/opt/hive/data/warehouse/test'             |
| TBLPROPERTIES (                                    |
|   'TRANSLATED_TO_EXTERNAL'='TRUE',                 |
|   'bucketing_version'='2',                         |
|   'current-schema'='{"type":"struct","schema-id":0,"fields":[{"id":1,"name":"id","required":false,"type":"int"}]}',  |
|   'external.table.purge'='TRUE',                   |
|   'format-version'='2',                            |
|   'iceberg.orc.files.only'='false',                |
|   'metadata_location'='file:/opt/hive/data/warehouse/test/metadata/00000-165e08e2-9cea-47b1-9eef-45921f238fe2.metadata.json',  |
```

Neer393: @okumin even I agree that storing the whole table object as a key might be too much heavy for cache. 
Your proposed solution of using metadata location as the key is good and would handle cases what you mentioned

The reason I did this was, we wanted to reduce the number of msc calls made here in IcebergHouseKeeperService.java. In the case where the cache mapping is between table name and Iceberg table object, we need to call ```getTableObjectByName()``` for each table name which increases the number of msc calls because from tableName we get hmsTable (making msc call) and from that we get IcebergTable.

Now if we make the cache mapping of metadata location and IcebergTable
First of all fetching all metadatalocations would be required. 
Secondly in case the metadatalocation key is not present in cache, we would need a function like ```IcebergTableUtil.getTableByMetadataLocation()``` which I don't know how we will create Iceberg Table object from metadata location .

The point is it still won't reduce our msc calls.

Neer393: > I'd also suspect the original `TableName` might not be sufficient because the table object might have been updated, or it might not be an Iceberg table already

I have a solution to this. What if we keep the current mapping same but along with this, we add a new field which indicates the timestamp this was added to cache. 
So if we want to know if the Iceberg Table in cache is stale or not, what we would do is compare the timestamps i.e the lastDDLTimestamp of the table and the cache timestamp. If the lastDDLTimestamp(saved as ```transient_lastDdlTime``` in the table) is greater, we would not return the cache object but instead force fetch the new IcebergTable.

How does this sound @okumin ?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -117,6 +119,35 @@ public List<TableName> getTables() throws Exception {
...
+
+    // if tableTypes is empty, then a list with single empty string has to specified to scan no tables.
+    if (tableTypes.isEmpty()) {
+      LOG.info("Table fetcher returns empty list as no table types specified");
+      return candidates;
+    }
+
+    List<String> databases = client.getDatabases(catalogName, dbPattern);
+
+    for (String db : databases) {
```
deniskuzZ, line 124: please avoid code duplication

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
deniskuzZ: we don't need advanced caching here. we use iceberg table ref to call the expire snapshots. cache TTL is 1 day, do you some issue with that?

deniskuzZ: also PR description seems to be absolutely unrelated to caching strategy

Neer393: This PR has nothing to do with caching strategy

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
Neer393: I have explained in this comment why this cannot be done [https://github.com/apache/hive/pull/6020#discussion_r2265871821](https://github.com/apache/hive/pull/6020#discussion_r2265871821)

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
vikramahuja1001: `      Database database = client.getDatabase(catalogName, db);`
`      if (MetaStoreUtils.checkIfDbNeedsToBeSkipped(database)) {`
`        LOG.debug("Skipping table under database: {}", db);`
`        continue;`
`      }`
`      if (MetaStoreUtils.isDbBeingPlannedFailedOver(database)) {`
`        LOG.info("Skipping table that belongs to database {} being failed over.", db);`
`        continue;`
`      }`
`      List<String> tablesNames = client.listTableNamesByFilter(catalogName, db, tableFilter, -1);`

The above part can be moved to a common method and can be used for both getTables and getTableNames method.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
deniskuzZ: wdym it cannot be done? try Cursor and see

Neer393: Reworked it 👍

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
Neer393: Yeah did this
Thanks for the detailed help @vikramahuja1001

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
okumin: You're right. I didn't notice that the `table` object is needed anyway on a cache miss; I assumed we could create `org.apache.iceberg.Table` from a metadata.json.

```
tableCache.get(table, key -> IcebergTableUtil.getTable(conf, table))
```

Or we might replace it with like this to reduce the memory pressure and speed up comparisons. I don't care too much.

```
org.apache.hadoop.hive.metastore.api.Table table = ???
String(or something better) cacheKey = getCacheKey(table);
tableCache.get(cacheKey, ignored -> IcebergTableUtil.getTable(conf, table))
```

> we don't need advanced caching here. we use iceberg table ref to call the expire snapshots. cache TTL is 1 day, do you some issue with that?

I assume we should expire snapshots on as up-to-date `org.apache.iceberg.Table` as possible to include all metadata.json and avoid conflicts(as ExpireSnapshot updates the latest metadata, I assume an operation against an obsolete Iceberg table might cause one additional retry); It means the current one seems to be better than the previous one that uses an table identifier as a cache key.
https://github.com/apache/iceberg/blob/apache-iceberg-1.9.2/api/src/main/java/org/apache/iceberg/ExpireSnapshots.java#L31-L32

Neer393: @okumin Thanks for that suggestion. Reverted back to TableName what we currently have. Now the cache is as light as it was previously and hence resolving this conversation

Neer393:
@deniskuzZ @okumin all comments have been addressed and all checks have passed.
We are good to merge this 👍

vikramahuja1001:
+1 non binding

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -79,35 +77,21 @@ public void run() {
...
-      GetTableRequest request = new GetTableRequest(tableName.getDb(), tableName.getTable());
-      try {
-        return IcebergTableUtil.getTable(conf, msc.getTable(request));
-      } catch (TException e) {
-        throw new RuntimeException(e);
-      }
-    });
+  private Table getIcebergTable(org.apache.hadoop.hive.metastore.api.Table table) {
+    TableName tableName = TableName.fromString(table.getTableName(), table.getCatName(), table.getDbName());
+    return tableCache.get(tableName, key -> IcebergTableUtil.getTable(conf, table));
```
okumin, line 94: I guess table metadata 24 hours ago might be stale, and there is a chance to optimize. However, it is not a scope of this pull request.

Neer393: Sure I will create a ticket for that. Here is the ticket for it:- [https://issues.apache.org/jira/browse/HIVE-29144](https://issues.apache.org/jira/browse/HIVE-29144)
In the meantime, can you merge this ?

Neer393: @okumin can we merge this now ?

okumin: Yes, we can. JFYI: We have a rule to wait for 24 hours after +1
https://hive.apache.org/community/resources/howtocommit/#commit

okumin:
Merged. @Neer393 Thanks for your contribution! @vikramahuja1001 @deniskuzZ Thanks for your review!

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -102,21 +104,47 @@ public List<TableName> getTables() throws Exception {
...
+      List<String> tablesNames = getTableNamesForDatabase(catalogName, db);
```
deniskuzZ, line 125: @Neer393 I don't understand what have you optimized here. 
You are still doing multiple calls: 1 to get table names and another to get table objects. Why not get table objects directly?

Also, have you considered the memory impact when loading everything into the heap? You could have iterated over TableIterable instead.  I don't think that is a robust solution, it can potentially lead to OOM.
cc @dengzhhu653, @wecharyu

Neer393: The earlier implementation had one msc call for getting table names and then one msc call each for getting the HMS table object for each table name. 

The newer implementation reduces the msc calls in a way that one msc call is made for getting all table names and then using TableIterable, the number of msc calls for getting table objects becomes ```Number of tables / (BATCH_MAX_RETRIEVE config value [Default is 300])```

So in the older implementation ```number of msc calls = 1 + number of tables```
whereas in the newer implementation ```number of msc calls = 1 + (number of tables / [BATCH_MAX_RETRIEVE])```

Neer393: In my earlier implementation I had the same proposal of directly getting table objects where I had implemented direct HMS API endpoint like ```listTableNamesByFilter``` but the idea was dropped by @vikramahuja1001

deniskuzZ: in order to use batching, you need to have the table list to fetch - that's ok. However, instead of working with the batches, you load everything into memory. 
Could you refactor to use Iterable (i.e make getTables return  `Iterable<Table>`)?

Neer393: Okay so for this fix should I create a new JIRA or as I am working on [https://issues.apache.org/jira/browse/HIVE-28974](https://issues.apache.org/jira/browse/HIVE-28974) which is related to IcebergHouseKeeperService only should I attach the fix in this JIRA ?
Whatever you suggest is fine to me

wecharyu: +1 Use the `TableIterator` is more reasonable to avoid possible OOM.

okumin: Let me summarize the points.
- This PR would reduce # of API calls from O(num-tables) to O(num-tables / batch size). It's neat 👍
- This PR would require the O(num-tables * tbl-size) space. We'd like to reduce it
- Regardless of this PR, we retain the O(num-tables * length-of-table-name) space. Is it OK or should we optimize it?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -79,35 +77,21 @@ public void run() {
...
-
-      for (TableName table : tables) {
-        try {
-          expireSnapshotsForTable(getIcebergTable(table, msc));
-        } catch (Exception e) {
-          LOG.error("Exception while running iceberg expiry service on catalog/db/table: {}/{}/{}",
-              catalogName, dbPattern, tablePattern, e);
-        }
+      for (org.apache.hadoop.hive.metastore.api.Table table : tables) {
+        expireSnapshotsForTable(getIcebergTable(table));
```
okumin, line 85: I recalled that we would like to retain the try-catch. We intentionally added it to avoid skipping everything when a single expiration fails.
See also: https://github.com/apache/hive/pull/5786#discussion_r2120885647

okumin:
@Neer393 @deniskuzZ I created a revert PR because we found two issues to be discussed.
https://github.com/apache/hive/pull/6033

Neer393:
Okay. In that case let me close the redundant JIRA that I created for fixing this. 
We will now ship this once and for all under HIVE-28952 only

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
Neer393: AFAIK, it's a tradeoff between the number of msc calls and the space. 
As if we try to decrease the number of tables stored in memory, we would increase the number of msc calls and then there would be no point of this JIRA. 
Please correct me if I am wrong but this is what I think

deniskuzZ: My concern was related to the fetch logic where we load all Hive table objects into the memory instead of using the batch iterator.

We also make O(num-tables) calls to load an Iceberg table. Can we optimize here? Then we put those into a separate cache. Maybe iwe could use CachingCatalog instead ?
````
tableCache.get(tableName, key -> IcebergTableUtil.getTable(conf, table))
````

deniskuzZ:
> Okay so for this fix should I create a new JIRA or as I am working on https://issues.apache.org/jira/browse/HIVE-28974 which is related to IcebergHouseKeeperService only should I attach the fix in this JIRA ?

@Neer393 [HIVE-28952](https://issues.apache.org/jira/browse/HIVE-28952) addendum should be OK. 
1 ask: can we use `TableIterator` in `IcebergTableOptimizer` as well?

Neer393:
> > Okay so for this fix should I create a new JIRA or as I am working on https://issues.apache.org/jira/browse/HIVE-28974 which is related to IcebergHouseKeeperService only should I attach the fix in this JIRA ?
> 
> @Neer393 [HIVE-28952](https://issues.apache.org/jira/browse/HIVE-28952) addendum should be OK. 1 ask: can we use `TableIterator` in `IcebergTableOptimizer` as well?

Okay in that case @okumin please do not revert the PR. As per @deniskuzZ I would put an addendum to it.

@deniskuzZ using `TableIterator` in `IcebergTableOptimizer` for `getTableNames()` is not a big deal. I mean we can add it but the question is do we want to as we added TableIterator as a single list of HMS API Table objects would be a burden as table objects are big in size but TableName objects are very small with only 3 strings for tablename, dbname and catalogname. 

So I don't think we require TableIterator as such for TableOptimizer but if you say so I can add it. I say the call is your's.

deniskuzZ:
@Neer393 `IcebergTableOptimizer` first retrieves the list of table names, then for each name it loads the corresponding Hive table object (O(num-tables) `getHiveTable()` call), followed by loading the Iceberg tables (O(num-tables)).
This is the same suboptimal implementation as in `IcebergHouseKeeperService`

Neer393:
> @Neer393 `IcebergTableOptimizer` first retrieves the list of table names, then for each name it loads the corresponding Hive table object (O(num-tables) `getHiveTable()` call), followed by loading the Iceberg tables (O(num-tables)). This is the same suboptimal implementation as in `IcebergHouseKeeperService`

Oh okay. In that case yes.
Will make changes there as well

PR #6021, opened 2025-08-08
HIVE-29128 : Allow show tables `format`="iceberg" command

vikramahuja1001:


vikramahuja1001:
@deniskuzZ , @chinnaraolalam  could you please take a look at this PR?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/info/show/tables/ShowTablesAnalyzer.java
```diff
@@ -53,17 +56,29 @@ public void analyzeInternal(ASTNode root) throws SemanticException {
     String tableNames = null;
     TableType tableTypeFilter = null;
     boolean isExtended = false;
```
Aggarwal-Raghav, line 58: nit: better to use `isIceberg`, to have uniform variable naming as above line `isExtended`

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -1974,6 +1996,19 @@ public List<String> getTablesByType(String dbName, String pattern, TableType typ
 
     try {
       List<String> result;
+      if (ifIceberg) {
+        if (type != null) {
+          throw new HiveException("Iceberg Tables by default is External Table");
+        } else {
+          if (pattern != null) {
```
Aggarwal-Raghav: nit: This looks cleaner.
```
if (ifIceberg) {
        if (type != null) {
          throw new HiveException("Iceberg tables are always of type EXTERNAL_TABLE");
        }
        String icebergPattern = (pattern != null) ? pattern : ".*";
        List<TableName> icebergTables = getIcebergTables(getMSC(), null, dbName, icebergPattern).getTables();
        return icebergTables.stream()
          .map(TableName::getTable)
      }
```

Aggarwal-Raghav:
Just thinking out loud: As the idea is to only do this for iceberg tables! Is it feasible to do something like this?
```show iceberg tables;``` instead of ```show tables where `format`='iceberg'```

**Comment by vikramahuja1001** [comment]
> Just thinking out loud: As the idea is to only do this for iceberg tables! Is it feasible to do something like this? `show iceberg tables;` instead of `` show tables where `format`='iceberg' ``

Yes, it can be done. But to do this, Iceberg has to be exposed as a keyword in Hive which it is not atm. That would require changes in the alter and stored_by grammar as well. I purposely kept it this way to not make any changes on those layers.  But yeah, i can make thee changes if there is a consensus here.

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java [hunk 3]
```diff
@@ -1974,6 +1996,15 @@ public List<String> getTablesByType(String dbName, String pattern, TableType typ
 
     try {
       List<String> result;
+      if (isIceberg) {
+        if (type != null) {
+          throw new HiveException("Iceberg Tables by default is External Table");
```
Indhumathi27, line 2001: should this be an exception in this case? 
1. can we show it like below?
```
# Table Name        	Table Type          
test_iceberg      	EXTERNAL_TYPE  
```
2. Or you could handle it in analyse step itself -- In ShowTablesAnalyzer

PR #6022, opened 2025-08-12
HIVE-29035: Fixing cache handling for REST catalog

henrib:
Fixing the catalog cache behavior (take 2) by checking that a cached table metadata location is the same as the latest known metadata location for that table in HMS database to avoid returning stale entries.

To ensure the cache does not return stale entries using HA or when direct HMS calls are concurrent with REST Catalog usage.

No

Unit test


henrib:
Continued from reverted https://github.com/apache/hive/pull/5882 . @deniskuzZ @okumin

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -434,6 +459,7 @@ public boolean tableExists(TableIdentifier identifier) {
     }
   }
 
+
```
okumin, line 462:
```suggestion
```

okumin (commented):
@henrib
I am experimentally adding caching Table Metadata on apache/iceberg so that the community can reuse the implementation. Do you think if it's reasonable and if Hive REST can cache them through HiveCatalog?
https://github.com/okumin/iceberg/commit/86287958d2365a120b76a348deea51e10b77023f

It is inspired by https://github.com/apache/iceberg/pull/4518

I expect we might potentially get the following feedback.
- Iceberg community rejects it because Table Metadata is not cacheable for our unknown reasons -> In this case, we will likely drop HIVE-29035
- Iceberg community does not want to merge it because RoI is not good -> In this case, we will implement it on our side
- Iceberg community accepts the patch -> Can we implement HIVE-29035 on top of the patch?

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
```diff
@@ -408,23 +410,46 @@ private void validateTableIsIcebergTableOrView(
...
+  /**
+   * Check whether table exists and return its current metadata location.
+   *
+   * <p>Note: If a hive table with the same identifier exists in catalog, this method will return
+   * {@code null}.
+   *
+   * @param identifier a table identifier
+   * @return the location of the table if it exists, null otherwise
+   */
+  public String getTableMetadataLocation(TableIdentifier identifier) {
```
deniskuzZ, line 425: please do not change anything in iceberg-catalog or submit an iceberg PR

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -35,24 +43,22 @@
...
+import org.slf4j.LoggerFactory;
 
 
 /**
  * Class that wraps an Iceberg Catalog to cache tables.
  */
 public class HMSCachingCatalog extends CachingCatalog implements SupportsNamespaces, ViewCatalog {
-  private final HiveCatalog hiveCatalog;
+  private static final Logger LOG = LoggerFactory.getLogger(HMSCachingCatalog.class);
+  protected final HiveCatalog hiveCatalog;
```
deniskuzZ, line 53: why change to protected?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -63,6 +69,96 @@ public List<Namespace> listNamespaces(Namespace nmspc) throws NoSuchNamespaceExc
     return hiveCatalog.listNamespaces(nmspc);
   }
 
+  protected void cacheInvalidateInc(TableIdentifier tid) {
+    // This method is intentionally left empty. It can be overridden in subclasses if needed.
```
deniskuzZ, line 76: what subclasses? please do not over-engineer

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -63,6 +69,96 @@ public List<Namespace> listNamespaces(Namespace nmspc) throws NoSuchNamespaceExc
...
+   * Gets the metadata file location of a table.
+   *
+   * @param table the table
+   * @return the location of the metadata file, or null if the table does not have a location
+   */
+  protected static String getMetadataLocation(final Table table) {
+    if (table instanceof HasTableOperations tableOps) {
+      final TableOperations ops = tableOps.operations();
+      final TableMetadata meta;
+      if (ops != null && (meta = ops.current()) != null) {
```
deniskuzZ: it can't be null, check iceberg code. I don't think we need method for that
````
if (table instanceof HasTableOperations tableOp) {
  metadataLocation = tableOp.operations().current().metadataFileLocation();
}
````

File: standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java [hunk 6]
```diff
@@ -63,6 +69,96 @@ public List<Namespace> listNamespaces(Namespace nmspc) throws NoSuchNamespaceExc
...
+        return meta.metadataFileLocation();
+      }
+    }
+    return null;
+  }
+
+  @Override
+  public Table loadTable(final TableIdentifier identifier) {
+    final Cache<TableIdentifier, Table> cache = this.tableCache;
```
deniskuzZ, line 213: what is the point in those local vars?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
```diff
@@ -63,6 +69,96 @@ public List<Namespace> listNamespaces(Namespace nmspc) throws NoSuchNamespaceExc
...
+    final HiveCatalog catalog = this.hiveCatalog;
+    final TableIdentifier canonicalized = identifier.toLowerCase();
+    Table cachedTable = cache.getIfPresent(canonicalized);
+    if (cachedTable != null) {
+      final String location = catalog.getTableMetadataLocation(canonicalized);
```
deniskuzZ, line 218: can't you use `catalog.loadTable(canonicalized)` and get the location instead of adding new method?

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/extension/HMSTestCatalogFactory.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.servlet.http.HttpServlet;
+import java.lang.ref.Reference;
+import java.lang.ref.SoftReference;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Factory for creating a test catalog that caches the last created catalog.
+ * <p>This class is used in tests to verify the behavior of the caching catalog.</p>
+ */
+public class HMSTestCatalogFactory extends HMSCatalogFactory {
```
deniskuzZ, line 38: why do you need TestCatalogFactory? use original Factory class

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/extension/HMSTestCachingCatalog.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A test class for HMSCachingCatalog that tracks cache metrics for test verifications.
+ * It extends HMSCachingCatalog and overrides methods to increment counters for cache hits,
+ * misses, invalidations, and loads.
+ */
+public class HMSTestCachingCatalog extends HMSCachingCatalog {
```
deniskuzZ, line 34: same here. we are not testing test classes

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: To write tests against a derived version (see test/java/org/apache/iceberg/rest/TestServerCatalogCache.java ).

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: Usual (old?) pattern of using locals instead of dereferencing members (marginally faster).

deniskuzZ: `tableCache` field is final so there is no point in defining local vars

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/extension/HMSTestCatalogFactory.java
henrib: So I can get access to the latest cache created and verify it performs as intended.

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/extension/HMSTestCachingCatalog.java
henrib: The methods being tested are the inherited ones. Having specific classes for tests is no different in effect that what mocking does.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: Wouldn't this imply the cache is rematerializing a table object to verify it might have avoided it ?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
henrib: The test-harnessed version of the cache uses those callback to check the feature.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
henrib: The cache in HIVE-29035 is limited to serving loadTable() for REST and resides server-side; the Table objects it serves are marshaled by to a client so there is no 'external' instance sharing. It is dependent upon HMS being the actual catalog implementation to acquire the latest known metadata location for a given table. This makes this PR pretty much tied to Hive; there is no need to involve Iceberg.
If/when the Iceberg community accepts the TableMetadata caching you propose, we can then assess what it means for this cache implementation.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
henrib: Iceberg PR: https://github.com/apache/iceberg/issues/13800

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
okumin: Without using a REST catalog, a client retrieves table metadata through XYZCatalog -> TableMetadataParser -> S3/HDFS/etc. With a REST catalog, a client does it through RESTCatalog, where REST API(in our case, the servlet) serves metadata via XYZCatalog(in our case, HiveCatalog or HMSCachingCatalog) -> TableMetadataParser -> S3/HDFS/etc. So, TableMetadataParser might be a better place to maintain. It can support our use case, and we can remove HMSCachingCatalog, which uses CachingCatalog introduced for client-side caching and utilized only in SparkCatalog and FlinkCatalog.
If my idea has some defects, I probably won't send the patch

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
deniskuzZ: thanks!

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCachingCatalog.java
deniskuzZ: sorry, I might be missing something. does the catalog.loadTable(canonicalized) use the cache?

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/extension/HMSTestCatalogFactory.java
deniskuzZ: can we keep the latest ref somewhere in a test and not it overloaded test class factory?

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/extension/HMSTestCachingCatalog.java
deniskuzZ: i think mocking would be a cleaner approach

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveCatalog.java
deniskuzZ: > Without using a REST catalog, a client retrieves table metadata through XYZCatalog -> TableMetadataParser -> S3/HDFS/etc. With a REST catalog, a client does it through RESTCatalog, where REST API(in our case, the servlet) serves metadata via XYZCatalog(in our case, HiveCatalog or HMSCachingCatalog) -> TableMetadataParser -> S3/HDFS/etc. So, TableMetadataParser might be a better place to maintain. It can support our use case, and we can remove HMSCachingCatalog, which uses CachingCatalog introduced for client-side caching and utilized only in SparkCatalog and FlinkCatalog. If my idea has some defects, I probably won't send the patch

@okumin are you saying `TableMetadataParser` is the common place for both direct and REST catalog invocations?

okumin: Yes, I am.

Using HiveCatalog directly, everything works on the client side as follows.
image

Using REST, some steps will be delegated to the server-side. In either case, TableMetadataParser would interact with an object storage.
image

deniskuzZ: some notes from iceberg dev lists:
````
Caching Metadata on the Client Side: Reloading table metadata for a
particular snapshot could leverage the ETag mechanism to reduce the amount
of network traffic.
````

okumin: I asked the Iceberg community with the sample PR.
https://github.com/apache/iceberg/pull/14137

deniskuzZ: @okumin, do you know if https://github.com/apache/iceberg/commit/72d5fd66c527c27c79880e051b7eaf46deb2fed6 solves the same problem?

PR #6023, opened 2025-08-12
HIVE-19996: Beeline performance poor with drivers having slow DatabaseMetaData.getPrimaryKeys impl

InvisibleProgrammer:
DatabaseMetaData.getPrimaryKeys can query the table metadata once for each column. It can cause significant slowness. 
Now we can query that info once, during the initialization of the Rows object.

To fix the performance issue.

No

- Added extra tests for TestTableOutputFormat
- Manual test on a test table

### DIFF
--- a/beeline/src/java/org/apache/hive/beeline/Rows.java
+++ b/beeline/src/java/org/apache/hive/beeline/Rows.java
@@ -29,9 +29,13 @@ import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 
 import org.apache.hadoop.hive.common.cli.EscapeCRLFHelper;
 
@@ -42,7 +46,8 @@ import org.apache.hadoop.hive.common.cli.EscapeCRLFHelper;
 abstract class Rows implements Iterator {
   protected final BeeLine beeLine;
   final ResultSetMetaData rsMeta;
-  final Boolean[] primaryKeys;
+  final boolean[] primaryKeyColumns;
+  boolean isPrimaryKeyColumnsInitialized;
   final NumberFormat numberFormat;
   private boolean convertBinaryArrayToString;
   private final String nullStr;
@@ -51,14 +56,15 @@ abstract class Rows implements Iterator {
     this.beeLine = beeLine;
     nullStr = beeLine.getOpts().getNullString();
     rsMeta = rs.getMetaData();
-    int count = rsMeta.getColumnCount();
-    primaryKeys = new Boolean[count];
     if (beeLine.getOpts().getNumberFormat().equals("default")) {
       numberFormat = null;
     } else {
       numberFormat = new DecimalFormat(beeLine.getOpts().getNumberFormat());
     }
     this.convertBinaryArrayToString = beeLine.getOpts().getConvertBinaryArrayToString();
+
+    int count = this.rsMeta.getColumnCount();
+    primaryKeyColumns = new boolean[count];
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


beeline/src/java/org/apache/hive/beeline/Rows.java
```diff
@@ -51,14 +55,14 @@ abstract class Rows implements Iterator {
...
-    int count = rsMeta.getColumnCount();
-    primaryKeys = new Boolean[count];
     if (beeLine.getOpts().getNumberFormat().equals("default")) {
       numberFormat = null;
     } else {
       numberFormat = new DecimalFormat(beeLine.getOpts().getNumberFormat());
     }
     this.convertBinaryArrayToString = beeLine.getOpts().getConvertBinaryArrayToString();
+
+    primaryKeys = initializePrimaryKeyMetadata();
```
kasakrisz: IIUC primary key info is queried every time a new `Rows` instance is created. Even if coloring is not needed. Could you please explore if this can be improved?

beeline/src/java/org/apache/hive/beeline/Rows.java
```diff
@@ -78,37 +82,43 @@ public void remove() {
...
-                primaryKeys[col] = Boolean.TRUE;
-                break;
-              }
-            }
-          } finally {
-            pks.close();
+  private boolean[] initializePrimaryKeyMetadata() throws SQLException {
+
+    int count = this.rsMeta.getColumnCount();
+    boolean[] primaryKeys = new boolean[count];
```
kasakrisz: Should this be initialized to all element false ?

beeline/src/java/org/apache/hive/beeline/Rows.java
```diff
@@ -78,37 +82,43 @@ public void remove() {
...
+    for (int i = 0; i < count; i++) {
+      String table = rsMeta.getTableName(i + 1);
+      String column = rsMeta.getColumnName(i + 1);
+
+      if (table == null || table.isEmpty() || column == null || column.isEmpty()) {
+        continue;
+      }
+
+      if (!tablePrimaryKeys.containsKey(table)) {
+        try (ResultSet pks = beeLine.getDatabaseConnection().getDatabaseMetaData().getPrimaryKeys(
```
kasakrisz: The original code caught `SQLException` and set the `primaryKeys[i]` to `false`. Was this intentionally removed?

beeline/src/java/org/apache/hive/beeline/Rows.java
InvisibleProgrammer: I don't see a use case when we create multiple Rows instances for a single ResultSet: https://github.com/apache/hive/blob/master/beeline/src/java/org/apache/hive/beeline/BeeLine.java#L2426

Is there any special use case in your mind that I don't know about? 

About querying it when coloring is disabled: it is an interesting question: The current structure of the code behaves in a way that the Rows object doesn't know if it will be used in a coloring or non-coloring use case. Also doesn't know if coloring is exists at all. 
The best alternative that I can think about is to fill the information only if the `isPrimaryKeyCol` method is called. But it can also make the code a little bit more complex if we introduce a lazy-loading mechanism. 
I think handling the situation only once and at the beginning of the class can be a logical place between doing hundreds of unnecessary calls and zero calls at all - and also keeping the complexity low. 

What are your opinions?

beeline/src/java/org/apache/hive/beeline/Rows.java
InvisibleProgrammer: It does.

beeline/src/java/org/apache/hive/beeline/Rows.java
InvisibleProgrammer: Nice catch, I fixed it: Now it follows the original practice: if an exception occurs, we just accept that we have no primary key information.

beeline/src/java/org/apache/hive/beeline/Rows.java
InvisibleProgrammer: Refactored to do lazy initialization.

kasakrisz (approved):

zhangbutao:
This fix may break the backward compatibility of beeline. I created a ticket for this issue HIVE-29213.

PR #6024, opened 2025-08-13
HIVE-29138: Authentication failure when connecting to Kerberized ZooK…

dengzhhu653:
…eeper

### DIFF
--- a/jdbc/pom.xml
+++ b/jdbc/pom.xml
@@ -241,7 +241,7 @@
                         <include>org/apache/hadoop/security/token/**</include>
                         <include>org/apache/hadoop/util/*</include>
                         <include>org/apache/hadoop/net/*</include>
-                        <include>org/apache/hadoop/io/*</include>
+                        <include>org/apache/hadoop/io/**</include>
                         <include>org/apache/hadoop/HadoopIllegalArgumentException.class</include>
                       </includes>
                     </filter>


jdbc/src/java/org/apache/hive/jdbc/Utils.java
```diff
@@ -145,6 +145,7 @@ public static class JdbcConnectionParams {
     public static final String SERVICE_DISCOVERY_MODE_ZOOKEEPER_HA = "zooKeeperHA";
     public static final String ZOOKEEPER_NAMESPACE = "zooKeeperNamespace";
     public static final String ZOOKEEPER_SSL_ENABLE = "zooKeeperSSLEnable";
+    public static final String ZOOKEEPER_SASL_CLIENT = "zookeeper.sasl.client";
```
dengzhhu653: need to check if we really need this new jdbc param

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/SecurityUtils.java
```diff
@@ -131,14 +166,22 @@ public AppConfigurationEntry[] getAppConfigurationEntry(String appName) {
...
+          }
         } else {
           krbOptions.put("doNotPrompt", "true");
-          krbOptions.put("storeKey", "true");
-          krbOptions.put("useKeyTab", "true");
-          krbOptions.put("keyTab", keyTabFile);
+          krbOptions.put("useTicketCache", "true");
+          if (keyTabFile != null) {
+            krbOptions.put("storeKey", "true");
+            krbOptions.put("useKeyTab", "true");
```
dengzhhu653, line 185: if both `useTicketCache` and `useKeyTab` are true, how they work together?

jdbc/src/java/org/apache/hive/jdbc/ZooKeeperHiveClientHelper.java
```diff
@@ -128,6 +126,13 @@ private static CuratorFramework getZkClient(JdbcConnectionParams connParams) thr
                 connParams.getZookeeperTrustStoreLocation(),
                 connParams.getZookeeperTrustStorePassword(), connParams.getZookeeperTrustStoreType()))
             .build();
+
+    Map<String, String> sessionVars = connParams.getSessionVars();
+    // If the client is requesting the Kerberos, then the ZooKeeper is mostly Kerberos-secured
+    if (sessionVars.containsKey(JdbcConnectionParams.AUTH_PRINCIPAL) &&
+        Boolean.parseBoolean(sessionVars.getOrDefault(HiveConf.ConfVars.HIVE_ZOOKEEPER_USE_KERBEROS.varname, "true"))) {
```
deniskuzZ: Wonder if we should get defaults from Hive conf?

jdbc/src/java/org/apache/hive/jdbc/ZooKeeperHiveClientHelper.java
```diff
@@ -195,38 +200,33 @@ static void configureConnParams(JdbcConnectionParams connParams) throws ZooKeepe
...
   private static void configureConnParamsHA(JdbcConnectionParams connParams) throws ZooKeeperHiveClientException {
     try {
+
       Configuration registryConf = new Configuration();
       registryConf.set(HiveConf.ConfVars.HIVE_ZOOKEEPER_QUORUM.varname, connParams.getZooKeeperEnsemble());
       registryConf.set(HiveConf.ConfVars.HIVE_SERVER2_ACTIVE_PASSIVE_HA_REGISTRY_NAMESPACE.varname,
         getZooKeeperNamespace(connParams));
+      Map<String, String> sessionVars = connParams.getSessionVars();
+      registryConf.setBoolean(HiveConf.ConfVars.HIVE_ZOOKEEPER_USE_KERBEROS.varname,
+          sessionVars.containsKey(JdbcConnectionParams.AUTH_PRINCIPAL) &&
```
deniskuzZ: can we extract this condition into the helper method and reuse?

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -187,7 +188,8 @@ public ZkRegistryBase(String instanceName, Configuration conf, String rootNs, St
 
   public static String getRootNamespace(Configuration conf, String userProvidedNamespace,
       String defaultNamespacePrefix) {
-    final boolean isSecure = ZookeeperUtils.isKerberosEnabled(conf);
+    final boolean isSecure = UserGroupInformation.isSecurityEnabled() &&
```
deniskuzZ: why not keep this logic in util class instead copy?

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
```diff
@@ -670,10 +674,19 @@ protected final synchronized PathChildrenCache ensureInstancesCache(
 
   public void start() throws IOException {
     if (zooKeeperClient != null) {
-      String principal = ZookeeperUtils.setupZookeeperAuth(
-          conf, saslLoginContextName, zkPrincipal, zkKeytab);
-      if (principal != null) {
-        userNameFromPrincipal = LlapUtil.getUserNameFromPrincipal(principal);
+      if (UserGroupInformation.isSecurityEnabled() &&
+          HiveConf.getBoolVar(conf, HiveConf.ConfVars.HIVE_ZOOKEEPER_USE_KERBEROS)) {
+        if (saslLoginContextName != null) {
```
deniskuzZ, line 675: why not move this inside of `setZookeeperClientKerberosJaasConfig` ?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -682,18 +683,17 @@ private void addConfsToPublish(HiveConf hiveConf, Map<String, String> confsToPub
...
-      String keyTabFile = hiveConf.getVar(ConfVars.HIVE_SERVER2_KERBEROS_KEYTAB);
-      if (keyTabFile.isEmpty()) {
-        throw new IOException("HiveServer2 Kerberos keytab is empty");
+  private static void setUpZooKeeperAuth(HiveConf hiveConf) {
+    try {
+      if (AuthType.isKerberosAuthMode(hiveConf) &&
+          StringUtils.isNotEmpty(HiveConf.getVar(hiveConf, HiveConf.ConfVars.HIVE_ZOOKEEPER_QUORUM)) &&
+          HiveConf.getBoolVar(hiveConf, HiveConf.ConfVars.HIVE_ZOOKEEPER_USE_KERBEROS)) {
+        // Install the JAAS Configuration for the runtime
+        Utils.setZookeeperClientKerberosJaasConfig(hiveConf.getVar(ConfVars.HIVE_SERVER2_KERBEROS_PRINCIPAL),
```
deniskuzZ: can we extract principal and keytab into local var?

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -682,18 +683,17 @@ private void addConfsToPublish(HiveConf hiveConf, Map<String, String> confsToPub
...
-      String principal = hiveConf.getVar(ConfVars.HIVE_SERVER2_KERBEROS_PRINCIPAL);
-      if (principal.isEmpty()) {
-        throw new IOException("HiveServer2 Kerberos principal is empty");
-      }
-      String keyTabFile = hiveConf.getVar(ConfVars.HIVE_SERVER2_KERBEROS_KEYTAB);
-      if (keyTabFile.isEmpty()) {
-        throw new IOException("HiveServer2 Kerberos keytab is empty");
+  private static void setUpZooKeeperAuth(HiveConf hiveConf) {
+    try {
+      if (AuthType.isKerberosAuthMode(hiveConf) &&
```
deniskuzZ, line 688: why not keep repeating condition logic in util class or helper method?
````
boolean isSecure =
        AuthType.isKerberosAuthMode(hiveConf) &&
        HiveConf.getBoolVar(hiveConf, HiveConf.ConfVars.HIVE_ZOOKEEPER_USE_KERBEROS)
````

File: service/src/java/org/apache/hive/service/server/KillQueryZookeeperManager.java [hunk 9]
```diff
@@ -101,11 +98,11 @@ public synchronized void start() {
       throw new ServiceException("Failed start zookeeperClient in KillQueryZookeeperManager");
     }
     try {
-      ZookeeperUtils.setupZookeeperAuth(this.getHiveConf(), SASL_LOGIN_CONTEXT_NAME, zkPrincipal, zkKeytab);
       zooKeeperClient.start();
       try {
         zooKeeperClient.create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT).forPath("/" + zkNameSpace);
-        if (ZookeeperUtils.isKerberosEnabled(conf)) {
+        if (UserGroupInformation.isSecurityEnabled() &&
```
deniskuzZ, line 104: could be reused from util

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
```diff
@@ -55,6 +56,21 @@ public class Utils {
 
   public static final String DISTCP_OPTIONS_PREFIX = "distcp.options.";
 
+
+  private static Method getKeytab = null;
+  static {
```
deniskuzZ: new line

can we reuse logic from SecurityUtils in metastore common?

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
```diff
@@ -55,6 +56,21 @@ public class Utils {
...
+    Class<?> clz = UserGroupInformation.class;
+
+    try {
+      getKeytab = clz.getDeclaredMethod("getKeytab");
+      getKeytab.setAccessible(true);
+    } catch(NoSuchMethodException nme) {
```
deniskuzZ: nit: space

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
```diff
@@ -55,6 +56,21 @@ public class Utils {
...
+      LOG.warn("Cannot find private method \"getKeytab\" in class:" +
```
deniskuzZ: do we really need this safegard?

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
```diff
@@ -73,15 +89,46 @@ public static UserGroupInformation getUGI() throws LoginException, IOException {
     return UserGroupInformation.getCurrentUser();
   }
 
+  /**
+   * This is used for Jdbc client to authenticate with zk, where we don't know the exact principal.
+   * @throws Exception
+   */
+  public static void setZookeeperClientKerberosJaasConfig() throws IOException {
```
deniskuzZ: should we pass UserGroupInformation as param?

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
```diff
@@ -117,14 +164,23 @@ public AppConfigurationEntry[] getAppConfigurationEntry(String appName) {
         Map<String, String> krbOptions = new HashMap<String, String>();
         if (IBM_JAVA) {
           krbOptions.put("credsType", "both");
-          krbOptions.put("useKeytab", keyTabFile);
+          if (keyTabFile != null) {
+            krbOptions.put("useKeytab", keyTabFile);
```
deniskuzZ: should it be "keyTab"?
krbOptions.put("useKeyTab", "true"); is missing

jdbc/src/java/org/apache/hive/jdbc/ZooKeeperHiveClientHelper.java
dengzhhu653: done

llap-client/src/java/org/apache/hadoop/hive/registry/impl/ZkRegistryBase.java
dengzhhu653: This llap `ZookeeperUtils` is dropped, as his work is taken over by the `SecurityUtils` which can be shared across the component. 
Created a static util method to reuse this logic

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
dengzhhu653: We can get the current UGI by UserGroupInformation.getLoginUser

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
dengzhhu653: this is for `IBM_JAVA`, it's using the `useKeyTab` to specify the keytab: https://www.ibm.com/docs/en/semeru-runtime-ce-z/11.0.0?topic=differences-jaas-options-kerberos-login

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
dengzhhu653: Just in case, this is called from static block, we need to catch this `NoSuchMethodException` explicitly.

service/src/java/org/apache/hive/service/server/HiveServer2.java
```diff
@@ -682,18 +683,18 @@ private void addConfsToPublish(HiveConf hiveConf, Map<String, String> confsToPub
...
+          StringUtils.isNotEmpty(HiveConf.getVar(hiveConf, HiveConf.ConfVars.HIVE_ZOOKEEPER_QUORUM)) &&
+          HiveConf.getBoolVar(hiveConf, HiveConf.ConfVars.HIVE_ZOOKEEPER_USE_KERBEROS)) {
+        // Install the JAAS Configuration for the runtime
+        String principal = hiveConf.getVar(ConfVars.HIVE_SERVER2_KERBEROS_PRINCIPAL);
```
deniskuzZ, line 692: should we keep `if (principal.isEmpty())` check? similar for keytab

shims/common/src/main/java/org/apache/hadoop/hive/shims/Utils.java
deniskuzZ: the naming is a bit confusing — in one place it’s a string, while in another it’s a boolean. maybe we should add a comment with link to https://www.ibm.com/docs/en/semeru-runtime-ce-z/11.0.0?topic=differences-jaas-options-kerberos-login ?
````
    krbOptions.put("useKeytab", keyTabFile);
} else {
  if (keyTabFile != null) {
    krbOptions.put("useKeyTab", "true");
````

service/src/java/org/apache/hive/service/server/HiveServer2.java
dengzhhu653: Here the HS2 is kerberized, neither principal nor keytab should be empty, otherwise the HS2 will fail to start

service/src/java/org/apache/hive/service/server/HiveServer2.java
dengzhhu653: AuthType.isKerberosAuthMode(hiveConf) is inferred from HS2's property:`hive.server2.authentication`, the client such as jdbc might not configure this property

PR #6027, opened 2025-08-14
HIVE-29122: Vectorization - Support IGNORE NULLS for FIRST_VALUE and …

soumyakanti3578:
…LAST_VALUE [ADDENDUM]

This is an addendum to https://issues.apache.org/jira/browse/HIVE-29122 and PR https://github.com/apache/hive/pull/6009

This is needed to improve tests for partitioned queries. For example, 
```
SELECT id, int_col,
  FIRST_VALUE(int_col IGNORE NULLS) OVER(PARTITION BY id ORDER BY id) AS first_int,
  LAST_VALUE(int_col IGNORE NULLS) OVER(PARTITION BY id ORDER BY id) AS last_int
FROM window_int_test
```
was changed to 
```
SELECT id, int_col,
  FIRST_VALUE(int_col IGNORE NULLS) OVER(PARTITION BY id ORDER BY id, int_col) AS first_int,
  LAST_VALUE(int_col IGNORE NULLS) OVER(PARTITION BY id ORDER BY id, int_col) AS last_int
FROM window_int_test
```
to deterministically sort the individual partitions. `PARTITION BY id ORDER BY id` doesn't guarantee an order as `id` has the same value for each partition. 

Also fixed a bug in LAST_VALUE for all three types.

See above

No

```
mvn test -pl itests/qtest -Pitests -Dtest=TestMiniLlapLocalCliDriver -Dtest.output.overwrite=true -Dqfile="vectorized_first_last_value_ignore_nulls_double.q,vectorized_first_last_value_ignore_nulls_int.q,vectorized_first_last_value_ignore_nulls_decimal.q"
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/ptf/VectorPTFEvaluatorDecimalLastValue.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/ptf/VectorPTFEvaluatorDecimalLastValue.java
@@ -67,7 +67,7 @@ public class VectorPTFEvaluatorDecimalLastValue extends VectorPTFEvaluatorBase {
         lastValue.set(decimalColVector.vector[0]);
         isGroupResultNull = false;
       } else {
-        isGroupResultNull = true;
+        isGroupResultNull = doesRespectNulls() || !lastValue.isSet();
       }
     } else if (decimalColVector.noNulls) {
       lastValue.set(decimalColVector.vector[size - 1]);
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/ptf/VectorPTFEvaluatorDoubleLastValue.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/vector/ptf/VectorPTFEvaluatorDoubleLastValue.java
@@ -64,7 +64,7 @@ public class VectorPTFEvaluatorDoubleLastValue extends VectorPTFEvaluatorBase {
         lastValue = doubleColVector.vector[0];
         isGroupResultNull = false;
       } else {
-        isGroupResultNull = true;
+        isGroupResultNull = doesRespectNulls() || lastValue == null;
       }
     } else if (doubleColVector.noNulls) {
       lastValue = doubleColVector.vector[size - 1];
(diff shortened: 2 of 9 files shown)


kasakrisz (approved):

kasakrisz:
@soumyakanti3578 
Could you please share some details and an example about the bug you fixed?

soumyakanti3578:
@kasakrisz 
Sorry I should have clarified it in more detail. The bug was fixed by this:
```
    if (longColVector.isRepeating) {
      if (longColVector.noNulls || !longColVector.isNull[0]) {
        lastValue = longColVector.vector[0];
        isGroupResultNull = false;
      } else {
--        isGroupResultNull = true;
++        isGroupResultNull = doesRespectNulls() || lastValue == null;
      }
    }
```

When `longColVector.isRepeating = true`, it means that the column vector contains the same value for all the rows in the batch and thus we just store the value in the `0th` index. 

When `if (longColVector.noNulls || !longColVector.isNull[0])` condition is `true`, we can simply pick up the value from the `0th` index because all rows in the batch has the same repeating non null value, so `lastValue` would be the same value too, and hence the group result is not null as indicated by `isGroupResultNull = false`.

But when the condition is `false`, i.e., when values in the batch are repeating for the column (`longColVector.isRepeating = true`) AND the value in the first element is null (`!(longColVector.noNulls || !longColVector.isNull[0])`) we were simply setting `isGroupResultNull = true;` as indeed then the value in the column vector is null. This was fine when we were not supporting `IGNORE NULLS`.

Now, we must check if we respect nulls with `doesRespectNulls()`. If we do, then the group result is indeed null and we can set it to `true`. Else if we are ignoring nulls, then we must check if lastValue was set by the previous batch or not. If it was set, then we must set `isGroupResultNull = false;`

In short, when we are in the `else` block, the value in the column vector is indeed `null`. But we must check the `lastValue` because it could have been set by the previous batch.

In the int test file, we have these three rows for id = 2: `(2, 20), (2, null), (2, null)`. Since we have `id` and `int_col` in the ORDER BY of `LAST_VALUE(int_col IGNORE NULLS) OVER(PARTITION BY id ORDER BY id, int_col) AS last_int` this would create two batches:
```
batch 1: [2, 20]
batch 2: [[2, null], [2, null]]
```
as we are dealing with `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`, i.e., both `(2, null)`s will be together. Also note that for batch 2, `longColVector.isRepeating = true`. While computing batch 2, we will go to the else block and we would have set the result to null before.

Hope this clarifies the bug!

PR #6028, opened 2025-08-14
HIVE-29146: Query with WITH clause fails during split generation when CTE materaliazation is enabled

kasakrisz:
When validating Read and WriteEntities after semantic analysis Hive looks for transactional (ACID) tables and sets the `ValidWriteIdList` if any found. However only the main query's entities are checked.
This patch propose that all Read and WriteEntities should be checked by iterating through all the CTE's too.

When 
* CTE materialization is enabled
* and the CTE is materialized
* and the CTE definition references at least one ACID table
* and the main query doesn't reference any ACID table
the `ValidWriteIdList` is not set hence exception is thrown when executing the CTE materialization plan

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestMiniLlapLocalCliDriver -Dqfile=cte_mat_acid_1.q -pl itests/qtest -Pitests
```


InvisibleProgrammer (approved):
LGTM.

Tests are pending. As I see, there are a couple of classes inherited from SemanticAnalyzer so let's see the test results.

PR #6029, opened 2025-08-15
HIVE-29148: Replace IcebergSplit#blockLocations with Iceberg's utility

okumin:

We would replace the custom `blockLocations` with the one in apache/iceberg.

https://issues.apache.org/jira/browse/HIVE-29148

For consistency and maintainability.

No.

Unit and integration tests

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergSplit.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergSplit.java
@@ -22,25 +22,17 @@ package org.apache.iceberg.mr.mapreduce;
 import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Set;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.BlockLocation;
-import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.mapreduce.InputSplit;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.ScanTaskGroup;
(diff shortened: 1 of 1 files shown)


iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/mapreduce/IcebergSplit.java
```diff
@@ -78,33 +70,14 @@ public String[] getLocations() {
...
-          locationSets.addAll(Arrays.asList(location.getHosts()));
-        }
-      } catch (IOException e) {
-        LOG.warn("Failed to get block locations for path {}", path, e);
-      }
-    });
-
-    return locationSets.toArray(new String[0]);
-  }
-
```
okumin, line 107: Just for your information, [this is the replacement as of 1.9.2](https://github.com/apache/iceberg/blob/apache-iceberg-1.9.2/core/src/main/java/org/apache/iceberg/hadoop/Util.java#L61-L76).

deniskuzZ: @okumin, shouldn't we upgrade the iceberg version? We are on 1.9.1 ATM
https://github.com/apache/hive/blob/master/iceberg/pom.xml#L29

okumin: Regarding this context, no, we shouldn't. My patch is included in 1.8.0 or newer versions.

deniskuzZ: ok, then i misunderstood the comment in https://github.com/apache/hive/pull/6029#discussion_r2278271445

deniskuzZ (approved):
+1, pending tests

okumin:
@deniskuzZ Thanks for reviewing this pull request 👍

PR #6032, opened 2025-08-16
HIVE-29149: Improve TableFetcher.getTables to return Iterable to prevent possible OOM

Neer393:
Modified TableFetcher.getTables to lazily return Iterable<Table> 

TableFetcher.getTables used to return a list of HMS API Table objects directly which could possibly strain the memory and cause OOM. This is fixed by this change to return an Iterable<Table> to prevent the possible OOM

No user facing changes were made here

Tested locally by executing the unit tests

CC:- @deniskuzZ @dengzhhu653 @wecharyu


Neer393:
Closed as no more needed as this was supposed to be delivered as a fix for HIVE-28952 but now this will be shipped as an addendum to the same ticket

PR #6033, opened 2025-08-16
Revert "HIVE-28952: TableFetcher to return Table objects instead of names"

okumin:
Reverts apache/hive#6020


Neer393:
@okumin Please close this PR because as per the discussions on #6020 we have decided to solve this as an addendum to HIVE-28952

PR #6035, opened 2025-08-18
HIVE-29154: Fix hive-exec shaded jar config for opencsv package change

armitage420:

Incomplete package migration in [HIVE-28248](https://issues.apache.org/jira/browse/HIVE-28248)

In [HIVE-28248](https://issues.apache.org/jira/browse/HIVE-28248), opencsv was upgraded to v5.9 and its package changed from net.sf.opencsv to com.opencsv.
However, the hive-exec shaded jar configuration was not updated accordingly, and still references the old package.
This Jira updates the shading config to relocate com.opencsv correctly and remove leftover net.sf.opencsv entries.

No

The patch was tested in my cluster. Basically, one such scenario would fail where we create a table with OpenCSVSerde, push some values, followed by running a "Select count(*) from the table" in tez execution environment. On replacing the hive-exec jar, with the current change, the query runs successfully with correct output. Without the same patch, it fails with "NoClassDefFound: com/opencsv/ICSVParser"

### DIFF
--- a/ql/pom.xml
+++ b/ql/pom.xml
@@ -1075,7 +1075,7 @@
                   <include>com.fasterxml.jackson.core:jackson-databind</include>
                   <include>com.fasterxml.jackson.dataformat:jackson-dataformat-yaml</include>
                   <include>com.google.guava:guava</include>
-                  <include>net.sf.opencsv:opencsv</include>
+                  <include>com.opencsv:opencsv</include>
                   <include>org.apache.hive:hive-storage-api</include>
                   <include>org.threeten:threeten-extra</include>
                   <include>org.apache.orc:orc-core</include>


Aggarwal-Raghav:
LGTM +1, pending tests

deniskuzZ:
@armitage420, could you please add a test

armitage420:
@Aggarwal-Raghav @zhangbutao @deniskuzZ Thanks for your time!

@deniskuzZ The test is not reproducible in q files, rather it is reproducible in my cluster with Hadoop - 3.4.1, Tez - 0.10.4.
Following are the steps followed:
```
set hive.execution.engine=tez;
 drop database if exists hive_header_footer_db cascade; create database if not exists hive_header_footer_db;
 use hive_header_footer_db;

CREATE EXTERNAL TABLE `testcase1`(id int, name string) ROW FORMAT SERDE 'org.apache.hadoop.hive.serde2.OpenCSVSerde' LOCATION '/user/hive/hiveData/hive_header_footer_compressed/testcase1' TBLPROPERTIES ("skip.header.line.count"="1", "skip.footer.line.count"="1");

LOAD DATA INPATH '/tmp/testcase1.csv' INTO TABLE testcase1;

select count(*) from testcase1;
```
The error is thrown during count(*) and not during simple reads.

Following is the testcase1.csv content:
```
1,2019-12-31
2,2019-12-31
3,2019-12-31
```
Here is the reproducible error without the changes:
Screenshot 2025-08-19 at 12 51 52 PM

Here is the result of the same command with replacing hive-exec jar with the new change:
image

deniskuzZ:
> @deniskuzZ The test is not reproducible in q files, rather it is reproducible in my cluster with Hadoop - 3.4.1, Tez - 0.10.4. 

I don't feel comfortable merging changes that aren’t reproducible through qtest, however, the fix looks valid.

PS: I can't reproduce the issue on our internal clusters. [HIVE-28248](https://issues.apache.org/jira/browse/HIVE-28248) was not backported. cc @difin 


Screenshot 2025-08-19 at 12 02 30

zhangbutao:
I did a test with Hive4.1.0 + Tez 0.10.5 + Hadoop 3.4.1. Foud the similar exception: `java.lang.NoClassDefFoundError: com/opencsv/ICSVParser`

```
create external table test_open_csv_serde (id int)row format serde 'org.apache.hadoop.hive.serde2.OpenCSVSerde';

insert into test_open_csv_serde values(12);
```

image

deniskuzZ:
ok, let's merge then

armitage420:
@deniskuzZ @zhangbutao thanks for you help and review!

I would also like to add add the reasoning behind this issue being not reproducible on qtest.
Qtest runs on local single JVM having all the necessary libraries present in classpath. Whereas for cluster, tez has separate containers launched, and the opencsv.jar required was not present for runtime environments.

deniskuzZ:
> I would also like to add add the reasoning behind this issue being not reproducible on qtest.
Qtest runs on local single JVM having all the necessary libraries present in classpath. Whereas for cluster, tez has separate containers launched, and the opencsv.jar required was not present for runtime environments.

I also checked the itests dependencies and we import everything there

PR #6036, opened 2025-08-19
HIVE-29155: Improve performance of TestVectorAggregation

thomasrebele:

The TestVectorAggregation took 63s on my laptop due to a slow implementation of getLinearRandomNumber. I propose to replace the for loop with an analytical solution (HIVE-29155). With the patch, TestVectorAggregation takes only 23s. I've added javadoc to the method to make it easier to understand what the method is doing.

I hope it is OK to remove some unused imports as well.

TestVectorAggregation is slower than necessary, therefore consuming resources (time & CPU) in the CI and machines of developers.

No

Manual test that the distribution of generated numbers stays the same. See attachment of HIVE-29155. TestVectorAggregation was executed and passes.

### DIFF
--- a/ql/src/test/org/apache/hadoop/hive/ql/exec/vector/aggregation/TestVectorAggregation.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/exec/vector/aggregation/TestVectorAggregation.java
@@ -18,28 +18,20 @@
 
 package org.apache.hadoop.hive.ql.exec.vector.aggregation;
 
-import java.lang.reflect.Constructor;
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Random;
 import java.util.Set;
-import java.sql.Timestamp;
 
 import org.apache.hadoop.hive.common.type.DataTypePhysicalVariation;
 import org.apache.hadoop.hive.ql.exec.vector.VectorRandomBatchSource;
 import org.apache.hadoop.hive.ql.exec.vector.VectorRandomRowSource;
 import org.apache.hadoop.hive.ql.exec.vector.VectorRandomRowSource.GenerationSpec;
-import org.apache.hadoop.hive.ql.exec.vector.expressions.aggregates.VectorAggregateExpression;
 import org.apache.hadoop.hive.ql.plan.ExprNodeColumnDesc;
 import org.apache.hadoop.hive.ql.plan.ExprNodeDesc;
 import org.apache.hadoop.hive.ql.udf.generic.GenericUDAFCount.GenericUDAFCountEvaluator;
 import org.apache.hadoop.hive.ql.udf.generic.GenericUDAFEvaluator;
-import org.apache.hadoop.hive.ql.udf.generic.GenericUDAFVariance;
-import org.apache.hadoop.hive.ql.udf.generic.GenericUDAFEvaluator.AggregationBuffer;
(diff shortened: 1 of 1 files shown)


ql/src/test/org/apache/hadoop/hive/ql/exec/vector/aggregation/TestVectorAggregation.java
```diff
@@ -314,19 +306,35 @@ private boolean checkDecimal64(boolean tryDecimal64, TypeInfo typeInfo) {
...
+
+    // Invert the problem: 1:*, 2:**, 3:***, 4:****
+    // So in the example, star index 0 becomes index 9 and vice versa
+    randomInt = triangularNumber - randomInt - 1;
+    // Use the formula for triangular numbers to convert from the star index to the number
+    // n*(n+1)/2 = s
+    // n*n + n - 2s = 0
+    // then use the greater solution of the quadratic formula
+    // n = ( -1 + sqrt(1-4*1*(-2s)) )/2
+    int n = (int)(-1 + Math.sqrt(1+8* (randomInt)))/2;
```
soumyakanti3578: nit: formatting is inconsistent inside `sqrt`; could change to `sqrt(1 + 8 * randomInt)`

soumyakanti3578 (approved):
The new formula looks good to me and it was explained clearly in the comments. I just have a nit about formatting but otherwise it's a **`+1`** from me.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/aggregation/TestVectorAggregation.java
thomasrebele: Thanks for spotting! Fixed.

ql/src/test/org/apache/hadoop/hive/ql/exec/vector/aggregation/TestVectorAggregation.java
```diff
@@ -314,19 +306,35 @@ private boolean checkDecimal64(boolean tryDecimal64, TypeInfo typeInfo) {
...
+
+    // Invert the problem: 1:*, 2:**, 3:***, 4:****
+    // So in the example, star index 0 becomes index 9 and vice versa
+    randomInt = triangularNumber - randomInt - 1;
+    // Use the formula for triangular numbers to convert from the star index to the number
+    // n*(n+1)/2 = s
+    // n*n + n - 2s = 0
+    // then use the larger solution of the quadratic formula
+    // n = ( -1 + sqrt(1-4*1*(-2s)) )/2
+    int n = (int)(-1 + Math.sqrt(1+8*randomInt))/2;
```
deniskuzZ, line 335: maybe `(int) Math.floor((-1 + Math.sqrt(1+8*randomInt))/2)`

thomasrebele: As far as I know, casting a double to an int truncates the decimal part. The floor function would be applied to a non-negative double, so the result would be the same as casting with `(int)`.

deniskuzZ: You’re casting an intermediate result to int, not the final value. However, since the expression is divided by 2, both approaches yield the same outcome.

thomasrebele: You're right, the `(int)` cast has a higher operator precedence than the division. It would have been cleaner to add another pair of parentheses.

PR #5855, opened 2025-06-10
HIVE-29000: Upgrade nimbus-jose-jwt

arorasimran0309:

PR #6037, opened 2025-08-19
HIVE-29000: Upgrade nimbus-jose-jwt to 10.3

saihemanth-cloudera:
Avoid CVE

Due to CVEs

No

Dependency Tree verification

### DIFF
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -25,7 +25,7 @@
   <name>Hive Service</name>
   <properties>
     <hive.path.to.root>..</hive.path.to.root>
-    <nimbus-jose-jwt.version>9.37.3</nimbus-jose-jwt.version>
+    <nimbus-jose-jwt.version>10.3</nimbus-jose-jwt.version>
   </properties>
   <dependencies>
     <!-- dependencies are always listed in sorted order by groupId, artifactId -->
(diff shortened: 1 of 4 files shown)


service/pom.xml
```diff
@@ -25,7 +25,7 @@
   <name>Hive Service</name>
   <properties>
     <hive.path.to.root>..</hive.path.to.root>
-    <nimbus-jose-jwt.version>9.37.3</nimbus-jose-jwt.version>
+    <nimbus-jose-jwt.version>10.3</nimbus-jose-jwt.version>
```
ayushtkn, line 28: We have 10.4.2 released as well, can we move to the latest
https://mvnrepository.com/artifact/com.nimbusds/nimbus-jose-jwt/10.4.2

arorasimran0309:
Appreciate the effort here. Although, this issue and its changes are being handled as a part of #5855

ayushtkn:
@saihemanth-cloudera can you check the PR from @arorasimran0309, I think that PR is older, if that is fine, we can abandon this in favour of that one

saihemanth-cloudera:
@ayushtkn @arorasimran0309 - I had to put up this patch because there was activity on the older patch and we need to address this soon. 
Since the other patch is revived now, I'll abandon this patch

PR #6039, opened 2025-08-21
HIVE-28952: Addendum - TableFetcher to return Table objects instead of names

Neer393:
In this addendum commit, I am now working with batches so that I do not load everything in memory in once. Also, added this Optimization to IcebergTableOptimimizer.java

It's an improvement to reduce the number of msc calls and also to work with batches to not overload the memory

No user facing changes as this is an optimization

Locally by executing unit tests

### DIFF
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
@@ -30,21 +30,19 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.HiveMetaStoreClient;
 import org.apache.hadoop.hive.metastore.api.CompactionType;
 import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.ShowCompactResponse;
 import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
 import org.apache.hadoop.hive.metastore.txn.TxnStore;
 import org.apache.hadoop.hive.metastore.txn.entities.CompactionInfo;
 import org.apache.hadoop.hive.ql.metadata.Partition;
 import org.apache.hadoop.hive.ql.session.SessionState;
-import org.apache.hadoop.hive.ql.txn.compactor.CompactorUtil;
 import org.apache.hadoop.hive.ql.txn.compactor.MetadataCache;
 import org.apache.hadoop.hive.ql.txn.compactor.TableOptimizer;
-import org.apache.hive.iceberg.org.apache.orc.storage.common.TableName;
 import org.apache.iceberg.ContentFile;
 import org.apache.iceberg.DataOperations;
 import org.apache.iceberg.Snapshot;
@@ -85,62 +83,47 @@ public class IcebergTableOptimizer extends TableOptimizer {
    */
   @Override
   public Set<CompactionInfo> findPotentialCompactions(long lastChecked, ShowCompactResponse currentCompactions,
-      Set<String> skipDBs, Set<String> skipTables) {
+                                                      Set<String> skipDBs, Set<String> skipTables) {
     Set<CompactionInfo> compactionTargets = Sets.newHashSet();
 
-    getTableNames().stream()
-        .filter(table -> !skipDBs.contains(table.getDb()))
-        .filter(table -> !skipTables.contains(table.getNotEmptyDbTable()))
-        .map(table -> {
-          org.apache.hadoop.hive.ql.metadata.Table hiveTable = getHiveTable(table.getDb(), table.getTable());
-          org.apache.iceberg.Table icebergTable = IcebergTableUtil.getTable(conf, hiveTable.getTTable());
-          return Pair.of(hiveTable, icebergTable);
-        })
-        .filter(t -> hasNewCommits(t.getRight(),
-            snapshotTimeMilCache.get(t.getLeft().getFullyQualifiedName())))
-        .forEach(t -> {
-          String qualifiedTableName = t.getLeft().getFullyQualifiedName();
-          org.apache.hadoop.hive.ql.metadata.Table hiveTable = t.getLeft();
-          org.apache.iceberg.Table icebergTable = t.getRight();
+    Iterable<Table> tables = getTables(skipDBs, skipTables);
 
-          if (icebergTable.spec().isPartitioned()) {
-            List<org.apache.hadoop.hive.ql.metadata.Partition> partitions = findModifiedPartitions(hiveTable,
-                icebergTable, snapshotTimeMilCache.get(qualifiedTableName), true);
+    for (Table table : tables) {
+      org.apache.hadoop.hive.ql.metadata.Table hiveTable = new org.apache.hadoop.hive.ql.metadata.Table(table);
+      org.apache.iceberg.Table icebergTable = IcebergTableUtil.getTable(conf, table);
+      String qualifiedTableName = hiveTable.getFullyQualifiedName();
 
-            partitions.forEach(partition -> addCompactionTargetIfEligible(hiveTable.getTTable(), icebergTable,
-                partition.getName(), compactionTargets, currentCompactions, skipDBs, skipTables));
+      if (hasNewCommits(icebergTable, snapshotTimeMilCache.get(qualifiedTableName))) {
+        if (icebergTable.spec().isPartitioned()) {
+          List<org.apache.hadoop.hive.ql.metadata.Partition> partitions = findModifiedPartitions(hiveTable,
+              icebergTable, snapshotTimeMilCache.get(qualifiedTableName), true);
 
-            if (IcebergTableUtil.hasUndergonePartitionEvolution(icebergTable) && !findModifiedPartitions(hiveTable,
-                icebergTable, snapshotTimeMilCache.get(qualifiedTableName), false).isEmpty()) {
-              addCompactionTargetIfEligible(hiveTable.getTTable(), icebergTable,
-                  null, compactionTargets, currentCompactions, skipDBs, skipTables);
-            }
-          } else {
-            addCompactionTargetIfEligible(hiveTable.getTTable(), icebergTable, null, compactionTargets,
-                currentCompactions, skipDBs, skipTables);
+          partitions.forEach(partition -> addCompactionTargetIfEligible(table, icebergTable,
+              partition.getName(), compactionTargets, currentCompactions, skipDBs, skipTables));
+
+          if (IcebergTableUtil.hasUndergonePartitionEvolution(icebergTable) && !findModifiedPartitions(hiveTable,
+              icebergTable, snapshotTimeMilCache.get(qualifiedTableName), false).isEmpty()) {
+            addCompactionTargetIfEligible(table, icebergTable,
+                null, compactionTargets, currentCompactions, skipDBs, skipTables);
           }
+        } else {
+          addCompactionTargetIfEligible(table, icebergTable, null, compactionTargets,
+              currentCompactions, skipDBs, skipTables);
+        }
 
-          snapshotTimeMilCache.put(qualifiedTableName, icebergTable.currentSnapshot().timestampMillis());
-        });
+        snapshotTimeMilCache.put(qualifiedTableName, icebergTable.currentSnapshot().timestampMillis());
+      }
+    }
 
     return compactionTargets;
   }
 
-  private List<org.apache.hadoop.hive.common.TableName> getTableNames() {
+  private Iterable<Table> getTables(Set<String> skipDBs, Set<String> skipTables) {
     try {
-      return IcebergTableUtil.getTableFetcher(client, null, "*", null).getTableNames();
+      int maxBatchSize = MetastoreConf.getIntVar(conf, MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX);
+      return IcebergTableUtil.getTableFetcher(client, null, "*", null).getTables(skipDBs, skipTables, maxBatchSize);
     } catch (Exception e) {
-      throw new RuntimeMetaException(e, "Error getting table names");
-    }
-  }
-
-  private org.apache.hadoop.hive.ql.metadata.Table getHiveTable(String dbName, String tableName) {
-    try {
-      Table metastoreTable = metadataCache.computeIfAbsent(TableName.getDbTable(dbName, tableName), () ->
-          CompactorUtil.resolveTable(conf, dbName, tableName));
-      return new org.apache.hadoop.hive.ql.metadata.Table(metastoreTable);
-    } catch (Exception e) {
-      throw new RuntimeMetaException(e, "Error getting Hive table for %s.%s", dbName, tableName);
+      throw new RuntimeMetaException(e, "Error getting tables");
     }
   }
 
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
@@ -18,13 +18,9 @@
 
 package org.apache.iceberg.mr.hive.metastore.task;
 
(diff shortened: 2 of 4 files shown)


standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -110,24 +114,29 @@ public List<TableName> getTableNames() throws Exception {
...
+      return Collections.emptyList();
     }
 
     List<String> databases = client.getDatabases(catalogName, dbPattern);
 
-    for (String db : databases) {
-      List<String> tablesNames = getTableNamesForDatabase(catalogName, db);
-      for (Table table : new TableIterable(client, db, tablesNames, maxBatchSize)) {
-        candidates.add(table);
+    return () -> databases.stream().flatMap(db -> {
```
deniskuzZ: This is not the correct implementation. We should avoid eagerly creating multiple `TableIterable` instances (1 per database).

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
```diff
@@ -85,52 +85,53 @@ public IcebergTableOptimizer(HiveConf conf, TxnStore txnHandler, MetadataCache m
...
-          }
-
-          snapshotTimeMilCache.put(qualifiedTableName, icebergTable.currentSnapshot().timestampMillis());
-        });
+    Iterable<Table> tables = getTables();
+
+    for (Table table : tables) {
+      if (skipDBs.contains(table.getDbName()) || skipTables.contains(
+          TableName.fromString(table.getTableName(), table.getCatName(), table.getDbName()).getNotEmptyDbTable())) {
+        continue;
```
deniskuzZ: do not use `continue` - it's a bad practice

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
```diff
@@ -85,52 +85,53 @@ public IcebergTableOptimizer(HiveConf conf, TxnStore txnHandler, MetadataCache m
...
-                  null, compactionTargets, currentCompactions, skipDBs, skipTables);
-            }
-          } else {
-            addCompactionTargetIfEligible(hiveTable.getTTable(), icebergTable, null, compactionTargets,
-                currentCompactions, skipDBs, skipTables);
-          }
-
-          snapshotTimeMilCache.put(qualifiedTableName, icebergTable.currentSnapshot().timestampMillis());
-        });
+    Iterable<Table> tables = getTables();
```
deniskuzZ: `getHiveTable` is no longer used

`skipDBs` and `skipTables` add an extra filter post select, maybe we can optimize, by propagating it to TableFetcher

````
List<String> databases = client.getDatabases(catalogName, dbPattern)
  .filter(dbName -> !skipDBs.contains(dbName)
  .toList()
````
and 
````
 List<String> tablesNames = getTableNamesForDatabase(catalogName, db);
  .filter(tblName -> !skipTables.contains(tblName)
  .toList()
````

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: also you can pass the `metadataCache` to avoid unnecessary data load

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java (hunk 1 above)
Comment (Neer393) [reply standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java]: TableIterable expects a db name so we need to create 1 TableIterable object per database

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java (hunk 3 above)
Comment (Neer393) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: The current implementation uses metadataCache so as to reduce the number of msc calls because here they were making one msc call per table for getting the HMS API Table object.

Now since using TableIterator, we have already reduced the number of msc calls, metadataCache does not seem to fit into the picture as regardless of whether we store the HMS API Table object in cache or not, TableFetcher will fetch it for us. So I don't see a point storing table object.

Please correct me if I am wrong

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java (hunk 2 above)
Comment (Neer393) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: Resolved

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java (hunk 1 above)
Comment (deniskuzZ) [reply standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java]: you shouldn't create it eagerly! only once done with 1 table, switch to the next. No need to open iterators for all of them at once

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java (hunk 3 above)
Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: I was thinking the same and you are probably right

Note `IcebergHouseKeeperService` uses both `tableCache` and Session `QueryState` to retrieve the cached icebergTable by name. One of them is redundant.  

````
IcebergTableUtil.getTable(Configuration, Properties, boolean skipCache) ->
  SessionStateUtil.getResource(configuration, tableIdentifier)
````
and 
````
private Table getIcebergTable(org.apache.hadoop.hive.metastore.api.Table table) {
  TableName tableName = TableName.fromString(table.getTableName(), table.getCatName(), table.getDbName());
  return tableCache.get(tableName, key -> IcebergTableUtil.getTable(conf, table));
}
````

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: IcebergTableUtil.getTable would be making 1 HMS call per table. Would be nice if we can optimize by caching the data (i.e. use `HiveMetaStoreClientWithLocalCache` in both `HiveTableOperations` and `TableFetcher`).  

Catalogs.loadTable -> HiveTableOperations.doRefresh -> metaClients.run(client -> client.getTable(database, tableName));

````
if (HiveConf.getBoolVar(conf, HiveConf.ConfVars.MSC_CACHE_ENABLED)) {
  HiveMetaStoreClientWithLocalCache.init(conf);
}

MSC_CACHE_ENABLED("hive.metastore.client.cache.v2.enabled", true, 
  "This property enables a Caffeine Cache for Metastore client")
````

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java (hunk 1 above)
Comment (Neer393) [reply standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java]: Oh okay, on it 👍

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java (hunk 3 above)
Comment (Neer393) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: > One of them is redundant.

Yes you are right. I believe the tableCache is redundant as anyways at the end of the day, requests are sent to IcebergTableUtil.getTable which has the following implementation where we can clearly see that `skipCache` is set to false. So indeed we first look into cache.
```
public static Table getTable(Configuration configuration, org.apache.hadoop.hive.metastore.api.Table hmsTable) {
    return getTable(configuration, hmsTable, false);
}
```

Let's get rid of tableCache

Comment (Neer393) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: > IcebergTableUtil.getTable would be making 1 HMS call per table. Would be nice if we can optimize by caching the data (i.e. use `HiveMetaStoreClientWithLocalCache` in both `HiveTableOperations` and `TableFetcher`).
> 
> Catalogs.loadTable -> HiveTableOperations.doRefresh -> metaClients.run(client -> client.getTable(database, tableName));
> 
> ```
> if (HiveConf.getBoolVar(conf, HiveConf.ConfVars.MSC_CACHE_ENABLED)) {
>   HiveMetaStoreClientWithLocalCache.init(conf);
> }
> 
> MSC_CACHE_ENABLED("hive.metastore.client.cache.v2.enabled", true, 
>   "This property enables a Caffeine Cache for Metastore client")
> ```

I did not understand much but what I understood is that you are asking me to make HiveTableOperations.java and TableFetcher.java make use of HiveMetaStoreClientWithLocalCache for which, I have to just add the following piece of code to both the files. Am I right?
```
if (HiveConf.getBoolVar(conf, HiveConf.ConfVars.MSC_CACHE_ENABLED)) {
      HiveMetaStoreClientWithLocalCache.init(conf);
}
```

Now comes the questions:-
1. How is Catalogs.loadTable invoking HiveTableOperations.doRefresh as when I see in HiveTableOperations.doRefresh, it shows no usages ? Am I missing something here ?
2. Even if we consider adding it to code, let's say I added this to HiveTableOperations' constructor but adding this to TableFetcher is not possible as TableFetcher does not have access to `conf`. One thing that can be done is adding this to IcebergHouseKeeperService.java but I don't know if it will add any value

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: > How is Catalogs.loadTable invoking HiveTableOperations.doRefresh as when I see in HiveTableOperations.doRefresh, it shows no usages ? Am I missing something here ?

try to run the test under debug

> Even if we consider adding it to code, let's say I added this to HiveTableOperations' constructor but adding this to TableFetcher is not possible as TableFetcher does not have access to conf. One thing that can be done is adding this to IcebergHouseKeeperService.java but I don't know if it will add any value

I don't follow why can't you propagate Configuration object to TableFetcher. 
This would introduce client-side caching, eliminating the need for multiple consecutive calls to hms.getTable(database, tableName).

Comment (Neer393) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: Adding HIveMetastoreClientWithLocalCache to TableFetcher is not possible as TableFetcher is a part to hive-metastore module and HiveMetaStoreClientWithLocalCache is under hive-exec module which has a dependency on hive-metastore module. 
The reason this is not possible is because adding this creates a circular dependency between the modules hive-metastore and hive-exec.
Anyways I have addressed rest of the comments

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java]: can't we create client with cache in metastore package or move the `HiveMetaStoreClientWithLocalCache` into the `hive-standalone-metastore-client` module?
probably that shouldn't be part of this PR

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java [hunk 4]
```diff
@@ -110,24 +112,29 @@ public List<TableName> getTableNames() throws Exception {
     return candidates;
   }
 
-  public List<Table> getTables(int maxBatchSize) throws Exception {
-    List<Table> candidates = new ArrayList<>();
-
+  public Iterable<Table> getTables(int maxBatchSize, Set<String> skipDBs, Set<String> skipTables) throws Exception {
```
deniskuzZ: could we please use next signatures
````
public Iterable<Table> getTables(Set<String> skipDBs, Set<String> skipTables, int maxBatchSize)

public Iterable<Table> getTables(int maxBatchSize) {
 return getTables(null, null, maxBatchSize)
}
````

File: standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java [hunk 5]
```diff
@@ -110,24 +112,29 @@ public List<TableName> getTableNames() throws Exception {
...
-
-    for (String db : databases) {
-      List<String> tablesNames = getTableNamesForDatabase(catalogName, db);
-      for (Table table : new TableIterable(client, db, tablesNames, maxBatchSize)) {
-        candidates.add(table);
-      }
-    }
-    return candidates;
+    List<String> databases = client.getDatabases(catalogName, dbPattern).stream()
+        .filter(dbName -> skipDBs == null || !skipDBs.contains(dbName)).toList();
```
deniskuzZ: please put `.toList()` on a new line

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
```diff
@@ -110,24 +112,29 @@ public List<TableName> getTableNames() throws Exception {
...
+
+    return () -> Iterators.concat(
+        Iterators.transform(databases.iterator(), db -> {
+          try {
+            List<String> tableNames = getTableNamesForDatabase(catalogName, db).stream()
+                .filter(tableName -> skipTables == null ||
+                    !skipTables.contains(TableName.fromString(tableName, catalogName, db).getNotEmptyDbTable()))
```
deniskuzZ: why not just use `TableName.getQualified(catalogName, db, tableName)`

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/metastore/task/IcebergHouseKeeperService.java
```diff
@@ -78,22 +68,17 @@ public void run() {
   private void expireTables(String catalogName, String dbPattern, String tablePattern) {
     try (IMetaStoreClient msc = new HiveMetaStoreClient(conf)) {
       int maxBatchSize = MetastoreConf.getIntVar(conf, MetastoreConf.ConfVars.BATCH_RETRIEVE_MAX);
-      List<org.apache.hadoop.hive.metastore.api.Table> tables =
-          IcebergTableUtil.getTableFetcher(msc, catalogName, dbPattern, tablePattern).getTables(maxBatchSize);
-      LOG.debug("{} candidate tables found", tables.size());
+      Iterable<org.apache.hadoop.hive.metastore.api.Table> tables =
+          IcebergTableUtil.getTableFetcher(msc, catalogName, dbPattern, tablePattern)
+              .getTables(maxBatchSize, null, null);
```
deniskuzZ: please keep the log

Neer393: We cannot keep the log as we no longer have access to tables.size() because we are now getting Iterable object instead of a list whose size is predefined

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/utils/TableFetcher.java
Neer393: getQualified adds the catalogname as the prefix but the skiptables set that we have contains tables in the format `dbname.tablename` so ideally we cannot use getQualified

```
public static String getQualified(String catName, String dbName, String tableName) {
    return catName + DatabaseName.CAT_DB_TABLE_SEPARATOR + dbName + DatabaseName.CAT_DB_TABLE_SEPARATOR + tableName;
  }
```

```
public String getNotEmptyDbTable() {
    String metaRefName = tableMetaRef == null ? "" : "." + tableMetaRef;
    return db == null || db.trim().isEmpty() ? table : db + DatabaseName.CAT_DB_TABLE_SEPARATOR + table + metaRefName;
  }
```

deniskuzZ: there is `TableName getDbTable` that doesn't use catalog

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/compaction/IcebergTableOptimizer.java
deniskuzZ: @Neer393, would you mind creating ticket for that, so we won't forget? Thanks!

Neer393: Sure @deniskuzZ 
Will create a ticket right away 👍 
Here it is:- HIVE-29163

deniskuzZ (approved):
+1

PR #6040, opened 2025-08-22
HIVE-29160: Replace forward() in vectorized operators with vectorForward().

ngsg:

This patch replaces `forward()` calls in `VectorAppMasterEventOperator` and `VectorPTFOperator` with `vectorForward()`.

Using `forward()` in a vectorized operator may cause incorrect results when there are multiple child operators. In addition, `forward()` treats a batch of rows as a single row, resulting in an incorrect count of processed rows.

No

I have checked that event operators now report the correct row counts in `EXPLAIN ANALYZE` output and in counters. This change can also be found in the updated qfile outputs.


ayushtkn (approved):
LGTM

ql/src/test/results/clientpositive/llap/tez_compile_counters.q.out
```diff
@@ -1501,7 +1501,7 @@ Stage-1 HIVE COUNTERS:
    RECORDS_OUT_INTERMEDIATE_Map_4: 240
    RECORDS_OUT_INTERMEDIATE_Reducer_2: 952
    RECORDS_OUT_INTERMEDIATE_Reducer_3: 0
-   RECORDS_OUT_OPERATOR_EVENT_30: 1
+   RECORDS_OUT_OPERATOR_EVENT_30: 309
```
InvisibleProgrammer, line 1504: That change revealed something that looks like a bug in thforward method: 
It doesn't calculate the size of the batch at all, it just adds 1 to the statistics: 

```java
  protected void forward(Object row, ObjectInspector rowInspector)
      throws HiveException {
    runTimeNumRows++;
```

I wonder, is it worth fixing in this ticket or a new one? Or it is intentional? 
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/exec/Operator.java#L875

ngsg: In the case of non-vectorized execution, Hive passes a single row via `forward()`, so it is expected that `runtimeNumRows` increases by 1.

ngsg:
Thanks @deniskuzZ, @ayushtkn, and @InvisibleProgrammer for the review!

ql/src/test/results/clientpositive/llap/tez_compile_counters.q.out
InvisibleProgrammer: vectorForward just got the exact same actual parameter as forward. And now the stat changed from 1 to 309. 
I think the batch size was 309 at the original call just forward ignored it. If it were received 1, that output shouldn't changed at all.

ngsg: As you pointed out, `vectorForward()` handles batch size correctly, while `forward()` ignores it. That's why switching to `vectorForward()` changes the stats from 1 to 309. However, since vectorized operators should use `vectorForward()`, I believe `forward()` does not need to consider `VectorizedRowBatch`, and both methods behave as intended. Please let me know if I misunderstood your point.

PR #6041, opened 2025-08-22
HIVE-29161: Correct the row count computation affected by Dynamic SemiJoin Reduction

ngsg:

This patch adds the missing parentheses in the calculation of the number of rows affected by Dynamic SemiJoin Reduction.
https://github.com/apache/hive/blob/df98c2e2f4e2370c603f8873f006ee8728afa8a6/ql/src/java/org/apache/hadoop/hive/ql/parse/TezCompiler.java#L1975

Without the parentheses, Hive sets `newNumRows` either to 1 or leaves it unchanged. This can lead to incorrect decisions about whether a SemiJoin branch is beneficial, potentially resulting in suboptimal query plans.
(FYI: In our 10TB TPC-DS experiment, we observed that the patch affected some query plans, but there was no change in query execution time.)

No

I manually checked the log files and verified that the number of rows is now computed correctly.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/parse/TezCompiler.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/parse/TezCompiler.java
@@ -1972,7 +1972,7 @@ public class TezCompiler extends TaskCompiler {
       for (SemijoinOperatorInfo roi : reductionFactorMap.values()) {
         // This semijoin will be kept
         // We are going to adjust the filter statistics
-        long newNumRows = (long) (1.0 - roi.reductionFactor) * roi.filterStats.getNumRows();
+        long newNumRows = (long) ((1.0 - roi.reductionFactor) * roi.filterStats.getNumRows());
         if (LOG.isDebugEnabled()) {
           LOG.debug("Old stats for {}: {}", roi.filterOperator, roi.filterStats);
           LOG.debug("Number of rows reduction: {}/{}", newNumRows, roi.filterStats.getNumRows());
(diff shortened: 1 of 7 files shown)


ngsg:
> +1, however, could we please change the commit message and JIRA title to something like "Correct the row count computation affected by Dynamic SemiJoin Reduction"

I have updated the JIRA title and the first commit message as you suggested.

zabetak (approved):
Nice finding! I am +1 with the patch since it fixes an obvious bug. 

In terms of actual benefit, I am not sure if we can consider it a net improvement or a regression. Per @ngsg observations on TPC-DS benchmark, execution time remains the same while the new plans have more semijoin branches. It seems that we are doing more work without significant gain in execution time. I am mentioning this just in case we want to follow-up with further improvements on the cost-model & SJ removal logic.

ngsg:
@zabetak, thank you for the comment. I also agree that we need to check whether this is an actual improvement or not. I revisited the experiment result and its logs, but I couldn't pinpoint the core reason why the SemiJoin branches were unbeneficial. Possible causes include the logic itself not being effective, stats misestimation leading to incorrect SJ benefit/cost computation, or something else I may have overlooked.

Looking closely at the code, the logic and the idea described in HIVE-20775 make sense to me, and the changes in `dynamic_semijoin_reduction_multicol.q.out` [1] seem to demonstrate a case where this logic is useful (though that occurs when `hive.tez.dynamic.semijoin.reduction.multicolumn` is disabled, which is not the general case).

I'll add the concern about actual benefit to my personal TODO list and will share results if I find further evidence for a follow-up ticket.

[1] https://github.com/apache/hive/pull/6041/files#diff-187118dfcb8b70a73f5682a2011de5c5cf4d3457789cc44b6d7216c188da5b97L332-R362

ngsg:
Thanks @deniskuzZ and @zabetak for the review!

PR #6042, opened 2025-08-23
HIVE-29073: Overlay modified session level metaconf to new hmsclient connection

Aggarwal-Raghav:
Check [HIVE-29073](https://issues.apache.org/jira/browse/HIVE-29073). When metaconf are modified in beeline session and HMS is restarted or call goes to another HMS (if multiple HMS are present) the modified configs are not honoured. 

This change will help the consistency of metaconf in a beeline session irrespective of HMS restart or call going to another HMS   

NO

Local testing, tested on 3 node 4.0.1 cluster.

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
@@ -6465,6 +6465,21 @@ private void constructOneLBLocationMap(FileStatus fSta,
 
   public void setMetaConf(String propName, String propValue) throws HiveException {
     try {
+      /*
+       * Updates the 'conf' object with session-level metastore variables
+       * ('metaConfVars'). This object is used to initialize the
+       * Thrift client connection to the Hive Metastore, ensuring that any
+       * session-specific overrides are propagated to the underlying connection.
+       *
+       * For reference on how this 'conf' object is consumed, see the client
+       * instantiation logic in:
+       * org.apache.hadoop.hive.ql.metadata.Hive#createMetaStoreClient()
+       */
+      if (Arrays.stream(MetastoreConf.metaConfVars)
+          .anyMatch(s -> s.getVarname().equals(propName))) {
+        // Storing varname prevents conflicts with HiveServer2-level configurations
+        conf.set(propName, propValue);
+      }
       getMSC().setMetaConf(propName, propValue);
     } catch (TException te) {
       throw new HiveException(te);
--- a/standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/ThriftHiveMetaStoreClient.java
+++ b/standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/ThriftHiveMetaStoreClient.java
@@ -763,6 +763,8 @@ public class ThriftHiveMetaStoreClient extends BaseMetaStoreClient {
           LOG.error(errMsg, e);
         }
         if (isConnected) {
+          // Set the beeline session modified metaConfVars for new HMS connection
+          overlaySessionModifiedMetaConf();
           break;
         }
       }
@@ -790,6 +792,23 @@ public class ThriftHiveMetaStoreClient extends BaseMetaStoreClient {
     }
 
     snapshotActiveConf();
+  }
+
+  private void overlaySessionModifiedMetaConf() {
+    for (MetastoreConf.ConfVars confVar : MetastoreConf.metaConfVars) {
+      String confVal = conf.get(confVar.getVarname());
+      if (!org.apache.commons.lang3.StringUtils.isBlank(confVal)) {
+        try {
+          setMetaConf(confVar.getVarname(), confVal);
+        } catch (TException e) {
+          LOG.error(
+              "Failed to set metastore config for {} with value {}",
+              confVar.getVarname(),
+              confVal,
+              e);
+        }
+      }
+    }
   }
 
   // wraps the underlyingTransport in the appropriate transport based on mode of authentication
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
@@ -225,7 +225,6 @@ public class MetastoreConf {
       ConfVars.HMS_HANDLER_ATTEMPTS,
       ConfVars.HMS_HANDLER_INTERVAL,
       ConfVars.HMS_HANDLER_FORCE_RELOAD_CONF,
-      ConfVars.PARTITION_NAME_WHITELIST_PATTERN,
       ConfVars.ORM_RETRIEVE_MAPNULLS_AS_EMPTY_STRINGS,
       ConfVars.USERS_IN_ADMIN_ROLE,
       ConfVars.HIVE_TXN_MANAGER,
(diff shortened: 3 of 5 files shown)


Aggarwal-Raghav:
CC @zhangbutao @deniskuzZ can you please provide your thoughts on this approach? this PR is to collect feedback on the approach.

ql/src/java/org/apache/hadoop/hive/ql/processors/SetProcessor.java
```diff
@@ -272,10 +272,25 @@ public Map<String, String> getHiveVariable() {
     }
     if (register) {
       ss.getOverriddenConfigurations().put(key, value);
+
+      if (metaConfOverlayList.contains(key)) {
+        try (Hive hive = Hive.get(ss.getConf())) {
+          // Resources will be closed automatically
+          hive.close(true);
```
Aggarwal-Raghav: This hive.close(true) is required becuase in **_org.apache.hadoop.hive.metastore.client.BaseMetaStoreClient#BaseMetaStoreClient()_**, if there are multiple session level modifed configs, only the top one is considered as post the connection creation conf object is not updated for same session

ql/src/java/org/apache/hadoop/hive/ql/processors/SetProcessor.java
```diff
@@ -272,10 +272,25 @@ public Map<String, String> getHiveVariable() {
     }
     if (register) {
       ss.getOverriddenConfigurations().put(key, value);
+
+      if (metaConfOverlayList.contains(key)) {
+        try (Hive hive = Hive.get(ss.getConf())) {
```
Indhumathi27: suggest using **needsRefresh** with get() API here instead. If **needsRefresh** alone not able to accomplish calling close, i suggest introducing a new variable to refresh Hive object.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -244,18 +243,26 @@ public String toString() {
...
       ConfVars.AGGREGATE_STATS_CACHE_MAX_FULL,
       ConfVars.AGGREGATE_STATS_CACHE_CLEAN_UNTIL,
-      ConfVars.DISALLOW_INCOMPATIBLE_COL_TYPE_CHANGES,
       ConfVars.FILE_METADATA_THREADS,
       ConfVars.METASTORE_CLIENT_FILTER_ENABLED,
       ConfVars.METASTORE_SERVER_FILTER_ENABLED,
       ConfVars.METASTORE_PARTITIONS_PARAMETERS_INCLUDE_PATTERN,
-      ConfVars.METASTORE_PARTITIONS_PARAMETERS_EXCLUDE_PATTERN
+      ConfVars.METASTORE_PARTITIONS_PARAMETERS_EXCLUDE_PATTERN,
+      // Add metaConfVars here as well
```
Aggarwal-Raghav, line 251: Please check the top comment during initializing `metaVars`. If the config value is changed for any of these Confvars, it will create new HMS client connection.
Code flow when user do: `set metaconf:metastore.try.direct.sql=false;`
```
SetProcessor.java (Hive hive = Hive.get(ss.getConf()) ->  Hive.java (getInternal(..)) -> Hive.java 
(isCompatible() ... -> ThriftHiveMetaStoreClient#isCompatibleWith() which returns false and causes to create new HMS Client connection
 ```

**Comment by Aggarwal-Raghav** [comment]
Before the patch:
Screenshot 2025-10-05 at 12 32 23 AM

After the patch:
Screenshot 2025-10-04 at 11 21 38 PM

**Comment by Aggarwal-Raghav** [comment]
@dengzhhu653 , can you please provide the feedback on this?

File: ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java [hunk 4]
```diff
@@ -6465,6 +6465,21 @@ public List<Function> getFunctionsInDb(String dbName, String pattern) throws Hiv
...
+       * Updates the 'conf' object with session-level metastore variables
+       * ('metaConfVars'). This object is used to initialize the
+       * Thrift client connection to the Hive Metastore, ensuring that any
+       * session-specific overrides are propagated to the underlying connection.
+       *
+       * For reference on how this 'conf' object is consumed, see the client
+       * instantiation logic in:
+       * org.apache.hadoop.hive.ql.metadata.Hive#createMetaStoreClient()
+       */
+      if (Arrays.stream(MetastoreConf.metaConfVars)
```
Indhumathi27, line 6478: i think this check is not required. getMSC().setMetaConf will validate if the meta conf is valid one or not. can move `conf.set(propName, propValue);` post setMetaConf

Aggarwal-Raghav: The updated logic permits only the 'varname' and disallows 'hivename'. For example:
```
set metaconf:metastore.try.direct.sql=false; -- this will pass the check
set metaconf:hive.metastore.try.direct.sql=false; -- this will fail the check
```
**Reason for this change**:
As conf object is created from HiveConf HS2 side this `hive.metastore.try.direct.sql` is present and `set metaconf:hive.metastore.try.direct.sql=false` will modify the HS2 conf as well resulting in failure of following q files.
```
TestNegativeLlapLocalCliDriver: allow_change_col_type_par_neg.q
TestMiniLlapLocalCliDriver: allow_change_col_type_par.q
TestMiniLlapLocalCliDriver: set_metaconf.q 
```
**NOTE**: HS2 side `hive.metastore.try.direct.sql` is marked deprecated and used only in test code.

Options:
1. We can rewrite these q files and their output (q.out) and support both varname or hivename.
2. Only support varname as done in this approach **to keep q files intact** but the `set metaconf:hivename=...; ` will not be overlayed on new HMS connection. Forcing user to use varname which is better in a way as it isolates /shows clear distinction between HiveConf and MetastoreConf.

Indhumathi27: I suggest to remove the deprecated Hivename MetaConf Variables and use Metastore configurations only. If not, both metaconf:hivename= and hive conf variables related to metastore should behave consistent, as they are intended for the same purpose.

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/client/ThriftHiveMetaStoreClient.java
```diff
@@ -763,6 +763,8 @@ private void open() throws MetaException {
           LOG.error(errMsg, e);
         }
         if (isConnected) {
+          // Set the beeline session modified metaConfVars for new HMS connection
+          overlaySessionModifiedMetaConf();
```
dengzhhu653, line 767: for Metastore clients, they can always set the meta conf after the connection is established, I'm not sure if it's necessary to set the meta conf on the initialization, for the socket timeout?

Aggarwal-Raghav: If the connection between HiveServer2 and the Hive Metastore is dropped, subsequent HMS calls invoke `org.apache.hadoop.hive.ql.metadata.Hive#getMSC()`. Because the `forceCreate` parameter is **false**, the existing `metaStoreClient` instance is reused rather than a new one being created.
The client's recovery logic only re-attempts the underlying thrift connection via `org.apache.hadoop.hive.metastore.client.ThriftHiveMetaStoreClient#open()`. Once this connection is re-established, the standard code path does not provide an opportunity to set the modified session-level metaconf.
Therefore, `overlaySessionModifiedMetaConf` is explicitly called immediately after a successful reconnection to ensure the session's modified configuration is correctly applied to the newly established HMS connection.

dengzhhu653: Thank you. I'm not aware of the retry, makes sense

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
dengzhhu653: could we file a jira to remove those deprecated Hive MetaConf Variables?

dengzhhu653 (approved):

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
Aggarwal-Raghav: Sure @dengzhhu653 , Have created [HIVE-29289](https://issues.apache.org/jira/browse/HIVE-29289), as part of this I'll remove the deprecated configuration from HiveConf related to MetaStore.

dengzhhu653:
Thank you @Aggarwal-Raghav and @Indhumathi27 for the contribution and review!

PR #6043, opened 2025-08-26
HIVE-28695: Implement Lineage information for windowing functions

InvisibleProgrammer:
Lineage information for PTF functions hasn't been implemented in Hive so far. Currently, the lineage information is the default one: it collects all the columns from all the acting tables and fills the lineage information with them. 
In this PR, I address this issue by collecting the accurate column info for the windowing functions.

The affected functions are: 
- row_number
- rank
- dense_rank
- ntile
- percent_rank
- lag
- lead
- first_value
- last_value
- cume_dist
- percentile_cont
- percentile_disc
- sum
- matchpath

There are two significant changes in the lineage info in `HookContext`:  
- dependency graph
- expression

The change affects the output of two different post exec hooks:  
- PostExecutePrinter: provides a simple string representation of the columns involved in the dependency graph. 
- LineageLogger: logs out the whole HookContext in a JSon format.

**Dependency graph**: 
Before this change, dependency graph contained ALL the columns from the affected tables, not only the columns involved in the PTF expression. 
For example, from `ptf.q`: 
POSTEXEC hook before the change: 
```
POSTHOOK: Lineage: part_4_n2.dr SCRIPT [(part)part.FieldSchema(name:p_partkey, type:int, comment:null), (part)part.FieldSchema(name:p_name, type:string, comment:null), (part)part.FieldSchema(name:p_mfgr, type:string, comment:null), (part)part.FieldSchema(name:p_brand, type:string, comment:null), (part)part.FieldSchema(name:p_type, type:string, comment:null), (part)part.FieldSchema(name:p_size, type:int, comment:null), (part)part.FieldSchema(name:p_container, type:string, comment:null), (part)part.FieldSchema(name:p_retailprice, type:double, comment:null), (part)part.FieldSchema(name:p_comment, type:string, comment:null), ]
```
And after the change: 
```
OSTHOOK: Lineage: part_4_n2.dr EXPRESSION [(part)part.FieldSchema(name:p_name, type:string, comment:null), (part)part.FieldSchema(name:p_mfgr, type:string, comment:null), ]
```

An example of a LineageLogger output:
```
{
    "version": "1.0",
    "engine": "tez",
    "database": "default",
    "hash": "dc1e91de973788e1c7bb767ab3e4f5f0",
    "queryText": "create view b_v_5_x as\nselect *\nfrom (select col_001, row_number() over (order by src.p1) as r_num\n from source_tbl2 src) v1",
    "edges": [
        {
            "sources": [
                2
            ],
            "targets": [
                0
            ],
            "edgeType": "PROJECTION"
        },
        {
            "sources": [
                3
            ],
            "targets": [
                1
            ],
            "expression": "row_number() over (order by src.p1 ROWS between unbounded and unbounded)",
            "edgeType": "PROJECTION"
        }
    ],
    "vertices": [
        {
            "id": 0,
            "vertexType": "COLUMN",
            "vertexId": "default.b_v_5_x.col_001"
        },
        {
            "id": 1,
            "vertexType": "COLUMN",
            "vertexId": "default.b_v_5_x.r_num"
        },
        {
            "id": 2,
            "vertexType": "COLUMN",
            "vertexId": "default.source_tbl2.col_001"
        },
        {
            "id": 3,
            "vertexType": "COLUMN",
            "vertexId": "default.source_tbl2.p1"
        }
    ]
}
```

## Additional notes about how the expression tag can differ from the script that Hive received: 

Compiler adds some modifications the original AST and at some cases, we can see them in the output of the expression tag of the LineageLogger: 

- At some cases, the PTF function can get the partitioning column as a parameter.
- At some cases, the PTF function gets a default ordering based on partitioning if order by clause is not explicitly defined to make sure the output is determenistic.
- The windowing expression is added (ROWS BETWEEN ... )

Example: 
```
rank() over (partition by src.p1) as r_num
-->
rank(src.p1) over (partition by src.p1 order by src.p1 ROWS between unbounded and unbounded)
```

Please note, the expression tag was some compiler default before this change.
For example, from lineage2.q.out: 

Original query text:
```
select sum(a.c1) over (partition by a.c1 order by a.id)\nfrom dest_l2 a\nwhere a.c2 != 10\ngroup by a.c1, a.c2, a.id\nhaving count(a.c2) > 0
```
Original expression tags:
```
(tok_function sum (. (tok_table_or_col $hdt$_0) c1) (tok_windowspec (tok_partitioningspec (tok_distributeby (. (tok_table_or_col $hdt$_0) c1)) (tok_orderby (tok_tabsortcolnameasc (tok_nulls_last (. (tok_table_or_col $hdt$_0) id))))) (tok_windowvalues (preceding unbounded) current)))
```
became: 
```
sum(a.c1) over (partition by a.c1 order by a.id RANGE between unbounded and unbounded)
```

To implement a missing feature.

If the user is Atlas or another third-party tool that consumes lineage information, yes. For regular end users, I would say no. 

Added a new qtest, `lineage_ptf.q`, containing simple examples for all affected functions. And also with different test cases for partitioning, grouping, or both of them. Added some a little bit more complex test cases that use more than one column where it is possible. 
Also the output is checked and re-generated for some additional qtests.


ql/src/test/queries/clientpositive/lineage_ptf.q
```diff
@@ -0,0 +1,125 @@
...
+--- Views for single PTF operators
+
+-- Test for ROW_NUMBER
+create view b_v_4_0 as
+select *
+from (select col_001,
+    row_number() over (partition by src.p1) as r_num,
+    row_number() over (partition by src.col_002) as r_num2,
+    row_number() over (partition by src.col_002) as r_num3
+
```
kasakrisz: nit.: unnecessary extra line

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
```diff
@@ -676,6 +683,92 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+        addAllMakeUniqueIfNotNull(columns, funcDef.getPartition().getExpressions().getFirst().getExprNode().getCols());
+      }
+      if (funcDef.getOrder() != null) {
+        addAllMakeUniqueIfNotNull(columns, funcDef.getOrder().getExpressions().getFirst().getExprNode().getCols());
+      }
+
+      if (!(funcDef.getTFunction() instanceof Noop)) {
+
+          if (funcDef instanceof WindowTableFunctionDef
+                  && ((WindowTableFunctionDef) funcDef).getWindowFunctions().getFirst().getArgs() != null) {
```
kasakrisz: There might be more than one window function here:
```
...
select 
    row_number() over (partition by src.p1) as r_num,
    rank() over (partition by src.col_002) as r_num2
...
```
Should we check all of them here?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
```diff
@@ -676,6 +683,92 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+
+              for (PTFExpressionDef arg : ((WindowTableFunctionDef) funcDef).getWindowFunctions().getFirst().getArgs()) {
+
+                  if (!(arg.getExprNode() instanceof ExprNodeConstantDesc)) {
+                      columns.addAll(arg.getExprNode().getCols());
```
kasakrisz: Should column uniqueness be enforced here? I saw we use `addAllMakeUniqueIfNotNull` a few lines earlier.
Could `columns` be a set instead of list ?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: No, if you see the first example in `lineage.ptf.q`: 
```sql
create view b_v_4_0 as
select *
from (select col_001,
    row_number() over (partition by src.p1) as r_num,
    row_number() over (partition by src.col_002) as r_num2,
    row_number() over (partition by src.col_002) as r_num3

        from source_tbl2 src) v1;
```
The output is: 
```
"edges":[
      {
         "sources":[
            4
         ],
         "targets":[
            0
         ],
         "edgeType":"PROJECTION"
      },
      {
         "sources":[
            5
         ],
         "targets":[
            1
         ],
         "expression":"(tok_function row_number (tok_windowspec (tok_partitioningspec (tok_distributeby (. (tok_table_or_col src) p1)) (tok_orderby (tok_tabsortcolnameasc (tok_nulls_first (. (tok_table_or_col src) p1))))) (tok_windowrange (preceding unbounded) (following unbounded))))",
         "edgeType":"PROJECTION"
      },
      {
         "sources":[
            6
         ],
         "targets":[
            2,
            3
         ],
         "expression":"(tok_function row_number (tok_windowspec (tok_partitioningspec (tok_distributeby (. (tok_table_or_col src) col_002)) (tok_orderby (tok_tabsortcolnameasc (tok_nulls_first (. (tok_table_or_col src) col_002))))) (tok_windowrange (preceding unbounded) (following unbounded))))",
         "edgeType":"PROJECTION"
      }
   ],
   "vertices":[
      {
         "id":0,
         "vertexType":"COLUMN",
         "vertexId":"default.b_v_4_0.col_001"
      },
      {
         "id":1,
         "vertexType":"COLUMN",
         "vertexId":"default.b_v_4_0.r_num"
      },
      {
         "id":2,
         "vertexType":"COLUMN",
         "vertexId":"default.b_v_4_0.r_num2"
      },
      {
         "id":3,
         "vertexType":"COLUMN",
         "vertexId":"default.b_v_4_0.r_num3"
      },
      {
         "id":4,
         "vertexType":"COLUMN",
         "vertexId":"default.source_tbl2.col_001"
      },
      {
         "id":5,
         "vertexType":"COLUMN",
         "vertexId":"default.source_tbl2.p1"
      },
      {
         "id":6,
         "vertexType":"COLUMN",
         "vertexId":"default.source_tbl2.col_002"
      }
   ]
```

The reason why it works in this way is because that process method is called on each node that is a PTF function. So if we add 3 PTF expressions into a select, we will get 3 calls. 
On the other side, getFirst is enough as there can be only one windowing function in the PTF expression. The same is true for partitions and order by. Taker order by as a simplest example: a simple order by can contain multiple columns. But cannot have multiple order by statements in a single select.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: Good idea. Let me check that.

ql/src/test/queries/clientpositive/lineage_ptf.q
InvisibleProgrammer: Thx.

InvisibleProgrammer:
I have found some edge cases (like over by or partition by clause can have multiple columns. And I started to implement the expr field in the lineage as well (it is a text representation of the clause). 

Converted back to Draft

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
```diff
@@ -676,6 +689,249 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+
+        if (windowFrameDef.getStart().isCurrentRow()) {
+          sb.append("current_row");
+        } else {
+          sb.append(windowFrameDef.getStart().isUnbounded() ? "unbounded" : windowFrameDef.getStart().getAmt() + " preceding");
+        }
+
+        sb.append(" and ");
+
+        if (windowFrameDef.getStart().isCurrentRow()) {
```
kasakrisz: Should this second call of `getStart()` be `getEnd()` ?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
```diff
@@ -676,6 +689,249 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+          sb.append("current_row");
+        } else {
+          sb.append(windowFrameDef.getStart().isUnbounded() ? "unbounded" : windowFrameDef.getStart().getAmt() + " following");
+        }
```
kasakrisz: This code is duplicated. How about extracting this to method like
```
appendBoundary(BoundaryDef boundaryDef, String precedingOrfollowing)
```
or just 
```
appendBoundary(BoundaryDef boundaryDef)
```
because `BoundaryDef` seems to have all the data we need to print.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
```diff
@@ -676,6 +689,249 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+      LineageCtx lCtx = (LineageCtx) procCtx;
+
+      // The operators
+      @SuppressWarnings("unchecked")
+      PTFOperator op = (PTFOperator)nd;
+      Operator<? extends OperatorDesc> inpOp = getParent(stack);
+      lCtx.getIndex().copyPredicates(inpOp, op);
+
+      Dependency dep = new Dependency();
+      DependencyType new_type = DependencyType.EXPRESSION;
```
kasakrisz: `new_type` - > `newType`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
```diff
@@ -676,6 +689,249 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+          sb.append("current_row");
+        } else {
+          sb.append(windowFrameDef.getStart().isUnbounded() ? "unbounded" : windowFrameDef.getStart().getAmt() + " following");
+        }
+      }
+
+      sb.append(")");
+      dep.setExpr(sb.toString());
+
+      LinkedHashSet<BaseColumnInfo> col_set = new LinkedHashSet<>();
```
kasakrisz: `col_set` -> `colSet`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
```diff
@@ -676,6 +689,249 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+      }
+
+      dep.setType(new_type);
+      dep.setBaseCols(col_set);
+
+      // This dependency is then set for all the colinfos of the script operator
+      for(ColumnInfo ci : op.getSchema().getSignature()) {
+        Dependency d = dep;
+          Dependency dep_ci = lCtx.getIndex().getDependency(inpOp, ci);
+          if (dep_ci != null) {
```
kasakrisz: `dep_ci` -> `depCI`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
```diff
@@ -676,6 +689,249 @@ public Object process(Node nd, Stack<Node> stack, NodeProcessorCtx procCtx,
...
+
+              sb.append("arg").append(counter++).append("(");
+              sb.append(ExprProcFactory.getExprString(funcDef.getRawInputShape().getRr().getRowSchema(), arg.getExprNode(), lCtx, inpOp, null));
+              sb.append("), ");
+            }
+
+            sb.delete(sb.length() - 2, sb.length());
+          }
+
+        }
```
kasakrisz: Could you please merge the blocks and else blocks of these if statements? Both has the same condition 
```
(funcDef instanceof WindowTableFunctionDef)
```

ql/src/test/queries/clientpositive/lineage_ptf.q
```diff
@@ -0,0 +1,142 @@
+set hive.exec.post.hooks=org.apache.hadoop.hive.ql.hooks.LineageLogger;
+
```
kasakrisz, line 2: Is there a test case when a column is referenced more than once in a window expression?
```
sum(src.col_001) over (partition by src.col_001 order by src.col_001)
```

kasakrisz (changes requested):

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: I changed it to Set.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: Personally, I totally agree. On the other hand, I want to keep the code consistent. If you search for this text in this file, that is the naming convention at all node processors in this file. 
If you agree with that, I can change all the occurrences  so that it will look like a Java code. But I also know, if I just rename those alone, somebody will add a review comment like why did you do unnecessary changes?

InvisibleProgrammer: Renamed the underscored variables.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: Sure, thx.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: Yes, nice catch!

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: Ok.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: Ok

ql/src/java/org/apache/hadoop/hive/ql/optimizer/lineage/OpProcFactory.java
InvisibleProgrammer: Ok

ql/src/test/queries/clientpositive/lineage_ptf.q
InvisibleProgrammer: I'm not completely sure if I completely understand this question. 
Do you mean something like `order by col_001, col_001` ?

kasakrisz: Thanks for adding a test case to cover this.

PR #6045, opened 2025-08-28
HIVE-28974: Implement pattern-based catalog retrieval in metastore client

Neer393:
This PR added support for pattern based catalog retrieval. The change is that a new method was added called ```getCatalogs(String catalogPattern)``` . If all catalogs are to be fetched, set catalogPattern to null, else if catalogPattern is set to empty string, no catalogs are returned. Also the existing `getCatalogs()` under the hood calls `getCatalogs(null)`.

These changes are needed for IcebergHouseKeeperService where the service is expiring snapshots according to the dbpattern and tablepattern but there was no support for catalogpattern yet. This is being added in this PR

No user facing change. Only updated getCatalogs to fetch catalogs based on pattern

This patch was tested locally by executing tests to ensure no breaking changes are introduced.


Neer393:
@deniskuzZ please have a look at this PR

PR #6046, opened 2025-08-28
HIVE-28974: Implement pattern-based catalog retrieval in metastore client

Neer393:
This PR added support for pattern based catalog retrieval. The change is that a new method was added called ```getCatalogs(String catalogPattern)``` . If all catalogs are to be fetched, set catalogPattern to null, else if catalogPattern is set to empty string, no catalogs are returned. Also the existing `getCatalogs()` under the hood calls `getCatalogs(null)`.

These changes are needed for IcebergHouseKeeperService where the service is expiring snapshots according to the dbpattern and tablepattern but there was no support for catalogpattern yet. This is being added in this PR

No user facing change. Only updated getCatalogs to fetch catalogs based on pattern

This patch was tested locally by executing tests to ensure no breaking changes are introduced.

### DIFF
--- a/standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
+++ b/standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
@@ -132,6 +132,15 @@ public interface IMetaStoreClient extends AutoCloseable {
   List<String> getCatalogs() throws MetaException, TException;
 
   /**
+   * Get a list of all catalogs known to the system.
+   * @param catalogPattern pattern for the catalog name to match
+   * @return list of catalog names
+   * @throws MetaException something went wrong, usually in the database.
+   * @throws TException general thrift exception.
+   */
+  List<String> getCatalogs(String catalogPattern) throws MetaException, TException;
+
+  /**
    * Drop a catalog.  Catalogs must be empty to be dropped, there is no cascade for dropping a
    * catalog.
    * @param catName name of the catalog to drop
--- a/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
+++ b/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
@@ -2619,7 +2619,7 @@ service ThriftHiveMetastore extends fb303.FacebookService
   void create_catalog(1: CreateCatalogRequest catalog) throws (1:AlreadyExistsException o1, 2:InvalidObjectException o2, 3: MetaException o3)
   void alter_catalog(1: AlterCatalogRequest rqst) throws (1:NoSuchObjectException o1, 2:InvalidOperationException o2, 3:MetaException o3)
   GetCatalogResponse get_catalog(1: GetCatalogRequest catName) throws (1:NoSuchObjectException o1, 2:MetaException o2)
-  GetCatalogsResponse get_catalogs() throws (1:MetaException o1)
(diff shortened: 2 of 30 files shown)


standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
```diff
@@ -2619,7 +2619,7 @@ service ThriftHiveMetastore extends fb303.FacebookService
   void create_catalog(1: CreateCatalogRequest catalog) throws (1:AlreadyExistsException o1, 2:InvalidObjectException o2, 3: MetaException o3)
   void alter_catalog(1: AlterCatalogRequest rqst) throws (1:NoSuchObjectException o1, 2:InvalidOperationException o2, 3:MetaException o3)
   GetCatalogResponse get_catalog(1: GetCatalogRequest catName) throws (1:NoSuchObjectException o1, 2:MetaException o2)
-  GetCatalogsResponse get_catalogs() throws (1:MetaException o1)
+  GetCatalogsResponse get_catalogs(1: GetCatalogRequest pattern) throws (1:MetaException o1)
```
wecharyu, line 2622: It seems a breaking change.  is the original `get_catalogs()` api still available for old clients?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
```diff
@@ -131,6 +131,15 @@ void alterCatalog(String catalogName, Catalog newCatalog)
    */
   List<String> getCatalogs() throws MetaException, TException;
```
wecharyu, line 132: We can deprecate it after introducing `getCatalogs(String catlogPattern)`

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
Neer393: No. Should I keep the existing `get_catalogs()` ?

standalone-metastore/metastore-client/src/main/java/org/apache/hadoop/hive/metastore/IMetaStoreClient.java
Neer393: We can deprecate this but we can use it to fetch all catalogs without providing a parameter for pattern. It's like in IMetaStoreClient where we have 2 methods for getDatabases
```
List<String> getDatabases(String databasePattern) throws MetaException, TException;
```
```
List<String> getDatabases(String catName, String databasePattern)
      throws MetaException, TException;
```

standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift
wecharyu: but if there are clients in old version call `get_catalogs()`, the server in new version could not response.

Neer393: Oh sorry I have made changes in all clients like IMetaStoreClient, BaseMetaStoreClient, MetaStoreClientWrapper and ThriftHiveMetastoreClient and this `get_catalogs()` under the hood calls `get_catalogs(null)` by passing catalogPattern as null. So it's non-breaking change

wecharyu: sorry, I mean if the server use new api `get_catalogs(pattern)`, but some engines that do not upgrade the client version may still call `get_catalog()` api, in this case the call would fail.

PR #6047, opened 2025-08-28
HIVE-29167: Use access() api to check permission for different filesystems

wecharyu:
Use filesystem's `access()` api to check permission.

The permission modes vary in different filesystems, we could not just check permissions based on the group mode.

No.

Removed the `supergroup` test and add a new test in `TestHdfsUtils`.
```bash
mvn test -Dtest.groups= -Dtest=org.apache.hadoop.hive.metastore.utils.TestHdfsUtils -pl :hive-standalone-metastore-server
```

### DIFF
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/HdfsUtils.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/HdfsUtils.java
@@ -63,14 +63,13 @@ public class HdfsUtils {
 
   /**
    * Check the permissions on a file.
+   * @param fs Filesystem the file is contained in
    * @param path the file path
-   * @param conf the {@link Configuration} used when checking permissions
    * @param action action to be performed
    * @throws IOException If thrown by Hadoop
    */
-  public static void checkFileAccess(Path path, Configuration conf, FsAction action)
+  public static void checkFileAccess(FileSystem fs, Path path, FsAction action)
       throws IOException {
-    FileSystem fs = path.getFileSystem(conf);
     checkFileAccess(fs, path, action, SecurityUtils.getUGI());
   }
 
--- a/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
+++ b/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java
@@ -512,7 +512,8 @@ public class Warehouse {
       return false;
     }
     try {
-      HdfsUtils.checkFileAccess(path, conf, FsAction.WRITE);
+      FileSystem fs = getFs(path);
+      HdfsUtils.checkFileAccess(fs, path, FsAction.WRITE);
       return true;
     } catch (FileNotFoundException fnfe){
       // File named by path doesn't exist; nothing to validate.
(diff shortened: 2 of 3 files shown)


standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/SecurityUtils.java
```diff
@@ -74,7 +73,7 @@
 public class SecurityUtils {
   private static final Logger LOG = LoggerFactory.getLogger(SecurityUtils.class);
 
-  public static UserGroupInformation getUGI() throws LoginException, IOException {
```
wecharyu, line 93: The `LoginException` would never be thrown, I consider to remove all uses of it in a new PR because it's not quite relative to current PR.

wecharyu:
@ayushtkn @deniskuzZ @zhangbutao could you help review this PR?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/HdfsUtils.java
```diff
@@ -65,57 +64,35 @@ public class HdfsUtils {
...
-      }
-    } else if (dirPerms.getOtherAction().implies(action)) {
-      return;
+    try {
+      ugi.doAs((PrivilegedExceptionAction<Void>) () -> {
+        fs.access(path, action);
+        return null;
+      });
+    } catch (InterruptedException e) {
+      throw new IOException(e);
```
zhangbutao, line 94: Why not throw `AccessControlException`?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/SecurityUtils.java
zhangbutao: That's ok

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/HdfsUtils.java
```diff
@@ -65,57 +64,35 @@ public class HdfsUtils {
...
-      }
-    } else if (arrayContains(groups, stat.getGroup())) {
-      if (dirPerms.getGroupAction().implies(action)) {
-        return;
-      }
-    } else if (dirPerms.getOtherAction().implies(action)) {
-      return;
+    try {
+      ugi.doAs((PrivilegedExceptionAction<Void>) () -> {
+        fs.access(path, action);
```
zhangbutao, line 90: **In case of hdfs filesystem**, If user A belongs to the superuser group `fs.getConf().get("dfs.permissions.supergroup", "")`, then user A should have all permissions. 
Can the `fs.access(path, action)` determine whether user A has all permissions?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/HdfsUtils.java
wecharyu: `AccessControlException` will be thrown directly, and `InterruptedException` is usually not related to permission.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/utils/HdfsUtils.java
wecharyu: Yes, hdfs `access()` API will also skip checkPermission for super user: 
https://github.com/apache/hadoop/blob/4d7825309348956336b8f06a08322b78422849b1/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/server/namenode/FSDirectory.java#L1954-L1960

zhangbutao:
LGTM 

@wecharyu Please rebase to fix the conflicts.

zhangbutao (approved):
LGTM +1

PR #6048, opened 2025-08-29
HIVE-29166: Fix the partition column update logic in ConvertJoinMapJoin#convertJoinBucketMapJoin.

ngsg:

Modify the partition column update logic in `ConvertJoinMapJoin#convertJoinBucketMapJoin` so that small table's partitionCols are updated if they do not match the big table's bucketCols. 

Previously, the partition columns were updated only when the number of partition columns is not equal to the number of bucket columns. However, this could cause incorrect bucket routing when the counts are equal but the column orders differ. For example, if the bucket columns are {id, part} and the partition columns are {part, id}, then the partition columns should be updated to {id, part}, but the current implementation does not update them.

The new logic now updates the partition columns if the counts are not equal or the column orders differ.

As described above, the current BucketMapJoin conversion logic may not update the partition columns properly, which leads to incorrect results, as reported in the JIRA.

No

I added two qfiles in which the current Hive returns incorrect results. I also checked that the patch fixes the originally reported duplicate-generation issue.

### DIFF
--- a/iceberg/iceberg-handler/src/test/results/positive/bucket_map_join_9.q.out
+++ b/iceberg/iceberg-handler/src/test/results/positive/bucket_map_join_9.q.out
@@ -24,76 +24,34 @@ SELECT * FROM tbl JOIN tbl tbl2 ON tbl.id = tbl2.id AND tbl.part = tbl2.part
 POSTHOOK: type: QUERY
 POSTHOOK: Input: default@tbl
 POSTHOOK: Output: hdfs://### HDFS PATH ###
-STAGE DEPENDENCIES:
-  Stage-1 is a root stage
-  Stage-0 depends on stages: Stage-1
+Plan optimized by CBO.
 
-STAGE PLANS:
-  Stage: Stage-1
-    Tez
-#### A masked pattern was here ####
-      Edges:
-        Map 1 <- Map 2 (CUSTOM_EDGE)
-#### A masked pattern was here ####
-      Vertices:
-        Map 1 
-            Map Operator Tree:
-                TableScan
-                  alias: tbl
-                  filterExpr: (id is not null and part is not null) (type: boolean)
-                  Statistics: Num rows: 2 Data size: 530 Basic stats: COMPLETE Column stats: COMPLETE
-                  Filter Operator
-                    predicate: (id is not null and part is not null) (type: boolean)
-                    Statistics: Num rows: 2 Data size: 530 Basic stats: COMPLETE Column stats: COMPLETE
-                    Select Operator
-                      expressions: foid (type: string), part (type: string), id (type: string)
-                      outputColumnNames: _col0, _col1, _col2
-                      Statistics: Num rows: 2 Data size: 530 Basic stats: COMPLETE Column stats: COMPLETE
-                      Map Join Operator
-                        condition map:
-                             Inner Join 0 to 1
-                        keys:
-                          0 _col1 (type: string), _col2 (type: string)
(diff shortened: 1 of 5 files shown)


ql/src/test/results/clientpositive/llap/bucketmapjoin14.q.out
```diff
@@ -0,0 +1,112 @@
...
+                    Statistics: Num rows: 2 Data size: 530 Basic stats: COMPLETE Column stats: COMPLETE
+                    Select Operator
+                      expressions: foid (type: string), part (type: string), id (type: string)
+                      outputColumnNames: _col0, _col1, _col2
+                      Statistics: Num rows: 2 Data size: 530 Basic stats: COMPLETE Column stats: COMPLETE
+                      Reduce Output Operator
+                        key expressions: _col1 (type: string), _col2 (type: string)
+                        null sort order: zz
+                        sort order: ++
+                        Map-reduce partition columns: _col2 (type: string), _col1 (type: string)
```
okumin, line 91: I verified that this list is inverted on the master branch, and the result would be empty.

```
Map-reduce partition columns: _col1 (type: string), _col2 (type: string)
```

iceberg/iceberg-handler/src/test/results/positive/bucket_map_join_9.q.out
```diff
@@ -0,0 +1,65 @@
...
+  Fetch Operator
+    limit:-1
+    Stage-1
+      Map 1 vectorized
+      File Output Operator [FS_53]
+        Map Join Operator [MAPJOIN_52] (rows=2 width=530)
+          BucketMapJoin:true,Conds:SEL_51._col1, _col2=RS_49._col1, _col2(Inner),Output:["_col0","_col1","_col2","_col3","_col4","_col5"]
+        <-Map 2 [CUSTOM_EDGE] vectorized
+          MULTICAST [RS_49]
+            PartitionCols:_col2, _col1
```
okumin, line 42: I verified that this list is inverted on the master branch, and the result would be empty.

```
PartitionCols:_col1, _col2
```

okumin (commented):
At first glance, this appears to be valid. I will take a look again on weenends because the updated part is complicated

difin (approved):
LGTM

aturoczy:
Thank you for the review. If it is OK, can we merge it to master?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/ConvertJoinMapJoin.java
```diff
@@ -656,32 +656,39 @@ private boolean convertJoinBucketMapJoin(JoinOperator joinOp, OptimizeTezProcCon
...
+    }
 
+    // If the number of partition columns is less than the number of bucket columns,
+    // then we cannot properly distribute small tables onto bucketized map tasks.
+    // Bail out.
+    if (positions.size() < listBucketCols.size()) {
+      return false;
+    }
+
+    CustomBucketFunction bucketFunction = opTraits.getCustomBucketFunctions().get(0);
```
deniskuzZ, line 690: @ngsg, @okumin shouldn't we apply `retainedColumns` to a `bucketFunction`?
````
final Optional<CustomBucketFunction> selected =
    opTraits.getCustomBucketFunctions().get(0).select(retainedColumns);
if (!selected.isPresent()) {
  LOG.info("{} can't keep itself only with {}", opTraits.getCustomBucketFunctions().get(0), retainedColumns);
  return false;
}
bucketFunction = selected.get();
````

PR #6051, opened 2025-08-31
HIVE-29170: Restrict the Query History service from being modified at the session level

abstractdog:
…


abstractdog:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=apache_hive&pullRequest=6051) **Quality Gate passed**
> Issues  [100 New issues](https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=6051&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=apache_hive&pullRequest=6051&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=apache_hive&pullRequest=6051&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0.0% Coverage on New Code](https://sonarcloud.io/component_measures?id=apache_hive&pullRequest=6051&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=apache_hive&pullRequest=6051&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=apache_hive&pullRequest=6051)

100 new issues, of course!

Aggarwal-Raghav:
Changes LGTM +1 (non-binding)

Was able to see the error
Screenshot 2025-09-01 at 7 59 19 PM

and after the patch, its working as intended:
```
0: jdbc:hive2://localhost:10000 (default)> set hive.query.history.enabled=true;
Error: Error while processing statement: Cannot modify hive.query.history.enabled at runtime. It is in the list of parameters that can't be modified at runtime or is prefixed by a restricted variable; Query ID: raghav_20250901200558_771d7f1a-40d1-421c-b486-c2ffba73766a (state=42000,code=1)
```

Aggarwal-Raghav (approved):

abstractdog:
thanks a lot @Aggarwal-Raghav for the review and testing so far!
rebased the patch to get a clean sonarqube result

InvisibleProgrammer:
I'm not sure about how that restricted list works but I wonder: what happens when we want to disable the service that is already enabled? I'm thinking about the test, `query_history.q`. 

The reason why I'm asking is that comment from the test:
```
-- any operation on query_history table might involve further record changes on query_history table,
-- so let's disable the feature while to reduce noise while checking the contents
set hive.query.history.enabled=false;
```

I also wonder, what's about challenging the solution: is it possible to do the opposite direction and create an instance when the user wants to enable the service?

abstractdog:
> I'm not sure about how that restricted list works but I wonder: what happens when we want to disable the service that is already enabled? I'm thinking about the test, `query_history.q`.
> 
> The reason why I'm asking is that comment from the test:
> 
> ```
> -- any operation on query_history table might involve further record changes on query_history table,
> -- so let's disable the feature while to reduce noise while checking the contents
> set hive.query.history.enabled=false;
> ```
> 
> I also wonder, what's about challenging the solution: is it possible to do the opposite direction and create an instance when the user wants to enable the service?

qtests run on a different codepath, where the commands are directly interpreted through the Driver, so the same session-related check is bypassed I think

regarding the opposite direction: on a real cluster, the user cannot enable the service after this patch on service level, which is the desired way, as the service needs to start and initialize together with the HS2 properly

InvisibleProgrammer:
> > I'm not sure about how that restricted list works but I wonder: what happens when we want to disable the service that is already enabled? I'm thinking about the test, `query_history.q`.
> > The reason why I'm asking is that comment from the test:
> > ```
> > -- any operation on query_history table might involve further record changes on query_history table,
> > -- so let's disable the feature while to reduce noise while checking the contents
> > set hive.query.history.enabled=false;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I also wonder, what's about challenging the solution: is it possible to do the opposite direction and create an instance when the user wants to enable the service?
> 
> qtests run on a different codepath, where the commands are directly interpreted through the Driver, so the same session-related check is bypassed I think
> 
> regarding the opposite direction: on a real cluster, the user cannot enable the service after this patch on service level, which is the desired way, as the service needs to start and initialize together with the HS2 properly

Got it, thank you.

ayushtkn (approved):
LGTM

PR #6052, opened 2025-09-01
HIVE-29171: Disable HBase web UI for related tests

KiranVelumuri:

Disable HBase web UI while setting up MiniHBaseCluster for HBase tests.

Improvement

No

mvn test -Dtest=TestHBaseQueries
mvn test -Dtest=TestHBaseCliDriver
mvn test -Dtest=TestHBaseNegativeCliDriver


ayushtkn (commented):
Like how much do we save here? these are basic integration tests in HBase. it has default values which enables bunch of things which we don't need, but still ok.

I believe that would be the case with Hadoop services as well. I am curious like how much is the improvement

KiranVelumuri:
Taking reference from [ManyMiniCluster](https://github.com/apache/hive/blob/6c7e972247957976fd12e9bbc4fb9343ffa78a71/itests/hcatalog-unit/src/test/java/org/apache/hive/hcatalog/hbase/ManyMiniCluster.java#L267-L269), I used it here for HBase tests. This is good to have when we upgrade javax.servlet.X classes and could help avoid loading of such classes in HBase tests.

There has been no improvement in test times as seen below. 
| Test | Avg. runtime (in sec) | Avg. runtime (in sec)|
| -------- | ------- | ------ |
| | Before | After |
| TestHBaseQueries | 48.353 | 44.529
| TestHBaseCliDriver | 758.333 | 762.667
| TestHBaseNegativeCliDriver | 59.440 | 58.109

ayushtkn:
thanx for the details. I don't feel the need for this, but I won't come in the way. If others feel it is good, I won't be blocking this

KiranVelumuri:
@ayushtkn Got it. Thank you for your review!

PR #6054, opened 2025-09-01
HIVE-29130: Remove jline 2.x

Aggarwal-Raghav:
Although Hive has moved to Jline3, 2.x was also getting shipped in packaging.

To have single Jline jar. Check [HIVE-29130](https://issues.apache.org/jira/browse/HIVE-29130) for more details

No

On local setup

### DIFF
--- a/packaging/src/main/assembly/beeline.xml
+++ b/packaging/src/main/assembly/beeline.xml
@@ -48,8 +48,7 @@
                 <include>org.apache.hive:hive-service-rpc:jar</include>
                 <include>commons-cli:commons-cli:jar</include>
                 <include>commons-io:commons-io:jar</include>
-                <include>commons-logging:commons-logging:jar</include>
-                <include>jline:jline:jar</include>
+                <include>org.jline:jline:jar</include>
                 <include>org.apache.thrift:libthrift:jar</include>
                 <include>org.slf4j:slf4j-api:jar</include>
                 <include>net.sf.supercsv:super-csv:jar</include>
--- a/pom.xml
+++ b/pom.xml
@@ -1732,6 +1732,13 @@
                   <searchTransitive>true</searchTransitive>
                   <message>Banned log4j:log4j dependency/transitive dependency was found!</message>
                 </bannedDependencies>
+                <bannedDependencies>
+                  <excludes>
+                    <exclude>jline:jline</exclude>
+                  </excludes>
+                  <searchTransitive>true</searchTransitive>
+                  <message>A banned jline2.x dependency/transitive dependency was found!</message>
+                </bannedDependencies>
               </rules>
               <fail>true</fail>
             </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -1736,6 +1736,9 @@
                   <excludes>
                     <exclude>jline:jline</exclude>
                   </excludes>
+                  <includes>
+                    <include>jline:jline:1.0</include>
+                  </includes>
                   <searchTransitive>true</searchTransitive>
                   <message>A banned jline2.x dependency/transitive dependency was found!</message>
                 </bannedDependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -1736,6 +1736,12 @@
                   <excludes>
                     <exclude>jline:jline</exclude>
                   </excludes>
+                  <!--
+                    Allowing jline 1.0 because pig 0.16.0 depends upon it. Otherwise
+                    causing UT failure in hcatalog/hcatalog-pig-adapter module. The
+                    below includes tag can be removed once pig moves to jline 3.x or
+                    support for pig is dropped from hive.
+                  -->
                   <includes>
                     <include>jline:jline:1.0</include>
                   </includes>
--- a/standalone-metastore/metastore-common/pom.xml
+++ b/standalone-metastore/metastore-common/pom.xml
@@ -253,10 +253,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>commons-logging</groupId>
-      <artifactId>commons-logging</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.apache.zookeeper</groupId>
       <artifactId>zookeeper</artifactId>
       <version>${zookeeper.version}</version>
@@ -564,29 +560,6 @@
         </executions>
       </plugin>
       <!-- TODO MS-SPLIT javadoc plugin -->
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-enforcer-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>enforce-banned-dependencies</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <configuration>
-              <rules>
-                <bannedDependencies>
-                  <excludes>
-                    <!--LGPL licenced library-->
-                    <exclude>com.google.code.findbugs:annotations</exclude>
-                  </excludes>
-                </bannedDependencies>
-              </rules>
-              <fail>true</fail>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
--- a/accumulo-handler/pom.xml
+++ b/accumulo-handler/pom.xml
@@ -32,6 +32,10 @@
       <artifactId>accumulo-core</artifactId>
       <exclusions>
         <exclusion>
+          <groupId>jline</groupId>
+          <artifactId>jline</artifactId>
+        </exclusion>
+        <exclusion>
           <groupId>commons-beanutils</groupId>
           <artifactId>commons-beanutils-core</artifactId>
         </exclusion>
--- a/hcatalog/core/pom.xml
+++ b/hcatalog/core/pom.xml
@@ -329,6 +329,10 @@
           <groupId>tomcat</groupId>
           <artifactId>jasper-runtime</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>jline</groupId>
+          <artifactId>jline</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
   </dependencies>
--- a/hcatalog/hcatalog-pig-adapter/pom.xml
+++ b/hcatalog/hcatalog-pig-adapter/pom.xml
@@ -103,6 +103,10 @@
           <groupId>tomcat</groupId>
           <artifactId>jasper-compiler</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>jline</groupId>
+          <artifactId>jline</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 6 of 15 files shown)


Aggarwal-Raghav:
As Hadoop has jline 3.9.0**, in my setup** it was getting picked first compared to 3.25.0 from hive. Causing **_NoSuchMethodError org.jline.reader.EndOfFileException#getPartialLine()_** check [here](https://issues.apache.org/jira/browse/HIVE-29130?focusedCommentId=18014311&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18014311).

**I would like the reviewer to confirm on this.**

Aggarwal-Raghav:
CC @ayushtkn @abstractdog

abstractdog:
thanks @Aggarwal-Raghav for taking care of this, would you be so kind to add an enforcer <bannedDependencies> item so ban this dependency even transitively in the future? that would be the perfect solution I believe

Aggarwal-Raghav:
> thanks @Aggarwal-Raghav for taking care of this, would you be so kind to add an enforcer item so ban this dependency even transitively in the future? that would be the perfect solution I believe

sure. Will update the PR.

packaging/src/main/assembly/beeline.xml
```diff
@@ -48,8 +48,7 @@
                 <include>org.apache.hive:hive-service-rpc:jar</include>
                 <include>commons-cli:commons-cli:jar</include>
                 <include>commons-io:commons-io:jar</include>
-                <include>commons-logging:commons-logging:jar</include>
```
Aggarwal-Raghav, line 51: removing commons-logging as it was getting packaged in beeline tarball. Check [HIVE-24691](https://issues.apache.org/jira/browse/HIVE-24691) and [HIVE-20019](https://issues.apache.org/jira/browse/HIVE-20019)

packaging/src/main/assembly/beeline.xml
```diff
@@ -48,8 +48,7 @@
...
-                <include>jline:jline:jar</include>
```
Aggarwal-Raghav, line 52: moved it to jline3.x groupId

standalone-metastore/metastore-common/pom.xml
```diff
@@ -564,29 +560,6 @@
...
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-enforcer-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>enforce-banned-dependencies</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <configuration>
```
Aggarwal-Raghav, line 576: removed the enforcer pluging from `standalone-metastore/metastore-common/pom.xml` and `standalone-metastore/metastore-server/pom.xml` and moved to `standalone-metastore/pom.xml` i.e. a common place and updated it with the same list of banned dependency as in parent `pom.xml`

Aggarwal-Raghav:
1. **The unit test failures in this PR are relevant.** The reason for this is that Pig version 0.16.0 uses jline 1.x. The jline dependency is being excluded from Pig due to an enforcer rule, which is causing failures specifically in the `hcatalog/hcatalog-pig-adapter` module.

2. I believe the current unit test scenario in the cluster/runtime environment might also be failing because we are not shipping jline 1.x. _Assumption_: This would only work if jline 2.x, which is being shipped without this PR, is backward compatible with version 1.x.

My recommendation for the next step is to remove the jline enforcer rule from the parent pom.xml, keeping the exclusions everywhere else except `hcatalog/hcatalog-pig-adapter`. Is that the approach acceptable? Still with this approach [2] i.e. pig working in cluster is unknown.

[possible_follow_up.patch](https://github.com/user-attachments/files/22110301/possible_follow_up.patch)

CC @abstractdog

ayushtkn:
> As Hadoop has jline 3.9.0**, in my setup** it was getting picked first compared to 3.25.0 from hive. Causing NoSuchMethodError org.jline.reader.EndOfFileException#getPartialLine() check [here](https://issues.apache.org/jira/browse/HIVE-29130?focusedCommentId=18014311&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18014311).

Hadoop's JLine 3.9.0 it was getting picked up first. So, we should make sure Hive Classpath is always before HADOOP jars.

I don't understand in that scope or for this problem why are we playing with Jline-2.x? 

I am not very convinced on enforcing some places and not on some. We might not have a use case or expertise with Pig, so better not to touch it in that case. We had a thread around Pig to depreacate/remove some time back & we were shot down, so people tend to use it

Why not upgrade JLine in Hadoop?

abstractdog:
> 1. **The unit test failures in this PR are relevant.** The reason for this is that Pig version 0.16.0 uses jline 1.x. The jline dependency is being excluded from Pig due to an enforcer rule, which is causing failures specifically in the `hcatalog/hcatalog-pig-adapter` module.
> 2. I believe the current unit test scenario in the cluster/runtime environment might also be failing because we are not shipping jline 1.x. _Assumption_: This would only work if jline 2.x, which is being shipped without this PR, is backward compatible with version 1.x.
> 
> My recommendation for the next step is to remove the jline enforcer rule from the parent pom.xml, keeping the exclusions everywhere else except `hcatalog/hcatalog-pig-adapter`. Is that the approach acceptable? Still with this approach [2] i.e. pig working in cluster is unknown.
> 
> [possible_follow_up.patch](https://github.com/user-attachments/files/22110301/possible_follow_up.patch)
> 
> CC @abstractdog

it is what it is, thanks @Aggarwal-Raghav for checking, but before proceeding with this, can you please check if there is an include/exclude possibility in bannedDependencies to allow old jline transitively through pig only, does it make sense?

Aggarwal-Raghav:
Thanks for the reply @ayushtkn.

> I don't understand in that scope or for this problem why are we playing with Jline-2.x?

In hive packaging, both jline 2.11 and jline 3.25.0 are getting shipped. IMO, its wrong. The aim of this PR is to ship only jline3.x jar. If we are ok shipping both the version then I'll close this PR. 

> Why not upgrade JLine in Hadoop?

It's good to have same dependency version across components. In my setup, Jline 3.9.0 is getting picked up first which is causing NoSuchMethodError. --  _I want to confirm if its due to my setup or not._

In conclusion,

- if Hadoop Classpath first, then jline 3.9.0 is picked => hadoop has to upgrade jline to 3.25.0
- if hive classpath is first, then we are all good.

ayushtkn:
> In conclusion,
if Hadoop Classpath first, then jline 3.9.0 is picked => hadoop has to upgrade jline to 3.25.0
if hive classpath is first, then we are all good.

So, the solution is like in ideal setup Hive Classpath should be first. In that case we are sorted. Else for the problem that you reported. Hadoop needs to upgrade JLine, correct?

And this PR is solving a different problem as reported in the ticket, this won't solve the problem that you reported?

Aggarwal-Raghav:
> > In conclusion,
> > if Hadoop Classpath first, then jline 3.9.0 is picked => hadoop has to upgrade jline to 3.25.0
> > if hive classpath is first, then we are all good.
> 
> So, the solution is like in ideal setup Hive Classpath should be first. In that case we are sorted. Else for the problem that you reported. Hadoop needs to upgrade JLine, correct?
> 
> And this PR is solving a different problem as reported in the ticket, this won't solve the problem that you reported?

yes and yes.
In addition, I found 2 problems:
1. commons-logggins is also getting shipped which is big no-no based on [HIVE-24691](https://issues.apache.org/jira/browse/HIVE-24691) and [HIVE-20019](https://issues.apache.org/jira/browse/HIVE-20019)
2. enforcer plugin defined in parent pom is not enforced on standalone-metastore module. IMO, the banned dependencies should be same.

Aggarwal-Raghav:
> can you please check if there is an include/exclude possibility in bannedDependencies to allow old jline transitively through pig only, does it make sense?

we can use `include` tag but there are 2 ways:
**1.** Do it on parent pom level: [include-tag.patch](https://github.com/user-attachments/files/22115645/include-tag.patch)
- Benefit: Single place enforcer is defined. So, easy to maintain.
- Drawback: For all the module jline 1.0 is enabled. But it won't be shipped in packaging. As we have excluded it from all problematic places flagged by enforcer plugin so it shouldn't be a problem.
 
**2.** Do it on hcatalog/hcatalog-pig-adapter/pom.xml level: [include-tag-child-pom.patch](https://github.com/user-attachments/files/22115654/include-tag-child-pom.patch)
- Benefit: Only for hcatalog/hcatalog-pig-adapter jline 1.0 is enabled, rest all modules are not impacted.
- Drawback: Copy of enforcer plugin needs to be defined and maintainability of enforcer plugin for new bannedDependencies can be missed for this child pom.

@abstractdog , please let me know which is better.

abstractdog:
> > can you please check if there is an include/exclude possibility in bannedDependencies to allow old jline transitively through pig only, does it make sense?
> 
> we can use `include` tag but there are 2 ways: **1.** Do it on parent pom level: [include-tag.patch](https://github.com/user-attachments/files/22115645/include-tag.patch)
> 
> * Benefit: Single place enforcer is defined. So, easy to maintain.
> * Drawback: For all the module jline 1.0 is enabled. But it won't be shipped in packaging. As we have excluded it from all problematic places flagged by enforcer plugin so it shouldn't be a problem.
> 
> **2.** Do it on hcatalog/hcatalog-pig-adapter/pom.xml level: [include-tag-child-pom.patch](https://github.com/user-attachments/files/22115654/include-tag-child-pom.patch)
> 
> * Benefit: Only for hcatalog/hcatalog-pig-adapter jline 1.0 is enabled, rest all modules are not impacted.
> * Drawback: Copy of enforcer plugin needs to be defined and maintainability of enforcer plugin for new bannedDependencies can be missed for this child pom.
> 
> @abstractdog , please let me know which is better.

I would go for 1), I don't like the whole copied enforcer plugin config in 2), thanks a lot for investigating all the possibilities!

Aggarwal-Raghav:
UT passed in local. 
```
cd hcatalog/hcatalog-pig-adapter ; mvn clean test -Dtest=TestTextFileHCatLoader,TestRCFileHCatLoader,TestAvroHCatLoader,TestSequenceFileHCatLoader,TestParquetHCatLoader,TestOrcHCatLoader -Drat.skip
```
Pushed the changes. Will see CI output.

pom.xml
```diff
@@ -1715,6 +1715,16 @@
                   <searchTransitive>true</searchTransitive>
                   <message>Banned log4j:log4j dependency/transitive dependency was found!</message>
                 </bannedDependencies>
+                <bannedDependencies>
+                  <excludes>
+                    <exclude>jline:jline</exclude>
+                  </excludes>
+                  <includes>
+                    <include>jline:jline:1.0</include>
```
abstractdog, line 1746: nit: please explain here with a brief comment why we still allow 1.0 to make it clear for the future if we can remove

Aggarwal-Raghav: Done

abstractdog:
thanks @Aggarwal-Raghav, I'm good now with the pom.xml changes, but haven't followed if you've reached consensus with @ayushtkn on the other aspect of this PR, please let me know guys

Aggarwal-Raghav:
@ayushtkn, can you please take a look once!

Aggarwal-Raghav:
gentle ping  for review @ayushtkn !

ayushtkn:
Stuck with some stuff. If @abstractdog is happy. I am happy. feel free to move ahead :-)

Aggarwal-Raghav:
Forced pushed to resolve merge conflicts

Aggarwal-Raghav:
@abstractdog , can you please help in merging this?

PR #6055, opened 2025-09-02
HIVE-29175: Wrong Anti join conversion leading to missing results

mdayakar:
As a part of anti join conversion, if the join condition operands are from left table only then not converting it to anti join to avoid wrong results.

To fix the wrong results.

No

Using q file tests
_mvn -Dtest=TestMiniLlapLocalCliDriver -Dqfile=antijoin_conversion.q -pl itests/qtest -Pitests_

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
@@ -1273,12 +1273,12 @@ public class HiveCalciteUtil {
   }
 
   /**
-   * Checks the operands in the join conditions are from right side or only from left side.
+   * Checks the operands in the join conditions are only from left side.
    *
    * @param joinRel Join node
    * @return true if the join condition operands are from right and left side, false otherwise.
    */
-  public static boolean hasJoinConditionOperandsFromLeftAndRightSide(Join joinRel) {
+  public static boolean checkIfJoinConditionOnlyUsesLeftOperands(Join joinRel) {
     RexNode condition = joinRel.getCondition();
     RelNode leftRel = joinRel.getLeft();
     int leftFieldCount = leftRel.getRowType().getFieldCount();
@@ -1288,10 +1288,10 @@ public class HiveCalciteUtil {
       ImmutableBitSet condBitmap = RelOptUtil.InputFinder.bits(cond);
       // here condition becomes true if both the operands are from left table
       if (leftBitmap.contains(condBitmap)) {
-        return false;
+        return true;
       }
     }
-    return true;
+    return false;
   }
 
   /**
--- a/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAntiSemiJoinRule.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAntiSemiJoinRule.java
@@ -98,7 +98,7 @@ public class HiveAntiSemiJoinRule extends RelOptRule {
     }
 
     // if one of the operand from join condition is not from right table then no need to convert to anti join.
-    if (!HiveCalciteUtil.hasJoinConditionOperandsFromLeftAndRightSide(join)) {
+    if (HiveCalciteUtil.checkIfJoinConditionOnlyUsesLeftOperands(join)) {
       return;
     }
 


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAntiSemiJoinRule.java
```diff
@@ -97,6 +97,12 @@ protected void perform(RelOptRuleCall call, Project project, Filter filter, Join
       return;
     }
 
+    // if one of the operand from join condition is not from right table then no need to convert to anti join.
+    boolean hasConditionMet = HiveCalciteUtil.hasJoinConditionOperandFromRightTable(join);
+    if (!hasConditionMet) {
```
kasakrisz: The local variable `hasConditionMet` can be removed and the function call `hasJoinConditionOperandFromRightTable` can be inlined.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
```diff
@@ -1272,6 +1272,28 @@ public static boolean hasAllExpressionsFromRightSide(RelNode joinRel, List<RexNo
...
+   */
+  public static boolean hasJoinConditionOperandFromRightTable(Join joinRel) {
+    RexNode condition = joinRel.getCondition();
+    RelNode leftRel = joinRel.getLeft();
+    int leftFieldCount = leftRel.getRowType().getFieldCount();
+    ImmutableBitSet leftBitmap = ImmutableBitSet.range(leftFieldCount);
+    List<RexNode> conditions = RelOptUtil.conjunctions(condition);
+    for (RexNode cond : conditions) {
+      ImmutableBitSet condBitmap = RelOptUtil.InputFinder.bits(cond);
+      // if condition becomes true if both the operands are from left table
```
kasakrisz: Typo: if ... if

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
```diff
@@ -1272,6 +1272,28 @@ public static boolean hasAllExpressionsFromRightSide(RelNode joinRel, List<RexNo
...
+   * Checks the operands in the join conditions are from right table or only from left table.
+   *
+   * @param joinRel Join node
+   * @return true if the join condition operands are from right and left table, false otherwise.
+   */
+  public static boolean hasJoinConditionOperandFromRightTable(Join joinRel) {
+    RexNode condition = joinRel.getCondition();
+    RelNode leftRel = joinRel.getLeft();
+    int leftFieldCount = leftRel.getRowType().getFieldCount();
+    ImmutableBitSet leftBitmap = ImmutableBitSet.range(leftFieldCount);
```
kasakrisz: nit: 
1. The method name says that it looks for references in the join condition coming from the right side. However the implementation search for references in the left side only.
It seems to be valid to assume that if a reference is not in the left side then it must be in the right. It just not what I expect when reading the method name and the implementation.

2. Is the right side always a table (TableScan), or can it be any subtree? If it’s the latter, then the word Table is misleading.

ql/src/test/queries/clientpositive/antijoin_conversion.q
```diff
@@ -20,3 +20,31 @@ select n.* from n left outer join t on (n.a=t.a) where cast(t.a as float) is nul
...
 
 
+create table tab1 (col1 int, col2 int, col3 int, col4 int);
+create table tab2 (col1 int, col2 int, col3 int, col4 int);
+
+insert into tab1 values (123, 1000, 5000, 9), (456, 1000, 7000, 7), (789, 1000, 5000, 8);
+insert into tab2 values (123, 1000, 5000, 2), (456, 1000, 7000, 7), (123, 5000, 4000, 2);
+
+select t1.col1, t1.col2, t1.col3 from tab2 t1
+left join tab1 t2
```
kasakrisz: Could you please swap the aliases so let
tab1 t1 and tab2 t2

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAntiSemiJoinRule.java
mdayakar: Modified code accordingly.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
mdayakar: Modified the comment

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
mdayakar: Changed the method name accordingly.

ql/src/test/queries/clientpositive/antijoin_conversion.q
mdayakar: Modified it.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveCalciteUtil.java
```diff
@@ -1272,6 +1272,28 @@ public static boolean hasAllExpressionsFromRightSide(RelNode joinRel, List<RexNo
     return true;
   }
 
+  /**
+   * Checks the operands in the join conditions are from right side or only from left side.
+   *
+   * @param joinRel Join node
+   * @return true if the join condition operands are from right and left side, false otherwise.
+   */
+  public static boolean hasJoinConditionOperandsFromLeftAndRightSide(Join joinRel) {
```
thomasrebele: The name of the method does not match what it does, as it checks for the LHS only.
I was wondering what happens in SQL when the filter is on the RHS only (e.g., t3.col2=t3.col2 in the scenario of the qfile). It behaves differently when moving the condition from the ON clause to the WHERE clause: [SQL fiddle](https://sqlfiddle.com/postgresql/online-compiler?id=074f8ebd-a80b-4d0f-8cb2-428a4310af68). Maybe we need to check whether Hive's AntiJoin works correctly in such a case.

An alternative name for the method is hasNoJoinConditionConjunctOnLeftSideOnly. Though maybe we can come up with something better.

mdayakar: Method name is changed to `checkIfJoinConditionOnlyUsesLeftOperands`, I feel this is ok which resembles the method implementation logic.

> I was wondering what happens in SQL when the filter is on the RHS only (e.g., t3.col2=t3.col2 in the scenario of the qfile). It behaves differently when moving the condition from the ON clause to the WHERE clause: [SQL fiddle](https://sqlfiddle.com/postgresql/online-compiler?id=074f8ebd-a80b-4d0f-8cb2-428a4310af68). Maybe we need to check whether Hive's AntiJoin works correctly in such a case.

In the above case it will not be converted to anti join. May be we can check the Hive's behaviour separately in the above case, if found any issue a bug can be raised and fixed.

PR #6056, opened 2025-09-03
AntiJoin rule fails to apply with an extra constant in the SELECT clause

thomasrebele:
Fix a bug in HiveCalciteUtil#hasAnyExpressionFromRightSide.

Adding constants to the SELECT clause should not prevent the Antijoin rule from doing its job

No

Extended antijoin.q: added a check for the query result and the CBO explanation


thomasrebele:
Closing this PR as I've decided to fix some other anti-join bug as well, see [HIVE-29176](https://issues.apache.org/jira/browse/HIVE-29176).

PR #6057, opened 2025-09-03
Set hive.security.authorization.enabled to true by default

zabetak:
In most real deployments the authorization is always enabled so it makes sense to set the default value to true. Changing the default is important for testing purposes since our testing should be focused on the most common deployment scenarios rather than the exceptions.

Yes/TBA

Existing tests


zabetak:
This an ambitious and probably useful change but it requires a lot of work and currently don't have enough bandwidth to pull this to completion.

PR #6058, opened 2025-09-03
HIVE-29179: Make the RELOAD event backward compatible after latest changes via HI…

saihemanth-cloudera:
…VE-28967

Making the RELOAD event backward compatible

Clients like impala are unable to use HIVE-28967

No

Remote cluster with Impala enabled

### DIFF
--- a/hcatalog/server-extensions/src/main/java/org/apache/hive/hcatalog/listener/DbNotificationListener.java
+++ b/hcatalog/server-extensions/src/main/java/org/apache/hive/hcatalog/listener/DbNotificationListener.java
@@ -1023,7 +1023,7 @@ public class DbNotificationListener extends TransactionalMetaStoreEventListener 
   public void onReload(ReloadEvent reloadEvent) throws MetaException {
     Table tableObj = reloadEvent.getTableObj();
     ReloadMessage msg = MessageBuilder.getInstance().buildReloadMessage(tableObj,
-            reloadEvent.getPartitions(), reloadEvent.isRefreshEvent());
+        reloadEvent.getPartitionObj(), reloadEvent.getPartitions(), reloadEvent.isRefreshEvent());
     NotificationEvent event =
             new NotificationEvent(0, now(), EventType.RELOAD.toString(),
                     msgEncoder.getSerializer().serialize(msg));
--- a/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/messaging/json/JSONReloadMessage.java
+++ b/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/messaging/json/JSONReloadMessage.java
@@ -41,7 +41,7 @@ public class JSONReloadMessage extends ReloadMessage {
     private Long timestamp;
 
     @JsonProperty
-    private String server, servicePrincipal, db, table, tableObjJson, refreshEvent;
+    private String server, servicePrincipal, db, table, tableObjJson, ptnObjJson, refreshEvent;
 
     @JsonProperty
     List<String> partitionListJson;
@@ -52,8 +52,8 @@ public class JSONReloadMessage extends ReloadMessage {
     public JSONReloadMessage() {
     }
 
-    public JSONReloadMessage(String server, String servicePrincipal, Table tableObj, List<Partition> ptns,
-                             boolean refreshEvent, Long timestamp) {
+    public JSONReloadMessage(String server, String servicePrincipal, Table tableObj, Partition ptnObj,
+                             List<Partition> ptns, boolean refreshEvent, Long timestamp) {
         this.server = server;
         this.servicePrincipal = servicePrincipal;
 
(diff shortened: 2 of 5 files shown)


standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/messaging/json/JSONReloadMessage.java
```diff
@@ -133,6 +138,11 @@ public Partition apply(@Nullable Object input) {
             });
     }
 
+    @Override
+    public Partition getPtnObj() throws Exception {
```
dengzhhu653, line 142: this method might return null somehow, will it break the compatibility? should we catch it on the impala side?

saihemanth-cloudera: I have added necessary guardrails on the Impala side to avoid this null pointer exception. i.e., if getPtnObj() returns null then we should check getPartitionObjs() method.

dengzhhu653: I have a concern here, if we can add the NPE check on the impala side, why do we still have the compatibility issue? 
Is it for the old message?

hcatalog/server-extensions/src/main/java/org/apache/hive/hcatalog/listener/DbNotificationListener.java
```diff
@@ -1023,7 +1023,7 @@ public void onDeletePartitionColumnStat(DeletePartitionColumnStatEvent deletePar
   public void onReload(ReloadEvent reloadEvent) throws MetaException {
     Table tableObj = reloadEvent.getTableObj();
     ReloadMessage msg = MessageBuilder.getInstance().buildReloadMessage(tableObj,
-            reloadEvent.getPartitions(), reloadEvent.isRefreshEvent());
+        reloadEvent.getPartitionObj(), reloadEvent.getPartitions(), reloadEvent.isRefreshEvent());
```
dengzhhu653, line 1026: nit: the `reloadEvent.getPartitionObj()` should always be null?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/messaging/json/JSONReloadMessage.java
dengzhhu653: If the above is true, then we can remove this `getPtnObj`, the client can still use getPartitionObjs() to retrieve the old `ptnObjJson`

stiga-huang: Impala is currently compiled with `getPtnObj()` so we have to add back this method for compatibility:
https://github.com/apache/impala/blob/ebbc67cf40bd856253d07c649028888d85c772cc/fe/src/compat-hive-3/java/org/apache/impala/compat/MetastoreShim.java#L600

When it returns null, the RELOAD event will be performed on the whole table:
https://github.com/apache/impala/blob/ebbc67cf40bd856253d07c649028888d85c772cc/fe/src/main/java/org/apache/impala/catalog/events/MetastoreEvents.java#L3450-L3454
I think this is fine since if the RELOAD event is on multiple partitions, reloading the whole table makes sure they are reloaded.

dengzhhu653: I have a little confusion here, if the getPtnObj() is null, then we would reload the whole table, would this ignore the partitions(the `ptns`) in this reload event?

stiga-huang: Yeah, in the current code of Impala, if there are multiple partitions in the RELOAD event, they (`ptns`) are ignored since the new API `getPartitionObjs()` is not used yet. The RELOAD event is then treated as a whole-table reload. That's OK since reloading the whole table also reloads these changed partitions. Though it has some redundant work of reloading other unchanged partitions, it ensures correctness, i.e. get rid of stale metadata cache.

dengzhhu653: @saihemanth-cloudera should we consider this https://github.com/apache/hive/pull/6058#discussion_r2370964841, and remove the always null parameter?

hcatalog/server-extensions/src/main/java/org/apache/hive/hcatalog/listener/DbNotificationListener.java
saihemanth-cloudera: It'll not be null due to: https://github.com/apache/hive/pull/6058/files#diff-0a5ff4393fcf55728e17e9a91fe9f2293a526e7fbf720e8a098d793eb122555aR74

dengzhhu653 (approved):

dengzhhu653:
Thank you for the PR and clarification, @saihemanth-cloudera and @stiga-huang

PR #6060, opened 2025-09-04
HIVE-29181: SELECT query on VIEW with IS NOT NULL operator producing unexpected result

mdayakar:
HIVE-29181: SELECT query on VIEW with IS NOT NULL operator producing unexpected result

Here when the data is selected from a view with `(([1008753865](tel:1008753865))/(0)) IS NOT NULL` filter condition which actually results to FALSE should not return any rows but actually it us returning rows, where as with same filter condition selecting data from a table working fine, not giving any rows as a result.

This is due to `HiveFilterProjectTransposeRule` rule which is the first rule in the list is getting applied for view query and finally the filter condition is getting removed so it is returning the rows where as for table query `ReduceExpressionsRule.FilterReduceExpressionsRule` is getting applied and the plan is getting changed to `HiveValues(tuples=[[]])` so no rows are getting fetched. 

So to fix the issue `ReduceExpressionsRule.FilterReduceExpressionsRule` can be added before `HiveFilterProjectTransposeRule` so that `ReduceExpressionsRule.FilterReduceExpressionsRule` will get applied for view query also and gives proper output.

To fix the issue mentioned in HIVE-29181

No

Using q file tests
_mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=view_with_where_exp.q -pl itests/qtest -Pitests_

### DIFF
--- a/ql/src/test/results/clientpositive/llap/ppd_gby_join.q.out
+++ b/ql/src/test/results/clientpositive/llap/ppd_gby_join.q.out
@@ -39,21 +39,21 @@ STAGE PLANS:
             Map Operator Tree:
                 TableScan
                   alias: src
-                  filterExpr: (((value < 'val_50') or (key > '2')) and (((key > '20') and (key < '4')) or ((key > '4') and (key < '400')))) (type: boolean)
+                  filterExpr: (((value < 'val_50') or key is not null) and (((key > '20') and (key < '4')) or ((key > '4') and (key < '400')))) (type: boolean)
                   Statistics: Num rows: 500 Data size: 89000 Basic stats: COMPLETE Column stats: COMPLETE
                   Filter Operator
-                    predicate: (((value < 'val_50') or (key > '2')) and (((key > '20') and (key < '4')) or ((key > '4') and (key < '400')))) (type: boolean)
-                    Statistics: Num rows: 72 Data size: 12816 Basic stats: COMPLETE Column stats: COMPLETE
+                    predicate: (((value < 'val_50') or key is not null) and (((key > '20') and (key < '4')) or ((key > '4') and (key < '400')))) (type: boolean)
+                    Statistics: Num rows: 110 Data size: 19580 Basic stats: COMPLETE Column stats: COMPLETE
                     Select Operator
                       expressions: key (type: string)
                       outputColumnNames: _col0
-                      Statistics: Num rows: 72 Data size: 6264 Basic stats: COMPLETE Column stats: COMPLETE
+                      Statistics: Num rows: 110 Data size: 9570 Basic stats: COMPLETE Column stats: COMPLETE
                       Reduce Output Operator
                         key expressions: _col0 (type: string)
                         null sort order: z
                         sort order: +
                         Map-reduce partition columns: _col0 (type: string)
-                        Statistics: Num rows: 72 Data size: 6264 Basic stats: COMPLETE Column stats: COMPLETE
+                        Statistics: Num rows: 110 Data size: 9570 Basic stats: COMPLETE Column stats: COMPLETE
             Execution mode: vectorized, llap
             LLAP IO: all inputs
         Map 4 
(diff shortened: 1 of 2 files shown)


ql/src/test/queries/clientpositive/view_with_where_exp.q
thomasrebele, line 1: Could we simplify the test case, please? E.g., by removing unnecessary columns, using simpler constants. The following test case reproduces the issue:

```
CREATE TABLE IF NOT EXISTS t0(col DOUBLE);

CREATE VIEW v0 AS (SELECT ALL (t0.col) AS col FROM t0);

INSERT INTO t0(col) VALUES(0.1);

explain cbo SELECT t0.col from t0 WHERE ((1000)/(0)) IS NOT NULL;
SELECT t0.col from t0 WHERE ((1000)/(0)) IS NOT NULL;

explain cbo SELECT v0.col FROM v0 WHERE ((1000)/(0)) IS NOT NULL;
SELECT v0.col FROM v0 WHERE ((1000)/(0)) IS NOT NULL;
```

ql/src/test/results/clientpositive/llap/explainuser_1.q.out
```diff
@@ -1331,9 +1331,9 @@ Stage-0
     Stage-1
       Reducer 3 llap
       File Output Operator [FS_16]
-        Select Operator [SEL_15] (rows=24 width=100)
+        Select Operator [SEL_15] (rows=21 width=100)
```
thomasrebele, line 1334: I guess the row count are just estimations, not the real row count of the executed query?

zabetak:
The description mentions that the filter condition is removed at some point. Who does the removal and why? It feels wrong to remove a filter that is always `false` so I want to understand a bit better what happens.

ql/src/test/queries/clientpositive/view_with_where_exp.q
zabetak: I am also curious to know if the datatype (i.e., `DOUBLE`) is important for reproducing the problem. IF not then probably I would pick something more straightforward like an `INT` or `STRING`.

ql/src/test/results/clientpositive/llap/ppd_gby_join.q.out
```diff
@@ -322,21 +322,21 @@ STAGE PLANS:
             Map Operator Tree:
                 TableScan
                   alias: src
-                  filterExpr: (((value < 'val_50') or (key > '2')) and (((key > '20') and (key < '4')) or ((key > '4') and (key < '400')))) (type: boolean)
+                  filterExpr: (((value < 'val_50') or key is not null) and (((key > '20') and (key < '4')) or ((key > '4') and (key < '400')))) (type: boolean)
```
zabetak, line 325: Is the simplification of `key > '2'` to `key is not null` valid?

ql/src/test/results/clientpositive/llap/subquery_notin.q.out
```diff
@@ -5300,20 +5300,27 @@ STAGE PLANS:
...
-                      Statistics: Num rows: 24 Data size: 4704 Basic stats: COMPLETE Column stats: COMPLETE
-                      Reduce Output Operator
-                        key expressions: _col0 (type: string), _col1 (type: string)
-                        null sort order: zz
-                        sort order: ++
-                        Map-reduce partition columns: _col0 (type: string), _col1 (type: string)
-                        Statistics: Num rows: 24 Data size: 4704 Basic stats: COMPLETE Column stats: COMPLETE
+                      Statistics: Num rows: 26 Data size: 5096 Basic stats: COMPLETE Column stats: COMPLETE
+                      Filter Operator
+                        predicate: UDFToDouble(_col0) is not null (type: boolean)
```
zabetak, line 5310: Rather minor but it seems that now some filter operators cannot be merged together.

PR #6061, opened 2025-09-04
HIVE-25952: Drop HiveRelMdPredicates::getPredicates(Project) in favor of RelMdPredicates

zabetak:


zabetak:
Technically it could be merged but for the reasons mentioned in the JIRA ticket we can wait a bit.

PR #6062, opened 2025-09-04
HIVE-29176: Wrong result when HiveAntiJoin is replacing an IS NULL filter on a nullable column

thomasrebele:
Improve the logic of HiveAntiSemiJoinRule to fix the problems mentioned in [HIVE-29176](https://issues.apache.org/jira/browse/HIVE-29176).

Wrong query plans introduced by HiveAntiSemiJoinRule.

No

Unit tests and qfile tests.

### DIFF
--- a/ql/src/test/results/clientpositive/llap/antijoin3.q.out
+++ b/ql/src/test/results/clientpositive/llap/antijoin3.q.out
@@ -177,7 +177,7 @@ POSTHOOK: Input: default@tabb
 POSTHOOK: Input: default@tabc
 #### A masked pattern was here ####
 CBO PLAN:
-HiveProject(id=[$1], id1=[null:VARCHAR(2147483647) CHARACTER SET "UTF-16LE"], id2=[$0])
+HiveProject(a.id=[$1], b.id=[null:VARCHAR(2147483647) CHARACTER SET "UTF-16LE"], c.id=[$0])
   HiveJoin(condition=[=($0, $1)], joinType=[right], algorithm=[none], cost=[not available])
     HiveProject(id=[$0])
       HiveFilter(condition=[IS NOT NULL($0)])


ql/src/test/queries/clientpositive/antijoin3.q
```diff
@@ -0,0 +1,39 @@
...
+(3, "val_3", "val_3"),
+(4, "val_4", "val_4");
+
+insert into antijoin3_t2 values
+(0, "val_0", null),
+(1, "val_1", null),
+(4, "val_4", "val_4"),
+(5, "val_5", "val_5");
+
+-- do not introduce anti-join if filtering a nullable column with IS NULL
```
amansinha100: One more suggestion for a test case would where the right column is not coming from a base table but from another Left Outer Join  so it becomes nullable even if it was nonnullable in the base table.  This is just to check if the nullable property is being properly propagated up the operators and if so, this PR's changes would work fine.  If for some reason it is not getting propagated, it could uncover a hidden bug.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAntiSemiJoinRule.java
```diff
@@ -135,43 +142,76 @@ private List<RexNode> getResidualFilterNodes(Filter filter, Join join) {
...
+      // In the following we check for filter nodes that let us deduce that
+      // "an (originally) not-null column of RHS IS NULL because the LHS row will not be matched"
+
+      boolean usesRHSFieldsOnly = rhsFields.contains(usedFields);
+      if (!usesRHSFieldsOnly) {
+        // If there is a mix between LHS and RHS fields, don't convert to anti-join
+        return Optional.empty();
+      }
+
+      if(filterNode.getKind() != SqlKind.IS_NULL) {
```
amansinha100: Since this is a quicker conditional check, it would be better to move this up a bit before line 160 so we can avoid executing the contains() if possible.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAntiSemiJoinRule.java
```diff
@@ -135,43 +142,76 @@ private List<RexNode> getResidualFilterNodes(Filter filter, Join join) {
...
+      if(!field.getType().isNullable()) {
+        builder.set(shift+field.getIndex());
+      }
     }
-    return filterList;
+    return builder.build();
   }
 
-  private boolean isStrong(RexNode rexNode) {
-    AtomicBoolean hasCast = new AtomicBoolean(false);
```
amansinha100, line 170: Previously, since AtomicBoolean was used, I am wondering there might have been a good reason.  Was there a concurrency issue that motivated the original implementation.  Changing it to a MutableBoolean means that this variable is not thread safe.

ql/src/test/queries/clientpositive/antijoin3.q
thomasrebele: Calcite handles the nullability propagation, so it should be correct. I've added a test query to be sure.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAntiSemiJoinRule.java
thomasrebele: Fixed.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HiveAntiSemiJoinRule.java
thomasrebele: A RexVisitor is single-threaded and the `hasCast` is a local variable, so no risk of concurrency issues. See `org.apache.calcite.rex.RexChecker#failCount` for a similar non-synchronized usage in Calcite.

I've stumbled upon `org.apache.calcite.util.Util.FoundOne`, see for example `org.apache.calcite.rex.RexUtil#isDeterministic`, which should be thread safe. I've replaced the hasCast variable with the FoundOne to make it future proof.

amansinha100: Yes, I couldn't think of a compelling reason why AtomicBoolean was used in the first place. The updated change looks good.

thomasrebele:
Some tests had been failing (antijoin.q, empty_result_outerjoin.q). I've fixed it by improving the logic that deduces which columns of the RHS are definitively NOT NULL.

_Update_: some other tests are failing now, I still need to fix those.

thomasrebele:
Tests are green. @amansinha100, if you have some time, would you mind reviewing the changes, please?

kasakrisz (approved):
LGTM

PR #6064, opened 2025-09-05
HIVE-28551 Stale results when executing queries over recreated transactional tables

thomasrebele:
A query result cache entry may only match if the table ids are the same for all tables used by the query.
This PR continues the work of an old PR: https://github.com/apache/hive/pull/5482.

Otherwise we get unexpected results, see [HIVE-28551](https://issues.apache.org/jira/browse/HIVE-28551).

No

qfile tests

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
@@ -65,7 +65,6 @@ import org.apache.hadoop.hive.metastore.messaging.MessageBuilder;
 import org.apache.hadoop.hive.ql.hooks.Entity.Type;
 import org.apache.hadoop.hive.ql.hooks.ReadEntity;
 import org.apache.hadoop.hive.ql.io.AcidUtils;
-import org.apache.hadoop.hive.ql.metadata.Hive;
 import org.apache.hadoop.hive.ql.metadata.SessionHiveMetaStoreClient;
 import org.apache.hadoop.hive.ql.metadata.Table;
 import org.apache.hadoop.hive.ql.metadata.events.EventConsumer;
@@ -86,13 +85,15 @@ public final class QueryResultsCache {
   private static final Logger LOG = LoggerFactory.getLogger(QueryResultsCache.class);
 
   public static class LookupInfo {
-    private String queryText;
-    private Supplier<ValidTxnWriteIdList> txnWriteIdListProvider;
+    private final String queryText;
+    private final Supplier<ValidTxnWriteIdList> txnWriteIdListProvider;
+    private final Set<Long> tableIds;
 
-    public LookupInfo(String queryText, Supplier<ValidTxnWriteIdList> txnWriteIdListProvider) {
+    public LookupInfo(String queryText, Supplier<ValidTxnWriteIdList> txnWriteIdListProvider, Set<Long> tableIds) {
       super();
       this.queryText = queryText;
       this.txnWriteIdListProvider = txnWriteIdListProvider;
+      this.tableIds = tableIds;
     }
 
     public String getQueryText() {
@@ -429,7 +430,7 @@ public final class QueryResultsCache {
     Set<CacheEntry> entriesToRemove = new HashSet<CacheEntry>();
     Lock readLock = rwLock.readLock();
     try {
-      // Note: ReentrantReadWriteLock deos not allow upgrading a read lock to a write lock.
+      // Note: ReentrantReadWriteLock does not allow upgrading a read lock to a write lock.
       // Care must be taken while under read lock, to make sure we do not perform any actions
       // which attempt to take a write lock.
       readLock.lock();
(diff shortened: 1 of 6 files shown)


ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
```diff
@@ -671,10 +672,15 @@ public void notifyTableChanged(String dbName, String tableName, long updateTime)
...
   private boolean entryMatches(LookupInfo lookupInfo, CacheEntry entry, Set<CacheEntry> entriesToRemove) {
     QueryInfo queryInfo = entry.getQueryInfo();
+
+    Set<Long> cacheTableIds = new HashSet<>();
     for (ReadEntity readEntity : queryInfo.getInputs()) {
-      // Check that the tables used do not resolve to temp tables.
       if (readEntity.getType() == Type.TABLE) {
         Table tableUsed = readEntity.getTable();
+        // collect the table ids of the cache entry
+        cacheTableIds.add(tableUsed.getTTable().getId());
```
zabetak: Do we need to collect **all** IDs in a temporary set? Can't we check if the current table ID is part of the `lookupInfo` and bail out immediately without going through all tables in this `entry`?

ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
```diff
@@ -671,10 +672,15 @@ public void notifyTableChanged(String dbName, String tableName, long updateTime)
    */
   private boolean entryMatches(LookupInfo lookupInfo, CacheEntry entry, Set<CacheEntry> entriesToRemove) {
     QueryInfo queryInfo = entry.getQueryInfo();
+
+    Set<Long> cacheTableIds = new HashSet<>();
     for (ReadEntity readEntity : queryInfo.getInputs()) {
-      // Check that the tables used do not resolve to temp tables.
       if (readEntity.getType() == Type.TABLE) {
         Table tableUsed = readEntity.getTable();
```
zabetak, line 678: Is there a way to use the table ID to invalidate an cache entry? We we are doing some kind of invalidation below based on writeids so I was wondering if we could/should do something similar when a table is dropped. This is not strictly related to the stale results problem so we can certainly defer it to a follow-up if it makes sense.

ql/src/test/queries/clientpositive/results_cache_invalidation4.q
```diff
@@ -0,0 +1,16 @@
+set hive.support.concurrency=true;
+set hive.txn.manager=org.apache.hadoop.hive.ql.lockmgr.DbTxnManager;
+
+set hive.query.results.cache.enabled=true;
+set hive.query.results.cache.nontransactional.tables.enabled=true;
```
zabetak, line 5: Just out of curiosity did you confirm that bug is also affecting nontransactional tables?

ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
thomasrebele: Ok, I guess we can assume there is a mapping id -> table, i.e., no two tables can have the same id. I'll update the PR.

ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
thomasrebele: I would move this to a follow-up ticket: HIVE-29193

ql/src/test/queries/clientpositive/results_cache_invalidation4.q
thomasrebele: Yes, this bug is also affecting nontransactional tables. Running results_cache_invalidation4.q on hive master (2c18ef2d5dde2a0c682eeb6dc9b7473b3b94386e) leads to the wrong result 'Victor' and 'Victor'.

ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
```diff
@@ -671,10 +672,20 @@ public void notifyTableChanged(String dbName, String tableName, long updateTime)
    */
   private boolean entryMatches(LookupInfo lookupInfo, CacheEntry entry, Set<CacheEntry> entriesToRemove) {
     QueryInfo queryInfo = entry.getQueryInfo();
+
+    Set<Long> cacheTableIds = new HashSet<>();
```
zabetak: The set variable is redundant and always empty.

ql/src/java/org/apache/hadoop/hive/ql/cache/results/QueryResultsCache.java
```diff
@@ -671,10 +672,20 @@ public void notifyTableChanged(String dbName, String tableName, long updateTime)
...
     for (ReadEntity readEntity : queryInfo.getInputs()) {
-      // Check that the tables used do not resolve to temp tables.
       if (readEntity.getType() == Type.TABLE) {
         Table tableUsed = readEntity.getTable();
+        // we want that the lookupInfo.tableIds are all covered by the table ids of the cache entry
+        // the query is used as cache key, so the lookup and the entry should use the same number of tables
+        // so it is enough to check whether every cache table id is contained in the lookup
+        if (!lookupInfo.tableIds.contains(tableUsed.getTTable().getId())) {
+          LOG.debug("Cached query not valid as it does not provide all table ids: expected {} but was {}", lookupInfo.tableIds, cacheTableIds);
```
zabetak: The `cacheTableIds` is always empty. Apart from that, I would simplify also a bit the message:
```suggestion
          LOG.debug("Cache entry contains a table (tableId={}) that is not present in the query", tableUsed.getTTable().getId());
```

zabetak:
Thanks @czxm for kicking of this work and @thomasrebele for pushing it to completion!

PR #6067, opened 2025-09-08
HIVE-28197: Add deserializer to convert JSON plans to RelNodes

soumyakanti3578:

Added a deserializer to convert JSON plans to logical plans (RelNodes)

While we can serialize a plan to JSON with `explain cbo formatted`, we didn't have a deserializer to convert back to a RelNode.

No

```
mvn test -pl ql -Dtest=org.apache.hadoop.hive.ql.optimizer.calcite.TestRelPlanParser
```


ql/src/java/org/apache/hadoop/hive/ql/metadata/NotNullConstraint.java
```diff
@@ -56,7 +56,7 @@ public NotNullConstraint(List<SQLNotNullConstraint> nns, String tableName, Strin
         String enable = pk.isEnable_cstr()? "ENABLE": "DISABLE";
         String validate = pk.isValidate_cstr()? "VALIDATE": "NOVALIDATE";
         String rely = pk.isRely_cstr()? "RELY": "NORELY";
-        enableValidateRely.put(pk.getNn_name(), ImmutableList.of(enable, validate, rely));
```
zabetak, line 59: Why is this change necessary?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveAggregate.java
```diff
@@ -51,6 +52,11 @@ public HiveAggregate(RelOptCluster cluster, RelTraitSet traitSet, RelNode child,
             groupSet, groupSets, aggCalls);
   }
 
+  public HiveAggregate(RelInput input) {
+    this(input.getCluster(), input.getTraitSet(), input.getInput(),
+        input.getBitSet("group"), input.getBitSetList("groups"), input.getAggregateCalls("aggs"));
```
zabetak, line 57: Can the following work?
```suggestion
    super(input);
```
If yes then can I do the same on the other RelNodes?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveMultiJoin.java
```diff
@@ -113,6 +115,18 @@ public HiveMultiJoin(
...
+        input.getCluster(),
+        input.getInputs(),
+        input.getExpression("condition"),
+        input.getRowType("rowType"),
+        (List<Pair<Integer, Integer>>) input.get("getJoinInputsForHiveMultiJoin"),
+        (List<JoinRelType>) input.get("getJoinTypesForHiveMultiJoin"),
+        input.getExpressionList("filters")
+    );
+  }
+
```
zabetak, line 129: Why do we to modify this class? Normally we shouldn't need to serialize/deserialize `MultiJoin` expressions cause they never appear in the final plan.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveRelNode.java
```diff
@@ -20,8 +20,19 @@
...
   Convention CONVENTION = new Convention.Impl("HIVE", HiveRelNode.class);
+
+  static Stream<RelNode> stream(RelNode node) {
+    return Stream.concat(
+        Stream.of(node),
+        node.getInputs()
+            .stream()
+            .flatMap(HiveRelNode::stream)
+    );
+  }
```
zabetak, line 37: If we keep this we should add appropriate Javadoc. In addition, putting static methods in interfaces is not a good pattern; it is better to move it to a utility class.

Other than that the most common way to traverse RelNode tree is via visitor and shuttles so not sure if this kind of Stream based traversal is something that will be well adopted.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/stats/FilterSelectivityEstimator.java
```diff
@@ -414,20 +415,14 @@ private Double computeConjunctionSelectivity(RexCall call) {
    * @return
    */
   private long getMaxNulls(RexCall call, HiveTableScan t) {
```
zabetak, line 417: Why are we changing the selectivity estimator?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/RexNodeConverter.java
```diff
@@ -699,15 +699,23 @@ protected RexNode convert(ExprNodeConstantDesc literal) throws CalciteSemanticEx
       calciteLiteral = rexBuilder.makeExactLiteral((BigDecimal) value, calciteDataType);
       break;
     case FLOAT:
-      calciteLiteral = rexBuilder.makeApproxLiteral(
+      // We mostly convert FLOAT to exact literal everywhere else, except when
+      // there is a suffix 'F'. See TypeCheckProcFactory.NumExprProcessor#process
+      // Converting to an exact literal helps with deserialization by avoiding
+      // scientific notation.
+      calciteLiteral = rexBuilder.makeExactLiteral(
```
zabetak, line 706: I am reluctant to change the type of literals just for the sake of serialization/deserialization. Is there another way to solve the problem? Even if it makes sense to change literal types it is a risky change that should be tested and land on its own.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelJsonImplWithStats.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.calcite.util.ImmutableBitSet;
+import org.apache.calcite.util.Pair;
+import org.apache.hadoop.hive.ql.plan.ColStatistics;
+
+import java.util.List;
+import java.util.Map;
+
+public class HiveRelJsonImplWithStats extends HiveRelJsonImpl{
```
zabetak, line 29: Since including the stats was the default behavior so far I would follow a slightly different approach. I would keep the original class intact and have the "nostats" serialization logic in this **new** class (changing the name accordingly). This makes it easier to follow what change and what not.

ql/src/java/org/apache/hadoop/hive/ql/parse/type/RexNodeExprFactory.java
```diff
@@ -617,6 +617,26 @@ protected RexNode createConstantExpr(TypeInfo typeInfo, Object constantValue)
           operands);
     }
     RelDataType finalType = TypeConverter.convert(typeInfo, rexBuilder.getTypeFactory());
+
+    // Make exact literals for float/double when the value is constant
+    // to avoid printing in scientific notation. This helps while
+    // deserializing the json plan.
+    switch (finalType.getSqlTypeName()) {
```
zabetak, line 624: As mentioned elsewhere this kind of change is risky so it should be part of another PR if necessary.

ql/src/java/org/apache/hadoop/hive/ql/parse/CalcitePlanner.java
```diff
@@ -516,34 +507,7 @@ public RelNode genLogicalPlan(ASTNode ast) throws SemanticException {
   }
 
   public static RelOptPlanner createPlanner(HiveConf conf) {
```
zabetak, line 509: It was already possible to create a planner through this method. Why did we move the entire logic in `HiveRelOptUtil`?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveSortExchange.java
```diff
@@ -87,6 +105,28 @@ public static HiveSortExchange create(RelNode input,
...
+      int index = relFieldCollation.getFieldIndex();
+      builder.add(input.getCluster().getRexBuilder().makeInputRef(input.getInput(), index));
+    }
+
+    return builder.build();
+  }
+
+  private static RelDistribution getDistribution(RelInput input) {
+    RelDistribution result = input.getDistribution();
+    if (RelDistribution.Type.ANY == result.getType()) {
```
zabetak, line 123: The `RelDistribution.Type.ANY` does not imply that distribution keys are empty. I am not sure what is the method trying to do here.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
```diff
@@ -1079,15 +1106,110 @@ public static Pair<RelOptTable, List<Integer>> getColumnOriginSet(RelNode rel, I
...
     if (rel == null) {
       return null;
     }
-
-    final HiveRelJsonImpl planWriter = new HiveRelJsonImpl();
     rel.explain(planWriter);
     return planWriter.asString();
   }
 
+  public static Optional<String> serializeToJSON(RelNode plan) {
```
zabetak, line 1120: Since this is a new public API in a commonly used class it should have Javadoc. Moreover, it really resembles to an existing method `toJsonString` but it's not obvious why we picked a different name. It's nto possible to decide which one to call without reading the code. When we are adding APIs to existing classes we try to keep things uniform so that other devs can find easier what they need.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
```diff
@@ -1079,15 +1106,110 @@ public static Pair<RelOptTable, List<Integer>> getColumnOriginSet(RelNode rel, I
...
+    }
+
+    JSONObject outJSONObject = new JSONObject(new LinkedHashMap<>());
+    outJSONObject.put("CBOPlan", serializeWithPlanWriter(plan, new HiveRelJsonImpl()));
+    String jsonPlan = outJSONObject.toString();
+
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Plan to serialize: \n{}", RelOptUtil.toString(plan));
+      LOG.debug("JSON plan: \n{}", jsonPlan);
+    }
```
zabetak, line 1132: Don't think we need logging at this stage since we are giving back the result and the receiver can decide what to do about it. Also the only caller at this stage is test code so there is no benefit in having the logs since the tests are explicitly validating this output.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
```diff
@@ -1079,15 +1106,110 @@ public static Pair<RelOptTable, List<Integer>> getColumnOriginSet(RelNode rel, I
...
+
+    return Optional.of(jsonPlan);
+  }
+
+  private static boolean isSerializable(RelNode plan) {
```
zabetak, line 1137: Alternative implem without new APIs.
```java
private static boolean isSerializable(RelNode plan) {
    boolean[] hasJdbcRel = new boolean[] { false };
    plan.accept(new RelHomogeneousShuttle() {
      @Override
      public RelNode visit(RelNode other) {
        hasJdbcRel[0] |= other instanceof JdbcRel;
        return hasJdbcRel[0] ? other : super.visit(other);
      }
    });
    return !hasJdbcRel[0];
  }
```

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
```diff
@@ -1079,15 +1106,110 @@ public static Pair<RelOptTable, List<Integer>> getColumnOriginSet(RelNode rel, I
...
+
+  public static RelNode deserializePlan(HiveConf conf, String jsonPlan) throws IOException {
+    RelOptPlanner planner = HiveRelOptUtil.createPlanner(conf, StatsSources.getStatsSource(conf), false);
+    RexBuilder rexBuilder = new RexBuilder(new HiveTypeFactory());
+    RelOptCluster cluster = RelOptCluster.create(planner, rexBuilder);
+
+    RelPlanParser parser = new RelPlanParser(cluster, conf);
+    RelNode deserializedPlan = parser.parse(jsonPlan);
+    // Apply partition pruning to compute partition list in HiveTableScan
+    deserializedPlan = applyPartitionPruning(conf, deserializedPlan, cluster, planner);
```
zabetak, line 1152: Why do we need the partition list? Can't we deserialize the plan without it?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
```diff
@@ -1079,15 +1106,110 @@ public static Pair<RelOptTable, List<Integer>> getColumnOriginSet(RelNode rel, I
...
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Deserialized plan: \n{}", RelOptUtil.toString(deserializedPlan));
```
zabetak, line 1154: Consider removing logging from this API. Same reasons as the one mentioned before.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveTableScan.java
```diff
@@ -326,10 +355,23 @@ public boolean isInsideView() {
...
     return tableScanTrait;
   }
 
+  private static HiveTableScanTrait createTableScanTrait(RelInput input) {
+    String enumName = input.getString("tableScanTrait");
+    if (enumName == null) {
+      return null;
+    }
+
+    return HiveRelEnumTypes.toEnum(enumName);
```
zabetak, line 372: The use of `HiveRelEnumTypes` seems a bit of an overkill. Can't we simply create the instance directly and drop the entire `RelEnumTypes` copy?
```suggestion
    return HiveTableScanTrait.valueOf(enumName);
```

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveTableScan.java
```diff
@@ -326,10 +355,23 @@ public boolean isInsideView() {
...
+
   public HiveTableScanTrait getTableScanTrait() {
     return tableScanTrait;
   }
 
+  private static HiveTableScanTrait createTableScanTrait(RelInput input) {
+    String enumName = input.getString("tableScanTrait");
+    if (enumName == null) {
+      return null;
+    }
```
zabetak, line 370: Are there cases where we don't serialize the trait? Can we ever have null here?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
```diff
@@ -1079,15 +1106,110 @@ public static Pair<RelOptTable, List<Integer>> getColumnOriginSet(RelNode rel, I
...
     return planWriter.asString();
   }
 
+  public static Optional<String> serializeToJSON(RelNode plan) {
+    if (!isSerializable(plan)) {
+      return Optional.empty();
+    }
+
+    JSONObject outJSONObject = new JSONObject(new LinkedHashMap<>());
+    outJSONObject.put("CBOPlan", serializeWithPlanWriter(plan, new HiveRelJsonImpl()));
```
zabetak, line 1126: I don't think we need the extra wrapping attribute for "CBOPlan".

zabetak (commented):
Didn't fully went through the changes but sending a first batch of comments in order not to lose them. Let me finalize the review before starting making code changes. For comments that simply require an answer feel free to share your thoughts.

ql/src/java/org/apache/hadoop/hive/ql/metadata/NotNullConstraint.java
soumyakanti3578: I don't remember exactly - I will have to look into it, but somewhere we were running into an error due to immutability of the list during serialization or deserialization.

I will try to find out exactly what the issue was.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveAggregate.java
soumyakanti3578: Yes, we can call `super` directly in some of them. We cannot call `super` for anything that extends `Join` as it doesn't have the constructor. There are others where we need to do some post-processing after calling `super`. I will do these changes in the next commit.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveMultiJoin.java
soumyakanti3578: I will look into it in detail as it was relevant in the earlier PR: https://github.com/apache/hive/pull/5131/commits/6b94a2065ae95e2119e1e013ac9afe57b762fe5f

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveRelNode.java
soumyakanti3578: I will try to avoid doing this and rely on a shuttle.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveSortExchange.java
soumyakanti3578: This is not needed anymore and can be removed. I believe this was a bug.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveTableScan.java
soumyakanti3578: Yes we don't always serialize `tableScanTrait`. We only serialize it when `pw.getDetailLevel() == SqlExplainLevel.ALL_ATTRIBUTES`

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/reloperators/HiveTableScan.java
soumyakanti3578: Actually `HiveRelEnumTypes` was getting used in another place earlier but I was able to remove it: https://github.com/apache/hive/pull/5131/files#diff-29a8fea85c2750c60547a7b4d3088d3f704a48f77f6a6b9eb933f1b5b527e033R334

I didn't realize that now this was the only place where it is used. I will update this in the next commit.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/stats/FilterSelectivityEstimator.java
soumyakanti3578: This can be reverted.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/translator/RexNodeConverter.java
soumyakanti3578: I couldn't think of another way to resolve this. The crux of the issue is that right now we almost always represents a double/float as a decimal, except when we use their respective suffix, or when we do constant folding. I believe the constant folding behavior is not intentional.

So the CBO plans can sometimes (rare) have the scientific notation and the deserialized plan will never have one (we have a choice between never and always, I chose to never use it as we rarely use scientific notation). 

We need this change, either here or in another PR, so that our tests pass, as we are relying on `Assertions.assertEquals` to compare the CBO and deserialized plans.

I believe we should create a separate PR and "fix" this, so that there is uniformity in the plans. There are not many plans which are affected - they all are present in this PR.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
soumyakanti3578: I think I will change it back to `toJsonString`. Maybe add a boolean variable `withStats`.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
soumyakanti3578: `explain cbo formatted` comes with the wrapper so I guess it's a good idea to keep it?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
soumyakanti3578: Yeah this will not be in the final commit. I will remove it.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/HiveRelOptUtil.java
soumyakanti3578: We can deserialize the plan but `partitionList` of `RelOptHiveTable` will be empty. Also, the plans won't match up as we do print `plKey` in CBO plans for `HiveTableScan`.

zabetak:
Hey bot, this is still relevant keep it open!

PR #6068, opened 2025-09-09
HIVE-29186: Compactor workers should be revoked only if running in me…

tanishq-chugh:
…tastore in case of HMS instance restart

Revoke the state of compaction jobs only in case they are running in metastore in the event of HMS instance restart (which runs the initiator).

In case there are multiple HS2 instances / worker threads and the HMS instance running the initiator restarts, it resets the state of all working compaction jobs in the compaction queue based on the host of metastore that crashed.

If workers are running on the HS2 on the same host as the one where HMS instance that crashed was running, and workers survive the HMS crash, due to state reset, the same job is picked up by another worker leading to duplicate dir/records. 

With the given solution, there can be multiple cases (considering the HMS instance that crashes is the one running initiator):
1. Multiple HS2 & HMS instances - HS2 will connect to new HMS instance and go ahead, state wont change in queue , second worker wont pick it up
2. Multiple HS2 but one HMS instance-
 - Case A : If HMS comes back up in time, HS2 worker will reconnect to same and go ahead, state won't change.
 - Case B : if it doesn't the HS2 worker thread will eventually die and revokeTimedoutWorkers will take care of it.

No

Manual testing on a cluster after reproducing the issue.

### DIFF
--- a/ql/src/test/org/apache/hadoop/hive/ql/txn/compactor/TestInitiator.java
+++ b/ql/src/test/org/apache/hadoop/hive/ql/txn/compactor/TestInitiator.java
@@ -39,7 +39,9 @@ import org.apache.hadoop.hive.metastore.api.ShowCompactResponseElement;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.metastore.api.hive_metastoreConstants;
 import org.apache.hadoop.hive.metastore.conf.MetastoreConf;
+import org.apache.hadoop.hive.metastore.txn.TxnHandler;
 import org.apache.hadoop.hive.metastore.txn.TxnUtils;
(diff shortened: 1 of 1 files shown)


ql/src/java/org/apache/hadoop/hive/ql/txn/compactor/Initiator.java
```diff
@@ -236,7 +236,10 @@ private TableOptimizer instantiateTableOptimizer(String className) {
   }
 
   private void recoverFailedCompactions(boolean remoteOnly) throws MetaException {
-    if (!remoteOnly) txnHandler.revokeFromLocalWorkers(ServerUtils.hostname());
+    if (!remoteOnly &&
```
deniskuzZ, line 239: @tanishq-chugh  we should drop `HIVE_METASTORE_RUNWORKER_IN` and not allow to run workers on HMS

PR #6069, opened 2025-09-09
HIVE-29183: Iceberg: Integrating Variant DataType in Hive (POC)

ayushtkn:

Adding support for Variant DataType in Hive-Iceberg

To  use Iceberg Variant datatype in Hive

Variant DataType can be used

UT


iceberg/iceberg-handler/src/test/queries/positive/variant_type.q
```diff
@@ -9,7 +9,7 @@
 CREATE EXTERNAL TABLE variant_test_basic (
     id INT,
     data VARIANT
-) STORED BY ICEBERG  tblproperties('format-version'='3');
+) STORED BY ICEBERG stored as avro tblproperties('format-version'='3');
```
deniskuzZ: should it be parquet here?

ayushtkn: Yes, leftover from my experiment to see if avro works as well. It does :-), I added a test below for avro table, forgot to remove it from here.

deniskuzZ:
hm , looks like I broke something, however, variant_type.q is passing if i revert the 0ed79baa5b04e4885f8b82c3d8163c004f3de48b

PR #6070, opened 2025-09-09
HIVE-29187: Unable to parse explain analyze statement when query has quoted identifier with special character

kasakrisz:

Quote all identifiers in the query passed to explain analyze. 

In case of explain analyze the AST subtree representing the query to analyze is transformed back to SQL text to pass it to the driver. During parsing of the original statement the quotation marks are stripped and when the AST is converted back to SQL text these are not added back. This leads to parsing exception when running the query if it has identifiers with special or whitespace characters.

No.

```
mvn test -Dtest.output.overwrite -Dtest=TestMiniLlapLocalCliDriver -Dqfile=quotedid_basic.q,quotedid_basic_standard.q,check_constraint.q -pl itests/qtest -Pitests
mvn test -Dtest.output.overwrite -Dtest=TestIcebergCliDriver -Dqfile=merge_iceberg_copy_on_write_unpartitioned.q -pl itests/qtest-iceberg -Pitests
```

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
@@ -28,7 +28,9 @@ import java.util.regex.Pattern;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.SetMultimap;
+import org.antlr.runtime.TokenRewriteStream;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.ql.exec.FunctionRegistry;
 import org.apache.hadoop.hive.ql.lib.CostLessRuleDispatcher;
 import org.apache.hadoop.hive.ql.lib.ExpressionWalker;
 import org.apache.hadoop.hive.ql.lib.Node;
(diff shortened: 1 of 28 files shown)


thomasrebele:
Thank you for the patch! I'm wondering whether it would make sense to insert quotes only for those identifiers where it is mandatory? A quick check that works for the majority of the identifiers would reduce the size of the patch. E.g., quotes could be avoided for identifiers that start with [a-zA-Z_] and continue with [a-zA-Z_0-9]*.

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
```diff
@@ -334,19 +343,19 @@ public static UnparseTranslator collectUnescapeIdentifierTranslations(ASTNode no
...
       UnparseTranslator unparseTranslator = ((QuotedIdExpressionContext)procCtx).getUnparseTranslator();
-      ASTNode tokTableOrColNode = (ASTNode) nd;
-      for (int i = 0; i < tokTableOrColNode.getChildCount(); ++i) {
-        ASTNode child = (ASTNode) tokTableOrColNode.getChild(i);
-        if (child.getType() == HiveParser.Identifier) {
-          unparseTranslator.addIdentifierTranslation(child);
-        }
+      ASTNode identifier = (ASTNode) nd;
+      String id = identifier.getText();
+      if (FunctionRegistry.getFunctionInfo(id) != null){
```
deniskuzZ, line 354: @kasakrisz, we skip from quoting just the function calls, right? In a code when !fi.isNative() we also sometimes add `addIdentifierTranslation`

kasakrisz:
> Thank you for the patch! I'm wondering whether it would make sense to insert quotes only for those identifiers where it is mandatory? A quick check that works for the majority of the identifiers would reduce the size of the patch. E.g., quotes could be avoided for identifiers that start with [a-zA-Z_] and continue with [a-zA-Z_0-9]*.

Thanks for the review @thomasrebele .
Currently, Hive quotes all identifiers (except function names) for other features where the SQL text has to be generated form the AST. Example: view expanded text, query result cache key, merge rewrite, check constraints etc...
If we want to introduce the check of the identifier name I would do for all similar features. In case of view expanded text it is not applicable because in that case one of the goal is to get all identifiers quoted.

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
kasakrisz: Yes. Hive doesn't support special and whitespace characters in function names.

PR #6071, opened 2025-09-10
HIVE-29188: [hiveACIDReplication] Add src and tgt commit time in replication metrics for better monitoring

shivaom02:
  * Currently when the Hive ACID replication is running we do not have any idea of how the replication is progressing.
  * Even when the replication is not running, we do not have any idea how much the target is behind the src.
  * This commit is to add this info in replication_metrics.

  * These changes are needed for better monitoring of Hive ACID replication from the target.

  * No.


pudidic:
+1 LGTM.

pudidic (approved):

PR #6073, opened 2025-09-11
HIVE-29191: Gracefully Handle Removed Configuration Properties in Hiv…

vikramahuja1001:
…e 4 to Avoid Job Failures

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -5607,16 +5607,11 @@ public class HiveConf extends Configuration {
     HIVE_MM_AVOID_GLOBSTATUS_ON_S3("hive.mm.avoid.s3.globstatus", true,
         "Whether to use listFiles (optimized on S3) instead of globStatus when on S3."),
 
-    HIVE_IGNORE_REMOVED_CONFIGS_LIST("hive.ignore.removed.configs.list",
-            "",
-            "Comma separated list of configuration options which are removed from hive code. Silently ignore if the user tries to set them"),
-
     // If a parameter is added to the restricted list, add a test in TestRestrictedList.Java
     HIVE_CONF_RESTRICTED_LIST("hive.conf.restricted.list",
         "hive.security.authenticator.manager,hive.security.authorization.manager," +
         "hive.security.metastore.authorization.manager,hive.security.metastore.authenticator.manager," +
         "hive.users.in.admin.role,hive.server2.xsrf.filter.enabled,hive.server2.csrf.filter.enabled,hive.security.authorization.enabled," +
-            "hive.ignore.removed.configs.list," +
             "hive.distcp.privileged.doAs," +
             "hive.server2.authentication.ldap.baseDN," +
             "hive.server2.authentication.ldap.url," +
--- a/ql/src/java/org/apache/hadoop/hive/ql/processors/SetProcessor.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/processors/SetProcessor.java
@@ -67,7 +67,11 @@ public class SetProcessor implements CommandProcessor {
           "hive.metastore.initial.metadata.count.enabled",
           "hive.cli.pretty.output.num.cols",
           "hive.debug.localtask",
-          "hive.timedout.txn.reaper.start"
+          "hive.timedout.txn.reaper.start",
+          "hive.repl.dumpdir.ttl",
+          "hive.repl.dumpdir.clean.freq",
+          "hive.llap.io.vrb.queue.limit.base",
+          "hive.llap.external.splits.order.by.force.single.split"
           );
   // Allow the user to set the ORC properties without getting an error.
   private static final Set<String> allowOrcConfigs = new HashSet<>();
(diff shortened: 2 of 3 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -5607,11 +5607,16 @@ public static enum ConfVars {
     HIVE_MM_AVOID_GLOBSTATUS_ON_S3("hive.mm.avoid.s3.globstatus", true,
         "Whether to use listFiles (optimized on S3) instead of globStatus when on S3."),
 
+    HIVE_IGNORE_REMOVED_CONFIGS_LIST("hive.ignore.removed.configs.list",
+            "",
+            "Comma separated list of configuration options which are removed from hive code. Silently ignore if the user tries to set them"),
+
```
zabetak: I don't think we need a new property since this complicates the deprecation and removal of properties even more. `hive.conf.validation` seems to be enough for skipping validation although I don't like much this flag either.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -7261,6 +7266,10 @@ public static String generateDeprecationWarning() {
         + "versions. Please adjust DDL towards the new semantics.";
   }
 
+  public static String generateRemovedWarning() {
+    return "This config does not exist in the current version of Hive. Consider removing this config.";
+  }
+
```
zabetak, line 7267: There are better ways to log warning about deprecated/removed properties. See: `org.apache.hadoop.conf.Configuration#addDeprecation`

ql/src/java/org/apache/hadoop/hive/ql/processors/SetProcessor.java
```diff
@@ -55,19 +56,26 @@ public class SetProcessor implements CommandProcessor {
...
   private static final String prefix = "set: ";
-  private static final Set<String> removedConfigs =
+  private static final Set<String> removedHiveConfigs =
       Sets.newHashSet("hive.mapred.supports.subdirectories",
           "hive.enforce.sorting","hive.enforce.bucketing",
           "hive.outerjoin.supports.filters",
           "hive.llap.zk.sm.principal",
-          "hive.llap.zk.sm.keytab.file"
+          "hive.llap.zk.sm.keytab.file",
+          "hive.stats.fetch.partition.stats",
```
Aggarwal-Raghav, line 65: there are other configs as well which I found doesn't exist in hive:
1. hive.repl.dumpdir.ttl
2. hive.repl.dumpdir.clean.freq
3. hive.llap.io.vrb.queue.limit.base
4. hive.llap.external.splits.order.by.force.single.split

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
vikramahuja1001: Thanks for the review @zabetak . Couldn't revert back/ rework due to limited connectivity. I have removed the property as per your comment.

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
vikramahuja1001: I checked the method `addDeprecation`. This particular method takes oldKeys and newKeys(which are replacing the oldKeys) and it also has a null check to make sure newKeys are not null. Now since we have removed the properties there are no newKeys in our case making the method unusable for us. 

Do let me know in case there is some other way to use the same functionality.

ql/src/java/org/apache/hadoop/hive/ql/processors/SetProcessor.java
vikramahuja1001: Thanks for pointing them out @Aggarwal-Raghav , i have added them to the code

ql/src/java/org/apache/hadoop/hive/ql/processors/SetProcessor.java
```diff
@@ -55,19 +56,30 @@ public class SetProcessor implements CommandProcessor {
   private static final SessionState.LogHelper console = SessionState.getConsole();
 
   private static final String prefix = "set: ";
-  private static final Set<String> removedConfigs =
+  private static final Set<String> removedHiveConfigs =
```
zabetak, line 59: I feel that maintaining all removed configs in the code base does not make much sense and beats the purpose of deprecating and removing a property. 

If every time that we remove a property we had to add it here and continue maintaining to some extend then why remove it in the first place. We could keep the property and just remove all references to it which also doesn't make sense.

Moreover, as history has shown we don't enrich this `removeConfigs` set in a consistent manner so sooner or later we will have more tickets proposing to enrich this set with more removed stuff.

I would propose to drop this completely and stop adding stuff here.

ql/src/java/org/apache/hadoop/hive/ql/processors/SetProcessor.java
```diff
@@ -251,7 +275,7 @@ public Map<String, String> getHiveVariable() {
           message.append("' FAILED in validation : ").append(fail).append('.');
           throw new IllegalArgumentException(message.toString());
         }
-      } else if (!removedConfigs.contains(key) && key.startsWith("hive.")) {
+      } else if (!allowOrcConfigs.contains(key) && key.startsWith("hive.")) {
```
zabetak, line 278: Going forward I see the following options (in personal preference order):
1. Leave exception as is and have users turn off property validation till they migrate their scripts which is somewhat acceptable given that a proper deprecation path was followed before removing a property
2. Add a very specific boolean property (e.g., `hive.conf.validation.missing.key.throws`) that fine tunes a bit if validation throws or not on missing keys
3. We revert HIVE-7211 and never throw if a "hive." property is missing (potentially just log a warn message)

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
zabetak: I guess we could pass an empty string or some special placeholder to achieve somewhat the desired behavior. However, since we are talking about removed properties I guess the deprecation API that I suggested does not make much sense. One way or the other at some point also entries in the deprecation context will be removed.

PR #6074, opened 2025-09-11
HIVE-29192: Iceberg: [V3] Add support for native default column type during create

ayushtkn:

Leverage the Iceberg V3 Column Defaults

To maintain the column defaults in the Iceberg Spec & able to use defaults for fields of Struct

Yes, now column defaults are persisted in Iceberg spec. + We can specify defaults for specific fields of a Struct type

UT

### DIFF
--- a/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
+++ b/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
@@ -337,11 +337,8 @@ public final class HiveSchemaUtil {
     }
   }
 
-  public static void setDefaultValues(List<Types.NestedField> fields, Record record, Set<String> missingColumns) {
-    // Pre-filter fields to only those that are missing or struct types
-    List<Types.NestedField> relevantFields =
-        fields.stream().filter(field -> missingColumns.contains(field.name()) || field.type().isStructType()).toList();
-    for (Types.NestedField field : relevantFields) {
+  public static void setDefaultValues(Record record, List<Types.NestedField> fields, Set<String> missingColumns) {
+    for (Types.NestedField field : fields) {
       Object fieldValue = record.getField(field.name());
 
       if (fieldValue == null) {
@@ -353,7 +350,7 @@ public final class HiveSchemaUtil {
             Record nestedRecord = GenericRecord.create(field.type().asStructType());
             record.setField(field.name(), nestedRecord);
             // For nested fields, we consider ALL fields as "missing" to apply defaults
-            setDefaultValuesForNestedStruct(field.type().asStructType().fields(), nestedRecord);
+            setDefaultValuesForNestedStruct(nestedRecord, field.type().asStructType().fields());
           } else if (field.writeDefault() != null) {
             Object defaultValue = convertToWriteType(field.writeDefault(), field.type());
             record.setField(field.name(), defaultValue);
@@ -362,13 +359,13 @@ public final class HiveSchemaUtil {
         // Explicit NULLs remain NULL
       } else if (field.type().isStructType() && fieldValue instanceof Record) {
         // For existing structs, apply defaults to any null nested fields
-        setDefaultValuesForNestedStruct(field.type().asStructType().fields(), (Record) fieldValue);
+        setDefaultValuesForNestedStruct((Record) fieldValue, field.type().asStructType().fields());
       }
     }
   }
 
   // Special method for nested structs that always applies defaults to null fields
-  private static void setDefaultValuesForNestedStruct(List<Types.NestedField> fields, Record record) {
+  private static void setDefaultValuesForNestedStruct(Record record, List<Types.NestedField> fields) {
     for (Types.NestedField field : fields) {
       Object fieldValue = record.getField(field.name());
 
@@ -378,7 +375,7 @@ public final class HiveSchemaUtil {
         record.setField(field.name(), defaultValue);
       } else if (field.type().isStructType() && fieldValue instanceof Record) {
         // Recursively process nested structs
-        setDefaultValuesForNestedStruct(field.type().asStructType().fields(), (Record) fieldValue);
+        setDefaultValuesForNestedStruct((Record) fieldValue, field.type().asStructType().fields());
       }
     }
   }
--- a/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
+++ b/iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
@@ -32,11 +32,13 @@ import org.apache.iceberg.io.OutputFileFactory;
 import org.apache.iceberg.mr.hive.FilesForCommit;
 import org.apache.iceberg.mr.hive.writer.WriterBuilder.Context;
 import org.apache.iceberg.mr.mapred.Container;
+import org.apache.iceberg.types.Types;
 
 class HiveIcebergRecordWriter extends HiveIcebergWriterBase {
 
   private final int currentSpecId;
   private final Set<String> missingColumns;
+  private final List<Types.NestedField> missingOrStructFields;
 
   HiveIcebergRecordWriter(Table table, HiveFileWriterFactory fileWriterFactory,
       OutputFileFactory dataFileFactory, Context context) {
@@ -44,12 +46,14 @@ class HiveIcebergRecordWriter extends HiveIcebergWriterBase {
 
     this.currentSpecId = table.spec().specId();
     this.missingColumns = context.missingColumns();
+    this.missingOrStructFields = specs.get(currentSpecId).schema().asStruct().fields().stream()
+        .filter(field -> missingColumns.contains(field.name()) || field.type().isStructType()).toList();
   }
 
   @Override
   public void write(Writable row) throws IOException {
     Record record = ((Container<Record>) row).get();
-    HiveSchemaUtil.setDefaultValues(specs.get(currentSpecId).schema().asStruct().fields(), record, missingColumns);
+    HiveSchemaUtil.setDefaultValues(record, missingOrStructFields, missingColumns);
 
     writer.write(record, specs.get(currentSpecId), partition(record, currentSpecId));
   }
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/constraint/ConstraintsUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/table/constraint/ConstraintsUtils.java
@@ -253,7 +253,7 @@ public final class ConstraintsUtils {
       } else if (childType == HiveParser.TOK_DEFAULT_VALUE) {
         // try to get default value only if this is DEFAULT constraint
         colTypeInfo = TypeInfoUtils.getTypeInfoFromTypeString(BaseSemanticAnalyzer.getTypeStringFromAST(typeChildForDefault));
-        Pair<TypeInfo, String> defaultValueTypeAndValue = getDefaultValue(grandChild, tokenRewriteStream);
+        Pair<TypeInfo, String> defaultValueTypeAndValue = getDefaultValueAndType(grandChild, tokenRewriteStream);
         defaultValueType = defaultValueTypeAndValue.getKey();
         checkOrDefaultValue = defaultValueTypeAndValue.getValue();
       } else if (childType == HiveParser.TOK_CHECK_CONSTRAINT) {
@@ -309,7 +309,7 @@ public final class ConstraintsUtils {
    * @param node AST node corresponding to default value
    * @return retrieve the default value and return it as string
    */
-  private static Pair<TypeInfo, String> getDefaultValue(ASTNode node, TokenRewriteStream tokenStream)
+  private static Pair<TypeInfo, String> getDefaultValueAndType(ASTNode node, TokenRewriteStream tokenStream)
       throws SemanticException{
     // first create expression from defaultValueAST
     TypeCheckCtx typeCheckCtx = new TypeCheckCtx(null);


iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaConverter.java
```diff
@@ -51,29 +57,56 @@ private HiveSchemaConverter(boolean autoConvert) {
     this.id = 1;
   }
 
-  static Schema convert(List<String> names, List<TypeInfo> typeInfos, List<String> comments, boolean autoConvert) {
+  static Schema convert(List<String> names, List<TypeInfo> typeInfos, List<String> comments, boolean autoConvert,
+      Map<String, String> defaultValues) {
     HiveSchemaConverter converter = new HiveSchemaConverter(autoConvert);
-    return new Schema(converter.convertInternal(names, typeInfos, comments));
+    return new Schema(converter.convertInternal(names, typeInfos, comments, defaultValues));
```
deniskuzZ: minor: `comments` probably should be the last arg

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaConverter.java
```diff
@@ -51,29 +57,56 @@ private HiveSchemaConverter(boolean autoConvert) {
...
+          Types.NestedField.builder()
+              .asOptional()
+              .withId(outerId - names.size() + i)
+              .withName(columnName)
+              .ofType(type)
+              .withDoc(comments.isEmpty() || i >= comments.size() ? null : comments.get(i));
+
+      if (defaultValues.containsKey(columnName)) {
+        if (type.isPrimitiveType()) {
+          Object icebergDefaultValue = getDefaultValue(stripQuotes(defaultValues.get(columnName)), type);
```
deniskuzZ: can we move `stripQuotes` inside of getDefaultValue

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaConverter.java
```diff
@@ -129,24 +162,42 @@ Type convertType(TypeInfo typeInfo) {
...
         return Types.ListType.ofOptional(listId, listType);
       case VARIANT:
         return Types.VariantType.get();
       default:
         throw new IllegalArgumentException("Unknown type " + typeInfo.getCategory());
     }
   }
+
+  private Map<String, String> getDefaultValuesMap(String defaultValue) {
+    if (defaultValue == null || defaultValue.isEmpty()) {
```
deniskuzZ: could be simplified with `StringUtils.isEmpty()`

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
```diff
@@ -60,18 +61,20 @@ public static List<FieldSchema> convert(Schema schema) {
...
+   *
+   * @param fieldSchemas  The list of the columns
+   * @param autoConvert   If <code>true</code> then TINYINT and SMALLINT is converted to INTEGER and VARCHAR and CHAR is
+   *                      converted to STRING. Otherwise if these types are used in the Hive schema then exception is
+   *                      thrown.
+   * @param defaultValues Default values for columns, if any. The map is from column name to default value.
    * @return An equivalent Iceberg Schema
    */
-  public static Schema convert(List<FieldSchema> fieldSchemas, boolean autoConvert) {
+  public static Schema convert(List<FieldSchema> fieldSchemas, boolean autoConvert, Map<String, String> defaultValues) {
```
deniskuzZ: autoConvert should be last arg

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/BaseHiveIcebergMetaHook.java
```diff
@@ -161,7 +164,10 @@ public void preCreateTable(CreateTableRequest request) {
             primaryKeys.stream().map(SQLPrimaryKey::getColumn_name).collect(Collectors.toSet()))
         .orElse(Collections.emptySet());
 
-    Schema schema = schema(catalogProperties, hmsTable, identifierFields);
+    Map<String, String> defaultValues = Stream.ofNullable(request.getDefaultConstraints()).flatMap(Collection::stream)
```
deniskuzZ: can we move this into schema method?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
```diff
@@ -310,7 +310,8 @@ private void doPreAlterTable(org.apache.hadoop.hive.metastore.api.Table hmsTable
       preAlterTableProperties = new PreAlterTableProperties();
       preAlterTableProperties.tableLocation = sd.getLocation();
       preAlterTableProperties.format = sd.getInputFormat();
-      preAlterTableProperties.schema = schema(catalogProperties, hmsTable, Collections.emptySet());
+      preAlterTableProperties.schema =
+          schema(catalogProperties, hmsTable, Collections.emptySet(), Collections.emptyMap());
```
deniskuzZ: is it ok to pass empty list here?

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaConverter.java
```diff
@@ -129,24 +162,42 @@ Type convertType(TypeInfo typeInfo) {
...
+      return Collections.emptyMap();
+    }
+    // For Struct, the default value is expected to be in key:value format
+    return Arrays.stream(stripQuotes(defaultValue).split(","))
```
deniskuzZ: would it be better to use guava splitter here?
````
Splitter.on(',').trimResults().withKeyValueSeparator(':')
  .split(stripQuotes(defaultValue))
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java [hunk 8]
```diff
@@ -303,6 +304,9 @@ public void configureOutputJobProperties(TableDesc tableDesc, Map<String, String
     // table-level and skipped only for output tables in HiveIcebergSerde. Properties from the map will be present in
     // the serde config for all tables in the query, not just the output tables, so we can't rely on that in the serde.
     tableDesc.getProperties().put(InputFormatConfig.OPERATION_TYPE_PREFIX + tableDesc.getTableName(), opType);
+    SessionStateUtil.getResource(conf, SessionStateUtil.MISSING_COLUMNS)
```
deniskuzZ, line 307: is that MISSING_COLUMNS or COLUMNS_WITH_DEFAULT?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergStorageHandler.java
```diff
@@ -2206,6 +2210,12 @@ public boolean hasUndergonePartitionEvolution(org.apache.hadoop.hive.ql.metadata
     return IcebergTableUtil.hasUndergonePartitionEvolution(table);
   }
 
+  @Override
+  public boolean supportsNativeColumnDefault(Map<String, String> tblProps) {
```
deniskuzZ: supportsDefaultColumnValues ?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
```diff
@@ -20,34 +20,131 @@
...
   private final int currentSpecId;
+  private final Set<String> missingColumns;
 
   HiveIcebergRecordWriter(Table table, HiveFileWriterFactory fileWriterFactory,
-      OutputFileFactory dataFileFactory, Context context) {
+      OutputFileFactory dataFileFactory, Context context, String missingColumns) {
     super(table, newDataWriter(table, fileWriterFactory, dataFileFactory, context));
 
     this.currentSpecId = table.spec().specId();
+    this.missingColumns = Optional.ofNullable(missingColumns)
```
deniskuzZ: why not pas as Set initially? also maybe set this in a Context

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
```diff
@@ -20,34 +20,131 @@
...
+        .map(columns -> Arrays.stream(columns.split(",")).collect(Collectors.toCollection(HashSet::new)))
+        .orElse(Sets.newHashSet());
   }
 
   @Override
   public void write(Writable row) throws IOException {
     Record record = ((Container<Record>) row).get();
+    setDefault(specs.get(currentSpecId).schema().asStruct().fields(), record, missingColumns);
```
deniskuzZ: should it be applied to every row? can we optimize?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
```diff
@@ -20,34 +20,131 @@
...
+
     writer.write(record, specs.get(currentSpecId), partition(record, currentSpecId));
   }
 
+  private static void setDefault(List<Types.NestedField> fields, Record record, Set<String> missingColumns) {
```
deniskuzZ: i think this helper methods should be extracted from the writer

ql/src/java/org/apache/hadoop/hive/ql/session/SessionStateUtil.java
```diff
@@ -38,6 +38,7 @@ public class SessionStateUtil {
   private static final String COMMIT_INFO_PREFIX = "COMMIT_INFO.";
   private static final String CONFLICT_DETECTION_FILTER = "conflictDetectionFilter.";
   public static final String DEFAULT_TABLE_LOCATION = "defaultLocation";
+  public static final String MISSING_COLUMNS = "missingColumns";
```
deniskuzZ, line 41: what do we store here?

deniskuzZ:
cc @kasakrisz for the compiler part.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
ayushtkn: We get one record only pushed & this is the place where we have the spec to extract the defaults from the Iceberg layer. I backtracked till `MapRecordProcessor` that pushes one-one record only

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
ayushtkn: We get one record only pushed & this is the place where we have the spec to extract the defaults from the Iceberg layer or did I catch your question wrong?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveIcebergMetaHook.java
ayushtkn: As of now we don't support this in ALTER, so passing emptyMap. I double checked ALTER doesn't override or remove any already set values.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
```diff
@@ -329,4 +336,85 @@ public static String convertToTypeString(Type type) {
         throw new UnsupportedOperationException(type + " is not supported");
     }
   }
+
+  public static void setDefault(List<Types.NestedField> fields, Record record, Set<String> missingColumns) {
```
deniskuzZ: maybe `setDefaultValues(record, fields, missingColumns)`?

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
```diff
@@ -329,4 +336,85 @@ public static String convertToTypeString(Type type) {
...
+        // Explicit NULLs remain NULL
+      } else if (field.type().isStructType() && fieldValue instanceof Record) {
+        // For existing structs, apply defaults to any null nested fields
+        setDefaultForNestedStruct(field.type().asStructType().fields(), (Record) fieldValue);
+      }
+    }
+  }
+
+  // Special method for nested structs that always applies defaults to null fields
+  private static void setDefaultForNestedStruct(List<Types.NestedField> fields, Record record) {
```
deniskuzZ: `setDefaultValuesForNestedStruct`

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaConverter.java
```diff
@@ -129,24 +162,40 @@ Type convertType(TypeInfo typeInfo) {
...
+        Type listType = convertType(listTypeInfo.getListElementTypeInfo(), defaultValue);
         return Types.ListType.ofOptional(listId, listType);
       case VARIANT:
         return Types.VariantType.get();
       default:
         throw new IllegalArgumentException("Unknown type " + typeInfo.getCategory());
     }
   }
+
+  private Map<String, String> getDefaultValuesMap(String defaultValue) {
```
deniskuzZ: static?

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
```diff
@@ -329,4 +336,85 @@ public static String convertToTypeString(Type type) {
         throw new UnsupportedOperationException(type + " is not supported");
     }
   }
+
+  public static void setDefault(List<Types.NestedField> fields, Record record, Set<String> missingColumns) {
+    for (Types.NestedField field : fields) {
```
deniskuzZ, line 341: can we pre-filter `fields` for just the `missingColumns` to avoid repetition for every record?

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
```diff
@@ -329,4 +336,85 @@ public static String convertToTypeString(Type type) {
...
+          return DateTimeUtil.dateFromDays((Integer) value);
+        }
+        break;
+      case TIMESTAMP:
+        // Convert microseconds since epoch (Long) to LocalDateTime
+        if (value instanceof Long) {
+          return DateTimeUtil.timestampFromMicros((Long) value);
+        }
+        break;
+      case TIMESTAMP_NANO:
```
deniskuzZ: looks like we support more types here than in `IcebergInternalRecordWrapper.converter`. could we extend it and reuse?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/constraint/ConstraintsUtils.java
```diff
@@ -290,10 +304,11 @@ private static List<ConstraintInfo> generateConstraintInfos(ASTNode child, List<
 
   /**
    * Validate and get the default value from the AST.
+   *
    * @param node AST node corresponding to default value
    * @return retrieve the default value and return it as string
    */
-  private static String getDefaultValue(ASTNode node, ASTNode typeChild, TokenRewriteStream tokenStream)
+  private static Object[] getDefaultValue(ASTNode node, TokenRewriteStream tokenStream)
```
deniskuzZ: maybe use Pair and rename getDefaultValueAndType?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/constraint/ConstraintsUtils.java
```diff
@@ -311,24 +326,22 @@ private static String getDefaultValue(ASTNode node, ASTNode typeChild, TokenRewr
...
     // throw an error if default value isn't what hive allows
     if (!isDefaultValueAllowed(defaultValExpr)) {
       throw new SemanticException(ErrorMsg.INVALID_CSTR_SYNTAX.getMsg("Invalid Default value: " + defaultValueText +
           ". DEFAULT only allows constant or function expressions"));
     }
-    return defaultValueText;
+    return new Object[]{defaultValExpr.getTypeInfo(), defaultValueText};
   }
 
+  public static void validateDefaultColumnType(TypeInfo colTypeInfo, TypeInfo defaultValTypeInfo,
```
deniskuzZ: validateDefaultValueType

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
```diff
@@ -866,7 +865,7 @@ public static List<FieldSchema> getColumns(
       ASTNode ast, boolean lowerCase, TokenRewriteStream tokenRewriteStream,
       List<SQLPrimaryKey> primaryKeys, List<SQLForeignKey> foreignKeys,
       List<SQLUniqueConstraint> uniqueConstraints, List<SQLNotNullConstraint> notNullConstraints,
-      List<SQLDefaultConstraint> defaultConstraints, List<SQLCheckConstraint> checkConstraints,
+      List<ConstraintsUtils.ConstraintInfo> defaultConstraintInfo, List<SQLCheckConstraint> checkConstraints,
```
deniskuzZ: minor: i think we can keep the naming

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
```diff
@@ -5064,6 +5070,7 @@ private RowResolver getColForInsertStmtSpec(Map<String, ExprNodeDesc> targetCol2
       }
       colListPos++;
     }
+    SessionStateUtil.addResource(conf, MISSING_COLUMNS, missingColumns);
```
deniskuzZ, line 5073: not sure it's a best place to populate missing columns. this is a getter with hidden side functionality.  
i might be wrong. is it called just once during the compilation? if yes, then it's probably Ok.

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/writer/HiveIcebergRecordWriter.java
deniskuzZ: can we optimize filtering part since it's a constant: https://github.com/apache/hive/pull/6074#discussion_r2358314550

ql/src/java/org/apache/hadoop/hive/ql/parse/SemanticAnalyzer.java
ayushtkn: I tried debugging my test and placed a breakpoint here—it hit once for the insert query. Since this code is part of getColForInsertStmtSpec, it makes sense that it should be invoked only once. There aren't multiple methods that call into this, and the native missing column handling is also done here:
```
String f = targetTableColNames.get(i);
if (targetCol2Projection.containsKey(f)) {
  // put existing column in new list to make sure it is in the right position
}
```
From what I observed during debugging, it should only be called once, since it’s essentially capturing column projection. If it’s being triggered multiple times, that would be surprise to me. There might be some specific case where it does get called more than once, but I’m not fully sure about that yet. Not much experience here :-(

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
deniskuzZ: could we make `record` 1st arg?

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
deniskuzZ: same https://github.com/apache/hive/pull/6074#discussion_r2360222286

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/constraint/ConstraintsUtils.java
deniskuzZ: getDefaultValueAndType

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java
deniskuzZ: idea was to pre-filter just once not for every row. setDefaultValues should get already pre-filtered fields (https://github.com/apache/hive/pull/6074#discussion_r2358552106)
````
setDefaultValues(Record record, List<Types.NestedField> fields)
````

PR #6075, opened 2025-09-15
HIVE-29201: Fix masking and re-enable query_iceberg_metadata_of_unpartitioned_table.q

armitage420:

Added explicit order by in queries

The test in itself uses SORT_QUERY_RESULTS to keep a deterministic ordering of queries. But there is still scope of non determinism, as SORT_QUERY_RESULTS sorts the output of each query lexicographically on unmasked rows, and if the present masked values change, then the output ordering changes as well. Hence, we need to add explicit order by on queries.

No

q.out file results, and test pipeline


deniskuzZ:
I would rather not select columns that are going to be masked

armitage420:
@deniskuzZ Not selecting masked columns is not feasible for this particular test, as the there are part of the column values are masked and not the column(related to metadata itself) as a whole.

deniskuzZ:
> @deniskuzZ Not selecting masked columns is not feasible for this particular test, as the there are part of the column values are masked and not the column(related to metadata itself) as a whole.

oh, ok. 

@armitage420 
> if the present masked values change

why would they change?

armitage420:
> @armitage420
> 
> > if the present masked values change
> 
> why would they change?

Thank you for your time @deniskuzZ !

Total size properties might change with file format upgrade, and in our case, it's orc here. Here's the jira for reference: [HIVE-25607](https://issues.apache.org/jira/browse/HIVE-25607)

Followed by the above mentioned jira, there was another jira that introduced masking for the same reason in iceberg qfiles: [HIVE-25658](https://issues.apache.org/jira/browse/HIVE-25658)

thomasrebele:
I had a look at this flaky test, too. If you look at the expected [query result](https://github.com/apache/hive/blob/96d1635f98d0cfd5b0cd01115bb738e1c6053b94/iceberg/iceberg-handler/src/test/results/positive/query_iceberg_metadata_of_unpartitioned_table.q.out#L104), the first columns are the same, but the first different column is not sorted lexicographically:

```
0	hdfs://### HDFS PATH ###	ORC	0	#Masked#	378	...
0	hdfs://### HDFS PATH ###	ORC	0	#Masked#	365	...
0	hdfs://### HDFS PATH ###	ORC	0	#Masked#	374	...
```

The problem is that it is sorted on the original value of `### HDFS PATH ###` and `#Masked#`, and after sorting the values are replaced.

Changing the query to make this deterministic is a workaround for this particular q file. A proposal for a more general fix: refactor the [masking](https://github.com/apache/hive/blob/96d1635f98d0cfd5b0cd01115bb738e1c6053b94/itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestUtil.java#L972) so that it is done before the [sorting](https://github.com/apache/hive/blob/96d1635f98d0cfd5b0cd01115bb738e1c6053b94/cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java#L286) (`out` is a `org.apache.hadoop.hive.common.io.SortPrintStream`).

armitage420:
@thomasrebele Thank you for your input!
You are correct—the lexicographical sorting is done on unmasked values. Therefore, a better (and more accurate) fix would be to apply masking before sorting the results.
Currently, [sorting](https://github.com/apache/hive/blob/96d1635f98d0cfd5b0cd01115bb738e1c6053b94/cli/src/java/org/apache/hadoop/hive/cli/CliDriver.java#L286) is performed for every single query, whereas [masking](https://github.com/apache/hive/blob/96d1635f98d0cfd5b0cd01115bb738e1c6053b94/itests/util/src/main/java/org/apache/hadoop/hive/ql/QTestUtil.java#L972) is only applied at the end, once we have collected all query results for the entire qfile. To implement the actual fix, we would need to change the test architecture so that masking is done per query, followed by sorting.
I'm not sure if this approach would be agreed upon, but if suggested, I can implement it!

@deniskuzZ @thomasrebele  Do let me know what both of you think!

thomasrebele:
I've been working on a draft of applying the masking before the sorting (in addition to applying the masking at the end of the processing) in https://github.com/thomasrebele/hive/tree/tr/HIVE-29201-v1. The design of `FetchConverter` makes it difficult to implement this cleanly. Alternatively, we could make `FetchConverter` an interface (and the old class would become `FetchConverterImpl`) to simplify the logic of LambdaFetchConverter. What do you think, @armitage420, @deniskuzZ?

deniskuzZ:
i think masking in this specific test isn’t very effective, as it bypasses validation for several iceberg metadata fields
query_iceberg_metadata_of_partitioned_table.q doesn't suffer from the same issue?

armitage420:
> i think masking in this specific test isn’t very effective, as it bypasses validation for several iceberg metadata fields query_iceberg_metadata_of_partitioned_table.q doesn't suffer from the same issue?

The masking is only done for HDFS paths, file_size_in_bytes and total file size of table properties, the masking doesn't really effect the validation of the test.

deniskuzZ:
> The masking is only done for HDFS paths, file_size_in_bytes and total file size of table properties, the masking doesn't really effect the validation of the test.

@armitage420 test adds some additional masking as well, try removing and see for yourself. Why do we mask row count instead of size_in_bytes?
https://iceberg.apache.org/docs/1.9.0/spark-queries/#all-data-files
````
0	hdfs://### HDFS PATH ###	ORC	0	5	        378	{1:7,2:30}	

0	hdfs://### HDFS PATH ###	ORC	0	#Masked#	378	{1:7,2:30}
````

armitage420:
> > The masking is only done for HDFS paths, file_size_in_bytes and total file size of table properties, the masking doesn't really effect the validation of the test.
> 
> @armitage420 test adds some additional masking as well, try removing and see for yourself. Why do we mask row count instead of size_in_bytes? https://iceberg.apache.org/docs/1.9.0/spark-queries/#all-data-files
> 
> ```
> 0	hdfs://### HDFS PATH ###	ORC	0	5	        378	{1:7,2:30}	
> 
> 0	hdfs://### HDFS PATH ###	ORC	0	#Masked#	378	{1:7,2:30}
> ```

My bad, I have fixed the wrong masking related to file format. I have also masked the file_size_in_bytes columns for the non-json rows that were missed. Although fixing this has no effect on the flaky nature of the output for this qfile.

zabetak:
Hey everyone, many thanks for working on this. Many recent runs are affected by this flaky test that makes our CI unstable. Since we haven't fully converged on the solution is it reasonable to disable this temporarily till the fix finally lands? 

PR for disabling the test: https://github.com/apache/hive/pull/6098

armitage420:
> Hey everyone, many thanks for working on this. Many recent runs are affected by this flaky test that makes our CI unstable. Since we haven't fully converged on the solution is it reasonable to disable this temporarily till the fix finally lands?
> 
> PR for disabling the test: #6098

I hope we can hold off on the following PR for a few days, to get a conclusion on the recent comments if @deniskuzZ has the time! :D

zabetak:
@armitage420 Every run that fails has a non-negligible cost in terms of money, time, and resources so allowing the flakiness to exist for multiple days/weeks is not OK so I disabled the test for now since the respective PR was approved.

In general, if we cannot solve a flakiness issue reasonably fast we opt to disable the test first and fix it afterwards; that's quite common practice in the Hive project.

deniskuzZ:
> In general, if we cannot solve a flakiness issue reasonably fast we opt to disable the test first and fix it afterwards; that's quite common practice in the Hive project.

True, however, now we have tons of disabled, flaky functional tests no one is looking at.

By the way, I couldn’t find any recent master branch builds where this test failed

deniskuzZ:
@armitage420, @thomasrebele, I agree that a better approach would be to apply masking before sorting the results. 
Since the test is currently disabled, we should aim to re-enable it as soon as possible to avoid unnoticed issues when upgrading to Iceberg 1.10.1 once it’s released. 
If the sort refactor will take longer, I’m fine with approving the current PR in the meantime.

@armitage420, just wondering, was the test failing a few months back?

armitage420:
@deniskuzZ  I had only seen this flakiness once in one of the PRs. While investigating, I found this JIRA that addressed a similar issue across Iceberg q-tests: [HIVE-25607](https://issues.apache.org/jira/browse/HIVE-25607)

I have also encountered other flaky tests (which seem fairly common in PRs run, although unrelated to current one) and created JIRAs in case anyone wants to address them: [HIVE-29158](https://issues.apache.org/jira/browse/HIVE-29158), [HIVE-29157](https://issues.apache.org/jira/browse/HIVE-29157)

Regarding the approach of masking before sorting, I have some concerns:
1. Sorting using SORT_QUERY_RESULTS is executed per query, whereas masking currently executes once per file. Changing the architecture to mask per query would increase the number of operations (and likely execution time for the pipeline).

2. The sorting discussed above is only applied to a few qfiles, so according to point 1, masking would still operate per query in those cases.

3. An approach where masking per query is done only when SORT_QUERY_RESULTS is used (but otherwise operates per file) might work better, though I'm not sure if this makes architectural sense.

deniskuzZ:
@armitage420, i would call this test broken (by some library version bump, etc), not flaky.

thomasrebele:
@armitage420: the QOutProcessor#maskPatterns applies the masking by copying the q.out file, reading the copy and overwriting the old file. Applying the masking on the string in memory will be much faster. I've created [HIVE-29227](https://issues.apache.org/jira/browse/HIVE-29227) to remove the masking from the post processing.

deniskuzZ:
[HIVE-29226](https://issues.apache.org/jira/browse/HIVE-29226) is merged 
@armitage420 could you please re-enable the test and remove the explicit sorting, but please keep the masking fix

deniskuzZ:
UPD: rebased and fixed masking

armitage420:
@deniskuzZ Thanks for your help :D

deniskuzZ (approved):
+1

PR #6078, opened 2025-09-15
HIVE-29202: Add HiveAuthzContext support to HiveMetaStoreAuthorizable…

jjiang037:
…Event for enhanced authorization

1. Add a `buildAuthzContext()` method to HiveMetaStoreAuthorizableEvent that constructs 
   HiveAuthzContext with available context information
2. Add a constructor to HiveMetaStoreAuthzInfo that accepts HiveAuthzContext
3. Update event implementations (starting with ReadDatabaseEvent) to use the new pattern

Allows authorization plugins to implement multi-tenancy, application-aware policies, and audit requirements

No

precommit-test suites; ut

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
@@ -51,6 +51,7 @@ public abstract class HiveMetaStoreAuthorizableEvent {
       builder.setCommandString(commandString);
     }
 
+    // TODO: refer to SessionManager/HiveSessionImpl for details on getting ipAddress and forwardedAddresses
     builder.setForwardedAddresses(new ArrayList<>());
 
     String ipAddress = HMSHandler.getIPAddress();
--- a/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
@@ -65,10 +65,6 @@ public abstract class HiveMetaStoreAuthorizableEvent {
     return builder.build();
   }
 
-  protected HiveAuthzContext buildAuthzContext() {
-    return buildAuthzContext(null);
-  }
-
   public abstract HiveMetaStoreAuthzInfo getAuthzContext();
 
   protected String getSdLocation(StorageDescriptor sd) {
--- a/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
@@ -19,6 +19,7 @@
 
 package org.apache.hadoop.hive.ql.security.authorization.plugin.metastore;
 
+import org.apache.hadoop.hive.metastore.HMSHandler;
 import org.apache.hadoop.hive.metastore.api.DataConnector;
 import org.apache.hadoop.hive.metastore.api.Database;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
@@ -27,6 +28,7 @@ import org.apache.hadoop.hive.metastore.events.PreEventContext;
 import org.apache.hadoop.hive.ql.security.authorization.plugin.HiveAuthzContext;
 import org.apache.hadoop.hive.ql.security.authorization.plugin.HivePrivilegeObject;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -42,8 +44,17 @@ public abstract class HiveMetaStoreAuthorizableEvent {
     this.preEventContext = preEventContext;
   }
 
-  protected HiveAuthzContext buildAuthzContext() {
+  protected HiveAuthzContext buildAuthzContext(String commandString) {
     HiveAuthzContext.Builder builder = new HiveAuthzContext.Builder();
+
+    if (commandString != null) {
+      builder.setCommandString(commandString);
+    }
+
+    builder.setForwardedAddresses(new ArrayList<>());
+
+    String ipAddress = HMSHandler.getIPAddress();
+    builder.setUserIpAddress(ipAddress);
 
     Map<String, Object> clientConfig = HiveMetaStoreAuthorizer.getClientConfig();
     if (clientConfig != null) {
@@ -51,6 +62,10 @@ public abstract class HiveMetaStoreAuthorizableEvent {
     }
 
     return builder.build();
+  }
+
+  protected HiveAuthzContext buildAuthzContext() {
+    return buildAuthzContext(null);
   }
 
   public abstract HiveMetaStoreAuthzInfo getAuthzContext();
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
```diff
@@ -40,6 +42,17 @@ protected HiveMetaStoreAuthorizableEvent(PreEventContext preEventContext) {
     this.preEventContext = preEventContext;
   }
 
+  protected HiveAuthzContext buildAuthzContext() {
```
wecharyu: Do you mean that different `HiveMetaStoreAuthorizableEvent` subclasses each provide their own `authzContext`? Could you also give some concrete usage scenarios as examples?

jjiang037: Yes, different HiveMetaStoreAuthorizableEvent subclasses can provide their own authzContext based on their specific authorization needs. Here are some concrete usage scenarios:

## Client Type Differentiation
### Different clients accessing HMS may require different authorization contexts:
REST Catalog clients vs Thrift clients: REST-based catalog services (like Iceberg REST catalog) may need additional context for audit logging to distinguish their requests from traditional Thrift-based HMS clients
This allows authorization plugins to apply different policies or logging based on the client type
## Enhanced Audit Context
The authorization context can carry metadata about:
Request origin: Whether the request came from a REST API, direct HMS client, or a proxy service
Client properties: Additional metadata like client version, request ID, or session information
Operation context: Specific context relevant to the operation type
## Example Implementation Pattern
```
// For REST catalog requests
public class ReadDatabaseEvent extends HiveMetaStoreAuthorizableEvent {
    @Override
    public HiveMetaStoreAuthzInfo getAuthzContext() {
        HiveAuthzContext authzContext = buildAuthzContext();
        // The context now includes client configuration that identifies
        // this as a REST catalog request vs regular HMS request
        return new HiveMetaStoreAuthzInfo(..., authzContext);
    }
```
}

wecharyu: Thanks for the details, it makes sense.

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
```diff
@@ -40,6 +42,17 @@ protected HiveMetaStoreAuthorizableEvent(PreEventContext preEventContext) {
...
+    HiveAuthzContext.Builder builder = new HiveAuthzContext.Builder();
+
+    Map<String, Object> clientConfig = HiveMetaStoreAuthorizer.getClientConfig();
+    if (clientConfig != null) {
+      builder.setClientConfig(clientConfig);
```
wecharyu, line 62: The `builder` only set `clientConfig`, other fields such as `commandString` and `ipAddress` are missing.

jjiang037: Good point about the other fields. The current implementation intentionally focuses on adding clientConfig as the minimum necessary change for the following reasons:

## Incremental Enhancement
This patch introduces the foundation for custom authorization contexts. Other fields like ipAddress and additional metadata can be added in follow-up patches as specific needs arise. This keeps the change focused and easier to review.

## Context-Specific Fields
Not all authorization events need all fields. For example:
- `commandString` is already provided by each event subclass (like `ReadDatabaseEvent.COMMAND_STR`)
- `ipAddress` may not be available in all HMS invocation paths
- Some fields might be better sourced from the `PreEventContext` when needed

## Current Requirements
The immediate need is to distinguish client types (REST catalog vs traditional HMS clients) for authorization and audit purposes, which clientConfig addresses. Other fields can be added when there are concrete requirements for them.

If you feel strongly that these fields should be included now, I can extend the implementation. However, I'd suggest we could also handle this as a follow-up enhancement to keep this patch focused on the core functionality of enabling customizable authorization contexts.

wecharyu: The original implement has set `commandString` and `ipAddress` by default. It might be better to keep this behavior, in case external components already depend on them.

jjiang037: @wecharyu I enhance this implementation by enabling setting commandString and ipAddress. Could you please review it again? Thanks a lot

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
```diff
@@ -40,6 +44,30 @@ protected HiveMetaStoreAuthorizableEvent(PreEventContext preEventContext) {
...
   }
 
+  protected HiveAuthzContext buildAuthzContext(String commandString) {
+    HiveAuthzContext.Builder builder = new HiveAuthzContext.Builder();
+
+    if (commandString != null) {
+      builder.setCommandString(commandString);
+    }
+
+    builder.setForwardedAddresses(new ArrayList<>());
```
wecharyu, line 55: nit: also add the comment
```suggestion
    // TODO: refer to SessionManager/HiveSessionImpl for details on getting ipAddress and forwardedAddresses
    builder.setForwardedAddresses(new ArrayList<>());
```

wecharyu (approved):
LGTM+1 (non-binding)

nrg4878:
@saihemanth-cloudera @dengzhhu653 Could you please review as well?

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/ReadDatabaseEvent.java
```diff
@@ -40,10 +41,14 @@ public ReadDatabaseEvent(PreEventContext preEventContext) {
...
-    HiveMetaStoreAuthzInfo ret =
-        new HiveMetaStoreAuthzInfo(preEventContext, HiveOperationType.SHOWDATABASES, getInputHObjs(), getOutputHObjs(),
-            COMMAND_STR);
+  @Override
+  public HiveMetaStoreAuthzInfo getAuthzContext() {
+    List<HivePrivilegeObject> inputHObjs = getInputHObjs();
+    List<HivePrivilegeObject> outputHObjs = getOutputHObjs();
+    HiveAuthzContext authzContext = buildAuthzContext(COMMAND_STR);
+    HiveMetaStoreAuthzInfo ret = new HiveMetaStoreAuthzInfo(preEventContext, HiveOperationType.SHOWDATABASES,
+        inputHObjs, outputHObjs, COMMAND_STR, authzContext);
```
saihemanth-cloudera: nit: getInputHObjs(), getOutputHObjs() as we are not really using these variables: inputHObjs, outputHObjs

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/HiveMetaStoreAuthorizableEvent.java
```diff
@@ -40,6 +44,31 @@ protected HiveMetaStoreAuthorizableEvent(PreEventContext preEventContext) {
...
+
+    Map<String, Object> clientConfig = HiveMetaStoreAuthorizer.getClientConfig();
+    if (clientConfig != null) {
+      builder.setClientConfig(clientConfig);
+    }
+
+    return builder.build();
+  }
+
+  protected HiveAuthzContext buildAuthzContext() {
```
saihemanth-cloudera: I don't think we need this method, as we already have buildAuthzContext(String)

jintongjiang: Fix in the commit: 6c5d6b990a479a65fbba29f6879de959ba8fc6b9

ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/metastore/events/ReadDatabaseEvent.java
jintongjiang: fix in the commit: 6c5d6b990a479a65fbba29f6879de959ba8fc6b9

PR #6079, opened 2025-09-17
HIVE-29206: Remove commons-lang from hive-exec shaded jar

KiranVelumuri:

[HIVE-29206](https://issues.apache.org/jira/browse/HIVE-29206)

To completely remove commons-lang

No

jar -tvf


KiranVelumuri:
> To be sure we don't have/use commons-lang in the project we should add an [enforcer rule](https://github.com/apache/hive/blob/d90574c9c8d5f06b6fb0ba3fd94431375a97b286/pom.xml#L1706).

Added the enforcer rule. We already have a [check-banned-imports](https://github.com/apache/hive/blob/d90574c9c8d5f06b6fb0ba3fd94431375a97b286/pom.xml#L1764) in place for commons-lang. We do not have commons-lang anywhere in the project dependency tree and would not allow from now on.

itests/hive-unit/pom.xml
```diff
@@ -558,6 +558,13 @@
             </goals>
             <configuration>
               <rules>
+                <bannedDependencies>
```
deniskuzZ: why don't we reuse the root pom declarations?

standalone-metastore/metastore-common/pom.xml
```diff
@@ -581,6 +581,13 @@
                     <exclude>com.google.code.findbugs:annotations</exclude>
                   </excludes>
                 </bannedDependencies>
+                <bannedDependencies>
```
deniskuzZ: can we define this in root metastore pom.xml?

Aggarwal-Raghav: @KiranVelumuri , please check https://github.com/apache/hive/pull/6054, I have handled the same in the PR but as its not merged, please sync the banned dependencies in parent pom and standalone-metastore pom. I discovered that commons-logging was still getting shipped. So, IMO, its better to same set of banned dependencies across project.

KiranVelumuri: I defined enforcer-plugin in root metastore pom.xml for now. Yes, I too feel that it's better to maintain the same set across parent pom and metastore pom.

itests/hive-unit/pom.xml
KiranVelumuri: there's only junit-jupiter declared here which is different from those in root pom.

PR #6081, opened 2025-09-17
HIVE-28965: tez.DagUtils: Failed to add credential supplier, ClassNotFoundException: org.apache.hadoop.hive.kafka.KafkaDagCredentialSupplier

KiranVelumuri:

Warn instead of throw error in logs when KafkaDagCredentialSupplier is not available in classpath

Stacktrace for ClassNotFoundException for KafkaDagCredentialSupplier is printed for cases not relevant to current scenario(might not be Kafka related or need a credential supplier)

No

mvn test -Dtest=TestBeeLineWithArgs#testRowsAffected

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
@@ -313,7 +313,7 @@ public class DagUtils {
         Class<? extends DagCredentialSupplier> c = Class.forName(s).asSubclass(DagCredentialSupplier.class);
         dagSuppliers.add(c.getConstructor().newInstance());
       } catch (ReflectiveOperationException e) {
-        LOG.error("Failed to add credential supplier", e);
+        LOG.warn("Failed to add credential supplier: {}", s);
       }
     }
     return dagSuppliers;


ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
```diff
@@ -313,7 +313,7 @@ private static List<DagCredentialSupplier> defaultCredentialSuppliers() {
         Class<? extends DagCredentialSupplier> c = Class.forName(s).asSubclass(DagCredentialSupplier.class);
         dagSuppliers.add(c.getConstructor().newInstance());
       } catch (ReflectiveOperationException e) {
-        LOG.error("Failed to add credential supplier", e);
+        LOG.warn("Failed to add credential supplier: {}", s);
```
deniskuzZ: that's ok, but would be great to init those CredentialSuppliers only when used

KiranVelumuri: sure, would look into this.

KiranVelumuri:
@deniskuzZ We needed some check at DagUtils to add the kafka credential supplier only when needed. So I referred to obtainToken and added those checks(isTokenRequired) here. Could you please tell if this approach is ok here?

ql/src/java/org/apache/hadoop/hive/ql/exec/tez/DagUtils.java
```diff
@@ -314,7 +350,46 @@ private static List<DagCredentialSupplier> defaultCredentialSuppliers() {
...
+   *   <li>kafka.consumer.security.protocol</li>
+   *   <li>kafka.producer.security.protocol</li>
+   * </ol>
+   * <p>and the first non null/not empty is returned.</p>
+   * <p>Defining multiple security protocols at the same time is invalid but this method is lenient and tries to pick
+   * the most reasonable option.</p>
+   * @param props the properties from which to obtain the protocol.
+   * @return the security protocol if one is defined in the properties and null otherwise.
+   */
+  static SecurityProtocol getSecurityProtocol(Properties props) {
```
deniskuzZ: i think this functionality should be extracted to util class like `KafkaUtils`

KiranVelumuri: This was from KafkaUtils, I put it here to avoid dependency on kafka-handler(since it causes cyclic dependency)

deniskuzZ:
in my opinion it's a bit of overkill. We shouldn't copy code from kafka-handler, instead we might initialize suppliers lazily only when UserGroupInformation.isSecurityEnabled 
cc @abstractdog 

````
private void getCredentialsFromSuppliers(BaseWork work, Set<TableDesc> tables, DAG dag, JobConf conf) {
    if (!UserGroupInformation.isSecurityEnabled()){
      return;
    }
    ..
    for (DagCredentialSupplier supplier : credentialSuppliers.get()) {
      ....
    }
}

private final Supplier<List<DagCredentialSupplier>> credentialSuppliers;

DagUtils(Supplier<List<DagCredentialSupplier>> credentialSuppliers) {
  this.credentialSuppliers = credentialSuppliers;
}

private static final DagUtils instance = new DagUtils(() -> defaultCredentialSuppliers());
````

Note, this is only a test classpath issue, in production deployment kafka-handler jar would be always present.

KiranVelumuri:
yes @deniskuzZ. This is what I felt too and wanted your opinion on.

abstractdog:
> in my opinion it's a bit of overkill. We shouldn't copy code from kafka-handler, instead we might initialize suppliers lazily only when UserGroupInformation.isSecurityEnabled cc @abstractdog
> 
> ```
> private void getCredentialsFromSuppliers(BaseWork work, Set<TableDesc> tables, DAG dag, JobConf conf) {
>     if (!UserGroupInformation.isSecurityEnabled()){
>       return;
>     }
>     ..
>     lazyInitCredentialSuppliers();
> }
> ```
> 
> Note, this is only a test classpath issue, in production deployment kafka-handler jar would be always present.

the original goal was to completely remove kafka dependency from ql, which is the case for every other third party that hive provides a storage handler for, so we should maintain that behavior
I agree with trying the lazy approach if security is disabled as it would easily solve the test log noise

deniskuzZ:
i've updated the snippet, @KiranVelumuri please take a look https://github.com/apache/hive/pull/6081#issuecomment-3385045320

KiranVelumuri:
> i've updated the snippet, @KiranVelumuri please take a look [#6081 (comment)](https://github.com/apache/hive/pull/6081#issuecomment-3385045320)

@deniskuzZ Thank you for the snippet. Could you please review? Sorry I was away travelling for the past 3 days.

PR #6082, opened 2025-09-17
HIVE-29208: Infinite loop while compiling query with filter predicate containing disjuncts on the same expression

kasakrisz:
`HivePointLookupOptimizerRule.RexMergeInClause` goal is to merge `IN` and `Equals` expressions into one `IN` expression if all of these expressions has a reference to the same expression at one side and has a constant expression on the other side.
```
OR(a IN(1, 2), a IN (2, 3))
```
```
a IN(1, 2, 3)
```
This patch try to keep the original order of operands of the input expression and add a check whether any merge can be performed or not. If not then return the original expression.

The current code always creates a new expression using the operands of the original expression but it changes the order of the operands. This leads to infinite loop and `StackOverflowException`.

Let’s look at the following expression:
```
OR(
  IN(MINUTE(FLAG(MINUTE), CAST($1):TIMESTAMP(9)), 2, 10),
  IS NULL(CAST($1):TIMESTAMP(9))
)
```
This OR expression has two parts (disjuncts):
* `IN(MINUTE(FLAG(MINUTE), CAST($1):TIMESTAMP(9)), 2, 10)`
* `IS NULL(CAST($1):TIMESTAMP(9))`

When `ReduceExpressionsRule.FILTER_INSTANCE` simplifies the expression using `RexSimplify`, it reorders the operands so that `IS NULL(CAST($1):TIMESTAMP(9))` comes first:

```
OR(
  IS NULL(CAST($1):TIMESTAMP(9)),
  IN(MINUTE(FLAG(MINUTE), CAST($1):TIMESTAMP(9)), 2, 10)
)
```
After that, `HivePointLookupOptimizerRule.RexMergeInClause` moves `IS NULL(...)` back to the end of the list.

This back-and-forth reordering between the two rules causes an infinite loop.

No.

```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile=pointlookup6.q -pl itests/qtest
```


ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
```diff
@@ -659,14 +658,17 @@ private static RexNode handleAND(RexBuilder rexBuilder, RexCall call) {
...
       }
+      if (inExpressionCount + eqExpressionCount == inLHSExprToRHSExprs.keySet().size()) {
+        // No IN and Equality expressions can be merged, bail out
+        return call;
+      }
       // Create IN clauses
-      final List<RexNode> newOperands = createInClauses(rexBuilder,
-          visitedRefs, inLHSExprToRHSExprs, inLHSExprToRHSNullableExprs);
-      newOperands.addAll(operands);
+      newOperands.addAll(createInClauses(rexBuilder, visitedRefs, inLHSExprToRHSExprs, inLHSExprToRHSNullableExprs));
```
deniskuzZ, line 670: i would prefer local var and then `newOperands.addAll(operands)`, however, could we at least move `createInClauses` on a next line 
````
newOperands.addAll(
    createInClauses(rexBuilder, visitedRefs, inLHSExprToRHSExprs, inLHSExprToRHSNullableExprs)
);
````

File: ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java [hunk 2]
```diff
@@ -734,26 +736,30 @@ private static boolean shareSameType(
...
       }
+      if (inExpressionCount == inLHSExprToRHSExprs.keySet().size()) {
+        // No IN expressions can be merged, bail out
+        return call;
+      }
       // Create IN clauses (fourth parameter is not needed since no expressions were removed)
-      final List<RexNode> newOperands = createInClauses(rexBuilder,
-          inLHSExprToRHSExprs.keySet(), inLHSExprToRHSExprs, null);
-      newOperands.addAll(operands);
+      newOperands.addAll(createInClauses(rexBuilder,
```
deniskuzZ: same https://github.com/apache/hive/pull/6082/files#r2360354039

deniskuzZ (approved):
LGTM +1, just code-style comment

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
```diff
@@ -599,15 +599,15 @@ private static RexNode handleAND(RexBuilder rexBuilder, RexCall call) {
...
+      final List<RexNode> newOperands = new ArrayList<>(operands.size());
+      int inExpressionCount = 0;
+      int eqExpressionCount = 0;
+      for (RexNode operand : operands) {
+        if (operand instanceof RexCall inCall && HiveIn.INSTANCE.equals(inCall.op)) {
+          if (!HiveCalciteUtil.isDeterministic(inCall.getOperands().getFirst())) {
             continue;
           }
-          RexNode ref = inCall.getOperands().get(0);
+          RexNode ref = inCall.getOperands().getFirst();
```
mdayakar, line 607: `RexNode ref = inCall.getOperands().getFirst();` local variable creation can be moved up to [607](https://github.com/apache/hive/pull/6082/files#diff-0400ba8e7a318a20ac5a3abdb58ab1d639813cd5212a64bf108898929035b4e0R607) line so that it can be used there also.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
```diff
@@ -659,14 +658,17 @@ private static RexNode handleAND(RexBuilder rexBuilder, RexCall call) {
...
             inLHSExprToRHSExprs.put(c.exprNode, new SimilarRexNodeElement(c.constNode));
           }
-          operands.remove(i);
-          --i;
+          ++eqExpressionCount;
+        } else {
+          newOperands.add(operand);
         }
       }
+      if (inExpressionCount + eqExpressionCount == inLHSExprToRHSExprs.keySet().size()) {
```
mdayakar, line 665: Here why can't we use `inLHSExprToRHSExprs .size()` instead of `inLHSExprToRHSExprs.keySet().size()`?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
```diff
@@ -734,26 +736,30 @@ private static boolean shareSameType(
...
-          RexCall inCall = (RexCall) operand;
-          if (!HiveCalciteUtil.isDeterministic(inCall.getOperands().get(0))) {
+      int inExpressionCount = 0;
+      for (RexNode operand : operands) {
+        if (operand instanceof RexCall inCall && HiveIn.INSTANCE.equals(inCall.op)) {
+          if (!HiveCalciteUtil.isDeterministic(inCall.getOperands().getFirst())) {
             continue;
           }
-          RexNode ref = inCall.getOperands().get(0);
+          RexNode ref = inCall.getOperands().getFirst();
```
mdayakar, line 743: Here also `RexNode ref = inCall.getOperands().getFirst();` can be moved up so that it can be used in the `if (!HiveCalciteUtil.isDeterministic(inCall.getOperands().getFirst()))` check.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
```diff
@@ -734,26 +736,30 @@ private static boolean shareSameType(
...
             inLHSExprToRHSExprs.put(ref, new SimilarRexNodeElement(inCall.getOperands().get(j)));
           }
-          operands.remove(i);
-          --i;
+          ++inExpressionCount;
+        } else {
+          newOperands.add(operand);
         }
       }
+      if (inExpressionCount == inLHSExprToRHSExprs.keySet().size()) {
```
mdayakar, line 754: Here also, `inLHSExprToRHSExprs.size()` can be used instead of `inLHSExprToRHSExprs.keySet().size()`.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
zabetak, line 665: Is this if check equivalent to: `if (visitedRefs.size() == operands.size())`? Checking to see if we need all those new counters and bookeeping.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
```diff
@@ -659,14 +658,17 @@ private static RexNode handleAND(RexBuilder rexBuilder, RexCall call) {
           } else {
             inLHSExprToRHSExprs.put(c.exprNode, new SimilarRexNodeElement(c.constNode));
           }
-          operands.remove(i);
-          --i;
+          ++eqExpressionCount;
+        } else {
+          newOperands.add(operand);
```
zabetak: Do we hit this `else` block when we skip an operand due to the `isDeterminstic` check? I have the impression that the current version may silently drop operands when non-deterministic functions are involved.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/TestHivePointLookupOptimizerRule.java
```diff
@@ -520,4 +521,27 @@ public void testSameDecimalLiteralDifferentPrecision() {
     System.out.println(condition);
     assertEquals("IN($1, 10000:DECIMAL(19, 5), 11000:DECIMAL(19, 5))", condition.toString());
   }
+
+  @Test
+  public void testNothingToBeMergedInOrExpressionAndOperandOrderIsUnchanged() {
```
zabetak, line 526: Since the PR touches also the handleAND method to introduce bail out code there should be tests covering that part as well.

ql/src/test/queries/clientpositive/pointlookup6.q
```diff
@@ -17,3 +17,12 @@ WHERE r_table.string_col = l_table.string_col AND l_table.string_col IN ('AAA111
...
 WHERE r_table.string_col = l_table.string_col AND l_table.string_col IN ('AAA111', 'BBB222') AND r_table.string_col IN ('AAA111', 'BBB222');
+
+explain cbo
+SELECT * FROM r_table
+WHERE ( (
+         MINUTE(string_col) = 2 OR
+         MINUTE(string_col) = 10
+        )
+        OR (MINUTE(string_col) IS NULL)
+      );
```
zabetak: Is the additional OR nesting important for the repro? If not I would remove it to avoid any potential confusion. There is also `set hive.optimize.point.lookup.min=2;` in this file which adds some nuances with respect to if the rule is firing or not in this case.

zabetak: Additionally, if casts and timestamps are not important for the repro I would simplify the query even more (e.g., use `UPPER(string_col)`).

zabetak (commented):
The main point that needs clarification is that pertaining to losing non-deterministic operands. The rest of the comments are mostly details.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
kasakrisz: A key might have multiple values hence `inLHSExprToRHSExprs.size()` and `inLHSExprToRHSExprs.keySet().size()` are not equal: the first is the number of the elements in the collection and the second is the number of keys.

kasakrisz: The collection `operands` does not contain nested expressions as a separate item hence the `size()` is not accurate. See the test case [TestHivePointLookupOptimizerRule.testRecursionIsNotObstructed](https://github.com/apache/hive/blob/c338904b80d4dc6df08981738b69fa136bf6c624/ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/TestHivePointLookupOptimizerRule.java#L317-L358)

ql/src/test/queries/clientpositive/pointlookup6.q
kasakrisz: With `UPPER(string_col)` a Search operator is introduced and the issue is not reproducible.
The additional OR nesting is not necessary.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
kasakrisz: I applied this change. Thanks.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
kasakrisz: Done

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
kasakrisz: I changed the code to not lose non-deterministic operands. Thank you for catching this.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/TestHivePointLookupOptimizerRule.java
kasakrisz: There is already coverage in TestHivePointLookupOptimizerRule.java

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
kasakrisz: Thanks for the review Denys.
AFAIK Hive doesn't have a predefined coding standard for a long time. I only add new lines into a statement if it would be longer than 120 characters in a single line. This was the rule when we had checkstyle in the past.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
kasakrisz: Thanks catching this I aligned the format to the previous occurrence.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
zabetak: Checkstyle is back and any violations will now be reported via Sonarqube (HIVE-29196).

ql/src/test/results/clientpositive/llap/bucketpruning1.q.out
```diff
@@ -750,13 +750,13 @@ STAGE PLANS:
             Map Operator Tree:
                 TableScan
                   alias: srcbucket_pruned
-                  filterExpr: ((ds = '2008-04-08') and (key) IN (2, 3)) (type: boolean)
+                  filterExpr: ((key) IN (2, 3) and (ds = '2008-04-08')) (type: boolean)
```
zabetak, line 753: It seems that the new order is the same with the order of the predicates in the query.
```sql
select * from srcbucket_pruned where key in (2,3) and ds='2008-04-08'
```
I guess this is expected given that the new logic tries to retain the original expression and order of predicates if there are no changes.

ql/src/test/queries/clientpositive/pointlookup6.q
zabetak: The fact that some calls are transformed to SEARCH and some don't is troublesome so I logged https://issues.apache.org/jira/browse/CALCITE-7194 as a follow-up.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/TestHivePointLookupOptimizerRule.java
zabetak: I was mainly referring to the new bail-out condition that was added in `handleAND` method but I guess the code is covered also through .q files.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
zabetak: I suppose that since we didn't catch through the tests we are lacking some coverage. It would be nice to add some tests with non-deterministic operands as a follow-up.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
mdayakar: my bad, I didn't observe the type of `inLHSExprToRHSExprs` is a Multimap.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/rules/HivePointLookupOptimizerRule.java
mdayakar: my bad, I didn't observe the type of `inLHSExprToRHSExprs` is a Multimap.

ql/src/test/org/apache/hadoop/hive/ql/optimizer/calcite/rules/TestHivePointLookupOptimizerRule.java
kasakrisz: Sorry for the late response: 
There are more than one test cases in TestHivePointLookupOptimizerRule to cover this:
* bail out: testSimpleCase, testCaseInAndEqualsWithConstantsOfDifferentType, testSameVarcharAndNullLiterals, testRecursionIsNotObstructed
* do the merge: testSameVarcharLiteralDifferentPrecision, testSameDecimalLiteralDifferentPrecision, testInExprsMergedMultipleOverlap, testInExprsAndEqualsMerged

And the q tests which q.out files affected by this PR

PR #6085, opened 2025-09-17
HIVE-29211: add LDAP callback handlers to enforce group filter for ke…

jjiang037:
…rberos users

This PR adds the capability to apply LDAP group filters to Kerberos-authenticated users in both HS2 and HMS.
Key changes:
1. New configuration: hive.server2.authentication.ldap.enableGroupCheckAfterKerberos (default: false) to enable LDAP filtering for Kerberos users
2. LdapGroupCallbackHandler: New SASL callback handler that intercepts Kerberos authentication and applies existing LDAP filters (userSearchFilter, groupSearchFilter, customLDAPQuery, userFilter, groupFilter)
3. HadoopThriftAuthBridge enhancement: Modified to accept custom callback handlers for both HMS and HiveServer2
4. ThriftHttpServlet enhancement: Added LDAP filter enforcement for HTTP transport mode
5. Filter reuse: Leverages LdapAuthenticationProviderImpl.resolveFilter() to ensure consistent filter resolution across authentication methods

Currently, Hive has an authorization inconsistency:

1. Users authenticating via LDAP are subject to LDAP group filters
2. Users authenticating via Kerberos bypass these filters entirely

This creates security and operational issues. This PR is created for fixing these issues.

Yes. This PR introduces a new optional configuration parameter.
Previous behavior: Kerberos-authenticated users bypass all LDAP group filters.
New behavior: When hive.server2.authentication.ldap.enableGroupCheckAfterKerberos=true:

1. Kerberos-authenticated users must pass configured LDAP filters
2. Requires valid LDAP bind credentials
3. Applies to both HiveServer2 and HMS
4. Does not affect proxy users

ut, regression tests, functional tests

### DIFF
--- a/service/src/test/org/apache/hive/service/auth/TestLdapKerberosWithGroupFilter.java
+++ b/service/src/test/org/apache/hive/service/auth/TestLdapKerberosWithGroupFilter.java
@@ -21,20 +21,17 @@ package org.apache.hive.service.auth;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hive.service.auth.ldap.DirSearch;
 import org.apache.hive.service.auth.ldap.DirSearchFactory;
-import org.apache.hive.service.auth.ldap.Filter;
 import org.apache.hive.service.auth.ldap.LdapGroupCallbackHandler;
 
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
 
 import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.CallbackHandler;
-import javax.security.sasl.AuthenticationException;
 import javax.security.sasl.AuthorizeCallback;
 
 import java.util.Collections;
@@ -164,82 +161,6 @@ public class TestLdapKerberosWithGroupFilter {
   }
 
   @Test
-  public void testKerberosAuthWithMultipleLdapGroupCheckPositive() throws Exception {
-    // Configure LDAP to allow users in either group1 or group2
-    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_GROUPFILTER, GROUP1_NAME + "," + GROUP2_NAME);
-    conf.setBoolVar(HiveConf.ConfVars.HIVE_SERVER2_LDAP_ENABLE_GROUP_CHECK_AFTER_KERBEROS, true);
-
-    // Test user1 in group1
-    when(dirSearch.findUserDn(USER1_ID)).thenReturn("uid=user1,dc=example,dc=com");
-    when(dirSearch.findGroupsForUser("uid=user1,dc=example,dc=com")).thenReturn(
-        Collections.singletonList("cn=group1,dc=example,dc=com"));
-
-    LdapGroupCallbackHandler callbackHandler1 = LdapGroupCallbackHandler.createForTesting(
-        conf, dirSearchFactory, delegateHandler);
-
-    AuthorizeCallback ac1 = new AuthorizeCallback(USER1_PRINCIPAL, USER1_PRINCIPAL);
-    callbackHandler1.handle(new Callback[]{ac1});
-    assertTrue("User1 should be authorized", ac1.isAuthorized());
-
-    // Reset mocks for user2 test
-    reset(dirSearch);
-    when(dirSearch.findUserDn(USER2_ID)).thenReturn("uid=user2,dc=example,dc=com");
-    when(dirSearch.findGroupsForUser("uid=user2,dc=example,dc=com")).thenReturn(
-        Collections.singletonList("cn=group2,dc=example,dc=com"));
-
-    // Need to reset dirSearchFactory mock to return the updated dirSearch
-    reset(dirSearchFactory);
-    when(dirSearchFactory.getInstance(any(HiveConf.class), anyString(), anyString()))
-        .thenReturn(dirSearch);
-
-    LdapGroupCallbackHandler callbackHandler2 = LdapGroupCallbackHandler.createForTesting(
-        conf, dirSearchFactory, delegateHandler);
-
-    AuthorizeCallback ac2 = new AuthorizeCallback(USER2_PRINCIPAL, USER2_PRINCIPAL);
-    callbackHandler2.handle(new Callback[]{ac2});
-    assertTrue("User2 should be authorized", ac2.isAuthorized());
-  }
-
-  @Test
-  public void testKerberosAuthWithUserGroupSearchFilter() throws Exception {
-    // Configure UserGroupSearchFilter
-    conf.setBoolVar(HiveConf.ConfVars.HIVE_SERVER2_LDAP_ENABLE_GROUP_CHECK_AFTER_KERBEROS, true);
-
-    String userSearchFilter = "(&(uid={0})(objectClass=person))";
-    String baseDn = "dc=example,dc=com";
-    String groupSearchFilter = "(&(memberUid={0})(objectClass=posixGroup))";
-    String groupBaseDn = "ou=groups,dc=example,dc=com";
-
-    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_USERSEARCHFILTER, userSearchFilter);
-    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_BASEDN, baseDn);
-    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_GROUPSEARCHFILTER, groupSearchFilter);
-    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_GROUPBASEDN, groupBaseDn);
-
-    String userDn = "uid=user1,dc=example,dc=com";
-
-    when(dirSearch.findUserDn(eq(USER1_ID), eq(userSearchFilter), eq(baseDn)))
-        .thenReturn(userDn);
-
-    when(dirSearch.executeUserAndGroupFilterQuery(
-        eq(USER1_ID),
-        eq(userDn),
-        eq(groupSearchFilter),
-        eq(groupBaseDn)))
-        .thenReturn(Collections.singletonList("cn=group1,ou=groups,dc=example,dc=com"));
-
-    LdapGroupCallbackHandler callbackHandler = LdapGroupCallbackHandler.createForTesting(
-        conf, dirSearchFactory, delegateHandler);
-
-    AuthorizeCallback ac = new AuthorizeCallback(USER1_PRINCIPAL, USER1_PRINCIPAL);
-    callbackHandler.handle(new Callback[]{ac});
-
-    assertTrue("User should be authorized with UserGroupSearchFilter", ac.isAuthorized());
-
-    verify(dirSearch).findUserDn(eq(USER1_ID), eq(userSearchFilter), eq(baseDn));
-    verify(dirSearch).executeUserAndGroupFilterQuery(eq(USER1_ID), eq(userDn), eq(groupSearchFilter), eq(groupBaseDn));
-  }
-
-  @Test
   public void testKerberosAuthWithDisabledLdapGroupCheck() throws Exception {
     // Disable LDAP group check
     conf.setBoolVar(HiveConf.ConfVars.HIVE_SERVER2_LDAP_ENABLE_GROUP_CHECK_AFTER_KERBEROS, false);
@@ -262,68 +183,4 @@ public class TestLdapKerberosWithGroupFilter {
     verifyNoInteractions(dirSearch);
   }
 
-  @Test
-  public void testDirectFilterApplication() throws Exception {
-    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_GROUPFILTER, GROUP1_NAME);
-
-    String userDn = "uid=user1,dc=example,dc=com";
-    String groupDn = "cn=group1,dc=example,dc=com";
-
-    when(dirSearch.findUserDn(USER1_ID)).thenReturn(userDn);
-    when(dirSearch.findGroupsForUser(eq(userDn))).thenReturn(Collections.singletonList(groupDn));
-
-    Filter filter = LdapAuthenticationProviderImpl.resolveFilter(conf);
-    assertNotNull("Filter should be resolved", filter);
-
-    filter.apply(dirSearch, USER1_ID);
-
-    verify(dirSearch, times(2)).findUserDn(USER1_ID);
-    verify(dirSearch).findGroupsForUser(eq(userDn));
-  }
-
-  @Test(expected = AuthenticationException.class)
-  public void testDirectFilterApplicationFailure() throws Exception {
-    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_GROUPFILTER, GROUP1_NAME);
-
-    String userDn = "uid=user2,dc=example,dc=com";
-    String wrongGroupDn = "cn=group3,dc=example,dc=com";
-
-    when(dirSearch.findUserDn(USER2_ID)).thenReturn(userDn);
-    when(dirSearch.findGroupsForUser(eq(userDn))).thenReturn(Collections.singletonList(wrongGroupDn));
-
-    Filter filter = LdapAuthenticationProviderImpl.resolveFilter(conf);
-    assertNotNull("Filter should be resolved", filter);
-
-    filter.apply(dirSearch, USER2_ID);
-  }
-
-  @Test
-  public void testKerberosAuthWithMixedAuthorizeCallbacks() throws Exception {
-    conf.setBoolVar(HiveConf.ConfVars.HIVE_SERVER2_LDAP_ENABLE_GROUP_CHECK_AFTER_KERBEROS, true);
-    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_GROUPFILTER, GROUP1_NAME);
-
-    String userDn = "uid=user1,dc=example,dc=com";
-    String groupDn = "cn=group1,dc=example,dc=com";
-
-    when(dirSearch.findUserDn(USER1_ID)).thenReturn(userDn);
-    when(dirSearch.findGroupsForUser(userDn)).thenReturn(Collections.singletonList(groupDn));
-
-    LdapGroupCallbackHandler callbackHandler = LdapGroupCallbackHandler.createForTesting(
-        conf, dirSearchFactory, delegateHandler);
-
-    AuthorizeCallback authorized = new AuthorizeCallback(USER1_PRINCIPAL, USER1_PRINCIPAL);
-    AuthorizeCallback delegated = new AuthorizeCallback(USER1_PRINCIPAL, USER2_PRINCIPAL);
-
-    Callback[] callbacks = {authorized, delegated};
-    callbackHandler.handle(callbacks);
-
-    assertTrue("Matching IDs should be authorized", authorized.isAuthorized());
-    ArgumentCaptor<Callback[]> captor = ArgumentCaptor.forClass(Callback[].class);
-    verify(delegateHandler).handle(captor.capture());
-    Callback[] delegatedCallbacks = captor.getValue();
-    assertEquals(2, delegatedCallbacks.length);
-    assertSame(authorized, delegatedCallbacks[0]);
-    assertSame(delegated, delegatedCallbacks[1]);
-    assertFalse("Delegated callback should remain unauthorized", delegated.isAuthorized());
-  }
 }
--- a/service/src/test/org/apache/hive/service/auth/ldap/TestLdapGroupCallbackHandler.java
+++ b/service/src/test/org/apache/hive/service/auth/ldap/TestLdapGroupCallbackHandler.java
@@ -29,8 +29,6 @@ import javax.security.auth.callback.Callback;
 import javax.security.auth.callback.UnsupportedCallbackException;
 import javax.security.sasl.AuthorizeCallback;
 
-import java.util.Collections;
-
 import static org.junit.Assert.*;
 import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.*;
@@ -58,8 +56,6 @@ public class TestLdapGroupCallbackHandler {
     conf = new HiveConf();
     conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_BIND_USER, "bindUser");
     conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_BIND_PASSWORD, "bindPassword");
-    when(dirSearchFactory.getInstance(any(HiveConf.class), anyString(), anyString()))
-        .thenReturn(dirSearch);
     doAnswer(invocation -> {
       Object[] args = invocation.getArguments();
       if (args.length > 0 && args[0] instanceof Callback[]) {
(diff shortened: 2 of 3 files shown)


common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
```diff
@@ -4353,6 +4353,13 @@ public static enum ConfVars {
     HIVE_SERVER2_PLAIN_LDAP_BIND_PASSWORD("hive.server2.authentication.ldap.bindpw", null,
         "The password for the bind user, to be used to search for the full name of the user being authenticated.\n" +
         "If the username is specified, this parameter must also be specified."),
+    HIVE_SERVER2_LDAP_ENABLE_GROUP_CHECK_AFTER_KERBEROS(
```
saihemanth-cloudera, line 4356: IMO, instead of adding a new config, we should enable group check by default if the auth mode is set to Kerberos

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,154 @@
...
+  public void handle(Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+    List<Callback> unhandledCallbacks = new ArrayList<>();
+
+    for (Callback callback : callbacks) {
+      if (callback instanceof AuthorizeCallback) {
+        AuthorizeCallback ac = (AuthorizeCallback) callback;
+        String authenticationID = ac.getAuthenticationID();
+        String authorizationID = ac.getAuthorizationID();
+
+        if (!authenticationID.equals(authorizationID)) {
```
saihemanth-cloudera: Is there a chance that authorizationID or authenticationID be null? If so, we'll get a NullPointerException here

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,154 @@
...
+          continue;
+        }
+
+        // Skip LDAP check for proxy users.
+        String proxyUser = SessionManager.getProxyUserName();
+        if (proxyUser != null && !proxyUser.isEmpty()) {
+          LOG.debug("Skipping LDAP filters for proxy user authorization");
+          ac.setAuthorized(true);
+          continue;
+        }
```
saihemanth-cloudera: This class executes at the service setup stage, do we really need this? because we anyway skip proxy user authorization check at run time.

service/src/test/org/apache/hive/service/auth/TestLdapKerberosWithGroupFilter.java
```diff
@@ -0,0 +1,312 @@
...
+    conf = new HiveConf();
+    conf.set("hive.root.logger", "DEBUG,console");
+
+    // Setup LDAP connection parameters
+    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_URL, "ldap://localhost:10389");
+    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_BIND_USER, "cn=admin,dc=example,dc=com");
+    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_BIND_PASSWORD, "admin");
+
+    // Configure Kerberos auth
+    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_AUTHENTICATION, "KERBEROS");
```
saihemanth-cloudera, line 78: Can you also verify multiple auth scenarios?

service/src/test/org/apache/hive/service/auth/TestLdapKerberosWithGroupFilter.java
```diff
@@ -0,0 +1,312 @@
...
+    conf.setVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_GROUPFILTER, GROUP1_NAME);
+
+    LdapGroupCallbackHandler callbackHandler = new LdapGroupCallbackHandler(
+        conf, dirSearchFactory, delegateHandler);
+
+    AuthorizeCallback ac1 = new AuthorizeCallback(USER1_PRINCIPAL, USER1_PRINCIPAL);
+    AuthorizeCallback ac2 = new AuthorizeCallback(USER2_PRINCIPAL, USER2_PRINCIPAL);
+
+    callbackHandler.handle(new Callback[]{ac1});
+    callbackHandler.handle(new Callback[]{ac2});
```
saihemanth-cloudera: nit: callbackHandler.handle(new Callback[]{ac1, ac2});

service/src/test/org/apache/hive/service/auth/ldap/TestLdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,288 @@
...
+        .when(delegateHandler).handle(any(Callback[].class));
+
+    callbackHandler = new LdapGroupCallbackHandler(conf, dirSearchFactory, delegateHandler);
+
+    try {
+      callbackHandler.handle(callbacks);
+      fail("Expected UnsupportedCallbackException");
+    } catch (UnsupportedCallbackException e) {
+      assertEquals(unsupportedCallback, e.getCallback());
+    }
```
saihemanth-cloudera: Instead of this, why not add: @Test(expected = UnsupportedCallbackException.class)

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,154 @@
...
+    } catch (AuthenticationException e) {
+      LOG.warn("User {} failed LDAP filter validation: {}", user, e.getMessage());
+      return false;
+    } catch (Exception e) {
+      LOG.error("Error applying LDAP filter for user {}", user, e);
+      return false;
+    }
+  }
+
+  private String extractUserName(@NotNull String principal) {
```
saihemanth-cloudera: make this public and add '@VisibleForTesting' and use this in TestThriftHttpKerberosLdapFilter.java#L299

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/AuthFactory.java
```diff
@@ -21,6 +21,7 @@
 import java.io.IOException;
 import java.util.HashMap;
 
+import javax.security.auth.callback.CallbackHandler;
```
saihemanth-cloudera: Is this required?

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
jjiang037: If we enable group checks by default, could this cause issues for users who are only relying on Kerberos?

service/src/test/org/apache/hive/service/auth/TestLdapKerberosWithGroupFilter.java
jjiang037: Address in the commit: df2496fda7f2c16fdc79b36187b9acd9b8a4c859

service/src/test/org/apache/hive/service/auth/TestLdapKerberosWithGroupFilter.java
jjiang037: Address in the commit: https://github.com/apache/hive/commit/df2496fda7f2c16fdc79b36187b9acd9b8a4c859

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: Address in the commit: https://github.com/apache/hive/commit/df2496fda7f2c16fdc79b36187b9acd9b8a4c859

service/src/test/org/apache/hive/service/auth/ldap/TestLdapGroupCallbackHandler.java
jjiang037: Address in the commit: https://github.com/apache/hive/commit/df2496fda7f2c16fdc79b36187b9acd9b8a4c859

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/AuthFactory.java
jjiang037: Address in the commit: https://github.com/apache/hive/commit/df2496fda7f2c16fdc79b36187b9acd9b8a4c859

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: Address in the commit: https://github.com/apache/hive/commit/df2496fda7f2c16fdc79b36187b9acd9b8a4c859

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: Address in the commit: https://github.com/apache/hive/commit/df2496fda7f2c16fdc79b36187b9acd9b8a4c859

common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
dengzhhu653: A new config makes sense to me, as not all of users should have configured both Kerberos and LDAP in HS2

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,153 @@
...
+    this.filter = enableLdapGroupCheck ? LdapAuthenticationProviderImpl.resolveFilter(conf) : null;
+
+    if (enableLdapGroupCheck && filter == null) {
+      LOG.warn("LDAP group check enabled but no filters configured");
+    }
+  }
+
+  @Override
+  public void handle(Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+    List<Callback> unhandledCallbacks = new ArrayList<>();
```
dengzhhu653: Can you try `delegate.handle(callbacks)` first? looks like this applies the LDAP filter before the authentication

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,153 @@
...
+        return false;
+      }
+
+      DirSearch dirSearch = this.dirSearchFactory.getInstance(conf, bindDN, bindPassword);
+
+      filter.apply(dirSearch, user);
+      LOG.debug("User {} passed LDAP filter validation", user);
+      return true;
+
+    } catch (AuthenticationException e) {
```
dengzhhu653: merge with the generic `catch (Exception e)` ?

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,153 @@
...
+      String bindDN = conf.getVar(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_BIND_USER);
+      char[] rawPassword = conf.getPassword(HiveConf.ConfVars.HIVE_SERVER2_PLAIN_LDAP_BIND_PASSWORD.varname);
+      String bindPassword = (rawPassword == null) ? null : new String(rawPassword);
+
+      if (StringUtils.isBlank(bindDN) || StringUtils.isBlank(bindPassword)) {
+        LOG.error("LDAP bind DN or password is not configured");
+        return false;
+      }
+
+      DirSearch dirSearch = this.dirSearchFactory.getInstance(conf, bindDN, bindPassword);
```
dengzhhu653: nit: move `dirSearch` as a global field?

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,153 @@
...
+  private final boolean enableLdapGroupCheck;
+  private final CallbackHandler delegateHandler;
+  private final DirSearchFactory dirSearchFactory;
+  private final Filter filter;
+
+  public LdapGroupCallbackHandler(HiveConf conf) {
+    this(conf, new LdapSearchFactory(), new SaslRpcServer.SaslGssCallbackHandler());
+  }
+
+  public LdapGroupCallbackHandler(HiveConf conf, DirSearchFactory dirSearchFactory, CallbackHandler delegateHandler) {
```
dengzhhu653: nit: add the annotation VisibleForTesting, and remove the public modifier?

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,153 @@
...
+      LOG.warn("User {} failed LDAP filter validation: {}", user, e.getMessage());
+      return false;
+    } catch (Exception e) {
+      LOG.error("Error applying LDAP filter for user {}", user, e);
+      return false;
+    }
+  }
+
+  @VisibleForTesting
+  public static String extractUserName(@NotNull String principal) {
```
dengzhhu653: use `SaslRpcServer.splitKerberosName(principal)[0]`?

service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpServlet.java
```diff
@@ -540,6 +553,54 @@ public String run() throws HttpAuthenticationException {
...
+        return;
+      }
+      // Check if this is a proxy user - we don't apply LDAP filters to proxy users
+      String proxyUser = SessionManager.getProxyUserName();
+      if (proxyUser != null && !proxyUser.isEmpty()) {
+        LOG.debug("Skipping LDAP filters for proxy user authentication");
+        return;
+      }
+
+      Filter filter = LdapAuthenticationProviderImpl.resolveFilter(hiveConf);
```
dengzhhu653: The following codes are similar to LdapGroupCallbackHandler, can we extract them to a new singleton class?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/security/HadoopThriftAuthBridge.java
```diff
@@ -374,6 +380,12 @@ protected Server(String keytabFile, String principalConf, String clientConf)
       }
     }
 
+    public Server(String keytabFile, String principalConf, String clientConf, CallbackHandler callbackHandler)
+            throws TTransportException {
+      this(keytabFile, principalConf, clientConf);
+      this.callbackHandler = callbackHandler;
```
dengzhhu653: nit: you can add a new method `setCallBackHandler` to init the callback compare to a new constructor

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: Fix in the commit: 720b5b8f9f1515767f83f0c1dd7ae1b76733ff02

service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpServlet.java
jjiang037: Extract them to https://github.com/apache/hive/pull/6085/commits/720b5b8f9f1515767f83f0c1dd7ae1b76733ff02#diff-b7ce549e080ea84afc3fb5a5b96e89c54c5a38a0ffeadb7aa715fe2c4adb1788

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: fix in the commit: 720b5b8f9f1515767f83f0c1dd7ae1b76733ff02

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: fix in the commit: https://github.com/apache/hive/commit/720b5b8f9f1515767f83f0c1dd7ae1b76733ff02

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/security/HadoopThriftAuthBridge.java
jjiang037: fix in the commit: https://github.com/apache/hive/commit/720b5b8f9f1515767f83f0c1dd7ae1b76733ff02

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: fix in the commit: https://github.com/apache/hive/commit/720b5b8f9f1515767f83f0c1dd7ae1b76733ff02

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: I think dirSearch holds stateful LDAP context objects that aren’t guaranteed to be thread-safe. Reusing a single instance could let concurrent Kerberos handshakes stomp on each other or reuse a closed LDAP connection. Considering this, I prefer to instantiate it inside enforceFilter

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,135 @@
...
+  @Override
+  public void handle(Callback[] callbacks) throws IOException, UnsupportedCallbackException {
+    delegateHandler.handle(callbacks);
+
+    for (Callback callback : callbacks) {
+      if (!(callback instanceof AuthorizeCallback)) {
+        continue;
+      }
+
+      AuthorizeCallback ac = (AuthorizeCallback) callback;
```
dengzhhu653, line 68: should we check the filter only when `ac.isAuthorized()`?

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,135 @@
...
+      }
+
+      if (!enableLdapGroupCheck || filter == null) {
+        ac.setAuthorized(true);
+        continue;
+      }
+
+      String user = extractUserName(authenticationID);
+      boolean authorized = applyLdapFilter(user);
+      ac.setAuthorized(authorized);
```
dengzhhu653, line 78: if `authorized` returned is false, we might need to set `ac.setAuthorized(false);` and remove the `AuthorizedID` from ac

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,135 @@
...
+
+    if (enableLdapGroupCheck && filter == null) {
+      LOG.warn("LDAP group check enabled but no filters configured");
+    }
+  }
+
+  @VisibleForTesting
+  public static LdapGroupCallbackHandler createForTesting(HiveConf conf, DirSearchFactory dirSearchFactory,
+      CallbackHandler delegateHandler) {
+    return new LdapGroupCallbackHandler(conf, dirSearchFactory, delegateHandler);
```
dengzhhu653, line 56: if we use a method to create the test instance, can we remove the `LdapGroupCallbackHandler(HiveConf conf, DirSearchFactory dirSearchFactory, CallbackHandler delegateHandler) `, in this method, we can 
```
 LdapGroupCallbackHandler handler = new LdapGroupCallbackHandler(conf);
 handler.delegateHandler = delegateHandler;
...

```

service/src/java/org/apache/hive/service/auth/ldap/KerberosLdapFilterEnforcer.java
```diff
@@ -0,0 +1,85 @@
...
+
+import javax.security.sasl.AuthenticationException;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.hive.conf.HiveConf;
+
+/**
+ * Singleton helper that applies configured LDAP filters for Kerberos-authenticated users.
+ */
+public final class KerberosLdapFilterEnforcer {
```
dengzhhu653, line 34: Personally I would like this class exposing the only `applyLdapFilter(username)` method, and holding all related to the Ldap authentication, initialization, like:

```
private KerberosLdapFilterEnforcer(HiveCon conf, DirSearchFactory factory) {
...
}
// authentication, 
boolean applyLdapFilter(String userName) {
...
}

public static synchronized KerberosLdapFilterEnforcer getInstance(HiveCon conf, DirSearchFactory factory) {
  return new KerberosLdapFilterEnforcer(conf, factory) 
}

```

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,135 @@
...
+    for (Callback callback : callbacks) {
+      if (!(callback instanceof AuthorizeCallback)) {
+        continue;
+      }
+
+      AuthorizeCallback ac = (AuthorizeCallback) callback;
+      String authenticationID = ac.getAuthenticationID();
+      String authorizationID = ac.getAuthorizationID();
+
+      if (StringUtils.isBlank(authenticationID) || StringUtils.isBlank(authorizationID)) {
```
dengzhhu653: this should always be true in case `ac.isAuthorized`

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,135 @@
...
+        LOG.debug("Missing authentication or authorization ID; skipping LDAP filter");
+        continue;
+      }
+
+      if (!authenticationID.equals(authorizationID)) {
```
dengzhhu653: this should always be true in case ac.isAuthorized

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
```diff
@@ -0,0 +1,135 @@
...
+        LOG.debug("Skipping LDAP filter for mismatched auth IDs");
+        continue;
+      }
+
+      if (!enableLdapGroupCheck || filter == null) {
```
dengzhhu653: nit: move this to `applyLdapFilter` as well

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: fix in the commit: 4783e43068d8656aa97fa93f6a5e7a716d6ff9f0

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: fix in the commit: https://github.com/apache/hive/commit/4783e43068d8656aa97fa93f6a5e7a716d6ff9f0

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: fix in the commit: https://github.com/apache/hive/commit/4783e43068d8656aa97fa93f6a5e7a716d6ff9f0

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: The tests still call that constructor directly so they can inject both the DirSearchFactory and delegate while keeping those fields final. If we removed it and set the delegate after construction, we’d either need setters or non-final fields, which I’d rather avoid. What do you think about it?

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: fix in the commit: https://github.com/apache/hive/commit/4783e43068d8656aa97fa93f6a5e7a716d6ff9f0

service/src/java/org/apache/hive/service/auth/ldap/LdapGroupCallbackHandler.java
jjiang037: fix in the commit: https://github.com/apache/hive/commit/4783e43068d8656aa97fa93f6a5e7a716d6ff9f0

service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpServlet.java
```diff
@@ -540,6 +556,35 @@ public String run() throws HttpAuthenticationException {
...
 
+    private void enforceLdapFilters(String shortName) throws HttpAuthenticationException {
+      boolean enableGroupCheck = hiveConf.getBoolVar(
+          HiveConf.ConfVars.HIVE_SERVER2_LDAP_ENABLE_GROUP_CHECK_AFTER_KERBEROS);
+      if (!enableGroupCheck) {
+        LOG.debug("No LDAP group check is enabled; skipping.");
+        return;
+      }
+      // Check if this is a proxy user - we don't apply LDAP filters to proxy users
+      String proxyUser = SessionManager.getProxyUserName();
```
dengzhhu653: Here we are building the connection, I guess the `proxyUser` is always null. In any case do we see this value as a non-empty string?

service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpServlet.java
```diff
@@ -60,7 +60,16 @@
 import org.apache.hive.service.auth.PasswdAuthenticationProvider;
 import org.apache.hive.service.auth.PlainSaslHelper;
 import org.apache.hive.service.auth.jwt.JWTValidator;
+import org.apache.hive.service.auth.ldap.CustomQueryFilterFactory;
```
dengzhhu653: some unused imports

service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpServlet.java
```diff
@@ -516,7 +528,11 @@ public String run() throws HttpAuthenticationException {
               "unable to establish context with the service ticket " +
               "provided by the client.");
         } else {
-          return getPrincipalWithoutRealmAndHost(gssContext.getSrcName().toString());
+          String principal = gssContext.getSrcName().toString();
+          String shortName = getPrincipalWithoutRealmAndHost(principal);
+          LOG.debug("Kerberos authentication successful");
+          enforceLdapFilters(principal);
```
dengzhhu653: can we call `boolean authorized = filterEnforcer.applyLdapFilter(principal); if (!authorized) throw HttpAuthenticationException` here?

service/src/java/org/apache/hive/service/auth/ldap/KerberosLdapFilterEnforcer.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /**
+   * Applies configured LDAP filters to authenticate a user principal.
+   *
+   * @param principal Kerberos principal to validate
+   * @return {@code true} if the principal passes all configured filters; {@code false} otherwise
+   */
+  public boolean applyLdapFilter(String principal) {
+    if (!enableLdapGroupCheck || filter == null) {
```
dengzhhu653: nit: `if (! isFilterConfigured())`

service/src/java/org/apache/hive/service/auth/ldap/KerberosLdapFilterEnforcer.java
```diff
@@ -0,0 +1,104 @@
...
+
+    String user = extractUserName(principal);
+    try (DirSearch dirSearch = createDirSearch()) {
+      filter.apply(dirSearch, user);
+      LOG.debug("Principal {} passed LDAP filter validation", principal);
+      return true;
+    } catch (AuthenticationException e) {
+      LOG.warn("Principal {} failed LDAP filter validation: {}", principal, e.getMessage());
+      return false;
+    } catch (Exception e) {
```
dengzhhu653, line 73: nit: as we don't care about what type of exception is in the caller, can we just use `catch (Exception e)` alone?

jintongjiang: fix in decc8688eb8a663d5d60a9481b55a726b9d4ab68

service/src/java/org/apache/hive/service/auth/ldap/KerberosLdapFilterEnforcer.java
jintongjiang: fix in decc8688eb8a663d5d60a9481b55a726b9d4ab68

service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpServlet.java
jintongjiang: fix in decc8688eb8a663d5d60a9481b55a726b9d4ab68

service/src/java/org/apache/hive/service/cli/thrift/ThriftHttpServlet.java
jintongjiang: fix in decc8688eb8a663d5d60a9481b55a726b9d4ab68

dengzhhu653:
+1. There is a test failure seems related, could you take a look please? also if we can remove some tests, that's better

dengzhhu653 (approved):

PR #6086, opened 2025-09-18
HIVE-29020: Support OAuth 2 in Iceberg REST Catalog

okumin:

Add RFC 7662 or RFC 9068-based OAuth 2 support.

https://issues.apache.org/jira/browse/HIVE-29020

OAuth 2 is the most common authorization method for protecting the Iceberg REST Catalog. As the Iceberg library officially supports it, Iceberg users would be able to use it without custom patches if we were to support it. Additionally, OAuth 2 is an industry-standard protocol as of 2025.

The security principles and considerations are written in the following document.

https://docs.google.com/document/d/1wOlmpKP4jZb4Je67wusdMoQ4VWgGfi5JLe8w3MP5mpE/edit?usp=sharing

No.

Added integration tests.


standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
```diff
@@ -127,8 +117,6 @@ public HMSCatalogAdapter(Catalog catalog) {
   }
 
   enum Route {
-    TOKENS(HTTPMethod.POST, "v1/oauth/tokens", null),
```
okumin, line 130: The initial specification of Iceberg REST allows the Iceberg REST Catalog to act as an Authorization Server, which is why the list contains this endpoint. As this does not follow security best practices, [it will be removed](https://github.com/apache/iceberg/blob/7eef46da0d00edb79b20d24cb7806f6ea8da4000/open-api/rest-catalog-open-api.yaml#L180-L181).

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/auth/jwt/JWTValidator.java
```diff
@@ -46,64 +47,35 @@
...
-    // verify claims
-    JWTClaimsSet claimsSet = parsedJwt.getJWTClaimsSet();
-    Date expirationTime = claimsSet.getExpirationTime();
-    if (expirationTime != null) {
-      Date now = new Date();
-      if (now.after(expirationTime)) {
-        LOG.warn("Rejecting an expired JWT: {}", parsedJwt.getPayload());
-        throw new AuthenticationException("JWT (ends with " + lastSevenChars + ") has been expired");
-      }
+  public JWTValidator(Set<JOSEObjectType> acceptableTypes, List<URL> jwksURLs, String expectedIssuer,
```
okumin, line 55: I replaced our custom implementation, following the article below.
https://connect2id.com/products/nimbus-jose-jwt/examples/validating-jwt-access-tokens

standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/auth/oauth2/TestTokenIntrospectionAuthenticator.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+  }
+
+  @BeforeClass
+  public static void setup() throws URISyntaxException, InterruptedException {
+    container = new GenericContainer<>(DockerImageName.parse("quay.io/keycloak/keycloak:26.3.4"))
+        .withEnv("KEYCLOAK_ADMIN","admin")
+        .withEnv("KEYCLOAK_ADMIN_PASSWORD","admin")
+        .withCommand("start-dev")
+        .withExposedPorts(8080);
```
okumin: I initially tried to use `spring-security-oauth2-authorization-server`, but gave it up because it has a hard dependency conflict on spring-core derived from Apache Atlas, and Nimbus we're using.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1873,8 +1873,55 @@ public enum ConfVars {
             " positive value will be used as-is."
     ),
     CATALOG_SERVLET_AUTH("metastore.catalog.servlet.auth",
-        "hive.metastore.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt"),
-        "HMS Catalog servlet authentication method (none, simple, or jwt)."
+        "hive.metastore.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt", "oauth2"),
```
deniskuzZ, line 1876: should we rename to `metastore.catalog.servlet.auth.type`? probably not, since it's already in use.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
```diff
@@ -100,7 +102,9 @@ private Catalog createCatalog() {
    */
   private HttpServlet createServlet(Catalog catalog) {
     String authType = MetastoreConf.getVar(configuration, ConfVars.CATALOG_SERVLET_AUTH);
-    ServletSecurity security = new ServletSecurity(AuthType.fromString(authType), configuration);
+    // Iceberg REST client uses "catalog" by default
```
deniskuzZ, line 105: could you please elaborate on the scope? what are the other values?

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/extension/HiveRESTCatalogServerExtension.java
```diff
@@ -69,6 +87,11 @@ public void beforeAll(ExtensionContext context) throws Exception {
     if (jwksServer != null) {
       jwksServer.start();
     }
+    if (authorizationServer != null) {
+      authorizationServer.start();
+      LOG.error(authorizationServer.getIssuer());
```
deniskuzZ: why not info?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1873,8 +1873,55 @@ public enum ConfVars {
...
+        "How to evaluate an access token. When your authorization server issues opaque tokens or you need " +
+        "to consider additional security requirements such as token revocations, use introspection."
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_AUDIENCE("metastore.catalog.servlet.auth.oauth2.audience",
+        "hive.metastore.catalog.servlet.auth.oauth2.audience", "",
+        "The acceptable name in the audience(aud) claim.  This is required when you use " +
+        "metastore.catalog.servlet.auth=oauth2"
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_CLIENT_ID("metastore.catalog.servlet.auth.oauth2.client.id",
+        "hive.metastore.catalog.servlet.auth.oauth2.client.id", "",
```
deniskuzZ, line 1896: maybe rename to
```
client-id
client-secret
```

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1873,8 +1873,55 @@ public enum ConfVars {
...
     ),
     CATALOG_SERVLET_AUTH("metastore.catalog.servlet.auth",
-        "hive.metastore.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt"),
-        "HMS Catalog servlet authentication method (none, simple, or jwt)."
+        "hive.metastore.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt", "oauth2"),
+        "HMS Catalog servlet authentication method (none, simple, jwt, or oauth2)."
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_ISSUER("metastore.catalog.servlet.auth.oauth2.issuer",
+        "hive.metastore.catalog.servlet.auth.oauth2.issuer", "",
+        "The issuer(iss)'s URI. This is required when you use metastore.catalog.servlet.auth=oauth2"
```
deniskuzZ: `The issuer URI of the IdP.`?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1873,8 +1873,55 @@ public enum ConfVars {
...
+        "to consider additional security requirements such as token revocations, use introspection."
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_AUDIENCE("metastore.catalog.servlet.auth.oauth2.audience",
+        "hive.metastore.catalog.servlet.auth.oauth2.audience", "",
+        "The acceptable name in the audience(aud) claim.  This is required when you use " +
+        "metastore.catalog.servlet.auth=oauth2"
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_CLIENT_ID("metastore.catalog.servlet.auth.oauth2.client.id",
+        "hive.metastore.catalog.servlet.auth.oauth2.client.id", "",
+        "The client ID of HMS as a resource server. This is required to use " +
```
deniskuzZ: `The public identifier of the HMS Catalog client` ?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1873,8 +1873,55 @@ public enum ConfVars {
...
+        "metastore.catalog.servlet.auth.oauth2.validation.method=introspection."
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_CLIENT_SECRET("metastore.catalog.servlet.auth.oauth2.client.secret",
+        "hive.metastore.catalog.servlet.auth.oauth2.client.secret", "",
+        "The client secret of HMS as a resource server. This is required to use " +
```
deniskuzZ: The OAuth2 client secret associated with the HMS’s registered client ID when acting as a resource server client

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: `This requires setting metastore.catalog.servlet.auth.oauth2.validation.method to introspection` ?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1873,8 +1873,55 @@ public enum ConfVars {
...
-        "hive.metastore.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt"),
-        "HMS Catalog servlet authentication method (none, simple, or jwt)."
+        "hive.metastore.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt", "oauth2"),
+        "HMS Catalog servlet authentication method (none, simple, jwt, or oauth2)."
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_ISSUER("metastore.catalog.servlet.auth.oauth2.issuer",
+        "hive.metastore.catalog.servlet.auth.oauth2.issuer", "",
+        "The issuer(iss)'s URI. This is required when you use metastore.catalog.servlet.auth=oauth2"
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_VALIDATION_METHOD("metastore.catalog.servlet.auth.oauth2.validation.method",
```
deniskuzZ, line 1889: should we define token type instead:

There are two types of access tokens, depending on how they are encoded:

**Identifier-based**: The token represents a random, hard-to-guess identifier associated with the authorization in the authorization server’s database.

**Self-contained**: The authorization is encoded within the token itself and is protected through cryptography to prevent tampering. JSON Web Token (JWT) is the common standard for this method.

For self-contained tokens, the authorization-related metadata can be directly parsed from the access token. **However, for identifier-based tokens, the authorization server’s token introspection functionality must be used to validate/retrieve the metadata**.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/auth/jwt/JWTValidator.java
deniskuzZ: cc @dengzhhu653, @henrib

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1873,8 +1873,55 @@ public enum ConfVars {
...
+        "hive.metastore.catalog.servlet.auth.oauth2.client.id", "",
+        "The client ID of HMS as a resource server. This is required to use " +
+        "metastore.catalog.servlet.auth.oauth2.validation.method=introspection."
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_CLIENT_SECRET("metastore.catalog.servlet.auth.oauth2.client.secret",
+        "hive.metastore.catalog.servlet.auth.oauth2.client.secret", "",
+        "The client secret of HMS as a resource server. This is required to use " +
+        "metastore.catalog.servlet.auth.oauth2.validation.method=introspection."
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_INTROSPECTION_CACHE_EXPIRY(
```
deniskuzZ, line 1905: is this a https://iceberg.apache.org/javadoc/1.9.1/org/apache/iceberg/rest/auth/OAuth2Properties.html#TOKEN_EXPIRES_IN_MS

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
okumin: The scope of OAuth 2 refers to the range of access from a client to protected resources. For example, let's say a web service (Client) wants to access the name, icon, and email (Protected Resources) of your (Resource Owner) Google account. The web service does not need to access your items in Google Photos, nor does it need to update Google's profiles. In this case, you can allow the web service to access only a limited set of resources.
image

The Iceberg client uses only "catalog" to give access to [all Iceberg-related endpoints](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml). We may protect HMS Thrift over HTTP with OAuth 2 and grant the "metastore" scope in the future. In that case, a user can create an access token that can use the HMS API but can not use Iceberg REST, or vice versa.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: Precisely, the format of a token is not tightly coupled with the validation method. I know two use cases.
(1) Some authorization servers can allow end users to revoke existing integrations. JWT can't express such a revocation. Resource servers can use the introspection endpoint to know the latest status.
(2) [Some authorization servers may not want to include all information in JWT](https://www.keycloak.org/docs/latest/server_admin/#_using_lightweight_access_token), not to expose PII or to reduce the token size. A token introspection response can contain more details than a JWT.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: Quickly checking, `OAuth2Properties.TOKEN_EXPIRES_IN_MS` is used for an Iceberg client to refresh an access token, meaning the process (2) below. `CATALOG_SERVLET_AUTH_OAUTH2_INTROSPECTION_CACHE_EXPIRY` is used to cache the response of (4).
image

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: This is an ID used for authentication, not from an HMS client, but rather from an HMS used against authorization servers. HMS is not only a server for an Iceberg client but also a client for an OAuth 2 authorization server.

standalone-metastore/metastore-rest-catalog/src/test/java/org/apache/iceberg/rest/extension/HiveRESTCatalogServerExtension.java
okumin: Fixed

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: Renamed

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: Aligned with the terms of RFC
https://datatracker.ietf.org/doc/html/rfc8414#section-2

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
deniskuzZ: do you think `catalog-read` and `catalog-write` scopes would be useful? This way we could restrict some of the clients from modifying the metadata?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: how about:
> The OAuth2 client identifier that HMS uses when registering as a resource server client with the Identity Provider (IdP) ?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: in our case HMS is both Client and a Resource Server, right?
Image Sep 25, 2025, 11_50_14 AM

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
```diff
@@ -1873,8 +1873,56 @@ public enum ConfVars {
             " positive value will be used as-is."
     ),
     CATALOG_SERVLET_AUTH("metastore.catalog.servlet.auth",
-        "hive.metastore.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt"),
-        "HMS Catalog servlet authentication method (none, simple, or jwt)."
+        "hive.metastore.catalog.servlet.auth", "jwt", new StringSetValidator("none", "simple", "jwt", "oauth2"),
+        "HMS Catalog servlet authentication method (none, simple, jwt, or oauth2)."
+    ),
+    CATALOG_SERVLET_AUTH_OAUTH2_ISSUER("metastore.catalog.servlet.auth.oauth2.issuer",
```
deniskuzZ, line 1879: issuer config is always paired with client-id & client-secret. should we group them together?

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: btw, don't we need to define these as well or that would be client's(caller) responsibility
https://datatracker.ietf.org/doc/html/rfc8414#section-2

- authorization_endpoint
redirects end-users to authenticate.

- token_endpoint – REQUIRED unless the AS supports only implicit.
to exchange codes for tokens.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: sorry @okumin, my bad. I checked number of other configs and they seem to follow the same naming convention

'fs.azure.account.oauth2.client.id': '${azure_client_id}',
'fs.azure.account.oauth2.client.secret': '${azure_client_secret}',

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: sorry @okumin, my bad. I checked number of other configs and they seem to follow the same naming convention

'fs.azure.account.oauth2.client.id': '${azure_client_id}',
'fs.azure.account.oauth2.client.secret': '${azure_client_secret}',

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: The entire authorization flow is decomposed into two steps: one to obtain an Access Token and another to access protected resources using the Access Token. I will skip explaining the first step because HMS is not involved at all. Anyway, a client exchanges some credentials for an Access Token with a limited scope and expiration. `authorization_endpoint` and `token_endpoint` are used there; That's why we don't need to configure them.
In the second step, HMS accepts and validates the Access Token. Token Introspection is a method to verify the latest state of Access Tokens through the network. `CATALOG_SERVLET_AUTH_OAUTH2_CLIENT_ID` and `CATALOG_SERVLET_AUTH_OAUTH2_CLIENT_SECRET` are used here so that only HMS can validate the access token and get the details.

image

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
okumin: It is an excellent question. It would be helpful. For example, it would allow you to give the read-only permission to a BI tool. I've not found the best practice for HMS and Iceberg users. I found some samples:

- [Apache Polaris](https://polaris.apache.org/in-dev/unreleased/configuring-polaris-for-production/)'s document often refers to `PRINCIPAL_ROLE:ALL` as a scope
- [Databricks Unity Catalog](https://docs.databricks.com/aws/en/external-access/iceberg) mentions `all-apis`
    - Most likely, [all-apis, sql, offline_access, openid, profile, email are allowed](https://docs.databricks.com/api/account/customappintegration/create)

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: Rephrased.

> The client ID|secret to authenticate HMS, as a resource server, to the introspection endpoint

This expression derives from [RFC](https://datatracker.ietf.org/doc/html/rfc7662#section-2.1).

> In this example, the protected resource uses a client identifier and
   client secret to authenticate itself to the introspection endpoint.
   The protected resource also sends a token type hint indicating that
   it is inquiring about an access token.

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: They are in their current location because they are only used when Token Introspection is employed.

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogFactory.java
deniskuzZ: that's probably needs more design thinking so we could handle it later

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
deniskuzZ: Thanks! Then I understood it right. 
But that would mean we need to add access token retrieval and refresh handling to the Hive RestCatalog OAuth2 client.

deniskuzZ (approved):
LGTM +1, thanks @okumin for adding OAuth2 support in RestCatalog

standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java
okumin: In the case of Iceberg REST, iceberg-core is responsible for it. In the case of Thrift over REST(Hive REST v2), as you say, IMetastoreClient needs to take the new responsibility, as you say. That's why I could not add `metastore.authentication=oauth2` in this pull request. I can add it if someone is willing to have it, though.

okumin:
I'm testing this in my local, and going to add how-to-setup to hive-site. I will merge this after that

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
henrib: It will be removed with Iceberg spec 2.0; at this point, it may be deprecated but should not be removed yet, should it ?

henrib (commented):
Should we remove the oauth/tokens endpoint yet ?

standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java
okumin: Yes, it should.
The token endpoint is not of a Resource Server but of an Authorization Server, i.e., typically Identity Provider such as Keycloak or Okta. The token endpoint is responsible for issuing a secure(e.g., cryptographically signed) Access Token.
Our endpoint is just a copied and pasted from iceberg-core, which is almost like an echo server(no authentication happens). It is possible to embed the Authorization Server roles in HMS; however, we would need to implement an RFC 6749, RFC 9068, or another RFC-compliant Authorization Server in that case. Otherwise, no user exists(who wants OAuth 2 that allows unauthenticated access?).
https://github.com/apache/hive/blob/96cf347ae4b449b6fb8118d5c588641ea1a12106/standalone-metastore/metastore-rest-catalog/src/main/java/org/apache/iceberg/rest/HMSCatalogAdapter.java#L243-L265

okumin:
@deniskuzZ @henrib Thanks for your reviews!

PR #6087, opened 2025-09-18
HIVE-29215: Fix owner info for view in authorizable events for alter …

saihemanth-cloudera:
…view query

Fixed owner info

Owner info incorrect in authorizable events

No

Remote cluster

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/view/create/AlterViewAsAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/view/create/AlterViewAsAnalyzer.java
@@ -67,14 +67,16 @@ public class AlterViewAsAnalyzer extends AbstractCreateViewAnalyzer {
     String expandedText = ctx.getTokenRewriteStream().toString(select.getTokenStartIndex(), select.getTokenStopIndex());
 
     AlterViewAsDesc desc = new AlterViewAsDesc(fqViewName, schema, originalText, expandedText);
-    validateCreateView(desc, analyzer);
+    Table oldView = validateCreateView(desc, analyzer);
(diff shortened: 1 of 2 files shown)


ql/src/java/org/apache/hadoop/hive/ql/ddl/view/create/AlterViewAsAnalyzer.java
```diff
@@ -67,14 +67,16 @@ public void analyzeInternal(ASTNode root) throws SemanticException {
...
-    validateCreateView(desc, analyzer);
+    Table oldView = validateCreateView(desc, analyzer);
+    oldView.setDbName(viewName.getDb());
+    oldView.setTableName(viewName.getTable());
+    oldView.setTableType(TableType.VIRTUAL_VIEW);
 
     rootTasks.add(TaskFactory.get(new DDLWork(getInputs(), getOutputs(), desc)));
-    DDLUtils.addDbAndTableToOutputs(getDatabase(viewName.getDb()), viewName, TableType.VIRTUAL_VIEW, false,
-        null, outputs);
+    DDLUtils.addDbAndTableToOutputs(getDatabase(viewName.getDb()), oldView, outputs);
```
dengzhhu653, line 76: I don't get it, before this fix `DDLUtils.addDbAndTableToOutputs` will create a dummy table based on `viewName`, it has the same `db` and `table` with the `oldView`, why it doesn't work before?

saihemanth-cloudera: The viewName has type TableName, which has dbname and table name. So the HivePrivilegeObject will be constructed with the current user as the owner of the view. This is incorrect behavior.

PR #6088, opened 2025-09-19
HIVE-29177: Implement default Catalog selection

zhangbutao:

**TODO: Add more test and cover more api.**


Neer393:
@zhangbutao I think we need a default database created whenever a user creates new catalog so that ideally when a user switches catalog, we point to the default database within it
What do you think ?

zhangbutao:
> @zhangbutao I think we need a default database created whenever a user creates new catalog so that ideally when a user switches catalog, we point to the default database within it What do you think ?

@Neer393 Currently(without this PR), when a new catalog is created, a default database will also be created. 

With this PR, when a user switches catalogs using `SET CATALOG catName`, it should not point to the `default database` by default. The user should explicitly specify a database through `USE dbName` or `USE catName.dbName xxx` to maintain maximum compatibility with previous usage patterns.
Please check **catalog_database.q** https://github.com/apache/hive/pull/6088/files#diff-0a99f71cb21eb68f9efe36ff59f2005f696ef9ea6b660aef5cc184152018f79b

deniskuzZ:
is the PR ready for review? 
cc @difin - once merged, we should probably update the RestCatalog client to use selected catalog configs

zhangbutao:
> is the PR ready for review? cc @difin - once merged, we should probably update the RestCatalog client to use selected catalog configs

This PR involves many database interfaces. Some of the interfaces I've marked with TODO tags and need a double check to avoid potential issues that could be hard to troubleshoot in future development.  But I welcome reviews of this PR during the period when I'm optimizing the code.

Additionally, this PR only addresses the interfaces for switching catalogs and databases, and does not cover interfaces related to `cat.db.table`. Therefore, even if the catalog and database are switched, the current client will still query tables from the default `hive `catalog. The code related to `cat.db.tbl` (part of the work in HIVE-29177) will also involve many table-related interfaces, which presents certain refactoring challenges.

If I have enough time in the near future, I might write down some thoughts on federated catalogs, including typical use cases and the code refactoring challenges involved. I hope this will attract more developers to join the discussion.

zhangbutao:
All the `TODO catalog` lable can be tracked by relative tickts, see https://github.com/apache/hive/pull/6088/commits/d2bd35394d82ef5dc778e3c2c193288e29e63187

We can safely address this `TODO catalog ` in the following tasks.

So I think this PR can be reviewed now. cc @deniskuzZ @difin

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
```diff
@@ -468,7 +468,7 @@ public enum ErrorMsg {
   RESOURCE_PLAN_ALREADY_EXISTS(10417, "Resource plan {0} already exists", true),
   RESOURCE_PLAN_NOT_EXISTS(10418, "Resource plan {0} does not exist", true),
   INCOMPATIBLE_STRUCT(10419, "Incompatible structs.", true),
-  OBJECTNAME_CONTAINS_DOT(10420, "Table or database name may not contain dot(.) character", true),
+  OBJECTNAME_CONTAINS_DOT(10420, "Catalog or table or database name may not contain dot(.) character", true),
```
deniskuzZ: maybe use comma: `Catalog, database or table names `

ql/src/java/org/apache/hadoop/hive/ql/ddl/catalog/use/SwitchCatalogDesc.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * DDL task description for SET CATALOG commands.
+ */
+@Explain(displayName = "Switch Catalog", explainLevels = { Explain.Level.USER, Explain.Level.DEFAULT, Explain.Level.EXTENDED })
+public class SwitchCatalogDesc implements DDLDesc, Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private final String catalogName;
+
+    public SwitchCatalogDesc(String databaseName) {
```
deniskuzZ: catalogName ?

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseDesc.java
```diff
@@ -45,13 +46,14 @@ public class CreateDatabaseDesc implements DDLDesc, Serializable {
   private final String remoteDbName;
   private final Map<String, String> dbProperties;
 
-  public CreateDatabaseDesc(String databaseName, String comment, String locationUri, String managedLocationUri,
+  public CreateDatabaseDesc(String catalogName, String databaseName, String comment, String locationUri, String managedLocationUri,
       boolean ifNotExists, Map<String, String> dbProperties) {
-    this(databaseName, comment, locationUri, managedLocationUri, ifNotExists, dbProperties, "NATIVE", null, null);
+    this(catalogName, databaseName, comment, locationUri, managedLocationUri, ifNotExists, dbProperties, "NATIVE", null, null);
```
deniskuzZ, line 51: for the external catalogs `dbtype` should be "REMOTE", right? It seems that we always hardcode "NATIVE" and below condition never returns true
````
 if (dbtype != null && dbtype.equalsIgnoreCase("REMOTE")) {
````

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java [hunk 4]
```diff
@@ -51,12 +53,17 @@ public DropDatabaseAnalyzer(QueryState queryState) throws SemanticException {
...
     boolean cascade = root.getFirstChildWithType(HiveParser.TOK_CASCADE) != null;
     boolean isSoftDelete = HiveConf.getBoolVar(conf, HiveConf.ConfVars.HIVE_ACID_LOCKLESS_READS_ENABLED);
 
-    Database database = getDatabase(databaseName, !ifExists);
+    String catalogName = catDbNamePair.getLeft();
+    if (catalogName != null && getCatalog(catalogName) == null) {
+      throw new SemanticException(ErrorMsg.CATALOG_NOT_EXISTS, catalogName);
+    }
+    String databaseName = catDbNamePair.getRight();
+    Database database = getDatabase(catDbNamePair.getLeft(), catDbNamePair.getRight(), !ifExists);
```
deniskuzZ: why not use local vars? 
````
getDatabase(catalogName, databaseName, !ifExists) 
````

File: ql/src/java/org/apache/hadoop/hive/ql/ddl/database/unlock/UnlockDatabaseAnalyzer.java [hunk 5]
```diff
@@ -41,14 +43,19 @@ public UnlockDatabaseAnalyzer(QueryState queryState) throws SemanticException {
...
+      throw new SemanticException(ErrorMsg.CATALOG_NOT_EXISTS, catalogName);
+    }
+    String databaseName = catDbNamePair.getRight();
 
-    inputs.add(new ReadEntity(getDatabase(databaseName)));
+    inputs.add(new ReadEntity(getDatabase(catalogName, databaseName, true)));
     // Unlock database operation is to release the lock explicitly, the operation itself don't need to be locked.
     // Set the WriteEntity as WriteType: DDL_NO_LOCK here, otherwise it will conflict with Hive's transaction.
-    outputs.add(new WriteEntity(getDatabase(databaseName), WriteType.DDL_NO_LOCK));
+    outputs.add(new WriteEntity(getDatabase(catalogName, databaseName, true), WriteType.DDL_NO_LOCK));
```
deniskuzZ, line 57: HIVE_LOCKS table doesn't have catalog column. 
Maybe some operation like Unlock/Lock DB, ShowDbLocks should only be supported by NATIVE DBs?

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/lock/show/ShowLocksOperation.java
```diff
@@ -160,6 +160,8 @@ private ShowLocksResponse getLocksForNewFormat(HiveLockManager lockMgr) throws H
       throw new HiveException("New lock format only supported with db lock manager.");
     }
 
+    // TODO catalog. Need to add catalog into ShowLocksRequest. But ShowLocksRequest doesn't have catalog field.
```
deniskuzZ, line 164: see https://github.com/apache/hive/pull/6088/files#r2454590955

ql/src/java/org/apache/hadoop/hive/ql/exec/repl/ReplLoadTask.java
```diff
@@ -448,8 +449,9 @@ private void setDbReadOnly() {
     Map<String, String> props = new HashMap<>();
     props.put(READONLY, Boolean.TRUE.toString());
 
+    // TODO catalog. Need to double check the actual catalog here. Depend on HIVE-29278
```
deniskuzZ, line 452: don't think we can replicate not NATIVE DBs

ql/src/java/org/apache/hadoop/hive/ql/exec/repl/incremental/IncrementalLoadTasksBuilder.java
```diff
@@ -264,8 +265,9 @@ private Task<?> dbUpdateReplStateTask(String dbName, String replState, Task<?> p
     HashMap<String, String> mapProp = new HashMap<>();
     mapProp.put(ReplicationSpec.KEY.CURR_STATE_ID_SOURCE.toString(), replState);
 
-    AlterDatabaseSetPropertiesDesc alterDbDesc = new AlterDatabaseSetPropertiesDesc(dbName, mapProp,
-        new ReplicationSpec(replState, replState));
+    // TODO catalog. Need to double check the actual catalog here. Depend on HIVE-29278
+    AlterDatabaseSetPropertiesDesc alterDbDesc = new AlterDatabaseSetPropertiesDesc(HiveUtils.getCurrentCatalogOrDefault(conf),
```
deniskuzZ, line 269: @harshal-16 could you please validate if this is ok?

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/HiveTxnManagerImpl.java
```diff
@@ -162,7 +167,8 @@ public int lockDatabase(Hive hiveDB, LockDatabaseDesc lockDb) throws HiveExcepti
             String.valueOf(System.currentTimeMillis()),
             "EXPLICIT", lockDb.getQueryStr(), conf);
 
-    HiveLock lck = lockMgr.lock(new HiveLockObject(dbObj.getName(), lockData), mode, true);
+    // Using the catalogName@databaseName format to uniquely identify a database.
+    HiveLock lck = lockMgr.lock(new HiveLockObject(catName + "@" +dbObj.getName(), lockData), mode, true);
```
deniskuzZ: oh, nice hack! however, maybe we should use "." ? 
I think it might actually work. i am not sure if `show locks` would handle this change, might need some tweaks: 
````
show locks db1
show locks cat1.db1
````
note, might be problematic if CU has multiple HMS versions on a cluster 

nit: please add space `+ dbObj.getName()`

File: ql/src/java/org/apache/hadoop/hive/ql/lockmgr/HiveTxnManagerImpl.java [hunk 10]
```diff
@@ -173,13 +179,15 @@ public int lockDatabase(Hive hiveDB, LockDatabaseDesc lockDb) throws HiveExcepti
...
+            HiveUtils.getCurrentCatalogOrDefault(conf));
     String dbName = unlockDb.getDatabaseName();
 
-    Database dbObj = hiveDB.getDatabase(dbName);
+    Database dbObj = hiveDB.getDatabase(catName, dbName);
     if (dbObj == null) {
       throw new HiveException("Database " + dbName + " does not exist ");
     }
-    HiveLockObject obj = new HiveLockObject(dbObj.getName(), null);
+    HiveLockObject obj = new HiveLockObject(catName + "@" +dbObj.getName(), null);
```
deniskuzZ: space

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -721,9 +726,10 @@ public void dropDatabase(String name, boolean deleteData, boolean ignoreUnknownD
    * @throws HiveException
    * @throws NoSuchObjectException
    */
+  @Deprecated
```
deniskuzZ, line 732: please add to java doc "@deprecated since 4.1.0, will be removed in 5.0.0"

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -1297,7 +1303,12 @@ public void renamePartition(Table tbl, Map<String, String> oldPartSpec, Partitio
     }
   }
 
-  // TODO: this whole path won't work with catalogs
+  /**
+   * When you call this method, you need to ensure that the catalog has been set in the db object.
```
deniskuzZ, line 1310: why not have fallback similar to https://github.com/apache/hive/pull/6088/files#diff-9d6bc2f1a2972d844b3107f29bc1731af7adf6f4c922e08904345c5ae88cd727R1443

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -1914,6 +1942,16 @@ public Table apply(org.apache.hadoop.hive.metastore.api.Table table) {
     }
   }
 
+  public List<Table> getTableObjects(String catName, String dbName, String pattern, TableType tableType) throws HiveException {
```
deniskuzZ, line 1946: @dengzhhu653 is it ok or we should use some request object like Database?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -1997,6 +2035,52 @@ public List<String> getTablesByType(String dbName, String pattern, TableType typ
...
+   * Returns all existing tables of a type (VIRTUAL_VIEW|EXTERNAL_TABLE|MANAGED_TABLE) from the specified
+   * database which match the given pattern. The matching occurs as per Java regular expressions.
+   * @param catName catalog name to find the tables in. if null, uses the current catalog in this session.
+   * @param dbName Database name to find the tables in. if null, uses the current database in this session.
+   * @param pattern A pattern to match for the table names.If null, returns all names from this DB.
+   * @param type The type of tables to return. VIRTUAL_VIEWS for views. If null, returns all tables and views.
+   * @return list of table names that match the pattern.
+   * @throws HiveException
+   */
+  public List<String> getTablesByType(String catName, String dbName, String pattern, TableType type)
```
deniskuzZ, line 2026: same as https://github.com/apache/hive/pull/6088/files#r2454680968
can we reuse it in getTablesByType(String dbName, String pattern, TableType type) ?

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
```diff
@@ -6463,6 +6564,18 @@ public List<Function> getFunctionsInDb(String dbName, String pattern) throws Hiv
     }
   }
 
+  public List<Function> getFunctionsInDb(String catName, String dbName, String pattern) throws HiveException {
```
deniskuzZ, line 6546: same as https://github.com/apache/hive/pull/6088/files#r2454681994

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
```diff
@@ -543,4 +546,10 @@ public static String getLowerCaseTableName(String refName) {
     }
     return refName.toLowerCase();
   }
+
+  public static String getCurrentCatalogOrDefault(Configuration conf) {
+    return SessionState.get() != null ?
```
deniskuzZ: you could try Optional.ofNullable

ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java
```diff
@@ -1198,6 +1198,10 @@ public String getCatalogName() {
     return this.tTable.getCatName();
   }
 
+  public void setCatalogName(String catalogName) {
```
deniskuzZ, line 1201: do we need this setter on a Table object?

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
```diff
@@ -1926,6 +1960,19 @@ protected Database getDatabase(String dbName, boolean throwException) throws Sem
     return database;
   }
 
+  protected Database getDatabase(String catalogName, String dbName, boolean throwException) throws SemanticException {
```
deniskuzZ, line 1927: try reuse code. you could call new method within the old one without catalogName

ql/src/java/org/apache/hadoop/hive/ql/parse/repl/load/message/CreateDatabaseHandler.java
```diff
@@ -56,22 +56,23 @@ public List<Task<?>> handle(Context context)
     Database db = metaData.getDatabase();
     String destinationDBName =
         context.dbName == null ? db.getName() : context.dbName;
+    String destinationCatalogName = db.getCatalogName(); // TODO catalog. Need to double check the catalog here. Depend on HIVE-29278
```
deniskuzZ: cc @harshal-16

ql/src/java/org/apache/hadoop/hive/ql/processors/HiveCommand.java
```diff
@@ -86,6 +86,8 @@ public static HiveCommand find(String[] command, boolean findOnlyForTesting) {
       } else if (command.length > 1 && "show".equalsIgnoreCase(command[0]) &&
           "processlist".equalsIgnoreCase(command[1])) {
         return PROCESSLIST;
+      } else if(command.length > 1 && "set".equalsIgnoreCase(command[0]) && "catalog".equalsIgnoreCase(command[1])) {
```
deniskuzZ: space

ql/src/test/queries/clientnegative/lockneg_try_lock_cat_db_in_use.q
```diff
@@ -0,0 +1,9 @@
+set hive.lock.numretries=0;
+set hive.support.concurrency=true;
```
deniskuzZ, line 2: that is only needed for ACID tables when DBTxnManager is used

ql/src/test/queries/clientpositive/catalog.q
```diff
@@ -45,3 +45,6 @@ SHOW CATALOGS LIKE 'test__';
 -- ALTER LOCATION
 ALTER CATALOG test_cat SET LOCATION '/tmp/test_cat_new';
 DESC CATALOG EXTENDED test_cat;
+
+-- DROP catalog at the end
+DROP CATALOG  test_cat;
```
deniskuzZ: redundant space

deniskuzZ:
thanks @zhangbutao! in general LGTM, added some comments

ql/src/java/org/apache/hadoop/hive/ql/exec/repl/incremental/IncrementalLoadTasksBuilder.java
harshal-16: Just checked the details in the main JIRA: https://issues.apache.org/jira/browse/HIVE-28879
As ACID tables can only reside in Hive (default), other catalogs don't make sense in case of replication here
Also, during dump operation itself, it filters only default catalog here: https://github.com/apache/hive/blob/d2bd35394d82ef5dc778e3c2c193288e29e63187/ql/src/java/org/apache/hadoop/hive/ql/exec/repl/ReplDumpTask.java#L923

deniskuzZ: thanks for checking, @harshal-16 ! what if we register multiple external HMS catalogs?

harshal-16: Oh, then I think we will need to update the incremental dump filter also to make it consistent with the load.

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseDesc.java
zhangbutao: > for the external catalogs dbtype should be "REMOTE", right?

I don't think so. `REMOTE ` dbtyp was introduced by HIVE-24396 data connector. If we implement all tasks about federated catalog, **we can deprecate data connector**. Because the data connector can only map external databases to internal databases by manually creating remote databases one by one to query external data sources,  like this:
`CREATE REMOTE DATABASE db_map_mysql USING mysqlconnector with DBPROPERTIES("connector.remoteDbName"="testmysqldb");`

whereas the federated catalog can retrieve all external databases at once without the need to manually create remote databases,  once we create a catalog.


> It seems that we always hardcode "NATIVE" and below condition never returns true

If you create a remote database with a specified dataconnector, this condition will return true.

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/unlock/UnlockDatabaseAnalyzer.java
zhangbutao: > HIVE_LOCKS table doesn't have catalog column.

Currently, this only retrieves a specific database under a particular catalog and does not involve any lock/unlock operations on the catalog.  
However, I believe future development should consider implementing lock/unlock operations for catalogs, though I haven't yet identified the specific use cases. It might be necessary during cross-catalog federated queries.

> Maybe some operation like Unlock/Lock DB, ShowDbLocks should only be supported by NATIVE DBs?

To be honest, I haven't fully thought this through. Locking/unlocking databases in external catalogs (such as a MySQL JDBC catalog) doesn't seem very meaningful, as these lock operations won't take effect on the MySQL source side. However, I believe allowing locks on databases in external catalogs in Hive might have some significance, particularly in scenarios involving cross-catalog and cross-database operations, to ensure task transactional consistency. Therefore, let's further explore this issue when we work on cross-catalog queries. An already created ticket, HIVE-29242, can be used to track this matter.

ql/src/java/org/apache/hadoop/hive/ql/ddl/table/lock/show/ShowLocksOperation.java
zhangbutao: Replied in https://github.com/apache/hive/pull/6088#discussion_r2465716095

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseDesc.java
deniskuzZ: should we introduce a flag like native/non-native similar to tables?

ql/src/java/org/apache/hadoop/hive/ql/exec/repl/ReplLoadTask.java
zhangbutao: You are right. Some task can only do in NATIVE DBs. Like this ReploadTask.

This TODO is added to verify if the correct native catalog name is obtained. BTW, besides the default `hive  ` native catalog, we still need to support other multiple native catalogs.

As for external catalogs, we need to skip those tasks that are exclusive to native catalogs. However, the specific types of external catalogs (e.g., JDBC, REST) have not yet been defined, and this part of the work can be refined and completed subsequently.  HIVE-29243 can be used to track this matter.

ql/src/java/org/apache/hadoop/hive/ql/exec/repl/incremental/IncrementalLoadTasksBuilder.java
zhangbutao: HIVE-29278  is used to track this matter. I think we can make the rep work with multiple native hms catalog.

dengzhhu653:
Sorry for the late reply.
Took a quick overview of the changes, looks we introduce a new syntax for the catalog, as well as the future items: 
>The code related to cat.db.tbl (part of the work in [HIVE-29177](https://issues.apache.org/jira/browse/HIVE-29177)) will also involve many table-related interfaces, which presents certain refactoring challenges.

> let's further explore this issue when we work on cross-catalog queries. An already created ticket, [HIVE-29242](https://issues.apache.org/jira/browse/HIVE-29242), can be used to track this matter.

It seems to me this is a significant/big change, and the user should follow a new way of catalog.db.table to request the table as the Trino/Presto does today.

I think we can have a more simple idea, let's push the catalog awareness down into the Metastore client. For example, 
SET CATALOG testcat;  -> change to a metaconf setting: set metaconf:metastore.catalog.default = testcat, default hive, this configuration will be propagated to HMS, every HMS API request will be appended with the `testcat`, both client and server can handle it.

For cross-catalog queries, we can also do it in a similar way, we can introduce a new catalog awareness lawyer between session client and the thrift client, as SessionHiveMetaStoreClient -> CatalogAwarenessMetastoreClient -> ThriftHiveMetaStoreClient. The CatalogAwarenessMetastoreClient can acknowledge of the catalog through configuration or properties stored in HMS(HIVE-27186).  Something as `hive.matastore.catalog.mapping.pattern`=`iceberg_catalog`:`ice_db1*,hive_db1,hive_db2.ice_tab*`;`hive`:`default,acid_db*`, the CatalogAwarenessMetastoreClient even can point to a third-party catalog partner.

By this way, we don't need to introduce the change on the Hive SQL syntax and test it throughly, and other engines(such as Impala) can also benefit from this way to add the support on cross-catalog queries.

dengzhhu653:
The biggest flaw is we cannot have two same db name/table name among catalogs, this might be a question in the future.
For cross-catalog queries e.g, db1.name1 join db1.name1 in this case, CatalogAwarenessMetastoreClient mightn't be able to figure out whether the all db1.name1 tables belong to the same catalog or not.
I don't have an elegant solution yet, maybe we can add a limit that the db is unique across the catalogs, which as we do in the Hive catalog.

deniskuzZ:
> The biggest flaw is we cannot have two same db name/table name among catalogs

I don't follow this: `FederatedMetastoreClient` should be able to resolve proper Catalog to connect to in order to fetch/update db/table metadata. 

PS: would be great if we could plan this feature for the next Hive/HMS release to support CU interest. 
https://www.linkedin.com/posts/mimounedjouallah_apache-hive-iceberg-rest-catalog-api-backed-activity-7398883260350529538-Go_T

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/00a75f9fe1d13cc9e49a00a3cd2b598dadef1f0b

common/src/java/org/apache/hadoop/hive/ql/ErrorMsg.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/00a75f9fe1d13cc9e49a00a3cd2b598dadef1f0b

ql/src/java/org/apache/hadoop/hive/ql/ddl/catalog/use/SwitchCatalogDesc.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/00a75f9fe1d13cc9e49a00a3cd2b598dadef1f0b

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/create/CreateDatabaseDesc.java
zhangbutao: > should we introduce a flag like native/non-native similar to tables?

This can be done when we implement specific catalogs (JDBC catalog & Iceberg catalog, etc.).

ql/src/java/org/apache/hadoop/hive/ql/ddl/database/drop/DropDatabaseAnalyzer.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/00a75f9fe1d13cc9e49a00a3cd2b598dadef1f0b

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/HiveTxnManagerImpl.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/88a24cd836431409358a643dfa570de2141aba5a

I suddenly realized that this hack isn't a good approach. I referenced the method for obtaining lock objects for tables and created a similar method for databases, which might be better.

Additionally, I noticed that the SHOW LOCKS statement could only display table-related locks but not database-related locks.
Here's the test query, referenced from lockneg_try_lock_db_in_use.q:
```
set hive.lock.numretries=0;
create database lockneg2;

lock database lockneg2 shared;
show locks;
```

After this code optimization https://github.com/apache/hive/pull/6088/commits/88a24cd836431409358a643dfa570de2141aba5a, the lock information for lockneg2 can now be seen:
```
0: jdbc:hive2://127.0.0.1:10000/default> show locks;
INFO  : Compiling command(queryId=hive_20251126170301_32696de3-7d1e-490f-a3b4-d956be706897): show locks
+----------------+---------+
|    tab_name    |  mode   |
+----------------+---------+
| hive@lockneg2  | SHARED  |
+----------------+---------+
```

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/HiveTxnManagerImpl.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/88a24cd836431409358a643dfa570de2141aba5a

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/00a75f9fe1d13cc9e49a00a3cd2b598dadef1f0b

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
zhangbutao: https://github.com/apache/hive/pull/6088/files#diff-0b11c3fb6f813b496b291de76cbdf98dff3e7f47d1ba4d8bb68c4dedb858da42R516

In this method, the call to` BaseMetaStoreClient::alterDatabase(String name, Database db)` will retrieve the specific catalog name based on the current session or the db object. 
However, we need to ensure that the db object preferably contains the catalog name, because references in the format "cat.db" might be unrelated to the current session's catalog.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
zhangbutao: I just follow other method pattern.  There are still many methods with numerous parameters like this. Perhaps we can incrementally optimize them later. However, these methods might be used by other components like Spark, so we may need to proceed with caution.

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/00a75f9fe1d13cc9e49a00a3cd2b598dadef1f0b

ql/src/java/org/apache/hadoop/hive/ql/metadata/HiveUtils.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/00a75f9fe1d13cc9e49a00a3cd2b598dadef1f0b

ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java
zhangbutao: Yes. We need. We may create a table like this  create table catName.dbName.tblName(id int);

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/cca5bdcfad23fa3653c3c51129c48dc21cce459d

ql/src/test/queries/clientpositive/catalog.q
zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/00a75f9fe1d13cc9e49a00a3cd2b598dadef1f0b

ql/src/test/queries/clientnegative/lockneg_try_lock_cat_db_in_use.q
zhangbutao: lock/unlock need this.

zhangbutao:
@dengzhhu653 Thanks for the suggesstion!

> It seems to me this is a significant/big change, and the user should follow a new way of catalog.db.table to request the table as the Trino/Presto does today.

That doesn't seem quite right. My goal is that even if users don't use this feature, they can still use the classic Hive approach, such as `select * from table` or `select * from db.table`. Users don't need to worry about catalogs because their current session has a default catalog. Of course, users can also specify a catalog name as a prefix to query a table, for example, `select * from hive.db.tbl`. 

Additionally, if users want to perform federated queries across multiple data sources in Hive, similar to Trino, they would need to create multiple catalogs and execute similar queries. This federated query capability is a new feature and does not pose compatibility issues for users.
```
SELECT 
    u.user_name,
    a.action_time
FROM mysql.test.users u
JOIN hive.default.user_actions a 
    ON u.user_id = a.user_id;
```


> I think we can have a more simple idea, let's push the catalog awareness down into the Metastore client. For example,
> SET CATALOG testcat; -> change to a metaconf setting: set metaconf:metastore.catalog.default = testcat, default hive, this configuration will be propagated to HMS, every HMS API request will be appended with the testcat, both client and server can handle it.

Although this allows switching catalogs, the current client can only access the databases and tables of one catalog at a time.



> For cross-catalog queries, we can also do it in a similar way, we can introduce a new catalog awareness lawyer between session client and the thrift client, as SessionHiveMetaStoreClient -> CatalogAwarenessMetastoreClient -> ThriftHiveMetaStoreClient. The CatalogAwarenessMetastoreClient can acknowledge of the catalog through configuration or properties stored in HMS([HIVE-27186](https://issues.apache.org/jira/browse/HIVE-27186)). Something as hive.matastore.catalog.mapping.pattern=iceberg_catalog:ice_db1*,hive_db1,hive_db2.ice_tab*;hive:default,acid_db*, the CatalogAwarenessMetastoreClient even can point to a third-party catalog partner.

This method allows accessing databases and tables across multiple catalogs, but it is quite inelegant. Moreover, as you mentioned, it cannot support having the same database names across different catalogs. This solution is somewhat similar to https://github.com/ExpediaGroup/waggle-dance. Additionally, it requires users to constantly modify parameters to add or remove database names in `hive.matastore.catalog.mapping.pattern`, which seems very cumbersome...


Finally, I want to reference the multi/federated catalog capabilities of Trino and StarRocks/Doris. I believe their design approaches are quite reasonable. I'm considering implementing this capability with the principle of minimizing code modifications. Considerable preliminary work has already been done on the HMS side with HIVE-18685. The subsequent work mainly involves connecting all catalog capabilities from the HS2 side to the HMS side. I've been thinking about writing down the related design considerations, but haven't had enough time yet... I will provide more thoughts on this later.

zhangbutao:
@deniskuzZ Thank you very much for your review. I've been quite busy with various matters recently, which has caused some delay in addressing these comments. My apologies.

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/HiveTxnManagerImpl.java
zhangbutao: Upon further inspection here, it was found that HIVE-2093 introduced the semantics for SHOW LOCKS DATABASE, specifically designed to retrieve the lock objects for a particular database. Therefore, I still need to spend some time confirming this.

deniskuzZ:
> @deniskuzZ Thank you very much for your review. I've been quite busy with various matters recently, which has caused some delay in addressing these comments. My apologies.

nw at all. thank you @zhangbutao for driving this feature. I think it would be a game-changer

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/HiveTxnManagerImpl.java
zhangbutao: There are some issues with the current database locks; I've just submitted another PR to fix these problems.  Please see https://github.com/apache/hive/pull/6222

zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/f066b1ea51f109982d0a7e960b531d44a912ee69

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/HiveTxnManagerImpl.java
```diff
@@ -38,8 +39,10 @@
 import org.apache.hadoop.hive.ql.lockmgr.HiveLockObject.HiveLockObjectData;
 import org.apache.hadoop.hive.ql.metadata.Hive;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
+import org.apache.hadoop.hive.ql.metadata.HiveUtils;
 import org.apache.hadoop.hive.ql.metadata.Partition;
 import org.apache.hadoop.hive.ql.metadata.Table;
+import org.stringtemplate.v4.ST;
```
deniskuzZ: redundant import

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
deniskuzZ: could we squash MetaException & TException handling

parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g
```diff
@@ -379,10 +379,13 @@ TOK_CATALOGLOCATION;
 TOK_CATALOGCOMMENT;
 TOK_ALTERCATALOG_LOCATION;
 TOK_ALTERCATALOG_PROPERTIES;
+TOK_SWITCHCATALOG;
 TOK_DESCDATABASE;
 TOK_DATABASELOCATION;
 TOK_DATABASE_MANAGEDLOCATION;
 TOK_PROPLIST;
+TOK_DBPROPLIST;
```
deniskuzZ: where is that used?

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DummyTxnManager.java
```diff
@@ -380,7 +389,7 @@ private HiveLockMode getWriteEntityLockMode (WriteEntity we) {
     }
   }
 
-  private List<HiveLockObj> getLockObjects(QueryPlan plan, Database db,
+  private List<HiveLockObj> getLockObjects(QueryPlan plan, Catalog catalog, Database db,
```
deniskuzZ: could we make method generic:
````
private <T> List<HiveLockObj> getLockObjects(QueryPlan plan, T target, HiveLockMode mode) 
  throws LockException {
List<HiveLockObj> locks = new LinkedList<HiveLockObj>();
    
switch (target) {
  case Catalog catalog -> {
    locks.add(new HiveLockObj(new HiveLockObject(catalog.getName(), lockData), mode));
  }
  case Database db -> {
    String catName = Objects.requireNonNullElse(db.getCatalogName(), 
        HiveUtils.getCurrentCatalogOrDefault(conf));
    locks.add(new HiveLockObj(new HiveLockObject(catName, lockData), mode));
    db.setCatalogName(catName);
    locks.add(new HiveLockObj(new HiveLockObject(db, lockData), mode));
  }
...
````

File: ql/src/test/queries/clientpositive/catalog_database.q [hunk 26]
```diff
@@ -0,0 +1,69 @@
...
+
+-- Check databases in catalog 'testcat',
+-- The list of databases in the catalog 'testcat' should contain default and testdb_1 and testdb_2.
+SHOW DATABASES;
+
+-- Switch database by catalog.db pattern
+USE testcat.testdb_1;
+
+-- Drop database by catalog.db pattern
+DROP DATABASE testcat.testdb_1;
```
deniskuzZ, line 43: Do we have tests ensuring that dropping the current catalog or the Hive default catalog is not allowed?

ql/src/test/results/clientpositive/llap/lock1.q.out
```diff
@@ -98,26 +98,26 @@ PREHOOK: query: SHOW LOCKS
...
 #### A masked pattern was here ####
 LOCK_MODE:EXPLICIT	 
 LOCK_QUERYSTRING:LOCK TABLE tstsrc_n1 SHARED	 
 PREHOOK: query: SHOW LOCKS tstsrc_n1
 PREHOOK: type: SHOWLOCKS
 POSTHOOK: query: SHOW LOCKS tstsrc_n1
 POSTHOOK: type: SHOWLOCKS
-default@tstsrc_n1	SHARED
-default@tstsrc_n1	SHARED
+hive@default@tstsrc_n1	SHARED
```
deniskuzZ: No need for catalog in show tables or show locks;

ql/src/java/org/apache/hadoop/hive/ql/parse/repl/load/message/CreateDatabaseHandler.java
```diff
@@ -55,22 +55,23 @@ public List<Task<?>> handle(Context context)
...
```
deniskuzZ: could we shorten `destinationCatalogName` -> `destCatalogName`. same for destDBName

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
```diff
@@ -421,6 +421,33 @@ public static String charSetString(String charSetName, String charSetString)
     }
   }
 
+  /**
+   *
+   * @param dbNameNode A root node that contains database fields
+   * @return Return a Pair object which includes catalogName and dbName
+   * @throws SemanticException
+   */
+  public static Pair<String, String> getCatDbNamePair(ASTNode dbNameNode) throws SemanticException {
```
deniskuzZ: it doesn't belong in SA, maybe move to util class

deniskuzZ:
@zhangbutao, I've added few more comments , otherwise LGTM.

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/HiveTxnManagerImpl.java
zhangbutao: Removed.

parser/src/java/org/apache/hadoop/hive/ql/parse/HiveParser.g
zhangbutao: Removed.

ql/src/java/org/apache/hadoop/hive/ql/lockmgr/DummyTxnManager.java
zhangbutao: Fixed! Thx!

ql/src/test/queries/clientpositive/catalog_database.q
zhangbutao: Added qtest ql/src/test/results/clientnegative/drop_default_catalog.q.out

ql/src/java/org/apache/hadoop/hive/ql/parse/repl/load/message/CreateDatabaseHandler.java
zhangbutao: Fixed.

ql/src/java/org/apache/hadoop/hive/ql/parse/BaseSemanticAnalyzer.java
zhangbutao: I'm not entirely sure if this method will be used in SA in the future. Given that it's currently only used in DB for DDL operations, I'll move this code to DDLUtils.

ql/src/test/results/clientpositive/llap/lock1.q.out
zhangbutao: I think it's necessary to display catalog information in the "show locks" command. Since we will have multiple native Hive catalogs in the future, each catalog will have its own locks. Displaying catalog information will help users determine which catalog a lock belongs to.

deniskuzZ: show locks should display locks only for the current catalog, similar to show tables for a specific database

zhangbutao: That's make sense. Fixed in https://github.com/apache/hive/pull/6088/commits/df94b419ed5c6e1017f3fd7eb53995a23848fbab

ql/src/test/queries/clientpositive/catalog_database.q
deniskuzZ: thanks!

ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java
deniskuzZ: @zhangbutao, looks like this comment slipped, but it's minor

zhangbutao: Fixed in https://github.com/apache/hive/pull/6088/commits/97e8af0dadfd36703499e72d9ecbfd3c9f84da0f 
Thanks!

zhangbutao:
> +1, Thanks, @zhangbutao. If anything was missed, we can fix it in a follow-up

Sure. We still have lots of works to do. :)

zhangbutao:
This PR currently introduces only a minor incompatibility issue: when using ZooKeeperHiveLockManager, the znode path for locks stored in ZooKeeper changes from `/database` to `/catalog/database`. This will cause older versions of Hive to be unable to properly read the lock information stored in ZooKeeper. However, to my knowledge, we typically use DbLockManager for ACID operations rather than ZooKeeperHiveLockManager, so this incompatibility risk is negligible.

Aside from this, the PR does not break any other default functionalities. I will proceed to merge this PR shortly and continue with the subsequent related development work.

zhangbutao:
Merged. Thanks @deniskuzZ @dengzhhu653 @harshal-16 @Neer393

parser/src/java/org/apache/hadoop/hive/ql/parse/FromClauseParser.g
```diff
@@ -231,6 +231,17 @@ uniqueJoinTableSource
     -> ^(TOK_TABREF $tabname $ts? $alias?)
     ;
 
+databaseName
+@init { gParent.pushMsg("database name", state); }
+@after { gParent.popMsg(state); }
+    :
+    catalog=identifier DOT db=identifier?
```
ayushtkn, line 238: why was ? needed here?

zhangbutao: @ayushtkn Thanks for finding this issue.
Yes, `? ` is not needed. We can remove it.

PR #6089, opened 2025-09-20
[WIP]HIVE-29203:get_aggr_stats_for doesn't aggregate stats when direct sql…

ramitg254:
… batch retrieve is enabled


ramitg254:
dropped the batched test added earlier as `set hive.metastore.direct.sql.batch.size=1000 `was not changing the value in `MetastoreConf` in the real time run of q test.
but an individual tpcds test can be checked with:
`mvn test -pl itests/qtest -Pitests -Dtest=TestTezTPCDS30TBPerfCliDriver  -Dqfile=query16.q -Dhive.metastore.direct.sql.batch.size=1000`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1880,14 +1880,8 @@ private List<ColumnStatisticsObj> columnStatisticsObjForPartitions(
...
       public List<ColumnStatisticsObj> run(final List<String> inputColNames) throws MetaException {
-        return Batchable.runBatched(batchSize, partNames, new Batchable<String, ColumnStatisticsObj>() {
-          @Override
-          public List<ColumnStatisticsObj> run(List<String> inputPartNames) throws MetaException {
-            return columnStatisticsObjForPartitionsBatch(catName, dbName, tableName, inputPartNames,
-                inputColNames, engine, areAllPartsFound, useDensityFunctionForNDVEstimation, ndvTuner,
-                enableBitVector, enableKll);
-          }
-        });
+        return columnStatisticsObjForPartitionsBatch(catName, dbName, tableName, partNames, inputColNames, engine,
```
dengzhhu653: nit: can we just aggrStatsUseJava directly here and remove the columnStatisticsObjForPartitionsBatch

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1953,306 +1942,6 @@ private List<ColumnStatisticsObj> aggrStatsUseJava(String catName, String dbName
         areAllPartsFound, useDensityFunctionForNDVEstimation, ndvTuner);
   }
 
-  private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
```
dengzhhu653, line 1960: +1 for removing the `aggrStatsUseDB`, if we enable the batch retrieval, then the stats might not be aggregated per column. If we don't, we might hit the limitation of maximum parameters for PreparedStatement for some dbs.

Let's see how the test going

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done

dengzhhu653:
+1. cc @kasakrisz @deniskuzZ @nrg4878 @saihemanth-cloudera @zhangbutao in case if have any other ideas.

zhangbutao:
> ### How was this patch tested?
> by running tpcds tests locally with setting hive.metastore.direct.sql.batch.size as 1000 in HiveConf.java and MetastoreConf.java.

Hi @ramitg254, Would the tpcds tests fail without this PR? BTW, What is the size of the tpcds test dataset?

ramitg254:
> > ### How was this patch tested?
> > by running tpcds tests locally with setting hive.metastore.direct.sql.batch.size as 1000 in HiveConf.java and MetastoreConf.java.
> 
> Hi @ramitg254, Would the tpcds tests fail without this PR? BTW, What is the size of the tpcds test dataset?

Hi @zhangbutao, without this pr tpcds tests will fail whenever `hive.metastore.direct.sql.batch.size` > 0 in cases partition size is greater than the batch size.
there were around 55 tpcds test failures when property is set to 1000, and one such example is `query16.q `

and i think dataset size would be 30tb as I was using `-Dtest=TestTezTPCDS30TBPerfCliDriver` and it is picking it up from docker container

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
deniskuzZ: do i get it right. we are moving stats aggregation from backend db to Java? what would be the impact on performance? 
someone was working on this optimization and now we drop it?

kasakrisz: `aggrStatsUseDB` can only be used if [hive.stats.fetch.bitvector and metastore.stats.fetch.kll](https://github.com/apache/hive/blob/008f4d11e230f30103dc156247f97bf4d18e0b33/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java#L1467-L1470) are `false`. Some tests enable them via `set` command in q files or entire tests suits via `hive-site` but the default is false. Aggregating the stats at the backend db is usually faster then doing it in java so we can lose performance with this patch in some cases. 

@ramitg254 
1. Could you please investigate how to aggregate the results of subsequent `aggrStatsUseDB` calls?
2. Seems that we have test coverage for this:  [TestObjectStore.java.testAggrStatsUseDB](https://github.com/apache/hive/blob/008f4d11e230f30103dc156247f97bf4d18e0b33/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestObjectStore.java#L1009-L1032) Should it be removed along with `aggrStatsUseDB`?

@dengzhhu653 
> If we don't, we might hit the limitation of maximum parameters for PreparedStatement for some dbs.
>
> Let's see how the test going

Do we have tests using other dbs than derby and the postgres image?

ramitg254: @kasakrisz I have added aggregation for `aggrStatsUseDB` earlier upto the commit https://github.com/apache/hive/pull/6089/commits/b56681625ebed7ff057847cdeedd52bd25dee41f but on later suggestion I removed it

saihemanth-cloudera: > Do we have tests using other dbs than Derby and the PostgreSQL image?

Oracle has a 1000 parameter limit. It would be nice to run performance testing on PostgreSQL, Oracle, mysql before removing this optimization.

saihemanth-cloudera: @ramitg254 - if you can address my above concern, I can +1 too. Thanks.

ramitg254: @saihemanth-cloudera  as per HIVE-16997 bit vector is set to true for tpcds tests even if default value is false making it go back to `aggrStatsUseJava` and even if i set it to false to fall back to `aggrStatsUseDB` the results are different for .out files of tpcds queries.
So my concern is do I need to first obtain those different results generated via `aggrStatsUseDB` and then run performance testing while validating with those results.

ramitg254:
Hi,
I ran a script which executes the queries of pattern cbo_* and query* under perf directory with `-Dhive.stats.fetch.bitvector=false`  and with the postgres dump the overall time elapsed results were like below:
Screenshot 2025-11-18 at 6 23 06 PM

and since there is significant gap in overall time elapsed to execute all the queries and also results of some queries varies from what I recorded from `aggrStatsUseDB` earlier after its drop  in case of `-Dhive.stats.fetch.bitvector=false`. Considering this performance drop and varied results I don't think we should  go ahead with dropping of `aggrStatsUseDB`.

Since currently that's the case so I have raised this current patch which addresses my original concern of handling the behaviour of merging in case of direct sql batch size defined and in this patch I am just adding merging logic which and no change in the earlier behaviour so there won't be any performance issues as well  but similar like I tested it out with same set of queries and config and results which I saved earlier originally get generated via `aggrStatsUseDB` didn't change with this logic as well and their time elapsed results are somewhat like this:
Screenshot 2025-11-18 at 6 36 57 PM

@dengzhhu653 @deniskuzZ @kasakrisz @saihemanth-cloudera  please share your thoughts on this

kasakrisz:
@ramitg254 
Thanks for reporting this bug and working on the fix. I have a few questions to get a better picture of the issue:

> I ran a script which executes the queries of pattern cbo_* and query* under perf directory with

1. Did you actually execute all TPC-DS queries or just compile them? The driver `TestTezTPCDS30TBPerfCliDriver` doesn't execute the queries since the data is not available. It uses a Postgres HMS backend db dump to simulate an environment where the TPC-DS schema exists and calls Hive's SQL compiler using the `explain` and `explain cbo` commands.

2. Do the numbers you shared in the tables (Apache master version, with aggrStatsUseDB/without aggrStatsUseDB and with batching/without batching) show the overall compilation time of all queries?

3. I haven't found any new tests in this patch, no golden file changes either. Could you please provide a minimal repro of the issue? Please don't copy-paste any q file of the tpc-ds queries. IIUC this issue should be reproducible with a table having a few partitions and a batch size smaller than the number of partitions. Unit tests are also welcome.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1942,6 +1937,43 @@ private List<ColumnStatisticsObj> columnStatisticsObjForPartitionsBatch(String c
     }
   }
 
+  private Batchable<String, Object[]> jobsBatching(final String queryText0, final String catName, final String dbName,
```
kasakrisz: Can this method be used for other than loading per partition column statistics? If not could you please rename it how more specific name.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1977,116 +2020,96 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+        + "count((\"LONG_HIGH_VALUE\"-\"LONG_LOW_VALUE\")/cast(\"NUM_DISTINCTS\" as decimal)),"
+        + "sum((\"DOUBLE_HIGH_VALUE\"-\"DOUBLE_LOW_VALUE\")/\"NUM_DISTINCTS\"),"
+        + "count((\"DOUBLE_HIGH_VALUE\"-\"DOUBLE_LOW_VALUE\")/\"NUM_DISTINCTS\"),"
+        + "sum((cast(\"BIG_DECIMAL_HIGH_VALUE\" as decimal)-cast(\"BIG_DECIMAL_LOW_VALUE\" as decimal))/\"NUM_DISTINCTS\"),"
+        + "count((cast(\"BIG_DECIMAL_HIGH_VALUE\" as decimal)-cast(\"BIG_DECIMAL_LOW_VALUE\" as decimal))/\"NUM_DISTINCTS\"),"
+        + "sum(\"NUM_DISTINCTS\")" + " from " + PART_COL_STATS + "" + " inner join " + PARTITIONS + " on "
+        + PART_COL_STATS + ".\"PART_ID\" = " + PARTITIONS + ".\"PART_ID\"" + " inner join " + TBLS + " on " + PARTITIONS
+        + ".\"TBL_ID\" = " + TBLS + ".\"TBL_ID\"" + " inner join " + DBS + " on " + TBLS + ".\"DB_ID\" = " + DBS
+        + ".\"DB_ID\"" + " where " + DBS + ".\"CTLG_NAME\" = ? and " + DBS + ".\"NAME\" = ? and " + TBLS
+        + ".\"TBL_NAME\" = ? ";
```
kasakrisz: Why was the formatting of the query changed? I think the original was easier to read.

```
select
from
inner join
...
where
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1977,116 +2020,96 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
-      start = doTrace ? System.nanoTime() : 0;
+      queryText =
+          "select \"COLUMN_NAME\", \"COLUMN_TYPE\", count(\"PART_COL_STATS\".\"PART_ID\") " + " from " + PART_COL_STATS
+              + " inner join " + PARTITIONS + " on " + PART_COL_STATS + ".\"PART_ID\" = " + PARTITIONS + ".\"PART_ID\""
+              + " inner join " + TBLS + " on " + PARTITIONS + ".\"TBL_ID\" = " + TBLS + ".\"TBL_ID\"" + " inner join "
+              + DBS + " on " + TBLS + ".\"DB_ID\" = " + DBS + ".\"DB_ID\"" + " where " + DBS + ".\"CTLG_NAME\" = ? and "
+              + DBS + ".\"NAME\" = ? and " + TBLS + ".\"TBL_NAME\" = ? " + " and " + PART_COL_STATS
+              + ".\"COLUMN_NAME\" in (%1$s)" + " and " + PARTITIONS + ".\"PART_NAME\" in (%2$s)" + " and "
+              + PART_COL_STATS + ".\"ENGINE\" = ? " + " group by " + PART_COL_STATS + ".\"COLUMN_NAME\", "
+              + PART_COL_STATS + ".\"COLUMN_TYPE\"";
```
kasakrisz: Why was the formatting of the query changed? I think the original was easier to read.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1942,6 +1937,43 @@ private List<ColumnStatisticsObj> columnStatisticsObjForPartitionsBatch(String c
     }
   }
 
+  private Batchable<String, Object[]> jobsBatching(final String queryText0, final String catName, final String dbName,
+      final String tableName, final List<String> partNames, final String engine, final boolean doTrace) {
+    return new Batchable<String, Object[]>() {
+      @Override
+      public List<Object[]> run(final List<String> inputColNames)
+          throws MetaException {
+        Batchable<String, Object[]> b2 = new Batchable<String, Object[]>() {
```
kasakrisz: Instead of `b2`, please use a more specific name. How about `partitionNameBatches`? That way the variable name reflects that we are batching on the partition names”

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1977,116 +2020,96 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+          "select \"COLUMN_NAME\", \"COLUMN_TYPE\", count(\"PART_COL_STATS\".\"PART_ID\") " + " from " + PART_COL_STATS
+              + " inner join " + PARTITIONS + " on " + PART_COL_STATS + ".\"PART_ID\" = " + PARTITIONS + ".\"PART_ID\""
+              + " inner join " + TBLS + " on " + PARTITIONS + ".\"TBL_ID\" = " + TBLS + ".\"TBL_ID\"" + " inner join "
+              + DBS + " on " + TBLS + ".\"DB_ID\" = " + DBS + ".\"DB_ID\"" + " where " + DBS + ".\"CTLG_NAME\" = ? and "
+              + DBS + ".\"NAME\" = ? and " + TBLS + ".\"TBL_NAME\" = ? " + " and " + PART_COL_STATS
+              + ".\"COLUMN_NAME\" in (%1$s)" + " and " + PARTITIONS + ".\"PART_NAME\" in (%2$s)" + " and "
+              + PART_COL_STATS + ".\"ENGINE\" = ? " + " group by " + PART_COL_STATS + ".\"COLUMN_NAME\", "
+              + PART_COL_STATS + ".\"COLUMN_TYPE\"";
+
+      Batchable<String, Object[]> b = jobsBatching(queryText, catName, dbName, tableName, partNames, engine, doTrace);
```
kasakrisz: Instead of `b` please use more specific variable name.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1977,116 +2020,96 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
-          colStats.add(prepareCSObjWithAdjustedNDV(row, 0,
-              useDensityFunctionForNDVEstimation, ndvTuner));
+      queryText = commonPrefix + " and \"COLUMN_NAME\" in (%1$s)" + " and " + PARTITIONS + ".\"PART_NAME\" in (%2$s)"
+          + " and \"ENGINE\" = ? " + " group by \"COLUMN_NAME\", \"COLUMN_TYPE\"";
+      Batchable<String, Object[]> b = jobsBatching(queryText, catName, dbName, tableName, partNames, engine, doTrace);
+      List<ColumnStatisticsObj> colStats = new ArrayList<>(colNames.size());
+      try {
+        List<Object[]> list = Batchable.runBatched(batchSize, colNames, b);
+        Map<String, List<Object[]>> colSubList = columnWiseSubList(list);
+        for (Map.Entry<String, List<Object[]>> entry : colSubList.entrySet()) {
```
kasakrisz: Could this list to map conversion be avoided? We could save memory in case of heavily partitioned tables with lots of columns.

IIUC a column can be appear in multiple batches so the stats has to be merged. This is what `columnStatisticsObjWithAdjustedNDV` do. Can it be called more than once per column.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2098,29 +2121,39 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
-        try (QueryWrapper query = new QueryWrapper(pm.newQuery("javax.jdo.query.SQL", queryText))) {
+        queryText =
+            "select \"COLUMN_NAME\", sum(\"NUM_NULLS\"), sum(\"NUM_TRUES\"), sum(\"NUM_FALSES\"), sum(\"NUM_DISTINCTS\")"
+                + " from " + PART_COL_STATS + " inner join " + PARTITIONS + " on " + PART_COL_STATS + ".\"PART_ID\" = "
+                + PARTITIONS + ".\"PART_ID\"" + " inner join " + TBLS + " on " + PARTITIONS + ".\"TBL_ID\" = " + TBLS
+                + ".\"TBL_ID\"" + " inner join " + DBS + " on " + TBLS + ".\"DB_ID\" = " + DBS + ".\"DB_ID\""
+                + " where " + DBS + ".\"CTLG_NAME\" = ? and " + DBS + ".\"NAME\" = ? and " + TBLS + ".\"TBL_NAME\" = ? "
+                + " and " + PART_COL_STATS + ".\"COLUMN_NAME\" in (" + makeParams(extraColumnNameTypeParts.size()) + ")"
+                + " and " + PARTITIONS + ".\"PART_NAME\" in (" + makeParams(partNames.size()) + ")" + " and "
+                + PART_COL_STATS + ".\"ENGINE\" = ? " + " group by " + PART_COL_STATS + ".\"COLUMN_NAME\"";
```
kasakrisz: Please restore the original formatting of the query.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2098,29 +2121,39 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
           extraColumnNames.addAll(extraColumnNameTypeParts.keySet());
-          Object qResult = executeWithArray(query.getInnerQuery(),
-              prepareParams(catName, dbName, tableName, partNames,
-                  extraColumnNames, engine), queryText);
-          if (qResult == null) {
-            return Collections.emptyList();
+          List<Object[]> unmergedList = Batchable.runBatched(batchSize, extraColumnNames, b);
+          Map<String, List<Object[]>> colSubList = columnWiseSubList(unmergedList);
+          List<Object[]> mergedList = new ArrayList<>();
+          for (Map.Entry<String, List<Object[]>> entry : colSubList.entrySet()) {
```
kasakrisz: Can the list to map conversion avoided?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2174,76 +2240,48 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+
+              queryText =
+                  "select \"" + colStatName + "\",\"PART_NAME\" from " + PART_COL_STATS + " inner join " + PARTITIONS
+                      + " on " + PART_COL_STATS + ".\"PART_ID\" = " + PARTITIONS + ".\"PART_ID\"" + " inner join "
+                      + TBLS + " on " + PARTITIONS + ".\"TBL_ID\" = " + TBLS + ".\"TBL_ID\"" + " inner join " + DBS
+                      + " on " + TBLS + ".\"DB_ID\" = " + DBS + ".\"DB_ID\"" + " where " + DBS
+                      + ".\"CTLG_NAME\" = ? and " + DBS + ".\"NAME\" = ? and " + TBLS + ".\"TBL_NAME\" = ? " + " and "
+                      + PART_COL_STATS + ".\"COLUMN_NAME\" in (%1$s)" + " and " + PARTITIONS
+                      + ".\"PART_NAME\" in (%2$s)" + " and " + PART_COL_STATS + ".\"ENGINE\" = ? " + " order by "
+                      + orderByExpr;
```
kasakrisz: Please format the query
```
select
from
inner join
...
where
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2254,6 +2292,69 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
     }
   }
 
+  private ColumnStatisticsObj columnStatisticsObjWithAdjustedNDV(List<Object[]> list, int i,
```
kasakrisz: Please use meaningful parameter names instead of `list` and `i`.
The actual parameter value of `i` is always 0. Can it be removed?

When does this method called? I attached a debugger but it never hit?
```
mvn test -pl itests/qtest -Pitests -Dtest=TestTezTPCDS30TBPerfCliDriver  -Dqfile=query16.q -Dhive.metastore.direct.sql.batch.size=1000 -Drat.skip -Dmaven.surefire.debug
```

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2254,6 +2292,69 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+    Object avglen = row[j++];
+    Object maxlen = row[j++];
+    Object trues = row[j++];
+    Object falses = row[j++];
+    Object sumLong = row[j++];
+    Object countLong = row[j++];
+    Object sumDouble = row[j++];
+    Object countDouble = row[j++];
+    Object sumDecimal = row[j++];
+    Object countDecimal = row[j++];
```
kasakrisz: Please use constant indices instead of `j++`.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2254,6 +2292,69 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+    Object sumDist = row[j];
+    for (int k = 1; k < list.size(); k++) {
```
kasakrisz: Can this be replaced with
```
for (Object[] row : list)
```
?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2254,6 +2292,69 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+      maxlen = MetastoreDirectSqlUtils.max(maxlen, row[j++]);
+      trues = MetastoreDirectSqlUtils.sum(trues, row[j++]);
+      falses = MetastoreDirectSqlUtils.sum(falses, row[j++]);
+      sumLong = MetastoreDirectSqlUtils.sum(sumLong, row[j++]);
+      countLong = MetastoreDirectSqlUtils.sum(countLong, row[j++]);
+      sumDouble = MetastoreDirectSqlUtils.sum(sumDouble, row[j++]);
+      countDouble = MetastoreDirectSqlUtils.sum(countDouble, row[j++]);
+      sumDecimal = MetastoreDirectSqlUtils.sum(sumDecimal, row[j++]);
+      countDecimal = MetastoreDirectSqlUtils.sum(countDecimal, row[j++]);
+      sumDist = MetastoreDirectSqlUtils.sum(sumDist, row[j]);
```
kasakrisz: Please use constant indices instead of `j++`.

ramitg254:
> @ramitg254 Thanks for reporting this bug and working on the fix. I have a few questions to get a better picture of the issue:
> 
> > I ran a script which executes the queries of pattern cbo_* and query* under perf directory with
> 
> 1. Did you actually execute all TPC-DS queries or just compile them? The driver `TestTezTPCDS30TBPerfCliDriver` doesn't execute the queries since the data is not available. It uses a Postgres HMS backend db dump to simulate an environment where the TPC-DS schema exists and calls Hive's SQL compiler using the `explain` and `explain cbo` commands.
> 2. Do the numbers you shared in the tables (Apache master version, with aggrStatsUseDB/without aggrStatsUseDB and with batching/without batching) show the overall compilation time of all queries?
> 3. I haven't found any new tests in this patch, no golden file changes either. Could you please provide a minimal repro of the issue? Please don't copy-paste any q file of the tpc-ds queries. IIUC this issue should be reproducible with a table having a few partitions and a batch size smaller than the number of partitions. Unit tests are also welcome.

This only addresses the fact that stats results shouldn't change after batch size is defined to some value.

1. I didn't ran the actual queries only the simulation used by the `TestTezTPCDS30TBPerfCliDriver` and ran those explain queries as you mentioned and recorded their time elapsed. 

2. yes this the overall compilation time of the queries executed via `TestTezTPCDS30TBPerfCliDriver` under different scenarios

3. I'll try to add some unit test to replicate this behaviour.

dengzhhu653:
I think we can make the `aggrStatsUseDB` be simpler, 
 1. run the below SQL batched by part names:
 ```java
    String query  = "select \"COLUMN_NAME\", \"COLUMN_TYPE\", "
        + "min(\"LONG_LOW_VALUE\"), max(\"LONG_HIGH_VALUE\"), min(\"DOUBLE_LOW_VALUE\"), max(\"DOUBLE_HIGH_VALUE\"), "
        + "min(cast(\"BIG_DECIMAL_LOW_VALUE\" as decimal)), max(cast(\"BIG_DECIMAL_HIGH_VALUE\" as decimal)), "
        + "sum(\"NUM_NULLS\"), max(\"NUM_DISTINCTS\"), "
        + "max(\"AVG_COL_LEN\"), max(\"MAX_COL_LEN\"), sum(\"NUM_TRUES\"), sum(\"NUM_FALSES\"), "
        // The following data is used to compute a partitioned table's NDV based
        // on partitions' NDV when useDensityFunctionForNDVEstimation = true. Global NDVs cannot be
        // accurately derived from partition NDVs, because the domain of column value two partitions
        // can overlap. If there is no overlap then global NDV is just the sum
        // of partition NDVs (UpperBound). But if there is some overlay then
        // global NDV can be anywhere between sum of partition NDVs (no overlap)
        // and same as one of the partition NDV (domain of column value in all other
        // partitions is subset of the domain value in one of the partition)
        // (LowerBound).But under uniform distribution, we can roughly estimate the global
        // NDV by leveraging the min/max values.
        // And, we also guarantee that the estimation makes sense by comparing it to the
        // UpperBound (calculated by "sum(\"NUM_DISTINCTS\")")
        // and LowerBound (calculated by "max(\"NUM_DISTINCTS\")")
        + "sum((\"LONG_HIGH_VALUE\"-\"LONG_LOW_VALUE\")/cast(\"NUM_DISTINCTS\" as decimal)),"
        + "sum((\"DOUBLE_HIGH_VALUE\"-\"DOUBLE_LOW_VALUE\")/\"NUM_DISTINCTS\"),"
        + "sum((cast(\"BIG_DECIMAL_HIGH_VALUE\" as decimal)-cast(\"BIG_DECIMAL_LOW_VALUE\" as decimal))/\"NUM_DISTINCTS\"),"
        + "count(PARTITIONS + ".\"PART_ID\"),"
        + "sum(\"NUM_DISTINCTS\")" + " from " + PART_COL_STATS + "" + " inner join " + PARTITIONS + " on "
        + PART_COL_STATS + ".\"PART_ID\" = " + PARTITIONS + ".\"PART_ID\"" + " inner join " + TBLS + " on " + PARTITIONS
        + ".\"TBL_ID\" = " + TBLS + ".\"TBL_ID\"" + " inner join " + DBS + " on " + TBLS + ".\"DB_ID\" = " + DBS
        + ".\"DB_ID\"" + " where " + DBS + ".\"CTLG_NAME\" = ? and " + DBS + ".\"NAME\" = ? and " + TBLS
        + ".\"TBL_NAME\" = ?  and \"COLUMN_NAME\" in (%1$s)" + " and " + PARTITIONS + ".\"PART_NAME\" in (%2$s)"
          + " and \"ENGINE\" = ? " + " group by \"COLUMN_NAME\", \"COLUMN_TYPE\"";
```
   2. For each batch, merge the returned rows as we does in this PR:
   ```java
  try {
        List<Object[]> list = Batchable.runBatched(batchSize, partNames, b);
        Map<String, List<Object[]>> colSubList = columnWiseSubList(list);
        for (Map.Entry<String, List<Object[]>> entry : colSubList.entrySet()) {
          colStats.add(columnStatisticsObjWithAdjustedNDV(entry.getValue(), 0, useDensityFunctionForNDVEstimation, ndvTuner));
          Deadline.checkTimeout();
        }
      } finally {
        b.closeAllQueries();
      }
```
 3.  for each `ColumnStatisticsObj` got from step 2, change the `sum` by `sum * partNames.size() / count(PARTITIONS + ".\"PART_ID\") for this column`
 
By this, we can make the  `aggrStatsUseDB` simple and clear, and don't send the unnecessary queries back and forth to backing db.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2254,6 +2292,69 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
     }
   }
 
+  private ColumnStatisticsObj columnStatisticsObjWithAdjustedNDV(List<Object[]> list, int i,
+                                                                 boolean useDensityFunctionForNDVEstimation, double ndvTuner)
+          throws MetaException {
```
dengzhhu653: can this method move into `StatObjectConverter`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done renamed to `columnWisePartitionBatcher`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, restored the original one

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, restored the original one

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, I went with `partitionBatchesFetcher`, it was sounding more clearer to me

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, renamed to `columnWisePartitionBatches`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done restored the formatting

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, renamed list to `columnBatchesOutput` and removed it as it was not needed
and you also need to use `-Dhive.stats.fetch.bitvector=false` also to step into this method as it is associated with `aggrStatsUseDB`. 
and if you want to compare results then first run the command without `-Dhive.metastore.direct.sql.batch.size=1000` and store the results and then compare those results with including `-Dhive.metastore.direct.sql.batch.size=1000` to compare the results produced after this.

this need to be done because `aggrStatsUseDB` sometimes produces different results from expected as with disabled bit vector and disabled kll sketch the exact ndv  estimation is not possible in some cases so approximate results are produced by `aggrStatsUseDB` and these changes are keeping the `aggrStatsUseDB` batched results consistent with `aggrStatsUseDB` unbatched results.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, added constants in place of it

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: I don't think we should do that as current loops start with `index =1`  and it is able to replace the method `prepareCSObjWithAdjustedNDV` and if we try to implement it in the other way the we will need to take care of the conditionals and return statement inside loop which will not be cleaner than current one

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, added constants in place of it

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: I am not sure about this as this method is replacing `prepareCSObjWithAdjustedNDV` method and it is only limited to `aggrStatsUseDB` and not other places

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: I understand but we will need map to store the redundant column results obtained from different batches for same column name and those needs to be merged and other issue is we can't be sure of the no. of elements for a particular column name which needs to be merged as there are else cases for `areAllPartsFound` otherwise we would have gone for sorting the obtained results and then merging the results obtained at a particular gap.

currently we are only calling `columnStatisticsObjWithAdjustedNDV` once per column as we are calling it when we obtained all the batched results and maintained that in the map and it is doing merging per column.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, restored

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: I don't think it can be avoided same as the above mentioned reason

ramitg254:
> I think we can make the `aggrStatsUseDB` be simpler,
> 
> 1. run the below SQL batched by part names:
> 
> ```java
>    String query  = "select \"COLUMN_NAME\", \"COLUMN_TYPE\", "
>        + "min(\"LONG_LOW_VALUE\"), max(\"LONG_HIGH_VALUE\"), min(\"DOUBLE_LOW_VALUE\"), max(\"DOUBLE_HIGH_VALUE\"), "
>        + "min(cast(\"BIG_DECIMAL_LOW_VALUE\" as decimal)), max(cast(\"BIG_DECIMAL_HIGH_VALUE\" as decimal)), "
>        + "sum(\"NUM_NULLS\"), max(\"NUM_DISTINCTS\"), "
>        + "max(\"AVG_COL_LEN\"), max(\"MAX_COL_LEN\"), sum(\"NUM_TRUES\"), sum(\"NUM_FALSES\"), "
>        // The following data is used to compute a partitioned table's NDV based
>        // on partitions' NDV when useDensityFunctionForNDVEstimation = true. Global NDVs cannot be
>        // accurately derived from partition NDVs, because the domain of column value two partitions
>        // can overlap. If there is no overlap then global NDV is just the sum
>        // of partition NDVs (UpperBound). But if there is some overlay then
>        // global NDV can be anywhere between sum of partition NDVs (no overlap)
>        // and same as one of the partition NDV (domain of column value in all other
>        // partitions is subset of the domain value in one of the partition)
>        // (LowerBound).But under uniform distribution, we can roughly estimate the global
>        // NDV by leveraging the min/max values.
>        // And, we also guarantee that the estimation makes sense by comparing it to the
>        // UpperBound (calculated by "sum(\"NUM_DISTINCTS\")")
>        // and LowerBound (calculated by "max(\"NUM_DISTINCTS\")")
>        + "sum((\"LONG_HIGH_VALUE\"-\"LONG_LOW_VALUE\")/cast(\"NUM_DISTINCTS\" as decimal)),"
>        + "sum((\"DOUBLE_HIGH_VALUE\"-\"DOUBLE_LOW_VALUE\")/\"NUM_DISTINCTS\"),"
>        + "sum((cast(\"BIG_DECIMAL_HIGH_VALUE\" as decimal)-cast(\"BIG_DECIMAL_LOW_VALUE\" as decimal))/\"NUM_DISTINCTS\"),"
>        + "count(PARTITIONS + ".\"PART_ID\"),"
>        + "sum(\"NUM_DISTINCTS\")" + " from " + PART_COL_STATS + "" + " inner join " + PARTITIONS + " on "
>        + PART_COL_STATS + ".\"PART_ID\" = " + PARTITIONS + ".\"PART_ID\"" + " inner join " + TBLS + " on " + PARTITIONS
>        + ".\"TBL_ID\" = " + TBLS + ".\"TBL_ID\"" + " inner join " + DBS + " on " + TBLS + ".\"DB_ID\" = " + DBS
>        + ".\"DB_ID\"" + " where " + DBS + ".\"CTLG_NAME\" = ? and " + DBS + ".\"NAME\" = ? and " + TBLS
>        + ".\"TBL_NAME\" = ?  and \"COLUMN_NAME\" in (%1$s)" + " and " + PARTITIONS + ".\"PART_NAME\" in (%2$s)"
>          + " and \"ENGINE\" = ? " + " group by \"COLUMN_NAME\", \"COLUMN_TYPE\"";
> ```
> 
> 2. For each batch, merge the returned rows as we does in this PR:
> 
> ```java
> try {
>      List<Object[]> list = Batchable.runBatched(batchSize, partNames, b);
>      Map<String, List<Object[]>> colSubList = columnWiseSubList(list);
>      for (Map.Entry<String, List<Object[]>> entry : colSubList.entrySet()) {
>        colStats.add(columnStatisticsObjWithAdjustedNDV(entry.getValue(), 0, useDensityFunctionForNDVEstimation, ndvTuner));
>        Deadline.checkTimeout();
>      }
>    } finally {
>      b.closeAllQueries();
>    }
> ```
> 
>  3. for each `ColumnStatisticsObj` got from step 2, change the `sum` by `sum * partNames.size() / count(PARTITIONS + ".\"PART_ID\") for this column`
> 
> By this, we can make the `aggrStatsUseDB` simple and clear, and don't send the unnecessary queries back and forth to backing db.

IIUC you are talking about the different queries for the `else` block of `if (areAllPartsFound)` but I don't have the unit test to test this behaviour as it is for a broken behaviour when a column is not present in all partiitions. So, I didn't change any implementation and queries for those case and only replaced the results of list with a map and then performed merging and add to the final stats list so that it won't break any functionality.

I can't change the implementation without having a unit test for that behaviour.

ramitg254:
Hi,
so it is regarding the unit tests added:
1. why I am going for driver based approach instead of simply adding a q file under an already present driver:
   -> the `-Dhive.stats.fetch.bitvector` and `-Dhive.metastore.direct.sql.batch.size` can't be set at the query time inside a q file and needed to be set in `hive-site.xml`
   
 2. why going for a docker image instead of a q file loading the data into the table:
   -> these changes are for scenario where statistics have been computed already and then someone specify this property `hive.metastore.direct.sql.batch.size` and restart hms so in that case results are different. But to replicate this in unit test it won't be possible if I define table and load data in q file as ` it will give ORM related exception when placing debug pointer although getting executed` . So what i did is created a `test_stats` table having  data in 13 partitions and then compute statistics for its columns and built a docker image with the stats dump and used that docker image.
  
the q files added are performing self join when `hive.metastore.direct.sql.batch.size=5` and the `hive.metastore.direct.sql.batch.size` is set according to the q file name.

you can validate the results via:
cherry-picking the unit tests added commit on the top of the current master branch and run:
`mvn test -pl itests/qtest -Pitests -Dtest=TestTezBatchedStatsCliDriver  -Dqfile=sketch_query.q -Dtest.output.overwrite`
similar for `no_sketch_query.q`
and you will see different results from expected one added in commit and if you just drop the property `hive.metastore.direct.sql.batch.size` from the `hive-site.xml`  added for this driver you will see the expected results which is the validation for corrected expected output with these changes in case of batching.

FYI: the stats are added in such a way that overlapping values lie in between batches of size 5 resulting in just taking maximum of ndv is not enough to evaluate ndv calculated without batching as it is the case failing only otherwise it passes even after having redundant per column elements which just simply gets concatenated via simple function like maximum and  all on the higher level

ramitg254:
and sonar is pointing out indentation issues which i don't think needs to be addressed as I kept it same as we have in those other drive files and config file and others are large brain methods like `aggrStatsUseDB` which are already part of code and splitting it up into separate methods with entire implementation intact can take up time and may break some scenarios addressed in that method

itests/qtest/src/test/java/org/apache/hadoop/hive/cli/TestTezBatchedStatsCliDriver.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.rules.TestRule;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import java.io.File;
+import java.util.List;
+
+@RunWith(Parameterized.class)
+public class TestTezBatchedStatsCliDriver {
```
kasakrisz: IMHO there is no need to add a new driver for testing the `ObjectStore.get_aggr_stats_for` method. Doing so would imply that we need new drivers to cover all combinations of `hive.stats.fetch.bitvector`, `hive.stats.fetch.kll`, `hive.metastore.direct.sql.batch.size`, etc. The number of drivers would explode if we followed this pattern.


Please take a look at an existing test case. It is not a q-test–based one, but a good integration test that focuses only on `ObjectStore`.
https://github.com/apache/hive/blob/ca105f8124072d19d88a83b2ced613d326c9a26b/standalone-metastore/metastore-server/src/test/java/org/apache/hadoop/hive/metastore/TestObjectStore.java#L1025-L1047
(The test method name `testAggrStatsUseDB` is misleading, though, because the test actually calls the `ObjectStore.get_aggr_stats_for` method, which eventually calls `MetaStoreDirectSql.testAggrStatsUseDB`.)

Configurations can be easily changed in these kinds of tests via the MetastoreConf object. We also have better ways to write assertions. These tests run faster than q-tests.

I temporarily added:
```
MetastoreConf.setLongVar(conf2, ConfVars.DIRECT_SQL_PARTITION_BATCH_SIZE, 2);
```
and I was able to reproduce the bug described in [HIVE-29203](https://issues.apache.org/jira/browse/HIVE-29203). Both with aggregating in the backend db (`aggrStatsUseDB`) and HMS side (`aggrStatsUseJava`).

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2174,86 +2270,132 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+    Object avgLong = row[AVGLONG];
+    Object avgDouble = row[AVGDOUBLE];
+    Object avgDecimal = row[AVGDECIMAL];
+    Object sumDist = row[SUMDIST];
+    if (row.length == 18) {
+      StatObjectConverter.fillColumnStatisticsData(cso.getColType(), data, llow, lhigh, dlow, dhigh, declow, dechigh,
+              nulls, dist, avglen, maxlen, trues, falses, avgLong, avgDouble, avgDecimal, sumDist,
+              useDensityFunctionForNDVEstimation, ndvTuner);
+      return cso;
+    }
```
kasakrisz: Why is this check needed? The schema of the query loading the stats is fix. The column list is defined in the query text.
Also when somehow the number of columns is 18 the aggregation is skipped and the `ColumnStatisticsData` objects is built by only the first row.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2174,86 +2270,132 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+    Object sumLong = row[AVGLONG];
+    Object countLong = row[AVGLONG + 1];
+    Object sumDouble = row[AVGDOUBLE + 1];
+    Object countDouble = row[AVGDOUBLE + 2];
+    Object sumDecimal = row[AVGDECIMAL + 2];
+    Object countDecimal = row[AVGDECIMAL + 3];
```
kasakrisz: Why are the constants used for indexing called `AVG...`? Aren't these `sums` and `counts`?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
kasakrisz: In both cases, whether `areAllPartsFound` is true or false, the same code is executed.
```
          list = Batchable.runBatched(batchSize, noExtraColumnNames, columnWisePartitionBatches);
          Map<String, List<Object[]>> colSubList = columnWiseSubList(list);
          for (Map.Entry<String, List<Object[]>> entry : colSubList.entrySet()) {
            colStats.add(
                    columnStatisticsObjWithAdjustedNDV(entry.getValue(), useDensityFunctionForNDVEstimation, ndvTuner));
            Deadline.checkTimeout();
          }

```
This can be extracted to a method. 

Maybe I'm missing something, but I don't see why the records can't be aggregated while iterating through them.
The map value could hold the partially aggregated data instead of an `ArrayList`, and the method `columnStatisticsObjWithAdjustedNDV` could merge the partially aggregated data with each record.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
kasakrisz: It seem the data is merged a few lines later:
```
for (Object[] row : subList) {
              mergedRow[1] = MetastoreDirectSqlUtils.sum(mergedRow[1], row[1]);
              mergedRow[2] = MetastoreDirectSqlUtils.sum(mergedRow[2], row[2]);
              mergedRow[3] = MetastoreDirectSqlUtils.sum(mergedRow[3], row[3]);
              mergedRow[4] = MetastoreDirectSqlUtils.sum(mergedRow[4], row[4]);
            }
```

Without transforming the `unmergedList` to `colSubList` the merge could be performed directly on the elements of  `unmergedList`: similar to above logic but instead of storing the partial results of the aggregation in one `mergeRow`, a map of `mergeRows` are need. One per column name.

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1954,9 +2007,20 @@ private List<ColumnStatisticsObj> aggrStatsUseJava(String catName, String dbName
...
   }
 
-  private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
-      String tableName, List<String> partNames, List<String> colNames, String engine,
-      boolean areAllPartsFound, boolean useDensityFunctionForNDVEstimation, double ndvTuner) throws MetaException {
+  private Map<String, List<Object[]>> columnWiseSubList(List<Object[]> list) {
+    Map<String, List<Object[]>> colSubList = new HashMap<>();
+    for (Object[] row : list) {
+      String colName = (String) row[0];
+      colSubList.putIfAbsent(colName, new ArrayList<>());
```
kasakrisz: Instead of adding the elements to an `ArrayList` could you please aggregate the elements and store the aggregated data in the map.

itests/qtest/src/test/java/org/apache/hadoop/hive/cli/TestTezBatchedStatsCliDriver.java
ramitg254: Thanks for pointing it, I agree with this and added corresponding test in `TestObjectStore.java` and removed earlier one

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: It was added to address the cases if merged avg or merged stats is calculated already like here https://github.com/apache/hive/commit/5cb03e6facd083058cffa60ab3e1795e9e99e196#:~:text=avg[i%20/%202]%20=%20MetastoreDirectSqlUtils.divide(sum%2C%20count) which just needed to filled later,

but I have removed this check as kept everything in terms of sum and count which can handled later on

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: I was keeping it according to https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/IExtrapolatePartStatus.java#L28, but everything is dealt in terms of sum and count so I have changed the indexes

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: Thanks for explaining it , I misunderstood it earlier so I have updated it with this implementation

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1977,47 +2034,29 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
         // And, we also guarantee that the estimation makes sense by comparing it to the
         // UpperBound (calculated by "sum(\"NUM_DISTINCTS\")")
         // and LowerBound (calculated by "max(\"NUM_DISTINCTS\")")
-        + "avg((\"LONG_HIGH_VALUE\"-\"LONG_LOW_VALUE\")/cast(\"NUM_DISTINCTS\" as decimal)),"
-        + "avg((\"DOUBLE_HIGH_VALUE\"-\"DOUBLE_LOW_VALUE\")/\"NUM_DISTINCTS\"),"
-        + "avg((cast(\"BIG_DECIMAL_HIGH_VALUE\" as decimal)-cast(\"BIG_DECIMAL_LOW_VALUE\" as decimal))/\"NUM_DISTINCTS\"),"
+        + "sum((\"LONG_HIGH_VALUE\"-\"LONG_LOW_VALUE\")/cast(\"NUM_DISTINCTS\" as decimal)),"
+        + "count((\"LONG_HIGH_VALUE\"-\"LONG_LOW_VALUE\")/cast(\"NUM_DISTINCTS\" as decimal)),"
```
dengzhhu653: looks a single count `count(1)` is enough for me, 
`count((\"LONG_HIGH_VALUE\"-\"LONG_LOW_VALUE\")/cast(\"NUM_DISTINCTS\" as decimal))`
`count((\"DOUBLE_HIGH_VALUE\"-\"DOUBLE_LOW_VALUE\")/\"NUM_DISTINCTS\")`
`count((cast(\"BIG_DECIMAL_HIGH_VALUE\" as decimal)-cast(\"BIG_DECIMAL_LOW_VALUE\" as decimal))/\"NUM_DISTINCTS\")`

dengzhhu653:
could you please consider this comments?
https://github.com/apache/hive/pull/6089#discussion_r2589091584
https://github.com/apache/hive/pull/6089#discussion_r2589103939

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2159,102 +2188,185 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+                  (row, useDensityFunctionForNDVEstimation, ndvTuner));
           Deadline.checkTimeout();
         }
       }
       return colStats;
     }
   }
 
-  private ColumnStatisticsObj prepareCSObj (Object[] row, int i) throws MetaException {
+  private void columnWiseStatsMerger(
```
ramitg254: separate method for the code chunk mentioned :https://github.com/apache/hive/pull/6089#discussion_r2589091584

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2108,33 +2126,44 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
+          List<String> extraColumnNames = new ArrayList<>();
           extraColumnNames.addAll(extraColumnNameTypeParts.keySet());
-          Object qResult = executeWithArray(query.getInnerQuery(),
-              prepareParams(catName, dbName, tableName, partNames,
-                  extraColumnNames, engine), queryText);
-          if (qResult == null) {
-            return Collections.emptyList();
+          List<Object[]> unmergedList = Batchable.runBatched(batchSize, extraColumnNames, columnWisePartitionBatches);
+          Map<String, Object[]> colSumStatsMap = new HashMap<>();
+          for (Object[] row : unmergedList) {
```
ramitg254: using unmergedList and map for mergedRow as mentioned in: https://github.com/apache/hive/pull/6089#discussion_r2589103939

ramitg254:
> could you please consider this comments? [#6089 (comment)](https://github.com/apache/hive/pull/6089#discussion_r2589091584) [#6089 (comment)](https://github.com/apache/hive/pull/6089#discussion_r2589103939)

@dengzhhu653 I have already considered and implemented these comments:
1. https://github.com/apache/hive/pull/6089#discussion_r2621892496
2. https://github.com/apache/hive/pull/6089#discussion_r2621896928

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: are you saying that instead of count(expression) we can use count(1) as whenever it is not null all rows will be counted?
but isn't this count(expr) is more readable and understandable

ramitg254: my bad, done updated with count(1)

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2029,65 +2064,48 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
           + " inner join " + TBLS + " on " + PARTITIONS + ".\"TBL_ID\" = " + TBLS + ".\"TBL_ID\""
           + " inner join " + DBS + " on " + TBLS + ".\"DB_ID\" = " + DBS + ".\"DB_ID\""
           + " where " + DBS + ".\"CTLG_NAME\" = ? and " + DBS + ".\"NAME\" = ? and " + TBLS + ".\"TBL_NAME\" = ? "
-          + " and " + PART_COL_STATS + ".\"COLUMN_NAME\" in (" + makeParams(colNames.size()) + ")"
-          + " and " + PARTITIONS + ".\"PART_NAME\" in (" + makeParams(partNames.size()) + ")"
+          + " and " + PART_COL_STATS + ".\"COLUMN_NAME\" in (%1$s)"
+          + " and " + PARTITIONS + ".\"PART_NAME\" in (%2$s)"
           + " and " + PART_COL_STATS + ".\"ENGINE\" = ? "
           + " group by " + PART_COL_STATS + ".\"COLUMN_NAME\", " + PART_COL_STATS + ".\"COLUMN_TYPE\"";
```
dengzhhu653: In this query, the `count("PART_COL_STATS"."PART_ID")` would be the same as `count(1)` in the `commonPrefix`, so this query can be merged as well.
- For each column name, if the count == partNames.size() or count < 2, then we simply merge the stats;
- otherwise, we need extrapolation for this column,

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -165,6 +166,26 @@ class MetaStoreDirectSql {
   private DirectSqlUpdatePart directSqlUpdatePart;
   private DirectSqlInsertPart directSqlInsertPart;
 
+  private static final int COLNAME = 0;
```
dengzhhu653: personally I dont like this number of static fields, maybe we could move them somewhere, like IExtrapolatePartStatus

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
dengzhhu653: the `areAllPartsFound` makes little sense here, we can remove this branch

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done added as an enum in the IExtrapolatePartStatus.java

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: I agree this is same and can be merged with common prefix but we should not do that as this query fetches way less parameters than `commonPrefix` one and it would be faster than it and also this query is necessary prerequisite for finding `noExtraColumnNames` later on which uses `commonPrefix` query so we can't just store the result of query execution of `commonPrefix` one once and use for both this prerequisite and for `noExtraColumnNames` as columns needed to be passed to both of them will be different.
but I have replaced `count("PART_COL_STATS"."PART_ID")` with `count(1)`

ramitg254: also this else branch of  `areAllPartsFound` should not be removed as it is helpful in cases when statistics are computed before and some new partitions are added later on so some partition stats will be missing in that we will need this branch for that.
I have added the test case `testMissingPartsStatsAggrWithBackendDB` for this scenario in `TestObjectStore.java` as well

dengzhhu653: This method aggregates the column stats for `partNames`, this is known beforehand, why do we should take account of the "new partitions"?

ramitg254: correct me if I'm wrong:
we have `ANALYZE TABLE {table name} PARTITION({part cols}) COMPUTE STATISTICS FOR COLUMNS {columns passed};`
so due to this query there can be scenarios where partitions found for a column of a table where stats have beeen computed can vary from column to column of a table and in that case `partNames ` will be fixed and `areAllPartsFound` can be false and we need to handle this scenario as well.

ramitg254: like let's say for colA stats computed when table has 5 parts and for colB stats computed when table has 8 parts
so here `areAllPartsFound` will be false

dengzhhu653: I think this is not fully right. let's take an example:
In db, colA has the stats for part1, part2, part3, and colB has stats for part1,part2,part3,par4,part5 
If we want to aggregate the column stats among the partition(part1,part2,part3), as colA and colB both have the partition column stats, we just fetch the stats and aggregate them.
however if we want the aggregated stats for partition(part1,part2,part3,part4,par5), as colB has all of them, so just aggregate them. While for colA, as part4 and part5 are missing, we need some assumptions on those two missing stats.

ramitg254: yes as you mentioned some assumption is need for missing two stats so this else block is taking care of that only as it is first finding the the available stats which are already there then extrapolate its results with some logic like:
```
if (IExtrapolatePartStatus.aggrTypes[colStatIndex] == IExtrapolatePartStatus.AggrType.Sum) {
              Object o = sumMap.get(colName).get(colStatIndex);
              // +3 only for the case of SUM_NUM_DISTINCTS which is after count rows index
              int rowIndex = (colStatIndex == 15) ? colStatIndex + 3 : colStatIndex + 2;
              if (o == null) {
                row[rowIndex] = null;
              } else {
                Long val = MetastoreDirectSqlUtils.extractSqlLong(o);
                row[rowIndex] = val / sumVal * (partNames.size());
              }
            }
```

dengzhhu653: this extrapolation is just based on the existing stats on colA, am I right?
here the `val` is sum(colA) of the partition: part1,part2,part3, sumVal is 3, and partNames.size() is 5.

ramitg254: yes

ramitg254: @dengzhhu653 is there anything else that I should address?

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -2159,102 +2187,180 @@ private List<ColumnStatisticsObj> aggrStatsUseDB(String catName, String dbName,
...
-                Object[] avg = (Object[]) (fqr.get(0));
-                // colStatIndex=12,13,14 respond to "AVG_LONG", "AVG_DOUBLE",
-                // "AVG_DECIMAL"
-                row[2 + colStatIndex] = avg[colStatIndex - 12];
-                end = doTrace ? System.nanoTime() : 0;
-                MetastoreDirectSqlUtils.timingTrace(doTrace, queryText, start, end);
+                if (min[0] == null || max[0] == null) {
+                  row[2 + colStatIndex] = null;
+                } else {
+                  row[2 + colStatIndex] = extrapolateMethod.extrapolate(min, max, colStatIndex, indexMap);
```
ramitg254: linear extrapolation used

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: @dengzhhu653 removed `areAllPartsFound` and tried to keep `aggrStatsUseDB` simple. Please have a look

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1957,307 +2010,224 @@ private List<ColumnStatisticsObj> aggrStatsUseJava(String catName, String dbName
...
-                    + " and " + PARTITIONS + ".\"PART_NAME\" in (" + makeParams(partNames.size()) + ")"
-                    + " and " + PART_COL_STATS + ".\"ENGINE\" = ? "
-                    + " order by \"" + colStatName + "\"";
-              } else {
-                queryText = "select \"" + colStatName + "\",\"PART_NAME\" from " + PART_COL_STATS
+          } else if (IExtrapolatePartStatus.aggrTypes[colStatIndex] == IExtrapolatePartStatus.AggrType.Min ||
+                  IExtrapolatePartStatus.aggrTypes[colStatIndex] == IExtrapolatePartStatus.AggrType.Max) {
+            // if the aggregation type is min/max, we extrapolate from the
+            // left/right borders
+            String orderByExpr = decimal ? "cast(\"" + colStatName + "\" as decimal)" : "\"" + colStatName + "\"";
```
dengzhhu653: looks like we have already calculated the min/max for those `partialStatsRows` by this query:
https://github.com/apache/hive/blob/417062b3d9b0e21b9d4bc0e2c9ed740a7b1e0b08/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java#L2019

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1957,307 +2010,224 @@ private List<ColumnStatisticsObj> aggrStatsUseJava(String catName, String dbName
...
+      for (Object[] unmergedRow : unmergedColStatslist) {
+        String colName = (String) unmergedRow[0];
+        Object[] mergedRow = mergedColStatsMap.getOrDefault(colName, new Object[21]);
+        mergeBackendDBStats(mergedRow, unmergedRow);
+        mergedColStatsMap.put(colName, mergedRow);
+      }
+
+      for (Map.Entry<String, Object[]> entry : mergedColStatsMap.entrySet()) {
+        BigDecimal partCount = new BigDecimal(entry.getValue()[COUNT_ROWS.idx()].toString());
+        if (partCount.equals(new BigDecimal(partNames.size()))) {
```
dengzhhu653: if partCount < 2, we also don't need to add this column to `partialStatsRows`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
```diff
@@ -1884,14 +1904,8 @@ private List<ColumnStatisticsObj> columnStatisticsObjForPartitions(
     return Batchable.runBatched(batchSize, colNames, new Batchable<String, ColumnStatisticsObj>() {
```
dengzhhu653: Could you please move all `aggrColStatsForPartitions` related methods to a new class, similar to `DirectSqlUpdatePart`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: actually its not the same min max so here what we are doing is:
let's say we have `LONG_LOW_VALUE `which is evaluated by taking min() of it so it is `IExtrapolatePartStatus.AggrType.Min` for which we have its minimum value but we need its maximum value as well which is not yet evaluated and is required for 
`row[2 + colStatIndex] = extrapolateMethod.extrapolate(min, max, colStatIndex, indexMap);`
and also i could have used directly MAX() in the query for that instead of order by but `PART_NAME` is also used in the extrapolate method so going with order by
and vice-versa the case with `IExtrapolatePartStatus.AggrType.Min`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done, corrected it, somehow got left out while simplification

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
ramitg254: done moved to `DirectSqlAggrStats.java`

standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/MetaStoreDirectSql.java
dengzhhu653: yeah, the `PART_NAME` is missing in the `partialStatsRows`. looks there has some room for optimization, the min/max extrapolation will send out the same query two times, each time only one column is extrapolated. If think it's beyond the scope, we can file another jira to optimize it.

can you take care of the code format, and go through the sonar issues please?

ramitg254: yes, regarding the optimization it can be taken care in follow up jira
and for sonar issues I have addressed most of them which made sense to me and the left ones no. should be around 60s and those mostly with names like TBLS, PARTITIONS etc which are used around most of the other files so I thought it shouldn't be changesd and regarding the line length of queries and some lines I left intentionally as readability is better in current scenario with that and others are related to brain method and having more parameter like 7 or 8 for a method which I think shouldn't be done in this refractor as it can break some existing functionality will deviate from purpose of ticket

dengzhhu653:
Filed the HIVE-29416 to track the optimization.

ramitg254:
Thanks! @dengzhhu653 @kasakrisz @saihemanth-cloudera @deniskuzZ for review

PR #6090, opened 2025-09-20
HIVE-29218: LOAD OVERWRITE PARTITION on muti-level partititoned exter…

JsZero:
…nal Iceberg table may unexpectedly delete other partitions

This PR addresses [HIVE-29218](https://issues.apache.org/jira/browse/HIVE-29218), which reports an issue when using LOAD OVERWRITE on multi-level partitioned Iceberg tables with appendFile as the underlying implementation.

The problem was that the overwrite logic could mistakenly delete metadata of partitions that should not be affected. For example, given a table partitioned by pcol1 and pcol2, with existing data in partition pcol1=x/pcol2=y, running LOAD OVERWRITE into another partition such as pcol1=x/pcol2=z would incorrectly remove the metadata for pcol1=x/pcol2=y just because it partially overlaps on pcol1=x.

The PR fix the overwrite handling logic in Iceberg LOAD implementation to ensure that only the target partition is overwritten.

The current implementation may inadvertently delete metadata that should be preserved.

No. This change only fixes incorrect behavior where unrelated partitions' metadata might be unintentionally deleted during LOAD OVERWRITE. The intended user-facing behavior remains unchanged.

ut


aturoczy:
@ayushtkn @deniskuzZ Could you please review it?

iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java
```diff
@@ -171,11 +172,13 @@ public static void appendFiles(URI fromURI, String format, Table icebergTbl, boo
       if (isOverwrite) {
         DeleteFiles delete = transaction.newDelete();
         if (partitionSpec != null) {
+          Expression andResult = Expressions.alwaysTrue();
```
deniskuzZ: could we rename local var to `partitionExpr` and use `IcebergTableUtil.generateExpressionFromPartitionSpec(...)`?

````
if (partitionSpec != null) {
  Expression partitionExpr = IcebergTableUtil.generateExpressionFromPartitionSpec(icebergTbl, partitionSpec, true);
  delete.deleteFromRowFilter(partitionExpr);
} else {
  delete.deleteFromRowFilter(Expressions.alwaysTrue());
}
````

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java]: would be nice to apply small refactor and remove redundant stuff
````
public static Expression generateExpressionFromPartitionSpec(Table table, Map<String, String> partitionSpec,
    boolean latestSpecOnly) throws SemanticException {
  Map<String, PartitionField> partitionFieldMap = getPartitionFields(table, latestSpecOnly).stream()
      .collect(Collectors.toMap(PartitionField::name, Function.identity()));
  Expression finalExp = Expressions.alwaysTrue();
  for (Map.Entry<String, String> part : partitionSpec.entrySet()) {
    String partColName = part.getKey();
    if (partitionFieldMap.containsKey(partColName)) {
      PartitionField partitionField = partitionFieldMap.get(partColName);
      TransformSpec.TransformType transformType = TransformSpec.fromString(partitionField.transform().toString());
      if (TransformSpec.TransformType.IDENTITY == transformType) {
        final Type partKeyType = table.schema().findField(partitionField.sourceId()).type();
        final Object partKeyVal = Conversions.fromPartitionString(partKeyType, part.getValue());
        Expression boundPredicate = Expressions.equal(partColName, partKeyVal);
        finalExp = Expressions.and(finalExp, boundPredicate);
      } else {
        throw new SemanticException(
            String.format("Partition transforms are not supported here: %s", partColName));
      }
    } else {
      throw new SemanticException(String.format("No partition column by the name: %s", partColName));
    }
  }
  return finalExp;
}
````

Comment (JsZero) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java]: @aturoczy @ayushtkn @deniskuzZ  Thanks a lot for the feedback!
@deniskuzZ That makes sense, I’ll update the code soon.

Comment (deniskuzZ) [reply iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/HiveTableUtil.java]: thanks!

**Review by deniskuzZ — APPROVED** [review]
+1 

PS: if not optimize/refactor `IcebergTableUtils.generateExpressionFromPartitionSpec` here, we’ll probably need a follow-up

**Comment by JsZero** [comment]
@deniskuzZ Thanks for approving! I’ve added some small refactorings to improve clarity. Does it look clearer now?

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java [hunk 2]
```diff
@@ -491,13 +491,11 @@ public static Expression generateExpressionFromPartitionSpec(Table table, Map<St
...
       if (partitionFieldMap.containsKey(partColName)) {
         PartitionField partitionField = partitionFieldMap.get(partColName);
-        Type resultType = partitionField.transform().getResultType(table.schema()
-            .findField(partitionField.sourceId()).type());
-        Object value = Conversions.fromPartitionString(resultType, entry.getValue());
-        TransformSpec.TransformType transformType = TransformSpec.fromString(partitionField.transform().toString());
-        Iterable<?> iterable = () -> Collections.singletonList(value).iterator();
-        if (TransformSpec.TransformType.IDENTITY == transformType) {
-          Expression boundPredicate = Expressions.in(partitionField.name(), iterable);
+        Transform<?, ?> transform = partitionField.transform();
```
deniskuzZ: please inline inside the if block. no need for the local var
````
 if (partitionField.transform().isIdentity()) {
````

File: iceberg/iceberg-handler/src/main/java/org/apache/iceberg/mr/hive/IcebergTableUtil.java [hunk 3]
```diff
@@ -491,13 +491,11 @@ public static Expression generateExpressionFromPartitionSpec(Table table, Map<St
...
+        if (transform.isIdentity()) {
+          Type resultType = transform.getResultType(table.schema().findField(partitionField.sourceId()).type());
```
deniskuzZ: since it's identity no need for getResultType
```
final Type partKeyType = table.schema().findField(partitionField.sourceId()).type();
final Object partKeyVal = Conversions.fromPartitionString(partKeyType, entry.getValue());
````

deniskuzZ: please update also the SemanticException messages

- "Partition transforms are not supported here: %s", partColName));
- "No partition column by the name: %s", partColName));

deniskuzZ:
> @deniskuzZ Thanks for approving! I’ve added some small refactorings to improve clarity. Does it look clearer now?

thanks @JsZero! looks good, added few minor comments

JsZero:
> > @deniskuzZ Thanks for approving! I’ve added some small refactorings to improve clarity. Does it look clearer now?
> 
> thanks @JsZero! looks good, added few minor comments

The code refactoring has been incorporated. Thanks a lot for the review @deniskuzZ !

PR #6092, opened 2025-09-22
HIVE-29217: Add configuration to choose materialization strategy for CTEs

zabetak:
Avoid interference between AST and CBO based materialization of common table expressions (CTEs). More details in HIVE-29217.

Yes but highly unlikely. Users may see plan/perf changes if they set/rely on the **experimental** property `hive.optimize.cte.suggester.class` and they expect both AST and CBO materialization to trigger sequentially one after the other.

Existing plus new tests.
```
mvn test -Dtest=TestMiniLlapLocalCliDriver -Dqfile_regex=.*cte.*
mvn test -Dtest=TestTezTPCDS30TBPerfCliDriver
```

### DIFF
--- a/ql/src/test/queries/clientpositive/cbo_self_join_ambiguous_alias_spool.q
+++ b/ql/src/test/queries/clientpositive/cbo_self_join_ambiguous_alias_spool.q
@@ -1,5 +1,6 @@
 create table t1 (key int, value int);
 
+set hive.optimize.cte.suggester.type=CBO;
 set hive.optimize.cte.materialize.threshold=1;
 set hive.optimize.cte.suggester.class=org.apache.hadoop.hive.ql.optimizer.calcite.CommonTableExpressionIdentitySuggester;
 set hive.optimize.cte.materialize.full.aggregate.only=false;


zabetak:
This is still relevant don't close it bot!

itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CliConfigs.java
```diff
@@ -349,7 +351,11 @@ public TPCDSCteCliConfig() {
       setClusterType(MiniClusterType.LLAP_LOCAL);
       setMetastoreType("postgres.tpcds");
       // At the moment only makes sense to check CBO plans
+      Set<Integer> skipQueries = ImmutableSet.of(64); // Skipped due to HIVE-29249
```
deniskuzZ: maybe add TODO: comment.
would the query 64 fail or return invalid result with CBO cte suggester? do we test AST as well?

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CommonTableExpressionPrintSuggester.java
```diff
@@ -35,10 +36,11 @@ public class CommonTableExpressionPrintSuggester implements CommonTableExpressio
   @Override
   public List<RelNode> suggest(final RelNode input, final Configuration configuration) {
     List<RelNode> result = internal.suggest(input, configuration);
-    // Ensure CTEs are printed deterministically to avoid test flakiness
-    result.stream().map(RelOptUtil::toString).sorted()
-        .forEach(cte -> SessionState.getConsole().printInfo("CTE Suggestion:\n" + cte, false));
-    return result;
+    // Ensure CTEs are printed and returned deterministically to avoid test flakiness
+    return result.stream()
```
deniskuzZ: why not apply sort in result var initialization and then do for-each print?

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query64.q.out
```diff
@@ -1,167 +0,0 @@
-CTE Suggestion:
```
deniskuzZ, line 1: can we fallback to AST here temporarily?

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query11.q.out
```diff
@@ -34,58 +34,50 @@ HiveSortLimit(sort0=[$0], sort1=[$1], sort2=[$2], sort3=[$3], dir0=[ASC], dir1=[
...
-                      HiveJoin(condition=[=($1, $3)], joinType=[inner], algorithm=[none], cost=[not available])
+                      HiveTableSpool(table=[[cte, cte_suggestion_2]])
                         HiveProject(ss_customer_sk=[$0], ss_sold_date_sk=[$3], $f8=[-($2, $1)])
                           HiveFilter(condition=[AND(IS NOT NULL($0), IS NOT NULL($3))])
                             HiveProject(ss_customer_sk=[$2], ss_ext_discount_amt=[$13], ss_ext_list_price=[$16], ss_sold_date_sk=[$22])
                               HiveTableScan(table=[[default, store_sales]], table:alias=[store_sales])
+                      HiveTableSpool(table=[[cte, cte_suggestion_1]])
                         HiveProject(d_date_sk=[$0])
-                          HiveFilter(condition=[=($1, 1999)])
+                          HiveFilter(condition=[=($1, 2000)])
```
deniskuzZ, line 44: filter seems different, but there is also condition=[=($1, 2000), so it may be ok

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query23.q.out
```diff
@@ -67,7 +68,7 @@ HiveProject(_c0=[$0])
                               HiveFilter(condition=[IS NOT NULL($22)])
                                 HiveTableScan(table=[[default, store_sales]], table:alias=[store_sales])
                             HiveProject(d_date_sk=[$0], d_date=[$2])
-                              HiveFilter(condition=[IN($6, 1999, 2000, 2001, 2002)])
+                              HiveFilter(condition=[OR(=($6, 1999), =($6, 2000), =($6, 2001), =($6, 2002))])
```
deniskuzZ, line 71: why OR is not rewritten with IN here?

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query23.q.out
```diff
@@ -90,7 +91,7 @@ HiveProject(_c0=[$0])
                                 HiveFilter(condition=[AND(IS NOT NULL($2), IS NOT NULL($22))])
                                   HiveTableScan(table=[[default, store_sales]], table:alias=[store_sales])
                               HiveProject(d_date_sk=[$0])
-                                HiveFilter(condition=[IN($6, 1999, 2000, 2001, 2002)])
+                                HiveFilter(condition=[OR(=($6, 1999), =($6, 2000), =($6, 2001), =($6, 2002))])
```
deniskuzZ, line 94: same

zabetak: Missed optimization due to https://issues.apache.org/jira/browse/HIVE-29469

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query23.q.out
zabetak: Missed optimization due to https://issues.apache.org/jira/browse/HIVE-29469

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query11.q.out
zabetak: The lines in the diff are misaligned/misleading (in IntelliJ the differences are easier to read). 

Line 57 on the left corresponds to Line 57 on the right. 
Line 42 on the left corresponds to Line 44 on the right.

The filters are identical so there is no issue here.

ql/src/java/org/apache/hadoop/hive/ql/optimizer/calcite/CommonTableExpressionPrintSuggester.java
zabetak: Applied the suggestion in https://github.com/apache/hive/pull/6092/commits/9709e8ba2b5508edb7c8a91ff68ace6a55b41cde

itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CliConfigs.java
zabetak: Added a TODO in https://github.com/apache/hive/pull/6092/commits/2767dbb5e61f9aa76edd03e667fdf23b720be9d3

I also updated the JIRA reference to point to HIVE-29466. The latter was easy to fix so raised a PR yesterday and if it is merged before this patch we won't have to skip anything.

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query64.q.out
zabetak: Since there is a PR ready for review about q64 there is no point in using a fallback.

itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CliConfigs.java
zabetak: The query64 simply fails it does not generate a wrong result.

zabetak:
@deniskuzZ Many thanks for the review. I believe I addressed all your comments so please take another look once you get the chance.

> HiveConf default AST, but we test with CBO, is that intentional?

Regarding the test coverage for `hive.optimize.cte.suggester.type` I kept the default to AST to avoid changes in behavior. In addition CBO is an experimental feature so not ready yet for being the default.

The AST is the default value and is tested as part of the main plan regression suite (TezTPCDS30TBCliConfig) as well as everywhere else where the config is not overridden.

The CBO is an experimental feature, and the main reason that we introduced a separate CTE suite (TPCDSCteCliConfig).

itests/util/src/main/java/org/apache/hadoop/hive/cli/control/CliConfigs.java
zabetak: HIVE-29466 has been fixed and merged so removed the TODO and enabled q64.

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query64.q.out
zabetak: Q64 now works so this conversation is obsolete.

ql/src/test/results/clientpositive/perf/tpcds30tb/cte/cbo_query80.q.out
```diff
@@ -80,14 +78,10 @@ HiveSortLimit(sort0=[$0], sort1=[$1], dir0=[ASC], dir1=[ASC], fetch=[100])
...
-                          HiveFilter(condition=[>($5, 50:DECIMAL(2, 0))])
-                            HiveTableScan(table=[[default, item]], table:alias=[item])
-                      HiveProject(p_promo_sk=[$0])
-                        HiveFilter(condition=[=($11, _UTF-16LE'N')])
-                          HiveTableScan(table=[[default, promotion]], table:alias=[promotion])
+                        HiveTableScan(table=[[cte, cte_suggestion_1]], table:alias=[cte_suggestion_1])
+                      HiveTableScan(table=[[cte, cte_suggestion_2]], table:alias=[cte_suggestion_2])
                     HiveProject(d_date_sk=[$0])
-                      HiveFilter(condition=[BETWEEN(false, CAST($2):TIMESTAMP(9), 1998-08-04 00:00:00:TIMESTAMP(9), 1998-09-03 00:00:00:TIMESTAMP(9))])
+                      HiveFilter(condition=[AND(<=(1998-08-04 00:00:00, CAST($2):TIMESTAMP(9)), <=(CAST($2):TIMESTAMP(9), 1998-09-03 00:00:00))])
```
deniskuzZ, line 84: is this also related to https://issues.apache.org/jira/browse/HIVE-29469?

zabetak: Yes, exactly!

zabetak:
Thanks for the review @deniskuzZ ! Much appreciated.

PR #6093, opened 2025-09-22
HIVE-26493: Hive throws error when querying a view that was created successfully

mdayakar:
HIVE-26493: Hive throws error when querying a view that was created successfully

While creating view, if column names are specified and the view definition is enclosed within the brackets then view creation is ok but while selecting the data from view, as here user provided column names the view definition is becoming like `select col1 as c1 from ((select col1 from tab)) v0`, here view definition is already enclosed within brackets and we also adding an extra brackets again, which is causing compile failure. 
So as a part of fix, if view definition is already enclosed within brackets then not adding the brackets again.

Without fix, the data from view cannot be selected.

No

Using q tests
_mvn -Dtest=TestMiniLlapLocalCliDriver -Dqfile=viewdef_within_brackets.q -pl itests/qtest -Pitests_

### DIFF
--- a/ql/src/java/org/apache/hadoop/hive/ql/ddl/view/create/CreateViewAnalyzer.java
+++ b/ql/src/java/org/apache/hadoop/hive/ql/ddl/view/create/CreateViewAnalyzer.java
@@ -113,6 +113,11 @@ public class CreateViewAnalyzer extends AbstractCreateViewAnalyzer {
     String expandedText = ctx.getTokenRewriteStream().toString(select.getTokenStartIndex(), select.getTokenStopIndex());
 
     if (imposedSchema != null) {
+      boolean isBracketAvailable = false;
+      String trimmedExpandedText = expandedText.trim();
+      if (trimmedExpandedText.startsWith("(") && trimmedExpandedText.endsWith(")")) {
+        isBracketAvailable = true;
+      }
       // Merge the names from the imposed schema into the types from the derived schema.
       StringBuilder sb = new StringBuilder();
       sb.append("SELECT ");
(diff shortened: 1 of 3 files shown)


ql/src/java/org/apache/hadoop/hive/ql/ddl/view/create/CreateViewAnalyzer.java
```diff
@@ -136,9 +141,14 @@ private String getExpandedText(List<FieldSchema> imposedSchema, ASTNode select,
         // We don't currently allow imposition of a type
         fieldSchema.setComment(imposedSchema.get(i).getComment());
       }
-      sb.append(" FROM (");
+      sb.append(" FROM ");
+      if (!isBracketAvailable) {
+        sb.append("(");
+      }
```
kasakrisz: How about adding more brackets?
```
create view vw_x (b) as (((select a from tbl_x)));
```

mdayakar: Currently in this case it will fail. So now fixed this scenario by removing all the brackets attached with view definition and adds a single bracket finally.

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/create/CreateViewAnalyzer.java
```diff
@@ -113,6 +113,12 @@ private String getExpandedText(List<FieldSchema> imposedSchema, ASTNode select,
     String expandedText = ctx.getTokenRewriteStream().toString(select.getTokenStartIndex(), select.getTokenStopIndex());
 
     if (imposedSchema != null) {
```
kasakrisz, line 121: Storing the extra brackets in the expanded text is not necessary even when column names are not specified so trimming them could be moved before this if statement.

kasakrisz (approved):

ql/src/java/org/apache/hadoop/hive/ql/ddl/view/create/CreateViewAnalyzer.java
mdayakar: Here issue comes when only column names are specified for the view otherwise there will not be any issue. 
For example below create view query
`create view vw_z as (
    (  (select a, b from tbl_x ) )
    );`
the expanded test is converted as `
    (  (select a, b from tbl_x ) )
    )` and compiling this statement doesn't create any issue.
    
 But for the below query
    `create view vw_y (col_b) as (
    ( ( select b from tbl_x ) )
    );`
    the expanded text is converted as `SELECT b AS col_b FROM (
    ( ( select b from tbl_x ) )
    ) vw_y` this results into compilation issue.
    
   So here handled only the case when issue is coming, other scenarios will work as previous.

mdayakar: Changed the code accordingly.

PR #6094, opened 2025-09-22
HIVE-29220: Iceberg: Enable write operations with REST Catalog HMS Client

difin:

Fixes to make Insert operations work with HMS REST Catalog Client.

Insert didn't work with HMS REST Catalog Client.

No

Updated `iceberg_rest_catalog.q` to perform Insert and Select operations.

### DIFF
--- a/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java
+++ b/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg.hive;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -32,6 +33,7 @@ import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.EnvironmentContext;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.SerDeInfo;
+import org.apache.hadoop.hive.metastore.api.SkewedInfo;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.metastore.utils.MetaStoreUtils;
@@ -157,6 +159,7 @@ public class MetastoreUtil {
     result.setSerdeInfo(getHiveSerdeInfo());
     result.setLocation(table.location());
     result.setParameters(Maps.newHashMap());
+    result.setSkewedInfo(new SkewedInfo(Collections.emptyList(), Collections.emptyList(), Collections.emptyMap()));
     return result;
   }
 
--- a/iceberg/iceberg-handler/src/test/queries/positive/iceberg_rest_catalog.q
+++ b/iceberg/iceberg-handler/src/test/queries/positive/iceberg_rest_catalog.q
@@ -3,6 +3,8 @@
 --! qt:replace:/(\s+neededVirtualColumns:\s)(.*)/$1#Masked#/
 -- Mask random uuid
 --! qt:replace:/(\s+'uuid'=')\S+('\s*)/$1#Masked#$2/
+-- Mask random uuid
+--! qt:replace:/(\s+uuid\s+)\S+(\s*)/$1#Masked#$2/
 -- Mask a random snapshot id
 --! qt:replace:/(\s+current-snapshot-id\s+)\S+(\s*)/$1#Masked#/
 -- Mask added file size
@@ -20,6 +22,7 @@
 -- Mask iceberg version
 --! qt:replace:/(\S\"iceberg-version\\\":\\\")(\w+\s\w+\s\d+\.\d+\.\d+\s\(\w+\s\w+\))(\\\")/$1#Masked#$3/
 
+set hive.stats.autogather=false;
 set metastore.client.impl=org.apache.iceberg.hive.client.HiveRESTCatalogClient;
 set metastore.catalog.default=ice01;
 set iceberg.catalog.ice01.type=rest;
@@ -30,7 +33,10 @@ set iceberg.catalog.ice01.type=rest;
 create database ice_rest;
 use ice_rest;
 
---! Creating table without catalog name in table properties
+-----------------------------------------------------------------------------
+--! Creating a table without a catalog name in table properties
+-----------------------------------------------------------------------------
+
 create table ice_orc1 (
     first_name string, 
     last_name string,
@@ -40,7 +46,10 @@ create table ice_orc1 (
 partitioned by (company_id bigint)
 stored by iceberg stored as orc;
 
---! Creating table with a valid catalog name in table properties
+-----------------------------------------------------------------------------
+--! Creating  table with a valid catalog name in table properties
+-----------------------------------------------------------------------------
+
 create table ice_orc2 (
     first_name string, 
     last_name string,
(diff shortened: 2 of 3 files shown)


iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java
```diff
@@ -157,6 +159,8 @@ private static StorageDescriptor getHiveStorageDescriptor(org.apache.iceberg.Tab
     result.setSerdeInfo(getHiveSerdeInfo());
     result.setLocation(table.location());
     result.setParameters(Maps.newHashMap());
+    result.setSkewedInfo(new SkewedInfo(new SkewedInfo(Collections.emptyList(),
+        Collections.emptyList(), Collections.emptyMap())));
```
okumin: Taking a glance, we don't need to wrap SkewedInfo with SkewedInfo

P.S. I verified the SkewedInfo is likely necessary. NPE would happen without this change.

```
java.lang.NullPointerException: Cannot invoke "org.apache.hadoop.hive.metastore.api.SkewedInfo.getSkewedColNames()" because the return value of "org.apache.hadoop.hive.metastore.api.StorageDescriptor.getSkewedInfo()" is null
	at org.apache.hadoop.hive.ql.metadata.Partition.getSkewedColNames(Partition.java:581)
	at org.apache.hadoop.hive.ql.parse.SemanticAnalyzer.genFileSinkPlan(SemanticAnalyzer.java:7769)
	at org.apache.hadoop.hive.ql.parse.SemanticAnalyzer.genPostGroupByBodyPlan(SemanticAnalyzer.java:11676)
	at org.apache.hadoop.hive.ql.parse.SemanticAnalyzer.genBodyPlan(SemanticAnalyzer.java:11539)
	at org.apache.hadoop.hive.ql.parse.SemanticAnalyzer.genPlan(SemanticAnalyzer.java:12474)
	at org.apache.hadoop.hive.ql.parse.SemanticAnalyzer.genPlan(SemanticAnalyzer.java:12340)
```

itests/qtest-iceberg/src/test/java/org/apache/hadoop/hive/cli/TestIcebergRESTCatalogLlapLocalCliDriver.java
```diff
@@ -18,101 +18,159 @@
...
 public class TestIcebergRESTCatalogLlapLocalCliDriver {
 
-  private static final Logger LOG = LoggerFactory.getLogger(
-      org.apache.hadoop.hive.cli.TestIcebergRESTCatalogLlapLocalCliDriver.class);
+  private static final String GRAVITINO_CONF_TEMPLATE = "gravitino-h2-test-template.conf";
+  private static final String GRAVITINO_CONF = "gravitino-h2-test.conf";
   private static final String CATALOG_NAME = "ice01";
   private static final CliAdapter adapter = new CliConfigs.TestIcebergRESTCatalogLlapLocalCliDriver().getCliAdapter();
+  private static final DockerImageName GRAVITINO_IMAGE =
+      DockerImageName.parse("apache/gravitino-iceberg-rest:1.0.0-rc3");
```
okumin: Just curious if the REST catalog of HMS does not work or not

itests/qtest-iceberg/src/test/resources/gravitino-h2-test-template.conf
```diff
@@ -0,0 +1,14 @@
+gravitino.iceberg-rest.httpPort = 9001
+
+# --- Iceberg REST Catalog Backend (set to JDBC) ---
+gravitino.iceberg-rest.catalog-backend = jdbc
```
okumin: The default backend is likely `memory`. Can't we use it? The `memory` mode sounds like we don't need an additional dependency.
https://gravitino.apache.org/docs/0.9.1/iceberg-rest-service

difin: I tried that initially and with memory mode it didn't work - as if commit after insert didn't happen, the inserted records were missing on select, switching the backend mode to h2/jdbc inside gravitino solved that.

itests/qtest-iceberg/src/test/java/org/apache/hadoop/hive/cli/TestIcebergRESTCatalogLlapLocalCliDriver.java
difin: Yes, it didn't work with REST catalog of HMS, insert and select succeeded, but select was empty. As far as I know REST catalog of HMS wasn't tested with write operations, and it looks like it doesn't handle them correctly at this time.

okumin: The REST API supports all required endpoints. I'd like you to recheck it and create a JIRA ticket if the issue persists. At least, on my laptop, the following change and integration tests succeeded.
https://github.com/okumin/hive/commit/367f68e7c49f041310445d4cb005a4255ef86c57

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/client/HiveRESTCatalogClient.java
```diff
@@ -228,4 +231,9 @@ public void dropDatabase(DropDatabaseRequest req) {
     validateCurrentCatalog(req.getCatalogName());
     restCatalog.dropNamespace(Namespace.of(req.getName()));
   }
+
+  @Override
+  public LockResponse lock(LockRequest request) {
+    return new LockResponse(request.getTxnid(), LockState.ACQUIRED);
+  }
```
okumin: What would happen if we implement this? My local test succeeded without this change

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java
difin: Thanks for verifying. I added the nest `SkewedInfo()` because it didn't work otherwise.

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/client/HiveRESTCatalogClient.java
difin: I retested on my local and it worked without it. Very strange, because before it was failing on an attempt to acquire a lock. Maybe on my local it was failing with gravitino, not with HMS REST server.

itests/qtest-iceberg/src/test/java/org/apache/hadoop/hive/cli/TestIcebergRESTCatalogLlapLocalCliDriver.java
difin: I rechecked, on my local it also succeeded with HMS REST server. Not sure why it didn't work before.
Thanks for checking that, @okumin!

iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java
okumin: I replaced this change with the following one and verified the local test succeeded

```java
result.setSkewedInfo(new SkewedInfo(Collections.emptyList(), Collections.emptyList(), Collections.emptyMap()));
```

okumin: @difin
Could you recheck it on your environment just in case?

difin: This worked on my local too, fixed.
Thanks, @okumin !

iceberg/iceberg-handler/src/test/queries/positive/iceberg_rest_catalog.q
```diff
@@ -20,6 +22,7 @@
 -- Mask iceberg version
 --! qt:replace:/(\S\"iceberg-version\\\":\\\")(\w+\s\w+\s\d+\.\d+\.\d+\s\(\w+\s\w+\))(\\\")/$1#Masked#$3/
 
+set hive.stats.autogather=false;
```
okumin, line 25: Not mandatory yet. Just curious why this does not work with autogather

okumin (approved):
LGTM!

iceberg/iceberg-handler/src/test/queries/positive/iceberg_rest_catalog.q
difin: I don't remember the error message, but without stats autogather it was failing in query analyzer due to missing table and column statistics.

PR #4403, opened 2023-06-10
HIVE-27317: Temporary (local) session files cleanup improvements

sercanCyberVision:
Unit test in the initial solution caused regression to the successors unit tests as it creates scratch dirs without all writable permission. So the solution https://github.com/apache/hive/pull/4293 has been reverted here https://github.com/apache/hive/pull/4399.

### What changes were proposed in this pull request?
Along with original solution, the unit test has been modified. Now, both scratch dirs have all writable permissions.
```
> ll ql/target/tmp
total 24
drwxrwxrwx  3 sercan sercan 4096 Jun 10 17:32 localscratchdir/
drwxrwxrwx  3 sercan sercan 4096 Jun 10 17:32 scratchdir/
```

### How was this patch tested?
Unit test.

@deniskuzZ Could you please review?

PR #6099, opened 2025-09-24
HIVE-29225: Premature deletion of scratch directories during output streaming

sercanCyberVision:
**ROOT-CAUSE:**
Please see https://issues.apache.org/jira/browse/HIVE-29225 for the detailed problem description and steps to reproduce. In short, scratch directories may be deleted prematurely while Hive is still streaming output to the client.

**SOLUTION:**
A new configuration property has been introduced:
```
hive.scratchdir.cleanup.grace.period.hours
```
This property defines a grace period to prevent cleanup of scratch directories that have been modified within the specified time window.

- Default value is `0h` (disabled by default).
- When Hive finishes output streaming, it updates the scratch directory’s last modified time.
- The cleanup logic now checks this timestamp and skips deletion if the directory was updated within the grace period.

This ensures directories are not removed prematurely while still in use.

**TESTING:**
Added a dedicated unit test to validate the new functionality:
the-test

Verified that all existing tests in `TestClearDanglingScratchDir` continue to pass:
scratch-dir-tests

### DIFF
--- a/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
+++ b/common/src/java/org/apache/hadoop/hive/conf/HiveConf.java
@@ -3822,6 +3822,11 @@ public class HiveConf extends Configuration {
         "To cleanup the Hive scratchdir when starting the Hive Server"),
     HIVE_SCRATCH_DIR_LOCK("hive.scratchdir.lock", false,
         "To hold a lock file in scratchdir to prevent to be removed by cleardanglingscratchdir"),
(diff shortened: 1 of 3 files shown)


sercanCyberVision:
Hello @deniskuzZ ,
You previously reviewed and approved the `ClearDanglingScratchDir.java` changes in PR #4403. 
Could you please take a look at this PR as well?

sercanCyberVision:
Hi @difin,
Can you please review this PR?

